PR #6940, opened 2022-02-12
add test for multiple text attachments but different charsets

yasserzamani:
monitors that multiple attachments with different charsets marshal and unmarshal doesn't damage non-ascii characters. Currently it doesn't.

### DIFF
--- a/components/camel-mail/src/test/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormatTest.java
+++ b/components/camel-mail/src/test/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormatTest.java
@@ -149,19 +149,19 @@ public class MimeMultipartDataFormatTest extends CamelTestSupport {
     @Test
     public void roundtripWithMultipleTextAttachmentsButDifferentCharsets() throws IOException {
         String att1ContentType = "text/plain; charset=ISO-8859-1";
-        String att1Text = new String("empfänger1".getBytes("ISO-8859-1"), "ISO-8859-1");
-        String att1FileName = "empfänger1";
+        String att1Text = new String("empf\u00e4nger1".getBytes("ISO-8859-1"), "ISO-8859-1");
+        String att1FileName = "empf\u00e4nger1";
         String att2ContentType = "text/plain; charset=ISO-8859-15";
-        String att2Text = new String("empfänger15".getBytes("ISO-8859-15"), "ISO-8859-15");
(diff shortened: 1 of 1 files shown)


components/camel-mail/src/test/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormatTest.java
```diff
@@ -146,6 +146,62 @@ public void roundtripWithTextAttachmentsAndSpecialCharacters() throws IOExceptio
...
     }
 
+    @Test
+    public void roundtripWithMultipleTextAttachmentsButDifferentCharsets() throws IOException {
+        String att1ContentType = "text/plain; charset=ISO-8859-1";
+        String att1Text = new String("empfänger1".getBytes("ISO-8859-1"), "ISO-8859-1");
+        String att1FileName = "empfänger1";
+        String att2ContentType = "text/plain; charset=ISO-8859-15";
+        String att2Text = new String("empfänger15".getBytes("ISO-8859-15"), "ISO-8859-15");
+        String att2FileName = "empfänger15";
```
davsclaus, line 156: When you have string text that uses non-ascii then you should unicode escape those, eg "aaa\u1234\u5555"

davsclaus (changes requested):

components/camel-mail/src/test/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormatTest.java
yasserzamani: Thanks! fixed

BTW I have same situation in my job which damages German characters. I though it's Camel issue. So I wrote this test. But I realized it's my fault. I MUST NOT convert body to string after `.marshal().mimemultipart()` where there're multiple different charsets in body and attachments. I MUST keep it as byte array when I push it into JMS. Then on other side when I read it as a byte array then Camel can unmarshal it correctly.

So I just wanted to not throw away this test. Please feel free to use or close it :)

davsclaus (approved):

components/camel-mail/src/test/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormatTest.java
davsclaus: Thanks for the test, we can surely use this

PR #6943, opened 2022-02-14
Add PMD scanning for PR code

orpiske:

### DIFF
--- a/.github/workflows/pmd-analysis.yml
+++ b/.github/workflows/pmd-analysis.yml
@@ -1,0 +1,26 @@
+name: pmd
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  pmd-code-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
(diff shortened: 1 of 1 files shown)


oscerd (approved):
I really hate PMD, but it's just me.

orpiske:
> I really hate PMD, but it's just me.

Yeah, there can be a lot of false negatives if we don't tune it accordingly. I'll definitely propose it with a ruleset that trying to balance this. 

BTW, the PR as it is cannot be accepted (I was testing if PMD action was whitelisted, but it's not).

orpiske:
> Eventually, if it won't be whitelisted, we just need to add the action as git submodule, like the backport action. So no problem at all.

Oh, nice, I was reading about this on the wiki and it seems a reasonable way to do it too. I will open a ticket asking for Infra to take a look and see if they can whitelist it.

orpiske:
Folks at Apache Infra approved the PMD action (INFRA-22883). Re-opening ...

orpiske:
I am merging this one so I can test it and tweak it during the week based on what we receive/review/etc. In case of any problems, I'll fix it and tune it.

PR #6946, opened 2022-02-15
Force some dependencies into the classpath

orpiske:
Removing the secretsmanager in change
(317a2baa00c666c27fe51f3996989a42607192fe) resulted in these
dependencies being ignored in the component / endpoint DSL classpath.
This forces them back into it.

### DIFF
--- a/dsl/camel-componentdsl/pom.xml
+++ b/dsl/camel-componentdsl/pom.xml
@@ -90,6 +90,26 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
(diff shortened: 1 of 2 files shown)


orpiske:
This is kinda of an ugly fix @davsclaus @oscerd. I am not sure if we have a better way to solve it ...

oscerd:
I'm moving the Secrets Manager aws properties Source in the aws secrets manager component. This should fix the situation a bit.

oscerd (changes requested):
Let's wait for move of the properties source and then check again.

orpiske:
> Let's wait for move of the properties source and then check again.

Ahhh, nice! Yeah, let's wait for that ... I was suspecting this was not a good approach for it. Thanks for the heads up!

oscerd:
It should be fixed with https://github.com/apache/camel/commit/cfa5189d83e1eff0afc738988f827f0603598f63 and https://github.com/apache/camel/commit/cfa5189d83e1eff0afc738988f827f0603598f63

oscerd:
cc @orpiske

orpiske:
Thanks @oscerd! Let's close this.

PR #6955, opened 2022-02-15
(chores) ci: use suggestion from SonarCloud to improve analysis time

orpiske:


davsclaus (approved):

orpiske:
I did not see any improvement from using this flag on my own CI, but let's see how it works on the Apache one.

PR #6956, opened 2022-02-15
[CAMEL-17651] fix reload strategy and related doc

JPMoresmau:
Running -Psourcecheck tells me `AbstractExchangeTest.java' has not been previously formatted.  Please format file and commit before running validation`, I have no idea what this mean, if it hasn't been previously formatted, why should it be formatted now? Contributing document only says to run this but gives no information about how to format, since the build fails without any helpful hint on what formatting was wrong. This is my first PR in Camel, please advise.

I tried to write a unit test for both bugs I'm fixing and reused the fake context created in the next test, not sure if there's a better way, since there is not many tests in this module.


davsclaus (approved):

oscerd (approved):

JPMoresmau:
Where do I find camel-checkstyle.xml (referenced in the build failure message) to ensure I pass the checkstyle validation?

JPMoresmau:
OK, in /buildingtools/src/main/resources/, github search failed me...

davsclaus:
next time, then unit tests should be in camel-core/src/test

PR #6960, opened 2022-02-16
CAMEL-17658 camel-debezium: additionalProperties are trimmed

JiriOndrusek:
Issue: https://issues.apache.org/jira/browse/CAMEL-17658

Problem is caused not entirely by debezium component. Automatic properties configuration can "trim" property in specific cases.

- Let say, that we have a property **map.A.B=value**, where  **map** is an URI property defined as ` @UriParam(label = "common", prefix = "map.", multiValue = true,`. Automatic configuration initialization will detect a HashMap object as a **map** property, but then it executes "set property A=value into HashMap (and property B is ignored)"
- I'm trying to change mechanism to not ignore the remaining part of the property name and execute "set property A.B=value into HashMap"

I'm not sure whether this change could be approved, Therefore this is only draft and I'd like to see whether there will be some failing tests.

In case that this approach is not desired or potentially dangerous, I will fix it directly on debezium components - which is safe.

### DIFF
--- a/components/camel-debezium/camel-debezium-mysql/src/test/java/org/apache/camel/component/debezium/DebeziumMySqlComponentTest.java
+++ b/components/camel-debezium/camel-debezium-mysql/src/test/java/org/apache/camel/component/debezium/DebeziumMySqlComponentTest.java
@@ -39,6 +39,7 @@ public class DebeziumMySqlComponentTest {
         params.put("databaseServerName", "test");
         params.put("databaseServerId", "1234");
         params.put("databaseHistoryFileFilename", "/db_history_file_test");
+        params.put("additionalProperties.database.connectionTimeZone", "CET");
 
         final String remaining = "test_name";
         final String uri = "debezium:mysql?name=test_name&offsetStorageFileName=/test&"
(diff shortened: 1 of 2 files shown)


davsclaus:
Ah so are you saying the key in the map is a string with a dot?

As if you did this in regular Java

Map map = ...
map.put("A.B", "myValue")

JiriOndrusek:
Yes, it is exacty the case. Value in the map is with dot (` "&additionalProperties.database.connectionTimeZone=CET"`). See https://github.com/apache/camel-quarkus/issues/3488.

JiriOndrusek:
The reason why it should work is fact, that it works in "non-specific debezium parent component (see https://github.com/apache/camel/blob/main/components/camel-debezium/camel-debezium-common/camel-debezium-common-component/src/test/java/org/apache/camel/component/debezium/DebeziumComponentTest.java#L48)"

The difference is caused by the fact, that thois parent component doesn't define `@UriParam(label = "common", prefix = "map.", multiValue = true`, so the **map** is not use configurer and the value is not trimmed. Therefore I think it should work in the same way with configurer.

JiriOndrusek:
@davsclaus Do you mean the test for "non-specific debezum component" https://github.com/apache/camel/blob/main/components/camel-debezium/camel-debezium-common/camel-debezium-common-component/src/test/java/org/apache/camel/component/debezium/DebeziumComponentTest.java#L48 ?

In that case I have no argument that we want this behavior during automatic configuration.

I can fix problem in component itself (not changing `PropertyBindingSupport`). This kind of fix is safe and easy, do you prefer it?

davsclaus:
Yeah its a tricky code the property binding - let me try your patch.

davsclaus:
Your code is correct, the newName should include the remainder parts - this is the first time we hit this because of the dot keys in that component.

davsclaus (approved):

PR #6962, opened 2022-02-16
CAMEL-17308: Remove "verb" as an allowed CAMEL REST YAML element.

klease:
This also affects the generated XML model which also will only accept
the XML elements for different methods such as get, put and post.
Remove test cases in RestTest.groovy using verb.

### DIFF
--- a/core/camel-core-model/src/generated/resources/org/apache/camel/model/rest/jaxb.index
+++ b/core/camel-core-model/src/generated/resources/org/apache/camel/model/rest/jaxb.index
@@ -26,4 +26,3 @@ RestSecurityOAuth2
 RestSecurityOpenIdConnect
 RestsDefinition
 SecurityDefinition
-VerbDefinition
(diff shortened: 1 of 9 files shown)


klease:
The generated YAML model now contains a "verbs" definition (as an array of VerbDefinition which is no longer defined in the Yaml). This doesn't seem consistent but doesn't prevent the test cases using the method names from running.

davsclaus:
Can you add a note in the 3.16 upgrade doc, about the yaml-dsl change (see root/docs folder).

klease:
The generated YAML model now contains a "verbs" definition (as an array of VerbDefinition which is no longer defined in the Yaml). This doesn't seem consistent but doesn't prevent the test cases using the method names from running.

> Can you add a note in the 3.16 upgrade doc, about the yaml-dsl change (see root/docs folder).

Done

davsclaus (approved):

PR #6982, opened 2022-02-19
Vault

davsclaus:


davsclaus:
@oscerd here is a PR that adds Vault configuration to camel-core / camel-main so its configurable for all runtimes. It needs maybe a little bit more polish and javadoc.

Then in the aws-secrets-manager you can then get access to the configuration via camel context instead of the custom code.

For spring boot we can then add auto configuration too, it would be similar to how its done of the other configurations we have.

core/camel-api/src/main/java/org/apache/camel/spi/VaultConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.apache.camel.spi;
+
+/**
+ * Configuration for access to Vaults.
+ */
+public class VaultConfiguration {
+
+    @Metadata(secret = true)
```
oscerd, line 24: Does it make sense to have specific VaultConfiguration for each provider? I mean AWSVaultConfiguration, GCPVaultConfiguration etc

oscerd (approved):
LGTM, I think the only missing bit is remapping the properties name in the aws-secrets-manager component, so they will be the same as VaultConfiguration

core/camel-api/src/main/java/org/apache/camel/spi/VaultConfiguration.java
davsclaus: Yeah maybe but lets start with this. Its a bit more tricky with different classes

davsclaus: You would then have to go one level down on the configuation for aws / gcp etc 

VaultConfiguration cfg = ...

cfg.aws().getXXX

Or that we have to specify which sub class to get

AwsVaultConfiguration cfg = camelContext.getVaultConfiguration(AwsVaultConfiguration.class);

oscerd: I think it would be a good solution to subclass, in the end we shouldn't introduce new dependency, just some properties for connecting and defining the pointers. Btw LGTM. We could start from here.

oscerd (approved):

core/camel-api/src/main/java/org/apache/camel/spi/VaultConfiguration.java
davsclaus: Okay I have a PR on the way with specific configuration per provider - AWS, Google, Azure etc

oscerd (approved):

davsclaus:
TODO: add vault.adoc documentation

oscerd:
We need to add the properties in the documentation. For the moment only env properties are listed. Ok for merging btw

oscerd:
@davsclaus, this should align also the AWSSecretManager Properties Function and we should be fine. I worked on the vault branch and force pushed.

oscerd:
Feel free to merge, if it's fine for you.

oscerd:
I think we are good to go. We need to reflect this on Camel Spring Boot.

PR #6993, opened 2022-02-21
CAMEL-17679: remove references to unsupported metrics types

orpiske:

### DIFF
--- a/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerEndpoint.java
+++ b/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerEndpoint.java
@@ -40,7 +40,7 @@ public class MicrometerEndpoint extends DefaultEndpoint {
 
     protected MeterRegistry registry;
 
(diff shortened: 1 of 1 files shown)


orpiske:
Hey @oscerd do you happen to know if this is enough to override the auto-generated documentation? I wonder if I need to create a setter for this and document it there ...

oscerd (approved):
This should work.

But you could also add enums to https://github.com/apache/camel/blob/fda3471595b63699ac00ab00809a760bb1a76b59/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerEndpoint.java#L43

in the uriPath annotation.

orpiske:
> This should work.
> 
> But you could also add enums to
> 
> https://github.com/apache/camel/blob/fda3471595b63699ac00ab00809a760bb1a76b59/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerEndpoint.java#L43
> 
> in the uriPath annotation.

Your suggestion is much better and cleaner than what I did. I will do it as you suggested. Thanks!

PR #7004, opened 2022-02-21
Attempt to force executions to be synchronized if JTA trans. is present

JiriOndrusek:
Issue https://issues.apache.org/jira/browse/CAMEL-17474

Problem is caused by the fact, that JTA transactions requires synchronous execution (see [comment](https://github.com/apache/camel/blob/main/components/camel-jta/src/main/java/org/apache/camel/jta/TransactionErrorHandler.java#L91))
There is a [change](https://github.com/apache/camel/commit/961ad0e56e9331e71c386415ec67676e586ea629#diff-592328b1c35a306fa6ba3875f63b872de065e6d973ea89870d579bbab505cfadR214) (a few months ago), which changed default executions to be asynchronous and causes deadlocks.

I added a test created by @zhfeng 

The whole code change is only POC whether this kind of change would help. (it seems to be working successfully)
Mainly I'm not sure about `JtaTransactionPolicy.beforeWrap`. But it is essential to set flag to the route's error handler factory, to make this idea work.

I would appreciate any kind of help/review//response to this PR. (as I'm not even sure whether this is the right idea to implement)

### DIFF
--- a/components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
+++ b/components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
@@ -42,12 +42,12 @@ public class JtaTransactionErrorHandler extends RedeliveryErrorHandler {
                                       JtaTransactionPolicy transactionPolicy,
                                       Predicate retryWhile, ScheduledExecutorService executorService,
                                       LoggingLevel rollbackLoggingLevel,
-                                      Processor onExceptionOccurredProcessor) {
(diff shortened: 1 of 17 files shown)


zhfeng:
Thanks @JiriOndrusek and I made some progress to mark the exchange transacted in ```JtaTransactionErrorHandler```. see https://github.com/zhfeng/apache-camel/blob/transacted_split/components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java#L72-L83 and that could be another solution.

zhfeng:
@JiriOndrusek can you close this PR since we alreday have a fix for camel-jta.

davsclaus:
closing PR

PR #7014, opened 2022-02-22
CAMEL-17474: camel-jta - Mark exchange transacted in JtaTransacionErr…

zhfeng:
…orHandler

My attempt to fix [CAMEL-17474](https://issues.apache.org/jira/browse/CAMEL-17474) and suppress https://github.com/apache/camel/pull/7004

### DIFF
--- a/components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
+++ b/components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
@@ -18,7 +18,9 @@ package org.apache.camel.jta;
 
 import java.util.concurrent.ScheduledExecutorService;
 
+import org.apache.camel.AsyncCallback;
 import org.apache.camel.CamelContext;
+import org.apache.camel.Exchange;
 import org.apache.camel.LoggingLevel;
 import org.apache.camel.Predicate;
 import org.apache.camel.Processor;
@@ -26,6 +28,8 @@ import org.apache.camel.processor.errorhandler.RedeliveryErrorHandler;
 import org.apache.camel.processor.errorhandler.RedeliveryPolicy;
 import org.apache.camel.spi.CamelLogger;
 import org.apache.camel.spi.ErrorHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This error handler does redelivering. If the transaction fails it can be retried if configured to do so. In the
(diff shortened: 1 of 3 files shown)


components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
```diff
@@ -64,4 +69,16 @@ public ErrorHandler clone(Processor output) {
         return answer;
     }
 
+    @Override
+    public boolean process(final Exchange exchange, final AsyncCallback callback) {
+        if (!exchange.isTransacted()) {
+            LOG.debug("mark " + exchange + " transacted");
```
davsclaus, line 75: Can you use logger placeholders: "Mark {} as transacted", exchange

components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
```diff
@@ -64,4 +69,16 @@ public ErrorHandler clone(Processor output) {
...
+            exchange.getUnitOfWork().beginTransactedBy("jta");
+            super.process(exchange, callback);
+            exchange.getUnitOfWork().endTransactedBy("jta");
```
davsclaus, line 78: I think you need to have the endTransacted in a try .. finally block to ensure it will be ended

components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
```diff
@@ -64,4 +69,16 @@ public ErrorHandler clone(Processor output) {
...
+        } else {
+            super.process(exchange, callback);
+        }
+        return false;
```
davsclaus, line 82: This may be an existing "issue" as the returned value should be the output from super.process and not hardcoded to false

components/camel-jta/src/main/java/org/apache/camel/jta/JtaTransactionErrorHandler.java
```diff
@@ -64,4 +69,16 @@ public ErrorHandler clone(Processor output) {
         return answer;
     }
 
+    @Override
+    public boolean process(final Exchange exchange, final AsyncCallback callback) {
+        if (!exchange.isTransacted()) {
+            LOG.debug("mark " + exchange + " transacted");
+            exchange.getUnitOfWork().beginTransactedBy("jta");
```
davsclaus, line 76: Maybe use camel-jta instead of jta

zhfeng:
Thanks @davsclaus  and I've done all the changes in your comments.

davsclaus (approved):

PR #7022, opened 2022-02-22
CAMEL-17694: improve handling of sendToAll header

prtnv:

### DIFF
--- a/components/camel-vertx/camel-vertx-websocket/src/test/java/org/apache/camel/component/vertx/websocket/VertxWebsocketTest.java
+++ b/components/camel-vertx/camel-vertx-websocket/src/test/java/org/apache/camel/component/vertx/websocket/VertxWebsocketTest.java
@@ -163,6 +163,9 @@ public class VertxWebsocketTest extends VertxWebSocketTestSupport {
 
     @Test
     public void testSendToAll() throws Exception {
+        MockEndpoint mockEndpoint = getMockEndpoint("mock:result");
+        mockEndpoint.expectedMessageCount(0);
+
         int expectedResultCount = 5;
         CountDownLatch latch = new CountDownLatch(expectedResultCount);
         List<String> results = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

components/camel-vertx/camel-vertx-websocket/src/test/java/org/apache/camel/component/vertx/websocket/VertxWebsocketTest.java
```diff
@@ -163,6 +163,9 @@ public void testSendWithMultipleConnectionKeys() throws Exception {
 
     @Test
     public void testSendToAll() throws Exception {
+        MockEndpoint mockEndpoint = getMockEndpoint("mock:result");
+        mockEndpoint.expectedMessageCount(0);
```
jamesnetherton, line 167: The expected message count being 0 implies a change in behavior, because prior to this change it would be 1 - which I think is more correct.

jamesnetherton: If I can find some time, I will try to dig a bit deeper into the `sendToAll` bits & compare with how some of the other Camel WS components do things.

prtnv: In addition to sending a message to all connected peers, VertxWebsocketProducer establishes an outgoing connection and also sends a message there. In this case, this outgoing connection was established to the local server. As a result, the message appeared in the mock endpoint. Is this correct behaviour?

jamesnetherton: > As a result, the message arrived to the mock endpoint. Is this correct behaviour?

I think so. But I wanted to compare with how it works in some of the other WS components to be sure.

jamesnetherton: I [compared](https://gist.githubusercontent.com/jamesnetherton/9b0849e48330e184491e281da8c64ac4/raw/6084befd2091dae8e41c6b9b4510e9f1cfffa9cd/SendToAllTest.java) with how `camel-undertow` does things. With `sendToAll` and it only sends the message to the connected peers and the consumer route does not run. So this change would make `vertx-websocket` in line with that behavior.

Still not sure if that's totally correct though. If I do `.to("vertx-websocket:localhost:8080/foo?sendToAll=true")`, then to me it means that I want to send a message to the thing listening on `localhost:8080` and any connected peers.

But maybe we merge this change as-is given that it fixes a specific problem and try to enhance things further in future.

davsclaus: +1

And I also agree with your last example that I want to send to "the thing" listening and connected peers. So maybe we can create a new JIRA about that.

jamesnetherton (approved):

PR #7040, opened 2022-02-25
CAMEL-17474: camel-core - set transacted when copying exchange in mul…

zhfeng:
…ticast processor

another fix in MultiProcessor.

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -910,6 +910,7 @@ public class MulticastProcessor extends AsyncProcessorSupport
         for (Processor processor : processors) {
             // copy exchange, and do not share the unit of work
             Exchange copy = processorExchangeFactory.createCorrelatedCopy(exchange, false);
+            copy.adapt(ExtendedExchange.class).setTransacted(exchange.isTransacted());
 
             if (streamCache != null) {
                 if (index > 0) {
--- a/tests/camel-itest/src/test/java/org/apache/camel/itest/tx/JtaRouteTest.java
+++ b/tests/camel-itest/src/test/java/org/apache/camel/itest/tx/JtaRouteTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.itest.tx;
 
+import java.util.Arrays;
+
 import org.apache.camel.AggregationStrategy;
 import org.apache.camel.EndpointInject;
 import org.apache.camel.Exchange;
@@ -25,6 +27,7 @@ import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.cdi.transaction.RequiresNewJtaTransactionPolicy;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.test.junit5.CamelTestSupport;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class JtaRouteTest extends CamelTestSupport {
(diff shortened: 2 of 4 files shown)


core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
```diff
@@ -910,6 +910,7 @@ protected Integer getExchangeIndex(Exchange exchange) {
         for (Processor processor : processors) {
             // copy exchange, and do not share the unit of work
             Exchange copy = processorExchangeFactory.createCorrelatedCopy(exchange, false);
+            copy.adapt(ExtendedExchange.class).setTransacted(exchange.isTransacted());
```
davsclaus, line 913: There would be other EIPs such as split etc that would need something similar. Maybe we can set the transacted in the createCorrelatedCopy (however wiretap EIP would need to set it to false as the tap exchange should not be transacted. 

So can you look at which other EIPs that use the createCorrelatedCopy API and build unit tests like you have for multicast.

jeremyross:
Doesn't seem to solve the issue when using spring transactions. Not sure if it's supposed to.

core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
zhfeng: Thansk @davsclaus and I will take a look.

zhfeng:
@jeremyross yeah, I think it could reslove the issue when using spring transactions.

zhfeng:
@davsclaus I add a similar fix for RecipientList and Split EIP. This issue only happens when it copies the exchange and sends it to more then one sub-route in the transacted context.

davsclaus:
Can you add a test for the Enricher EIP also.

zhfeng:
Sure, I will do.

jeremyross:
@zhfeng ok just tested with my reproducer project and it looks like it does fix the problem.

zhfeng:
Thanks @jeremyross for checking it !

tests/camel-itest/src/test/java/org/apache/camel/itest/tx/JtaRouteTest.java
```diff
@@ -38,23 +65,94 @@ protected RouteBuilder createRouteBuilder() throws Exception {
...
+                        .log("will never get here");
+
+                from("direct:multicast").routeId("r.route2")
+                        .log(LoggingLevel.DEBUG, "Entering route: ${routeId}")
+                        .multicast()
+                        .to("log:r.test", "direct:route3", "mock:test")
+                        .end();
+
+                from("direct:route3").routeId("r.route3")
+                        .log(LoggingLevel.DEBUG, "Entering route: ${routeId}");
```
davsclaus, line 91: I think we should add a inlined processor that tests that the exchange is transacted, in all of these sub routes (not only here). This will help test the copied exchanges are marked as transactional.

Something ala
.process(e -> Assertions.assertTrue(e.isTransacted())

zhfeng: OK, I get it and is there any way to do similar check with mock endpoint ?

davsclaus: oh yeah you can do something with onExchangeReceived or what the name is

davsclaus: @zhfeng can you work on this so we can get this PR ready and merged and backported as we want to build new releases of Camel LTS versions with this included.

zhfeng: @davsclaus yeah, I'm working on it.

zhfeng: Done

davsclaus:
Sorry can you make this PR compile with JDK8 also so we can backport this to older branches

zhfeng:
Oh, I just missed it.

zhfeng:
We removed the CI build for JDK 8 ?

davsclaus:
There are CI for 3.14 and 3.11 branches.
Camel 3.15 onwards are JDK11 only.

davsclaus:
Though the PR may work with JDK8 .. I frankly may be mistaken. We can see when we backport as you will get CI errors then

zhfeng:
OK, I will check it with JDK8 locally before backport it to 3.14 and 3.11

davsclaus (approved):

zhfeng:
@davsclaus It works with JDK 8 and I just backport to 3.14.x

PR #7041, opened 2022-02-25
CAMEL-17690: camel-test-main - Annotation based testing

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17690

## Motivation

It is already possible to test a Camel main application by extending the base class `CamelMainTestSupport` but we would like to go even further by proposing an approach based on annotations like `SpringBootTest`

## Modifications

* Adds the annotation `CamelMainTest` to configure and trigger the Camel Main test extension which supports out of the box:
  * Allow to advice rules easily -> attribute `advices`
  * Add support of `mockEndpointsAndSkip` and `mockEndpoints` -> attributes of the same name
  * Allow to override properties -> attribute `properties`
  * Add support of replaceRouteFromWith -> attribute of the same name
  * Allow to configure easily the property placeholder locations with a file name or all the locations  -> attributes `propertyPlaceholderFileName` and `propertyPlaceholderLocations` respectively
  * Allow to provide a main class  -> attribute `mainClass`  
  * Allow to add configuration classes  -> attribute `configurationClasses`  
  * Add support of dumpRouteCoverage -> attribute of the same name
* Adds the annotation `Configure` to mark all the methods to call to configure the Camel Main Application
* Adds the annotation `ReplaceInRegistry` to mark all the methods / fields whose return value / value should replace the beans with the same name and same type
* Adds the annotation `AdviceRouteMapping` to define a mapping between a route and the route builder to call to advice the route
* Adds the interface `DebuggerCallback` to implement in order to enable the debug mode


davsclaus:
I wonder for `ReplaceInRegistry` if we can reuse `BindToRegistry` instead, and potentially have an override (or replace if better name) = true|false attribute if needed. I just want to avoid adding too many new annotations.

However this can be looked at in another PR.

essobedo:
I agree with you, having too many annotations doing more or less the same can be confusing for the end user but on the other hand it would mean modifying an annotation that is part of the Camel API for something that sounds really specific to tests. If it is specific to test and it is used abusively outside a test, I think that it could bring even more confusion. To avoid confusion, IMHO, we should avoid mixing the project API with its test API. Moreover having a specific test API allows to take precedence over the project API which is particularly helpful in this use case where you need to replace a bean of a normal application with a bean for testing purpose.

WDYT?

davsclaus:
Yeah okay

davsclaus (approved):

PR #7073, opened 2022-03-01
CAMEL-17720: Add option to allow unmarshalling null body

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17720

## Motivation

It is common to unmarshal data to json via Jackson, and the DataFormat we have will then fail if the message body is empty. We should add an option that can be turned on to allow null body and ideally propose the same option for all Dataformats.

## Modifications

* Add the new option `allowNullBody` at unmarshal EIP level to be able to apply it for all existing DataFormat
* Skip the call to the method `unmarshal` if `allowNullBody` is `true` and the body is `null`
* Add `allowNullBody` to the Java DSL

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/processor/UnmarshalProcessor.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/processor/UnmarshalProcessor.java
@@ -43,9 +43,15 @@ public class UnmarshalProcessor extends AsyncProcessorSupport implements Traceab
     private String routeId;
     private CamelContext camelContext;
     private final DataFormat dataFormat;
+    private final boolean allowNullBody;
 
     public UnmarshalProcessor(DataFormat dataFormat) {
+        this(dataFormat, false);
+    }
+
+    public UnmarshalProcessor(DataFormat dataFormat, boolean allowNullBody) {
         this.dataFormat = dataFormat;
+        this.allowNullBody = allowNullBody;
     }
 
     @Override
@@ -55,13 +61,20 @@ public class UnmarshalProcessor extends AsyncProcessorSupport implements Traceab
         InputStream stream = null;
         Object result = null;
         try {
-            stream = exchange.getIn().getMandatoryBody(InputStream.class);
+            final Message in = exchange.getIn();
+            final Message out;
+            if (allowNullBody && in.getBody() == null) {
+                // The body is null, and it is an allowed value so let's skip the unmarshalling
+                out = exchange.getOut();
+            } else {
+                stream = in.getMandatoryBody(InputStream.class);
 
-            // lets setup the out message before we invoke the dataFormat so that it can mutate it if necessary
-            Message out = exchange.getOut();
-            out.copyFrom(exchange.getIn());
+                // lets set up the out message before we invoke the dataFormat so that it can mutate it if necessary
+                out = exchange.getOut();
+                out.copyFrom(in);
 
-            result = dataFormat.unmarshal(exchange, stream);
+                result = dataFormat.unmarshal(exchange, stream);
+            }
             if (result instanceof Exchange) {
                 if (result != exchange) {
                     // it's not allowed to return another exchange other than the one provided to dataFormat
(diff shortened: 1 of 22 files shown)


essobedo:
@davsclaus Is-it more or less what you have in mind?

core/camel-api/src/main/java/org/apache/camel/spi/DataFormat.java
```diff
@@ -54,4 +54,13 @@
      * @throws Exception can be thrown
      */
     Object unmarshal(Exchange exchange, InputStream stream) throws Exception;
+
+    /**
+     * Indicates whether {@code null} is allowed as value of a body to unmarshall.
+     *
+     * @return {@code true} if {@code null} is allowed, {@code false} otherwise.
+     */
+    default boolean allowNullBody() {
```
oscerd, line 63: Is this correct? It should be specific to Jackson and not generic to data format.

I'm from mobile so I might losing something

oscerd (changes requested):
Just one doubt

core/camel-api/src/main/java/org/apache/camel/spi/DataFormat.java
essobedo: That's a good remark, actually, the failure doesn't happen in the Jackson `DataFormat` itself but in the `UnmarshalProcessor`.  Indeed, The failure is due to the fact that it cannot find a converter to convert a `null` value into an `InputStream` which occurs when calling `in.getMandatoryBody(InputStream.class)`. That is why I had to manage it at a higher level, moreover @davsclaus proposed in the same ticket to eventually have it for all the existing DataFormat which could give more sense to this approach. Do you see a better/simpler approach? If I need to apply the same logic to the other DataFormats, I had better to have the best approach.

oscerd: Well, I didn't read the suggestion from @davsclaus, in that case it makes completely sense as a basic step to then provide the feature on all the dataformat. So fine with me. Thanks for the explanation.

core/camel-support/src/main/java/org/apache/camel/support/processor/UnmarshalProcessor.java
```diff
@@ -55,13 +55,16 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
         InputStream stream = null;
         Object result = null;
         try {
-            stream = exchange.getIn().getMandatoryBody(InputStream.class);
+            final Message in = exchange.getIn();
+            final Message out = exchange.getOut();
```
davsclaus, line 59: Do not call getOut() to early as it will create an empty message, and cause the exchange to have this empty message as the response, even if the message body is null and allow null enabled. So move this out to later where its first in use

core/camel-api/src/main/java/org/apache/camel/spi/DataFormat.java
davsclaus: Yeah I can't think of a better way as there is no common DataFormatConfiguration class that they all extend etc where we could have common options. 

But this is really a useful feature to let the data format not fail on null bodies.

Especially for use-cases with json where we would often use jackson.

essobedo:
@davsclaus @oscerd What do you think of this new approach? As it is not related to DataFormat but rather to the way to Unmarshal, I propose to add an option directly in the Unmarshal EIP, WDYT? Is that acceptable?

davsclaus (approved):

davsclaus:
Yeah this is better.

Would be good to have an test that the yaml-dsl works too
https://github.com/apache/camel/blob/main/dsl/camel-yaml-dsl/camel-yaml-dsl/src/test/groovy/org/apache/camel/dsl/yaml/UnmarshalTest.groovy

You could maybe add a new test method (and keep existing as-is)

essobedo:
@davsclaus Good remark, done

davsclaus (approved):

oscerd:
Sorry, late to the party. Thanks for this @essobedo, keep up the good work!

PR #7091, opened 2022-03-02
CAMEL-17555: Flag to include/exclude routes in route templates

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17555

## Motivation

To make route templates (aka kamelets) more flexible it can benefit that we allow to declare routes whether they should be included or excluded. This allows to parameterize these routes and therefore only include them if the template has a parameter included for this.

## Modifications

* Add a new parameter called `precondition` allowing to provide a predicate in Simple language to evaluate in order to know if a given route should be excluded or not.
* Remove the definition of the route if the `precondition` fails

### DIFF
--- a/core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultCamelContext.java
+++ b/core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultCamelContext.java
@@ -1000,7 +1000,7 @@ public class DefaultCamelContext extends SimpleCamelContext implements ModelCame
     private boolean includedRoute(RouteDefinition definition) {
         final String precondition = definition.getPrecondition();
         if (precondition == null) {
-            LOG.debug("No precondition found, the route is included by default");
+            LOG.trace("No precondition found, the route is included by default");
             return true;
         }
         final ExpressionDefinition expression = new SimpleExpression(precondition);
@@ -1010,8 +1010,8 @@ public class DefaultCamelContext extends SimpleCamelContext implements ModelCame
         predicate.initPredicate(this);
 
         boolean matches = predicate.matches(new DefaultExchange(this));
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("The precondition has been evaluated to {}, thus the route is {}", matches,
+        if (LOG.isTraceEnabled()) {
+            LOG.trace("The precondition has been evaluated to {}, consequently the route is {}", matches,
                     matches ? "included" : "excluded");
         }
         return matches;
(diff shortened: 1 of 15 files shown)


core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultCamelContext.java
```diff
@@ -871,28 +875,40 @@ public void startRouteDefinitions(List<RouteDefinition> routeDefinitions) throws
                     // need to reset auto assigned ids, so there is no clash when creating routes
                     ProcessorDefinitionHelper.resetAllAutoAssignedNodeIds(routeDefinition);
                 }
+                // Check if the route is included
+                if (includedRoute(routeDefinition)) {
```
essobedo, line 879: Need to evaluate the precondition here as we need the parameters to be set

core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultCamelContext.java
```diff
@@ -975,6 +991,32 @@ protected synchronized boolean removeRoute(String routeId, LoggingLevel loggingL
...
+    /**
+     * Indicates whether the route should be included according to the precondition.
+     *
+     * @param  definition the definition of the route to check.
+     * @return            {@code true} if the route should be included, {@code false} otherwise.
+     */
+    private boolean includedRoute(RouteDefinition definition) {
+        final String precondition = definition.getPrecondition();
+        if (precondition == null) {
+            LOG.debug("No precondition found, the route is included by default");
```
davsclaus, line 1003: I think trace level is fine here, as otherwise debug is a bit spammed as 99,9% of routes are included

core/camel-core-model/src/main/java/org/apache/camel/model/RouteDefinition.java
```diff
@@ -456,6 +457,18 @@ public RouteDefinition autoStartup(boolean autoStartup) {
         return this;
     }
 
+    /**
+     * Sets the expression of the precondition in simple language to evaluate to determine if this route should be
```
davsclaus, line 461: I think you should say predicate instead of expression - also further below

docs/user-manual/modules/ROOT/pages/routes.adoc
```diff
@@ -54,6 +54,44 @@ rb -> rb.from("kafka:cheese").to("jms:queue:foo");
...
+[source,xml]
+----
+<route precondition="'{{format}}' == 'xml'">
+  <from uri="direct:in"/>
+  <unmarshal><jaxb/></unmarshal>
+  <to uri="direct:out"/>
+</route>
+----
+
+And in YAML DS:
```
davsclaus, line 81: DSL

davsclaus (commented):
Great work only some small comments

essobedo:
Many thanks @davsclaus for your remarks, I addressed them all and I added a test for Spring

davsclaus (approved):

oscerd (approved):

PR #7093, opened 2022-03-03
CAMEL-17727: camel-kafka - Add readiness health check for kafka consumer

davsclaus:

### DIFF
--- a/components/camel-kafka/pom.xml
+++ b/components/camel-kafka/pom.xml
@@ -38,6 +38,10 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-support</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-health</artifactId>
+        </dependency>
 
         <!-- kafka java client -->
         <dependency>
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConsumer.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConsumer.java
@@ -24,9 +24,12 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
+import org.apache.camel.CamelContextAware;
 import org.apache.camel.Processor;
 import org.apache.camel.ResumeAware;
 import org.apache.camel.component.kafka.consumer.support.KafkaConsumerResumeStrategy;
+import org.apache.camel.health.HealthCheck;
+import org.apache.camel.health.HealthCheckAware;
 import org.apache.camel.spi.StateRepository;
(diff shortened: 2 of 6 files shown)


oscerd (approved):

orpiske (approved):
Nice, this looks cool.

orpiske:
> A first prototype of readiness health check for kafka consumers.
> 
> I wonder if there is more details we should gather in the details map when it is DOWN.
> 
> Also this is only on consumer level, for producer its likely a bit harder to do.

We can probably play a bit w/ the PartitionAssignmentListener to check for some rebalancing events (if relevant - I am not sure yet).

orpiske:
For producers, I think we can look at the commit managers. We have centralized the commit handling, so I think it would be possible to look for failed commit events and act/report on it.

orpiske:
> 

I think that would be nice. IMHO, this will be necessary if/once we decide to support partitioning of sources (i.e.: so that other members of an ensemble can react if one or more of them fails/stalls).

davsclaus:
Also this PR likely need some feedback/thoughts on that we have built-in self-healing by re-connect/re-try attempts in the fetcher that may go against a k8s container platform, where you may want to signal ready = false, and let the k8s container restart the pod.

davsclaus:
I had to refine the consumer check as kafka is a real *** where it will just log WARNs and not surface the problem to its own kafka consumer and therefore not to camel kafka consumer.

davsclaus:
Here is how you can try this locally
https://github.com/apache/camel-examples/tree/main/examples/jbang-kafka-health

davsclaus:
I found a better way to check for connectivity with kafka in the lower level network client (but requires reflection to access it) because apache kafka does not expose an API for this :(

But this works better as that is the real network connection that it's checking, for READY connection state.

davsclaus:
btw love @oscerd kamelets and its easier to just run jbang with a kamelet to consume from kafka than it is to setup a java project ;)

oscerd:
I'm planning to add Kafka Kamelets for SSL too. It's really nice to play with them.

I think, with a bit of work, we might also try to work on a set of integration tests, done through test-infra contiainers and camel-jbang/kamelet-main

davsclaus:
Screenshot 2022-03-04 at 10 06 00

essobedo:
That's very nice, as far as I understand this PR, it means that for each endpoint (consumer or producer), we will have a specific health check, is-that correct? 

If so, it means that if several endpoints access to the same brokers, we will actually do several times the same task so would it make sense to let the components decide when to create a new Health Check? This way even if we have several endpoints accessing to the same brokers, we would have only one heath check.

davsclaus:
yeah at first we will have specific per consumer/producer that uses the same actual network client that kafka-client uses, so we check the same thing that would be doing the network communcation.

if we do not do that then we are not really 100% checking the correct place. 

you can argue if kafka-client had some connection-pool or shared network clients then reuse could be higher in their client and the camel health-check could check on the pool instead

PR #7104, opened 2022-03-04
CAMEL-17741: Improper way to get the default name of a field annotated with XmlElement

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17741

## Motivation

A field annotated with `@XmlElement` without the attribute `name` should have as XML name its name according to the JAXB specification which is not currently the case.

## Modifications:

* Rewrite the method `fieldName` to make sure that the attribute name of the annotation `@XmlRootElement` is only retrieved in case if a field is annotated with `@XmlElementRef` (or `@DslProperty`) as stated by the spec
* Remove the annotation `@XmlValue` from the method `fieldName` as it doesn't hold any name
* Remove the method `annotationValue` for a `MethodInfo` as it was not used


essobedo:
Please note that this change has no effect on current mapping

davsclaus (approved):

PR #7105, opened 2022-03-04
CAMEL-17717: REST DSL securityRequirements cleanup

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17717

## Motivation

The class `SecurityRequirementsDefinition` contains only one property which should be moved directly into `RestDefinition` for the sake of simplicity.

## Modifications:

* Move all the fields and methods of `SecurityRequirementsDefinition` into `RestDefinition`

## Result

The mappings (XML, YAML...) remain the same but the class `SecurityRequirementsDefinition` has been removed.


essobedo:
Waiting for https://github.com/apache/camel/pull/7104

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
```diff
@@ -606,6 +619,10 @@ public RestDefinition security(String key, String scopes) {
         return this;
     }
 
+    public Collection<SecurityDefinition> securityRequirements() {
```
davsclaus, line 622: What is this method for?

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
```diff
@@ -80,11 +84,13 @@
...
     private RestSecuritiesDefinition securityDefinitions;
-    @XmlElement(name = "securityRequirements") // use the name Swagger/OpenAPI uses
+    @XmlElement
     @Metadata(label = "security")
-    private SecurityRequirementsDefinition securityRequirements;
+    private List<SecurityDefinition> securityRequirements = new ArrayList<>();
     @XmlElementRef
     private List<VerbDefinition> verbs = new ArrayList<>();
+    @XmlTransient
+    private Map<String, SecurityDefinition> itemsMap = new HashMap<>();
```
davsclaus, line 93: Can we avoid this internal special map as this is not common practice in the model classes, so can you try to find another way.

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
```diff
@@ -170,15 +176,17 @@ public void setSecurityDefinitions(RestSecuritiesDefinition securityDefinitions)
...
+    public List<SecurityDefinition> getSecurityRequirements() {
         return securityRequirements;
     }
 
     /**
      * Sets the security requirement(s) for all endpoints.
      */
-    public void setSecurityRequirements(SecurityRequirementsDefinition securityRequirements) {
-        this.securityRequirements = securityRequirements;
+    public void setSecurityRequirements(Collection<SecurityDefinition> securityRequirements) {
```
davsclaus, line 186: The getter/setter should be same type, eg List

davsclaus (changes requested):

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
essobedo: This comes from https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/rest/SecurityRequirementsDefinition.java#L46-L47

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
essobedo: This comes from https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/rest/SecurityRequirementsDefinition.java#L76-L78

core/camel-core-model/src/main/java/org/apache/camel/model/rest/RestDefinition.java
essobedo: This comes from https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/rest/SecurityRequirementsDefinition.java#L87-L91

essobedo:
I will add the proper commit to remove the class as expected

essobedo:
@davsclaus What should I do? Apply your comments within the context of this PR or keep it as it is?

essobedo:
I finally assumed that your answer was yes so I addressed your comments

davsclaus (approved):

davsclaus:
Ah okay this code was contributed by community user so I guess we just accepted it back then - to make rest-dsl have a way to configure security

PR #7107, opened 2022-03-04
[CAMEL-17743] Set correct javaTypes in schema

johnpoth:
https://issues.apache.org/jira/browse/CAMEL-17743

Thanks !


davsclaus:
Can you check the other circuit breaker, mp fault tolerance

oscerd (approved):

davsclaus:
Is the java.util methods not present, so I think it should only be those CB impl classes that should be changed

davsclaus:
The first one also has the JDK thread pools which should be as before, almost there ;)

johnpoth:
Thanks @davsclaus updated! 

I'm still unsure of the javaType's role in the model generation. It seems we loose the underlying Java Type when specified. For example the model for the [circuitBreaker](https://github.com/apache/camel/blob/main/core/camel-core-model/src/generated/resources/org/apache/camel/model/faultToleranceConfiguration.json#L15) property doesn't mention that the underlying type is a [String](https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/FaultToleranceConfigurationCommon.java#L30). So it seems the annotation should never be used or we need to update how the model is generated? I'm probably missing something ...

johnpoth:
Ah right.. updated and rebased to single commit, thanks !

davsclaus (approved):

PR #7118, opened 2022-03-05
CAMEL-17751: Extracting Camel-Saga-Compensate & Camel-Saga-Complete f…

vagpap:
If the verifyRequest method is not able to retrieve:

`Camel-Saga-Compensate` & `Camel-Saga-Complete`

as individual headers, try to extract them from the `CamelHttpQuery` header.
If found, add them also as headers to the exchange.getIn since if they are not there, neither the compensation or the completion is called correctly.

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
@@ -16,10 +16,7 @@
  */
 package org.apache.camel.service.lra;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLDecoder;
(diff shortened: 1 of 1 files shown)


davsclaus (changes requested):
Can you use URISupport from camel camel-util which has APIs to extract http headers

vagpap:
The code currently throws a RuntimeCamelException in case the URISupport.parseQuery method throws the URISyntaxException.

Is this appropriate? Should it throw something else?

oscerd (approved):
LGTM, we need to backport on main too

vagpap:
@oscerd should I do it and then open a new PR for main?

davsclaus (approved):

davsclaus:
I think we can cherry pick to main so lets try this first

PR #7127, opened 2022-03-07
CAMEL-17602: camel-aws-sqs - consumer should map headers back to thei…

davsclaus:
…r original type such as boolean, integer etc.


oscerd (approved):

oscerd:
Maybe in the migration guide we might add a line about this.

davsclaus:
Good idea I added to the docs

PR #7134, opened 2022-03-07
camel-core - Resumable DSL keep same type in Java DSL as it has no mo…

davsclaus:
…re configuration, and changing types can cause problems for Java DSL. Add string parameter variant to make it easy to use.


davsclaus:
Ah yeah just updated the docs

PR #7146, opened 2022-03-08
CAMEL-17593: camel-aws-sqs - Added option so you can choose what to d…

davsclaus:
…o when sending to sqs when hitting maximum number of message headers.


oscerd (approved):

oscerd:
I think this needs to be backported on 3.14.x too and also add a line on migration

PR #7190, opened 2022-03-11
CAMEL-16628: Allow to generate doc about message headers

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-16628

## Motivation

Like we have for endpoint options with @UriParam, we should have a new SPI annotation to markup headers that consumer/producers support.

## Modifications

This PR adds a new element called `headersClass` to the annotation `@UriEndpoint` to provide the class that contains the constants corresponding to the name of the headers supported by the consumers/producers. 

In the example below, the endpoint `FooEndpoint` indicates that the class that contains the name of all the headers supported by the consumers/producers is `FooConstants`.

```
@UriEndpoint(scheme = "foo", syntax = "foo", title = "foo", headersClass = FooConstants.class)
public class FooEndpoint extends Endpoint {
    // Rest of the class
}
```

A headers class can be any class but by convention, we expect a name of type `xxxConstants` where `xxx` is the name of the corresponding component like for example `FtpConstants` for the component `camel-ftp`.

The metadata of a given header are retrieved from the annotation `@Metadata` added to the `String` constant representing its name.

In the example below, the header `FooConstants.FILE_NAME` specify a description, a default value and a java type.
```
public final class FooConstants {

    @Metadata(description = "Some description of CamelFileName", defaultValue = "Some default value", 
                        javaType = "java.lang.String")
    public static final String FILE_NAME = "CamelFileName";
    // Rest of the class
}
```

This PR adds a new asciidoctor template called `component-endpoint-headers.adoc` that can be called from a doc file with:
```
// component headers: START
include::partial$component-endpoint-headers.adoc[]
// component headers: END
```

## Result

image


davsclaus:
Really good start, its not so far again.

I wonder instead of having @Metadata(headers = array here) on the endpoint class which then leads to duplicating effort as you need to add each here here, and as well in the headers class (xxxConstants).

What about instead you add an attribute on @UriEndpoint that refers to the headers class

@UriEndpoint(....
                         headersClass = KafkaConstants.class)

essobedo:
Hum, it is a totally different approach, so I'll need to rewrite nearly everything but ok as you wish. It's my bad, I should have validated my approach first before implementing it. Regarding your suggestion how many headers class do you want to allow? Only one or several classes?

davsclaus:
@essobedo oh my word is not a rule in law - you are the one that are actual implementing this and find out on the way what is a better solution.

Unfortunately the Camel components was very _open ended_ when Camel was created/designed and this leads to where we are today with endpoint headers that are not well documented, defined consistently in a class named XXX or defined directly on the endpoint class, etc.

But we have the power in our hands where we can implement what you work on, and then "fix" all the components in Apache Camel. We have done this over the years with adding documentation for every endpoint options etc.

In the camel-catalog that prints a summary we can write a report which lists all the components that does not yet have headers marked up with @Metadata - then we can see how many we need to fix. This is what we did in the past.
And some components will not have headers, and if so, then should we still have a header class, but keep it empty?

About number of header classes, then that should be 1 class.

essobedo:
@davsclaus I applied your remarks, please check again

davsclaus:
Awesome work, this looks easy to use and also makes it possible for us to migrate existing components.

oscerd (approved):
Amazing work. This is really nice. Thanks a lot!

oscerd:
Once this has been merged, we should create an epic with multiple issue for all the components.

essobedo:
@oscerd here is the task https://issues.apache.org/jira/browse/CAMEL-17792

PR #7192, opened 2022-03-12
CAMEL-17774 Added check for null inputStream

tsantakismanos:
This PR is to add a check for null during `ZipIterator` creation which will prevent the assignment of a `new ZipInputStream(new BufferedInputStream(inputStream))` to `zipInputStream` variable. 
Such case may occur when a message is converted to `InputStream` but no `TypeConverter` is present (e.g in `ZipSplitter`)

### DIFF
--- a/components/camel-zipfile/src/main/java/org/apache/camel/dataformat/zipfile/ZipIterator.java
+++ b/components/camel-zipfile/src/main/java/org/apache/camel/dataformat/zipfile/ZipIterator.java
@@ -48,6 +48,11 @@ public class ZipIterator implements Iterator<Message>, Closeable {
     public ZipIterator(Exchange exchange, InputStream inputStream) {
         this.exchange = exchange;
         this.allowEmptyDirectory = false;
+
+        if (inputStream == null) {
+            throw new IllegalStateException("Input stream cannot be null");
+        }
+
         if (inputStream instanceof ZipInputStream) {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

components/camel-zipfile/src/main/java/org/apache/camel/dataformat/zipfile/ZipIterator.java
```diff
@@ -48,6 +48,11 @@
     public ZipIterator(Exchange exchange, InputStream inputStream) {
         this.exchange = exchange;
         this.allowEmptyDirectory = false;
+
+        if (inputStream == null) {
+            throw new IllegalStateException("Input stream cannot be null");
+        }
```
essobedo, line 54: I’m not sure that an IllegalStateException is the right choice in this case, I would rather use ´Objects.requireNonNull´ Instead.

tsantakismanos: Indeed , changed to the proposed one.

oscerd:
In case you'll need to check a null value, you could have a look at ObjectHelper class methods. It's not for this case, but it might help for future PRs.

davsclaus:
Thanks you are welcome to look into camel-tarfile if this component also should have such a null check too.

davsclaus (approved):

PR #7209, opened 2022-03-15
CAMEL-17798 - Offsets resetting when another Camel node is shutdown

rgala:
This is for Jira CAMEL-17798


orpiske (changes requested):
LGTM, but I think it would be better to rebase your branch to ensure we have a clean commit log. (Only reason why I am marking as request changes).

oscerd (changes requested):
LGTM, but please rebase.

orpiske (approved):

davsclaus (approved):

oscerd (approved):

PR #7213, opened 2022-03-16
CAMEL-17812: Unfetchable mails should not be counted as part of a poll

mash-sap:
https://issues.apache.org/jira/browse/CAMEL-17812

### DIFF
--- a/components/camel-mail/src/main/java/org/apache/camel/component/mail/MailConsumer.java
+++ b/components/camel-mail/src/main/java/org/apache/camel/component/mail/MailConsumer.java
@@ -161,17 +161,8 @@ public class MailConsumer extends ScheduledBatchPollingConsumer {
         try {
             int count = folder.getMessageCount();
             if (count > 0) {
-                List<KeyValueHolder<String, Message>> messages = retrieveMessages();
-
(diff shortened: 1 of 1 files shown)


components/camel-mail/src/main/java/org/apache/camel/component/mail/MailConsumer.java
```diff
@@ -388,63 +387,57 @@ private SearchTerm computeSearchTerm() {
...
-                    throw e;
+                return exchange;
+            } else {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Skipping message as it was flagged as deleted: {}", MailUtils.dumpMessage(message));
                 }
             }
+        } catch (Exception e) {
+            if (skipFailedMessage) {
+                LOG.debug("Skipping failed message at index due {}", e.getMessage(), e);
```
davsclaus, line 433: There is no longer an index so this log message should be updated

davsclaus (changes requested):

components/camel-mail/src/main/java/org/apache/camel/component/mail/MailConsumer.java
mash-sap: Done

davsclaus (approved):

PR #7217, opened 2022-03-17
CAMEL-17762: improve the log message to include more information on failure

orpiske:
I was going to fix a typo on the log message and thought about extending the error message to give more details. 

Any thoughts?

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
@@ -70,6 +70,7 @@ public class ResumableCompletion implements Synchronization {
 
     @Override
     public void onFailure(Exchange exchange) {
-        LOG.warn("Skipping offset update for due to failure in processing");
+        LOG.warn("Skipping offset update due to failure in processing: {}", exchange.getException().getMessage(),
+                exchange.getException());
     }
 }


orpiske:
@davsclaus @oscerd FYI ... please let me know if you think this is going to be overly verbose (it wasn't on my current test, though).

oscerd (approved):

oscerd:
LGTM.

davsclaus:
In what circumstance can that onFailure be called? Is it when failing to update the offset itself in some storage, or is it for example when a camel exchange is failed during routing or what?

orpiske:
> In what circumstance can that onFailure be called? Is it when failing to update the offset itself in some storage, or is it for example when a camel exchange is failed during routing or what?

I think this will happen primarily when an exchange is failing during routing.

davsclaus:
Ah okay so maybe we can regard the resume offset as a kind of transaction and if we look at the TX error handler then it has a rollbackLoggingLevel you can configure
https://github.com/apache/camel/blob/028a6b8cdc86b39d3a11dfbf7f5aa4ab96b3cce0/components/camel-spring/src/main/java/org/apache/camel/spring/spi/TransactionErrorHandlerBuilder.java#L40

Which logs like this (without stacktrace)
https://github.com/apache/camel/blob/743d226a07b6926a4935d772b19f65c0e71a69a7/components/camel-spring/src/main/java/org/apache/camel/spring/spi/TransactionErrorHandler.java#L283

As regular Camel error handler does the error handling / logging so you can see the stacktrace there.
If there is stacktrace then it would likely be duplicated and can become very verbose.

orpiske:
> Ah okay so maybe we can regard the resume offset as a kind of transaction and if we look at the TX error handler then it has a rollbackLoggingLevel you can configure
> 
> https://github.com/apache/camel/blob/028a6b8cdc86b39d3a11dfbf7f5aa4ab96b3cce0/components/camel-spring/src/main/java/org/apache/camel/spring/spi/TransactionErrorHandlerBuilder.java#L40
> 
> Which logs like this (without stacktrace)
> 
> https://github.com/apache/camel/blob/743d226a07b6926a4935d772b19f65c0e71a69a7/components/camel-spring/src/main/java/org/apache/camel/spring/spi/TransactionErrorHandler.java#L283
> 
> As regular Camel error handler does the error handling / logging so you can see the stacktrace there. If there is stacktrace then it would likely be duplicated and can become very verbose.

Cool, thanks Claus. I think this will do the trick and I'll take a closer look at it. I'll close this PR for now.

orpiske:
Closing. TBI.

PR #6750, opened 2022-01-14
(RFC) (chores) ci: added support for running quick tests

orpiske:
This adds a custom action that detects if the PR changes a component and, if it does, then run a quick test of only the modified component.

PR #7252, opened 2022-03-23
DO NOT MERGE: ci - added support for running quick tests

orpiske:
Trying again the idea on PR #6750. 

This adds a custom action that detects if the PR changes a component and, if it does, then run a quick test of only the modified component.


orpiske:
We don't seem to have comment permission on this repo. I opened a ticket for that: https://issues.apache.org/jira/browse/INFRA-23027

orpiske:
Comment is not allowed, so I removed that part.

orpiske:
The PR needs to be recreated. Closing.

PR #7254, opened 2022-03-23
[CAMEL-17845] Use single Kafka server when running IT

johnpoth:
https://issues.apache.org/jira/browse/CAMEL-17845

Speeds up from ~14 min to ~4 min for a full build with IT

I've also fixed a couple of flaky ITs. KafkaConsumerTopicIsPatternIT is still failing for me locally when I run it with other tests (was failing before too),  requires further investigation. 

Thanks !

### DIFF
--- a/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerManualCommitIT.java
+++ b/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerManualCommitIT.java
@@ -29,9 +29,11 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.TestInstance;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
 public class KafkaConsumerManualCommitIT extends BaseEmbeddedKafkaTestSupport {
 
     public static final String TOPIC = "testManualCommitTest";
--- a/test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
+++ b/test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
@@ -23,7 +23,6 @@ import org.testcontainers.utility.DockerImageName;
 
 public class ContainerLocalSingletonKafkaService extends ContainerLocalKafkaService
         implements ExtensionContext.Store.CloseableResource {
-    private static boolean started;
 
     public ContainerLocalSingletonKafkaService(KafkaContainer container) {
         super(container);
@@ -35,11 +34,10 @@ public class ContainerLocalSingletonKafkaService extends ContainerLocalKafkaServ
 
     @Override
     public void beforeAll(ExtensionContext extensionContext) {
-        if (!started) {
-            started = true;
-            extensionContext.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).put("kafka", this);
+        extensionContext.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).getOrComputeIfAbsent("kafka", s -> {
             super.initialize();
-        }
+            return this;
+        });
     }
 
     @Override
@@ -53,9 +51,8 @@ public class ContainerLocalSingletonKafkaService extends ContainerLocalKafkaServ
     }
 
     public static ContainerLocalSingletonKafkaService kafka3Container() {
-        KafkaContainer container = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.1"));
-        container = container.withEmbeddedZookeeper();
-
+        KafkaContainer container
+                = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.1")).withEmbeddedZookeeper();
         return new ContainerLocalSingletonKafkaService(container);
     }
 }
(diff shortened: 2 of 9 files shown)


oscerd (approved):

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerManualCommitIT.java
```diff
@@ -29,9 +29,11 @@
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.TestInstance;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
```
essobedo, line 36: AFAIK it is the default, so it should not be needed

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
```diff
@@ -0,0 +1,61 @@
...
+        super();
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext extensionContext) {
+        if (!started) {
+            started = true;
+            extensionContext.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).put("kafka", this);
+            super.initialize();
+        }
```
essobedo, line 42: It is not really thread safe, in case of parallel build we could face issues with it. To simplify and make it thread safe, you could use `getOrComputeIfAbsent` instead to initialize it only once, like I did [here](https://github.com/apache/camel/blob/main/components/camel-test/camel-test-main-junit5/src/main/java/org/apache/camel/test/main/junit5/CamelMainExtension.java#L76)

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public void close() {
+        super.shutdown();
+    }
+
+    public static ContainerLocalSingletonKafkaService kafka3Container() {
+        KafkaContainer container = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.1"));
+        container = container.withEmbeddedZookeeper();
```
essobedo, line 57:
```suggestion
        KafkaContainer container = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.1"))
                .withEmbeddedZookeeper();
```

essobedo (commented):
Sounds good, only small remarks

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerManualCommitIT.java
johnpoth: Hi @essobedo ! Thanks for reviewing! 
It's no longer the default (PER_CLASS inherited from the [parent](https://github.com/apache/camel/pull/7254/commits/b54c747b22baaec4e5faecc5588223d4c380b7d4#diff-51302f36d8ba1b24821b40d3b2b95b37d06a07537e8d965ddfe11c075210239dR36))
Thanks!

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
johnpoth: Hmm I doubt these ITs were run in parallel. That would mean that multiple Kafka instances were started in parallel :)

But in the spirit of running these in parallel in the future I'll add the suggested changes :) thanks !

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
johnpoth: Done, thanks !

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
essobedo: I maybe wrong but AFAIK, the test-infra artifacts are part of the Camel API for the tests which means that it is meant to be used by the camel community so it could potentially be used in a parallel build even if we don't do it within the context of Camel itself. 

Here as you tied it to the `GLOBAL` namespace, if it is thread safe (by using `getOrComputeIfAbsent` for example), it should start only one Kafka instance even in the context of a parallel build

orpiske (approved):
Wow, that's a nice improvement in test time. Thanks @johnpoth !

davsclaus (approved):

test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/ContainerLocalSingletonKafkaService.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void close() {
+        super.shutdown();
+    }
+
+    public static ContainerLocalSingletonKafkaService kafka3Container() {
+        KafkaContainer container
+                = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.0.1")).withEmbeddedZookeeper();
```
essobedo, line 55: It would be nice, if we could move the name of the image in static field like this example https://github.com/apache/camel/blob/main/test-infra/camel-test-infra-cassandra/src/test/java/org/apache/camel/test/infra/cassandra/services/CassandraLocalContainerService.java#L30

orpiske: Hey @essobedo, not meaning to overrule you, but I am going to merge the code as is. I'll take responsibility of doing the change you suggested (I also talked w/ @johnpoth about it and it's fine for him). 

I want to apply the same idea to other services in the test-infra and I have time right now to focus on it.

essobedo: No worries, I only make suggestions, you are always free to follow them or not

PR #7261, opened 2022-03-24
[CAMEL-17844] Handle FunctionalInterface (lambas) in BeanInfo

bjornomy:
This PR aims to fix a regression regarding FunctionalInterfaces in BeanInfo.

* let BeanInfo ignore methods from synthetic classes (proxies and lambdas) when choosing what method to call on a bean
* remove BeanInfoTypeResolver, as it's not longer necesary
* add a doc snippet on `@FunctionalInterface` usage

- [X] Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.


bjornomy:
Is there a way to automatically format the source in such a way that `sourcecheck` passes?

oscerd:
> Is there a way to automatically format the source in such a way that `sourcecheck` passes?

If you run a full build the code will autoformat.

davsclaus:
Also need to test that camel-karaf would still work, as it has special case for beans, the interface that was deleted was in use by camel-karaf, so we need to update its source code also

essobedo:
> Is there a way to automatically format the source in such a way that `sourcecheck` passes?

What I personally do for this is to launch the formatter (`formatter:format`) when I build, so I end up with the next command `mvn formatter:format clean install -Psourcecheck`

davsclaus:
Your changed with checking for synthetic later could potentially make no need for the camel-bean-osgi module which we then can remove - I am testing this now

davsclaus:
Thanks for the PR I have merged it manually

PR #7275, opened 2022-03-28
CAMEL-17802: remove the autoCommitOnStop because the consumer already auto-commits on the call to close

orpiske:
Please don't merge it yet. I'd like to see if the reporter can
review/test it.


oscerd (approved):

oscerd:
@rgala have a look if you have time.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/CommitManagers.java
```diff
@@ -34,13 +34,7 @@ public static CommitManager createCommitManager(
...
 
         if (configuration.isAutoCommitEnable()) {
-            if ("async".equals(configuration.getAutoCommitOnStop())) {
-                return new AsyncCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-            } else if ("sync".equals(configuration.getAutoCommitOnStop())) {
-                return new SyncCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-            } else if ("none".equals(configuration.getAutoCommitOnStop())) {
-                return new NoopCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-            }
+            return new AsyncCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
```
rgala, line 37: I don't think it is needed because no matter what the commit manager does, the offset will be handled automatically on consumer close. Wouldn't it be better to return NoopCommitManager here?

orpiske: I think it could be. There's more to it, though.

davsclaus (approved):

orpiske:
Looking good. Let's merge.

PR #7281, opened 2022-03-29
CAMEL-17859: increase logs in critical smpp code for diagnosing the issue root cause

orpiske:
This is a WIP and available for the user to help diagnose and fix the
issue

### DIFF
--- a/components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConsumer.java
+++ b/components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConsumer.java
@@ -194,6 +194,9 @@ public class SmppConsumer extends DefaultConsumer {
             } finally {
                 reconnectLock.unlock();
             }
+        } else {
+            LOG.warn("Thread {} could not acquire a lock for creating the session during consumer reconnection",
+                    Thread.currentThread().getId());
         }
     }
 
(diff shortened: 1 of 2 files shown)


orpiske:
c/c: @PascalSchumacher if you can give it a try and see if we can collect more information with these ... Thanks in advance.

PascalSchumacher:
Sorry I am not able to reduce the issue. 😞 

I am able to simulate part of the events from the log in the issue (existing connection, enquire link timeout) using a hacked copy of `DefaultSmppServer` from https://github.com/fizzed/cloudhopper-smpp.

But I do not know how to simulate `SocketException` (and the `InterruptedException` with the right timing).

orpiske:
> Sorry I am not able to reduce the issue. 😞
> 
> I am able to simulate part of the events from the log in the issue (existing connection, enquire link timeout) using a hacked copy of `DefaultSmppServer` from https://github.com/fizzed/cloudhopper-smpp.
> 
> But I do not know how to simulate `SocketException` (and the `InterruptedException` with the right timing).

That's OK. For now, let's increase the logs in the `tryLock` failures so we are not in the dark in the case it fails. Based on our discussion on the ticket, I also believe they could play a role in this, but I don't want to change the code without a solid evidence of the root cause - primarily to avoid making it even worse.

orpiske:
And thanks for taking the time to try to reproduce it!

PascalSchumacher:
Yes, the logging should make it easier to diagnose the problem if it occurs again.

PR #7286, opened 2022-03-29
Add singleton support for more test infra services:

orpiske:
- ArangoDB
- MongoDB
- HBase


oscerd (approved):

orpiske:
> Hi @orpiske, it is not related directly to your change but I was wondering why we need to call [`getLogger` in the function `defaultCreator`](https://github.com/apache/camel/blob/63d6a8bf425145450a827d60cdbf25d194a4a381/test-infra/camel-test-infra-common/src/test/java/org/apache/camel/test/infra/common/services/SingletonService.java#L39). Could it be a static field instead?

Thanks for pointing it out. It should be static field.

PR #7299, opened 2022-03-30
CAMEL-17687 - Create a Camel Azure Key Vault component - Purge Deleted Secret Operation

oscerd:
Let's try the new stuff from @orpiske :-D


orpiske:
Hm, I think I need to fix the mask 🤨

orpiske:
Based on what I see on #7300, it should be fixed now. Sorry to hijack your thread @oscerd 😆

PR #7300, opened 2022-03-30
CAMEL-17866: fix CamelSqlGeneratedKeyRows issue.

klease:
CamelSqlGeneratedKeyRows is not populated if already exists because input
headers were overwriting the new output header.
Change code to set headers on input message instead of using deprecated
getOut() method. Ensure that input body is set to null in case of EMPTY_RESULT
on SelectOne queries.


oscerd (approved):

klease:
Thanks @oscerd; I'll merge when the pr build check is done. The checkstyle failure is not due to my changes.

davsclaus (approved):

PR #7310, opened 2022-03-31
Modify MainHelper.optionKey not to camel case components names in keys.

klease:


davsclaus:
Ah you need to do this for languages and dataformats also in case they have a dash

oscerd (approved):

klease:
> Ah you need to do this for languages and dataformats also in case they have a dash
Thanks. In fact, now I see all these cases should already be handled in MainHelper.computeProperties() but the property keys have already been transformed in BaseMainSupport.autoConfigurationFromProperties() using optionkey before calling computeProperties.
Maybe that's what should be removed instead. I'll have another look at it.

PR #7311, opened 2022-03-31
CAMEL-17890: fix handling of component, language or data-format prope…

klease:
…rties with a dash.

Remove initial mapping using optionKey when loading property keys; this is
done in computerProperties.


klease:
The optionKey mapping is also used in other methods but I think it only causes issues in autoConfigurationFromProperties().

PR #7318, opened 2022-04-01
CAMEL-17792: Add documentation about the message headers (K-M)

essobedo:
https://issues.apache.org/jira/browse/CAMEL-17792

## Motivation

It is now possible to document a message header using annotations, let's leverage it to document the headers of the components.

 ## Modifications

For all the components whose name starts with a K, L or M

* Document the messages headers using the annotations
* Migrate the existing documentation of message headers to use the annotations instead
* Re-define shared constants like those in `Exchange` in the local constants class to be able to document it within the context of current component.
* In case of common headers class for several endpoint, replace the scheme of each endpoint with a constant to facilitate usage of the element `applicableFor`
* Move constants corresponding to headers to the headers class.


essobedo:
Depends on https://github.com/apache/camel/pull/7317

oscerd (approved):

PR #7319, opened 2022-04-01
CAMEL-17879: camel-beanio + leftovers from camel-beanstalk

orpiske:


oscerd (approved):
I think we should take care of Spring Boot starters removal too

orpiske:
> I think we should take care of Spring Boot starters removal too

Oh, that's a good reminder. Thanks!

PR #7327, opened 2022-04-01
CAMEL-17893: camel-yaml-dsl - Remove endpoint-dsl notation which was …

davsclaus:
…not well known and is also causing problems for tooling and yaml schema validation.


davsclaus:
@lburgazzoli makes yaml-dsl more consistent and similar to other DSLs and much easier for tooling (cc @mgubaidullin)

davsclaus:
Okay found that we had endpoint-dsl on the consumer side as well which was not documented or tested so I haven't seen it before. But this means we can drop the source generator too.

PR #7339, opened 2022-04-04
Camel 17515

davsclaus:


davsclaus:
Luca / Antonie - I think it makes the most sense to report DOWN for readiness check when camel is shutting down. Suppose the shutdown process takes a longer time, then its better to be DOWN so k8s does not send new traffic to Camel.

oscerd (approved):

astefanutti (approved):
I agree, that makes perfect sense.

lburgazzoli (approved):

PR #7353, opened 2022-04-04
CAMEL-17879: remove deprecated component camel-hystrix

orpiske:


orpiske:
It might be better to wait until the whole set of tests run on my own CI before merging, so I think it would be better to wait until tomorrow for merge.

davsclaus (approved):

orpiske:
The build completely cleanly and, apparently, without any problems on my own CI tests. We should be good to merge this one.

PR #7370, opened 2022-04-05
CAMEL-17879: remove deprecated legacy XML code from camel-cdi

orpiske:


orpiske:
@davsclaus @oscerd maybe it would be good to double check this one. I don't know much about this feature.

davsclaus (approved):

PR #7376, opened 2022-04-06
CAMEL-17897: camel-exec - add exitValues parameter

sbalmer:
Add `exitValues` parameter to `camel-exec` component as documented in https://issues.apache.org/jira/browse/CAMEL-17897


oscerd (approved):
cc @davsclaus

oscerd:
There are some conflicts, can you please rebase?

sbalmer:
Unresolved issues:

* I just used the name from the underlying library. But I'm not too happy with the `exitValues` name. Something like `expectedExitValues`, `acceptedExitValues` or just `failOnNonzeroExitValue` would be other possibilities.
* There is no functional test because I couldn't get `ExecScriptTest` to work on my installation. It complained about a missing JAVA_HOME which is set. I'm unclear on how to proceed. I hacked `ExecJava8IssueTest` to test whether it fails with `exitValues=1` locally, and that worked.

sbalmer:
Improvements left out for now:

- Catching the ExecException and throwing an exception with better diagnostics. I'd prefer using it first as-is and then see what would be useful.
- Syntax for ranges: `exitValues=0-100`. Because the underlying API supports list, adding support for comma-separated lists was easy. But more parsing would complicate matters and I don't have a use case. My use will be `exitValues=0` in all cases.

davsclaus (approved):

PR #7387, opened 2022-04-07
CAMEL-17051: extend the Resume API to include support for pausable consumers

orpiske:
This adds a new API - working similarly to the resume API - that allow
pausing Kafka consumption based on external factors (ie.: such as an
external system being unavailable). 

It is done in generic way to that we can extend it for other components.
Additionally, it should provide the ground work for implementing
(distributed) partitioning support in the future. 

What do you think?

Obs.: a few known things to be fixed before merge

- [x] Documentation
- [x] There's some poor code in the component hot path that can be optimized
- [x] And + any comments from review


davsclaus:
Hi Great work so far

I think pausing should be possible both manually and automatic. The former is to integrate with Camel's existing Lifecycle with the suspend/resume on the Camel Service.

1) manual
KafkaConsumer (from kafka-client) have its own pause/resume API, which seems to not be in use. I think camel end users would want camel-kafka to use these APIs for pausing.

Camel has suspend/resume operations on a route (can call via Java or JMX) so users can force a pauses/resume. I think the camel kafka consumer should then implement Suspendable and then the doSuspend/doResume operations should then trigger the kafka-client pauses/resume methods.

2) automatic
The generic way with a "boolean" seems a bit more more like a readiness check that would break out the poll loop, which you can then "configure" or "customize" to make it automatic. I wonder if the BooleanSupplier API is a bit too simplistic? You may need access to the Camel Consumer so you have a way to use information from there to know if you can continue or not.

orpiske:
Thanks!

> Hi Great work so far
> 
> I think pausing should be possible both manually and automatic. The former is to integrate with Camel's existing Lifecycle with the suspend/resume on the Camel Service.
> 
>     1. manual
>        KafkaConsumer (from kafka-client) have its own pause/resume API, which seems to not be in use. I think camel end users would want camel-kafka to use these APIs for pausing.

Yes, that's was the primary goal for CAMEL-17051. I think this is a common problem for other components too even if they don't have native pause support like the Kafka client does. 
I also think that native pause support may be important in the future once we start exploring partitioning support (granted: this is just a speculation ATM, but I supose it will be helpful) 

> 
> 
> Camel has suspend/resume operations on a route (can call via Java or JMX) so users can force a pauses/resume. I think the camel kafka consumer should then implement Suspendable and then the doSuspend/doResume operations should then trigger the kafka-client pauses/resume methods.

So, I think I should explore extending this API to support that.


> 
>     2. automatic
>        The generic way with a "boolean" seems a bit more more like a readiness check that would break out the poll loop, which you can then "configure" or "customize" to make it automatic. I wonder if the BooleanSupplier API is a bit too simplistic? You may need access to the Camel Consumer so you have a way to use information from there to know if you can continue or not.

I think you raise a good point here. I will take a closer look at whether the BooleanSupplier can be enough or whether I can come up with a more flexible design for this part of the API.

Thanks for the review!

orpiske:
Resolved TODO + comments. 

Marking it as ready for review, to avoid packing too much into it.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -499,4 +510,19 @@ long getCurrentRecoveryInterval() {
...
+    /*
+     * This is for manually pausing the consumer
+     */
+    public void pause() {
+        consumer.pause(consumer.assignment());
+    }
+
+    /*
+     * This is for manually resuming the consumer (not to be confused w/ the Resume API). This is
+     * mostly used for directly calling pause from Java code or via JMX
```
davsclaus, line 523: should this javadoc not be added to the pause method above

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/pausable.json
```diff
@@ -0,0 +1,20 @@
+{
+  "model": {
+    "kind": "model",
+    "name": "pausable",
+    "title": "Pausable",
+    "description": "Resume EIP to support resuming processing from last known offset.",
```
davsclaus, line 6: I think the description is a copy from the resume EIP. It should be something about pausing

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConfiguration.java
```diff
@@ -1612,15 +1612,15 @@ public void setValueDeserializer(String valueDeserializer) {
...
     }
 
-    public String getSeekTo() {
+    public SeekPolicy getSeekTo() {
         return seekTo;
     }
 
     /**
      * Set if KafkaConsumer will read from beginning or end on startup: beginning : read from beginning end : read from
      * end This is replacing the earlier property seekToBeginning
```
davsclaus, line 1621: Remove that outdated comment about seekToBeginning

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/PollExceptionStrategy.java
```diff
@@ -24,6 +24,20 @@
...
+     * Reset any error flags set by a previous error condition
+     */
+    default void reset() {
+
+    }
+
+    /**
+     * Whether to continue fetching or not
+     * 
+     * @return
```
davsclaus, line 37: polish javadoc

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/PollExceptionStrategy.java
```diff
@@ -24,6 +24,20 @@
  */
 public interface PollExceptionStrategy {
```
davsclaus, line 25: Add class javadoc if missing as its an interface Camel end users can implement

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/KafkaConsumerListener.java
```diff
@@ -0,0 +1,89 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean afterProcess(ProcessingResult result) {
+        if (result.isFailed()) {
+            LOG.warn("Pausing consumer due to error on the last processing");
+            consumer.pause(consumer.assignment());
+
+            if (seekPolicy == SeekPolicy.BEGINNING) {
```
davsclaus, line 76: Is this desired behaviour? That in case of an failure (what can trigger that failure in ProcessingResult?)

That the consumer is seeking to either beginning or end.?

core/camel-api/src/main/java/org/apache/camel/ConsumerListener.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.camel;
+
+import java.util.function.Predicate;
+
+/**
+ * Proxies between the consumer within Camel to
```
davsclaus, line 23: Polish javadoc

core/camel-api/src/main/java/org/apache/camel/ConsumerListener.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public interface ConsumerListener<C, P> {
+
+    /**
+     * Whether processing can resume or not
+     *
+     * @return true if it can, or false otherwise
+     */
+    void setResumableCheck(Predicate<?> afterConsumeEval);
```
davsclaus, line 32: Method does not return anything. Can you update javadoc on these methods, such as tell what returning true or false means in the following methods

core/camel-core-model/src/main/java/org/apache/camel/model/PausableDefinition.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @XmlAttribute(required = true)
+    @Metadata(required = true, javaType = "org.apache.camel.ConsumerListener")
+    private String consumerListener;
+
+    @XmlTransient
+    private Predicate<?> untilCheck;
+
+    @XmlAttribute(required = true)
+    private String untilCheckRef;
```
davsclaus, line 50: Avoid using xxxRef style, but just name it xxx
Also set javaType = "...Predicate" as its a known class

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: Yeap. Fixing it, thanks!

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/pausable.json
orpiske: Note to self: avoid copying class headers. 

Good catch, thanks!

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/KafkaConsumerListener.java
orpiske: Yes. If there is a failure in processing in subsequent steps. For example, from the integration test: 

```
from(from)
                        .pausable(testConsumerListener, o -> canContinue())
                        .routeId("pausable-it")
                        .process(exchange -> LOG.info("Got record from Kafka: {}", exchange.getMessage().getBody()))
                        .to(intermediate);

                from(intermediate)
                        .process(exchange -> {
                            LOG.info("Got record on the intermediate processor: {}", exchange.getMessage().getBody());

                            if (getCount() <= RETRY_COUNT) {
                                throw new RuntimeCamelException("Error");
                            }
                        })
                        .to(to);
```

Such failure will be added to the processing result (which is internal to the component). IOW, the downstream failures should end up at this point which then (based on the policy) will rewind the consumer to the desired point in the partition.

core/camel-core-model/src/main/java/org/apache/camel/model/PausableDefinition.java
orpiske: In this case it would clash with the untilCheck variable above, but I will name it something else.

orpiske: Ah, I'll do like the resume api definition and rename the predicate instead ;)

davsclaus (approved):

orpiske:
Will merge as soon as we have a clean build on this one.

PR #7392, opened 2022-04-08
CAMEL-17762: allow adjusting log level for exception messages during resume updates

orpiske:
This is a follow up fix from the exception handling discussion we had in
PR #7217.


orpiske:
@davsclaus I hope you don't mind, but I marked you as reviewer since you provided some comments there. I hope I got your suggestion correctly.

core/camel-core-model/src/main/java/org/apache/camel/model/ResumableDefinition.java
```diff
@@ -40,6 +41,13 @@
     @Metadata(required = true, javaType = "org.apache.camel.ResumeStrategy")
     private String resumeStrategy;
 
+    @XmlTransient
+    private LoggingLevel loggingLevel = LoggingLevel.WARN;
```
davsclaus, line 45: See hos we do this in LogDefinition with just a String type (not a Ref name)

org.apache.camel.model.LogDefinition

davsclaus (changes requested):

core/camel-core-model/src/main/java/org/apache/camel/model/ResumableDefinition.java
davsclaus: I had a mistake too ;)
https://github.com/apache/camel/commit/55bc0f2430fa2df1569a8c70b853e0d9e2c5acd3

orpiske: :D Nice, thanks for the tip and I glad you found an issue!

core/camel-core-model/src/main/java/org/apache/camel/model/ResumableDefinition.java
```diff
@@ -40,6 +40,11 @@
     @Metadata(required = true, javaType = "org.apache.camel.ResumeStrategy")
     private String resumeStrategy;
 
+    @XmlAttribute(required = true)
```
davsclaus, line 43: I do not think this should be required = true, as the default value is ERROR

(PS later you use WARN as the default in the resolveLogger)

core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
```diff
@@ -70,6 +73,29 @@ public void onComplete(Exchange exchange) {
...
     @Override
     public void onFailure(Exchange exchange) {
-        LOG.warn("Skipping offset update for due to failure in processing");
+        Exception e = exchange.getException();
+        Object offset = exchange.getMessage().getHeader(Exchange.OFFSET);
+
+        if (offset instanceof Resumable) {
+            Resumable<?, ?> resumable = (Resumable<?, ?>) offset;
+
+            if (loggingLevel == LoggingLevel.DEBUG && LOG.isDebugEnabled()) {
```
davsclaus, line 82: This would only make logging level work for DEBUG and any other value will log is hardcoded to log at WARN in the else block.

Take a look at CamelLogger class which can log at configured level automatic, so you can use call its log method.

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/ResumableReifier.java
```diff
@@ -47,6 +49,17 @@ protected ResumeStrategy resolveResumeStrategy() {
...
         }
+
         return strategy;
     }
+
+    protected LoggingLevel resolveLoggingLevel() {
+        LoggingLevel loggingLevel = parse(LoggingLevel.class, definition.getLoggingLevel());
+
+        if (loggingLevel == null) {
+            loggingLevel = LoggingLevel.WARN;
```
davsclaus, line 60: Here default value is WARN but in the model it is ERROR.

davsclaus (changes requested):

core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
orpiske: Thanks!

So, what I want here is for the logger to log the exception only if the API log level is debug or lower. Because, then, it makes it easier to figure out what caused the update to not happen.

Do I understand correctly that you are suggesting me to do something like the code below?

```
if (LOG.isDebugEnabled()) {
      CamelLogger.log(LOG, "Skipping offset update with address '{}' and offset value '{}' due to failure in processing: {}",
              resumable.getAddressable(), resumable.getLastOffset().offset(), e.getMessage(), e);
  } else {
      CamelLogger.log(LOG, "Skipping offset update with address '{}' and offset value '{}' due to failure in processing: {}",
              resumable.getAddressable(), resumable.getLastOffset().offset(), e.getMessage());
  }
```

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/ResumableReifier.java
orpiske: Ahh, good catch. I think we can default to ERROR.

core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
davsclaus: Ah yeah that makes sense to have hardcoded DEBUG then with the stacktrace included. 

We sometimes add `(stacktrace in DEBUG logging level)` to the "other" message so users can know how to get the stacktrace.

core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
```diff
@@ -70,6 +74,34 @@ public void onComplete(Exchange exchange) {
...
+
+        if (offset instanceof Resumable) {
+            Resumable<?, ?> resumable = (Resumable<?, ?>) offset;
+
+            String logMessage = String.format(
+                    "Skipping offset update with address '%s' and offset value '%s' due to failure in processing: %s",
+                    resumable.getAddressable(), resumable.getLastOffset().offset(), e.getMessage());
+
+            if (LOG.isDebugEnabled()) {
+                CamelLogger.log(LOG, loggingLevel, logMessage, e);
```
davsclaus, line 88: We would actually have 2 logging levels

- LOG instance that are controlled by log4j properties
- loggingLevel which is EIP option

So I think you need to log with stacktraces if one of them is DEBUG, and hence you need to test both.

And then the `loggingLevel` you pass in to CamelLogger should be hardcoded to DEBUG so its always logged at that level.

davsclaus (changes requested):
see my prev comment

core/camel-core-processor/src/main/java/org/apache/camel/processor/resume/ResumableCompletion.java
davsclaus: Yes

orpiske: Great, thanks for confirming. I'll adjust.

davsclaus (approved):

PR #7395, opened 2022-04-08
CAMEL-17879: remove camel-etcd leftovers

orpiske:


davsclaus (approved):

davsclaus:
You can also delete `src/generated` in the endpoint and component DSL to have it fully regen from clean in case some left overs still

orpiske:
> 

Thanks for the tip!

PR #7435, opened 2022-04-13
CAMEL-17763: cleaned up unused exceptions in test code (0)

orpiske:


orpiske:
Zookeeper failure is unrelated, so merging.

PR #7442, opened 2022-04-13
CAMEL-17962: fix build error caused by cleanup of duplicated charset features in camel-netty

orpiske:


orpiske:
Merging it directly, as it is causing a build failure on some of the Netty code.

PR #7444, opened 2022-04-13
[CAMEL-17954]fix integration tests failure

ffang:

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/test/java/org/apache/camel/component/salesforce/CompoundTypesIntegrationTest.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/test/java/org/apache/camel/component/salesforce/CompoundTypesIntegrationTest.java
@@ -23,7 +23,7 @@ import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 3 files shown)


components/camel-salesforce/it/resources/CamelSalesforceIntegrationTests.connectedApp
```diff
@@ -21,8 +21,8 @@
...
-        <consumerKey>${salesforce.client.id}</consumerKey>
-        <consumerSecret>${salesforce.client.secret}</consumerSecret>
+        <!--consumerKey>${salesforce.client.id}</consumerKey>
+        <consumerSecret>${salesforce.client.secret}</consumerSecret-->
```
zhfeng, line 25: Is there any reason to comment these two propreties? they are not used in the test?

ffang: Hi @zhfeng ,

No, they are not used in the test, the test will pick up the salesforce.client.id and salesforce.client.secret properties from test-salesforce-login.properties directly.

Plus, having consumerKey and consumerSecret in CamelSalesforceIntegrationTests.connectedApp actually breaks the latest version of ant migration tool(ant-salesforce.jar complains those and failed). Only removing those can setup the salesforce server and run the integration test.

Cheers
Freeman

zhfeng: So it should be better to remove them but not just comment. Or add some words to explain this reason.

components/camel-salesforce/camel-salesforce-component/src/test/java/org/apache/camel/component/salesforce/CompoundTypesIntegrationTest.java
```diff
@@ -23,8 +23,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.junit.Assert.assertEquals;
```
jeremyross, line 26: Please use assertions from org.junit.jupiter.api.Assertions.

ffang: Good catch, revised accordingly.

Thanks @jeremyross !

davsclaus (approved):

jeremyross:
Merged.

PR #7486, opened 2022-04-22
CAMEL-17894: stop running impsort and checkstyle by default

orpiske:
Folks, I'd like to hear your opinions about this rather disruptive
change. 

The motivation for removing those 2 plugins from the default build
execution is because they are very slow. On my test machine, this
reduces the build time by between 3 to 4 minutes with "normal Maven"
(roughly 7.6% improvement). The numbers with Maven Daemon are not so 
dramatic, but still reduces a bit (varying a lot here, but roughly
between 20 to 60 seconds). 

The (possible) downside is that this stops the code from automatically
reformatting the code and sorting the imports by default (ie.: when
running `mvn -Pfastinstall install`). 

Instead, this introduces a new profile `format` that can be run
along side the `fastinstall` to do the same. 

Because this can be disruptive to the workflow of others, I'd like to
hear some thoughts about this.

### DIFF
--- a/.github/workflows/master-pr-build.yml
+++ b/.github/workflows/master-pr-build.yml
@@ -72,7 +72,7 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: 'maven'
       - name: maven build
-        run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -l build.log -Dmvnd.threads=2 -V -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 --no-transfer-progress -Pfastinstall -Dcheckstyle.skip=true -DskipTests install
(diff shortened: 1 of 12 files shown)


orpiske:
I've marked a few folks for some thoughts, but everyone is welcome to give their thoughts.

oscerd (approved):
I agree totally. We could also add a github action running each hour or two, doing the format profile.

orpiske:
> 

Thanks. Yeah, I think that would be a cool idea to work-around any leftovers!

orpiske:
I am marking this one as ready for review.

davsclaus (approved):

orpiske:
Thanks for the reviews/approvals. Let's merge this and, if there's problems, I'll rollback.

orpiske:
Something is not looking OK in some cases 😩. I am looking at it.

orpiske:
> Something is not looking OK in some cases 😩. I am looking at it.

Fixed. https://github.com/apache/camel/commit/8b18abb2db30b0484c4d576e91c2009dadb05a6f

PR #7493, opened 2022-04-25
CAMEL-17861

muellerch2293:
CAMEL-17861: adapt method to determine size of stream
the old version loaded the entire bytestream into memory to determine its length
new implementation is copied from: 
https://github.com/apache/camel/blob/abb1ad84fac520b00339533a2eafdeb901e22d87/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/utils/AWS2S3Utils.java#L83

### DIFF
--- a/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobUtils.java
+++ b/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobUtils.java
@@ -16,13 +16,13 @@
  */
 package org.apache.camel.component.azure.storage.blob;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 
 import org.apache.camel.Exchange;
 import org.apache.camel.Message;
 import org.apache.camel.util.ObjectHelper;
-import org.apache.commons.io.IOUtils;
 
 public final class BlobUtils {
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus (approved):

components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobUtils.java
```diff
@@ -33,11 +33,26 @@ public static Message getInMessage(final Exchange exchange) {
...
     }
 
-    public static Long getInputStreamLength(final InputStream inputStream) throws IOException {
-        final long length = IOUtils.toByteArray(inputStream).length;
-        inputStream.reset();
-
-        return length;
+    public static long getInputStreamLength(InputStream is) throws IOException {
+        if (!is.markSupported()) {
+            return -1;
```
essobedo, line 38: I'm afraid that returning -1 as length could cause side effects in some parts of the code

components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobUtils.java
```diff
@@ -33,11 +33,26 @@ public static Message getInMessage(final Exchange exchange) {
...
+        }
+        if (is instanceof ByteArrayInputStream) {
+            return is.available();
+        }
+        long size = 0;
+        try {
+            is.mark(1024);
+            int i = is.available();
```
essobedo, line 46: I'm not sure that relying on the method `available` is really safe especially when we read the Javadoc about it that is quite flexible regarding the implementation

davsclaus (approved):

PR #7502, opened 2022-04-27
CAMEL-18003: fix incorrect concurrent access to the Kafka consumer

orpiske:
This prevents the code from throwing ConcurrentModificationException
when suspnding

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -50,6 +50,19 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class KafkaFetchRecords implements Runnable {
+    /*
+     This keeps track of the state the record fetcher is. Because the Kafka consumer is not thread safe, it may take
+     some time between the pause or resume request is triggered and it is actually set.
+     Some of the states may be set but not read. This is done deliberately to avoid the code to enter multiple times
+     the branches that handle the *_REQUESTED states.
+     */
+    private enum State {
+        RUNNING,
+        PAUSE_REQUESTED,
+        PAUSED,
+        RESUME_REQUESTED,
+    }
+
     private static final Logger LOG = LoggerFactory.getLogger(KafkaFetchRecords.class);
 
     private final KafkaConsumer kafkaConsumer;
@@ -71,6 +84,8 @@ public class KafkaFetchRecords implements Runnable {
     private long currentBackoffInterval;
     private boolean reconnect; // must be false at init (this is the policy whether to reconnect)
     private boolean connected; // this is the state (connected or not)
+
+    private volatile State state = State.RUNNING;
 
     KafkaFetchRecords(KafkaConsumer kafkaConsumer,
                       BridgeExceptionHandlerToErrorHandler bridge, String topicName, Pattern topicPattern, String id,
@@ -313,6 +328,7 @@ public class KafkaFetchRecords implements Runnable {
                     setConnected(false);
                 }
 
+                updateTaskState();
             }
 
             if (!isConnected()) {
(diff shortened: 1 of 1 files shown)


components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -359,6 +369,23 @@ protected void startPolling() {
...
+                state = State.PAUSED;
+                break;
+            case RESUME_REQUESTED:
+                LOG.info("Resuming the consumer as a response to a resume request");
+                consumer.resume(consumer.assignment());
+                state = State.RUNNING;
+                break;
+            default:
+                break;
+        }
```
essobedo, line 386: I don't know if it is intended but it seems that the states `RUNNING` and `PAUSED` are set but never used. Maybe you can set the state to `null` once done?

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -516,14 +543,16 @@ public BridgeExceptionHandlerToErrorHandler getBridge() {
      * or via JMX
      */
     public void pause() {
-        consumer.pause(consumer.assignment());
+        LOG.info("A pause request was issued and the consumer thread will pause after current processing has finished");
+        state = State.PAUSE_REQUESTED;
```
essobedo, line 547: Maybe we should only allow to pause if the state is `RUNNING `?

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -516,14 +543,16 @@ public BridgeExceptionHandlerToErrorHandler getBridge() {
...
     }
 
     /*
      * This is for manually resuming the consumer (not to be confused w/ the Resume API). This is
      * mostly used for directly calling resume from Java code or via JMX
      */
     public void resume() {
-        consumer.resume(consumer.assignment());
+        LOG.info("A resume request was issued and the consumer thread will resume after current processing has finished");
+        state = State.RESUME_REQUESTED;
```
essobedo, line 556: Maybe we should only allow to pause if the state is `PAUSED `?

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: This is intended, but I think I can do better. I want to be explicit about the state it is.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: Actually, I think it should be either `RUNNING` or `RESUME_REQUESTED` (or, basically !PAUSED) . There's always a chance that it could receive a pause request almost immediately after a pause one.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: Actually, I think it's better to keep it as is. Otherwise we'll just unnecessarily implement a  state machine without necessity (we can probably get rid of 1 of the states, though).

orpiske: I'll add a comment though. This, indeed, can be confusing.

oscerd (approved):

orpiske:
Per logs, the checkstyle issue is not on this code, so it's safe to merge.

essobedo:
BTW speaking of thread safety, I see several fields in this class that should be declared as `volatile` such as `connected`, `terminated` and `reconnect`

orpiske:
> BTW speaking of thread safety, I see several fields in this class that should be declared as `volatile` such as `connected`, `terminated` and `reconnect`

I don't think `connected` and `reconnect` should be set as volatile. All their read/write access are done within the scope of the same thread. It may be the case for `terminated`, though since it is read on the scope of the health check.

orpiske:
> > BTW speaking of thread safety, I see several fields in this class that should be declared as `volatile` such as `connected`, `terminated` and `reconnect`
> 
> I don't think `connected` and `reconnect` should be set as volatile. All their read/write access are done within the scope of the same thread. It may be the case for `terminated`, though, since it is read on the scope of the health check.

Oh no, they are leaking via the `isRecoverable()` which is called from the Health Check thread. 😯

orpiske:
Thanks for raising the warning, BTW.

PR #7503, opened 2022-04-27
CAMEL-18019: ensure some fields in the Kafka record fetcher are safe for concurrent read

orpiske:

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -65,9 +65,12 @@ public class KafkaFetchRecords implements Runnable {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaFetchRecords.class);
 
+    // There are a few of volatile fields here because they are usually read from other threads,
+    // like from the health check thread. They are usually only read on those contexts.
+
     private final KafkaConsumer kafkaConsumer;
     private org.apache.kafka.clients.consumer.Consumer consumer;
-    private String clientId;
+    private volatile String clientId;
     private final String topicName;
     private final Pattern topicPattern;
     private final String threadId;
@@ -77,13 +80,14 @@ public class KafkaFetchRecords implements Runnable {
     private final BridgeExceptionHandlerToErrorHandler bridge;
     private final ReentrantLock lock = new ReentrantLock();
     private CommitManager commitManager;
-    private Exception lastError;
+    private volatile Exception lastError;
     private final KafkaConsumerListener consumerListener;
 
-    private boolean terminated;
-    private long currentBackoffInterval;
-    private boolean reconnect; // must be false at init (this is the policy whether to reconnect)
-    private boolean connected; // this is the state (connected or not)
+    private volatile boolean terminated;
+    private volatile long currentBackoffInterval;
+
+    private volatile boolean reconnect; // The reconnect must be false at init (this is the policy whether to reconnect).
+    private volatile boolean connected; // this is the state (connected or not)
 
     private volatile State state = State.RUNNING;
 
@@ -490,7 +494,7 @@ public class KafkaFetchRecords implements Runnable {
         this.connected = connected;
     }
 
-    public boolean isReady() {
+    private boolean isReady() {
         if (!connected) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


oscerd (approved):

essobedo:
I see also a potential bug [here](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L104). Indeed the `pollExceptionStrategy` is set in the constructor using the value of the field `consumer` that is always `null` at this stage

essobedo:
One more[ potential issue](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L75), the map `lastProcessedOffset` seems to be read and write by different threads, It looks like it should be a `ConcurrentHashMap`

orpiske:
> What about `clientId` and `lastError`? They seem to be shared also

Hm, `clientId` seems to be written only within the scope of the same thread (in 

> I see also a potential bug [here](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L104). Indeed the `pollExceptionStrategy` is set in the constructor using the value of the field `consumer` that is always `null` at this stage



> I see also a potential bug [here](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L104). Indeed the `pollExceptionStrategy` is set in the constructor using the value of the field `consumer` that is always `null` at this stage

Please open a separate ticket for this one.

orpiske:
I think we should probably abstract the task state that is exposed to the HealthCheck, so it's easier to know what is being accessed elsewhere.

essobedo:
> > I see also a potential bug [here](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L104). Indeed the `pollExceptionStrategy` is set in the constructor using the value of the field `consumer` that is always `null` at this stage
> 
> Please open a separate ticket for this one.

https://issues.apache.org/jira/browse/CAMEL-18020

orpiske:
> One more[ potential issue](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L75), the map `lastProcessedOffset` seems to be read and write by different threads, It looks like it should be a `ConcurrentHashMap`

I think we could try to remove this altogether and move its functionality to the commit manager. This is a remainder for the earlier versions of this component (pre 3.7), so I believe what it is trying to do overlaps with what we have there.

orpiske:
We can also probably remove the reconnect variable at this point (or leave in the appropriate PollExceptionStrategy). Anyway, let's do it in baby steps so it's easier to bisect if we have problems.

orpiske:
> One more[ potential issue](https://github.com/apache/camel/blob/b5361d38f0d56590d151a654b11920c463beaa3b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java#L75), the map `lastProcessedOffset` seems to be read and write by different threads, It looks like it should be a `ConcurrentHashMap`

I logged this one here: https://issues.apache.org/jira/browse/CAMEL-18020 (as explained, I think this one should be reworked and moved to a more appropriate location within the code).

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -516,28 +520,23 @@ public boolean isReady() {
...
 
-    boolean isRecoverable() {
+    private boolean isRecoverable() {
         return (pollExceptionStrategy.canContinue() || isReconnect()) && isKafkaConsumerRunnable();
     }
 
-    long getCurrentRecoveryInterval() {
-        return currentBackoffInterval;
+    // concurrent access happens here
+    public TaskHealthState healthState() {
```
essobedo, line 536: private package instead of public as it seems to be an implementation detail?

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/TaskHealthState.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.apache.camel.util.TimeUtils;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+
+/*
+ * This one is used to avoid exposing methods from this class to the health checker. Fields and methods used
+ * used to build an instance of this class, must be made thread-safe (i.e.: most importantly, read fields
+ * should be marked as volatile).
+ */
+public class TaskHealthState {
```
essobedo, line 30: private package instead of public as it seems to be an implementation detail?

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: +1 from me, but let's leave it like this for now. I am considering moving the whole set of health check code into a separate package (or, at least, within each of `consumer` and `producer` packages).

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/TaskHealthState.java
orpiske: +1 from me, but let's leave it like this for now. I am considering moving the whole set of health check code into a separate package (or, at least, within each of `consumer` and `producer` packages).

orpiske:
I'm going to merge this first bit of changes. Please keep the reviews, comments, tickets going ...

PR #7508, opened 2022-04-27
CAMEL-17950: camel-debezium - Upgrade to 1.9

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17950

## Motivation

Debezium 1.9 is available so we would like to use it in Camel

## Modifications:

* Modifies all the related pom file to set the new version
* Propose the new generated class based on this new version


essobedo:
Depends on https://github.com/apache/camel/pull/7506

PR #7511, opened 2022-04-28
CAMEL-18023. camel-jbang command options from application.properties

mgubaidullin:
https://issues.apache.org/jira/browse/CAMEL-18023

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/PropertiesHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/PropertiesHelper.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/PropertiesHelper.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static Properties readProperties() {
+        File defaultsFile = new File(APPLICATION_PROPERTIES_FILE);
+        Properties properties = new Properties();
+        if (defaultsFile.exists()) {
+            try (FileInputStream fis = new FileInputStream(defaultsFile)) {
+                properties.load(fis);
+            } catch (IOException e) {
+                LOG.error("Error", e);
```
orpiske, line 104: I am probably nitpicking a bit, but ... maybe wouldn't it be better to also log the exception message? I.e.: 

```
LOG.error("Error reading defaults file: {}", e.getMessage, e); 
```

mgubaidullin: Thanks @orpiske

orpiske (approved):

PR #7521, opened 2022-04-28
CAMEL-18020: fix a possible NPE due to uninitialized consumer in poll exception strategies

orpiske:


orpiske:
The test failure is related to the other PR that fixed the flaky test.

PR #7523, opened 2022-04-28
(chores) camel-kafka: fix flaky idempotent test

orpiske:


orpiske:
The failure is unrelated to the fixed flaky test, so merging.

PR #7562, opened 2022-05-06
[CAMEL-18072] mongodb: add a type converter to convert from byte[] to Document

lburgazzoli:

### DIFF
--- a/components/camel-mongodb/src/main/java/org/apache/camel/component/mongodb/converters/MongoDbBasicConverters.java
+++ b/components/camel-mongodb/src/main/java/org/apache/camel/component/mongodb/converters/MongoDbBasicConverters.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import org.apache.camel.Converter;
 import org.apache.camel.Exchange;
 import org.apache.camel.converter.IOConverter;
+import org.apache.camel.support.ExchangeHelper;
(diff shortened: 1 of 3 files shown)


oscerd (approved):

components/camel-mongodb/src/main/java/org/apache/camel/component/mongodb/converters/MongoDbBasicConverters.java
```diff
@@ -78,6 +81,21 @@ public static Document fromFileToDocument(File f, Exchange exchange) throws Exce
         return fromInputStreamToDocument(new FileInputStream(f), exchange);
     }
 
+    @Converter
+    public static Document fromByteArrayToDocument(byte[] data) {
```
davsclaus, line 85: In some of the other converters we take in exchange as parameter, and use

ExchangeHelper.getCharsetName(exchange)

that defaults to utf-8

See for example IOConverter

lburgazzoli: Will fix it

lburgazzoli: done

davsclaus (approved):

PR #7575, opened 2022-05-09
CAMEL-18085 ManagementStatisticsLevel.RoutesOnly should not register processors

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-18085

Processor statistics should not be registered if `ManagedStatisticsLevelRoutesOnlyTest` is set.


oscerd (approved):

davsclaus (changes requested):
See comment in JIRA

PR #7577, opened 2022-05-10
Add doc for kafka transaction

zhfeng:

### DIFF
--- a/components/camel-kafka/src/main/docs/kafka-component.adoc
+++ b/components/camel-kafka/src/main/docs/kafka-component.adoc
@@ -423,5 +423,16 @@ from("kafka:my_topic?headerFilterStrategy=#myStrategy")
 
 `myStrategy` object should be subclass of `HeaderFilterStrategy` and must be placed in the Camel registry, either manually or by registration as a bean in Spring/Blueprint, as it is `CamelContext` aware.
 
+== Kafka Transaction
+
+You need to add `transactional.id`, `enable.idempotence` and `retries` in `additional-properties` to enable kafka transaction with the producer.
+[source,java]
+----
+from("direct:transaction")
+.to("kafka:my_topic?additional-properties[transactional.id]=1234&additional-properties[enable.idempotence]=true&additional-properties[retries]=5");
+----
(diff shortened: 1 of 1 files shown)


components/camel-kafka/src/main/docs/kafka-component.adoc
```diff
@@ -423,5 +423,16 @@ from("kafka:my_topic?headerFilterStrategy=#myStrategy")
 
 `myStrategy` object should be subclass of `HeaderFilterStrategy` and must be placed in the Camel registry, either manually or by registration as a bean in Spring/Blueprint, as it is `CamelContext` aware.
 
+== Kafka Transaction
+
+It needs to add `transactional.id`, `enable.idempotence` and `retries` in `additional-properties` to enable kafka transaction with the producer.
+[source,java]
+----
+from("direct:transaction")
+.to("kafka:my_topic?additional-properties[transactional.id]=1234additional-properties[enable.idempotence]=trueadditional-properties[retries]=5");
```
davsclaus, line 432: You need & to separate the options

components/camel-kafka/src/main/docs/kafka-component.adoc
```diff
@@ -423,5 +423,16 @@ from("kafka:my_topic?headerFilterStrategy=#myStrategy")
 
 `myStrategy` object should be subclass of `HeaderFilterStrategy` and must be placed in the Camel registry, either manually or by registration as a bean in Spring/Blueprint, as it is `CamelContext` aware.
 
+== Kafka Transaction
+
+It needs to add `transactional.id`, `enable.idempotence` and `retries` in `additional-properties` to enable kafka transaction with the producer.
```
davsclaus, line 428: It needs -> You need

zhfeng: Done

davsclaus (approved):

davsclaus:
I think it would be good if we could make configuring the TX options as new endpoint options so its easier and more tooling friendly. Additional properties is only intended for special and rare use-cases.

zhfeng:
Thanks @davsclaus 

- `transactional.id`->`addtional-properties[transactional.id]`
- `idempotence`->`additional-properties[enable.idempotence]`
- `retries`->`additional-properties[retries]`

Does it make sense?

davsclaus:
Yeah but name it transactionId without the dot in the uri option

PR #7584, opened 2022-05-10
CAMEL-18085 Introduction of ManagementMBeansLevel

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-18085

I added a property **mbeansLevel** with values:
* `Default` - Camel will register MBeans for the context, all the routes and the processors.
* `RoutesOnly` - Camel will register MBeans for the context and for all the routes. (not for any processor).
* `ContextOnly` -  Camel will register MBeans for the context (neither for any route nor for any processor).

I updated the doc and added JUnit tests (ManagedMBeansLevel*Test)

Corner cases:

- If a route is added when context is already running, MBeans are registered only if allowed (covered by `ManagedMBeansLevelContextOnlyWithAddedRouteTest` and `ManagedMBeansLevelDefaultWithAddedRouteTest`). 
- I didn't add any validation that MBeansLevel is compatible with statistics level. Current behavior is that statistics can not be kept for the resource without MBeans. (Covered by `ManagedMBeansAndStatisticsConflictTest`)


JiriOndrusek:
I added the configuration into `DefaultConfigurationProperties` and also prepared a PR for camel-spring-boot https://github.com/apache/camel-spring-boot/pull/557

PR #7595, opened 2022-05-11
CAMEL-18002

klease:
Prevent a reused Exchange with non-null UoW from causing loss of onCompletion
which is added in DefaultNettyHttpBinding.toCamelMessage().
This onCompletion releases the ByteBuf in the FullHttpRequest.


klease:
@davsclaus Do you think this is an acceptable solution?

davsclaus:
Fixed by CAMEL-18101.

Thanks Karen for showing the real problem and letting us find the real bug.

PR #7602, opened 2022-05-12
Fix netty tests

orpiske:
The false positive test failure on NettyHttpSimpleUriParametersTest has been causing a lot of Netty HTTP tests to be skipped. This is caused by reusing forks (which, apparently, prevents some of the leak detection to run correctly).

### DIFF
--- a/components/camel-netty-http/pom.xml
+++ b/components/camel-netty-http/pom.xml
@@ -170,4 +170,20 @@
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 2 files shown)


davsclaus:
I found the solution to fix the test, also the one that is disabled. So this PR can be closed

orpiske:
> I found the solution to fix the test, also the one that is disabled. So this PR can be closed

Oh, that's awesome news. These tests have been broken for a while. I'm going to close it, thanks!!!

PR #7605, opened 2022-05-12
CAMEL-17723 camel-ssh - Dynamic Header is not resolved in pollCommand.

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-17723

Hi @davsclaus , I implemented  the proposal from the https://issues.apache.org/jira/browse/CAMEL-17723 about raw-url (not encoded).
Simple JUnit covers the basic scenario. (`RawUrlPollEnrichTest`)
Do you think of any problems, which may be caused by not encoded urls in java DSL? (I was thinking about some possible security implications but I can not come with an example, where it could cause any damage)

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/builder/EndpointConsumerBuilder.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/builder/EndpointConsumerBuilder.java
@@ -29,9 +29,14 @@ import org.apache.camel.Expression;
  */
 public interface EndpointConsumerBuilder extends EndpointConsumerResolver {
     /**
-     * Builds the url of this endpoint. This API is only intended for Camel internally.
+     * Builds the encoded url of this endpoint. This API is only intended for Camel internally.
      */
     String getUri();
+
+    /**
+     * Builds the raw url of this endpoint. This API is only intended for Camel internally.
+     */
+    String getRawUri();
 
     /**
      * Adds an option to this endpoint. This API is only intended for Camel internally.
--- a/core/camel-core-model/src/main/java/org/apache/camel/builder/EndpointProducerBuilder.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/builder/EndpointProducerBuilder.java
@@ -29,9 +29,14 @@ import org.apache.camel.Expression;
  */
 public interface EndpointProducerBuilder extends EndpointProducerResolver {
     /**
-     * Builds the url of this endpoint. This API is only intended for Camel internally.
+     * Builds the encoded url of this endpoint. This API is only intended for Camel internally.
      */
     String getUri();
+
+    /**
+     * Builds the raw url of this endpoint. This API is only intended for Camel internally.
+     */
+    String getRawUri();
 
     /**
      * Adds an option to this endpoint. This API is only intended for Camel internally.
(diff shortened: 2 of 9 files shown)


davsclaus:
Yeah possible I need to take a closer look - but if we can switch to raw, then we should do it the other places where EndpointProducerBuilder#getUri and EndpointConsumerBuilder#getUri are used today

JiriOndrusek:
I'll search for such places and add them to this PR

JiriOndrusek:
I changed getUri -> getRawUri in several more places (all I could fine in java code)

davsclaus:
Thanks for updating the other place.

@JiriOndrusek I wonder if you could also do a unit test per EIP that has been changed, eg toD, wireTap, enrich ...

JiriOndrusek:
I will do that, thanks for pointing it out. I tried to make a test with ssh (based on the summary of the ticket) and it was possible but it required a lot of security stuff so I kept only seda test. Those EIPs which are directly modified should be tested I agree (it didn't occurred to me)

davsclaus:
@JiriOndrusek Camel 3.17.0 is to be build over the weekend - it would be good to have this included but if you dont have time today to work on the additional tests, then we have time for 3.18 which also is the LTS release (more important)

JiriOndrusek:
@davsclaus I'll try to prepare them during today. (I hope it should be possible) I'll comment here in case I'm not able to finish this

JiriOndrusek:
@davsclaus I started to work on random EIP (wireTap), but I'm facing a problem (probably because I'm missing some type of configuratio)

-  In PollEnrich case, as you can see i. e. [here](https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/ProcessorDefinition.java#L3129), PollEnricher saves raw url as expression and evaluates it later
- WireTapReifier (ad probably other reifiers) handles rawUri like an uri, as you can see i.e. [here](https://github.com/apache/camel/blob/main/core/camel-core-reifier/src/main/java/org/apache/camel/reifier/WireTapReifier.java#L59), therefore it is not **evaluated**

I'm probably missing some kind of configuration, which will switch `wireTap` to dynamic. Do you have an idea how it can be achieved?

JiriOndrusek:
I'll dig around to find a solution (it would be nice if the DSL could use dynamicUri for EIP), but it will take me probably more time.

davsclaus:
Also it seems that wiretap has some special "simple" support which it checks at line 72
boolean simple = LanguageSupport.hasSimpleFunction(uri);

I just wanted to ensure that if you use endpoint-dsl and have a simple function in the uri, that it still works.

JiriOndrusek:
Thanks for the tips!
I'll look at them.
(TBH I'm not sure why I reverted line 72 in WireTapReifier in my second commit, I'll recheck it again and covert with the proper test)

JiriOndrusek:
@davsclaus I'd say that both items are already prepared in this PR. Let me explain:

---

```
Because the endpoint-dsl is also changed (uri -> raw) then it would be good to have unit tests added there - you could potentially copy those examples you have and then make them use the EndpointRouteBuilder instead and build the URIs via that instead of normal URI strings
https://github.com/apache/camel/tree/main/dsl/camel-endpointdsl
```
There are 2 test in camel-endpointdsl (`RawUrlWireTapDslTest` and `RawUrlPollEnrichTest`), they cover all changes caused by uri -> raw

---

```
Also it seems that wiretap has some special "simple" support which it checks at line 72
boolean simple = LanguageSupport.hasSimpleFunction(uri);

I just wanted to ensure that if you use endpoint-dsl and have a simple function in the uri, that it still works.
```
Line 72 in wiretapReifier is covered by `RawUrlWireTapDslTest`. If the change in line 72 is reverted -> the test will fail (with "cannot convert simple function to integer").

---

I think that this misunderstanding was caused by the fact, that I separated the changes in 2 commits (with intension to make it easily understandable but the result was completely opposite) - I'm sorry for that.

davsclaus:
Ah perfect, LTGM

PR #7618, opened 2022-05-16
CAMEL-18110: DeliverSm message payload opt param fix

memaskal:
The complete issue description can be found here: https://issues.apache.org/jira/browse/CAMEL-18110

I want this issue to be merged with the camel-3.14.x (LTS) branch but I had some problems with the auto-generated files.

The first commit resolves this issue, the other 2 are some minor issues I found along the way.

### DIFF
--- a/components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppBinding.java
+++ b/components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppBinding.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.smpp;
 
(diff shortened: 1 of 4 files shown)


memaskal:
Another issue I found along the way and probably could be verified and resolved in a new issue !? 

SmppMessage.createBody is dead code. The SmppMessage object is created only from the SmppBinding
class when a deliverSm, AlertNotification, or dataSm arrives at Camel. Only the deliverSm is an instance of MessageRequest
and the org.apache.camel.component.smpp.SmppBinding#createSmppMessage(org.apache.camel.CamelContext, org.jsmpp.bean.DeliverSm) always sets the body, thus the createBody is never used.

davsclaus (approved):

PR #7645, opened 2022-05-19
CAMEL-18130: Fix move-file post-processing

mash-sap:
Evaluation of file-expressions work on the inMessage. This changes makes sure that the properties of a file are bound to the inMessage of a exchange. This makes sure that the file-expressions can be evaluated correctly and post-processing can be executed.


davsclaus:
Thanks for this PR. We used it for resolving this in a little bit cleaner way, and added your unit test.

mash-sap:
Thanks @davsclaus! We still use Camel 2.x: I will down- port this change to our 2.X fork. Do you see any problem in doing so (i.e. maybe this change is only possible in Camel 3.X)?

davsclaus:
I think you can backport this to 2.x, as the file and ftp components are fundamental similar in 3.x and 2.x.

PR #7664, opened 2022-05-21
CAMEL-18135: fixed incorrect sync commit of the async manual commit

orpiske:
This also includes a fix for a broken test case for the Async manual
commit.


orpiske:
Please don't merge even if tests are OK. I am waiting for user feedback.

ludovic-boutros:
Hi @orpiske , thank you for investigating this.
I just checked the code. Indeed, just recording the offsets in the cache seems to do the trick because it will be taken in account later on if everything is ok. I had to understand the code refactoring. 
On my laptop, the IT Case seems to fail sometimes.

I'm sorry to not have enough time to help you more on this. :(

orpiske:
> Hi @orpiske , thank you for investigating this. I just checked the code. Indeed, just recording the offsets in the cache seems to do the trick because it will be taken in account later on if everything is ok. I had to understand the code refactoring. On my laptop, the IT Case seems to fail sometimes.

Thanks for reviewing it. Let's merge as it is and see how it goes on CI. 

> 
> I'm sorry to not have enough time to help you more on this. :(

No worries. I appreciate the review, thanks!

PR #7681, opened 2022-05-26
[CAMEL-18146] Fixes bug in caml-kafka SslEndpointAlgorithm config

jakemehring:
Allows this property to be an empty string so that server host name verification can be disabled.

- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] If you're unsure, you can format the pull request title like `[CAMEL-XXX] Fixes bug in camel-file component`, where you replace `CAMEL-XXX` with the appropriate JIRA issue.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.


zhfeng (approved):

davsclaus:
This is not good to set option to an empty value like that, instead we should add a special value that is regarded as no algorightm, such as `"none"` etc.

davsclaus (changes requested):
See prev comment

davsclaus:
You can now use "none" or "false" to disable this

PR #7686, opened 2022-05-28
Dynamic handling of should continue and should handle evaluation in RedeliveryErroHandler.RedeliveryTask

graben:


graben:
Hi @davsclaus,

if you think this might be useful / looks good, I'll open a jira ticket.

Regards,
Benjamin

davsclaus:
At first thought I dont see this as good because there is already functionality for continue/handled that are well established for a long time.

This is an internal thingy that makes things more complex (yet another way of doing something).

Its also seldom use-case to control this via exchange properties.

Also a big reason is that for Camel 3 we have profiled and optimized the routing engine and the overhead of checking for a new exchange property that 99.999% are not in use is too high.

graben:
Hi @davsclaus, thanks for your feedback. IMHO, modelling exception handling is not powerful enough if you need more complex error handling. For example, if you need to evaluate DB error codes, you have no chance because you always get the same SQL exception class.

Unfortunately, the actual implementation of RedeliveryTask has no real API / factory that could be customized for special purposes. The easiest way I found was the suggested one with temporal exchange properties set by an OnExceptionOccurred processor. I tried to keep the change as most noninvasive as possible.

Maybe you see a better way?

davsclaus:
Thanks for the feedback, I will take a look later this week - since its inside the routing engine we need to be a bit careful what changes we introduce.

davsclaus:
Here is how you can do it today
https://github.com/apache/camel/commit/092010e1fc226d8c43bdaceab25d818a5130d134

graben:
Nice but does not fit for our use case. We use an generic error handling approach for > 1000 routes therefore modelling is not an option.

davsclaus:
You can also use a generic onException

graben:
Well, this is only one aspect of a broad approach implemented, yet. I'll give it a try maybe it fits.

graben:
I find no way to get an onException clause globally into my routes without modelling. It must be something like a factory way to get it shared by jar into multiple Spring-Boot applications.

graben:
Thx for the tip

davsclaus:
Did you have luck with the route configurations?

graben:
Not yet, errorhandling config and onexception policy seems to not play well in our environment. Needs further investigation. after my vacation next week.

PR #7688, opened 2022-05-30
CAMEL-18127: implement adapter auto-configuration in the resume API

orpiske:
This is the first part of CAMEL-18127 and the work to make the resume
api simpler to use (see apache/camel-examples#92 for details). 

There are a few things missing in this PR, that will be completed in
subsequent PRs (otherwise it would be too big): 

- Update the remaining components to implement auto-configuration for
  them
- endpoint self-configuration (CAMEL-18128)
- improve the serialization (CAMEL-18148) 
- generate the adapter configuration automatically (CAMEL-18152)


oscerd (approved):

orpiske:
> ❌ Finished component verification: **1 component(s) test failed** out of 7 component(s) tested

Test failure is unrelated to this change and a work-around is provided on #7690.

components/camel-atom/src/main/resources/org/apache/camel/resume/adapter
```diff
@@ -0,0 +1 @@
+org.apache.camel.component.atom.UpdatedDateFilter
```
davsclaus, line 1: You need to have ASF license header on these files too

davsclaus (approved):

components/camel-atom/src/main/resources/org/apache/camel/resume/adapter
orpiske: Thanks! I'll need to adjust the loader to filter out the license.

orpiske:
Please don't merge it yet. It needs a fix to load the config files w/ license information.

PR #7689, opened 2022-05-30
s390x jenkinsfile

Nayana-ibm:
Adding jenkinsfile to run jenkins job on s390x node using jdk11 
We have removed camel-kudu and docs module as these are not supported on s390x.

### DIFF
--- a/Jenkinsfile.s390x
+++ b/Jenkinsfile.s390x
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+def AGENT_LABEL = 's390x'
+def JDK_NAME = env.JDK_NAME ?: 'jdk_11_latest'
+
+def MAVEN_PARAMS = "-B -e -fae -V -Dnoassembly -Dmaven.compiler.fork=true -Dsurefire.rerunFailingTestsCount=2"
+
+pipeline {
+
+    agent {
+        label AGENT_LABEL
+    }
+
+    tools {
+        jdk JDK_NAME
+    }
+
+    environment {
+        MAVEN_SKIP_RC = true
+    }
+
+    options {
+        buildDiscarder(
+            logRotator(artifactNumToKeepStr: '5', numToKeepStr: '10')
+        )
+        disableConcurrentBuilds()
+    }
+
+    parameters {
+        booleanParam(name: 'CLEAN', defaultValue: true, description: 'Perform the build in clean workspace')
+    }
+
+    stages {
+
+        stage('Clean workspace') {
+             when {
+                 expression { params.CLEAN }
+             }
+             steps {
+                 sh 'git clean -fdx'
+                 // Remove camel-kudu and docs as these are not supported on s390x.
+                 sed -i '85d' pom.xml
+                 sed -i '212d' components/pom.xml
+           }
+        }
+
+        stage('Build & Deploy') {
+            when {
+                branch 'main'
+            }
+            steps {
+                sh "./mvnw -U $MAVEN_PARAMS -Dmaven.test.skip.exec=true clean deploy"
+            }
+        }
+
+        stage('Checks') {
+            steps {
+                sh "./mvnw $MAVEN_PARAMS -Psourcecheck -Dcheckstyle.failOnViolation=false checkstyle:check"
+            }
+        }
+
+        stage('Test') {
+            steps {
+                sh "./mvnw $MAVEN_PARAMS -Dmaven.test.failure.ignore=true -Dcheckstyle.skip=true verify"
+            }
+            post {
+                always {
+                    junit allowEmptyResults: true, testResults: '**/target/surefire-reports/*.xml'
+                    junit allowEmptyResults: true, testResults: '**/target/failsafe-reports/*.xml'
+                }
+            }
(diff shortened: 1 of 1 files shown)


orpiske:
> 

Hello, thanks for contributing this. I do have a question, though: is this for running on s390 at Apache? I don't see any s390 node there.

oscerd:
As far as I know we don't have s390 instances. So if this is used in a fork, it must be there. Upstream on ASF it makes less sense

orpiske:
> As far as I know we don't have s390 instances. So if this is used in a fork, it must be there. Upstream on ASF it makes less sense

+1. 

In the case you want to include that node into the Apache CI (which, IMHO, I would love that), so that we have a s390x slave, you probably need to create a ticket to the Apache Infra. Here's a sample ticket you can look at [1] (that was for adding ppc64le nodes). If you already have a ticket, it would be good to link it here. 

Thanks!


1. https://issues.apache.org/jira/browse/INFRA-22612

Nayana-ibm:
> > As far as I know we don't have s390 instances. So if this is used in a fork, it must be there. Upstream on ASF it makes less sense
> 
> +1.
> 
> @Nayana-ibm In the case you want to include that node into the Apache CI (which, IMHO, I would love that), so that we have a s390x slave, you probably need to create a ticket to the Apache Infra. Here's a sample ticket you can look at [1] (that was for adding ppc64le nodes for Kafka, but the process would be similar for Camel). If you already have a ticket, it would be good to link it here.
> 
> Thanks!
> 
>     1. https://issues.apache.org/jira/browse/INFRA-22612

We have already added s390x nodes to Apache CI 
listed here:  https://jenkins-ccos.apache.org/view/all/  (jenkins-s390x-1, jenkins-s390x-2, jenkins-s390x-3, jenkins-s390x-4)
issue: https://issues.apache.org/jira/browse/INFRA-21433

orpiske:
> > > As far as I know we don't have s390 instances. So if this is used in a fork, it must be there. Upstream on ASF it makes less sense
> > 
> > 
> > +1.
> > @Nayana-ibm In the case you want to include that node into the Apache CI (which, IMHO, I would love that), so that we have a s390x slave, you probably need to create a ticket to the Apache Infra. Here's a sample ticket you can look at [1] (that was for adding ppc64le nodes for Kafka, but the process would be similar for Camel). If you already have a ticket, it would be good to link it here.
> > Thanks!
> > ```
> > 1. https://issues.apache.org/jira/browse/INFRA-22612
> > ```
> 
> We have already added s390x nodes to Apache CI listed here: https://jenkins-ccos.apache.org/view/all/ (jenkins-s390x-1, jenkins-s390x-2, jenkins-s390x-3, jenkins-s390x-4) issue: https://issues.apache.org/jira/browse/INFRA-21433

Thank you for clarifying this! In this case, it's +1 from me.

Jenkinsfile.s390x
```diff
@@ -0,0 +1,103 @@
...
+
+        stage('Clean workspace') {
+             when {
+                 expression { params.CLEAN }
+             }
+             steps {
+                 sh 'git clean -fdx'
+                 // Remove camel-kudu and docs as these are not supported on s390x.
+                 sed -i '/camel-kudu/d' components/pom.xml
+                 sed -i '0,/docs/{//d;}' pom.xml
```
orpiske, line 60: Maybe it's easier to run: 
`mvn -pl "!camel-kudu,!docs" -P install` 

Are you able to check if that works?

orpiske (approved):
+1 from me, but I did add a note for checking if can avoid some part of the CI logic by skipping incompatible modules via Maven CLI.

Jenkinsfile.s390x
Nayana-ibm: yes. below command worked:
mvn -Dmaven.test.skip.exec=true clean install -pl \!docs,\!components/camel-kudu

I will make changes to file

Nayana-ibm: made changes. please review

orpiske: Looking good, thanks.

Nayana-ibm:
Thanks @orpiske 
Just wondering when s390x job will be executed on https://builds.apache.org/job/Camel/
Anything else needs to be done?

orpiske:
> Thanks @orpiske Just wondering when s390x job will be executed on https://builds.apache.org/job/Camel/ Anything else needs to be done?

I think I need to work with @oscerd to include that job on the Apache CI before the job is available. I'll check this today.

orpiske:
@Nayana-ibm actually, I could just copy/edit the CI configuration and added the job. You can watch it here: https://ci-builds.apache.org/job/Camel/job/Camel%20JDK11%20Build%20(S390x)/job/main/6/console

orpiske:
@Nayana-ibm FYI, the build seems stuck waiting for the s390x nodes to come online.

Nayana-ibm:
@orpiske Yes..nodes are offline right now. I will check with @gmcdonald

Nayana-ibm:
@orpiske nodes are online now.  
Build fails with an error: _Error: JAVA_HOME is not defined correctly._
Do I need to update jenkinsfile to set PATH or jenkins job configuration will handle this?

orpiske:
> @orpiske nodes are online now. Build fails with an error: _Error: JAVA_HOME is not defined correctly._ Do I need to update jenkinsfile to set PATH or jenkins job configuration will handle this?

I believe the JAVA_HOME has to be configured at the node configuration, no? If not, I think it will need to be set on the Jenkinsfile, indeed.

Nayana-ibm:
> > @orpiske nodes are online now. Build fails with an error: _Error: JAVA_HOME is not defined correctly._ Do I need to update jenkinsfile to set PATH or jenkins job configuration will handle this?
> 
> I believe the JAVA_HOME has to be configured at the node configuration, no? If not, I think it will need to be set on the Jenkinsfile, indeed.

I think it should be set on node configuration.  As I don't see any other camel jenkinsfile sets the JAVA_HOME

Nayana-ibm:
@orpiske Can you please trigger now?

orpiske:
> @orpiske Can you please trigger build now?

Failed with: 

```
Error: JAVA_HOME is not defined correctly.
  We cannot execute /home/jenkins/tools/java/latest11/bin/java
```

orpiske:
@Nayana-ibm BTW, if you want, feel free to ping me on Zulip chat and we can try to sort it out quicker. You can find me at https://camel.zulipchat.com/.

gmcdonald:
Looks like there is a bit more work to do on these nodes, taking a look now

Nayana-ibm:
issue is resolved now. Builds are executing https://builds.apache.org/job/Camel/job/Camel%20JDK11%20Build%20(S390x)/

PR #7692, opened 2022-05-31
(chores) camel-netty-http: fix test execution on Github actions

orpiske:


orpiske:
The failure is not on netty code, so this should be safe to merge.

PR #7693, opened 2022-05-31
camel-kafka test fixes

orpiske:

### DIFF
--- a/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAsyncManualCommitIT.java
+++ b/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAsyncManualCommitIT.java
@@ -16,8 +16,9 @@
  */
 package org.apache.camel.component.kafka.integration;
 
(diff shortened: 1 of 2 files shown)


components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAsyncManualCommitIT.java
```diff
@@ -74,8 +83,6 @@ public void after() {
         if (producer != null) {
             producer.close();
         }
-        // clean all test topics
-        kafkaAdminClient.deleteTopics(Collections.singletonList(TOPIC));
```
zhfeng, line 78: Why we need to remove these clean codes?

orpiske: In this case, we do want the data to remain in the topic between tests. If we delete the topic between test executions, then the 3rd test scenario will have no uncommitted data to fetch from.

PR #7694, opened 2022-05-31
Backport CAMEL-17949 and netty-http test fixes

orpiske:

### DIFF
--- a/components/camel-netty-http/pom.xml
+++ b/components/camel-netty-http/pom.xml
@@ -156,6 +156,11 @@
             <artifactId>camel-jetty</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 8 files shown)


orpiske:
Tests passed.

davsclaus (approved):
The HeaderFilterStrategy that some component uses is potentially a better and more general solution. But this strategy is also a little bit confusing to use for component developers.

orpiske:
> The HeaderFilterStrategy that some component uses is potentially a better and more general solution. But this strategy is also a little bit confusing to use for component developers.

Thanks. I created a ticket so we can take a look at that when time allow us: https://issues.apache.org/jira/browse/CAMEL-18158

PR #7696, opened 2022-05-31
Missing assertion fixes and proper manual IT test naming

orpiske:
This prevents bogus warnings for tests in SonarCloud.


orpiske:
> ❌ Finished component verification: **1 component(s) test failed** out of 10 component(s) tested

The failure is not related to this change (nats is not running well on GH actions).

PR #7697, opened 2022-05-31
Update camel-jbang.adoc

brunoNetId:
broken link to example caused by residual closing parenthesis sign


orpiske:
Thanks @brunoNetId

PR #7705, opened 2022-06-01
(chores) camel-aws2-kinesis: add IT for AWS Kinesis consumer

orpiske:


orpiske:
Checkstyle issue not caused by this change.

PR #7708, opened 2022-06-01
[CAMEL-18459] Fixes incorrect JmsSendDynamicAware parsing for destinations that starts with "schema://" and dose not have queue: or topic: specificed

artemse:


davsclaus:
This is potential in other like kafka, sjms, spring-rabbitmq etc that are also having this. Can you check these as well?

davsclaus:
Thanks for the PR I merged and backported to 3.14.x branch

PR #7713, opened 2022-06-02
[CAMEL-18159] SendDynamicAware of several components parses uri that starts with schema:// incorectly

artemse:


components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/FtpSendDynamicAwareTest.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.camel.component.file.remote;
```
davsclaus, line 1: This class looks like its not complete, or added in the git PR as a mistake

davsclaus, line 1: This class looks like its not complete, or added in the git PR as a mistake

artemse: Yes, the file was added by mistake. Removed it.

components/camel-rabbitmq/src/test/java/org/apache/camel/component/rabbitmq/RabbitMQSendDynamicAwareTest.java
```diff
@@ -0,0 +1,41 @@
+package org.apache.camel.component.rabbitmq;
```
davsclaus, line 1: You need to add the ASF license header in top of these new files. Look at one of the existing java source files and copy its header.

davsclaus (approved):
ASF license headers is missing in some files

artemse:
I believe it's last component with this problem.

davsclaus (approved):

PR #7714, opened 2022-06-02
CAMEL-18127: added adapter auto-configuration for AWS2 Kinesis

orpiske:
Demo (to be added to the examples repository)

[asciicast](https://asciinema.org/a/499053)


davsclaus:
How are the adapters discovered? Some kind of classpath discovery? If so we have the existing FactoryFinder that we use for everything in Camel ala components, languages, health checks, and other things.

The manifest files are stored in META-INF/services/org/apache/camel/...

I noticed that in this PR that it may not be the location for this?

davsclaus (approved):

PR #7715, opened 2022-06-02
CAMEL-14847: create a camel-jq component

lburgazzoli:


lburgazzoli:
@davsclaus @oscerd this is a first rough draft of a new language based on JQ (https://issues.apache.org/jira/browse/CAMEL-14847).

As it's been a while since I created a language/component, I'm asking for advice on how to do it properly :)

bom/camel-bom/pom.xml
```diff
@@ -1082,6 +1082,11 @@
         <artifactId>camel-jackson-avro</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.apache.camel</groupId>
+        <artifactId>camel-jackson-jq</artifactId>
```
davsclaus, line 1087: Maybe this is wrong as its called `camel-jq` in all the other places

components/camel-jq/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-support</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>net.thisptr</groupId>
+            <artifactId>jackson-jq</artifactId>
```
davsclaus, line 54: Add `version` here

components/camel-jq/src/generated/resources/META-INF/services/org/apache/camel/other.properties
```diff
@@ -0,0 +1,7 @@
+# Generated by camel build tools - do NOT edit this file!
```
davsclaus, line 1: I think the other.properties is wrong, maybe try to `rm -rf src/generated` to regen from fresh

components/camel-jq/src/main/java/org/apache/camel/language/jq/JqExpression.java
```diff
@@ -0,0 +1,244 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(JqExpression.class);
+
+    private final String expression;
+    private final Scope scope;
+
+    private ObjectMapper objectMapper;
+    private String resultTypeName;
+    private Class<?> resultType;
+
+    private boolean autoDiscoverObjectMapper;
```
davsclaus, line 39: I wonder if you want/need for these fine grained auto discover functions to be able to turn them on|off individually?

Would you for example turn off fromJq only? Or from serviceLoader only? And are these two built-in functions from JQ itself?

core/camel-core-model/src/main/java/org/apache/camel/model/language/JqExpression.java
```diff
@@ -0,0 +1,115 @@
...
+    public void setAutoDiscoverFunctionsFromServiceLoader(String autoDiscoverFunctionsFromServiceLoader) {
+        this.autoDiscoverFunctionsFromServiceLoader = autoDiscoverFunctionsFromServiceLoader;
+    }
+
+    public String isAutoDiscoverFunctionsFromJq() {
+        return autoDiscoverFunctionsFromJq;
+    }
+
+    /**
+     * Set if JQ functions should be discovered from JQ's json file.
```
davsclaus, line 99: What does this mean? Maybe the doc page should have more details? For users that dont know JQ to well. Does it mean you can have built-in functions in a JQ script file?

core/camel-core-model/src/main/java/org/apache/camel/model/language/JqExpression.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
+import org.apache.camel.spi.Metadata;
+
+/**
+ * Evaluates a JQ expression.
+ */
+@Metadata(firstVersion = "3.18.0", label = "language", title = "Jq")
```
davsclaus, line 30: title = `JQ`

core/camel-core-model/src/main/java/org/apache/camel/model/language/JqExpression.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
+import org.apache.camel.spi.Metadata;
+
+/**
+ * Evaluates a JQ expression.
```
davsclaus, line 28: Maybe do like jsonpath:

Evaluates a JQ expression against a JSON message body.

label = "language,json"

davsclaus:
Great work you have not lost the knowledge how to build new languages.

Also maybe add support for `headerName` that some languages have that makes it possible to supply the json in a header instead of message body. See JSonPath language

lburgazzoli:
@davsclaus I've fixed most of the findings (doc is still not up to date)

One additional question I have is that in the camel-jackson component, the Jackson Type converter [must explicitly be enabled ](https://github.com/apache/camel/blob/main/components/camel-jackson/src/main/java/org/apache/camel/component/jackson/converter/JacksonTypeConverters.java#L72).

I wonder if some basic conversion (i.e. InputStream/String/byte[]/Map <--> JsonNode) can be enabled by default.
The reason for that is that the jq extension requires the input to be a JsonNode and of course I can do my own set if/else but maybe it would be better to leverage the type converter, WDYT ? as alternative the doc can mention that the JacksonTypeConverters shuld be explicitly enabled.

davsclaus:
Ah yeah the problem is that its a fallback type converter that is too generic.

We can add explict type converters that uses `JSonObject` `JsonNode` or what there is that are 100% json then we can use regular type converters that are always enabled.

And we can modify the fallback to be for special use-cases.

You are welcome to create a JIRA about this

lburgazzoli:
> 
> You are welcome to create a JIRA about this

https://issues.apache.org/jira/browse/CAMEL-18165

davsclaus:
Out of the box converters added

davsclaus (approved):

oscerd (approved):

davsclaus (approved):

PR #7727, opened 2022-06-06
CAMEL-18167: camel-aws2-s3 - Show a message when download link url is…

zhfeng:
… not browser compatible


oscerd (approved):

components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Producer.java
```diff
@@ -579,6 +579,10 @@ private void createDownloadLink(Exchange exchange) {
                 .build();
 
         PresignedGetObjectRequest presignedGetObjectRequest = presigner.presignGetObject(getObjectPresignRequest);
+        if (!presignedGetObjectRequest.isBrowserExecutable()) {
```
davsclaus, line 582: So if every message is like this you have a INFO logging per message, that can spam the logs if you have a lot of files to upload.

Either reduce the logging level to DEBUG or we need a way to only log INFO once and then tell that this can happen for other files too. 

Maybe there can be a header added in the response that has this status? Then users can use this in their camel routes as well.

zhfeng: good point - `DEBUG` level makes sense and if `checksum` validation enabled, users have to send a signedPayload. such codes refers to https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/presigner/S3Presigner.html
```
// Send any request payload that is needed by the service (not needed when isBrowserExecutable is true)
     if (presignedRequest.signedPayload().isPresent()) {
         connection.setDoOutput(true);
         try (InputStream signedPayload = presignedRequest.signedPayload().get().asInputStream();
              OutputStream httpOutputStream = connection.getOutputStream()) {
             IoUtils.copy(signedPayload, httpOutputStream);
         }
     }

```
Is it possible to use a header to hold a `InputStream` value?

davsclaus: Oh I was thinking about a header to store a boolean whether it was browser compatible or not

Something ala:

setHeader("s3browsercompatible", true or false)

zhfeng: yeah, I get it. so if `s3browsercompatible` is true, user have to send some signed payload. I just wonder if we can pass it through the router?

davsclaus: Yes you can use a header for that

zhfeng: use a header to store `InputStream` value for signed payload? :)

zhfeng: Hi @davsclaus, I add a header `DOWNLOAD_LINK_BROWSER_COMPATIBLE` to store the staus. But I'm not very sure that it is possible to add another header to store a signed payload which type is `InputStream`. Is it safe to copy it when adding in message headers?

zhfeng:
```
presignedGetObjectRequest.signedPayload().ifPresent(payload -> {
           message.setHeader(AWS2S3Constants.DOWNLOAD_LINK_SIGNED_PAYLOAD, payload.asInputStream());
        });
```

davsclaus: Ah so what can / should we do? If the camel component is to get a url for download - so output is a String with the url

davsclaus: and if its not browser compatible then the url cannot really be used to download from a brwoser, and you need all this workaround code you pointed to?

And because the presigner is closed in camel, then I would not think you can return a header with an input stream as the stream would then be closed

So maybe we do

1) return nothing as a download url is not possible
2) add option to allow to eager download if not browser compabile, but this will then return a blob instead of an url and the returned data can be very big if the s3 file is very big (this can be confusing as either you get a url or a big payload)
3) something else ...

zhfeng: I'd like to add two headers

- `DOWNLOAD_LINK_HTTP_REQUEST_HEADERS` with `presignedGetObjectRequest.httpRequest().headers()`
- `DOWNLOAD_LINK_SIGNED_PAYLOAD` with `presignedGetObjectRequest.signedPayload().asUtf8String()` if present

I just find that signed payload could be read as `String` so it should be fine to be added into message headers. And if the users really need to enable `checksum validation` when getting the dowonload link, it is their repsonsibleity to get these two from the message headers and add them into the http request for downloading the file. We can add this information as a `NOTE` in our aws2-s3.adoc

WDYT?

davsclaus (approved):

PR #7733, opened 2022-06-07
CAMEL-17424: do handle Authorization/Authentication issues when failling to poll

orpiske:


orpiske:
I think we are good to go on this one.

PR #7734, opened 2022-06-07
(chores) Rename all manual tests

orpiske:
Manual tests commmonly use practices that deviate from the standard
good practices - with reason - but that causes them to be flagged in
SonarQube analysis. Renaming them should help us reduce the incidence of
false alarms in the code analysis.


davsclaus (approved):

orpiske:
The checkstyle issue is not caused by this change. I'll merge and keep an eye just in case.

PR #7756, opened 2022-06-09
CAMEL-18170 - Enabling to serviceAccountKey as base64 to google-drive…

rhuan080:
… component

Enabling to use service account key as base64.


davsclaus:
There should already be support for this out of the box
org.apache.camel.impl.engine.DefaultResourceResolvers.Base64Resolver

Does it not already work?

rhuan080:
Hi @davsclaus,

Yes! Actually, it is working. Thus, I think the issue was a misunderstanding caused by the option name. The BigQuery use the `serviceAccountKey` name and the google drive (and remaining components) uses `keyResource`. Thus, I think the changes are just about docs. 

I think we should rename the option `keyResource` to `serviceAccountKey`.

davsclaus:
> I think we should rename the option keyResource to serviceAccountKey.

Yeah if its for this then a better name is a good idea. You are welcome to rename the option.
We then need to add a note in the 3.18 upgrade guide at about this:
https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/camel-3x-upgrade-guide-3_18.adoc

rhuan080:
Updated! I`ll upgrade the guide! Thank you @davsclaus :-)

davsclaus (approved):

rhuan080:
I sent a PR to update the doc --> https://github.com/apache/camel/pull/7763

PR #7758, opened 2022-06-10
CAMEL-18180 - Force LF endings in .adoc files

rgala:
Fore LF line endings in .adoc files during checkout.


orpiske:
I think it's fine, but I'll need to double check this one manually.

orpiske:
Looks good. Thanks for the fix @rgala!

orpiske (approved):

PR #7761, opened 2022-06-10
[CAMEL-17130] camel-file - Use NIO

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-17130

Changes:
1. There are simple Java IO -> NIO upgrades such as replace File usage with Path (in places it actually helps and doesn't make the code less readable)
2. FileConsumer - changed pollDirectory() and listFiles() methods to stream processing
3. FileResumeSet - completely rewritten to suit the stream processing above
4. FileUtils - some methods oveloads for Path for convenient usage

Notes:
- I've kept File as a class that is being transferred in an Exchange. I tried to replace it with Path, but run into problems with converters. If it makes sence, we can add this change in a next step
- the changes might not handle corner cases well, since for example File#getName() might return an empty string and Path#getFileName() return null in the same case. I suppose this doesn't pose much problem so I left it as it is.

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
@@ -17,7 +17,6 @@
 package org.apache.camel.component.file;
 
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -29,6 +28,7 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.PosixFilePermission;
@@ -74,15 +74,15 @@ public class FileOperations implements GenericFileOperations<File> {
 
     @Override
     public boolean deleteFile(String name) throws GenericFileOperationFailedException {
-        File file = new File(name);
-        return FileUtil.deleteFile(file);
+        Path path = Path.of(name);
+        return FileUtil.deleteFile(path);
     }
 
     @Override
     public boolean renameFile(String from, String to) throws GenericFileOperationFailedException {
         boolean renamed;
-        File file = new File(from);
-        File target = new File(to);
+        Path file = Path.of(from);
+        Path target = Path.of(to);
         try {
             if (endpoint.isRenameUsingCopy()) {
                 renamed = FileUtil.renameFileUsingCopy(file, target);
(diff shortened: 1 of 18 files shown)


orpiske:
@rnetuka Nice one!

I'm wondering if it would make sense to plan for, in a future work, implement a better/more modern mechanism for watching for file notifications. That way, maybe it would not be necessary to poll all the time and, instead, use something like the [WatchService](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/WatchService.html). I believe this would be more resource efficient as well, as the JVM probably hooks to the native file system modification mechanism for each OS (i.e.: like inotify on Linux or something like that).

bmarcj:
Great idea! 

Using Java's NIO features exclusively opens the door to one day using something like the filesystem provider for s3 or minio e.g.
https://github.com/Upplication/Amazon-S3-FileSystem-NIO2

components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
```diff
@@ -423,38 +423,38 @@ private void keepLastModified(Exchange exchange, File file) {
...
+    private void writeFileByFile(Path source, Path target, Exchange exchange) throws IOException {
         // in case we are using file locks as read-locks then we need to use
         // file channels for copying to support this
-        String path = source.getAbsolutePath();
+        String path = source.toAbsolutePath().toString();
         FileChannel channel
                 = exchange.getProperty(asExclusiveReadLockKey(path, Exchange.FILE_LOCK_CHANNEL_FILE), FileChannel.class);
         if (channel != null) {
-            try (FileChannel out = new FileOutputStream(target).getChannel()) {
+            try (FileChannel out = new FileOutputStream(target.toFile()).getChannel()) {
```
bmarcj, line 446: `FileChannel.open(target)`?

davsclaus:
> @rnetuka Nice one!
> 
> I'm wondering if it would make sense to plan for, in a future work, implement a better/more modern mechanism for watching for file notifications. That way, maybe it would not be necessary to poll all the time and, instead, use something like the [WatchService](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/WatchService.html). I believe this would be more resource efficient as well, as the JVM probably hooks to the native file system modification mechanism for each OS (i.e.: like inotify on Linux or something like that).

There is already camel-file-watch

davsclaus:
Make sure camel-ftp and camel-jsch keeps working after changing in camel-file

orpiske:
> > @rnetuka Nice one!
> > I'm wondering if it would make sense to plan for, in a future work, implement a better/more modern mechanism for watching for file notifications. That way, maybe it would not be necessary to poll all the time and, instead, use something like the [WatchService](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/WatchService.html). I believe this would be more resource efficient as well, as the JVM probably hooks to the native file system modification mechanism for each OS (i.e.: like inotify on Linux or something like that).
> 
> There is already camel-file-watch

Yeah, I know. My point is whether we can, maybe, consolidate code/features/etc where possible and opportunistically reduce duplicated feature set / code. It would make it easier to test and maintain, IMHO.

components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
rnetuka: I've updated the code. Thanks!

davsclaus:
There are merge conflicts - can you update the PR thanks

oscerd (changes requested):
This probably needs to be tested manually.

gitokus:
In FileConsumer filtering filenames before creating GenericFile would make big performance benefit when processing network shares with lots of files.

oscerd:
@rnetuka can you please rebase?

davsclaus:
closing old and outdated PRs

PR #7771, opened 2022-06-13
Update test.adoc

yvesdm:
Fix typo in documentation. Return type is RouteBuilder and not RoutesBuilder

### DIFF
--- a/components/camel-test/camel-test/src/main/docs/test.adoc
+++ b/components/camel-test/camel-test/src/main/docs/test.adoc
@@ -42,7 +42,7 @@ public class SimpleMockTest extends CamelTestSupport {
     }
 
     @Override
-    protected RoutesBuilder createRouteBuilder() throws Exception {
+    protected RouteBuilder createRouteBuilder() throws Exception {
         return new RouteBuilder() {
             @Override
             public void configure() throws Exception {


components/camel-test/camel-test/src/main/docs/test.adoc
```diff
@@ -42,7 +42,7 @@ public class SimpleMockTest extends CamelTestSupport {
     }
 
     @Override
-    protected RoutesBuilder createRouteBuilder() throws Exception {
+    protected RouteBuilder createRouteBuilder() throws Exception {
```
essobedo, line 45: Seems to be correct as it is as you can see https://github.com/apache/camel/blob/main/components/camel-test/camel-test/src/main/java/org/apache/camel/test/junit4/CamelTestSupport.java#L878

yvesdm: Indeed, you a right. I through it was a mistake since we build a new RouteBuilder[] and I got an error from mi IDE.
So I first through a copy/paste typo mistake instead of a class required to be imported...
One of my learned of the day ;)

davsclaus:
Thanks closing this as the current docs is correct

PR #7782, opened 2022-06-14
s390x jenkinsfile update

Nayana-ibm:
Skip the test case execution of modules which are either not supported on s390x or vendor images are not available for s390x.
Also added - xpathExprGrpLimit and - xpathExprOpLimit parameters


orpiske (approved):
Thanks for your contribution!

PR #7802, opened 2022-06-16
(chores) test fixes and cleanups

orpiske:


orpiske:
> x Finished component verification: **1 component(s) test failed** out of 7 component(s) tested

This is not related to this build, PR, so it's safe to merge.

PR #7813, opened 2022-06-17
CAMEL-18201 Enhance CamelTestSupport to turn off context stopping

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-18201

This PR adds an option to turn  off the context stopping. It is possible by junit5's `ConfigurationProperty` (which falls back also to system variable - used in the junit test). If context does not stop automatically method `doSetUp` is not called if context already exists.

I added 2 tests to cover both option values (off and on).

### DIFF
--- a/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
+++ b/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
@@ -413,7 +413,7 @@ public abstract class CamelTestSupport
         }
     }
 
-    private void doSetUp() throws Exception {
+    protected void doSetUp() throws Exception {
         LOG.debug("setUp test");
         // jmx is enabled if we have configured to use it, or if dump route
         // coverage is enabled (it requires JMX)
@@ -706,7 +706,7 @@ public abstract class CamelTestSupport
         doStopCamelContext(context, camelContextService);
     }
 
-    private static void doStopCamelContext(CamelContext context, Service camelContextService) {
+    protected void doStopCamelContext(CamelContext context, Service camelContextService) {
         if (camelContextService != null) {
             if (camelContextService == threadService.get()) {
                 threadService.remove();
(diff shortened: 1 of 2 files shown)


JiriOndrusek:
WDYT? @davsclaus  This change will allow camel quarkus to run `CamelTestSupport` based tests (See https://github.com/apache/camel-quarkus/issues/3511) and will  allow users to reuse old camel tests in camel-quarkus JVM mode (without need of bigger refactor)..

davsclaus:
Hmmm so this seems a bit like the camelContextPerClass functionality that is already there.
And the name of this option is maybe a bit unclear as it will reuse the previous context in the following tests.

JiriOndrusek:
Functionality may look similar but it is not. PerClass functionality will stop context after each class. This option won't stop context after each class, it will never stop context.

As for the name, I see that it may be a bit unclear, what about `NEVER_STOP_CONTEXT`/`DONT_STOP_CONTEXT` or `SHARE_CONTEXT_FOR_ALL_TEST`/`SHARE_CONTEXT`?

davsclaus:
Maybe what CEQ needs is being able to override some protected methods in CamelTestSupport to control more fine grained what CEQ needs to do in the doSetup and teardown methods.

Then CEQ can override these methods and do something different.

Maybe the existing doStopCamelContext and setup is useable as-is, eg you can try to implement this in CEQ and then override doStopCamelContext to not stop camel, or only stop when really needed. And doSetup you can then skip calling super if already started.

Then maybe this is possible to just do in CEQ only and keep camel core as-is

JiriOndrusek:
Overriding method in CEQ was mt preferred solution, unfortunatelly method I need to override is private static ` private static void doStopCamelContext(`

If making it non-static protected is a better solution, I will use that. (I chose a configurable option, as I saw similar options there)

If you agree, I'll modify the PR to change method signature and allow overriding of it.

davsclaus:
Yeah its better to make the method protected

JiriOndrusek:
Thank you! I change only 2 signatures and keep 1 test covering the overriding of both methods.

davsclaus (approved):

davsclaus:
LGTM great work @JiriOndrusek

JiriOndrusek:
I'll check it.

JiriOndrusek:
I've found the reason in https://github.com/apache/camel/actions/runs/2515347416

```
[ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.17.1:validate (validate) on project camel-test-junit5: File '/home/runner/work/camel/camel/components/camel-test/camel-test-junit5/src/test/java/org/apache/camel/test/junit5/CamelTestSupporOneContextForAllTest.java' has not been previously formatted.  Please format file and commit before running validation! -> [Help 1]
```
I'fixed by running `mvn process-resources -Pformat`

PR #7841, opened 2022-06-21
CAMEL-18125 fix multi node Kafka resume strategy empty cache

orpiske:
In multi-node setups (clustering), loading the cache should also be delayed until the component is starting


orpiske:
Checkstyle issue is unrelated to this change, so merging.

PR #7845, opened 2022-06-21
CAMEL-17911: fix creation of duplicate apiProxy

klease:
The check for equality always failed because the configuration is only set on the superclass AbstractApiComponent but was read from Olingo2Component.
The same issue existed in camel-olingo4.
Modify AbstractOlingo2TestSupport to fix flaky test due to reused Camel context with lingering polling threads.


oscerd (approved):

davsclaus (approved):

davsclaus:
I also backported to 3.14.x branch

PR #7855, opened 2022-06-22
CAMEL-18218 Fix camel-jira not adding components field

claudio4j:
* the "components" field was not updated due to a bug in handling the field
* Added tests for the component field
* Fixed the "components" field description
* Updated google-guava accordingly to jira-rest-java-client-parent

https://issues.apache.org/jira/browse/CAMEL-18218


claudio4j:
`main pr build / build (11) (pull_request)` fails with `Error: Process completed with exit code 1.` but I couldn't find the error file.

orpiske:
> `main pr build / build (11) (pull_request)` fails with `Error: Process completed with exit code 1.` but I couldn't find the error file.

Thanks for your contribution!

Unfortunately the org restrictions don't allow public access to the log files, so we upload them on request. I have done that for you. 

[test-logs.zip](https://github.com/apache/camel/files/8963765/test-logs.zip)

It's a checkstyle issue. You can usually fix those by running the auto format with `mvn -Pformat,fastinstall install`

orpiske (approved):

components/camel-jira/src/main/java/org/apache/camel/component/jira/producer/AddIssueProducer.java
```diff
@@ -25,6 +27,7 @@
 import com.atlassian.jira.rest.client.api.domain.IssueType;
 import com.atlassian.jira.rest.client.api.domain.Priority;
 import com.atlassian.jira.rest.client.api.domain.input.IssueInputBuilder;
+import com.google.common.base.Strings;
```
davsclaus, line 30: Do not use google/guava for this - Camel has API for checking if a string is empty, or since Camel is now Java 11 based, you can use blank() method on Strings from the JDK itself

davsclaus (changes requested):

components/camel-jira/src/main/java/org/apache/camel/component/jira/producer/AddIssueProducer.java
davsclaus: org.apache.camel.util.ObjectHelper isEmpty method

As we need to backport to 3.14.x then we cannot use Java 11 APIs as 3.14 supports JDK8

claudio4j: Fixed that, thanks.

PR #7867, opened 2022-06-23
CAMEL-18217: debugger - Allow to suspend messages

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-18217

## Motivation

Up to now when using the BacklogDebugger there is no easy way to suspend all messages before ensuring that all breakpoints have been defined. This can particularly be a problem when debugging an application with a very short lifetime like a test.

## Modifications

* Define a system variable `CAMEL_DEBUGGER_SUSPEND` to suspend the message processing at startup
* Add methods `attach` and `detach` to respectively resume and suspend message processing when `CAMEL_DEBUGGER_SUSPEND` is set to `true` 
* Add `junit-pioneer` as test dependency to be able to set the value of an environment variable within the context of a Unit test.
* Add some doc to describe it

## Result

I could debug following the steps described in [the blog post](https://camel.apache.org/blog/2022/06/HowToUseCamelRouteTextualDebuggerWithUnitTest/) without the additional `sleep` by setting the environment variable and invoking manually `attach` (will be done by https://github.com/camel-tooling/camel-debug-adapter once this PR is merged).

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -355,6 +355,7 @@
     <jt400-version>11.0</jt400-version>
     <jta-api-1.2-version>1.2</jta-api-1.2-version>
     <junit-jupiter-version>5.8.2</junit-jupiter-version>
+    <junit-pioneer-version>1.7.1</junit-pioneer-version>
     <junit-toolbox-version>2.4</junit-toolbox-version>
     <junit-version>4.13.2</junit-version>
     <jxmpp-version>0.6.4</jxmpp-version>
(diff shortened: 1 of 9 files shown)


davsclaus (approved):

essobedo:
> Yeah - We may consider a
> 
> mvn camel:debug goal that can set this automatic, so users can run Camel in debug mode with this out of the box.
> 
> https://github.com/apache/camel/tree/main/tooling/maven/camel-maven-plugin/src/main/java/org/apache/camel/maven

@davsclaus Interesting, maybe it can be done in another PR, WDYT?

davsclaus:
Yeah in another PR

essobedo:
The ticket for this feature https://issues.apache.org/jira/browse/CAMEL-18223

PR #7874, opened 2022-06-24
Update split-eip.adoc

MKdir98:
fix misspelling (change multicast to split)


davsclaus:
Thanks for the PR. I cherry picked it to main branch as 3.17.x is a non LTS branch

PR #7903, opened 2022-06-27
(chores) camel-jms: removed a few Thread.sleep in tests

orpiske:

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsMultipleConsumersTest.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsMultipleConsumersTest.java
@@ -41,9 +41,6 @@ public class JmsMultipleConsumersTest extends CamelTestSupport {
         });
         context.start();
 
-        // give it a bit time to setup both topic listeners
-        Thread.sleep(2000);
-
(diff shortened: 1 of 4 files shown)


components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsMultipleConsumersTest.java
```diff
@@ -42,7 +45,7 @@ public void configure() {
         context.start();
 
         // give it a bit time to setup both topic listeners
-        Thread.sleep(2000);
+        Awaitility.await().atMost(2, TimeUnit.SECONDS).until(context::isStarted);
```
essobedo, line 48: After `context.start()`, `context.isStarted()` should return `true`, right? If so I guess it is a no op

davsclaus: Yeah correct - maybe the sleep can be removed all together

orpiske: Thanks, I'll give it a try tomorrow without the loop.

PR #7920, opened 2022-06-28
[DOCS] Update Bean and FHIR components.

mmajerni:


docs/components/modules/ROOT/partials/component-endpoint-options.adoc
```diff
@@ -70,7 +70,7 @@ ifeval::[{apicount} != 0]
 
 == API Parameters ({apicount} APIs)
 
-The @{doctitle} endpoint is an API based component and has additional parameters based on which API name and API method is used.
+The {doctitle} endpoint is an API based component and has additional parameters based on which API name and API method is used.
```
davsclaus, line 73: Is there a reason for this change? Because it could be that the `@` is on purpose as this is a partials file.

mmajerni: I thought that it was a typo.
Docs: https://camel.apache.org/components/3.14.x/fhir-component.html#_api_parameters_13_apis
Text: **The @FHIR endpoint is an API based component and has additional ...**

 I can reverse the change if it is intended.

davsclaus: Ah okay yeah then its a typo since its part of the generated doc.

davsclaus (approved):

PR #7928, opened 2022-06-29
CAMEL-18246: Upgrade hadoop-common >= 2.10.2, 3.3.3

aldettinger:

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -244,7 +244,7 @@
     <guava-eventbus-version>28.2-jre</guava-eventbus-version>
     <guice3-version>3.0</guice3-version>
     <hadoop3-guava-version>29.0-jre</hadoop3-guava-version>
-    <hadoop3-version>3.3.2</hadoop3-version>
+    <hadoop3-version>3.3.3</hadoop3-version>
     <hamcrest-version>2.2</hamcrest-version>
     <hapi-base-version>2.3</hapi-base-version>
     <hapi-fhir-version>4.2.0</hapi-fhir-version>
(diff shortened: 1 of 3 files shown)


davsclaus:
I think we can merge this as its good to upgrade on main branch. Then you can follow up with new PRs for more work if needed.

Is that okay?

aldettinger:
Ok, let's merge then.

orpiske:
@davsclaus @aldettinger I was talking to @aldettinger earlier today and I think this upgrade has broken the hbase component. 

We probably need to decide if we want to revert the upgrade or if there's still time to investigate and fix it: 

- Before: https://ci-builds.apache.org/job/Camel/job/Apache%20Camel/job/main/437/testReport/org.apache.camel.component.hbase/
- After: https://ci-builds.apache.org/job/Camel/job/Apache%20Camel/job/main/438/testReport/org.apache.camel.component.hbase/

aldettinger:
Oh, you mean even upgrading to hadoop-common 2.10.2 from this pr has broken the build.
Shouldn't we both revert AND investigate the fix then ?

davsclaus:
Ah yeah we can revert

It may be worth to check that the 2.10.2 in test-infra-hbase aligns so every dependency it uses are 2.10.2 also, as mixed versions could maybe cause this problem.

Also we can try to keep the 3.3.3 upgrade but revert the 2.10.x upgrade.

orpiske:
> Oh, you mean even upgrading to hadoop-common 2.10.2 from this pr has broken the build. Shouldn't we both revert AND investigate the fix then ?

Yeah, so per CI, it seems this one broke the hbase tests. I am fine with both approaches, although I don't really know how much time we have for 3.18.0 release (I guess we need to finish all the remaining fixes today).

davsclaus:
Yeah its fine to revert and then fixes can come in patch releases in the near future

aldettinger:
Ok, let's revert and I should probably have time to respin this next week.

PR #7943, opened 2022-06-30
CAMEL-18246: Upgrade hadoop-common version in camel-gora

aldettinger:

### DIFF
--- a/components/camel-gora/pom.xml
+++ b/components/camel-gora/pom.xml
@@ -72,6 +72,11 @@
 
         <!-- apache gora -->
         <dependency>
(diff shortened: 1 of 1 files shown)


aldettinger:
@orpiske Then I would do the same trick in camel-gora and all camel components would be covered.

What do you think ?

orpiske:
> @orpiske Then I would do the same trick in camel-gora and all camel components would be covered.
> 
> What do you think ?

Thanks for double checking. I don't know much about hadoop, but at first sight it LGTM.

PR #7964, opened 2022-07-01
CAMEL-18253: prevent the code from reporting an invalid number of duplicates

orpiske:
This removes the counter from the KafkaIdempotentRepository since duplicates are already correctly accounted for in the IdempotentConsumer class


orpiske:
Not a major one and it can wait .1 if there's no time.

davsclaus (approved):

PR #7980, opened 2022-07-06
CAMEL-18049 Fix autoregister when using Telegram webhook

zbendhiba:
The auto registration has moved from POST to GET in Telegram API.
I've tested this code with Telegram Token with this example : https://github.com/apache/camel-spring-boot-examples/tree/main/webhook

If this is Okay, this could be cherry-picked in LTS versions too, to correct the bug


orpiske:
@zbendhiba hello, we have just fixed the infra issue preventing the checkstyle from running. Please, would you mind rebasing the code and pushing it again? This should ensure the checks run with the updated code.

Thanks for you contribution!

oscerd (approved):

PR #7982, opened 2022-07-07
CAMEL-18237: fix incorrect round-to-zero

orpiske:
Resequence definition must use precise time definition, otherwise the
milliseconds get rounded to zero which causes tests such as the ones
in ResequenceStreamNotIgnoreInvalidExchangesTest to fail.

Ref: cf65ab21dbe3cfa598fcb83e7397c5c39718b3a4

----

This should fix some of our CI issues with the resequencer tests.


orpiske:
Checkstyle issue is unrelated, so it's safe to merge.

PR #7986, opened 2022-07-07
[CAMEL-18186] - Fix Camel-Saga tracing propagation

chiochuan:
When new exchange is being create during completion/compensation, previous exchange information is not copy. 
The changes is extract the tracer information from previous exchange and activate the new span when new exchange is created.

### DIFF
--- a/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfComponent.java
+++ b/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfComponent.java
@@ -43,7 +43,7 @@ public class CxfComponent extends HeaderFilterStrategyComponent implements SSLCo
     private Boolean allowStreaming;
     @Metadata(label = "security", defaultValue = "false")
     private boolean useGlobalSslContextParameters;
-    
+
     private Map<String, BeanCacheEntry> beanCache = new HashMap<String, BeanCacheEntry>();
 
     public CxfComponent() {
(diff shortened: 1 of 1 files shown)


orpiske (changes requested):
Thanks for your contribution. The code is not building. I believe you may need to rebase the code and adjust it to work with the recent changes.

Thanks!

core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
```diff
@@ -229,4 +235,11 @@ private <T> List<T> reversed(List<T> list) {
         Collections.reverse(reversed);
         return reversed;
     }
+
+    private void copyTrace(Exchange exchange, Exchange previousExchange) {
+        SpanAdapter span = ActiveSpanManager.getSpan(previousExchange);
```
zhfeng, line 240: I belive that `SpanAdapter` and `ActiveSpanManager` are in `camel-tracing` which can not be used here.

oscerd (changes requested):
I don't think this is correct at all, because you're using staff from camel-tracing.

core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
chiochuan: for this case, do you thing is just feasible if we just copy all the properties from old exchange to new exchange?

davsclaus (changes requested):
are you able to work on this ?

chiochuan:
> are you able to work on this ?

if copy the old exchange to new exchange without the exchangeId, will that work? i can try for this

PR #7991, opened 2022-07-08
CAMEL-18260: debugger - Reflect exchange changes on BacklogTracerEventMessage

essobedo:
fixes https://issues.apache.org/jira/browse/CAMEL-18260

## Motivation

While investigating on https://github.com/camel-tooling/camel-idea-plugin/issues/734, I realized that is was actually due to the fact that when a change is made on an exchange, the corresponding backlog tracer event message is not modified such that the change is not reflected in the debugger variables. 

## Modifications:

* Adds the new method `refreshBacklogTracerEventMessage` to refresh the content of the message when the exchange is modified
* Fixes warnings raised by IntelliJ (not related to the issue)

## Result

The changes are reflected directly on the `BacklogTracerEventMessage` as we can see in the screenshot below:

variables-updated-on-set-value

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogDebugger.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogDebugger.java
@@ -107,7 +107,7 @@ public final class BacklogDebugger extends ServiceSupport {
         private final CountDownLatch latch;
 
         /**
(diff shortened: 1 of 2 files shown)


oscerd (approved):

essobedo:
@oscerd Does it worth a backport in the 3.18 branch?

oscerd:
Yes, it makes sense to me

PR #8015, opened 2022-07-19
CAMEL-18275: Address issue of completions not being run in SEDA pipeline

klease:
When synchronizations are handed over, add a method to allow some housekeeping
to be performed. In the OnCompletionProcessor.OnCompletionSynchronizationAfterConsumer
this method stores the routeId on the Exchange.

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/SynchronizationVetoable.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/SynchronizationVetoable.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.spi;
 
+import org.apache.camel.Exchange;
+
 /**
  * A vetoable {@link org.apache.camel.spi.Synchronization}.
  * <p/>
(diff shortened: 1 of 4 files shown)


klease:
There is some analysis in the JIRA issue from Eduard and myself.

jeremyross:
My tests for this issue are passing with this PR. Also did some regression testing with another decent size integration. All tests pass.

davsclaus (approved):

oscerd (approved):

Gems:
Thank you for your effort, guys.
I'd like to understand the way the system is supposed to work after the fix.

Looking at the code I assume that all the completions will be executed when the SEDA pipelist gets completed. Is my assumption correct?

I want to add here that in case "before consumer" mode is used, I'd expect the completions to be executed before the message gets processed in the next queue of the defined pipeline. Does the fix provide such behavior?

klease:
>  Looking at the code I assume that all the completions will be executed when the SEDA pipelist gets completed. Is my assumption correct?

Yes, this is correct.
> I want to add here that in case "before consumer" mode is used, I'd expect the completions to be executed before the message gets processed in the next queue of the defined pipeline. Does the fix provide such behavior?

Unfortunately not. In fact, the current PR doesn't handle the before consumer mode at all. I can add this but unfortunately the completions are still called at the end of the entire pipeline. The change in the PR doesn't prevent the handover from occurring; it only ensures that when the handed-over completions execute, they will perform the logic for all the routes traveled by the exchange and not only the last one.

klease:
I think https://github.com/apache/camel/pull/8105 will make the beforeConsumer mode work as expected by preventing the handover.

PR #8023, opened 2022-07-24
CAMEL-18292: Addresses ujson.IncompleteParseException: exhausted input error when …

javaduke:
…body is null or empty

### DIFF
--- a/components/camel-datasonnet/src/main/java/org/apache/camel/language/datasonnet/DatasonnetExpression.java
+++ b/components/camel-datasonnet/src/main/java/org/apache/camel/language/datasonnet/DatasonnetExpression.java
@@ -114,12 +114,18 @@ public class DatasonnetExpression extends ExpressionAdapter implements Expressio
         }
 
         Document<?> body;
+
+        String bodyAsString = MessageHelper.extractBodyAsString(exchange.getMessage());
+
         if (exchange.getMessage().getBody() instanceof Document) {
             body = (Document<?>) exchange.getMessage().getBody();
+        } else if (exchange.getMessage().getBody() == null || "".equals(bodyAsString)) {
+            //Empty body, force type to be application/java
+            body = new DefaultDocument<>("", MediaTypes.APPLICATION_JAVA);
         } else if (MediaTypes.APPLICATION_JAVA.equalsTypeAndSubtype(bodyMT) || bodyMT == null) {
             body = new DefaultDocument<>(exchange.getMessage().getBody());
         } else {
-            body = new DefaultDocument<>(MessageHelper.extractBodyAsString(exchange.getMessage()), bodyMT);
+            body = new DefaultDocument<>(bodyAsString, bodyMT);
         }
 
         // the mapper is pre initialized
--- a/components/camel-datasonnet/src/test/resources/org/apache/camel/language.datasonnet/camel-context.xml
+++ b/components/camel-datasonnet/src/test/resources/org/apache/camel/language.datasonnet/camel-context.xml
@@ -191,6 +191,15 @@
             <to uri="mock:direct:end"/>
         </route>
 
+        <route id="transformEmptyBody">
+            <from uri="direct:transformEmptyBody"/>
+            <transform>
+                <datasonnet bodyMediaType="application/xml" outputMediaType="application/json"
+                            resultType="java.lang.String">{ "hello": "world" }</datasonnet>
+            </transform>
+            <to uri="mock:direct:end"/>
+        </route>
+
     </camelContext>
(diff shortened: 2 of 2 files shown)


components/camel-datasonnet/src/main/java/org/apache/camel/language/datasonnet/DatasonnetExpression.java
```diff
@@ -114,7 +114,11 @@ private Document<?> doEvaluate(Exchange exchange) {
         }
 
         Document<?> body;
-        if (exchange.getMessage().getBody() instanceof Document) {
+
+        if (exchange.getMessage().getBody() == null || "".equals(MessageHelper.extractBodyAsString(exchange.getMessage()))) {
```
davsclaus, line 118: This logic should be moved last as extracting body as string would then always be performed and that is not good

davsclaus (changes requested):
Its not good to

components/camel-datasonnet/src/main/java/org/apache/camel/language/datasonnet/DatasonnetExpression.java
javaduke: Oh, I see, but then how do I verify if the body is an empty string?

davsclaus: Ah yeah do you need to do this even if there is a Document object and its empty? Because if you convert that to a string you do an xml dump which would not be empty, it will have <xml ...> something as output.

The extractBody ... method deals with stream caching so that is good. It may not be as much overkill, just that the message body can be InputStream, byte[], XML Document and many other things, and in Java everything can be converted to a String via Object.toString so you would essentailly always get some output.

javaduke: No, if it's an empty Document, it's a different story (a Document cannot really be empty). Datasonnet doesn't like empty inputs, and I'm really just trying to deal with the situation where message body is not used in mapping but Datasonnet expects it anyway. If an input is an XML document and the mapping expects something else, there should be a legitimate error thrown.
So I really just need to make sure that if it's null or an empty string, we force the empty string with java object mime type, to avoid an error.

javaduke: OK, I changed the logic slightly, but I still need to ensure that if it's null or an empty string, I force APPLICATION_JAVA with empty string to DS.

components/camel-datasonnet/src/main/java/org/apache/camel/language/datasonnet/DatasonnetExpression.java
```diff
@@ -114,14 +114,20 @@ private Document<?> doEvaluate(Exchange exchange) {
...
+
         if (exchange.getMessage().getBody() instanceof Document) {
             body = (Document<?>) exchange.getMessage().getBody();
+        } else if (exchange.getMessage().getBody() == null || "".equals(MessageHelper.extractBodyAsString(exchange.getMessage()))) {
+            //Empty body, force type to be application/java
+            body = new DefaultDocument<>("", MediaTypes.APPLICATION_JAVA);
         } else if (MediaTypes.APPLICATION_JAVA.equalsTypeAndSubtype(bodyMT) || bodyMT == null) {
             body = new DefaultDocument<>(exchange.getMessage().getBody());
         } else {
             body = new DefaultDocument<>(MessageHelper.extractBodyAsString(exchange.getMessage()), bodyMT);
```
davsclaus, line 126: You would essentially call this method twice, first when checking for empty, and then afterwards here too. I think it would be good to do this only once

javaduke: Makes sense, I refactored it.

davsclaus (approved):

jam01:
DataSonnet does accept `null`, you can try this:
```
System.out.println(new Mapper("payload").transform(new DefaultDocument<>(null)).getContent());
// the output is the String "null"
```

Further, changing a `null` to empty string is a potentially breaking change for anyone that's checking for `null` payload.

I propose we roll this change back and figure out the root cause of the ujson exception. Perhaps checking for null body and if so setting the `DefaulDocument.NULL_INSTANCE`

@javaduke @davsclaus

javaduke:
Actually, I agree, upon further investigation I think it should be addressed in datasonnet, I'm on it.

PR #8035, opened 2022-07-25
(chores) ci: fixes invalid log instantiation

orpiske:
The hadoop-common dependency causes the code to instantiate a different
Logger type which, leads to an invalid evaluation of the log level. This
causes the exchange formatter to not run which causes the test to fail.

This was introduced by upgrading hadoop libraries in 7fdf3ac65fb8d57682a565b9f31a03d046a182e0

### DIFF
--- a/dsl/camel-endpointdsl/pom.xml
+++ b/dsl/camel-endpointdsl/pom.xml
@@ -83,6 +83,10 @@
                     <groupId>org.apache.drill.exec</groupId>
                     <artifactId>*</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.hadoop</groupId>
+                    <artifactId>hadoop-common</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>


orpiske:
@aldettinger @davsclaus I was investigating the current CI failure today and I found - after bisecting the code - that the log ones are related to #7928.

It looks like the new hadoop libraries cause a conflict which prevent the correct log class from being instantiated (see the attachments, please).

I am not entirely sure if this is the correct fix. Maybe it would be better to filter the exclusions directly on the camel-hbase component, but I couldn't get a good set of exclusions there. 

Do you mind to take a look/review this, please?

with-hadoop-common
without-hadoop-common

orpiske:
And these are the related CI failures: https://ci-builds.apache.org/job/Camel/job/Apache%20Camel/job/main/481/testReport/

oscerd (approved):

davsclaus (approved):

orpiske:
Thanks for the review, folks. I think @aldettinger is on PTO, so I believe he won't see this. I'll merge and follow up later. 

Thanks!

PR #8039, opened 2022-07-26
CAMEL-18299: use the FactoryFinder to lookup the adapter

orpiske:


davsclaus (commented):
This is better. After this can you look at @JdkService annotation that we use together with factory finder - then you can delete all the manual coded marker files as they will be auto-generated instead into src/generated.

And these marker files should be in META-INF/services/org/apache/camel/resume folder, for example as: 
https://github.com/apache/camel/blob/main/components/camel-debug/src/generated/resources/META-INF/services/org/apache/camel/debugger-factory

orpiske:
> This is better. After this can you look at @JdkService annotation that we use together with factory finder - then you can delete all the manual coded marker files as they will be auto-generated instead into src/generated.
> 
> And these marker files should be in META-INF/services/org/apache/camel/resume folder, for example as: https://github.com/apache/camel/blob/main/components/camel-debug/src/generated/resources/META-INF/services/org/apache/camel/debugger-factory

Thanks for the review. 

Re "JdkService": sure thing. I had originally created CAMEL-18152 for this, so I'll do it along with it in one of the upcoming PRs this week. Thanks!

oscerd (approved):

davsclaus (approved):

PR #8042, opened 2022-07-26
CAMEL-18300: camel-google-storage - Content-Type and other headers sh…

davsclaus:
…ould be stored via their specific setter than general metdata that leads to duplicate data in google cloud.


oscerd (approved):

mwajdzik:
There are multiple bugs introduced with this commit. Variables cd, ce, cc are not used - ct is used instead.

PR #8059, opened 2022-07-27
CAMEL-18315: camel-hyperledger-aries - Add support for single tenancy…

tdiesler:
… agent


components/camel-hyperledger-aries/src/main/java/org/apache/camel/converter/aries/ConnectionRecordConverter.java
```diff
@@ -0,0 +1,70 @@
+/*
```
davsclaus, line 1: There is a duplicated license header here

davsclaus (approved):

components/camel-hyperledger-aries/src/main/java/org/apache/camel/converter/aries/ConnectionRecordConverter.java
tdiesler: Fixed

davsclaus:
The GH action says there is a test error

tdiesler:
It seems that checkstyle issues (which should be fixed now) are reported as test errors

PR #8087, opened 2022-08-01
(chores) camel-endpointdsl: use unique data directory for tests

orpiske:
This helps prevents misbehaved tests from causing other tests to fail

### DIFF
--- a/dsl/camel-endpointdsl/src/test/java/org/apache/camel/builder/endpoint/BaseEndpointDslTest.java
+++ b/dsl/camel-endpointdsl/src/test/java/org/apache/camel/builder/endpoint/BaseEndpointDslTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.builder.endpoint;
 
+import java.util.concurrent.ThreadLocalRandom;
+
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

essobedo:
What about using the [TempDirectory extension](https://www.baeldung.com/junit-5-temporary-directory)?

orpiske:
> What about using the [TempDirectory extension](https://www.baeldung.com/junit-5-temporary-directory)?

I confess I don't know about this one. Let me take a look and see how it goes. (so, please, don't merge it yet).

essobedo:
If you don't want it to be merged, you should convert it to a draft

orpiske:
> What about using the [TempDirectory extension](https://www.baeldung.com/junit-5-temporary-directory)?

So, this is a very cool feature of JUnit. I didn't know about it. However, I won't do this one for now. We have a lot more code that could be converted / cleanup with this, so I want to look at it with more time (please, do feel free to create an issue on Jira if you want).

oscerd (approved):

PR #8089, opened 2022-08-02
chore : add a a special hapi-base-version so different versions of hapi-base and hapi-structures can be used (#7519)

cunningt:
cherry-pick over from main


cunningt:
(left the hapi.version upgrade from 4.1.0 -> 4.2.0 here

PR #8105, opened 2022-08-03
CAMEL-18275: prevent completion in beforeConsumer mode from being handed over

klease:


core/camel-core/src/test/java/org/apache/camel/processor/OnCompletionBeforeChainedSedaRoutes.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.camel.builder.RouteBuilder;
+import org.junit.jupiter.api.Test;
+
+public class OnCompletionBeforeChainedSedaRoutes extends ContextTestSupport {
+
+    @Test
+    public void testOnCompletionChained() throws Exception {
+        String body = "<body>";
+
+        getMockEndpoint("mock:end").expectedBodiesReceived(body);
```
Gems, line 29: I'd expect `"mock:end"` to arrive only after `"mock:cdone"`.

oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/processor/OnCompletionBeforeChainedSedaRoutes.java
klease: Due to the fact that the SEDA routes are run asynchronously, the order of arrival is indeterminate in the beforeConsumer mode. Also the order in which the expectedBodiesRecieved method is invoked does not control the order in which the assertions are actually evaluated. 
In the afterConsumer mode, the order of arrival is determinate: end, ddone, cdone, bdone, adone.

Gems: Indeed 👍🏻
I only later noticed that those are expectations for a set of different endpoints.

I'll write a test that delays processing on the second queue and ensures that the "on completion" gets executed meanwhile.

klease: That would be great. It was beyond my vacation-mode motivation. -:)

Gems: [Here](https://github.com/apache/camel/pull/8120) you go. Looks nice! Thank you for your effort 👍🏻

PR #8112, opened 2022-08-04
CAMEL-16287 camel-aws2-sqs should use pagination for deciding which aws sqs queues it should create

ladutsko:


oscerd (changes requested):
If this something for main too, please open the fix there and we'll backport appropriately

PR #8141, opened 2022-08-10
CAMEL-18377: camel-jpa - resue an EntityManager from the current tran…

zhfeng:
…saction if possible

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaHelper.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaHelper.java
@@ -60,6 +60,14 @@ public final class JpaHelper {
             em = getEntityManagerMap(exchange).get(getKey(entityManagerFactory));
         }
 
+        // then try reuse any entity manager from the transaction context
+        if (em == null && exchange != null && exchange.isTransacted()) {
+            Map<String, Object> data = getTransactionContextData(exchange);
+            if (data != null) {
+                em = (EntityManager) data.get(getKey(entityManagerFactory));
+            }
+        }
+
         if (em == null && useSharedEntityManager) {
             em = SharedEntityManagerCreator.createSharedEntityManager(entityManagerFactory);
         }
@@ -82,9 +90,25 @@ public final class JpaHelper {
             // we want to reuse the EM so store as property and make sure we close it when done with the exchange
             Map<String, EntityManager> entityManagers = getEntityManagerMap(exchange);
             entityManagers.put(getKey(entityManagerFactory), em);
+
+            // we want to reuse the EM in the same transaction
+            if (exchange.isTransacted()) {
+                Map<String, Object> data = getTransactionContextData(exchange);
+                if (data != null) {
+                    data.put(getKey(entityManagerFactory), em);
+                }
+            }
             exchange.adapt(ExtendedExchange.class).addOnCompletion(new JpaCloseEntityManagerOnCompletion(em));
         }
         return em;
+    }
+
+    private static Map<String, Object> getTransactionContextData(Exchange exchange) {
+        Map<String, Object> data = null;
+        if (exchange.isTransacted()) {
+            data = exchange.getProperty(Exchange.TRANSACTION_CONTEXT_DATA, Map.class);
+        }
+        return data;
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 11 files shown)


davsclaus (approved):

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaHelper.java
```diff
@@ -60,6 +62,15 @@ public static EntityManager getTargetEntityManager(
             em = getEntityManagerMap(exchange).get(getKey(entityManagerFactory));
         }
 
+        // then try to reuse one from the current transaction if possible
+        if (em == null && TransactionSynchronizationManager.hasResource(entityManagerFactory)) {
```
zhfeng, line 66: It's not good to reuse the one from JpaTransactionManager. I just wonder if there is any way to reuse one in current transaction

```java
if (em == null && exchange != null && exhange.isTransacted()) {
    // get em from the transaction context in camel?
}
```

@davsclaus any thoughts here?

davsclaus: When the splitter creates sub exchanges it copies from the parent.

To reuse the same transaction manager, then we could

1) Add a constant for this in Exchange (so its in core)
2) Copy over the existing TM in the sub exchange if its transacted (in the splitter) (see link below)
3) There are other EIPs that should do #2 (Multicast, Recipient List, Enricher) but not wiretap

https://github.com/apache/camel/blob/991f9ca7898bfea362821ef115d930fece317230/core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java#L231

zhfeng: Thanks @davsclaus - I'm thinking to share an EntityManager in the sub exchanges during the transaction. Can you take a review about the changes in Splitter.java? If it's OK, I will apply the same change to other EIPs.

core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
```diff
@@ -203,6 +205,8 @@ public Iterator<ProcessorExchangePair> iterator() {
                 private int index;
                 private boolean closed;
 
+                private final Map<String, Object> data = new ConcurrentHashMap<>();
```
davsclaus, line 208: You should only create this map if its transacted (otherwise its waste of cpu for regular non transacted). And the name data should be txData to better indicate its for transaction

core/camel-api/src/main/java/org/apache/camel/Exchange.java
```diff
@@ -254,6 +254,7 @@ public interface Exchange {
     String TRACING_HEADER_FORMAT = "CamelTracingHeaderFormat";
     @Deprecated
     String TRACING_OUTPUT_FORMAT = "CamelTracingOutputFormat";
+    String TRANSACTION_CONTEXT_DATA = "CamelTransactionContextData";
```
davsclaus, line 257: its a bit generic this name? Do you need to store something else than the EM factory?

TRANSACTION_MANAGER

or 

TRANSACTION_ENTITY_MANAGER

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaHelper.java
davsclaus: Yes I added some comments see below

core/camel-api/src/main/java/org/apache/camel/Exchange.java
zhfeng: Currently it only needs to store an EntityManager but I'd like to keept it more generic since it could be used for sharing any data in sub exchanges during the transaction.

core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
zhfeng: yeah, I will update it.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaHelper.java
zhfeng: I add the similar changes to `multicast` and `recipientList` processors. Aslo an unit test has been added to test with a pooling data source. For `enrich`, the default aggreate stragtegy is using the new exchange, so it works with transacted. But if the custom aggregate stragtegy is used, the users have to copy  `JpaConstants.ENTITY_MANAGE` propery from the newExchange by themseleves. I also provide `JpaHelper.copyEntityManagers` method to do this copying.

davsclaus (approved):

PR #8151, opened 2022-08-11
CAMEL-15520: remove unused deprecated classes in core

orpiske:


orpiske:
@davsclaus it might be good to double check this one. They seem unused and have been marked deprecated for a long time ...

davsclaus (approved):

PR #8156, opened 2022-08-12
HttpSendDynamicAware not optimizing for http:hostname[:port][/resourc…

jeroenbellen:
…eUri][?options]


davsclaus:
Thanks for the test - I found the bug and backported to 3.18.x too

PR #8163, opened 2022-08-15
Convert non-persistent JMS tests to the test infra

orpiske:
- this reworks several tests relying on a non-persistent broker to use the equivalent service in the test-infra
- moves some of the legacy code to test-infra
- minor test cleanups (mostly logs)
- avoid sharing resources as a pre-work for future parallelization

-----

Note: there's still a lot more to be done here, but given the number of
tests, let's do it in small steps.


davsclaus (approved):

orpiske:
Checkstyle failure is not on this code, so merging.

PR #8171, opened 2022-08-17
feat: add a global route builder interceptors

lburgazzoli:

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/builder/RouteBuilder.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/builder/RouteBuilder.java
@@ -18,6 +18,7 @@ package org.apache.camel.builder;
 
 import java.io.Reader;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


core/camel-core-model/src/main/java/org/apache/camel/builder/RouteBuilder.java
```diff
@@ -665,7 +666,10 @@ protected void checkInitialized() throws Exception {
                         camelContext.adapt(ExtendedCamelContext.class).getErrorHandlerFactory());
             }
 
-            for (RouteBuilderLifecycleStrategy interceptor : lifecycleInterceptors) {
+            Collection<RouteBuilderLifecycleStrategy> strategies = new ArrayList<>(lifecycleInterceptors);
+            strategies.addAll(camelContext.getRegistry().findByType(RouteBuilderLifecycleStrategy.class));
+
```
davsclaus, line 671: I wonder if we should sort these strategies, so a strategy can implement Ordered from Camel to influence if its run first or last etc.

davsclaus (approved):

core/camel-core-model/src/main/java/org/apache/camel/builder/RouteBuilder.java
lburgazzoli: done

PR #8190, opened 2022-08-19
Align to spring-boot 2.6.11

cunningt:
Hibernate version goes from 5.4.32.Final -> 5.6.10.Final - is that okay?    All the other version changes looked minor.


davsclaus (approved):

davsclaus:
Yes this is fine Hibernate is only used for testing purposes in camel-jpa

PR #8192, opened 2022-08-21
CAMEL-18131 - Adding health check to test conectivity

rhuan080:


components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3ComponentVerifierExtension.java
```diff
@@ -20,14 +20,7 @@
...
-import org.apache.camel.component.extension.verifier.ResultErrorBuilder;
 import org.apache.camel.component.extension.verifier.ResultErrorHelper;
-import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
-import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
-import software.amazon.awssdk.core.exception.SdkClientException;
-import software.amazon.awssdk.regions.Region;
-import software.amazon.awssdk.services.s3.S3Client;
-import software.amazon.awssdk.services.s3.S3ClientBuilder;
 
 public class AWS2S3ComponentVerifierExtension extends DefaultComponentVerifierExtension {
```
oscerd, line 25: I think we could also remove the class

oscerd (approved):

davsclaus:
Yes we can remove the component verifier which also means it should be removed on the component at
```
registerExtension(new AWS2S3ComponentVerifierExtension());
```

davsclaus (approved):

orpiske (approved):

davsclaus:
The PR keeps failing, wonder what that error is?

Also mind that this essentially makes the S3 component do health-checks with a live check to AWS every time its invoked. Not sure if AWS would rate limit you - meaning that the interval for triggering health checks should be a reasonable (5 or 10s or longer)

rhuan080:
@davsclaus I have tested it in my local environment and it works fine. I think it is some issue about the environment. Mainly because the issue is in another component that is not related to this update.

rhuan080:
Hi @orpiske,

Based on the file changed list, the files updated were just about the aws2-s3. Thus, I think I don`t need to rebase it. However, I will do that.

orpiske:
> Hi @orpiske,
> 
> Based on the file changed list, the files updated were just about the aws2-s3. Thus, I think I don`t need to rebase it. However, I will do that.

Screenshot 2022-08-30 at 19 16 51

Just as a matter of explanation: the `rebase` from upstream syncs your branch with the other changes that had been merged to `upstream/main`. Looking at your branch right, it was 227 commits behind `upstream/main`. When our diff script runs to find out what has changed, it compares the PR changes to `upstream/main`, therefore the greater the distance from main, the more will be tested (and this is why the label bot adds tags for jms, kafka and others even if the PR hasn't changed those).  

On a side note: I should probably add this to the contribution guide 🫣

PR #8212, opened 2022-08-24
CAMEL-18355: Fix bug when both proxy and basic authentication credent…

klease:
…ials provided

Add helper class to manage credentials to set on the HttpClient.
Add a unit test reproducing the issue.

### DIFF
--- a/components/camel-http/src/main/java/org/apache/camel/component/http/BasicAuthenticationHttpClientConfigurer.java
+++ b/components/camel-http/src/main/java/org/apache/camel/component/http/BasicAuthenticationHttpClientConfigurer.java
@@ -27,12 +27,15 @@ public class BasicAuthenticationHttpClientConfigurer implements HttpClientConfig
     private final String password;
     private final String domain;
     private final String host;
+    private final HttpCredentialsHelper credentialsHelper;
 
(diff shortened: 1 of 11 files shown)


davsclaus:
The HttpCredentialsHelper is static which means you cant have 2+ http components in the same JVM with different settings.

klease:
> The HttpCredentialsHelper is static which means you cant have 2+ http components in the same JVM with different settings.

True. I'll address that.

davsclaus:
If you do a hot reload updating routes that used http, then the map could be growing forever causing a memory leak.

A better approach is to store that help as a instance on the http component, then its private to this component. You can create the helper in the doInit method etc.

klease:
Thanks for the feedback Claus.
There isn't a doInit method on HttpComponent so I just create it in the method which creates the ClientBuilder and pass it to the private methods creating the configurers.
I also had to modify HttpComponentVerifierExtension.java which creates a ClientBuilder.
Perhaps it would be better to move the Helper to the main http package and give it package visiblity following these changes?

davsclaus:
Yeah if you think this is better and have time to do so

davsclaus (approved):

PR #8214, opened 2022-08-24
[CAMEL-18428]-Fixed password option description

pedrobalao:
Trivial PR, just a small correction on the FHIR component documentation


davsclaus:
You need to fix this in the java source code, those files you changed are automatic generated, and would therefore be changed back on next build.

pedrobalao:
@davsclaus Thanks for the review. I just updated the component configuration with the correct description.

davsclaus (approved):

PR #8221, opened 2022-08-26
Revert surefire + failsafe updates

orpiske:
I think Camel is affected by SUREFIRE-2096 (and, maybe, the related
issues mentioned on the ticket), causing longer test run times and the
fork timeout issues.


davsclaus (approved):

davsclaus:
Can you add a comment where we say M5 is buggy and update that to M7

orpiske:
> Can you add a comment where we say M5 is buggy and update that to M7

There is a comment on the pom file. Do you think it should be somewhere else? I can add it in another place too ...

orpiske:
Checkstyle issue not related to this, so it's safe to merge.

PR #8266, opened 2022-09-02
CAMEL-18452: Integrate azure-sdk-bom in a non-conflicting manner

bvahdat:

### DIFF
--- a/components/camel-jms/pom.xml
+++ b/components/camel-jms/pom.xml
@@ -39,7 +39,7 @@
 
         <camel.surefire.reuseForks>true</camel.surefire.reuseForks>
         <camel.surefire.forkCount>1</camel.surefire.forkCount>
-        <camel.surefire.forkTimeout>340</camel.surefire.forkTimeout>
+        <camel.surefire.forkTimeout>1200</camel.surefire.forkTimeout>
         <camel.surefire.parallel>true</camel.surefire.parallel>
         <camel.surefire.parallel.factor>0.5</camel.surefire.parallel.factor>
     </properties>


oscerd (changes requested):
This looks good to me. The concern I have mainly is about the other runtimes. Especially Karaf. It won't be possible to have all the needed bundles for each non release. So we either remove the features from Karaf or we find a way to support this approach

orpiske:
> ❌ Finished component verification: **1 component(s) test failed** out of 1 component(s) tested

One additional comment regarding this ... In the past I remember fixing conflicts caused by different Netty versions used in these vendor BOMs and Camel. We may also need to make sure it doesn't cause conflicts. 


The error in this PR kinda reminds me of the issues I had fixed in the past. I am not saying it is the cause, but it reminded me of those. 😅

bvahdat:
> This looks good to me. The concern I have mainly is about the other runtimes. Especially Karaf. It won't be possible to have all the needed bundles for each non release. So we either remove the features from Karaf or we find a way to support this approach

Why would we want to make use of the needed bundles not belonging to the same BOM as that would cause all sorts of problems, no mater if Karaf, Quarkus or SB runtime? If we would continue doing that, then we would run into the same issue as currently on the main branch described in CAMEL-18452 in regards to CAMEL-18451.

Let's compare azure-XXX versions in 3.18.1:
https://github.com/apache/camel/blob/camel-3.18.1/parent/pom.xml#L86-L96

With the corresponding azure-XXX bundle versions:
https://github.com/apache/camel-karaf/blob/main/pom.xml#L130-L140

They are out-of-sync. E.g. 

`<azure-eventhubs-version>5.11.2</azure-eventhubs-version>`

Versus:

`<azure-messaging-eventhubs-bundle-version>5.8.0_1</azure-messaging-eventhubs-bundle-version>`

But that doesn't need to be that way. If we rely on a given azure-sdk-bom version we can see in one shot what are the corresponding bundles we should come up with:

https://mvnrepository.com/artifact/com.azure/azure-sdk-bom/1.2.5

Just thinking about a given Apache Camel based integration where camel-azure-eventhubs + camel-azure-servicebus + camel-azure-blobstorage of a given Camel version are all inside the same POM and if their versions are not aligned / consistent then that integration would run into all sorts of problems at runtime with conflicting transitive dependencies.

bvahdat:
>  it is on the shoulders of servicemix and the release cycle timing is different

I see. I am not familiar with the SMX and it's release cycles, but I would expect if the intention is there to provide a bundle for azure-XXX, then that activity should always happen in the context of a given azure-sdk-bom version with all the other azure-YYY bundles belonging to the same azure-sdk-bom version. Otherwise it doesn't help much.

bvahdat:
> The error in this PR kinda reminds me of the issues I had fixed in the past. I am not saying it is the cause, but it reminded me of those.

Thanks @orpiske for your feedback. The root cause of the error was because of the ancient version of Azurite from February in use. The build is now successful.

BTW I find test-infra a tremendous added value to the Apache Camel project, KUDOS :raised_hands:

orpiske:
Thanks! I am glad that you found the issue!

PR #8281, opened 2022-09-05
CAMEL-18455: dozer project is deprecated and unmaintained

orpiske:

### DIFF
--- a/components/camel-dozer/pom.xml
+++ b/components/camel-dozer/pom.xml
@@ -28,7 +28,7 @@
 
     <artifactId>camel-dozer</artifactId>
     <packaging>jar</packaging>
-    <name>Camel :: Dozer</name>
+    <name>Camel :: Dozer (deprecated)</name>
     <description>Camel Support for the Dozer type conversion framework</description>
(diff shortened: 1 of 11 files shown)


components/camel-dozer/pom.xml
```diff
@@ -28,7 +28,7 @@
 
     <artifactId>camel-dozer</artifactId>
     <packaging>jar</packaging>
-    <name>Camel :: Dozer</name>
+    <name>Camel :: Dozer (Deprecated)</name>
```
davsclaus, line 31: It should be in lower-case, then we have tooling that detects this and mark it officially as deprecated.

oscerd (approved):

components/camel-dozer/pom.xml
orpiske: Ahhhh, nice. I had seem 2 classes previously marked as deprecated so I assumed we had done that in the past. 

I'll adjust. Thanks for the heads up.

davsclaus (approved):

PR #8301, opened 2022-09-06
CAMEL-16488: ensure resource setup/cleanup on the right time

orpiske:


davsclaus (approved):

orpiske:
Unrelated issues, so we are good to go.

PR #8303, opened 2022-09-06
[CAMEL-18472] lock 'model' in 'removeRoute' first to avoid deadlock

gerdailger:
Obtain the lock on `model` before locking `camelContext` in `removeRoute` to avoid a potential deadlock with concurrent `addRoutes` calls.

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/issues/DynamicallyConcurrentlyAddRoutesTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/issues/DynamicallyConcurrentlyAddRoutesTest.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.issues;
+
+import java.util.List;
+import java.util.UUID;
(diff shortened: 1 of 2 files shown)


core/camel-core/src/test/java/org/apache/camel/issues/DynamicallyConcurrentlyAddRoutesTest.java
```diff
@@ -0,0 +1,141 @@
...
+        Supplier<Callable<Boolean>> addRouteSupplier = () -> {
+            Callable<Boolean> addRouteTask = () -> {
+                String routeId = UUID.randomUUID().toString();
+                context.addRoutes(new RouteBuilder() {
+                    @Override
+                    public void configure() throws Exception {
+                        String endpointUri = "tmp:start_" + routeId;
+                        MySlowEndpoint mySlowEndpoint = new MySlowEndpoint(endpointUri, context);
+                        from(mySlowEndpoint).id(routeId).to("mock:result");
+                        log.info("Route " + endpointUri + " configured");
```
orpiske, line 65: Can you please use log markers here? As in:

log.info("Route {} configured", endpointUri);

orpiske (commented):
Thanks for the contribution. I've left one advisory comment. 

@davsclaus @oscerd I'm not sure if you are planning to test this one, but FYI I'll trigger some manual tests on this as the core code is not tested on PRs.

core/camel-core/src/test/java/org/apache/camel/issues/DynamicallyConcurrentlyAddRoutesTest.java
gerdailger: Oh yes of course, thank you!

orpiske (approved):
Manually tested with: 

- camel-ftp
- camel-jms
- camel-kafka
- camel-couchbase
- core/camel-core 

There were no issues.

core/camel-core/src/test/java/org/apache/camel/issues/DynamicallyConcurrentlyAddRoutesTest.java
orpiske: Thanks!

oscerd (approved):

oscerd:
Let's wait for @davsclaus

davsclaus:
The method `disposeModel` is used when in lightweight mode to clear the definition classes for runtimes where you do not dynamic add/remove routes.

So even if model is null then an exception would be thrown so that should be okay

davsclaus:
LGTM

davsclaus (approved):

PR #8310, opened 2022-09-07
AWS refresh secret

davsclaus:


davsclaus:
@oscerd here is a PR that adds AWS refresh as first-class. You need to set `camel.vault.aws.refresh-enabled=true` and it checks every 30 sec by default (there is a setting to custom that). Also you need to specify which secrets to trigger (we should auto-detect from property placeholder in another PR).

`camel.vault.aws.secrets=foo,bar,dude*`

its a pattern so you can use * for wildcards and comma for multi valued.

Maybe you can try with PR with the example
https://github.com/apache/camel-examples/tree/main/examples/aws/aws-secrets-manager-reloading

When we get AWS done then we can do similar for the other clouds.

davsclaus:
And we also need to update spring boot starter with the new config values

oscerd (approved):
Looks really good, thanks a lot! Can we also support more task? Like not aws-secret-refresh but selectable from multiple? We could maybe add a resolver or set an enum. Looks perfect by the way. Please merge and I'll try the example with this. Many thanks

oscerd:
> Looks really good, thanks a lot! Can we also support more task? Like not aws-secret-refresh but selectable from multiple? We could maybe add a resolver or set an enum. Looks perfect by the way. Please merge and I'll try the example with this. Many thanks

I think we could make it configurable, like having some kind of policy or strategy parameter for the task to use. But it's an improvement

davsclaus:
At the moment its one task per cloud, but we can look at adding support for multiple. Lets take one step at a time.

davsclaus:
merged to main, give it a try @oscerd and we can improve this in new PRs as we learn

PR #8322, opened 2022-09-07
CAMEL-18481: Attachment Handling: Allow the Mail Consumer to handle attachments wi…

mash-sap:
…th duplicate or missing filenames


mash-sap:
This is a first draft to enhance the attachment handling.

The property introduced in this change (handleDuplicateAttachmentNames) could also be extended to replace the recent enhancement (failOnDuplicateAttachment): https://issues.apache.org/jira/browse/CAMEL-17586
For example (handleDuplicateAttachmentNames=throwException)

mash-sap:
This change introduced 2 new settings to improve the attachment handling:
- generateMissingAttachmentNames: To generate a uuid as attachment name for attachments that have no name
- handleDuplicateAttachmentNames: To prefix attachment names with  a uuid in order to be able to add these attachment into the attachment map

mash-sap:
Hi everyone,

please provide some feedback to this PR.

There are still some checks failing. I will try to resolve them once I got some feedback for this PR.

Thank you &
Best regards,
Manuel

https://issues.apache.org/jira/browse/CAMEL-18481

davsclaus:
In reality how many emails have file attachments where the file name has NO name. 

Instead of trying to make a component has options for everything, then IMHO its better to keep them simpler but allow an interface (MailBinding) where end users can implement very very special use-cases.

mash-sap:
> In reality how many emails have file attachments where the file name has NO name.
> 
> Instead of trying to make a component has options for everything, then IMHO its better to keep them simpler but allow an interface (MailBinding) where end users can implement very very special use-cases.

Thanks for the feedback @davsclaus 

There is this Content-Type of message/rfc822 which has no name (one example can be found in the tests of this change). When saving a email (from outlook) and attaching this mail to a new mail, the Content-Type of this attached mail will be message/rfc822 and no attachment name is available.

davsclaus:
Thanks for the explanation

davsclaus (approved):

PR #8325, opened 2022-09-08
CAMEL-18456: consolidate Jetty usages in test code

orpiske:
Adds the test infra + converts camel-ahc-ws


oscerd (approved):

orpiske:
Checkstyle failing because the new test infra module is not on the repos yet. It should be fine to go.

PR #8368, opened 2022-09-14
CAMEL-18505 Camel-coap: Version conflict (californium-legal and californium-connector)

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-18505

I'm excluding californium-legal and element-connector with version 2.7.2 and adding it as 2.7.3 to prevent conflict.
(I can not change  <californium-version>2.7.2</californium-version> to 2.7.3, because org.eclipse.californium:element-connector-tcp-netty:jar:2.7.3 does not exists)

I used `californium-scandium-version` as a version number for   californium-legal and element-connector, I can not find a better name of that property. But in the future, once all californium dependencies will have the same version, property will be removed. So I think it is acceptable.

### DIFF
--- a/components/camel-coap/pom.xml
+++ b/components/camel-coap/pom.xml
@@ -44,6 +44,16 @@
             <groupId>org.eclipse.californium</groupId>
             <artifactId>californium-core</artifactId>
             <version>${californium-version}</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>element-connector</artifactId>
+                    <groupId>org.eclipse.californium</groupId>
+                </exclusion>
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus (approved):

JiriOndrusek:
The build error does not seem to be related to this change `Camel :: Box :: Component: Could not acquire write lock for '/home/runner/.m2/repository/.locks/commons-codec~commons-codec~1.15.resolverlock'`

davsclaus:
LGTM to merge

PR #8396, opened 2022-09-20
Adds tests to demonstrate bug within the toD optimisation

jeroenbellen:
The toD optimisation does not work when a path variable contains a space.

The HTTP component correctly encodes spaces to %20, but the HttpSendDynamicAware does not follow this logic.
This triggers an URISyntaxException because java.net.URI  (on line 168) does not accept an unencoded URL.

It might be worthwhile to add some logging when an URISyntaxException is handled. Currently, it goes unnoticed.


davsclaus:
Can you create a JIRA ticket about this

davsclaus:
I created a ticket
https://issues.apache.org/jira/browse/CAMEL-18544

jeroenbellen:
@davsclaus Thanks for the fast response and fix! Any changes that the fix can be included within the 3.14.x ?

PR #8447, opened 2022-09-28
CAMEL-18350: fix bug causing all messages to be reconsumed

klease:
When "breakOnFirstError" = "true", if an error occurs on the first message in a set of polled messages, then the offset is set to -1, causing all messages to be refetched and not only the one which failed. Modify KafkaFetchRecords to remember the result returned from the the processing of the previous batch when processing a new batch of polled messages.


davsclaus (approved):

orpiske (approved):
Thank you for the fix! LGTM

orpiske:
I believe we might need to backport this to 3.18.x

oscerd (approved):
Maybe on 3.14.x too.

orpiske:
> Maybe on 3.14.x too.

Indeed! 

I'm on PTO ... but, I can take care of it when I am back next week (unless, of course, @klease or someone else wants to do it - which if perfectly fine as well).

klease:
> Maybe on 3.14.x too.

I cherry-picked to 3.18.x  but that causes conflicts on 3.14.x; I'll look into it.

PR #8452, opened 2022-09-30
CAMEL-10173: camel-etcd3 - Add an implementation based on jetcd

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-10173

## Motivation

The API v3 of etcd should be supported by Camel and for this, we would like to propose a new component based on https://github.com/etcd-io/jetcd.

## Modifications:

* Add a new test infra for Integration tests with etcd v3
* Convert the test for the AggregationRepository into an integration test
* Propose a consumer and a producer based on jetcd and the legacy code.
* Propose a ServiceDiscovery based on jetcd and the legacy code.
* Propose a RoutePolicy based on jetcd and the legacy code. The leadership management relies on a lease with TTL.


oscerd (approved):

essobedo:
@lburgazzoli it would be great if you could review it as you know well jetcd, you are the reporter of the ticket and you seem to be very familiar with route policies. Thx in advance 🙏

essobedo:
@davsclaus @oscerd I've just proposed an asynchronous/non-blocking producer instead of a synchronous/blocking one because the client allows us to do it easily. What do you prefer in this case and in general?

oscerd:
There is no preferences in general. It's up to the contributor.

davsclaus (approved):

PR #8458, opened 2022-10-03
CAMEL-18575: use TempDir annotation from JUnit

orpiske:


orpiske:
Checkstyle issues caused by unpublished artifact, so merging.

PR #8462, opened 2022-10-04
(chores) Removed outdated Java 1.8 checks in tests

orpiske:

### DIFF
--- a/components/camel-xmlsecurity/src/test/java/org/apache/camel/component/xmlsecurity/ECDSASignatureTest.java
+++ b/components/camel-xmlsecurity/src/test/java/org/apache/camel/component/xmlsecurity/ECDSASignatureTest.java
@@ -59,8 +59,7 @@ public class ECDSASignatureTest extends CamelTestSupport {
 
     static {
         boolean includeNewLine = true;
-        if (TestSupport.getJavaMajorVersion() >= 9
-                || TestSupport.isJava18_261_later() && !TestSupport.isJavaVendor("Azul")) {
+        if (!TestSupport.isJavaVendor("Azul")) {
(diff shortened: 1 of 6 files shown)


oscerd (approved):

components/camel-xmlsecurity/src/test/java/org/apache/camel/component/xmlsecurity/ECDSASignatureTest.java
```diff
@@ -59,8 +59,7 @@ public class ECDSASignatureTest extends CamelTestSupport {
 
     static {
         boolean includeNewLine = true;
-        if (TestSupport.getJavaMajorVersion() >= 9
-                || TestSupport.isJava18_261_later() && !TestSupport.isJavaVendor("Azul")) {
+        if (TestSupport.getJavaMajorVersion() >= 9 || !TestSupport.isJavaVendor("Azul")) {
```
essobedo, line 62: It sounds like it is always true since we only support Java 11+, don't you agree?

essobedo (approved):

components/camel-xmlsecurity/src/test/java/org/apache/camel/component/xmlsecurity/ECDSASignatureTest.java
orpiske: Agreed. Let's simplify this one.

essobedo: For me, it is always true whatever the java vendor since it is a logical OR and we have the first operand that is true

PR #8475, opened 2022-10-04
(chores) NPE fixes in camel-mock

orpiske:


oscerd (approved):

orpiske:
FYI: I am testing this one on my own CI, so I'll merge it when/if it's successful.

essobedo:
Why not adding an assert directly into the method `getReceivedExchange`?
The assert could be the same as this one https://github.com/apache/camel/blob/1a10081dc71d483020482c51eda815e8af764ac8/components/camel-mock/src/main/java/org/apache/camel/component/mock/MockEndpoint.java#L958

WDYT?

orpiske:
> Why not adding an assert directly into the method `getReceivedExchange`? The assert could be the same as this one
> 
> https://github.com/apache/camel/blob/1a10081dc71d483020482c51eda815e8af764ac8/components/camel-mock/src/main/java/org/apache/camel/component/mock/MockEndpoint.java#L958
> 
> WDYT?

I think this is a good idea. I looked again and it seems it would not have an impact. Thanks!

PR #8383, opened 2022-09-15
CAMEL-18516: camel-yaml-dsl - bannedDefinitions doesn't work for gene…

igarashitm:
…rate-yaml-schema

Submitting a draft PR - currently it has test failures due to `FromDefinition` missing and looking into whether we have to adjust the test cases or remove `FromDefinition` from ban list.
```
[INFO] Running org.apache.camel.dsl.yaml.RouteConfigurationTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.004 s <<< FAILURE! - in org.apache.camel.dsl.yaml.RouteConfigurationTest
[ERROR] org.apache.camel.dsl.yaml.RouteConfigurationTest.route-configuration-id  Time elapsed: 0.003 s  <<< ERROR!
com.github.fge.jsonschema.core.exceptions.ProcessingException: 
fatal: JSON Reference "#/items/definitions/org.apache.camel.model.FromDefinition" cannot be resolved
    level: "fatal"
    schema: {"loadingURI":"#","pointer":"/items/definitions/org.apache.camel.model.RouteDefinition/properties/from"}
    ref: "#/items/definitions/org.apache.camel.model.FromDefinition"

	at com.github.fge.jsonschema.core.load.RefResolver.rawProcess(RefResolver.java:123)
	at com.github.fge.jsonschema.core.load.RefResolver.rawProcess(RefResolver.java:51)
	at com.github.fge.jsonschema.core.processing.RawProcessor.process(RawProcessor.java:77)
	at com.github.fge.jsonschema.core.processing.RawProcessor.process(RawProcessor.java:41)
	at com.github.fge.jsonschema.core.processing.ProcessorChain$ProcessorMerger.process(ProcessorChain.java:189)
	at com.github.fge.jsonschema.core.processing.ProcessingResult.of(ProcessingResult.java:79)
	at com.github.fge.jsonschema.core.processing.CachingProcessor$1.load(CachingProcessor.java:141)
	at com.github.fge.jsonschema.core.processing.CachingProcessor$1.load(CachingProcessor.java:133)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3628)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2336)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2295)
[...]

PR #8487, opened 2022-10-05
CAMEL-18516: camel-yaml-dsl - bannedDefinitions doesn't work for gene…

igarashitm:
…rate-yaml-schema

Following up #8383 
Cc: @lburgazzoli 

This fix enables the `bannedDefinitions` configuration for `generate-yaml-schema` as well. As a result, generated schema removes the `FromDefinition` which is not expected, so removed the `FromDefinition` from the `bannedDefinitions` in the POM for generating YAML schema.

`FromDefinition` is excluded when deserializers are generated here
https://github.com/apache/camel/blob/main/dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/pom.xml#L127
And the custom/handmade `FromDefinitionDeserializer` is honored
https://github.com/apache/camel/blob/main/dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/src/main/java/org/apache/camel/dsl/yaml/deserializers/FromDefinitionDeserializer.java
There's no `FromDefinitionDeserializer` in the generated deserializers
https://github.com/apache/camel/blob/main/dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/src/generated/java/org/apache/camel/dsl/yaml/deserializers/ModelDeserializers.java

So I think we don't need to exclude `FromDefinition` when generating the YAML schema.

### DIFF
--- a/dsl/camel-yaml-dsl/camel-yaml-dsl-maven-plugin/src/main/java/org/apache/camel/maven/dsl/yaml/GenerateYamlSchemaMojo.java
+++ b/dsl/camel-yaml-dsl/camel-yaml-dsl-maven-plugin/src/main/java/org/apache/camel/maven/dsl/yaml/GenerateYamlSchemaMojo.java
@@ -102,6 +102,11 @@ public class GenerateYamlSchemaMojo extends GenerateYamlSupportMojo {
                     .map(values -> Stream.of(values).collect(Collectors.toCollection(TreeSet::new)))
                     .orElseGet(TreeSet::new);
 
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@igarashitm I would need to test it because now I don't recall all the machinery.

davsclaus:
@igarashitm with this PR then the existing yaml schema files are the same as before ?

igarashitm:
Yes, verified the generated schema files stay unchanged.

davsclaus (approved):

davsclaus:
LGTM

PR #8489, opened 2022-10-06
CAMEL-18594: fixed a NPE on the master component

orpiske:
This prevents the component from taking leadership if a resume strategy is unset for the route.


orpiske:
Backport is not needed for this one.

PR #8492, opened 2022-10-07
CAMEL-18596: Use async checkpoint updating method from Azure EventCon…

jylipaa:
Invoke asynchronous methods and subscribe to them instead of calling blocking methods.

https://issues.apache.org/jira/browse/CAMEL-18596

### DIFF
--- a/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
+++ b/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
@@ -174,12 +174,13 @@ public class EventHubsConsumer extends DefaultConsumer {
             var completionCondition = processCheckpoint(exchange);
             if (completionCondition.equals(COMPLETED_BY_SIZE)) {
                 eventContext.updateCheckpointAsync()
-                        .subscribe(unused -> LOG.debug("Processed one event..."), error -> {
-                            LOG.debug("Error when updating Checkpoint: {}", error.getMessage());
-                            exchange.setException(error);
-                        }, () -> LOG.debug("Checkpoint updated."));
+                        .subscribe(unused -> LOG.debug("Processed one event..."),
+                                error -> LOG.debug("Error when updating Checkpoint: {}", error.getMessage()),
+                                () -> {
+                                    processedEvents.set(0);
+                                    LOG.debug("Checkpoint updated.");
+                                });
 
-                processedEvents.set(0);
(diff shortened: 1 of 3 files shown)


components/camel-azure/camel-azure-eventhubs/src/test/java/org/apache/camel/component/azure/eventhubs/EventHubsCheckpointUpdaterTimerTaskTest.java
```diff
@@ -0,0 +1,44 @@
+package org.apache.camel.component.azure.eventhubs;
```
oscerd, line 1: Please add license header

oscerd (changes requested):
Just a minor finding.

oscerd (approved):

orpiske:
> ❌ Finished component verification: **1 component(s) test failed** out of 1 component(s) tested

Please, can you rebase the code with the latest changes from main, so it can be tested automatically? Thanks!

orpiske (approved):

components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
```diff
@@ -173,7 +173,12 @@ private void processCommit(final Exchange exchange, final EventContext eventCont
         try {
             var completionCondition = processCheckpoint(exchange);
             if (completionCondition.equals(COMPLETED_BY_SIZE)) {
-                eventContext.updateCheckpoint();
+                eventContext.updateCheckpointAsync()
+                        .subscribe(unused -> LOG.debug("Processed one event..."), error -> {
+                            LOG.debug("Error when updating Checkpoint: {}", error.getMessage());
+                            exchange.setException(error);
```
davsclaus, line 179: Since the callback is async then in case of an error, then storing that on the exchange does not make sense as the exchange would likely no longer be in use. Instead the error should be logged, or better yet processed by exception handler on the consumer - there should be a getExceptionHandler.handleException(...)

davsclaus (changes requested):

components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
jylipaa: I decided to just log the error. Going deeper seems to be out of my experience level within this matter..

davsclaus (approved):

PR #8494, opened 2022-10-07
[CXF-8762] CXF client sends the SOAPAction header without quotes

rnetuka:
Issue: https://issues.apache.org/jira/browse/CXF-8762

Actually, there are two SOAPAction headers.
1. SOAPAction Camel header (stored top-level in the context)
2. SOAPAction Protocol header (stored in the context inside PROTOCOL_HEADERS)

The value inside PROTOCOL_HEADERS is used in CXF when creating a HTTP request. It should be quoted as stated in this specification. http://www.ws-i.org/Profiles/BasicProfile-1.1.html#SOAPAction_HTTP_Header

I believe we should leave the value stored top-level in the context unquoted, since it has nothing to do with the HTTP protocol. Because of that, I'm removing the quotes inside _propagateHeadersFromCxfToCamel_.

Any comment are welcome.

### DIFF
--- a/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/DefaultCxfBinding.java
+++ b/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/DefaultCxfBinding.java
@@ -25,6 +25,7 @@ import java.nio.charset.StandardCharsets;
 import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


davsclaus:
There is an util method to remove quotes at
https://github.com/apache/camel/blob/main/core/camel-util/src/main/java/org/apache/camel/util/StringHelper.java#L144

Can you use this method, thanks

ffang:
Hi @rnetuka ,

Looks good to me!

And please use the util method in StringHelper.java to remove the quote as @davsclaus mentioned about.

Thanks!
Freeman

rnetuka:
Thanks for the review guys! I've added the StringHelper's removeLeadingAndEndingQuotes method call as you suggested.

davsclaus:
I backported to 3.18.x

PR #8495, opened 2022-10-07
CAMEL-17438: camel-kubernetes - Add replace operation to producers

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-17438

## Motivation

We would like to be able to replace an existing resource type and for this, we would need to add the replace operation to all existing producers that support it.

## Modifications:

* Fix warnings and deprecations
* Improve the code coverage by adding tests to validate the behavior of mainly create operations
* Add replace operation to producers


oscerd (approved):

davsclaus (approved):

davsclaus:
yeah awesome we can now do CRUD with camel-kubernetes, great work guys

PR #8497, opened 2022-10-08
CAMEL-18582: Remove SocketTimeoutException from non retryable classes

essobedo:
Related to https://issues.apache.org/jira/browse/CAMEL-18582

## Motivation

With the latest fix, we now get errors of type `SocketTimeoutException: Read timed out` indicating that the runner loses regularly the connection with maven central for some period of time but as it can recover it, to workaround it, we need to make sure that it can retry when it faces this kind of error.

## Modifications:

* Use `default` as retry handler instead of `standard` to be able to change the non-retryable classes
* By default, `SocketTimeoutException` as a subclass of `InterruptedIOException` is part of the non-retryable classes which is the reason why no retries are made, so we need to redefine the non-retryable classes without `InterruptedIOException` to make sure that it will retry in case of `Read timed out`


oscerd (approved):

davsclaus (approved):

essobedo:
During the last build, we can see in the log file that several `read timed out`s occurred but the build still passed, so it should be good now

PR #8498, opened 2022-10-09
[camel-18587] It will only disconnect on isBreakOnErrorHit if consumer is not already paused

geekrupam:


davsclaus (approved):

orpiske (changes requested):
The change LGTM, but can you please adjust the commits so there's only one? Merge commits usually mess up with the log history and it would make it easier to me to back-port the fixes.

PR #8499, opened 2022-10-09
[camel-18588] Added condition around commitManager.forceCommit method to not reset from zero  if  there is no information on the last offset committed

geekrupam:


davsclaus (approved):

oscerd (approved):

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -131,8 +131,11 @@ private boolean processException(
                 LOG.warn("Will seek consumer to offset {} and start polling again.", partitionLastOffset);
             }
 
-            // force commit, so we resume on next poll where we failed
-            commitManager.forceCommit(partition, partitionLastOffset);
+            // force commit, so we resume on next poll where we failed except when the failure happened
+            // at the first message in a poll
+            if(partitionLastOffset != -1) {
```
orpiske, line 136: Please use `AbstractCommitManager.START_OFFSET` instead of -1, so it's easier to read. 


Thanks!

orpiske (changes requested):
LGTM, but requesting one small cosmetic change.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
davsclaus: Also the code should be formatted, you can do that with `mvn compile -P format`

PR #8501, opened 2022-10-10
[CAMEL-18555] Replace Shrinkwrap with maven-resolver

grgrzybek:


grgrzybek:
It's quite complex, but that's what Shrinkwrap was doing under the covers. Now we use maven-resolver (latest version 1.8.2) directly.
I squashed several commits, but I wanted to keep the one introducing `org.apache.camel.main.injection.DIRegistry` as separate one.

davsclaus (approved):

davsclaus:
Really great work, its a lot of code to use the real maven for dependency resolution. And now it will work even better with using maven security settings. Thanks for adding the docs for that too.

davsclaus:
I think one of the classes takes 1st place for the most java imports ;)

oscerd (approved):

PR #8512, opened 2022-10-11
CAMEL-16909: camel-kubernetes - Event support

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-16909

## Motivation

The events are not yet supported by the camel-kubernetes component and we would like to add its support.

## Modifications:

* Add a new component called `kubernetes-events` based on other components of the same type 
* Remove all classes from the package `org.apache.camel.component.kubernetes.consumer.common` as they are useless (not directly related to the initial issue)


oscerd (approved):

davsclaus (approved):

components/camel-kubernetes/src/main/java/org/apache/camel/component/kubernetes/events/KubernetesEventsConsumer.java
```diff
@@ -0,0 +1,140 @@
...
+                throw new RuntimeCamelException("Consumer label key or consumer resource name need to be set.");
+            }
+            watch = w.watch(new Watcher<>() {
+
+                @Override
+                public void eventReceived(Action action, Event resource) {
+                    Exchange exchange = createExchange(false);
+                    exchange.getIn().setBody(resource);
+                    exchange.getIn().setHeader(KubernetesConstants.KUBERNETES_EVENT_ACTION, action);
+                    exchange.getIn().setHeader(KubernetesConstants.KUBERNETES_EVENT_TIMESTAMP, System.currentTimeMillis());
```
orpiske, line 113: Just out of curiosity: wouldn't it be better/possible to use a more elaborate data type for the timestamp? 

My concern here is that this data may be confusing / hard to work with in scenarios where events may happen in different timezone.

Would it make sense or is it irrelevant in this context?

orpiske (approved):
LGTM - I did left a comment though.

components/camel-kubernetes/src/main/java/org/apache/camel/component/kubernetes/events/KubernetesEventsConsumer.java
essobedo: This timestamp is actually the same as other resource types, for the sake of consistency, I think that we should keep the same type. BTW, please note that you have a more specific type in the event itself stored in the message body.

orpiske: Thanks for clarifying!

PR #8521, opened 2022-10-11
CAMEL-18576: camel-core - Escape a placeholder

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-18576

## Motivation

The elastic search queries can contain double braces which are seen as a property placeholder by Camel, we need a way to escape it.

## Modifications:

* Propose to use the backslash character to escape double braces


davsclaus:
Thanks I will look closer tomorrow.

davsclaus (approved):

PR #8549, opened 2022-10-13
Cleanup jpa component

gilvansfilho:
Solving some pointings of sonar.

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
@@ -44,7 +44,7 @@ public class JpaPollingConsumer extends PollingConsumerSupport {
 
     private static final Logger LOG = LoggerFactory.getLogger(JpaPollingConsumer.class);
 
-    private transient ExecutorService executorService;
+    private volatile ExecutorService executorService;
     private final EntityManagerFactory entityManagerFactory;
     private final TransactionStrategy transactionStrategy;
     private String query;
@@ -137,19 +137,19 @@ public class JpaPollingConsumer extends PollingConsumerSupport {
                     entityManager.joinTransaction();
                 }
 
-                Query query = getQueryFactory().createQuery(entityManager);
-                configureParameters(query);
+                Query innerQuery = getQueryFactory().createQuery(entityManager);
+                configureParameters(innerQuery);
 
                 if (getEndpoint().isConsumeLockEntity()) {
-                    query.setLockMode(getLockModeType());
+                    innerQuery.setLockMode(getLockModeType());
                 }
 
-                LOG.trace("Created query {}", query);
+                LOG.trace("Created query {}", innerQuery);
 
                 Object answer;
 
                 try {
-                    List<?> results = query.getResultList();
+                    List<?> results = innerQuery.getResultList();
 
                     if (results != null && results.size() == 1) {
                         // we only have 1 entity so return that
--- a/components/camel-jpa/src/test/java/org/apache/camel/component/jpa/JpaComponentTest.java
+++ b/components/camel-jpa/src/test/java/org/apache/camel/component/jpa/JpaComponentTest.java
@@ -34,74 +34,80 @@ public class JpaComponentTest extends CamelTestSupport {
 
     @Test
     public void testJpaComponentConsumerHasLockModeType() throws Exception {
-        JpaComponent comp = new JpaComponent();
-        comp.setCamelContext(context);
-        assertNull(comp.getEntityManagerFactory());
-        assertNull(comp.getTransactionManager());
+        try (JpaComponent comp = new JpaComponent()) {
+            comp.setCamelContext(context);
+            assertNull(comp.getEntityManagerFactory());
+            assertNull(comp.getTransactionManager());
 
-        JpaEndpoint jpa
-                = (JpaEndpoint) comp.createEndpoint("jpa://" + SendEmail.class.getName() + "?lockModeType=PESSIMISTIC_WRITE");
-        JpaConsumer consumer = (JpaConsumer) jpa.createConsumer(null);
+            JpaEndpoint jpa
+                    = (JpaEndpoint) comp
+                            .createEndpoint("jpa://" + SendEmail.class.getName() + "?lockModeType=PESSIMISTIC_WRITE");
+            JpaConsumer consumer = (JpaConsumer) jpa.createConsumer(null);
 
-        assertEquals(LockModeType.PESSIMISTIC_WRITE, consumer.getLockModeType());
+            assertEquals(LockModeType.PESSIMISTIC_WRITE, consumer.getLockModeType());
+        }
     }
 
     @Test
     public void testJpaComponentCtr() throws Exception {
-        JpaComponent comp = new JpaComponent();
-        comp.setCamelContext(context);
-        assertNull(comp.getEntityManagerFactory());
-        assertNull(comp.getTransactionManager());
+        try (JpaComponent comp = new JpaComponent()) {
+            comp.setCamelContext(context);
+            assertNull(comp.getEntityManagerFactory());
+            assertNull(comp.getTransactionManager());
 
-        JpaEndpoint jpa = (JpaEndpoint) comp.createEndpoint("jpa://" + SendEmail.class.getName());
-        assertNotNull(jpa);
-        assertNotNull(jpa.getEntityType());
+            JpaEndpoint jpa = (JpaEndpoint) comp.createEndpoint("jpa://" + SendEmail.class.getName());
+            assertNotNull(jpa);
+            assertNotNull(jpa.getEntityType());
+        }
     }
 
     @Test
     public void testJpaComponentEMFandTM() throws Exception {
-        JpaComponent comp = new JpaComponent();
-        comp.setCamelContext(context);
-        assertNull(comp.getEntityManagerFactory());
-        assertNull(comp.getTransactionManager());
+        try (JpaComponent comp = new JpaComponent()) {
+            comp.setCamelContext(context);
+            assertNull(comp.getEntityManagerFactory());
+            assertNull(comp.getTransactionManager());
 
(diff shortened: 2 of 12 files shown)


gilvansfilho:
@orpiske as we talked about

orpiske:
Folks: I am helping @gilvansfilho understand about the contribution workflow for Camel and how to get started contributing with the project, so I'll take a look. Nonetheless, please do feel free to review and give insights about his contribution.

orpiske (approved):
LGTM! Thanks for your contribution!

oscerd (approved):

orpiske:
My only suggestion is regarding the commit message, that could be more specific with regards to the changes, so it makes it easier for us when bisecting or backporting.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaEndpoint.java
```diff
@@ -185,6 +185,16 @@ protected String createEndpointUri() {
         return "jpa" + (entityType != null ? "://" + entityType.getName() : "");
     }
 
+    @Override
```
davsclaus, line 188: I wonder why sonarcloud reports a problem that equals/hashCode needs to be overridden when they just call super?

gilvansfilho: Sonar cares about additional [subclass fields and the needeed of add them in equals and hash code methods](https://rules.sonarsource.com/java/RSPEC-2160). In that case this is not necessary as equality is based on context name and endpoint Uri so additional fields dont need to be considered. However sonar points as issue.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
```diff
@@ -44,7 +44,7 @@ public class JpaPollingConsumer extends PollingConsumerSupport {
 
     private static final Logger LOG = LoggerFactory.getLogger(JpaPollingConsumer.class);
 
-    private transient ExecutorService executorService;
+    private ExecutorService executorService;
```
zhfeng, line 47: why remove `transient` here? Is there an issue sonar reported?

gilvansfilho: Yes, there are. [If class is not `Serializable` then dont need to use transient](https://sonarcloud.io/project/issues?issues=AX4xQtpX2wn0z3gD78Js&open=AX4xQtpX2wn0z3gD78Js&id=apache_camel)

zhfeng: Hmm, it is interesting! I wonder if the modifier should be `volatile` since there might be multi polling consumers? @orpiske wdyt?

davsclaus: Yes correct it should be volatile

gilvansfilho: Should I change this?

davsclaus: Yes pleaase

gilvansfilho: This will introduce another sonar pointing as [`volatile` should be used only with primitive fields](https://rules.sonarsource.com/java/RSPEC-3077).

zhfeng: I think we could ignore this issue as `volatile` here is used to avoiding creating multi `ExecutorService` in `receive(long timeout)`.

orpiske: > I think we could ignore this issue as `volatile` here is used to avoiding creating multi `ExecutorService` in `receive(long timeout)`.

I don't think volatile is enough to protect the code in this particular scenario. A synchronized block / lock would be the correct to ensure this. AFAIK, it only guarantees the visibility updates of itself (and other variables declared around it) but cannot protect against concurrent writes of the variable. And, since we have 2 operations happening on that variable in that particular block (a fetch - from memory - and a write - to memory) it would still be entirely possible for 2 threads to concurrently fetch a null value at the same time.  

```
// need to use a thread pool to perform the task so we can support timeout
if (executorService == null) {
    executorService = getEndpoint().getComponent().getOrCreatePollingConsumerExecutorService();
}
```

That said ... I have seen this pattern used elsewhere in the code and I always wanted to fix it, but given a lot of other things we should take into consideration before fixing it, I've always left it aside as suggested fix for a major version. 

So, TL,DR: IMHO, `volatile` is not OK, but we can leave it be for a larger fix in the future.

zhfeng: I agree and is there CAMEL issue for tracking these fixes?

orpiske: I am not aware of any ticket tracking this item in particular. However, it is [listed on our SonarCloud instance](https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS3077&id=apache_camel).

Please, feel free to create a Jira ticket if you think we should track it.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaEndpoint.java
gilvansfilho: @orpiske `equals` and `hashCode` removed from last commit.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
gilvansfilho: @orpiske `volatile` added as with talked about

components/camel-jpa/src/test/java/org/apache/camel/component/jpa/JpaComponentTest.java
```diff
@@ -44,6 +44,7 @@ public void testJpaComponentConsumerHasLockModeType() throws Exception {
         JpaConsumer consumer = (JpaConsumer) jpa.createConsumer(null);
 
         assertEquals(LockModeType.PESSIMISTIC_WRITE, consumer.getLockModeType());
+        comp.close();
```
essobedo, line 47: What about using a try-with-resource statement instead?

gilvansfilho: Code will be something like: 

```
@Test
public void testJpaComponentConsumerHasLockModeType() throws Exception {
    try (JpaComponent comp = new JpaComponent()) {
        comp.setCamelContext(context);
        assertNull(comp.getEntityManagerFactory());
        assertNull(comp.getTransactionManager());
        JpaEndpoint jpa
            = (JpaEndpoint) comp.createEndpoint("jpa://" + SendEmail.class.getName() + "?lockModeType=PESSIMISTIC_WRITE");
        JpaConsumer consumer = (JpaConsumer) jpa.createConsumer(null);
        assertEquals(LockModeType.PESSIMISTIC_WRITE, consumer.getLockModeType());
    } catch (Exception e) {
        throw e; //or e.printStackTrace();
    }
}
```

Once we will not handle the exception I believe is not necessary use try-with-resource.

gilvansfilho:
Tests failed here too because error when downloading some dependencies: 
- https://repository.apache.org/snapshots/org/apache/camel/spi-annotations/3.20.0-SNAPSHOT/spi-annotations-3.20.0-20221018.103417-15.jar
- https://repository.apache.org/snapshots/org/apache/camel/camel-management-api/3.20.0-SNAPSHOT/camel-management-api-3.20.0-20221018.103417-15.jar
- https://repository.apache.org/snapshots/org/apache/camel/camel-support/3.20.0-SNAPSHOT/camel-support-3.20.0-20221018.103417-15.jar

All errors are like below:
```
[ERROR] Failed to execute goal on project camel-jpa: Could not resolve dependencies for project org.apache.camel:camel-jpa:jar:3.20.0-SNAPSHOT: The following artifacts could not be resolved: org.apache.camel:camel-support:jar:3.20.0-SNAPSHOT, org.apache.camel:camel-management-api:jar:3.20.0-SNAPSHOT, org.apache.camel:camel-util:jar:3.20.0-SNAPSHOT, org.apache.camel:camel-core-languages:jar:3.20.0-SNAPSHOT: Could not transfer artifact org.apache.camel:camel-support:jar:3.20.0-20221018.103417-15 from/to apache.snapshots (https://repository.apache.org/snapshots/): transfer failed for https://repository.apache.org/snapshots/org/apache/camel/camel-support/3.20.0-SNAPSHOT/camel-support-3.20.0-20221018.103417-15.jar: peer not authenticated -> [Help 1]
```

With wget I was able to download these dependencies and after put them on my local repo tests were successfully executed

components/camel-jpa/src/test/java/org/apache/camel/component/jpa/JpaComponentTest.java
essobedo: Why do you catch the exception ? You could simply throw it

gilvansfilho: That is true. But don't is a bit confusing use try without catch?

essobedo: No because the try-with-resources statement has been created for this purpose. Read this https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html for more details.

gilvansfilho: I know how to use this but I really find it a bit confusing. But no problem, I will do it.

PR #8553, opened 2022-10-14
[CAMEL-18612] Inconsistency in JsonPath component causes problems with databinding

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-18612

### DIFF
--- a/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathExpression.java
+++ b/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathExpression.java
@@ -16,9 +16,6 @@
  */
 package org.apache.camel.jsonpath;
 
-import java.util.Collection;
-import java.util.List;
-
 import com.jayway.jsonpath.Option;
 import org.apache.camel.CamelContext;
 import org.apache.camel.Exchange;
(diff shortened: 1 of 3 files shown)


oscerd (changes requested):
What happens with a list with a single element if you removed that block? It seems to be a corner case we won't cover anymore.

rnetuka:
@oscerd nothing, the result will be a list with a single element. A one-element list is still a list in my opinion. Is there some case when you actually _need_ the unwrapping?

oscerd:
As far as It seems there is no specific needs for that, but maybe I would git blame around to find something. Maybe there is some specific reason or something we don't need anymore.

rnetuka:
@oscerd I doubt this corner case ever was supported. When I try the reproducer from the issue, the list element unwraps as a LinkedHashMap and a type converter calls LinkedHashMap#toString to create the return value, which is not a valid Json string.

davsclaus:
Is there an unit test that can reproduce this that can be added to camel-jsonpath. As a PR that removes code claiming to fix a bug, then I would like to see if possible a testcase for such bug first.

rnetuka:
@davsclaus I've added a unit test. Without the changes, it throws an error. With the changes, it works.

rnetuka:
Based on the JIRA discussion, this has been replaced by https://github.com/apache/camel/pull/8571

PR #8554, opened 2022-10-14
(chores) camel-rest-swagger: fixing a few code violations from SonarCloud

gilvansfilho:
Fixed:
- shadowing member variables
- override equals in subclass that add fields

### DIFF
--- a/components/camel-rest-swagger/src/main/java/org/apache/camel/component/rest/swagger/RestSwaggerEndpoint.java
+++ b/components/camel-rest-swagger/src/main/java/org/apache/camel/component/rest/swagger/RestSwaggerEndpoint.java
@@ -308,14 +308,24 @@ public final class RestSwaggerEndpoint extends DefaultEndpoint {
         return (RestSwaggerComponent) getComponent();
     }
 
+    @Override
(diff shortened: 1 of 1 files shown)


components/camel-rest-swagger/src/main/java/org/apache/camel/component/rest/swagger/RestSwaggerEndpoint.java
```diff
@@ -308,14 +308,24 @@ RestSwaggerComponent component() {
         return (RestSwaggerComponent) getComponent();
     }
 
+    @Override
```
davsclaus, line 311: I wonder why sonarcloud reports a problem that equals/hashCode needs to be overridden when they just call super?

gilvansfilho: Sonar cares about additional [subclass fields and the needeed of add them in equals and hash code methods](https://rules.sonarsource.com/java/RSPEC-2160). In that case this is not necessary as equality is based on context name and endpoint Uri so additional fields dont need to be considered. However sonar points as issue.

PR #8562, opened 2022-10-17
Jdk17 support

gnodet:
Ongoing work for JDK 17 support


essobedo (approved):

davsclaus:
Ah we need to upgrade the maven archetypes to JDK11 as JDK8 is dropped since 3.14.x. 
That would be good to get into the 3.20 and 3.18.x releases.

davsclaus:
3.18.x is later this week, so would be good to do that first.
3.20.0 is Nov/Dec depending on how active we are on this release

oscerd (approved):

gnodet:
> 3.18.x is later this week, so would be good to do that first.
> 3.20.0 is Nov/Dec depending on how active we are on this release

I've pushed the commit to update the archetypes on 3.18.x and `main`.

I'm going to have a look at supporting java `record`s , however, those have been introduced in JDK 14. Is the goal to raise the requirement to JDK 17 (as done in Spring 6) , or do we want to support `record` without raising the minimum requirements ? @davsclaus @oscerd

davsclaus:
The goal is for camel 4 which is JDK17 as minimum

oscerd:
I would target them for Camel 4.

gnodet:
Closed in favour of #8579

PR #8571, opened 2022-10-18
[CAMEL-18612] Inconsistency in JsonPath component causes problems with databinding

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-18612

Please, check that I didn't miss any place to inlude/support the new _unpackArray_ option.

Also, the option name is only a proposal. Another proposals are:
- unwrapSingleArray (original)
- unwrapArray
- unwrapSingleElements
- unpackSingleArray
- unpackArray
- unpackSingleElements

Which is the best?

### DIFF
--- a/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathLanguage.java
+++ b/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathLanguage.java
@@ -38,6 +38,7 @@ public class JsonPathLanguage extends LanguageSupport implements PropertyConfigu
     private boolean allowSimple = true;
     private boolean allowEasyPredicate = true;
     private boolean writeAsString;
+    private boolean unpackArray;
     private String headerName;
     private Option[] options;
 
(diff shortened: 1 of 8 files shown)


components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathLanguage.java
```diff
@@ -134,8 +140,9 @@ public Expression createExpression(String expression, Object[] properties) {
         answer.setAllowSimple(property(boolean.class, properties, 2, allowSimple));
         answer.setAllowEasyPredicate(property(boolean.class, properties, 3, allowEasyPredicate));
         answer.setWriteAsString(property(boolean.class, properties, 4, writeAsString));
-        answer.setHeaderName(property(String.class, properties, 5, headerName));
-        String option = (String) properties[6];
+        answer.setWriteAsString(property(boolean.class, properties, 5, unpackArray));
```
davsclaus, line 143: setUnpackArray

davsclaus (changes requested):
We also need to add this new option in 
https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/model/language/JsonPathExpression.java

rnetuka:
@davsclaus What about https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/builder/ExpressionClause.java#L466 and  https://github.com/apache/camel/blob/main/core/camel-core-model/src/main/java/org/apache/camel/builder/ExpressionClauseSupport.java#L477?

davsclaus (approved):

davsclaus:
Yeah ideally we need a new method in those java-dsl classes to expose this new option. However its likely seldom in use, so you can just add 1 new method.

PR #8574, opened 2022-10-18
CAMEL-18593: Platform-http : add reverse proxy feature

zbendhiba:


oscerd (approved):

davsclaus:
A problem is that this is only in the vertx implementation so this will not work for others, such as with spring boot that uses servlet.

So it would be good to make this clear in the documentation that currently its only supported when using camel-platform-http-vertx.

zbendhiba:
> A problem is that this is only in the vertx implementation so this will not work for others, such as with spring boot that uses servlet.
> 
> So it would be good to make this clear in the documentation that currently its only supported when using camel-platform-http-vertx.

@davsclaus  I've just updated the documentation

davsclaus (approved):

davsclaus:
Hmm there is an error in components/camel-swagger-java maybe its already there and not due to this PR

zbendhiba:
Yes I don't think so, All I changed since last build is the doc file.
I wonder which test is failing, can't reproduce it locally

davsclaus:
@zbendhiba I wonder if you can send a PR for 3.18.x branch.

zbendhiba:
> @zbendhiba I wonder if you can send a PR for 3.18.x branch.

there it is : https://github.com/apache/camel/pull/8581

PR #8575, opened 2022-10-19
fix whatsapp doc, the "since" field to 3.19

claudio4j:
This component was merged after 3.18


oscerd (changes requested):
It needs to be modified here https://github.com/apache/camel/blob/main/components/camel-whatsapp/src/main/java/org/apache/camel/component/whatsapp/WhatsAppEndpoint.java#L40 and it will be updated during the build

PR #8579, opened 2022-10-19
Camel 4.x

gnodet:
Start working on camel 4.x


oscerd (changes requested):
Wasn't better to create directly a camel-4.x from main?

gnodet:
> Wasn't better to create directly a [camel-4](https://issues.apache.org/jira/browse/CAMEL-4).x from main?

I wasn't sure about that.  Do we prefer keep main for `3.x` line for some time and create a `camel-4.x` branch now ? Or create a `3.20.x` branch and keep main for `4.x.`...

For some time, keeping a PR for 4.x has the benefit of easily rebasing against master to keep up-to-date.  That avoids having to back port all changes, at the cost of having to merge conflicts during the rebase operation.

oscerd:
I think we could always rebase here, the problem is we're going to forget about it.

gnodet:
> I think we could always rebase here, the problem is we're going to forget about it.

So you'd rather back porting every change made to `main` manually to the `camel-4.x` branch ?

oscerd:
It's the same, we are going to forget to do that anyway. So let's try to rebase here everyday.

davsclaus:
I think its fine to keep a PR here in the beginning when gnodet is jump started on this. When we later ramp up and have more working on this then a branch is better.

A PR is maybe also easier to see what is changed etc.

orpiske:
I think one benefit of having a branch is that we can setup a CI job, so we can keep an eye on tests, code quality, etc as we progress. Given the large scope of the work, doing it from the beginning can save us from a lot of trouble far ahead.

That said, as 3.x is still moving at full speed, rebasing can be painful, indeed.

orpiske:
Oh, and I believe we might be able to automate the rebase work. At least a good part of it.

PR #8587, opened 2022-10-19
[camel-18627] added null check to check if there was previously committed offset before resuming

geekrupam:


davsclaus (approved):

oscerd (approved):

orpiske (approved):
LGTM, thanks for your contribution!

PR #8600, opened 2022-10-20
CAMEL-14832: Data format for SWIFT financial message conversion support

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-14832

## Motivation

Able to support SWIFT financial message conversion with Prowide Core library. This library has Apache License 2.0. Available from https://github.com/prowide/prowide-core.

## Modifications

* Add a new module called `camel-swift`
* Add 2 data formats which are `swiftMt` and `swiftMx` for respectively the SWIFT MT messages and the SWIFT MX messages.


core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/SwiftMtDataFormat.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import org.apache.camel.model.DataFormatDefinition;
+import org.apache.camel.spi.Metadata;
+
+/**
+ * Encode and decode SWIFT MT messages.
+ */
+@Metadata(firstVersion = "3.20.0", label = "dataformat,transformation,swift,mt", title = "SWIFT MT")
```
davsclaus, line 30: Label = mt is maybe too far. eg users will find this via swift

"dataformat,transformation,swift"

core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/SwiftMxDataFormat.java
```diff
@@ -0,0 +1,156 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
+import org.apache.camel.model.DataFormatDefinition;
+import org.apache.camel.spi.Metadata;
+
+/**
+ * Encode and decode SWIFT MX messages.
+ */
+@Metadata(firstVersion = "3.20.0", label = "dataformat,transformation,swift,mx", title = "SWIFT MX")
```
davsclaus, line 31: Same as prev comment

davsclaus (approved):

core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/SwiftMtDataFormat.java
essobedo: Just to make it clear, you would rather prefer that I remove the tags mt and mx, is that correct?

davsclaus: Yes correct - as we should have a limited set of labels. If you search for MT then you find that anyway via the name. A label is like a common label for N+ - not a unique label.

essobedo:
The checkstyle build fails because no snapshot of `camel-swift` has been published  yet

PR #8602, opened 2022-10-21
camel-fhir: Override commons-text version

jamesnetherton:


davsclaus (approved):

davsclaus:
We need to backport this to 3.18 and 3.14 branches

jamesnetherton:
> We need to backport this to 3.18 and 3.14 branches

There's another PR for 3.18.x:

https://github.com/apache/camel/pull/8603

I will cherry-pick the change back to 3.14.x after merging.

components/camel-fhir/camel-fhir-api/pom.xml
```diff
@@ -40,6 +40,11 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-support</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+            <version>${commons-text-version}</version>
+        </dependency>
```
essobedo, line 47: Even if it is not really necessary, I would tend to exclude it from `hapi-fhir-base` to make it clear that it is an override or at least add a comment to explain that it is an override because otherwise over time we could remove it by mistake since it is not directly called by the component. For the component, it is a runtime/transitive dependency.

jamesnetherton: Latest commit adds an exclusion + comment.

essobedo (approved):
@jamesnetherton Thx 🙏

PR #8611, opened 2022-10-24
(chores) Use log guards

orpiske:


orpiske:
The failure is unrelated, so it's safe to merge.

PR #8646, opened 2022-10-31
[CAMEL-18618] Milo component: Return status from write

rangoy:
- [ ] ? Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.

In our usecase we need to get the response from the server when writing values to OPC UA via Milo. 

I've tried to ask on [zulipchat|https://camel.zulipchat.com/#narrow/stream/257298-camel/topic/Result.20from.20camel-milo.20endpoint] and also create an issue on jira without any response.

I tried to run a clean build in the component directory but tests failed both on master and my branch

### DIFF
--- a/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientProducer.java
+++ b/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientProducer.java
@@ -91,7 +91,15 @@ public class MiloClientProducer extends DefaultAsyncProducer {
         final Boolean await = msg.getHeader(MiloConstants.HEADER_AWAIT, this.defaultAwaitWrites, Boolean.class);
 
         if (TRUE.equals(await)) {
-            future.whenComplete((v, ex) -> async.done(false));
+            future.whenComplete((result, throwable) -> {
+                        if (throwable != null) {
+                            msg.getExchange().setException(throwable);
+                        } else {
+                            msg.setBody(result);
+                        }
+                        async.done(false);
+                    }
+            );
             return false;
         } else {
             async.done(true);
--- a/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/internal/SubscriptionManager.java
+++ b/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/internal/SubscriptionManager.java
@@ -904,7 +904,7 @@ public class SubscriptionManager {
                 if (e != null) {
                     handleConnectionFailure(e);
                 }
-                return null;
+                return status;
             }, this.executor);
         }
     }


davsclaus:
At first glance this seems a bit wrong as future.get() is a blocking operation. And the method returns false which is signalling to camel that the response is async and completes later.

So for me it looks the old code, in that whenComplete you then need to deal with exception and set on the exchange before calling async.done(false)

rangoy:
Hi, thanks for the reply. 

As I said, when I asked at [zulipchat](https://camel.zulipchat.com/#narrow/stream/257298-camel/topic/Result.20from.20camel-milo.20endpoint/near/300812919)  , I was not sure how to do that. Do you have any suggestion how to do that?

> I did an attempt of patching this to be able to get the status from the server, not sure how to get the new body applied without using get()

davsclaus:
I would assume future get is available during whenComplete is called

rangoy:
> I would assume future get is available during whenComplete is called

I finally got to this again. For some reason I did not get that to work last time, but this seems to work as intended. I guess parsing of the status value can be left for the consumer. 

```
  future.whenComplete((o, throwable) -> {
                        final DefaultMessage newMessage = new DefaultMessage(exchange);
                        newMessage.setBody(o);
                        exchange.setMessage(newMessage);
                        async.done(false);
                    }
            );
            return false;
```

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientProducer.java
```diff
@@ -91,7 +92,13 @@ public boolean process(Exchange exchange, AsyncCallback async) {
         final Boolean await = msg.getHeader(MiloConstants.HEADER_AWAIT, this.defaultAwaitWrites, Boolean.class);
 
         if (TRUE.equals(await)) {
-            future.whenComplete((v, ex) -> async.done(false));
+            future.whenComplete((o, throwable) -> {
+                        final DefaultMessage newMessage = new DefaultMessage(exchange);
```
davsclaus, line 96: I dont think you need to set a new message as this leads to loosing headers and whatnot.

You usually just set the result directly on the message, eg

exchange.getMessage().setBody(...)

davsclaus (changes requested):

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientProducer.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.camel.Message;
 import org.apache.camel.component.milo.MiloConstants;
 import org.apache.camel.support.DefaultAsyncProducer;
+import org.apache.camel.support.DefaultMessage;
```
davsclaus, line 28: Remove this unusued import

davsclaus (approved):

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientProducer.java
```diff
@@ -91,7 +92,11 @@ public boolean process(Exchange exchange, AsyncCallback async) {
         final Boolean await = msg.getHeader(MiloConstants.HEADER_AWAIT, this.defaultAwaitWrites, Boolean.class);
 
         if (TRUE.equals(await)) {
-            future.whenComplete((v, ex) -> async.done(false));
+            future.whenComplete((result, throwable) -> {
+                        msg.setBody(result);
```
davsclaus, line 96: Ah what if throwable != null, eg I assume there was an error, then its likely better to set response, on the exchange

if (throwable != null) {
msg.getExchange().setException(throwable);
} else {
  msg.setBody(result);
}

rangoy: Ok, I amended the commit where I added that and also removed an unnecessary import also.

davsclaus (approved):

rangoy:
Rebased onto latest main

PR #8659, opened 2022-11-03
CAMEL-18677: allow setting the cache on the configuration builder

orpiske:


oscerd (approved):

orpiske:
Checkstyle issue is not caused by this change, therefore merging.

PR #8661, opened 2022-11-04
[CAMEL-18646] Provide custom configuration

gilvansfilho:
Allows user to provide a custom git config file through a endpoint query param.
Related to [Jira issue](https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-18646).


components/camel-git/src/main/docs/git-component.adoc
```diff
@@ -83,5 +83,17 @@ from("git:///tmp/testRepo?type=commit")
     .to(....)
 ---------------------------------------
 
+== Custom config file
+By default camel-git will load .gitconfig file from user home folder. You
```
orpiske, line 87: Use quotes to force the formatting to recognize this as a file. For instance, to make it like this `.gitconfig`.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        return getRepository(endpoint, DEFAULT_INSTANCE);
+    }
+
+    private static Repository fromPathToConfigFile(GitEndpoint endpoint, String pathToConfigFile) throws IOException {
+        if (ObjectHelper.isEmpty(pathToConfigFile)) {
+            throw new IllegalArgumentException("Path to git config file must be supplied");
+        }
+        File gitConfigFile;
+        if (pathToConfigFile.startsWith("/")) { //load from system
```
orpiske, line 54: Wouldn't this break on Windows?

oscerd (changes requested):
Maybe we can also point a remote configfile through resource helper and copy locally in the git repository before using it.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,76 @@
...
+        FileRepositoryBuilder builder = new FileRepositoryBuilder();
+        try {
+            SystemReader.setInstance(instance);
+            // scan environment GIT_* variables
+            return builder.setGitDir(new File(endpoint.getLocalPath(), ".git")).readEnvironment()
+                    .findGitDir() // scan up the file system tree
+                    .build();
+        } catch (IOException e) {
+            LOG.error("There was an error, cannot open {} repository", endpoint.getLocalPath());
+            throw e;
```
orpiske, line 72: Maybe describe the error that caused the problem. Something like: `LOG.error("There was an error opening the {} repository: {}", endpoint.getLocalPath(), e.getMessage);` 

Or ... just wrap the exception.

orpiske (changes requested):
Thanks for your contribution @gilvansfilho. I've left some comments.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
gilvansfilho: Sure, this will. I will change.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
gilvansfilho: `LOG.error("There was an error opening the {} repository: {}", endpoint.getLocalPath(), e.getMessage);` This will create a sonar issue.

So wrap on wich exception?

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
gilvansfilho: @oscerd 
> Maybe we can also point a remote configfile through resource helper and copy locally in the git repository before using it.

Do you think We do it using same config param (`gitConfigFile`) or other? like `remoteGitConfigFile`

oscerd: Probably with remoteGitConfigFile. With resource helper you could able to grab file content directly from classpath, filesystem, http etc.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
orpiske: I think `RuntimeCamelException` should do the trick.

components/camel-git/src/main/docs/git-component.adoc
gilvansfilho: Done.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
gilvansfilho: Done.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
gilvansfilho: Done.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,111 @@
...
+        String path = uri.substring(schema.length());
+
+        File gitConfigFile;
+        if (ResourceHelper.isClasspathUri(uri)) {
+            gitConfigFile = new File(endpoint.getClass().getClassLoader().getResource(path).getFile());
+        } else if (ResourceHelper.isHttpUri(uri)) {
+            try {
+                gitConfigFile = getTempFileFromHttp(uri);
+            } catch (IOException e) {
+                throw new RuntimeCamelException(String.format("Something goes wrong when loading: %s", uri), e);
```
orpiske, line 77: Just a minor grammatical nitpick: "Something went wrong when loading".

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+        } else { //load from system
+            gitConfigFile = new File(path);
+            if (!gitConfigFile.exists() || gitConfigFile.isDirectory()) {
```
orpiske, line 81: This fails to take into account access permissions for file. I'd recommend using [Files.isReadable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#isReadable(java.nio.file.Path)) to ensure that the file is a regular file that exists and is also readable.

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,111 @@
...
+                throw new IllegalArgumentException(String.format("Invalid file: %s", path));
```
orpiske, line 82: And based on the comment above, I'd reword the exception message to: `"The configuration file at %s is unreadable (either missing, lacking proper access permission os is not a regular file")`

components/camel-git/src/main/java/org/apache/camel/component/RepositoryFactory.java
```diff
@@ -0,0 +1,111 @@
...
+        FileRepositoryBuilder builder = new FileRepositoryBuilder();
+        try {
+            SystemReader.setInstance(instance);
+            // scan environment GIT_* variables
+            return builder.setGitDir(new File(endpoint.getLocalPath(), ".git")).readEnvironment()
+                    .findGitDir() // scan up the file system tree
+                    .build();
+        } catch (IOException e) {
+            throw new RuntimeCamelException(
+                    String.format("There was an error, cannot open %s repository", endpoint.getLocalPath()), e);
```
orpiske, line 99: Nitpicking a bit: "There was an error opening the repository at %s."

orpiske (changes requested):
Thanks for the changes. I think you are on the right track. 

I did leave some comments, but I think they are all simple to fix.

components/camel-git/src/main/java/org/apache/camel/component/git/GitEndpoint.java
```diff
@@ -77,6 +77,9 @@ public class GitEndpoint extends DefaultEndpoint {
               label = "producer")
     private String operation;
 
+    @UriParam(description = "A String with path to a .gitconfig file", label = "producer,consumer,advanced")
```
davsclaus, line 80: producer and consumer is default, so only use "advanced"

orpiske (approved):
Thanks for your contribution.

PR #8673, opened 2022-11-05
CAMEL-18549 backport of Dynamic Router EIP component to 3.14.x

Steve973:
Trivial pull request that backports the code from CAMEL-18549 to 3.14.x, adding the Dynamic Router EIP component code to this earlier version.


oscerd (approved):

davsclaus:
Ah okay so the EIP was added in 3.15. And because of that then this should not be backported, sorry for the noise.

Going to close this PR.

Steve973:
> Ah okay so the EIP was added in 3.15. And because of that then this should not be backported, sorry for the noise.
All right.  I only did it because you requested it in the ticket.  I was not sure how you and this project handle backports.

davsclaus:
Yeah really sorry for making you do all this effort - I should have looked more careful that this bug in the EIP was not part of that release.

We cannot make such a big change into a patch LTS branch and hence have to close this PR.

Steve973:
No worries!  It was not a big effort.  I only checked out 3.14.x, and pulled the changes from main into that module, and that's it.  Only took a minute.  A cherry-pick would have been even faster.  I have been using this component at work and made it (and, of course, Camel) a central part of all module intercommunication, so I may very well have some useful improvements to add to this.  I'll probably need your input for the best ways to do some of it.  Is there a preferred discussion space for this, other than a ticket?

davsclaus:
Yeah dev mailing list, zulip chat, or a JIRA. Usually a conversations starts with the former, and then when more concrete end up as a JIRA

PR #8678, opened 2022-11-07
CAMEL-17505: camel-core - Propose primitive types in builder methods

essobedo:
Related to https://issues.apache.org/jira/browse/CAMEL-17505

## Motivation

For primitive types like int, boolean and long, the builders propose only methods that accept a `String` as parameter, we need to overload those methods to propose more appropriate types in case of primitive types.

## Modifications:

* Add a method with a more convenient parameter type for each parameter of a primitive type
* Avoid setting unnecessary default values to keep the code easy to read and maintain
* Add the missing method `aliases` to `XStreamDataFormat`


oscerd (approved):

essobedo:
The checkstyle issue is not related to those changes:
```
Could not find artifact org.apache.camel:camel-snmp:jar:3.20.0-SNAPSHOT
```

PR #8687, opened 2022-11-08
CAMEL-18696: camel-ldap - Make filter a bit easier to use

davsclaus:


oscerd (approved):

davsclaus:
the test error is due to no tests executed as its a manual process to test against a ldap server.

noren95:
Can you confirm that the fix was merged into 3.14.x? I see nothing was changed on that branch. Thanks!

davsclaus:
Yeah it was a hickup it will be in 3.14.7 release that will be done later. You can use camel-spring-ldap instead.

PR #8706, opened 2022-11-09
CAMEL-18678: ensure strategies can access the Camel context

orpiske:


oscerd (approved):

davsclaus:
You may need != null check in stop/shudown if executorService is null. This can happen if you start and then something else fails, and then camel is shutting down existing services.

orpiske:
> You may need != null check in stop/shudown if executorService is null. This can happen if you start and then something else fails, and then camel is shutting down existing services.

Ohhh, good catch. I'll fix it tomorrow. Thanks for pointing it.

orpiske:
Checkstyle issue is not on this code, so merging.

PR #8717, opened 2022-11-15
Camel 18719 : camel-jbang - Adding the dependency of camel-quarkus-core creates two entries in exported POM.xml

mrinalsharma:
This pull request fixes [CAMEL-18719](https://issues.apache.org/jira/browse/CAMEL-18719)

- `dep = dep.substring(0, dep.lastIndexOf(","));` removed extra `,` that was causing an empty dependency in pom.xml
- quarkus-pom.tmpl already have camel-quarkus-core, camel-quarkus-platform-http, camel-quarkus-microprofile-health  included so there was no need to add it again if  dependency was passed as `--deps`


oscerd (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
```diff
@@ -342,6 +342,7 @@ private int run() throws Exception {
         if (dep == null) {
             dep = dependencies;
         } else if (dependencies != null && !dependencies.equals(dep)) {
+            dep = dep.substring(0, dep.lastIndexOf(","));
```
davsclaus, line 345: What if there is no comma in dep, then indexOf returns -1 and substring will fail. I think we need to check if dep ends with comma we can avoid adding the comma

davsclaus:
@mrinalsharma do you have time to fix that last comment?

mrinalsharma:
> 

Sorry, got stuck in a personal issue. Will try to finish it today.

davsclaus:
@mrinalsharma no rush, we need this within 2 weeks before we cut the 3.20 release

essobedo (changes requested):
Hi, thx for your contribution. Could you please rebase your branch or recreate it? Indeed your changes are drowned in the middle of many other changes that are not related which makes the PR very hard to review

davsclaus:
Ah yeah this PR seems wrong now, you can try to create a new clean PR with only your changes, and then close this

mrinalsharma:
created a new Pull request https://github.com/apache/camel/pull/8749

PR #8726, opened 2022-11-18
CAMEL-18724: fix display of default values in website

klease:
Corrects "false" not shown for EIP option defaults and all default values for options for languages and dataformats


oscerd (approved):

davsclaus (approved):

davsclaus:
I wonder if this can/need to be backported to 3.18.x branch and 3.14.x branch as well

PR #8727, opened 2022-11-18
(chores) camel-dataset: avoid checking data set index when creating the exchanges

orpiske:

### DIFF
--- a/components/camel-dataset/src/main/java/org/apache/camel/component/dataset/DataSetConsumer.java
+++ b/components/camel-dataset/src/main/java/org/apache/camel/component/dataset/DataSetConsumer.java
@@ -35,11 +35,13 @@ public class DataSetConsumer extends DefaultConsumer {
     private DataSetEndpoint endpoint;
     private Processor reporter;
     private ExecutorService executorService;
+    private final boolean withIndexHeader;
 
     public DataSetConsumer(DataSetEndpoint endpoint, Processor processor) {
         super(endpoint, processor);
         this.endpoint = endpoint;
         this.camelContext = endpoint.getCamelContext();
+        this.withIndexHeader = !endpoint.getDataSetIndex().equals("off");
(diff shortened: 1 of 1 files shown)


components/camel-dataset/src/main/java/org/apache/camel/component/dataset/DataSetConsumer.java
```diff
@@ -35,11 +35,13 @@ public class DataSetConsumer extends DefaultConsumer {
     private DataSetEndpoint endpoint;
     private Processor reporter;
     private ExecutorService executorService;
+    private boolean dataSetIndex;
```
essobedo, line 38: let's make it final

essobedo: Another remark, don't you believe that `dataSetIndexOff` or something similar could be more appropriate as name?

orpiske: +1. Those are good remarks.

essobedo: Or if we take directly the negation, it could rather be `withHeader`

davsclaus (approved):

components/camel-dataset/src/main/java/org/apache/camel/component/dataset/DataSetConsumer.java
orpiske: > Or if we take directly the negation, it could rather be `withHeader`

I went with a similar one: `withIndexHeader` so it's more specific. Thanks for pointing it out.

oscerd (approved):

essobedo (approved):

PR #8728, opened 2022-11-18
[CAMEL-18728]: Fix elasticsearch-rest component name to prevent component name clash

ludovic-boutros:
Fix elasticsearch-rest component name in order to prevent component name clash.

- [ ] Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.


essobedo (changes requested):
Thx for the contribution and for reporting. Regarding the PR, it is a lot of changes, I hope we can find a smaller fix. First of all, we need to avoid moving classes in `camel-elasticsearch-rest` as it is a very old component it is in maintenance mode so we are not supposed to change it a lot.

essobedo:
I rather prefer that we add an exception in the maven plugin directly as it has already been done [here](https://github.com/apache/camel/blob/main/tooling/maven/camel-package-maven-plugin/src/main/java/org/apache/camel/maven/packaging/EndpointDslMojo.java#L1000-L1005)

Simply adds:
```
            case "org.apache.camel.component.elasticsearch.ElasticsearchComponent":
                return "ElasticsearchRest";
```

oscerd (changes requested):
I think this is a bit overkilling, I would follow the way suggested by @essobedo: the main reason is just because we are planning to remove the rest component at some point.

ludovic-boutros:
I will follow the way you all suggested.
I did not know it. It's far more simple !

ludovic-boutros:
It should be ok now. So easy when you understand the build process a bit more 👍 .
Thank you.

essobedo (changes requested):
Much better thx 👍. 

One last remark, could you please revert the changes in `camel-spring.xsd`, they are not related and are due to some kind of reformatting that will be reverted anyway by the next "Regen commit". Let's keep it as small as possible. Thx in advance 🙏

essobedo:
Could you please also remove "in order" from your commit message?

ludovic-boutros:
@essobedo it should be good now.

essobedo (approved):
That's it, great job, many thx 👍

PR #8735, opened 2022-11-20
CAMEL-18707: Upgrade to Jandex 3.0.3

klease:
Change maven group to io.smallrye.
Adapt to minor API changes in Jandex 3.


klease:
Jandex is also declared as a dependency in the spring-boot project by camel-spring-boot-generator-maven-plugin but it isn't used.
Camel-quarkus uses the jandex version from quarkus (currently 3.0.1) in extensions, but it uses version 1.0.8 of the jandex-maven-plugin in some integration-test pom files. Should that be updated to use 3.0.3 of the jandex plugin from io.smallrye?

davsclaus:
camel-quarkus team will take care of jandex upgrade in that project
If jandex is not used in spring boot, then you are welcome to remove it there.

PR #8746, opened 2022-11-21
CAMEL-18717: do save to offset repository when using a commit manager

orpiske:


orpiske:
I asked the reporter to try this patch. So, please wait a bit before merging. 

Also, this probably needs to be back-ported to 3.18.x.

oscerd (approved):

davsclaus (approved):

orpiske:
So, the user will test this once released. Let's merge and back-port this.

PR #8749, opened 2022-11-21
Fixed Camel 18719 : camel-jbang - Adding the dependency of camel-quarkus-core creates two entries in exported POM.xml

mrinalsharma:
This pull request fixes [CAMEL-18719](https://issues.apache.org/jira/browse/CAMEL-18719)

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
@@ -344,7 +344,7 @@ class Run extends CamelCommand {
         // merge existing dependencies with --deps
         String deps = RuntimeUtil.getDependencies(profileProperties);
         if (deps.isBlank()) {
-            deps = dependencies;
+            deps = dependencies != null ? dependencies : "";
         } else if (dependencies != null && !dependencies.equals(deps)) {
             deps += "," + dependencies;
         }
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
@@ -342,15 +342,15 @@ class Run extends CamelCommand {
         writeSetting(main, profileProperties, "camel.jbang.console", console ? "true" : "false");
         writeSetting(main, profileProperties, "camel.main.routesCompileDirectory", WORK_DIR);
         // merge existing dependencies with --deps
-        String dep = profileProperties != null ? profileProperties.getProperty("camel.jbang.dependencies") : null;
-        if (dep == null) {
-            dep = dependencies;
-        } else if (dependencies != null && !dependencies.equals(dep)) {
-            dep += "," + dependencies;
+        String deps = RuntimeUtil.getDependencies(profileProperties);
+        if (deps.isBlank()) {
+            deps = dependencies;
+        } else if (dependencies != null && !dependencies.equals(deps)) {
+            deps += "," + dependencies;
         }
-        if (dep != null) {
-            main.addInitialProperty("camel.jbang.dependencies", dep);
-            writeSettings("camel.jbang.dependencies", dep);
+        if (deps != null) {
+            main.addInitialProperty("camel.jbang.dependencies", deps);
+            writeSettings("camel.jbang.dependencies", deps);
         }
 
         // command line arguments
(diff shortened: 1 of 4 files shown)


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/RuntimeUtil.java
```diff
@@ -83,4 +83,20 @@ public static void loadProperties(Properties properties, File file) throws IOExc
         }
     }
 
+    public static String getDependencies(Properties properties) {
+        String deps = properties.getProperty("camel.jbang.dependencies");
```
essobedo, line 87: Warning `properties` could be null according to https://github.com/apache/camel/pull/8749/files#diff-b62a4b0acb50014174399c284b5540cf84016571e5c789a23edc8d843692721cL345

essobedo (changes requested):
Thx for recreating your PR, it is much easier to read. Just one small remark

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/RuntimeUtil.java
```diff
@@ -83,4 +83,20 @@ public static void loadProperties(Properties properties, File file) throws IOExc
...
+        if (deps != null) {
+            deps = deps.trim();
+            if (deps.length() > 0 && deps.charAt(0) == ',') {
+                deps = deps.substring(1);
+            }
+            if (deps.length() > 0 && deps.charAt(deps.length() - 1) == ',') {
+                deps = deps.substring(0, deps.lastIndexOf(","));
+            }
+        } else {
+            deps = new String("");
```
essobedo, line 97:
```suggestion
            deps = "";
```

mrinalsharma:
In the method getDependencies(), we are trying to read `property camel.jbang.dependencies`. If its null then the method should return empty string.

Run.java:
if getDependencies() doesn't return blank then merge the dependencies read from file and the dependencies passes as --deps parameter.

ExportQuarkus.java
`camel-quarkus-core,camel-quarkus-platform-http,camel-quarkus-microprofile-health` are  already defined in `quarkus-pom.tmpl'` so there is no need to define them again. Adding again in pom.xml cause a compilation error.

ExportBaseCommand.jaja
The change avoids adding a blank dependency in pom.xml. Blank dependency causes a compilation error.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
```diff
@@ -342,15 +342,15 @@ private int run() throws Exception {
...
-        if (dep != null) {
-            main.addInitialProperty("camel.jbang.dependencies", dep);
-            writeSettings("camel.jbang.dependencies", dep);
+        String deps = RuntimeUtil.getDependencies(profileProperties);
+        if (deps.isBlank()) {
+            deps = dependencies;
+        } else if (dependencies != null && !dependencies.equals(deps)) {
+            deps += "," + dependencies;
+        }
+        if (!deps.isBlank()) {
```
essobedo, line 351: Warning `deps` could be `null` in case `deps.isBlank()` returns initially `true` and `dependencies` is `null`. In this case, we would end up with an NPE

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/RuntimeUtil.java
essobedo: To be clear, if `properties` is `null`, calling `properties.getProperty("camel.jbang.dependencies")` will throw a `NullPointerException`

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
mrinalsharma: Thanks for the clarification.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/RuntimeUtil.java
mrinalsharma: My understanding is that properties are expected to be not null, if that's not the case we can see NullPointerException at so many other places.

essobedo: If you check these lines https://github.com/apache/camel/blob/59c6450582fa662c168021360c42f1e78a4cc438/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java#L266-L269, you will see that `profileProperties` is `null` if the file doesn't exist which can occur so I confirm that `properties` can be `null`

mrinalsharma: Sure, I should have the fix soon.

davsclaus (approved):

oscerd (approved):

essobedo (approved):

PR #8753, opened 2022-11-21
CAMEL-18736: Ensure cxfEndpoint schemas are uniquely named for JaxRS and JaxWS to avoid classpath conflicts

driseley:
Prevent classpath conflicts occurring when both camel-cxf-spring-soap and camel-cxf-spring-rest modules are included in a spring boot application, and the spring schemas are loaded.

I wasn't able to think of a sensible test for this change, but all existing tests pass and a local build of this resolves our issue


oscerd (approved):

davsclaus (approved):

davsclaus:
The public namespaces are not affected so Camel end users should not have anything to change in their applications.

PR #8759, opened 2022-11-22
allow ES modules in JS DSL

hguerrero:
- [x ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
- [x ] Each commit in the pull request should have a meaningful subject line and body.
- [x ] If you're unsure, you can format the pull request title like `[CAMEL-XXX] Fixes bug in camel-file component`, where you replace `CAMEL-XXX` with the appropriate JIRA issue.
- [x ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x ] Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.

This PR allows using ES modules with the JavaScript DSL by adding the IO permission to the GraalVM context. Also, we are using an ES module name for processing. This will allow importing `node_modules` from the npm repository and use them in the route through process functions.

Example of the new route:

```js
// camel-k: language=js
import { faker } from "./node_modules/@faker-js/faker/dist/esm/index.mjs";

const proc = (e) => {
  e.getIn().setBody(faker.name.fullName());
};

// Write your routes here, for example:
from("timer:js?period=30000").routeId("js").process(proc).log("${body}");
```

### DIFF
--- a/dsl/camel-js-dsl/src/main/java/org/apache/camel/dsl/js/JavaScriptRoutesBuilderLoader.java
+++ b/dsl/camel-js-dsl/src/main/java/org/apache/camel/dsl/js/JavaScriptRoutesBuilderLoader.java
@@ -48,6 +48,7 @@ public class JavaScriptRoutesBuilderLoader extends EndpointRouteBuilderLoaderSup
                 .allowExperimentalOptions(true)
                 .allowHostClassLookup(s -> true)
                 .allowPolyglotAccess(PolyglotAccess.NONE)
+                .allowIO(true)
                 .option("engine.WarnInterpreterOnly", "false");
 
         final Context context = contextBuilder.build();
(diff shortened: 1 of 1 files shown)


rhuan080:
My suggestion is to provide a unit test. Could you provide a unit test?

hguerrero:
Let me check if I can find an example package we can use for a test, as I don't think we can reuse faker.

davsclaus (approved):

oscerd (approved):

davsclaus:
We should make camel-js-del depend on camel-javascript and reuse its way to create that JS Context instance so its the same way.

davsclaus:
@hguerrero I had to remove ".mjs" from the name as otherwise the script fails to compile

PR #8774, opened 2022-11-24
camel-splunk-hec - Add endpoint url suffix configuration to enable op…

kremers:
…tions and raw transports

Trivial change, adding option to configure endpoint url suffix


oscerd (approved):

kremers:
Added test and default value in [60c65f8]

kremers:
Fixed checkstyle error from automated testing in [b999da5]

components/camel-splunk-hec/src/main/java/org/apache/camel/component/splunkhec/SplunkHECConfiguration.java
```diff
@@ -99,6 +101,20 @@ public void setHost(String host) {
         this.host = host;
     }
 
+    /**
+     * Splunk endpoint
+     *   Defaults to /services/collector/event
+     *   To write RAW data like JSON use /services/collector/raw
+     *   To extract timestamps in Splunk>8.0 /services/collector/event?auto_extract_timestamp=true
+     */
+    public void setEndpoint(String endpoint) {
```
davsclaus, line 110: Are there other endpoints? And endpoint can maybe confuse as Camel also has a concept of endpoint. So maybe we can come up with a different name?

For the last with the query parameter then that will not work as Camel think its another parameter.

So maybe come up

1) a new name, with 3 enum values: default, raw, timestamp
2) And then you can set the url based on the endpoint inside the camel component

davsclaus (changes requested):

components/camel-splunk-hec/src/main/java/org/apache/camel/component/splunkhec/SplunkHECConfiguration.java
kremers: Hi Claus, thank you very much for your quick reply and check of the PR!

- Thank you very mich for the hint with the name, okay to set it to "SplunkRESTEndpoint"? Which would be a better naming. 

- If the url includes a question mark the user can utilize RAW{} https://camel.apache.org/manual/faq/how-do-i-configure-endpoints.html#HowdoIconfigureendpoints-Configuringparametervaluesusingrawvalues so it will work. Modules like azure-eventhub do make excessive use of it for connection strings, passwords, ... for example:

`.from(azure-eventhubs:?connectionString=RAW({{camel.component.azure-eventhubs.connection-string}}) ...`

- For the reason of extendability, enum does not sound like a good fit to me, since it requires changes to the software and different versions in place on every new Splunk version/endpoint. The ENUM may also suggest RESTEndpoints that are not available in older Splunk versions. So I suggest to stick with a strong default that does not introduce a breaking change, but full flexibility to adpot to new Splunk versions without changing the camel module. See: https://docs.splunk.com/Documentation/SplunkCloud/8.2.2203/Data/HECRESTendpoints

- Naming in the module can be improved a lot, since hostAndIp are written aus SplunkURL where users normally expect an URL compliant to RFC1738 (https://datatracker.ietf.org/doc/html/rfc1738) including the REST endpoint including http or https (including RAW() for handling slashes). One possible way to improve and keep backwards compability would be introduction of a configuration parameter "connectionString" that has first priority if set and includes all information. But I guess that is better adressed in a second pull request.

davsclaus: Thanks for the detailed feedback. Yeah components can surely be improved, you are welcome to send more PRs.

components/camel-splunk-hec/src/main/java/org/apache/camel/component/splunkhec/SplunkHECConfiguration.java
```diff
@@ -36,6 +36,8 @@ public class SplunkHECConfiguration {
     private String source = "camel";
     @UriParam
     private String host;
+    @UriParam(defaultValue = "/services/collector/event")
+    private String splunkRESTEndpoint = "/services/collector/event";
```
davsclaus, line 40: Cane we use `splunkEndpoint` instead as upper-case REST make it a bit more unusual and a bit harder to read.

davsclaus (changes requested):

components/camel-splunk-hec/src/main/java/org/apache/camel/component/splunkhec/SplunkHECConfiguration.java
kremers: Yes, I'll change it.

davsclaus (approved):

components/camel-splunk-hec/src/main/java/org/apache/camel/component/splunkhec/SplunkHECConfiguration.java
davsclaus: Thanks that was quick

oscerd (approved):

PR #8777, opened 2022-11-25
CAMEL-18761 camel-open-api - Warning log when using boolean type

Kinae:
Trivial change

The generation of the Schema produce a warn log when it should not for the boolean type : 
`WARN  RestModelConverters            - Encountered unexpected type boolean in processing schema.`

The OpenApi spec says : https://swagger.io/docs/specification/data-models/data-types

```
string (this includes dates and files)
number
integer
boolean
array
object
```

Since it just an error log, it's hard to test, so I added a boolean in the User.java, the log is no longer there.


oscerd (approved):

davsclaus (approved):

Kinae:
Thank you !

PR #8778, opened 2022-11-25
CAMEL-18731: Add result type and different sources of input data to languages

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-18731

## Motivation:

We need to have `resultTypeName`, `resultType`, `headerName`, and/or `propertyName` for all languages for which it can make sense.

## Modifications:

* Add `resultTypeName` and `resultType` to bean, groovy, hl7terser, mvel, ognl, ref, spel, ref and XQuery
* Add `headerName` and/or `propertyName` to xquery, hl7terser, jq, jsonpath, tokenize, xpath, XQuery, xtokenize
* Remove `headerName` from js and python since they already provide the headers as part of the input data
* Add several classes for type "Typed..." as the base class for all the languages that support the result type
* Add several classes for type "SingleInputTyped..." as the base class for all the languages that support the result type and different sources of input data
* Add several classes for type "SingleInput..." as the base class for all the languages that support different sources of input data but not the result type
* Add several abstract test classes to extend to validate the behavior of each language affected by these changes
* Add missing attributes to the annotations `@Joor`, `@JsonPath`, `@XQuery`, `@XPath`
* Set the result type when applicable in the methods `xpath` and `xquery` of `ExpressionClauseSupport`


oscerd (approved):

davsclaus:
@essobedo you cannot have camel-core dependency in camel-bean when testing due to build order problems. So ideally have tests in core/camel-core, and only some basic tests in camel-bean

essobedo:
> @essobedo you cannot have camel-core dependency in camel-bean when testing due to build order problems. So ideally have tests in core/camel-core, and only some basic tests in camel-bean

@davsclaus Hi Claus, thx for the remark, however, I don't get it, I have added camel-core in camel-bean and the test for the bean language is already in core, so I'm confused. Could you please clarify a bit more?

davsclaus:
camel-bean was in v2 in core, and tests for bean are 99% in camel-core and should stay there.

Also you can end up with maven not able to build the project from 100% clean due to camel-bean is needed in core, and core is not build yet - only because your local m2 repo have camel-core test-jar already it can use that

essobedo:
I don't add any dependency to `camel-bean`, from `camel-bean`, the only new class that refers to is `TypedLanguageSupport` which is part of `camel-support` not `camel-core`.
Regarding the test for the `bean` language, it is not in `camel-bean` but in `camel-core`.

In both cases, I'm sorry but it is still not clear to me what's wrong. Could you please share with me the line(s) in the code where the problem is?

davsclaus (approved):

davsclaus:
Thanks for this great work to cleanup and align the languages.

PR #8782, opened 2022-11-27
CAMEL-18764: camel-core - templatedRoute builder from Java cannot use…

davsclaus:
… templates from other DSLs


oscerd (approved):

davsclaus:
The problem is that `RouteBuilder` now with templatedRoute can trigger too early. So we need to move this out of its normal route add logic, and make templated routes last.

It would be good to add unit test for this in core.

If you look at how `ConfigureRouteTemplates` was in use previously then its invoked last after all the regular route add stuff, so this PR does like what it does.

https://github.com/apache/camel/blob/0fbb0d62f8eb68aeca0cfda898f0b82804402d78/core/camel-main/src/main/java/org/apache/camel/main/RoutesConfigurer.java#L212

essobedo (approved):

PR #8783, opened 2022-11-27
CAMEL-18757 Specify type for pom-only artifacts like camel-*-parent

cunningt:
https://issues.apache.org/jira/browse/CAMEL-18757

Check the model of each component for the packaging type.    If a pom, express that in the parent and the bom.

Did some parent/pom.xml cleanup as well - I checked through the entries for core, dsl, and catalog and added dependency entries for any missing artifacts under those sections.     Fixed some spacing issues in parent/pom.xml generation.


davsclaus:
I wonder if maven plugins such as
```
 <dependency>
        <artifactId>camel-vertx-kafka-maven-plugin</artifactId>
      </dependency>
```

Should have

```
<type>maven-plugin</type>
```
https://maven.apache.org/ref/3.8.6/maven-core/artifact-handlers.html

cunningt:
Can definitely make that change but in the other projects that I've checked (cxf, karaf) the type isn't included for maven-plugin.   ` <type>pom</type>` is an issue because no jars exist for that dependency type and it causes errors on certain bom resolution programs that default to looking for a jar.    In the case of maven-plugins, a jar definitely exists.

davsclaus:
Okay its fine if others dont do that - also I dont think its a real problem, for end users that import this BOM

davsclaus (approved):

PR #8785, opened 2022-11-28
CAMEL-18766 Fixes bug in camel-support

thuri:
- [X] Run `mvn clean install -Psourcecheck` in your module with source check enabled to make sure basic checks pass and there are no checkstyle violations. A more thorough check will be performed on your pull request automatically.

The changes in this PR fix a bug in the BackgroundTask class which results in a thread that loops forever doing nothing but log a warning. 

The fix is essentially ending the thread when the budget as been consumed. What that means is dependent on the actual implementation of the Budget instance that is passed to the BackgroundThread. It may mean that the `maxDuration` has been exceeded or that the task has been run for `maxIteration` times.

The new implementation saves the result of the runTaskWrapper method in an wrapper instance so that it can be used as return value of the run method implementation.

This is necessary because the Future, returned by the scheduleAtFixedRate method can't be used for that as the invocation of its `get` method would result in an exception.

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
@@ -106,11 +106,6 @@ public class BackgroundTask implements BlockingTask {
     }
 
     @Override
-    public <T> boolean run(Predicate<T> predicate, T payload) {
-        return this.run(() -> predicate.test(payload));
-    }
-
-    @Override
     public boolean run(BooleanSupplier supplier) {
         final CountDownLatch latch = new CountDownLatch(1);
         // we need a wrapper for the actual result that will be defined in the runTaskWrapper method which 
--- a/core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
@@ -24,7 +24,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
-import java.util.function.Predicate;
 
 import org.apache.camel.support.task.budget.TimeBoundedBudget;
 import org.apache.camel.support.task.budget.TimeBudget;
@@ -77,7 +76,10 @@ public class BackgroundTask implements BlockingTask {
     private final TimeBudget budget;
     private final ScheduledExecutorService service;
     private final String name;
+    private final CountDownLatch latch = new CountDownLatch(1);
+
     private Duration elapsed = Duration.ZERO;
+    private boolean completed;
 
     BackgroundTask(TimeBudget budget, ScheduledExecutorService service, String name) {
         this.budget = budget;
@@ -85,7 +87,7 @@ public class BackgroundTask implements BlockingTask {
         this.name = name;
     }
 
-    private void runTaskWrapper(CountDownLatch latch, BooleanSupplier supplier, ExecutionResult result) {
+    private void runTaskWrapper(BooleanSupplier supplier) {
         LOG.trace("Current latch value: {}", latch.getCount());
         if (latch.getCount() == 0) {
             return;
@@ -93,13 +95,13 @@ public class BackgroundTask implements BlockingTask {
 
         if (!budget.next()) {
             LOG.warn("The task {} does not have more budget to continue running", name);
-            result.completed = false;
+            completed = false;
             latch.countDown();
             return;
         }
 
         if (supplier.getAsBoolean()) {
-            result.completed = true;
+            completed = true;
             latch.countDown();
             LOG.trace("Task {} succeeded and the current task won't be schedulable anymore: {}", name, latch.getCount());
         }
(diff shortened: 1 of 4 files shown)


orpiske:
GH actions won't test all code affected by this, so I will test it manually.

oscerd (approved):

core/camel-support/src/test/java/org/apache/camel/support/task/BackgroundIterationTimeTaskTest.java
```diff
@@ -42,7 +42,8 @@ void testRunNoMoreSupplier() {
                         .withMaxIterations(3)
                         .withInterval(Duration.ofSeconds(1))
                         .withInitialDelay(Duration.ZERO)
-                        .withMaxDuration(Duration.ofSeconds(5))
+                        // use unlimitedduration so we're sure that the task is really canceled after maxIterations
```
essobedo, line 45:
```suggestion
                        // use unlimited duration so we're sure that the task is really canceled after maxIterations
```

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
```diff
@@ -85,78 +85,58 @@ public BackgroundTask build() {
...
         }
 
         if (!budget.next()) {
             LOG.warn("The task {} does not have more budget to continue running", name);
-
-            return;
-        }
-
-        if (predicate.test(payload)) {
+            result.completed = false;
```
essobedo, line 96: Is it really needed since `false` is the default value for a `boolean`?

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
```diff
@@ -85,78 +85,58 @@ public BackgroundTask build() {
...
 
     @Override
     public <T> boolean run(Predicate<T> predicate, T payload) {
-        CountDownLatch latch = new CountDownLatch(1);
-
-        Future<?> task = service.scheduleAtFixedRate(() -> runTaskWrapper(latch, predicate, payload),
-                budget.initialDelay(), budget.interval(), TimeUnit.MILLISECONDS);
-
-        return waitForTaskCompletion(latch, task);
+        return this.run(() -> predicate.test(payload));
```
essobedo, line 110: I'm wondering if it could be moved into the interface `BlockingTask` as the default implementation?

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
```diff
@@ -167,12 +147,14 @@ private boolean waitForTaskCompletion(CountDownLatch latch, Future<?> task) {
...
-        return completed;
     }
 
     @Override
     public Duration elapsed() {
         return elapsed;
     }
+
+    private static class ExecutionResult {
+        private boolean completed;
```
essobedo, line 158: @orpiske This makes me wonder if `BackgroundTask` is stateful or stateless. I mean can run several times the same task? because if not, I guess that the `CountDownLatch` and this field could be moved directly as fields of `BackgroundTask` otherwise maybe we could add the `CountDownLatch` to that class and rename it to something like `ExecutionState`

essobedo (commented):
Thx for the PR, the fix makes sense 👍 Just small remarks

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
essobedo: @orpiske any opinion on this?

thuri: That's a good idea. Should I add this to the PR?

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
orpiske: Yes, it's stateful because of `budget` and `elapsed`.

I think your suggestion is +/- in line with what I commented on the Zulip chat: to investigate moving the latch into the budget and consider it as part of that. Granted, I did not comment on the elapsed, but I think it would eventually become a natural move as well. 

IMHO, making (all) the tasks stateless - keeping the state on the budget itself - could potentially be a better approach for the future (maybe as part of Camel 4?), but I think the current approach is fine.

core/camel-support/src/main/java/org/apache/camel/support/task/BackgroundTask.java
orpiske: +1. That would be better / cleaner.

orpiske (approved):
I tested it manually on my own CI and the proposed PR seems fine, not causing problems on other components using this.

Just please apply the other suggestions and it should be good to be merged.

thuri:
I have squashed the changes that actual fix the issue and the small commit that has been made by github.
I'd leave the other changes as separate commits because they're not really related to the issue.
One commit is the default implementation of the run(Predicate) method.
The other commit is moving the latch and the completed flag to the object level.

@orpiske If possible it would be good to repeat the manual test especially because the completed flag and latch are now member variables.

orpiske:
> 

Thanks for the heads up. I'll retest tonight and I'll comment tomorrow morning about how it is going.

thuri:
Not sure why the build fails here. I need to have a deeper look into this. The checksource maven profile didn't show me any error and i can't find the build log for the failing test in camel-cdi ... May take a while. 
Sorry :-(

essobedo:
Click on Details then Summary, you will then see the link to download the different log files at the bottom of the page.
In your case the checkstyle error is:
``` 
[camel-support] [INFO] Starting audit...
[ERROR] /home/runner/work/camel/camel/core/camel-support/src/main/java/org/apache/camel/support/task/BlockingTask.java:40:5: File contains tab characters (this is the first instance). [FileTabCharacter]
```

Always tries to launch the command `mvn clean install -Psourcecheck` against the modules in which you have updated files

orpiske:
> Click on Details then Summary, you will then see the link to download the different log files at the bottom of the page. In your case the checkstyle error is:
> 
> ```
> [camel-support] [INFO] Starting audit...
> [ERROR] /home/runner/work/camel/camel/core/camel-support/src/main/java/org/apache/camel/support/task/BlockingTask.java:40:5: File contains tab characters (this is the first instance). [FileTabCharacter]
> ```
> 
> Always tries to launch the command `mvn clean install -Psourcecheck` against the modules in which you have updated files

Actually, only us committters have access to those files (due to Apache rules). So, in cases like these, we have to upload them in the comments.

orpiske:
> > 
> 
> Thanks for the heads up. I'll retest tonight and I'll comment tomorrow morning about how it is going.

The tests passed on my own CI. We can merge it (and backport) once you fix the formatting problem.

essobedo:
> Actually, only us committters have access to those files (due to Apache rules). So, in cases like these, we have to upload them in the comments.

I was not aware of it, thx for the info. This can be annoying in the long term

essobedo (approved):
LGTM 👍

orpiske:
> ### Components tested:
> 13 	13 	1 	13

Failed caused by a well known flakiness on the core-management tests. Therefore, merging.

orpiske:
BTW, I am going to backport this. I already have a build in progress and will open the PR soon. 

@thuri thank you for contributing with this fix!

PR #8789, opened 2022-11-29
Fix the camel project export dir option typo

WillemJiang:
`camel export` command directory parameter should be `-dir` instead of `--dir`


zhfeng (approved):

tadayosi (approved):

oscerd (approved):

davsclaus:
Ah lets use `--directory` so all args are using double `--`

davsclaus (approved):

PR #8811, opened 2022-11-30
[CAMEL-18737]: camel-kamelet parameter substitution does not work in …

luigidemasi:
…bean instantiation when constructor is used


davsclaus:
If you can provide an unit test with this in camel-kamelet, that would be good IMHO

luigidemasi:
> If you can provide an unit test with this in camel-kamelet, that would be good IMHO

done

davsclaus (approved):

davsclaus:
This is great work, can you add the unit test to the 3.18.x PR

PR #8812, opened 2022-11-30
[CAMEL-18737]: camel-kamelet parameter substitution does not work in …

luigidemasi:
…bean instantiation when constructor is used


davsclaus:
Ah there is already tests for 3.18.x, thanks

PR #8828, opened 2022-12-03
CAMEL-18792: camel-hdfs - Make integration tests runnable without a real cluster

sekikn:

### DIFF
--- a/components/camel-hdfs/pom.xml
+++ b/components/camel-hdfs/pom.xml
@@ -145,6 +145,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-minicluster</artifactId>
+            <version>${hadoop3-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.lz4</groupId>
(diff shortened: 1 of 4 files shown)


oscerd (approved):

davsclaus:
Thanks for helping with this. I wonder if the test used a hardcoded port number (9000 ?) then can we try using AvailablePortFinder which can find a free port number, as CI server may have "port number in use" problem.

sekikn:
Thanks for the comment! I'll fix the PR to leverage AvailablePortFinder.
In addition, I've just noticed that there's already camel-test-infra-hdfs in test-infra, so it looks better to fix it to use minicluster (it requires an external cluster now) just like hbase, rather than include minicluster directly into the integration tests. I'll update the PR.

orpiske:
> Thanks for the comment! I'll fix the PR to leverage AvailablePortFinder. In addition, I've just noticed that there's already camel-test-infra-hdfs in test-infra, so it looks better to fix it to use minicluster (it requires an external cluster now) just like hbase, rather than include minicluster directly into the integration tests. I'll update the PR.

Thanks for your contribution!

+1 for adding it to the `test-infra` component. This could - potentially - also benefit the Camel Kafka Connector project which uses it.

davsclaus (approved):

PR #8830, opened 2022-12-04
Upgrade opentelemetry to the version 1.20.1

bvahdat:
- Will separately look into `camel-opentelemetry-starter` and provide PR by it's related repo if required.

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -442,7 +442,7 @@
     <openjpa-version>3.2.2</openjpa-version>
     <openstack4j-version>3.10</openstack4j-version>
     <opentelemetry-alpha-version>${opentelemetry-version}-alpha</opentelemetry-alpha-version>
-    <opentelemetry-version>1.17.0</opentelemetry-version>
+    <opentelemetry-version>1.20.1</opentelemetry-version>
     <opentracing-tracerresolver-version>0.1.8</opentracing-tracerresolver-version>
     <opentracing-version>0.33.0</opentracing-version>
     <optaplanner-version>8.31.0.Final</optaplanner-version>
--- a/components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/SetCorrelationContextProcessor.java
+++ b/components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/SetCorrelationContextProcessor.java
@@ -98,7 +98,7 @@ public class SetCorrelationContextProcessor extends AsyncProcessorSupport implem
     }
 
     public String getBaggageName() {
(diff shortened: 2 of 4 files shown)


davsclaus:
Why do you remove the public in all that source code ?

bvahdat:
I have only removed the `public` modifiers in Tests, please see Sam Brannen's answer here:

https://stackoverflow.com/a/55230350

Who is the main contributor to JUnit 5:

https://github.com/junit-team/junit5/graphs/contributors

And I think @essobedo also applies the same pattern by all his PRs which makes sense to me as well.

Also please see the INFO box here:

https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

oscerd (approved):

davsclaus:
Please keep as before, this PR is about dependency upgrade.

We should not change `public` in test methods which causes backport problems as the code is changed compared to older branches.

bvahdat:
> Please keep as before, this PR is about dependency upgrade.
> 
> We should not change `public` in test methods which causes backport problems as the code is changed compared to older branches.

Sure, is done.

davsclaus (approved):

oscerd (approved):

davsclaus:
I am not sure why the CI server failed the 2nd time

bvahdat:
> I am not sure why the CI server failed the 2nd time

These are test failures in core not relevant to this PR. Though not sure why those tests are run in the first place, as this PR has no effect on e.g. `camel-core`, but maybe @orpiske knows more about it.

Screen Shot 2022-12-04 at 18 56 49

Screen Shot 2022-12-04 at 18 57 06

Anyway going to rebase & merge.

orpiske:
> > I am not sure why the CI server failed the 2nd time
> 
> These are test failures in core not relevant to this PR. Though not sure why those tests are run in the first place, as this PR has no effect on e.g. `camel-core`, but maybe @orpiske knows more about it.


One of the core tests is flaky on Github actions. Let's see if #8835 improves a bit.

orpiske:
> We should not change public in test methods which causes backport problems as the code is changed compared to older branches.

IMHO, that's a cleanup we need to do separately and in one shot per component (so we can cherry-pick) and only removing the `public` modifier. Maybe for Camel 4 ...

bvahdat:
> > We should not change public in test methods which causes backport problems as the code is changed compared to older branches.
> 
> IMHO, that's a cleanup we need to do separately and in one shot per component (so we can cherry-pick) and only removing the `public` modifier. Maybe for Camel 4 ...

Thanks @orpiske for your feedback. This has been now taken care of separately through https://github.com/apache/camel/pull/8831

PR #8837, opened 2022-12-05
CAMEL-18743 - camel-jbang - Command to see only if everything is okay…

oscerd:
… or there has been a failure

Signed-off-by: Andrea Cosentino <ancosen@gmail.com>


davsclaus (approved):

davsclaus:
We may want some kind of option that can make parsing from shell easier, in case you just want a single count without table header etc

camel count <pid> --total
2

camel count <pid> --fail
0

camel count <pid> --total --fail
2,0

oscerd:
Yeah, maybe if you don't specificy options go with the actual output otherwise no headers and just output

oscerd:
@davsclaus should be as we wish

PR #8849, opened 2022-12-06
camel-kafka: added an auth-based IT test

orpiske:


orpiske:
Just for the record: there are cleanups to do here, but I'll do them in a separate PR (this test is a pre-work for a Camel Kafka fix / bug verification)

PR #8854, opened 2022-12-07
(chores) replacing more inner classes with lambda

gilvansfilho:
Solving some sonar issues about [replacing inner classes with lambdas](https://rules.sonarsource.com/java/tag/java8/RSPEC-1604).


davsclaus (approved):

davsclaus:
Okay since we started on this - however it makes backporting to older branches harder as then the code is changed. Ideally it would have been a 4.0 change.

gilvansfilho:
> Okay since we started on this - however it makes backporting to older branches harder as then the code is changed. Ideally it would have been a 4.0 change.

@davsclaus  There are a lot of this to be done however I'll wait 4.0 as you said.

PR #8870, opened 2022-12-08
CAMEL-18796: fix for the issue + test case / reproducer

orpiske:


orpiske:
If merging, please do not squash. I need to cherry-pick a few of the commits to 3.18.x.

PR #8879, opened 2022-12-11
(chores) documentation: add a section about creating the project

orpiske:

### DIFF
--- a/docs/main/modules/getting-started/pages/index.adoc
+++ b/docs/main/modules/getting-started/pages/index.adoc
@@ -5,7 +5,7 @@ This document will guide you through the fundamentals of the Apache Camel Core. 
 [[BookGettingStarted-eip-book]]
 
 [[BookGettingStarted-TheEnterpriseIntegrationPatternsEIPBook]]
-== The _Enterprise Integration Patterns_ (EIP) book
+== The _Enterprise Integration Patterns_ (E.I.P.) book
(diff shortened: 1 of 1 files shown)


oscerd (approved):

docs/main/modules/getting-started/pages/index.adoc
```diff
@@ -5,7 +5,7 @@ This document will guide you through the fundamentals of the Apache Camel Core.
 [[BookGettingStarted-eip-book]]
 
 [[BookGettingStarted-TheEnterpriseIntegrationPatternsEIPBook]]
-== The _Enterprise Integration Patterns_ (EIP) book
+== The _Enterprise Integration Patterns_ (E.I.P.) book
```
davsclaus, line 8: We use EIP as the reference name, and not with dots

davsclaus (approved):

docs/main/modules/getting-started/pages/index.adoc
orpiske: So, Grammarly complains about using it this way. I wonder if we should reduce usage of the term, as this  indicates it affects readability.

Any thoughts about this?

orpiske: Btw, I'll merge this for now and we can think about this later

PR #8881, opened 2022-12-12
[Camel-18808] add support for CamelAzureStorageBlobBlobUploadSize in uploadBlockBlob operation

jwwallin:
This PR adds support for `CamelAzureStorageBlobBlobUploadSize`-header in the `camel-azure-storage-blob`-component producer. 

Specifically this adds the ability to override `BlobUtils.getInputStreamLength()` in `BlobStreamAndLength` by providing a value to the `CamelAzureStorageBlobBlobUploadSize`-header. To allow this change to work with the Azure SDK client, the `BloblClientWrapper` needed a change in how it calls `BlockBlobClient.uploadWithResponse()`. Functionally this change to the wrapper should be negligible as it does things very close to how the convenience method in the SDK, which was previously used, works.


davsclaus:
is there a problem if the same header is used in both consumer and producer, if you go ala

from s3-blob
   set Body
         // change body here
    to s3-blob

Then the old header with the length may be in use

jwwallin:
> is there a problem if the same header is used in both consumer and producer, if you go ala
> 
> from s3-blob set Body // change body here to s3-blob
> 
> Then the old header with the length may be in use, instead of calculating a new length because the body is changed. We have this problem in some of the other components, where you then have to remove headers

It would then propagate through, that's true... Would it be more convenient to then use a separate header for producer operations?

jwwallin:
I've made the suggested change. I'm not 100% certain the placement of the header removal is exactly the same as other components. If there is a better way of handling that, I will obviously make that change.

I have verified that this solution fixes the problem I was having with our integration.

components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobStreamAndLength.java
```diff
@@ -40,14 +40,16 @@ private BlobStreamAndLength(InputStream inputStream, long streamLength) {
     @SuppressWarnings("rawtypes")
     public static BlobStreamAndLength createBlobStreamAndLengthFromExchangeBody(final Exchange exchange) throws IOException {
         Object body = exchange.getIn().getBody();
+        Long blobSize = exchange.getIn().getHeader(BlobConstants.BLOB_UPLOAD_SIZE, () -> null, Long.class);
```
davsclaus, line 43: You can maybe remove the header here instead?

davsclaus: As the remove will give you back the header

jwwallin: I was wondering about that as-well. It's at least better than the current spot, since it's in context. :+1:

jwwallin: Change done.

jwwallin:
Would this get backported to 3.18.x as well? That's the version we are currently using.

davsclaus (approved):

PR #8883, opened 2022-12-12
camel-kafka: prevent exceptions in close from leaking (CAMEL-18796)

orpiske:


oscerd (approved):

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/SingleNodeKafkaResumeStrategy.java
```diff
@@ -396,6 +400,8 @@ public void stop() {
             IOHelper.close(producer, "Kafka producer", LOG);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
+        } catch (Exception e) {
+            LOG.error("Error closing the Kafka producer: {} (this error will be ignored)", e.getMessage(), e);
```
davsclaus, line 404: Should this be WARN like the other

orpiske: Ah, definitely it should! Thanks for pointing it

davsclaus (approved):

PR #8907, opened 2022-12-15
CAMEL-18816: camel-ahc component crashes when traffic starts too early

krzysztof-mackowiak:
Fix for bug CAMEL-18816.

Problem is caused by the fact that  `createProducer()` method in `AhcEndpoint` class may be executed before `doStart()` method.

This PR contains 3 tests that reproduce a problem and a fix based on solution from HttpEndpoint/HttpDestination.

Creation of `client` field in `AhcEndpoint` has been moved from `doStart()` method to `getOrCreateClient()` method. This `getOrCreateClient()` method is invoked from `AhcProducer(AhcEndpoint endpoint)` constructor.

### DIFF
--- a/components/camel-ahc/src/main/java/org/apache/camel/component/ahc/AhcEndpoint.java
+++ b/components/camel-ahc/src/main/java/org/apache/camel/component/ahc/AhcEndpoint.java
@@ -16,7 +16,9 @@
  */
 package org.apache.camel.component.ahc;
 
+import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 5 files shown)


davsclaus (approved):

davsclaus:
Ideally you would just create the http client in doStart in the producer, and not create it via a getter (bad practice). Then there is no need for synchrnoization etc.

However since the component is deprecated and removed in newer releases then this is good for a patch release.

oscerd (approved):

PR #8909, opened 2022-12-15
Upgrade to kubernetes client 6.3.1

davsclaus:


davsclaus:
the client is possible not fully synced to central, so we may need to re-kick this job later

oscerd (approved):

PR #8914, opened 2022-12-18
camel-jbang - Upgrade to Camel 3.20.0

tadayosi:


davsclaus:
Yes when the 3.20 is GA and in maven central then we can merge

oscerd (approved):

PR #8928, opened 2022-12-19
CAMEL-18829: camel-yaml-dsl - generate-yaml-schema - Support `additio…

igarashitm:
…nalProperties: false`


igarashitm:
An example of having `additionalProperties` to be `false`
https://github.com/igarashitm/example-generate-yaml-schema/blob/main/pom.xml#L71

And generated YAML schema from it
https://github.com/igarashitm/example-generate-yaml-schema/commit/85441fdcc6c641a2ef02aea7149a66a0429b3c92#diff-b1d8ab8b59f212acab62fab8cab033f98736c4bd9e92cc34fcf9b53457dbc1d0

davsclaus (approved):

PR #8911, opened 2022-12-16
CAMEL-18661: clean-ups

bvahdat:
As follow-up with https://github.com/apache/camel/pull/8713

PR #8934, opened 2022-12-20
Tracing: fix flaky test and add scope checks before closing

lmolkova:
Follow up for #8713 and #8911.

Fixes flaky `CurrentSpanTest.testContextDoesNotLeak` test.
I was able to repro failures locally and with this change, they are no longer reproducible (tried on 1000 iterations).

 There are several problems:
1. We can't guarantee that `ExchangeAsyncProcessingStartedEvent` is fired once and only if a span was created previously
2. We can't guarantee that span ends on the same thread it was started, i.e. scope is closed on the same thread
3. Due to concurrency in async scenarios we can call close on scope two times

As a result, we're closing scopes multiple times and from the wrong threads. Changing these on Camel side, AFAIK would be quite difficult.

Unfortunately, OTel scope closing works differently than I thought - https://github.com/open-telemetry/opentelemetry-java/issues/5055. If called multiple times and on a different thread than the scope was started, it restores the wrong context on the other thread.

While the solution on OTel side is evaluated, I suggest guarding scope closing with thread check and making it idempotent.

This PR also adds logging in case there is more flakiness.

### DIFF
--- a/components/camel-opentelemetry/src/test/resources/log4j2.properties
+++ b/components/camel-opentelemetry/src/test/resources/log4j2.properties
@@ -24,6 +24,6 @@ appender.out.name=out
 appender.out.layout.type=PatternLayout
 appender.out.layout.pattern=%d [%-15.15t] %-5p %-30.30c{1} - %m%n
 logger.opentelemetry.name=org.apache.camel.opentelemetry
-logger.opentelemetry.level=INFO
+logger.opentelemetry.level=DEBUG
 rootLogger.level=INFO
 rootLogger.appenderRef.file.ref=file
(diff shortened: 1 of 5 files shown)


components/camel-opentelemetry/src/test/resources/log4j2.properties
```diff
@@ -24,6 +24,6 @@ appender.out.name=out
 appender.out.layout.type=PatternLayout
 appender.out.layout.pattern=%d [%-15.15t] %-5p %-30.30c{1} - %m%n
 logger.opentelemetry.name=org.apache.camel.opentelemetry
-logger.opentelemetry.level=INFO
+logger.opentelemetry.level=DEBUG
```
oscerd, line 27: I think should be INFO again

oscerd (approved):

components/camel-opentelemetry/src/test/resources/log4j2.properties
lmolkova: sure, let's just see if tests pass in CI

davsclaus (approved):

PR #8936, opened 2022-12-21
Problem when parameter type of bean method is String.

Luke-hbk:
[camel-bean] 
Problem when parameter type of bean method is String.
Problems with more than 2 parameters.

Expression to evaluate the bean parameter parameters and provide the correct values when the method is invoked.

There are several problems:
1. There is a problem with quotes being removed in the splitSafeQuote method used inside the evaluate method.
   splitSafeQuote("'Camel', 'Test'", ",", true)  -> ["Camel","Test"]   
2. The quote is removed and false is returned from the isValidParameterValue method. 
   isValidParameterValue("Camel") -> return false
3. When “ClassLoader” is working which make thread blocked it caused by Disk I/O.
   So it behavior make that performance deteriorates.

This PR also adds TestCase.

### DIFF
--- a/components/camel-bean/src/test/java/org/apache/camel/component/bean/BeanInfoTest.java
+++ b/components/camel-bean/src/test/java/org/apache/camel/component/bean/BeanInfoTest.java
@@ -17,6 +17,7 @@
 package org.apache.camel.component.bean;
 
 import java.util.Collections;
+import java.util.Iterator;
 
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.description.modifier.Ownership;
@@ -27,6 +28,8 @@ import net.bytebuddy.implementation.bind.annotation.RuntimeType;
 import org.apache.camel.CamelContext;
 import org.apache.camel.Handler;
 import org.apache.camel.spi.Registry;
+import org.apache.camel.support.ObjectHelper;
+import org.apache.camel.util.StringQuoteHelper;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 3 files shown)


essobedo:
Hi thank you for your contribution, would you mind providing an example of a route to illustrate the problem?

Luke-hbk:
I am using XML DSL.

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<route xmlns="http://camel.apache.org/schema/spring" customId="true" id="Chatbot" streamCache="true">
   <from uri="rest:post:credit"/>
    <to id="to55" uri="bean:PropertyProcessor?method=RAW(combineHeaderAndBody(*,'start','test'))"/>
</route>


image

essobedo:
I'm sorry but your problem is still not clear to me, you are obviously mixing several things (quote and perf issues) which makes it harder to understand.

Could you please clarify each problem you meet?

essobedo:
At least, please describe what you get and what you expect.

components/camel-bean/src/test/java/org/apache/camel/component/bean/BeanInfoTest.java
```diff
@@ -108,6 +111,28 @@ public String myMethod() {
...
+    @Test
+    public void testSplitSafeQuoteRemoveQuoteAfterValidParameterTypeCheck() {
+        String[] out = StringQuoteHelper.splitSafeQuote("'\"Hello,Camel\"','\"Bye,World\"'", ',');
+        Iterator<?> it = ObjectHelper.createIterator(out, ",", true);
+        Object parameterValue = it != null && it.hasNext() ? it.next() : null;
+        while (parameterValue != null) {
+            assertTrue(BeanHelper.isValidParameterValue(parameterValue.toString()));
+            parameterValue = it != null && it.hasNext() ? it.next() : null;
+        }
+    }
```
essobedo, line 134: The tests for `StringQuoteHelper` are [here](https://github.com/apache/camel/blob/main/core/camel-core/src/test/java/org/apache/camel/util/StingQuoteHelperTest.java)

essobedo: You will see in the different tests that it is already possible to get `'Hello'` with single quotes if you use double quotes in your expression like `"'Hello'"`.

Luke-hbk:
Thank you for your answer.
I understand because of your answer. 

But I have one question.
If there is only one parameter, the quote is not removed.

StringQuoteHelper.java 
if (input.indexOf(separator) == -1) {
            // no separator in data, so return single string with input as is
            return new String[] { trim ? input.trim() : input };
}

The quote is not removed because it is returned in the above code.

Example
 out = StringQuoteHelper.splitSafeQuote("'Hello'", ',', true);
 assertEquals(**"'Hello'"**, out[0]); -> true

 out = StringQuoteHelper.splitSafeQuote("'Hello', 'Camel'", ',', true);
 assertEquals(**"Hello"**, out[0]); -> true

Check if this is the intended behavior.

essobedo:
> Example out = StringQuoteHelper.splitSafeQuote("'Hello'", ',', true); assertEquals(**"'Hello'"**, out[0]); -> true
> 
> out = StringQuoteHelper.splitSafeQuote("'Hello', 'Camel'", ',', true); assertEquals(**"Hello"**, out[0]); -> true
> 
> Please check if this is the intended behavior.

The first result sounds weird indeed. @davsclaus could you please confirm if it is a bug or not? I'm also wondering if there is a way to escape a single quote if single quotes have chosen like in the expression `"'John's'"`

davsclaus:
Yeah the 1st example is a little bug.
For 2nd, then you MUST start with double quotes

    "John's"

essobedo:
> Yeah the 1st example is a little bug. For 2nd, then you MUST start with double quotes
> 
> ```
> "John's"
> ```

We still need to be able to escape the quotes otherwise what can we do with an expression with both type of quotes like `"John's \"big\" house"`

davsclaus:
I took a look the string quote helper, and created a PR with a fix for the little bug when there is only 1 element. This change affected kamelets and route templates / property binding via contrustor arguments that shouldn't be trimmed

Here is a PR
https://github.com/apache/camel/pull/8941

davsclaus:
> > Yeah the 1st example is a little bug. For 2nd, then you MUST start with double quotes
> > ```
> > "John's"
> > ```
> 
> We still need to be able to escape the quotes otherwise what can we do with an expression with both type of quotes like `"John's \"big\" house"`

That is NOT intended to be supported, - its for normal use-cases

davsclaus:
@Luke-hbk did you find out something else in item no 3 and what was that?

PR #8944, opened 2022-12-23
[CAMEL-15111] Fixes bug in camel-as2 via not forcing ascii charset when not specified

yasserzamani:
... otherwise you will get `java.nio.charset.MalformedInputException: Input length = 1` when content transfer encoding is binary. This PR mainly does AS2SessionInputBuffer.setCharsetDecoder(null) instead when charset is not specified so it doesn't reach MalformedInputException. Then it convert returned result back to byte[] from String when charset is not specified. String were required to find multipart boundaries. Generally this PR keeps previous behavior otherwise, i.e. when charset is specified. It also tests all 8 different situations for encrypt, sign and compress via a parameterized test. as2-lib third-party library is used for these tests as at the moment, binary content transfer encoding is not supported by Camel AS2 client.

### DIFF
--- a/components/camel-as2/camel-as2-api/pom.xml
+++ b/components/camel-as2/camel-as2-api/pom.xml
@@ -100,7 +100,7 @@
         <dependency>
             <groupId>com.helger.as2</groupId>
             <artifactId>as2-lib</artifactId>
-            <version>4.11.0</version>
+            <version>${as2-lib-version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/components/camel-as2/camel-as2-api/pom.xml
+++ b/components/camel-as2/camel-as2-api/pom.xml
@@ -96,6 +96,13 @@
             <artifactId>camel-test-junit5</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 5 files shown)


components/camel-as2/camel-as2-api/pom.xml
```diff
@@ -96,6 +96,13 @@
             <artifactId>camel-test-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- added for binary content transfer encoding test because Camel AS2 client doesn't support binary currently -->
+        <dependency>
+            <groupId>com.helger.as2</groupId>
+            <artifactId>as2-lib</artifactId>
+            <version>4.11.0</version>
+            <scope>test</scope>
```
oscerd, line 104: This should be in a property placeholder in parent pom. Don't use a static version inside the component pom

oscerd (approved):

oscerd (changes requested):
Just a minor one

components/camel-as2/camel-as2-api/pom.xml
yasserzamani: Thanks! There were a lot to do and I forgot to do so while I was going to. Nice catch :+1: fixed

davsclaus (approved):

PR #8955, opened 2022-12-28
[CAMEL-5963] camel-smpp: add Transceiver (TRX) support

yasserzamani:
Fixes CAMEL-5963 via adding a new producer specific uri param so called `messageReceiverRouteId` which is however optional and so backward compatible (i.e. if not set then previous TX behavior will be seen exactly as before). If set then the user wouldn't define a corresponding redundant consumer. Camel will use the specified route above as corresponding consumer - internally it uses one and same SmppSession as producer.
Example:
```java
                from("direct:start")
                        .to("smpp://j@localhost:8056?password=jpwd&systemType=producer" +
                            "&messageReceiverRouteId=testMessageReceiverRouteId");

                from("direct:messageReceiver").id("testMessageReceiverRouteId")
                        .to("mock:result");
```

CamelContext startup would fail in the first place on a wrong `messageReceiverRouteId` value via registering a StartupListener when no route can be found for the specified id. This has integrated tested in org/apache/camel/component/smpp/integration/SmppTRXProducerIT.java

NOTE: When user's SMPP Server doesn't support TRX, still possible and she/he has to define separate producer (TX by default) and consumer (unchanged and RX by default) as before.


yasserzamani:
I'm not sure why checkstyle fails. I had verified (and also verified it now) locally. It's fine locally!

davsclaus (approved):

simonalexander2005:
Does this also work with `toD`, where the `messageReceiverRouteId` is configured in a parameter; or just with `to`?

PR #8964, opened 2023-01-03
(chores) camel-arangodb: avoid risky unboxing

orpiske:


components/camel-arangodb/src/main/java/org/apache/camel/component/arangodb/ArangoDbProducer.java
```diff
@@ -116,7 +116,7 @@ private Function<Exchange, Object> saveDocument() {
                 Boolean isMultiInsert = (Boolean) exchange.getMessage().getHeader(MULTI_INSERT, false);
```
davsclaus, line 116: getHeader(x,y, Boolean.class) would then avoid boxing also

orpiske: Yeah, that's even better.

davsclaus (approved):

PR #8971, opened 2023-01-04
camel-plc4x: Added autoReconnect uri parameter

SteinOv:
I don't have a JIRA account yet, so I couldn't create an issue (I send an email to private@camel.apache.org, is that the right approach?).

I added an uri parameter "autoReconnect". If enabled, when a request is made and there is no connection, an attempt to reconnect is made.


oscerd (approved):

components/camel-plc4x/src/main/java/org/apache/camel/component/plc4x/Plc4XConsumer.java
```diff
@@ -85,6 +86,15 @@ protected void doStart() throws Exception {
     }
 
     private void startUnTriggered() {
+        if (plc4XEndpoint.isAutoReconnect() && !plcConnection.isConnected()) {
+            try {
+                plc4XEndpoint.reconnect();
+                LOGGER.debug("Successfully reconnected");
+            } catch (PlcConnectionException e) {
+                LOGGER.error("Unable to reconnect, skipping request", e);
```
davsclaus, line 94: Maybe tone down the logging to WARN as you are dealing with this error by skipping.

davsclaus:
In case the connection is NOT connected and auto reconnect = false, would the connection be "bad" all the time, or can the connection somehow re-establish itself and become working again ?

SteinOv:
> In case the connection is NOT connected and auto reconnect = false, would the connection be "bad" all the time, or can the connection somehow re-establish itself and become working again ?

The connection could be re-established manually, for example in a custom processor by using Plc4XEndpoint.getConnection().connect(). The connection cannot re-establish itself in any way before my changes.

davsclaus (approved):

PR #9015, opened 2023-01-08
[CAMEL-4616] camel-jms: test not loosing messages in async TX sends

yasserzamani:
Invalidates CAMEL-4616.

I think that the test provided by this PR proves that camel JMS doesn't loose messages in transaction when AtiveMQ uses `async send` internally and also asynchronously fails. See the test comments to understand what the pull request does, how, and why.

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/issues/AsyncJmsProducerExceptionInTXTest.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/issues/AsyncJmsProducerExceptionInTXTest.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (approved):

yasserzamani:
@davsclaus yes I also saw same thing when I ran camel-jms tests locally on my machine. But I also saw same case for an another test not related at all to my changes (which passed in second run). So I thought maybe it's normal and why camel-jms tries tests multiple times. Should I fix it?

davsclaus:
No you do not have to fix it unless you can easily spot what can be done.

yasserzamani:
@davsclaus by the way, another reason that made me happy to see such exception and to not fix it was that it's an another confirmation and proof for what I'm going to prove with this PR. It showed me that in case of wrong username/password, camel-jms doesn't loose messages when it's in an async transaction send :) maybe I should OR two assertions together.

yasserzamani:
@davsclaus I could fixed it by using a new broker instead of shared one: see #9021 please :)

PR #9016, opened 2023-01-08
CAMEL-18864: java-joor-dsl - Drop support of class files

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-18864

## Motivation

I think we can remove the .class also in camel core for `java-joor` dsl as it is not really needed anymore. I think (I am getting older so it is maybe no longer there) there is an option in camel-main to configure a "compile directory" so removing would also need to remove that.

## Modifications

* Remove the `ClassRoutesBuilderLoader`
* Remove "compile directory" and "compile load first" everywhere it is defined and used
* Remove all related unit tests
* Add a note in the upgrade guide


oscerd (approved):

davsclaus (approved):

davsclaus:
I think the option is also in camel spring boot

essobedo:
The checkstyle issue is not related to those changes, the problem is in `camel-jbang-core`:
```
[camel-jbang-core] [INFO] Starting audit...
[ERROR] /home/runner/work/camel/camel/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java:270:5: Method run length is 253 lines (max allowed is 250). [MethodLength]
Audit done.
```

essobedo:
@davsclaus I can't see it in camel spring boot, could you please share with me a link to it?

davsclaus:
Ah yeah its not in spring boot as it was only needed via standalone via camel-main / jbang at the time. So this is done

PR #9042, opened 2023-01-11
CAMEL-18878 Autowiring on endpoint works even if is disabled on component

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-18878

Fix + JUnit test

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/impl/DefaultComponentAutowiredFalseTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/DefaultComponentAutowiredFalseTest.java
@@ -22,12 +22,17 @@ import java.net.CookiePolicy;
 import java.util.Map;
 
 import org.apache.camel.CamelContext;
+import org.apache.camel.Consumer;
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.Endpoint;
+import org.apache.camel.Processor;
+import org.apache.camel.Producer;
 import org.apache.camel.spi.GeneratedPropertyConfigurer;
 import org.apache.camel.spi.PropertyConfigurerGetter;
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus:
I am not sure this is correct, as the endpoint option should be able to overrule the component.

eg the component can be = false, but endpoint = true or vice versa.
Usually camel will set the endpoint option to the same value as component when creating the endpoint, and then set endpoint specific value (if provided in uri)

JiriOndrusek:
@davsclaus  I found the text - https://github.com/apache/camel/blob/main/core/camel-support/src/main/java/org/apache/camel/support/DefaultEndpoint.java#L59

which says that component has higheh priority with aurowiredEnabled

"Whether autowiring is enabled. This is used for automatic autowiring options (the option must be marked as autowired)"
                            + " by looking up in the registry to find if there is a single instance of matching type, which then gets configured on the component."
                            + " This can be used for automatic configuring JDBC data sources, JMS connection factories, AWS Clients, etc."
                            + " Important: If a component has the same option defined on both component and endpoint level, then disabling"
                            + " autowiring on endpoint level would not affect that the component will still be autowired, and therefore the endpoint"
                            + " will be configured with option from the component level. In other words turning off autowiring would then require to turn it off on the component level.

JiriOndrusek:
My intention is to allow disable auwired on component and then endpoint default value `true` for the property is ignored  Because without the fix, if I disable component, endpoint overrides it with default `true` and the component disablement has no effect.

davsclaus:
Ah yeah good point as the component would be autowired before creating the endpoint, and therefore turning it off on endpoint is too late.

davsclaus (approved):

PR #9059, opened 2023-01-12
(chores) camel 4: adjust the contribution template to help users identify the correct branch

orpiske:
# Description

# Target

# Tracking


orpiske:
I adjusted the template to take into consideration Camel 3 and Camel 4. 

@oscerd / @davsclaus should we keep this templated enabled by default during Camel 4 development? Any thoughts?

oscerd (approved):

oscerd:
Sounds good to me. We should harmonize the template even in the subprojects though. Not only for Camel 3/Camel 4, but in general

orpiske:
> Sounds good to me. We should harmonize the template even in the subprojects though. Not only for Camel 3/Camel 4, but in general

Yeah. I've been thinking about cleaning up our contribution guides. Having a common template would make things easier too.

PR #9086, opened 2023-01-13
CAMEL-18912: Reuse client from consumer

simon-ras:
Before this change, autowired clients would fail to be detected.

Also small optimization added by only requesting a single queue when listing the queues.

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

davsclaus:
@simon-ras I wonder if you have time to check the other health checks in the AWS as they may also benefit from reusing client

simon-ras:
I can give it a shot - though it might take a few days due to limited spare time. 

* Do you prefer 1 PR per component fixed, or 1 commit per component, or just a single commit including everything?
* Also, what is the best procedure for backporting this - should I make available PR with cherry-picked commit for each version, or better let you take care of this when merging the `main` PR?

davsclaus:
1 per component as if you do not get time to complete them all then others can take over later

davsclaus:
we will backport

oscerd:
Don't worry about the backport operations, we'll do it. Thanks for your time and for reporting.

oscerd (approved):

simon-ras:
Forgot to format prior to pushing. Will reformat and push again.

simon-ras:
Yes, for all health checks. There is still a few components which does not have health checks, they are listed here https://issues.apache.org/jira/browse/CAMEL-18131. They continue to use ComponentVerifierExtension.

I just discovered that there are some test which fail though, so this PR is not quite ready yet...

simon-ras:
I believe `AWS2S3ConsumerHealthCustomClientTest` - which is the failing test - is invalid. These tests all use the client which is set on the `configuration` and since the healthcheck now uses the client of the endpoint, then such case cannot be tested.

Do you mind @oscerd do you agree?

oscerd:
It's important to make it works in all the cases. I think your fix will work only in case of client set on endpoint.. what happens if I specify keys secret and access? It should work too..

oscerd:
Ah right. You're right Simon.

oscerd (approved):

davsclaus (approved):

davsclaus:
@simon-ras can we cherry pick this commit to camel-3.x or would you be able to do another PR for that branch so we get this into 3.x releases too

simon-ras:
> @simon-ras can we cherry pick this commit to [camel-3](https://issues.apache.org/jira/browse/CAMEL-3).x or would you be able to do another PR for that branch so we get this into 3.x releases too

Sure, please see https://github.com/apache/camel/pull/9105

PR #9115, opened 2023-01-16
CAMEL-18928: removed component camel-datasonnet

orpiske:


oscerd (changes requested):
Don't we disable them only? Without removing them?

zhfeng:
Where can we put these disabled components?

orpiske:
> Don't we disable them only? Without removing them?

My suggestion to @davsclaus earlier today was to: tag the code, remove them and keep the code base and git history clean. Once they are available, we can simply revert the change and do that on top of it. 

@davsclaus did I understand it differently?

orpiske:
> Where can we put these disabled components?

There's a list of components to be disabled here: https://issues.apache.org/jira/browse/CAMEL-18928. You can add them to the list and I'll take care of removing them cleanly.

davsclaus (approved):

zhfeng (approved):

orpiske:
Folks, can I merge this one? 

> ### Components tested:
> 2 	2 	0 	0

PR #9143, opened 2023-01-17
CAMEL-14008 Another try to add a connection pool to camel-as2

McIntozh:
As CAMEL-14008 has two unsuccessfully closed pull requests, I'm adding a third one that will hopefully be accepted.

# Description

I've fixed some issues with the other pull requests and made timeouts and pool size configurable via URL parameters.
Also the we now respect the connection keep-alive/close strategy from the server when releasing it back to the pool (that was for example an issue when sending multiple requests to the mendelson as2 test instance). 
Also resolved a deprecation for Exchange.getOut().

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-14008) filed for the change (usually before you start working on it).


components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2ClientConnection.java
```diff
@@ -41,21 +55,30 @@
...
-                               Integer targetPortNumber) throws IOException {
+                               Integer targetPortNumber, Duration socketTimeout, Duration connectionTimeout,
+                               Integer connectionPoolMaxSize, Duration connectionPoolTtl) throws IOException {
 
         this.as2Version = Args.notNull(as2Version, "as2Version");
         this.userAgent = Args.notNull(userAgent, "userAgent");
         this.clientFqdn = Args.notNull(clientFqdn, "clientFqdn");
         this.targetHost = new HttpHost(
                 Args.notNull(targetHostName, "targetHostName"), Args.notNull(targetPortNumber, "targetPortNumber"));
+        Args.notNull(socketTimeout, "socketTimeout");
```
davsclaus, line 78: Use Camels ObjectHelper.notNull (in camel-util)

davsclaus (changes requested):
only small nitpick about to use camel apis for null check

McIntozh:
OK, didn't know that ObjectHelper.notNull exitsts. I've committed the fix.

davsclaus:
Did you replace all of `Args`  with ObjectHelper ?

McIntozh:
Looks like I missed some..
When is Args.notNull to be used and when ObjectHelper.notNull?
Should I replace all Args.notNull in the project? Because most of them already existed before my change.
Or should validating via Args.notNull be used when assigning a method parameter to a class property and ObjectHelper.notNull for parameters that are not assigned?

davsclaus:
Ah okay so if there are more Args then we should use Camel APIs instead as we do that in the other components.

davsclaus (approved):

PR #9160, opened 2023-01-18
CAMEL-18932: fixed camel-solr tests due to sl4fj upgrade

orpiske:


oscerd (approved):

orpiske:
Bypassing checks so we can get this on the next CI build.

PR #9177, opened 2023-01-20
(chores) camel-kafka: disable connection check test on Apache CI

orpiske:
# Description

This test runs OK locally, but fails on Apache CI.

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

orpiske:
Actually, I think this test is wrong as it seems to depend on a complex chain of events to be correct (service needs to be *fully* up, Camel - and the Kafka client - needs to be connect, partition/client metadata has to be properly synced/up-to-date). It seems to work well just on machines fast enough for this to happen in a timely manner. 

I'll remove it.

PR #9181, opened 2023-01-20
Prevent high channel churn in a queue full scenario

groundhog2k:
In a scenario with a high message publishing rate and a full queue the publisher will get "nack" (not acknowledged) as feedback from RabbitMQ. 

The method `Channel#waitForConfirmsOrDie(long)` in https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/ChannelN.java#L241 will close the channel. This closed channel is then given back to the channel pool.  Next time the channel is taken out of the pool it is detected as closed and will be removed. So a new channel needs to be opened and this too will be closed directly after usage when the same queue is still full. There will be a high rate of opened and closed communication channels and the channel pool gets unusable. 

This is described as high channel churn (https://www.rabbitmq.com/channels.html#high-channel-churn) and can have side effects on RabbitMQ (memory and cpu resource usage) and performance too.

It was also discussed with folks from rabbitmq-java-client (rabbitmq/rabbitmq-java-client#942) and the usage of Channel#waitForConfirmsOrDie(long) is "highly NOT recommended" by them.

This behavior affects ALL versions of the camel-rabbitmq component!

### DIFF
--- a/components/camel-rabbitmq/src/main/java/org/apache/camel/component/rabbitmq/RabbitMQMessagePublisher.java
+++ b/components/camel-rabbitmq/src/main/java/org/apache/camel/component/rabbitmq/RabbitMQMessagePublisher.java
@@ -162,11 +162,25 @@ public class RabbitMQMessagePublisher {
     private void waitForConfirmation() throws IOException {
         try {
             LOG.debug("Waiting for publisher acknowledgements for {}ms", endpoint.getPublisherAcknowledgementsTimeout());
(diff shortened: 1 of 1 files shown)


oscerd (approved):
Thanks for this.

This should go also on camel-3.x. This component is deprecated and it has been removed in camel 4.x, in favor of camel-spring-rabbitmq.

Please open a PR again camel-3.x too.

oscerd:
Please don't close this one, this is a bug, so it should go on 3.20.x too, thanks.

davsclaus (approved):

PR #9183, opened 2023-01-20
Prevent high channel churn in a queue full scenario

groundhog2k:
In a scenario with a high message publishing rate and a full queue the publisher will get "nack" (not acknowledged) as feedback from RabbitMQ.

The method Channel#waitForConfirmsOrDie(long) in https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/ChannelN.java#L241 will close the channel. This closed channel is then given back to the channel pool. Next time the channel is taken out of the pool it is detected as closed and will be removed. So a new channel needs to be opened and this too will be closed directly after usage when the same queue is still full. There will be a high rate of opened and closed communication channels and the channel pool gets unusable.

This is described as high channel churn (https://www.rabbitmq.com/channels.html#high-channel-churn) and can have side effects on RabbitMQ (memory and cpu resource usage) and performance too.

It was also discussed with folks from rabbitmq-java-client (rabbitmq/rabbitmq-java-client#942) and the usage of Channel#waitForConfirmsOrDie(long) is "highly NOT recommended" by them.

This behavior affects ALL versions of the camel-rabbitmq component!

### DIFF
--- a/components/camel-rabbitmq/src/main/java/org/apache/camel/component/rabbitmq/RabbitMQMessagePublisher.java
+++ b/components/camel-rabbitmq/src/main/java/org/apache/camel/component/rabbitmq/RabbitMQMessagePublisher.java
@@ -179,7 +179,7 @@ public class RabbitMQMessagePublisher {
                 // Because we don't know why timeout happend (Maybe a communication problem)
                 channel.close(AMQP.PRECONDITION_FAILED, "TIMEOUT WAITING FOR ACK");
             } catch (Exception ce) {
-              LOG.warn("Caught exception during closing of channel", ce);
+                LOG.warn("Caught exception during closing of channel", ce);
             }
             LOG.warn("Acknowledgement error for {}", camelExchange);
             throw new RuntimeCamelException(e);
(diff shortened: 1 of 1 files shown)


oscerd (approved):

groundhog2k:
What is the reason that the pipeline failed? I can't see any root cause in the logs - only "error 1" !? Anything I can do/fix here?

essobedo:
The checkstyle issue is not related to your changes but is more in `camel-jbang-core`:

```
[ERROR] /home/runner/work/camel/camel/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/CodeRestGenerator.java:108:22: Name 'CODE' must match pattern '^[a-z][a-zA-Z0-9]*$'. [LocalFinalVariableName]
[ERROR] /home/runner/work/camel/camel/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/CodeRestGenerator.java:109:22: Name 'GENERATOR_NAME' must match pattern '^[a-z][a-zA-Z0-9]*$'. [LocalFinalVariableName]
[ERROR] /home/runner/work/camel/camel/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/CodeRestGenerator.java:110:22: Name 'LIBRARY' must match pattern '^[a-z][a-zA-Z0-9]*$'. [LocalFinalVariableName]
```

davsclaus (approved):

PR #9208, opened 2023-01-25
CAMEL-18977: camel-joor - Make the code more flexible for native mode

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-18977

## Motivation

With the current code base, it is hard to change the behavior of the jOOR language in order to move most of the logic at build time and/or at static initialization time, without having to use ugly hacks like reflection. The goal of this task is to make the code more flexible to avoid using hacks. 

## Modifications:

* Add a constructor to the class `JoorLanguage` to be able to inject specific jOOR compiler and jOOR script compiler
* Add getters for the jOOR compiler and jOOR script compiler
* Make the method `evalCode` public to give access to the generated code.


essobedo:
Waiting https://github.com/apache/camel-quarkus/pull/4440 to be approved and merged first

oscerd (approved):

PR #9210, opened 2023-01-25
Fix tests

Croway:
# Description

# Target

# Tracking

- Fix jms flaky test
- CAMEL-18978
- Fix tests in camel-atmosphere-websocket, after jetty and atmosphere update some messages were consumed by wrong endpoints, renaming them did the trick


Croway:
@oscerd @orpiske if this PR is approved, useStreaming -> streaming rename should be added to the release notes, where are those?

oscerd (approved):

components/camel-atmosphere-websocket/src/main/java/org/apache/camel/component/atmosphere/websocket/WebsocketEndpoint.java
```diff
@@ -86,15 +86,15 @@ public void setSendToAll(boolean sendToAll) {
...
+    public boolean isStreaming() {
+        return streaming;
     }
 
     /**
      * To enable streaming to send data as multiple text fragments.
      */
-    public void setUseStreaming(boolean useStreaming) {
-        this.useStreaming = useStreaming;
+    public void setStreaming(boolean streaming) {
```
essobedo, line 96: I hope there is a better way, renaming a property should be avoided

essobedo:
> @oscerd @orpiske if this PR is approved, useStreaming -> streaming rename should be added to the release notes, where are those?

In the migration guide user-manual/modules/ROOT/pages/camel-4-migration-guide.adoc

components/camel-atmosphere-websocket/src/test/java/org/apache/camel/component/atmosphere/websocket/WebsocketRouteWithInitParamTest.java
```diff
@@ -110,18 +114,18 @@ void testWebsocketSingleClientBroadcastMultipleClientsGuaranteeDelivery() throws
...
-        WebsocketTestClient wsclient1 = new WebsocketTestClient("ws://localhost:" + PORT + "/hola3", 2);
-        WebsocketTestClient wsclient2 = new WebsocketTestClient("ws://localhost:" + PORT + "/hola3", 2);
-        WebsocketTestClient wsclient3 = new WebsocketTestClient("ws://localhost:" + PORT + "/hola3", 2);
+        WebsocketTestClient wsclient1 = new WebsocketTestClient("ws://localhost:" + PORT + "/guarantee", 2);
+        WebsocketTestClient wsclient2 = new WebsocketTestClient("ws://localhost:" + PORT + "/guarantee", 2);
+        WebsocketTestClient wsclient3 = new WebsocketTestClient("ws://localhost:" + PORT + "/guarantee", 2);
 
         wsclient1.connect();
-        wsclient1.await(awaitTime);
+        wsclient1.await(1);
```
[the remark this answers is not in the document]
Croway: in this way the test runs faster, I noticed that when connecting a high await time is not useful, moreover, there are 3 consequent await

components/camel-atmosphere-websocket/src/test/java/org/apache/camel/component/atmosphere/websocket/WebsocketRouteWithInitParamTest.java
```diff
@@ -41,6 +43,8 @@ public class WebsocketRouteWithInitParamTest extends WebsocketCamelRouterWithIni
     private static String[] broadcastMessageTo = {};
     private static Map<String, String> connectionKeyUserMap = new HashMap<>();
 
+    private ExecutorService executor = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
Croway: nice catch, it is just a leftover

components/camel-atmosphere-websocket/src/main/java/org/apache/camel/component/atmosphere/websocket/WebsocketEndpoint.java
Croway: I've spent some time, and the only solution I found is renaming that property, I still cannot explain why, but, when useStreaming=true is used, all the messages are consumed by the same endpoint. https://issues.apache.org/jira/browse/CAMEL-18978

Croway:
> > @oscerd @orpiske if this PR is approved, useStreaming -> streaming rename should be added to the release notes, where are those?
> 
> In the migration guide [user-manual/modules/ROOT/pages/camel-4-migration-guide.adoc](https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/camel-4-migration-guide.adoc)

Thanks!, I've updated the migration guide https://github.com/apache/camel/pull/9210/commits/9afb26287ed1c765d8d94b40ed744f3e89b7a068

components/camel-atmosphere-websocket/src/test/java/org/apache/camel/component/atmosphere/websocket/WebsocketRouteWithInitParamTest.java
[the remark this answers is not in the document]
orpiske: Hm, I am afraid this may cause problems in our Apache CI. Our nodes there are not so fast, so the the time needed to receive the messages could be higher.

components/camel-atmosphere-websocket/src/main/java/org/apache/camel/component/atmosphere/websocket/WebsocketEndpoint.java
orpiske: TBH, this is still puzzling me

orpiske (changes requested):
Thanks @Croway. I left a note about a potential impact of the change on the nodes in our CI. Marking as request changes, but please let me know in case I misunderstood something.

docs/user-manual/modules/ROOT/pages/camel-4-migration-guide.adoc
```diff
@@ -59,3 +59,7 @@ All the `camel-test` modules that were JUnit 4.x based has been removed. All tes
 === camel-caffeine
 
 The `keyType` parameter has been removed. The Key for the cache will now be only `String` type. More information in CAMEL-18877.
+
+=== camel-atmosphere-websocket
+
+The `useStreaming` parameter was renamed to `streaming`. More information in CAMEL-18978.
```
essobedo, line 65: Could be removed

components/camel-atmosphere-websocket/src/main/java/org/apache/camel/component/atmosphere/websocket/WebsocketEndpoint.java
Croway: Hi @essobedo thanks a lot for your review, you were right, useStreaming was not the issue, after some time debugging it, I noticed that Atmosphere builds an in-memory map for each consumer but, without this information https://github.com/apache/camel/pull/9210/files#diff-c20b148a113e2656ed364867af72660ee3204358f5ab036e1f4ac3c810f5c88fR62, the path, the logic in Atmosphere is not enough, and messages are redirected to the first entry of the map.

essobedo: Well done! Great news 🎆

components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsRequestReplyTempQueueMultipleConsumersTest.java
```diff
@@ -45,7 +45,7 @@
 public class JmsRequestReplyTempQueueMultipleConsumersTest extends CamelTestSupport {
 
     @RegisterExtension
-    public ArtemisService service = ArtemisServiceFactory.createPersistentVMService();
+    public ArtemisService service = ArtemisServiceFactory.createVMService();
```
essobedo, line 48: Is-this change required?

Croway: The test behave the same, the persistence part is not used anyway, the VMService is a little bit faster, but I can restore PersistentVM

orpiske: I think using the non-persistent service is fine on this case. I don't think the broker needs to persist messages in this scenario.

essobedo: The question is more: is it an integration or a unit test? If it is a unit test, this change makes sense otherwise in the case of an integration test, it is better to keep the persistence even if it is faster.

orpiske: This is a unit test. It is named as so (our integration tests are named as `*IT.java`).

orpiske (approved):

essobedo (approved):

PR #9219, opened 2023-01-27
CAMEL-18917 camel-as2: add validating signature support to server

McIntozh:
# Description

A first try of solving CAMEL-18917 "Signature is not validated".
Added a new configuration option to supply certificates that are used for validation. If no certificates are supplied (the default configuration), signatures are not validated. This way the previous behavior where signatures are not validated, is not affected if nothing is explicitly configured.

Had to do quite some code changes and fix certain unit tests.
I haven't implemented validation for MDNs on client side yet, because there are still other open issues regarding the MDN that have to be solved first.

# Target

# Tracking


davsclaus:
Are there use-cases where you would have more than 1 certificate in that chain? And if so must it be an array type?

McIntozh:
I would assume, that when you set up a server, you would have one endpoint that receives messages from multiple parties. As every party has their own KeyPair for signing, the server would need the public certificates from every parity to validate every possible parties signature.
Perhaps my assumption/use case is wrong? Is this module designed to set up one endpoint per parity? Then we don't need an array.

Also the other certificate parameters `signingCertificateChain` and `encryptingCertificateChain` are also arrays (whereas I don't exactly know why we need more than one certificate for encrypting. `EncryptingUtils` seems to be the only place the parameter is used and it just takes the first entry from the array).

davsclaus:
Thanks for the explanation, that makes sense thanks

davsclaus (approved):

davsclaus:
Thanks I backported this for 3.21 as well. You are surely welcome to keep improving camel-as2

hiteshkaushik28:
Hi @McIntozh / @davsclaus , please help me understand if its possible to set up one endpoint per party? I have this usecase where each party has a separate keypair, and the server should be able to use the corresponding keypair and validatingSignatureChain?

PR #9233, opened 2023-01-27
CAMEL-18979: Add delimiter and prefix to ListObjectsRequest

alexandredavi:
# Description

Add delimiter and prefix to AWS S3 list objects operation

# Target

# Tracking


components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Producer.java
```diff
@@ -562,7 +562,13 @@ private void listObjects(S3Client s3Client, Exchange exchange) throws InvalidPay
                 message.setBody(objectList.contents());
             }
         } else {
-            ListObjectsResponse objectList = s3Client.listObjects(ListObjectsRequest.builder().bucket(bucketName).build());
+            final ListObjectsRequest listObjectsRequest = ListObjectsRequest
+                    .builder()
+                    .bucket(bucketName)
+                    .delimiter(getConfiguration().getDelimiter())
```
oscerd, line 568: I would add a guard to check if the delimiter is set on the configuration, same for the prefix, otherwise setting them on the request doesn't make sense. Also it would be a good idea to have a way to provide this parameter as headers, for being a bit more dynamic.

You need also to modify https://github.com/apache/camel/blob/main/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Configuration.java#L45-L48
 
and remove the label consumer, since now they are common.

oscerd (changes requested):
This needs to go also on main

components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Producer.java
alexandredavi: I did not checked if the delimiter and prefix are setted on the configuration because it already is null on the ListObjectsRequest builder and it will stay as null if is null on the configuration, but i can add the check to avoid an unecessary operation on the builder.

About the need of go also to main, should i do a cherry-pick?

oscerd: > I did not checked if the delimiter and prefix are setted on the configuration because it already is null on the ListObjectsRequest builder and it will stay as null if is null on the configuration, but i can add the check to avoid an unecessary operation on the builder.
> 

Effectively we are not checking even on the consumer side, so I would maintain the same behavior. So ignore my first review on that part.

> About the need of go also to main, should i do a cherry-pick?

Yes, it should be straightforward I think.

oscerd (approved):

davsclaus (approved):

zhfeng (approved):

PR #9270, opened 2023-01-31
(chores) camel-core: tests cleanups

orpiske:
- remove unnecessary override
- simplify list creation
- use try-with-resources


oscerd (approved):

essobedo (approved):

orpiske:
Also, @essobedo and everyone else: please do continue reviewing. Just keep in mind that sometimes I may need to merge ealier (to ensure it gets into the next CI build - where we have most of the instability) and save fixes for later. 

On these tests alone there are more than 11.000 items to review. So, simple ones or more likely to cause instability goes first, and more elaborate and not-so-easy to automate go later, as I need to balance getting the cleanup done with my availability to backtrace and bisect the changes in case of problems in the future.

PR #9271, opened 2023-01-31
(chores) camel-core: 3rd round of test cleanups

orpiske:
Lots of assertion cleanups on this one, sprinkled with some minor rework in array handling and a few more static inner classes.

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
@@ -43,7 +43,12 @@ import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class ConverterTest extends TestSupport {
 
@@ -58,7 +63,7 @@ public class ConverterTest extends TestSupport {
     }
 
     @Test
-    public void testIntegerPropertyEditorConversion() throws Exception {
+    public void testIntegerPropertyEditorConversion() {
         Integer value = converter.convertTo(Integer.class, "1000");
         assertNotNull(value);
         assertEquals(Integer.valueOf(1000), (Object) value, "Converted to Integer");
@@ -68,7 +73,7 @@ public class ConverterTest extends TestSupport {
     }
 
     @Test
-    public void testConvertStringToAndFromByteArray() throws Exception {
+    public void testConvertStringToAndFromByteArray() {
         byte[] array = converter.convertTo(byte[].class, "foo");
         assertNotNull(array);
 
@@ -79,7 +84,7 @@ public class ConverterTest extends TestSupport {
     }
 
     @Test
-    public void testConvertStringToAndFromCharArray() throws Exception {
+    public void testConvertStringToAndFromCharArray() {
         char[] array = converter.convertTo(char[].class, "foo");
         assertNotNull(array);
 
@@ -90,7 +95,7 @@ public class ConverterTest extends TestSupport {
     }
 
     @Test
-    public void testConvertStringAndStreams() throws Exception {
+    public void testConvertStringAndStreams() {
         InputStream inputStream = converter.convertTo(InputStream.class, "bar");
         assertNotNull(inputStream);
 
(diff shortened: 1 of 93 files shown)


oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
```diff
@@ -168,38 +174,38 @@ public void testFileToString() throws Exception {
     }
 
     @Test
-    public void testPrimitiveBooleanConversion() throws Exception {
-        boolean value = converter.convertTo(boolean.class, null);
+    public void testPrimitiveBooleanConversion() {
+        boolean value = assertDoesNotThrow(() -> converter.convertTo(boolean.class, null));
```
essobedo, line 178: Brings more complexity to the test and reduces its readability for no real gain IMHO

core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
```diff
@@ -168,38 +174,38 @@ public void testFileToString() throws Exception {
...
         int value = converter.convertTo(int.class, 4);
         assertEquals(4, value, "value");
     }
 
     @Test
-    public void testPrimitiveIntPropertySetter() throws Exception {
+    public void testPrimitiveIntPropertySetter() {
         MyBean bean = new MyBean();
-        IntrospectionSupport.setProperty(converter, bean, "foo", "4");
+        assertDoesNotThrow(() -> IntrospectionSupport.setProperty(converter, bean, "foo", "4"));
```
essobedo, line 191: ditto

core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
```diff
@@ -223,46 +229,41 @@ public void testInstanceMethodConversionWithExchange() throws Exception {
...
+        Exception ex = assertThrows(NoTypeConversionAvailableException.class,
+                () -> converter.mandatoryConvertTo(InputStream.class, exchange),
+                "Expected to get a NoTypeConversionAvailableException here");
     }
 
     @Test
-    public void testStringToChar() throws Exception {
-        char ch = converter.convertTo(char.class, "A");
+    public void testStringToChar() {
+        char ch = assertDoesNotThrow(() -> converter.convertTo(char.class, "A"));
```
essobedo, line 241: ditto

core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
```diff
@@ -223,46 +229,41 @@ public void testInstanceMethodConversionWithExchange() throws Exception {
...
         assertEquals('A', (int) ch);
 
-        ch = converter.convertTo(char.class, " ");
+        ch = assertDoesNotThrow(() -> converter.convertTo(char.class, " "));
```
essobedo, line 244: ditto

essobedo (commented):
LGTM only small remarks

core/camel-core/src/test/java/org/apache/camel/converter/ConverterTest.java
orpiske: It is an expectation for this test that when receiving a primitive type, the conversion succeeds. By allowing the exception to leak, it could potentially be misinterpreted as some generic fail. Let's add a fail message here, and be clear about out expectations (same for the other ones).

orpiske: One more thing in favor of the `assertDoesNotThrow` with the fail message: we get more details on the CI job in case of failure. That prevents us from having to hunt down the logs on the CI.

davsclaus (approved):

orpiske:
> ### Components tested:
> 2 	2 	1 	2

The failure is unrelated. We should be good to merge.

orpiske:
Hey @essobedo are you fine with my comments, so I can merge this?

essobedo:
> Hey @essobedo are you fine with my comments, so I can merge this?

TBH, I'm not really convinced because you replaced a unit test that could be inlined to make a one-line test with 4 lines of code just to avoid throwing an Exception, for me, it is a bit over-engineered. The problem, I see here, is the fact as you duplicate the pattern, to remain homogenous, future similar tests are supposed to apply the same logic which is not really easy to read.

But go ahead, if it is fine for you and others, you can merge it as it is my point of view nothing more 😃

orpiske:
> > Hey @essobedo are you fine with my comments, so I can merge this?
> 
> TBH, I'm not really convinced because you replaced a unit test that could be inlined to make a one-line test with 4 lines of code just to avoid throwing an Exception, for me, it is a bit over-engineered. The problem, I see here, is the fact that, as you duplicate the pattern, to remain homogenous, future similar tests are supposed to apply the same logic which is not really easy to read (easy to write = easy to read = easy to maintain).

Yeah. TBH, I kinda agree with you that the way we assert for throws / does not throw can be a bit verbose and convoluted. It's one feature I believe was cleaner with JUnit 4. 

> 
> But go ahead, if it is fine for you and others, you can merge it as it is only my point of view nothing more 😃

Thanks.

PR #9272, opened 2023-01-31
Remove deprecated XmlRoutesDefinitionLoader

gnodet:
# Description

# Target

# Tracking


oscerd (approved):

orpiske (approved):

davsclaus:
Is this being replaced by https://github.com/apache/camel/pull/9295 and if so then we can close this PR

gnodet:
> Is this being replaced by #9295 and if so then we can close this PR

It's not. I've splitted them correctly.

davsclaus (approved):

core/camel-support/src/main/java/org/apache/camel/support/CachedResource.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        if (data == null) {
+            try (InputStream is = IOHelper.buffered(delegate.getInputStream())) {
+                ByteArrayOutputStream bos = new ByteArrayOutputStream();
+                IOHelper.copyAndCloseInput(is, bos);
+                data = bos.toByteArray();
+            }
```
essobedo, line 65: Any reason not to simply call `readAllBytes()` on the `InputStream`?

dsl/camel-xml-io-dsl/src/main/java/org/apache/camel/dsl/xml/io/XmlRoutesBuilderLoader.java
```diff
@@ -41,45 +44,36 @@ public XmlRoutesBuilderLoader() {
...
-                new ModelParser(resource)
-                        .parseTemplatedRoutesDefinition()
-                        .ifPresent(this::setTemplatedRouteCollection);
-                new ModelParser(resource)
-                        .parseRestsDefinition()
-                        .ifPresent(this::setRestCollection);
-                new ModelParser(resource)
-                        .parseRoutesDefinition()
-                        .ifPresent(this::addRoutes);
+                for (String ns : List.of("", NAMESPACE)) {
```
essobedo, line 54: The list could be a static final field to avoid creating it at each call

dsl/camel-xml-jaxb-dsl-test/definition/src/test/java/org/apache/camel/dsl/xml/jaxb/definition/LoadRestFromXmlTest.java
```diff
@@ -54,10 +52,9 @@ public void testLoadRestsDefinitionFromXml() throws Exception {
         foo.assertIsSatisfied();
 
         // load rest from XML and add them to the existing camel context
-        InputStream is = getClass().getResourceAsStream("barRest.xml");
         ExtendedCamelContext ecc = context.adapt(ExtendedCamelContext.class);
-        RestsDefinition rests = (RestsDefinition) ecc.getXMLRoutesDefinitionLoader().loadRestsDefinition(ecc, is);
-        context.addRestDefinitions(rests.getRests(), true);
+        Resource resource = ecc.getResourceLoader().resolveResource("org/apache/camel/dsl/xml/jaxb/definition/barRest.xml");
+        ecc.getRoutesLoader().loadRoutes(resource);
```
essobedo, line 57: To avoid duplicating the same code, we could have a helper class that proposes a method allowing us to load the routes from a file of a given name located in a given package. This helper class could be added to the test classes of camel-core (the test classes of camel-core are packaged as a jar) to be reusable in different modules.

essobedo (commented):
Nice 👍 just small remarks/questions

dsl/camel-xml-jaxb-dsl-test/definition/src/test/java/org/apache/camel/dsl/xml/jaxb/definition/LoadRestFromXmlTest.java
gnodet: I think this should be done in a separate commit if there's really a need to make a reusable method, as it would span a bunch of additional files.  
But if that's really something that should be reused, maybe a better place would be on the `ExtendedCamelContext` directly, as it would benefit end users too.

essobedo (approved):

PR #9295, opened 2023-02-03
Generated xml writer

gnodet:
Generate a lightweight xml writer.  There are a few changes in the model to fix a few discrepancies and avoid too much hacks.

# Description

# Target

# Tracking


components/camel-openapi-java/src/main/java/org/apache/camel/openapi/jmx/JmxRestDefinitionsResolver.java
```diff
@@ -16,11 +16,8 @@
  */
 package org.apache.camel.openapi.jmx;
 
-import java.io.InputStream;
 import java.lang.management.ManagementFactory;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
oscerd, line 20: This will fail with sourcecheck profile.

oscerd (changes requested):
Just one minor finding.

davsclaus (approved):

core/camel-xml-io/src/main/java/org/apache/camel/xml/io/PrettyPrintXMLWriter.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Implementation of XMLWriter which emits nicely formatted documents.
+ */
+public class PrettyPrintXMLWriter {
+    /**
+     * Line separator ("\n" on UNIX)
+     */
+    protected static final String LS = System.getProperty("line.separator");
```
orpiske, line 32: Minor thing, but `System.lineSeparator()` is a bit clearer

core/camel-xml-io/src/main/java/org/apache/camel/xml/io/PrettyPrintXMLWriter.java
```diff
@@ -0,0 +1,369 @@
...
+
+    private final Writer writer;
+    private final String lineIndenter;
+    private final String lineSeparator;
+    private final String encoding;
+    private final String docType;
+    private final LinkedList<String> elementStack = new LinkedList<>();
```
orpiske, line 39: Maybe it would be better to use a List here, so to give more flexibility.

core/camel-xml-io/src/test/java/org/apache/camel/xml/out/ModelWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.xml.out;
+
+import java.io.*;
```
orpiske, line 19: This may fail the checkstyle check

core/camel-xml-io/src/test/java/org/apache/camel/xml/out/ModelWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+import static org.junit.jupiter.api.AssertionFailureBuilder.assertionFailure;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ModelWriterTest {
+
+    public static final String NAMESPACE = "http://camel.apache.org/schema/spring";
+
+    @ParameterizedTest
+    @MethodSource("routes")
```
orpiske, line 49: Given that this code is doing something complex, I think it would be good to annotate it with `@Display` describing the expectations (i.e.: `@Display("Tests that given an XML file containing X, will generate/process/something Y")`)

core/camel-xml-io/src/test/java/org/apache/camel/xml/out/ModelWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+            RoutesDefinition expected = new ModelParser(is, NAMESPACE).parseRoutesDefinition().get();
+            StringWriter sw = new StringWriter();
+            new ModelWriter(sw, ns).writeRoutesDefinition(expected);
+            RoutesDefinition actual = new ModelParser(new StringReader(sw.toString()), ns).parseRoutesDefinition().get();
+            assertDeepEquals(expected, actual, sw.toString());
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("rests")
```
orpiske, line 61: Same suggestion about `@Display` here and on the other tests.

core/camel-xml-io/src/test/java/org/apache/camel/xml/out/ModelWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+        String name = field.getName();
+        try {
+            Method method = field.getDeclaringClass().getDeclaredMethod(
+                    "get" + name.substring(0, 1).toUpperCase() + name.substring(1));
+            if (method.getAnnotation(XmlTransient.class) != null) {
+                return true;
+            }
+        } catch (Throwable t) {
+            // ignore
+        }
```
orpiske, line 206: We should not catch `Throwable`. This would also catch JVM errors (such as OOME) and could hide some serious problems. In general, catching just Exception should be fine.

core/camel-xml-io/src/test/java/org/apache/camel/xml/out/ModelWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+        try {
+            Method method = field.getDeclaringClass().getDeclaredMethod(
+                    "set" + name.substring(0, 1).toUpperCase() + name.substring(1),
+                    field.getType());
+            if (method.getAnnotation(XmlTransient.class) != null) {
+                return true;
+            }
+        } catch (Throwable t) {
```
orpiske, line 214: Same note about catching `Throwable`

orpiske (commented):
Hi, I added some suggestions.

PR #9332, opened 2023-02-10
CAMEL-19014 - Fixing issue about concurrence on the SimpleLanguage

rhuan080:

# Description

Fixing the issue [CAMEL-19014](https://issues.apache.org/jira/browse/CAMEL-19014)

# Target

# Tracking


essobedo (changes requested):
Thx for the PR, but first we need to agree that the use case is worth being fixed, I personally tend to say it doesn't as it will add some overhead to fix a very uncommon use case.

If we decide to fix it, this fix has to be reviewed because it will have a dramatic impact on the performance, especially in the `evaluate` method which is frequently called, and itself calls `evaluate` on children's expressions. This fix assumes that evaluating an expression is very fast but in practice, we don't know, indeed if we call a method that is very slow for some reason (IO or CPU), it will block parallel evaluations just to fix once again a non-standard use case.

rhuan080:
Hi @essobedo, thank you for reviewing the PR.

I have changed the synchronized block to lock. Now the evaluate method tries to get the ReadLock that does not block the thread for multiple ReadLock. The `init` method is using the WriteLock, thus the `evaluate` method is blocked only in case of a WriteLock on `init` method. It avoids the `evaluate` method to be blocked unnecessarily. 

I think we should develop this method as thread-safe, as it is cached and can be used in a multithreaded task (like parallel or some internal multithread task). Thus, to avoid an issue now or in future, I think it should be immutable or thread-safe

essobedo:
> I have changed the synchronized block to lock. Now the evaluate method tries to get the ReadLock that does not block the thread for multiple ReadLock. The `init` method is using the WriteLock, thus the `evaluate` method is blocked only in case of a WriteLock on `init` method. It avoids the `evaluate` method to be blocked unnecessarily.
> 

It if needs to be fixed, it is not the right way to fix it, you don't even need locks here. If you really want to fix it, you will have to rewrite the code of the `concatExpression`, to have 2 classes instead of one (for constants only and not only constants). The problem is with the case of the not only constants, for it, you simply need to set the content of `col` once for all while initializing the instance instead of doing it in the init method. Moreover, you would need to add a unit test to prove that the bug is really fixed, [see TDD](https://en.wikipedia.org/wiki/Test-driven_development).

> I think we should develop this method as thread-safe, as it is cached and can be used in a multithreaded task (like parallel or some internal multithread task). Thus, to avoid an issue now or in future, I think it should be immutable or thread-safe

This decision doesn't really belong to us, especially if we disagree, I guess that someone like @davsclaus or @oscerd could eventually decide. But, here, according to your stack trace, if we get this problem it is because the expressions are misused such that the threads spend their time to re-initialize the same expression over and over again instead of only once at startup as the framework does, I don't believe that for a production-ready code, we would really want this behavior, so it tends to prove that it is more a misuse than a bug but it is only my point of view, let's see what others think.

essobedo:
> The best approach for me is to use an immutable object, however, if it is not possible I think we should consider using a lock or synchronized block.
> 

Actually, immutability is the idea behind what I proposed

rhuan080:
Understood and I agree. However, I think we don't need 2 classes, but just rewrite the concatExpression to have 1 class immutable. What do you think about it?

In the case of 2 classes, we will need to define when using immutable e when using mutable. For me, it is an unneeded complex.  
> > The best approach for me is to use an immutable object, however, if it is not possible I think we should consider using a lock or synchronized block.
> 
> Actually, immutability is the idea behind what I proposed

rhuan080:
Hi @essobedo reviewing the class, I agree with you. This class already knows if the expression is constant or not. Thus, for me you are right! Thanks a lot.

I'll write the unit test and use the approach you said, and we can discuss it again.

essobedo:
Well, we don't necessarily need 2 classes, but the code will be cleaner and thus easier to maintain, it will then follow the Single-responsibility principle.

Regarding, the unit test, you can simply adapt what the OP provided as it helps to reproduce the bug.

oscerd:
From what I read this is really a corner case. I'm not sure we should fix that.

essobedo:
@rhuan080 +1 I agree, if we can end up with a solution based on immutable classes, it is a good tradeoff

oscerd:
Using immutable classes sounds good and I'm +1 on that

rhuan080:
Hi,

I added the unit test. Now, I'll rewrite the concatExpression to be immutable. @essobedo and @oscerd If you have questions or improvement about the unit test, please, tell me :-)

essobedo:
> Hi,
> 
> I added the unit test. Now, I'll rewrite the concatExpression to be immutable. @essobedo and @oscerd If you have questions or improvements about the unit test, please, tell me :-)

I don't understand why you don't use the test case provided by the OP?

rhuan080:
The test uses components that are not available in this module. Thus, I created the test using what is available inside the support module. I have reproduced the issue with my test.

essobedo:
> The test provided uses components that are not available in this module. Thus, I created the test using what is available inside the camel-support module. I have reproduced the issue with my test.

You can put it in this package https://github.com/apache/camel/tree/main/core/camel-core/src/test/java/org/apache/camel/language

rhuan080:
Hi @essobedo! 

Added the test according to your suggestion.

core/camel-core/src/test/java/org/apache/camel/builder/ExpressionBuilderConcurrencyTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ExpressionBuilderConcurrencyTest extends ContextTestSupport {
+
+    @Test
+    public void testConcatExpressionConcurrency() throws Exception {
+        MockEndpoint mockWithFailure = getMockEndpoint("mock:result");
+        mockWithFailure.expectedMessageCount(102);
+        mockWithFailure.assertIsSatisfied();
+        //assertMockEndpointsSatisfied();
```
essobedo, line 38: Could be removed

core/camel-core/src/test/java/org/apache/camel/builder/ExpressionBuilderConcurrencyTest.java
```diff
@@ -0,0 +1,106 @@
...
+        List<Exchange> exchanges = mockWithFailure.getExchanges();
+        exchanges.stream()
+                .forEach(exchange -> Assertions.assertTrue(exchange.getMessage()
+                        .getHeader("#CustomHeader", String.class)
+                        .equals("This is a test a with startLabel: `Document` endLabel: `Document` and label: `ALabel`")));
```
essobedo, line 43: For clarity, it should rather be a simple `for` loop. Moreover, please use `assertEquals` instead.

core/camel-core/src/test/java/org/apache/camel/builder/ExpressionBuilderConcurrencyTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.camel.model.language.SimpleExpression;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ExpressionBuilderConcurrencyTest extends ContextTestSupport {
+
+    @Test
+    public void testConcatExpressionConcurrency() throws Exception {
+        MockEndpoint mockWithFailure = getMockEndpoint("mock:result");
+        mockWithFailure.expectedMessageCount(102);
```
essobedo, line 36: Let's put `100` to avoid making it sound like a magic number

core/camel-core/src/test/java/org/apache/camel/builder/ExpressionBuilderConcurrencyTest.java
```diff
@@ -0,0 +1,106 @@
...
+    protected RouteBuilder createRouteBuilder() throws Exception {
+
+        return new RouteBuilder() {
+            Map body = new HashMap() {
+                {
+                    put("label", "ALabel");
+                    put("startLabel", "Document");
+                    put("endLabel", "Document");
+                }
+            };
```
essobedo, line 56: Let's use, `Map.of("label", "ALabel", "startLabel", "Document", "endLabel", "Document")` instead

essobedo (commented):
Many thx for the uni test 🙏 Some small remarks

rhuan080:
Hi @essobedo, I'll fix the unit tests. Thanks a lot! 

I have committed the refractory of the concatExpression and created one method to build the optimized expression (immutable) and the unoptimized expression. Let me know if you have any questions about the solution.

core/camel-core/src/test/java/org/apache/camel/builder/ExpressionBuilderConcurrencyTest.java
rhuan080: Done!

core/camel-support/src/main/java/org/apache/camel/support/builder/ExpressionBuilder.java
```diff
@@ -1577,74 +1578,122 @@ public String toString() {
...
+     * @param context the Camel context
+     * @param expressions the expression to be concatenated dynamically
+     * @param description the text description of the expression
+     * @return an expression which when evaluated will return the concatenated values
+     */
+    private static Expression concatExpressionOptimized(final CamelContext context, final Collection<Expression> expressions, final String description) {
+        Collection<Object> preprocessedExpression = new ArrayList<>(expressions.size());
+        for (Expression expression : expressions) {
+            if (expression instanceof ConstantExpressionAdapter) {
+                expression.init(context);
```
essobedo, line 1664: I did not realize that for some `ConstantExpressionAdapter` we had to initialize it first to get its value, I don't know if it is acceptable to call `init` so early? @oscerd WDYT?

If it is not acceptable, the other way could be to use a volatile boolean field to know if the expression has been initialized or not and use the Double-checked locking approach to ensure that it is initialized once. Even if it adds a locking mechanism it is only during the initialization phase which happens only once per expression when properly used so the overhead is limited.

core/camel-support/src/main/java/org/apache/camel/support/builder/ExpressionBuilder.java
```diff
@@ -1577,74 +1578,122 @@ public String toString() {
...
                     }
                 }
                 return buffer.toString();
             }
 
             @Override
             public void init(CamelContext context) {
-                boolean constant = false;
                 for (Expression expression : expressions) {
                     expression.init(context);
```
essobedo, line 1696: Assuming that it is acceptable to call `init` when creating the expression, we could avoid calling init on expressions of type `ConstantExpressionAdapter` as they are already initialized

essobedo (commented):
That's more or less what I have in mind, but it makes me realize that it is more complex than I thought, see my comments for details

core/camel-support/src/main/java/org/apache/camel/support/builder/ExpressionBuilder.java
rhuan080: Hi @essobedo, Looking at the `ConstantExpression`, I think that calling the `init` so early is not a problem. Waiting for the @oscerd to confirm it for us. 

About the lock, the problem is the `evaluate` method being called concurrently with the `init`calling.  Thus, I think the lock should be put at the `evaluate` method as well. However, we came back to the penalty we said before.

essobedo: > About the lock, the problem is the `evaluate` method being called concurrently with the `init`calling. Thus, I think the lock should be put at the `evaluate` method as well. However, we came back to the penalty we said before.

To be thread-safe, we have to consider the state changes and the visibility of those changes. Here only the `init` method affects the state and the `init` method is always called before `evaluate` so applying a double-check locking approach in the `init` method is good enough as it ensures that the state is changed only once at the very first call and that all threads thanks to a volatile read or synchronized block will always have the right state before calling `evaluate`. I hope it is clear enough.

davsclaus:
be careful to not overcomplicate this - the init of services in Camel are used during bootstrap and are called by Camel in sequence (no locking needed).

We should not do special weird code for one corner case with simple language that been in Camel since 1.x.

If you can come up with a better parser (then yeah) but do not sacrifice the init service lifecycle of Camel.

core/camel-support/src/main/java/org/apache/camel/support/builder/ExpressionBuilder.java
```diff
@@ -1593,59 +1594,109 @@ public static Expression concatExpression(final Collection<Expression> expressio
...
-                    // evaluate them but use their constant value as-is directly
-                    // this can be common with the simple language where you use it for templating
-                    // by mixing string text and simple functions together (or via the log EIP)
-                    col = new ArrayList<>(expressions.size());
+
+                if(col == null) {
+                    Collection<Object> preprocessedExpression = new ArrayList<>(expressions.size());
                     for (Expression expression : expressions) {
                         if (expression instanceof ConstantExpressionAdapter) {
+                            expression.init(context);
```
essobedo, line 1690: it's already initialized so it is not needed

rhuan080: My fault. I'm sorry. Fixed.

oscerd (approved):

essobedo (approved):
LGTM 👍 Even if the `ExpressionAdapter` of `concatExpressionOptimized` is not thread-safe, I believe that it is a good tradeoff compared to not fixing at all

rhuan080:
I'm sorry for closing this issue. It was just a click error, and the button to close does not have a double check 🤦‍♂️

Thank you @essobedo and @oscerd ! 

As the @davsclaus that the init is just called at the Camel bootstrap and the init method is not called inside a parallel process, I think just preventing the collection from being re-initialized in case of calling the init again is sufficient. I will update the issue in Jira with these information soon..

davsclaus:
what is the status of this? And I think its safer to do this in 3.21 release only, and not 3.20.x to keep it as-is.
This is a corner case and simple language is used a lot in Camel routes.

essobedo:
@rhuan080 Would you mind creating a new PR for the branch `camel-3.x` for Camel 3.21?

davsclaus:
This PR causes NPE errors on main, so hold until we fixed this or we need to revert and drop this PRs

davsclaus (changes requested):
see prev comments

davsclaus:
okay the optimized can be even better, if there are only ConstantExpressionAdapter then you can build the result string in init and return its constant value in evaluate

davsclaus:
Okay this commit fixes the NPEs
https://github.com/apache/camel/commit/d6ca5fca99d528b5394113433c5f4e9a1b0b5296

davsclaus:
And this commit optimizes when there are only constant values
https://github.com/apache/camel/commit/f2a3cc20d07271cc2bfba26cb3ecdcf4c119866d

davsclaus:
So we need a PR for camel-3.x branch and leave 3.20.x as-is.
And in 3.x we need this PR and the 2 commits I linked.

rhuan080:
Hi @davsclaus, ok! I'll send a PR to camel-3.x. I'll update the 3.x to add the commits.

rhuan080:
PR 3.x: https://github.com/apache/camel/pull/9531

PR #9339, opened 2023-02-12
avoid java assert keyword as disabled per default

bvahdat:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
@@ -26,6 +26,7 @@ import org.apache.camel.resume.OffsetKey;
 import org.apache.camel.resume.ResumeAdapter;
 import org.apache.camel.resume.cache.ResumeCache;
 import org.apache.camel.spi.annotations.JdkService;
+import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.kinesis.model.GetShardIteratorRequest;
@@ -46,8 +47,8 @@ public class KinesisDefaultResumeAdapter implements KinesisResumeAdapter, Cachea
 
     @Override
     public void resume() {
-        assert streamName != null;
-        assert resumable != null;
+        ObjectHelper.notNull(streamName, "streamName");
+        ObjectHelper.notNull(resumable, "resumable");
 
         KinesisOffset offset = cache.get(streamName, KinesisOffset.class);
 
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
@@ -26,7 +26,6 @@ import org.apache.camel.resume.OffsetKey;
 import org.apache.camel.resume.ResumeAdapter;
 import org.apache.camel.resume.cache.ResumeCache;
 import org.apache.camel.spi.annotations.JdkService;
-import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.kinesis.model.GetShardIteratorRequest;
@@ -47,8 +46,8 @@ public class KinesisDefaultResumeAdapter implements KinesisResumeAdapter, Cachea
 
     @Override
     public void resume() {
-        ObjectHelper.notNull(streamName, "streamName");
-        ObjectHelper.notNull(resumable, "resumable");
+        assert streamName != null;
+        assert resumable != null;
 
         KinesisOffset offset = cache.get(streamName, KinesisOffset.class);
 
--- a/components/camel-caffeine/src/main/java/org/apache/camel/component/caffeine/resume/CaffeineCache.java
+++ b/components/camel-caffeine/src/main/java/org/apache/camel/component/caffeine/resume/CaffeineCache.java
@@ -25,6 +25,7 @@ import java.util.function.Function;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import org.apache.camel.resume.cache.ResumeCache;
+import org.apache.camel.util.ObjectHelper;
 
 /**
  * This is a simple cache implementation that uses Caffeine to store the resume offsets
@@ -57,7 +58,7 @@ public class CaffeineCache<K> implements ResumeCache<K> {
 
     @Override
     public boolean contains(K key, Object entry) {
-        assert key != null;
+        ObjectHelper.notNull(key, "key");
         Object cachedEntry = cache.getIfPresent(key);
 
         return entry.equals(cachedEntry);
--- a/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
@@ -21,6 +21,7 @@ import java.time.Duration;
 import java.util.Properties;
 
 import org.apache.camel.resume.ResumeStrategyConfiguration;
+import org.apache.camel.util.ObjectHelper;
 
 /**
  * A configuration suitable for using with the {@link KafkaResumeStrategy} and any of its implementations
@@ -37,9 +38,7 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setProducerProperties(Properties producerProperties) {
-        assert producerProperties != null;
-
-        this.producerProperties = producerProperties;
+        this.producerProperties = ObjectHelper.notNull(producerProperties, "producerProperties");
     }
 
     public Properties getConsumerProperties() {
@@ -47,9 +46,7 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setConsumerProperties(Properties consumerProperties) {
-        assert consumerProperties != null;
-
-        this.consumerProperties = consumerProperties;
+        this.consumerProperties = ObjectHelper.notNull(consumerProperties, "consumerProperties");
     }
 
     public String getTopic() {
@@ -57,9 +54,7 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setTopic(String topic) {
-        assert topic != null;
-
-        this.topic = topic;
+        this.topic = ObjectHelper.notNull(topic, "topic");
     }
 
     public Duration getMaxInitializationDuration() {
--- a/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
@@ -21,7 +21,6 @@ import java.time.Duration;
 import java.util.Properties;
 
 import org.apache.camel.resume.ResumeStrategyConfiguration;
-import org.apache.camel.util.ObjectHelper;
 
 /**
  * A configuration suitable for using with the {@link KafkaResumeStrategy} and any of its implementations
@@ -38,7 +37,9 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setProducerProperties(Properties producerProperties) {
-        this.producerProperties = ObjectHelper.notNull(producerProperties, "producerProperties");
+        assert producerProperties != null;
+
+        this.producerProperties = producerProperties;
     }
 
     public Properties getConsumerProperties() {
@@ -46,7 +47,9 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setConsumerProperties(Properties consumerProperties) {
-        this.consumerProperties = ObjectHelper.notNull(consumerProperties, "consumerProperties");
+        assert consumerProperties != null;
+
+        this.consumerProperties = consumerProperties;
     }
 
     public String getTopic() {
@@ -54,7 +57,9 @@ public class KafkaResumeStrategyConfiguration extends ResumeStrategyConfiguratio
     }
 
     void setTopic(String topic) {
-        this.topic = ObjectHelper.notNull(topic, "topic");
+        assert topic != null;
+
+        this.topic = topic;
     }
 
     public Duration getMaxInitializationDuration() {
--- a/components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
+++ b/components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.apache.camel.RuntimeCamelException;
+import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -239,7 +240,7 @@ public final class LogWriter implements AutoCloseable {
 
          Trying to update a persisted entry here is not acceptable
          */
-        assert layerInfo != null;
+        ObjectHelper.notNull(layerInfo, "layerInfo");
 
         final LogEntry logEntry = transactionLog.update(layerInfo, state);
 
--- a/components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
+++ b/components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.apache.camel.RuntimeCamelException;
-import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -240,7 +239,7 @@ public final class LogWriter implements AutoCloseable {
 
          Trying to update a persisted entry here is not acceptable
          */
-        ObjectHelper.notNull(layerInfo, "layerInfo");
+        assert layerInfo != null;
 
         final LogEntry logEntry = transactionLog.update(layerInfo, state);
 
--- a/core/camel-support/src/main/java/org/apache/camel/support/resume/AdapterHelper.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/resume/AdapterHelper.java
@@ -29,7 +29,6 @@ import org.apache.camel.resume.ResumeStrategy;
 import org.apache.camel.resume.ResumeStrategyConfiguration;
 import org.apache.camel.resume.cache.ResumeCache;
 import org.apache.camel.spi.FactoryFinder;
-import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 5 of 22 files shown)


oscerd (approved):

bvahdat:
The 2 failed tests by the `camel-management` module are not relevant to this PR and fail the same way on the main branch:

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.369 s - in org.apache.camel.management.ManagedThrottlerTest
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR] org.apache.camel.management.ManagedFromRestGetTest.testFromRestModel
[ERROR]   Run 1: ManagedFromRestGetTest.testFromRestModel:67 expected: <true> but was: <false>
[ERROR]   Run 2: ManagedFromRestGetTest.testFromRestModel:67 expected: <true> but was: <false>
[ERROR]   Run 3: ManagedFromRestGetTest.testFromRestModel:67 expected: <true> but was: <false>
[INFO] 
[ERROR] org.apache.camel.management.ManagedFromRestPlaceholderTest.testFromRestModelPlaceholder
[ERROR]   Run 1: ManagedFromRestPlaceholderTest.testFromRestModelPlaceholder:68 expected: <true> but was: <false>
[ERROR]   Run 2: ManagedFromRestPlaceholderTest.testFromRestModelPlaceholder:68 expected: <true> but was: <false>
[ERROR]   Run 3: ManagedFromRestPlaceholderTest.testFromRestModelPlaceholder:68 expected: <true> but was: <false>
```

davsclaus (approved):

orpiske (changes requested):
I am not sure this is fully correct.

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
```diff
@@ -46,8 +47,8 @@ public void setRequestBuilder(GetShardIteratorRequest.Builder resumable) {
 
     @Override
     public void resume() {
-        assert streamName != null;
-        assert resumable != null;
```
orpiske, line 50: Using Java's assert is the correct away to assert for nullity in non-public code as is the case here (obs.: the code is internally public, but not public in the to Camel users).

components/camel-caffeine/src/main/java/org/apache/camel/component/caffeine/resume/CaffeineCache.java
```diff
@@ -57,7 +58,7 @@ public CaffeineCache(Cache<K, Object> cache, long cacheSize) {
 
     @Override
     public boolean contains(K key, Object entry) {
-        assert key != null;
+        ObjectHelper.notNull(key, "key");
```
orpiske, line 61: Assert is the correct check here, as this is set internally by the resume code and it should not be null. The assert only adds a layer of check when running in tests as surefire/failsafe will enable native Java assertions by default in those cases.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
```diff
@@ -37,29 +38,23 @@ public Properties getProducerProperties() {
     }
 
     void setProducerProperties(Properties producerProperties) {
-        assert producerProperties != null;
-
-        this.producerProperties = producerProperties;
+        this.producerProperties = ObjectHelper.notNull(producerProperties, "producerProperties");
```
orpiske, line 41: Assert is the correct one: this not user-facing public code.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
```diff
@@ -37,29 +38,23 @@ public Properties getProducerProperties() {
...
     }
 
     public Properties getConsumerProperties() {
         return consumerProperties;
     }
 
     void setConsumerProperties(Properties consumerProperties) {
-        assert consumerProperties != null;
-
-        this.consumerProperties = consumerProperties;
```
orpiske, line 52: Same as above

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
```diff
@@ -37,29 +38,23 @@ public Properties getProducerProperties() {
...
     }
 
     public String getTopic() {
         return topic;
     }
 
     void setTopic(String topic) {
-        assert topic != null;
-
-        this.topic = topic;
```
orpiske, line 62: Same as above.

components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
```diff
@@ -239,7 +240,7 @@ public void updateState(EntryInfo.CachedEntryInfo entryInfo, LogEntry.EntryState
 
          Trying to update a persisted entry here is not acceptable
          */
-        assert layerInfo != null;
+        ObjectHelper.notNull(layerInfo, "layerInfo");
```
orpiske, line 243: This is also not correct and also impacts the performance of the log writer as the assert is native and would not be enabled in production code (whereas the ObjectHelper will incur the cost of `notNull` check on a critical section of this code).

core/camel-support/src/main/java/org/apache/camel/support/resume/AdapterHelper.java
```diff
@@ -41,9 +42,9 @@ private AdapterHelper() {
     }
 
     public static ResumeAdapter eval(CamelContext context, ResumeAware resumeAware, ResumeStrategy resumeStrategy) {
-        assert context != null;
-        assert resumeAware != null;
-        assert resumeStrategy != null;
+        ObjectHelper.notNull(context, "context");
+        ObjectHelper.notNull(resumeAware, "resumeAware");
+        ObjectHelper.notNull(resumeStrategy, "resumeStrategy");
```
orpiske, line 47: Assert is the correct one here as this is an internal interface where the nullity condition should not happen in any circumstance.

test-infra/camel-test-infra-core/src/test/java/org/apache/camel/test/infra/core/DefaultAnnotationProcessor.java
```diff
@@ -151,13 +152,12 @@ private void doInvokeFixture(Annotation annotation, Field field, Object instance
...
 
-                    assert mockEndpoint != null;
-
+                    ObjectHelper.notNull(mockEndpoint, "mockEndpoint");
                     field.set(instance, mockEndpoint);
                 } else {
                     final Endpoint endpoint = context.getEndpoint(uri);
 
-                    assert endpoint != null;
+                    ObjectHelper.notNull(endpoint, "endpoint");
```
orpiske, line 160: A bit torn on this one, as the assert is missing a message saying what's going on.

test-infra/camel-test-infra-core/src/test/java/org/apache/camel/test/infra/core/api/CamelTestSupportHelper.java
```diff
@@ -30,7 +31,7 @@ public interface CamelTestSupportHelper {
 
     default <T extends Endpoint> T getMandatoryEndpoint(String uri, Class<T> type) {
         T endpoint = getCamelContextExtension().getContext().getEndpoint(uri, type);
-        assert endpoint != null : "No endpoint found for uri: " + uri;
```
orpiske, line 33: Assert is OK here.

orpiske (changes requested):
I've left a few comments. In general `assert` is the right way to check for nullity on internal code where we know it won't usually be null.
This is particular important on performance sensitive code, as the Java's native `assert` is only enabled during tests (or when passing the `-ea` flag to the JVM).

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
bvahdat: Thanks for your feedback.

IMHO no matter if it's about a user-facing pubic or not-public logic, at the end of the day this code would run in a production environment where  typically Java assertion is not explicitly enabled. Then having these null guards using Java assertion would simply get skipped, then users would only see the **side effects** of these null values **later on** on their call stack and they / we would need to guess and find out the **root cause** (which was null values not being checked early enough). On the other hand having this check eagerly through `ObjectHelper#notNull` would not end up in such a scenarios.

components/camel-caffeine/src/main/java/org/apache/camel/component/caffeine/resume/CaffeineCache.java
bvahdat: > The assert only adds a layer of check when running in tests as surefire/failsafe will enable native Java assertions by default in those cases.

But this is not a test running under surefire/failsafe control. So the same reasoning applies here as :point_up_2:.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
bvahdat: My same reasoning as very top :point_up_2:.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
bvahdat: My same reasoning as very top :point_up_2:.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
bvahdat: My same reasoning as very top :point_up_2:.

components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
bvahdat: > impacts the performance of the log write

Not sure if a null check through `ObjectHelper.notNull`will have any reasonable performance impact, or do we maybe have any benchmark for that which proves it otherwise?

core/camel-support/src/main/java/org/apache/camel/support/resume/AdapterHelper.java
bvahdat: My same reasoning as very top 👆.

Other than that when we run Camel's own tests (where Java assertion through surefire/failsafe plugins are enabled) then it could be that all these assertions are properly fulfilled which is given through a **concrete test setup**, but what if the setup and data of a given Camel user's production environment is different than what we had by our tests which suddenly they would end up with null values here. So using `ObjectHelper.notNull` guarantees we always verify these, no matter if through Camel test **setup and configuration** or the ones from the users.

test-infra/camel-test-infra-core/src/test/java/org/apache/camel/test/infra/core/DefaultAnnotationProcessor.java
bvahdat: I think [the message here](https://github.com/apache/camel/blob/camel-4.0.0-M1/core/camel-util/src/main/java/org/apache/camel/util/ObjectHelper.java#L156) would be clear enough as:

```
endpoint must be specified
```

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
orpiske: These 2 are [invariants which is why checking for assertions is the correct approach](https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#usage-invariants). In normal circumstances they should be defined as final members of the `KinesisDefaultResumeAdapter` class. Unfortunately our code is made in a way that does not favor usage of final variables in many places (which then, would make the assert unnecessary)

components/camel-caffeine/src/main/java/org/apache/camel/component/caffeine/resume/CaffeineCache.java
orpiske: I am wrong about this one. `ObjectHelper` is OK here.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
orpiske: Same as the other one.

components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/KafkaResumeStrategyConfiguration.java
orpiske: Also not a public interface.

components/camel-wal/src/main/java/org/apache/camel/component/wal/LogWriter.java
orpiske: The impact of `ObjectHelper.nonNull` is pretty small. Almost close to none, but still a bit faster. 

```
AssertTest.benchmarkAssert   avgt   10  183.115 ± 0.844  us/op
AssertTest.benchmarkHelper   avgt   10  186.937 ± 0.779  us/op
```

Of all the items I commented, this one is the only one which is a big -1 from me. The log writer adds a small bit of latency on the Resume process due to the I/O, so every bit of performance that can be saved here, is important.

core/camel-support/src/main/java/org/apache/camel/support/resume/AdapterHelper.java
orpiske: The reasoning is the other way around: is to ensure that we catch misbehaved tests, tests that may affect this piece of code with these or any thing non-production that may be doing something nasty. 

The nullity check for most of these is already performed by the calling code. This one is a good example of that:
- [MasterConsumer.java](https://github.com/apache/camel/blob/main/components/camel-master/src/main/java/org/apache/camel/component/master/MasterConsumer.java#L143-L152)
- [DefaultRoute.java](https://github.com/apache/camel/blob/main/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultRoute.java#L646-L649)

So, the native assertion allow us to add a small layer of "protection" without adding too many runtime checks.

test-infra/camel-test-infra-core/src/test/java/org/apache/camel/test/infra/core/api/CamelTestSupportHelper.java
orpiske: I am fine with either assert or ObjectHelper here.

test-infra/camel-test-infra-core/src/test/java/org/apache/camel/test/infra/core/DefaultAnnotationProcessor.java
orpiske: Yeah, that would be ok.

orpiske:
One last reason for using the native assert instead of `ObjectHelper.notNull`: the native assert throws an [`AssertionError`](https://docs.oracle.com/javase/8/docs/api/java/lang/AssertionError.html) instead of an [`IllegalArgumentException`](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html). 

As a descendant of `Error` it properly conveys the abnormality of the condition that raised it (granted: many places try to incorrectly catch `Throwable` - but that's a completely different discussion).

davsclaus:
ObjectHelper is intended for startup/initiaizliation to check if Camel has been configured correctly. 

During high speed processing then its not intended to be in use. At that place there can be `assert` code from JDK that can benefit during trouble shooting / debugging. However we very rarely use `assert` in the camel source code. So if Otavio have some of them in camel-kafka during his performance improvement work a while back then that is fine to keep.

bvahdat:
> One last reason for using the native assert instead of `ObjectHelper.notNull`: the native assert throws an [`AssertionError`](https://docs.oracle.com/javase/8/docs/api/java/lang/AssertionError.html) instead of an [`IllegalArgumentException`](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html).
> 
> As a descendant of `Error` it properly conveys the abnormality of the condition that raised it (granted: many places try to incorrectly catch `Throwable` - but that's a completely different discussion).

Thanks for your clarifications. So should I close this PR completely or is there any changes we want to apply? If so which ones?

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisDefaultResumeAdapter.java
bvahdat: > These 2 are [invariants which is why checking for assertions is the correct approach](https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#usage-invariants). In normal circumstances they should be defined as final members of the `KinesisDefaultResumeAdapter` class. Unfortunately our code is made in a way that does not favor usage of final variables in many places (which then, would make the assert unnecessary)

In the very same page [it says](https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#putting-assertions-into-your-code):

> Do not use assertions for argument checking in public methods.
Argument checking is typically part of the published specifications (or contract) of a method, and these specifications must be obeyed whether assertions are enabled or disabled. Another problem with using assertions for argument checking is that erroneous arguments should result in an appropriate runtime exception (such as IllegalArgumentException, IndexOutOfBoundsException, or NullPointerException). An assertion failure will not throw an appropriate exception.

And by Camel we additionally make a differentiation of about if that given public method is a _user-facing public code_ or not and then we apply the assert approach or not depending on that?  As a concrete example consider the change by *this* conversation here. Even if this class is not a _user-facing public code_, according to the recommendation :point_up_2: that assert is done inside it's contract implementation method, namely `org.apache.camel.resume.ResumeAdapter#resume` which is wrong according to the recommendation :point_up_2:.

```
@Override
    public void resume() {
        assert streamName != null;
        assert resumable != null;
       ...
       ...
}
```

But I will apply the changes the way you are asking me to do, it's alright 👍

orpiske:
> > One last reason for using the native assert instead of `ObjectHelper.notNull`: the native assert throws an [`AssertionError`](https://docs.oracle.com/javase/8/docs/api/java/lang/AssertionError.html) instead of an [`IllegalArgumentException`](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html).
> > As a descendant of `Error` it properly conveys the abnormality of the condition that raised it (granted: many places try to incorrectly catch `Throwable` - but that's a completely different discussion).
> 
> Thanks for your clarifications. So should I close this PR completely or is there any changes we want to apply? If so which ones?

Thanks! I don't think you need to close the PR. I think removing the ones I mentioned would be enough and then just double checking sure we are not replacing the `assert` with `ObjectHelper` in the hot path.

bvahdat:
> > > One last reason for using the native assert instead of `ObjectHelper.notNull`: the native assert throws an [`AssertionError`](https://docs.oracle.com/javase/8/docs/api/java/lang/AssertionError.html) instead of an [`IllegalArgumentException`](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html).
> > > As a descendant of `Error` it properly conveys the abnormality of the condition that raised it (granted: many places try to incorrectly catch `Throwable` - but that's a completely different discussion).
> > 
> > 
> > Thanks for your clarifications. So should I close this PR completely or is there any changes we want to apply? If so which ones?
> 
> Thanks! I don't think you need to close the PR. I think rolling back the ones I mentioned would be enough and then just double checking to make sure we are not replacing the `assert` with `ObjectHelper` in the hot path.

@orpiske sorry for not getting back earlier today, as was much busy with other tasks.

I think in the meanwhile I have applied all your suggestions, so please verify.

orpiske (approved):

PR #9342, opened 2023-02-13
CAMEL-15105: decouple the ExtendedExchange from the Exchange

orpiske:


oscerd (approved):

orpiske:
@davsclaus @oscerd  This is the pre-work for a simpler interface to manage extensions, plugins in both the `Exchange` as well as the `CamelContext` interfaces. 

This provides only the changes for the `Exchange`, as I would like to hear your and the Community's opinion about the approach.

oscerd:
Lgtm.

zhfeng (commented):
@orpiske I just curious that what is the motivation here?

orpiske:
> @orpiske I just curious that what is the motivation here?

The core idea is to provide and API that provides an uniform and consistent way for handling extensions to the CamelContext. 

However, the same pattern also applies to the ExtendedExchange, Message and other classes that handle external properties, plugins and traits. 

This PR provides an archetype of how I propose we solve the problem:

- Decouple the ExtendedExchange from the Exchange hierarchy and move them to an external class ExtendedExchangeExtension 

Later on, the idea is to apply the same pattern to the CamelContext:

- Decouple the ExtendedContext from the CamelContext hierarchy and move them to an external class ExtendedCamelContextExtension 

And, *possibly*, but not yet discussed, apply the same to the Message as well.

Lastly, after all this is done, we can define, Plugin, Extension and other interfaces that settles the consistent interface for handling this type of problem throughout the code.

As a side benefit, this also allow us to simplify the code on those cores classes (including their hierarchy) and opening up for the implementation of even more lightweight instances of them.

davsclaus (approved):

orpiske:
> ↩️ There are either **too many** changes to be tested in this PR or the code **needs be rebased**: (46 components likely to be affected)

This one won't be able to be tested on GH actions, so I am waiting for the results on my own CI.

orpiske:
We're good to go. Also, the checkstyle issue is unrelated.

PR #9344, opened 2023-02-14
CAMEL-19031: Check current saga status is RUNNING before add a step

zhfeng:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
@@ -16,14 +16,20 @@
  */
 package org.apache.camel.processor;
 
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.camel.CamelExecutionException;
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.model.SagaCompletionMode;
+import org.apache.camel.model.SagaPropagation;
 import org.apache.camel.saga.InMemorySagaService;
 import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class SagaTimeoutTest extends ContextTestSupport {
 
@@ -64,6 +70,29 @@ public class SagaTimeoutTest extends ContextTestSupport {
         compensate.assertIsNotSatisfied();
     }
 
+    @Test
+    public void testTimeoutMultiParticipants() throws Exception {
+
+        MockEndpoint compensate = getMockEndpoint("mock:compensate");
+        compensate.expectedMessageCount(1);
+
+        MockEndpoint complete = getMockEndpoint("mock:complete");
+        complete.expectedMessageCount(0);
+
+        MockEndpoint end = getMockEndpoint("mock:end");
+        end.expectedMessageCount(1);
+
+        CamelExecutionException ex = assertThrows(CamelExecutionException.class,
+                () -> {
+                    template.sendBody("direct:saga-multi-participants", "Hello");
+            });
+
+        assertEquals("Cannot begin: status is COMPENSATED", ex.getCause().getMessage());
+        end.assertIsSatisfied();
+        complete.assertIsSatisfied();
+        compensate.assertIsSatisfied();
+    }
+
     @Override
     protected RouteBuilder createRouteBuilder() throws Exception {
 
@@ -81,6 +110,27 @@ public class SagaTimeoutTest extends ContextTestSupport {
                         .compensation("mock:compensate").completion("mock:complete")
                         .to("mock:end");
 
+                from("direct:saga-multi-participants")
+                        .process(exchange -> {
+                            exchange.getMessage().setHeader("id", UUID.randomUUID().toString());
+                        })
+                        .saga()
+                        .propagation(SagaPropagation.REQUIRES_NEW)
+                        .to("direct:service1")
+                        .to("direct:service2");
+
+                from("direct:service1")
+                        .saga().option("id", header("id"))
+                        .propagation(SagaPropagation.MANDATORY).timeout(100, TimeUnit.MILLISECONDS)
+                        .compensation("mock:compensate").completion("mock:complete")
+                        .delay(300L)
+                        .to("mock:end");
+
+                from("direct:service2")
+                        .saga().option("id", header("id"))
+                        .propagation(SagaPropagation.MANDATORY).timeout(500, TimeUnit.MILLISECONDS)
+                        .compensation("mock:compensate").completion("mock:complete")
+                        .to("mock:end");
             }
         };
     }
--- a/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
+++ b/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
@@ -75,6 +75,13 @@ public class InMemorySagaCoordinator implements CamelSagaCoordinator {
 
     @Override
     public CompletableFuture<Void> beginStep(Exchange exchange, CamelSagaStep step) {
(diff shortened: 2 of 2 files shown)


oscerd (approved):

davsclaus (approved):

core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
```diff
@@ -75,6 +75,13 @@ public String getId() {
 
     @Override
     public CompletableFuture<Void> beginStep(Exchange exchange, CamelSagaStep step) {
+        Status status = currentStatus.get();
+        if (status != Status.RUNNING) {
+            CompletableFuture<Void> res = new CompletableFuture<>();
+            res.completeExceptionally(new IllegalStateException("Cannot begin: status is " + status));
+            return res;
+        }
+
```
essobedo, line 84: Would you mind adding a test case somewhere in https://github.com/apache/camel/blob/main/core/camel-core/src/test/java/org/apache/camel/processor to prove that it fixes the bug?

zhfeng: Sure and thanks for pointing the src directory out. I was thinking about a test case but have not find the right place.

davsclaus (approved):

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
```diff
@@ -81,6 +110,29 @@ public void configure() throws Exception {
...
+                        .to("direct:service1")
+                        .to("direct:service2");
+
+                from("direct:service1")
+                        .saga().option("id", header("id"))
+                        .propagation(SagaPropagation.MANDATORY).timeout(100, TimeUnit.MILLISECONDS)
+                        .compensation("mock:compensate").completion("mock:complete")
+                        .process(exchange -> {
+                            Thread.sleep(300L, 0);
+                        })
```
essobedo, line 128: You could use the `delay` EIP instead https://camel.apache.org/components/3.20.x/eips/delay-eip.html

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
```diff
@@ -64,6 +70,29 @@ public void testTimeoutHasNoEffectIfCompleted() throws Exception {
         compensate.assertIsNotSatisfied();
     }
 
+    @Test
+    public void testTimeoutMultiParticipants() throws Exception {
+
+        MockEndpoint compensate = getMockEndpoint("mock:compensate");
+        compensate.expectedMessageCount(1);
```
essobedo, line 77: That should be `2`, when a timeout expires all participants should be compensated

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
```diff
@@ -64,6 +70,29 @@ public void testTimeoutHasNoEffectIfCompleted() throws Exception {
...
+
+        MockEndpoint complete = getMockEndpoint("mock:complete");
+        complete.expectedMessageCount(2);
```
essobedo, line 80: That should be `0` since the Saga participants are compensated none can be completed. The participants cannot be both completed and compensated.

essobedo (changes requested):
Thanks for the unit test 🙏 Unless I miss something, the behavior doesn't seem correct, see my comments. We would also need to modify the incorrect doc regarding the timeout, especially this part "The minimum value of those timeouts is taken as timeout for the saga when they are composed together."

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
zhfeng: Nice catch! I will change to use `deplay`. Thanks!

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
zhfeng: There is only `service1` involved in the saga transction. `service2` can not join because the status is `COMPENSATED`. So I think it should be `1`.

zhfeng:
Thanks @essobedo for reviewing. In term of saga doc, we have opened a discussion on [narayana zulip](https://narayana.zulipchat.com/#narrow/stream/323714-users/topic/Saga.20timeout.20value.20depend.20on.20what.20.3F) and there is also a tiny configuation issue with `camel-lra-starter`, I just raised https://github.com/apache/camel-spring-boot/pull/752. I'd like to fix the doc issue in a seperated PR and maybe @chen19980 can help us.

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
essobedo: Oh, I missed that, it would be definitively more clear 😄 It makes me wonder in which case `assertIsNotSatisfied` can be useful, it sounds like a very dangerous method.

chen19980:
@zhfeng Yes, I can help to fix the doc about saga timeout description.But how do I edit or fix this issue ?

zhfeng:
@chen19980 No, you don't need to edit this PR. You can create a new PR to fix `core/camel-core-engine/src/main/docs/modules/eips/pages/saga-eip.adoc`. Feel free to ping me if you have any problem.

core/camel-core/src/test/java/org/apache/camel/processor/SagaTimeoutTest.java
zhfeng: Done

essobedo (approved):
👍 LGTM thx

essobedo:
I still believe that the way timeouts are managed when set on saga participants, doesn't seem to be correct as they should only apply to the participant itself while for now, it affects the whole saga, but maybe it can be done in different PR.

zhfeng:
@essobedo Did you check this [discussion](https://narayana.zulipchat.com/#narrow/stream/323714-users/topic/Saga.20timeout.20value.20depend.20on.20what.20.3F) on narayana zulip? 

>When a participant joins an existing LRA it is allowed to specify a new timeout, provided it is earlier than the one currently in effect, because the ability of a participant to be able to compensate may be time bound, i.e. if the new timelimit is earlier than the current one than it becomes the new limit and the LRA will be cancelled when the limit is reached if still active.

I think it follows the LRA Spec [1]

[1] https://download.eclipse.org/microprofile/microprofile-lra-2.0-RC1/microprofile-lra-spec-2.0-RC1.html#timing-out-lras-and-compensators

essobedo:
> @essobedo Did you check this [discussion](https://narayana.zulipchat.com/#narrow/stream/323714-users/topic/Saga.20timeout.20value.20depend.20on.20what.20.3F) on narayana zulip?
> 
> > When a participant joins an existing LRA it is allowed to specify a new timeout, provided it is earlier than the one currently in effect, because the ability of a participant to be able to compensate may be time bound, i.e. if the new timelimit is earlier than the current one than it becomes the new limit and the LRA will be cancelled when the limit is reached if still active.
> 
> I think it follows the LRA Spec [1]
> 
> [1] https://download.eclipse.org/microprofile/microprofile-lra-2.0-RC1/microprofile-lra-spec-2.0-RC1.html#timing-out-lras-and-compensators

No I didn't but it is not even the behavior, we have right now, instead of replacing an existing timeout, the timeout is extended. I mean, if I have 2 participants, and the timeouts set are 2 s for the first participant and 1 s for the second participant, instead of having the minimum of the 2 timeouts which is 1 s, we have the minimum of (2 s and the time of the first action + 1 s).

zhfeng:
SagaPropagationTest#testPropagationSupports is failing with this fix and it should be fix in
- https://github.com/apache/camel/pull/9368

I'm just curious why the CI did not find this failing test when checking this PR? @oscerd is there anything we are missing?

oscerd:
It could be the module containing saga propagation test was not part of the tests triggered

zhfeng:
It runs the `coreTest` but it looks ilke does not fail the build.

PR #9359, opened 2023-02-16
CAMEL-19049 - unwrap aries proxy & bytebuddy proxy classes when introspecting

rastislavpapp:
Fixes problem introduced in CAMEL-18411 in `BeanInfo` class, in 3.14.x, where methods from synthetic classes are no longer considered as overriding the original methods (via a change in the `#findMostSpecificOverride` method). 

This causes ambiguity when trying to invoke them. I'm mainly concerned about usage in Karaf, where the proxy class is created via aries-proxy. For simplicity, test case added uses bytebuddy.

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanInfo.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanInfo.java
@@ -63,7 +63,9 @@ public class BeanInfo {
     private static final Logger LOG = LoggerFactory.getLogger(BeanInfo.class);
     private static final String CGLIB_CLASS_SEPARATOR = "$$";
     private static final String CGLIB_METHOD_MARKER = "CGLIB$";
+    private static final String BYTE_BUDDY_CLASS_SEPARATOR = "$ByteBuddy$";
     private static final String BYTE_BUDDY_METHOD_MARKER = "$accessor$";
+    private static final String ARIES_PROXY_CLASS_PREFIX = "Proxy";
     private static final String[] EXCLUDED_METHOD_NAMES = new String[] {
             "clone", "equals", "finalize", "getClass", "hashCode", "notify", "notifyAll", "wait", // java.lang.Object
             "getInvocationHandler", "getProxyClass", "isProxyClass", "newProxyInstance" // java.lang.Proxy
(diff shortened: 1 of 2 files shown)


davsclaus:
Can you test with 3.18, or 3.20 before special PRs against an old Camel version.
We have fixed stuff and its likely there is another existing PR that can be backported instead.

davsclaus (changes requested):
Can you test with 3.18, or 3.20 before special PRs against an old Camel version.
We have fixed stuff and its likely there is another existing PR that can be backported instead.

components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanInfo.java
```diff
@@ -63,7 +63,9 @@ public class BeanInfo {
     private static final Logger LOG = LoggerFactory.getLogger(BeanInfo.class);
     private static final String CGLIB_CLASS_SEPARATOR = "$$";
     private static final String CGLIB_METHOD_MARKER = "CGLIB$";
+    private static final String BYTE_BUDDY_CLASS_SEPARATOR = "$ByteBuddy$";
     private static final String BYTE_BUDDY_METHOD_MARKER = "$accessor$";
+    private static final String ARIES_PROXY_CLASS_PREFIX = "Proxy";
```
davsclaus, line 68: This is not accepted as Proxy has no special marker that would cause problems for end users that have normal classes that are named `ProxyMyWorld` or something.

Why does Aries not use a special marker like the others do

rastislavpapp: The class name of aries-proxy class is created as:
```
className = "Proxy" + AbstractWovenProxyAdapter.getSanitizedUUIDString();
```
So it would be possible to create a regex for it, if we wanted to be more precise. But the class name is created in default java package, so it should be pretty resistant to real world user classes. At least that was how I thought about it.

rastislavpapp:
In addition to the problems above, I've found out that the solution does not work if the proxy is based on an interface.

components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanInfo.java
rastislavpapp: Or the check could be done via reflection, checking for some other markers.

oscerd: Reflection should be avoided.

davsclaus:
closing old PRs

PR #9366, opened 2023-02-16
CAMEL-19049 - fix for 3.18.x

rastislavpapp:
Fix of CAMEL-19049 based on #9359

Fixes problem introduced in [CAMEL-18411](https://issues.apache.org/jira/browse/CAMEL-18411) in BeanInfo class, in 3.14.x, where methods from synthetic classes are no longer considered as overriding the original methods (via a change in the #findMostSpecificOverride method).

This causes ambiguity when trying to invoke them. I'm mainly concerned about usage in Karaf, where the proxy class is created via aries-proxy. For simplicity, test case added uses bytebuddy.

### DIFF
--- a/components/camel-bean/pom.xml
+++ b/components/camel-bean/pom.xml
@@ -52,5 +52,30 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.aries.proxy</groupId>
+            <artifactId>org.apache.aries.proxy</artifactId>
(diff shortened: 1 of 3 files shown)


orpiske:
@davsclaus @oscerd @rastislavpapp Hi folks, do you happen to know the state of this PR? I wonder if it can be merged (after fixing the conflicts ...)

oscerd:
3.18 is EOL.

PR #9372, opened 2023-02-17
CAMEL-19070: camel-elasticsearch - Increase startup timeout

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-19070

## Motivation

The integration tests of the component `camel-elasticsearch` fail on the build pipeline and need to be fixed.

## Modifications:

* Upgrade elasticsearch (server and client) to the latest version
* Increase the startup timeout from 60 seconds to 90 seconds to fit with a slow-build pipeline


orpiske (approved):
Nice one, thanks!

PR #9374, opened 2023-02-17
CAMEL-19065: camel-kafka - Add KafkaSagaIT test

zhfeng:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaSagaIT.java
+++ b/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaSagaIT.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


davsclaus:
Hmm so there is a a special for this in camel-kafka - I dont like that, what about camel-jms and so on.
Does kafka not just store the header as-is, eg a byte[]. And if so can you not just in camel-saga convert the header to a String in case its byte[] or something else.

oscerd (changes requested):
This is wrong because we are adding logic not related to kafka in the Kafka component. This should be done internally in the LRA component. Same for any other headers with the same problem

zhfeng:
For `camel-jms`, it seems that it allows `String` as a value. So I think it should be OK in propagating the `SAGA_LONG_RUNNING_ACTION`
https://github.com/apache/camel/blob/f955a62d128804b45791d4f7e40b58ead58c866b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsBinding.java#L489-L491

zhfeng:
Hmm, I added a `KafkaSagaIT` test and it works. After debugging, it looks like
https://github.com/apache/camel/blob/f955a62d128804b45791d4f7e40b58ead58c866b/core/camel-base/src/generated/java/org/apache/camel/converter/CamelBaseBulkConverterLoader.java#L329-L331

which does this `byte[]` to `String` convertion. 
Anyway the reproducer https://github.com/codependent/booking-service is based on `camel-spring-boot`, so it might be some issues there and I will keep investigating.

zhfeng:
@davsclaus How to make sure such a `Converter` is also available for `camel-spring-boot`?

zhfeng:
`camel-spring-boot` also should work. There is an isue in the reproducer https://github.com/codependent/booking-service and it needs to use
```java
log("LRA id is \${headers.Long-Running-Action}")
```
to convert it to `String` explicitly.

oscerd (approved):

oscerd:
Now it's just an it test. Good to merge.

PR #9377, opened 2023-02-17
Micrometer Observation component

marcingrzejszczak:
A list of questions and TODOs related to this component that we must resolve before merging

Discuss:

- The `Tracer` abstraction from Camel has a notion of only high cardinality tags, we are missing low cardinality tags for metrics (that means that metrics will not have any tags at this point)
- Are the changes in the `Tracer` in the `tracing` component acceptable?
- Version should be 3.21.0 not 3.20.3 but there's no such branch

TODO:

- [ ] Spring Boot Starter is missing
- [ ] Polish the documentation
- [ ] Currently the behaviour is almost identical to OpenTelemetry one, that requires the presence of custom handlers - I'm considering to remove them so that the additional code will not be necessary. That will mean that with Micrometer Observation with OTel Tracer the spans will be a little bit different than with the OpenTelemetry component


oscerd (commented):
Still have to review everything. But I think we need two PRs. One against 3.x and one against main. 3.20.x is LTS and we won't introduce a new component there. Thanks. I'll try to find some time this week to fully review.

PR #9389, opened 2023-02-20
 Micrometer Observation component

marcingrzejszczak:
A list of questions and TODOs related to this component that we must resolve before merging

Discuss:

- The `Tracer` abstraction from Camel has a notion of only high cardinality tags, we are missing low cardinality tags for metrics (that means that metrics will not have any tags at this point)
- Are the changes in the `Tracer` in the `tracing` component acceptable?

TODO:

- [ ] Spring Boot Starter is missing
- [ ] Polish the documentation
- [ ] Currently the behaviour is almost identical to OpenTelemetry one, that requires the presence of custom handlers - I'm considering to remove them so that the additional code will not be necessary. That will mean that with Micrometer Observation with OTel Tracer the spans will be a little bit different than with the OpenTelemetry component

Recrated from https://github.com/apache/camel/pull/9377


components/camel-observation/src/test/java/org/apache/camel/observation/CamelMicrometerObservationTestSupport.java
```diff
@@ -0,0 +1,255 @@
...
+        verify(expected, async);
+    }
+
+    protected List<SpanData> verify(SpanTestData[] expected, boolean async) {
+        List<SpanData> spans = inMemorySpanExporter.getFinishedSpanItems();
+        spans.forEach(mockSpan -> {
+            System.out.println("Span: " + mockSpan);
+            System.out.println("\tComponent: " + mockSpan.getAttributes().get(COMPONENT_KEY));
+            System.out.println("\tTags: " + mockSpan.getAttributes());
+            System.out.println("\tLogs: ");
```
orpiske, line 136: This should go to a log file, to make it easier for us in the CI.

orpiske (changes requested):
Just adding a comment about something that caught my eye (but haven't reviewed it fully).

components/camel-observation/src/test/java/org/apache/camel/observation/CamelMicrometerObservationTestSupport.java
marcingrzejszczak: Just FYI I've literally copied that from OpenTelemetry code.

orpiske: Thanks for the heads up! I will fix it there.

oscerd (approved):
This looks good to me with a rapid review.

oscerd:
> A list of questions and TODOs related to this component that we must resolve before merging
> 
> Discuss:
> 
> * The `Tracer` abstraction from Camel has a notion of only high cardinality tags, we are missing low cardinality tags for metrics (that means that metrics will not have any tags at this point)
> * Are the changes in the `Tracer` in the `tracing` component acceptable?
> 
> TODO:
> 
> * [ ]  Spring Boot Starter is missing

This is not a problem, we could do that later

> * [ ]  Polish the documentation

This could be done with a subsequent PR

> * [ ]  Currently the behaviour is almost identical to OpenTelemetry one, that requires the presence of custom handlers - I'm considering to remove them so that the additional code will not be necessary. That will mean that with Micrometer Observation with OTel Tracer the spans will be a little bit different than with the OpenTelemetry component

On this, I'm not really sure about it: if we goes this way it's important to document and notes the difference, otherwise users might think the result will be the same with Micrometer and Opentelemetry and it won't be the case

> Recrated from #9377

oscerd:
Please rebase on 3.x, so we are good. Waiting for @orpiske and @davsclaus reviews.

orpiske (approved):
Thanks for the contribution!

davsclaus (approved):

davsclaus:
Thanks for the big work, there is a little merge conflict, I wonder if you could resolve that and push update to this PR so it can be merged cleanly.

Also we would need this to be ported to Camel v4 (main) but that is JDK17 and JakartaEE; which I assume micrometer also works with, but just wanted to mention the difference to Camel v3.

marcingrzejszczak:
I fixed the conflicts and did some additional changes

- modified the tracer api to introduce a low cardinality tag (those would be used for metrics and traces)
- modified the current micrometer observation behaviour to remove the custom handlers

Below you have a screenshot of a trace view for a very simple route

```
@Component
class MyRoute extends RouteBuilder {

	@Override
	public void configure() throws Exception {
		from("direct://foo")
				.log("hello")
				.to("http://httpbin.org/headers");
	}

}
```

image

With the Micrometer Observation you also get the following metric.

```
MeterId{name='camel.operation', tags=[tag(HTTP_METHOD=GET),tag(HTTP_STATUS=200),tag(component=camel-http),tag(error=none)]}
```

I also changed the behaviour so that it is close to the OpenTelemetry one but not the same - I will document the differences plus I've left the code in the tests that can be used to have the same behaviour.

So the question to you is - do you find such information valuable (traces and metrics) in its current form if you were to troubleshoot any issues within a Camel based system? I can try a more complex scenario and we can discuss what metrics / tags can be more useful in your opinion. We can also play around with the names if you think e.g. name of the component should be there in a span name.

As for Jakarta - Micrometer is working fine with JDK17.

cc @bclozel

marcingrzejszczak:
So I've been playing around with this and I'm really surprised with how it's actually working currently with OpenTelemetry. Let's assume that I have the following routes

```java
from("direct://foo")
				.log("hello")
				.to("direct://bar")
				.to("http://example.org");
		from("direct://bar")
				.log("hello")
				.to("direct://micrometer")
				.to("http://example.org");
		from("direct://micrometer")
				.setHeader(MicrometerConstants.HEADER_METRIC_NAME, constant("new.name"))
				.setHeader(MicrometerConstants.HEADER_METRIC_TAGS, constant(Tags.of("dynamic-key", "dynamic-value")))
				.to("micrometer:counter:name.not.used?tags=key=value")
				.to("direct://baz");
		from("direct://baz")
				.log("hello")
				.to("bean:com.example.cameldemo.MyBean")
				.to("exec:wc?args=--words /usr/share/dict/words")
				.process(exchange -> {
					// Use the Camel Exec String type converter to convert the ExecResult to String
					// In this case, the stdout is considered as output
					String wordCountOutput = exchange.getIn().getBody(String.class);
					// do something with the word count
					System.out.println(wordCountOutput);
				})
				.to("file:///tmp/camel-outputdir?flatten=true")
				.to("http://example.org");
```

I would assume that for the whole flow (e.g. from `producerTemplate.send("file:///tmp/camel-outputdir?flatten=true", defaultExchange);`) to the end I would have a single trace. From what I see this is not the case cause all `direct://` components are completely ignored in terms of continuing a trace (look at `TracingEventNotifier:shouldExclude` method).

image

image

Is this really the intent of how this should actually work? I would assume that there should be 1 trace for the whole processing because otherwise this doesn't make a lot of sense (at least to me).

oscerd:
As far as I remember it was not really easy to have a full trace, in the end we were able to have one trace for each route.

This means from..to, when you pass to another from..to, like for example

		from("direct://bar")
				.log("hello")
				.to("direct://micrometer")

This will be another trace. I don't remember the reason, but I think instrumenting it to take into account all the routes for a single trace was too hard to implement or too invasive. I might remember wrongly nevertheless

marcingrzejszczak:
> This will be another trace. I don't remember the reason, but I think instrumenting it to take into account all the routes for a single trace was too hard to implement or too invasive. I might remember wrongly nevertheless

This will actually not be a trace because `direct` seems to be an internal component that gets ignored.

oscerd:
> > This will be another trace. I don't remember the reason, but I think instrumenting it to take into account all the routes for a single trace was too hard to implement or too invasive. I might remember wrongly nevertheless
> 
> This will actually not be a trace because `direct` seems to be an internal component that gets ignored.

This looks wrong then. Can you open an issue?

oscerd:
Thanks

marcingrzejszczak:
IMO the graph should like this (please correct me if I'm wrong)

```java
from("direct://foo")
				.log("hello")
				.to("direct://bar")
				.to("http://example.org");
		from("direct://bar")
				.log("hello")
				.to("direct://micrometer")
				.to("http://example.org");
		from("direct://micrometer")
				.setHeader(MicrometerConstants.HEADER_METRIC_NAME, constant("new.name"))
				.setHeader(MicrometerConstants.HEADER_METRIC_TAGS, constant(Tags.of("dynamic-key", "dynamic-value")))
				.to("micrometer:counter:name.not.used?tags=key=value")
				.to("direct://baz");
		from("direct://baz")
				.log("hello")
				.to("bean:com.example.cameldemo.MyBean")
				.to("exec:wc?args=--words /usr/share/dict/words")
				.process(exchange -> {
					// Use the Camel Exec String type converter to convert the ExecResult to String
					// In this case, the stdout is considered as output
					String wordCountOutput = exchange.getIn().getBody(String.class);
					// do something with the word count
					System.out.println(wordCountOutput);
				})
				.to("file:///tmp/camel-outputdir?flatten=true")
				.to("http://example.org");
```

image

We have a red graph beacuse I'm explicitly throwing an exception in a controller in my application. 

I've achieved this graph by removing the instanceof check here (https://github.com/apache/camel/blob/camel-3.x/components/camel-tracing/src/main/java/org/apache/camel/tracing/Tracer.java#L296) but obviously I might be missing sth. I'm not committing that change, but if you confirm my assumptions I can file an issue and point to a potential solution.

marcingrzejszczak:
image

Actually everything is fine after doing the before mentioned change. I didn't have the file or exec only because I was throwing an exception in the bean component and the exchange processing was stopped :facepalm: 

So how do we proceed? I mean should I commit the change with the exclusion or not?

davsclaus:
The latest observation screenshot looks really great, with all the span details.

@marcingrzejszczak are you referring to the `shouldExclude` method you linked above, where you can only make this work if you remove that entirely or what?

davsclaus:
Ah okay, so `AbstractInternalSpanDecorator` is intended for camel components that are used for internal communication like `direct` / `seda` etc. And in your example above you call routes from routes via direct:bar/direct:baz. Do you not see spans for those or how does it filter those?

marcingrzejszczak:
So without my change afair you'll see spans but not for direct components and you'll not see a single trace for the whole flow of a message but you'll see multiple different traces. That's because spans are no longer maintaining parent child relationship.

davsclaus:
Okay I think we can accept that code change. Are all the unit test passing with this PR in both camel-tracing and camel-micrometer ?

marcingrzejszczak:
> Okay I think we can accept that code change. Are all the unit test passing with this PR in both camel-tracing and camel-micrometer ?

If I do the change to the `shouldExclude` a lot of tests in `camel-opentelemetry` are failing because number of spans is invalid (previously they were ignored, now they are not).

oscerd:
Maybe we could mark them as disabled and get back to them once we have the fix in place.

marcingrzejszczak:
It's up to you - I don't have a problem with adding back the code from the `shouldExclude` method so that we merge this with all the tests passing and then filing a new issue that most likely `shouldExclude` method is broken and then we could fix `micrometer`, `opentracing` and `opentelemetry` module tests. WDYT?

oscerd:
If you time to do that, it would be nice, we are probably in the same situation on main, 3.x and 3.20.x, so we should fix the shouldExclude method on LTS branch too (3.20.x). To me it's +1. Thanks a lot for your time.

marcingrzejszczak:
OK so I'm changing this PR as ready for review

marcingrzejszczak:
Hopefully I fixed checkstyle issues and also I've created the new issue here https://issues.apache.org/jira/browse/CAMEL-19124

davsclaus:
Can you work on the SB starter for the 3.x branch, so we have all together. Then afterwards we can possible cherry pick to main branch. If not then we appreciate if you would then work on PRs for main so we have this in v4 as well.

davsclaus:
merging this for 3.x so its included in Camel 3.21 onwards.

davsclaus:
Also I wonder if you would contribute an example such as for camel-spring-boot-examples. And you are welcome to include screenshots in the example as it helps users to quickly see this awesome feature

davsclaus:
Screenshots is also welcome in docs. However they may be harder to do, and the img may need to go in docs root folder where we have images for other pages such as EIPs

marcingrzejszczak:
Sure I can work on the starter. Of course the screenshots should be done after we fix the issue with shouldExclude method because the trace view will be broken

marcingrzejszczak:
Spring Boot configuration https://github.com/apache/camel-spring-boot/pull/784
Spring Boot example https://github.com/apache/camel-spring-boot-examples/pull/98

PR #9399, opened 2023-02-23
CAMEL-18988: camel-core - tests failing due to OOM (part 2)

essobedo:
fixes https://issues.apache.org/jira/browse/CAMEL-18988 (part 2)

## Motivation

Since we upgraded to surefire + failsafe M8, some of our tests have started to fail with `OnOutOfMemoryError`.

## Modifications

In practice, only the JVM option `-XX:+HeapDumpOnOutOfMemoryError` is needed as a thread dump can be extracted from an `hprof` file.

* Ensure that an `hprof` file is always generated on OOME
* Remove duplicated JVM options
* Reduce the max pool size of the thread pool used to launch the tests to limit the impact on memory since there is no real memory leak so the OOMEs are likely due to an excessive amount of threads 

## Result

The heap size doesn't exceed 1 Go during the tests whereas it could reach 1.5 Go before the change.

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -41,7 +41,7 @@
         <compiler.fork>true</compiler.fork>
         <cxf.codegenplugin.forkmode>true</cxf.codegenplugin.forkmode>
         <cxf.codegen.jvmArgs/>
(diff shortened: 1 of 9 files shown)


orpiske (approved):
Nice one @essobedo !

PR #9417, opened 2023-02-27
(chores) camel-kafka: Avoid misreporting liveness checks

orpiske:


orpiske:
@davsclaus FYI, this is the strange behavior I mentioned on the Zulip Chat. It seems that invoking `HealthCheckHelper.invokeLiveness(context);` returns an empty result. Maybe we are doing something wrong on the test?

oscerd (approved):

davsclaus (approved):

davsclaus:
Yeah we should revisit the health checks, as 98% should be readiness only (connectivity problems etc).

Liveness is harder to write as health check as its when you run low on memory, file descriptors, or some other weird errors, where the container needs to be restarted.

So we should likely make liveness = false by default. And then if possible write a basic liveness check to have it report as UP by default.

oscerd (approved):

davsclaus:
I created a ticket about liveness = false by default for health checks
https://issues.apache.org/jira/browse/CAMEL-19118

davsclaus (approved):

orpiske:
> I created a ticket about liveness = false by default for health checks https://issues.apache.org/jira/browse/CAMEL-19118

Oh, cool! Thanks @davsclaus

PR #9426, opened 2023-02-27
(perf chores) camel-seda: cache queue reference to avoid costly operations in the hot path

orpiske:
This one seems to add a very small (but still measurable) throughput improvement for Seda by avoiding a few synchronized calls. It's part of several other perf patches I will be sending 

Compared w/ 4.0.0-M1 (both w/ 4 consumers and 1 producer): 

```
--- Patch
Mean: 484459.8571428573
Geometric mean: 484272.99632561835
Standard deviation: 13570.203212282813

--- 4.0.0-M1
Mean: 480957.58035714296
Geometric mean: 479545.8325907664
Standard deviation: 33212.22104132737
```

### DIFF
--- a/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaEndpoint.java
+++ b/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaEndpoint.java
@@ -100,6 +100,7 @@ public class SedaEndpoint extends DefaultEndpoint implements AsyncEndpoint, Brow
     private boolean discardIfNoConsumers;
 
     private BlockingQueueFactory<Exchange> queueFactory;
+    private volatile QueueReference ref;
 
     public SedaEndpoint() {
         queueFactory = new LinkedBlockingQueueFactory<>();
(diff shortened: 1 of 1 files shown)


oscerd (approved):

components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaEndpoint.java
```diff
@@ -213,8 +215,14 @@ protected BlockingQueue<Exchange> createQueue() {
      * @return the reference, or <tt>null</tt> if no queue reference exists.
      */
     public synchronized QueueReference getQueueReference() {
-        String key = getComponent().getQueueKey(getEndpointUri());
-        QueueReference ref = getComponent().getQueueReference(key);
+        if (key == null) {
```
davsclaus, line 218: I wonder if this can be done in doStart - I assume getEndpointUri is static at the moment when its started, and maybe then this can be computed once, and then the synchronization may not be need at all

orpiske: 🤔 Thanks! Sounds like a good idea and I can give it a try. 

I'll dispatch the other PRs I have on the queue and then I'll go back to this one.

orpiske: It turns out we can do that. I updated the PR with your suggestion, thanks!

components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaEndpoint.java
```diff
@@ -100,6 +100,8 @@ public class SedaEndpoint extends DefaultEndpoint implements AsyncEndpoint, Brow
     private boolean discardIfNoConsumers;
 
     private BlockingQueueFactory<Exchange> queueFactory;
+    private String key;
+    private QueueReference ref;
```
essobedo, line 104: Only `ref` seems to be needed here. Moreover, please note that it should be `volatile` otherwise you have unfortunately no guarantee that its value will be visible from a thread executed by a different core

orpiske: I've just checked the code. We do have a few occasions where this could happen. Let's make it `volatile` here. Thanks @essobedo!

davsclaus (approved):

essobedo (approved):

PR #9429, opened 2023-02-27
CAMEL-19060: move the isFailureHandled logic to a field in the ExchangeExtension to avoid costly operations

orpiske:
This is a small optimization aimed at that avoiding costly operations in AbstractExchange#getProperty. It brings a small but noticeable improvement to SEDA (and probably others): 

```
-- Patch
Mean: 457300.04464285704
Geometric mean: 457246.17949513963
Standard deviation: 7040.578210525716

-- 4.0.0-M1
Maximum: 490420.0
Mean: 453472.6964285715
Geometric mean: 453389.6965851105
Standard deviation: 8802.284218119967
```


orpiske:
@davsclaus @oscerd I guess one of you might one to double check this PR.

oscerd (approved):

core/camel-support/src/main/java/org/apache/camel/support/ExtendedExchangeExtension.java
```diff
@@ -30,6 +30,7 @@
 
 public class ExtendedExchangeExtension implements ExchangeExtension {
     private final AbstractExchange exchange;
+    private boolean failureHandled;
```
davsclaus, line 33: The other flags are in AbstractExchange, so maybe we should move it to where they are.

orpiske: Noted, thanks! I'll take a look at that.

core/camel-core-processor/src/main/java/org/apache/camel/processor/OnCompletionProcessor.java
```diff
@@ -190,8 +190,8 @@ protected static void doProcess(Processor processor, Exchange exchange) {
         } finally {
             // restore the options
             exchange.setRouteStop(stop);
-            if (failureHandled != null) {
-                exchange.setProperty(ExchangePropertyKey.FAILURE_HANDLED, failureHandled);
+            if (failureHandled) {
```
davsclaus, line 193: I wonder if we have anywhere in the code, where we call: setFailureHandled(false), as then this code will not restore that value.

However I suspect we do not have that, as we use Boolean.TRUE before, and "remove" or "null" to say false/none. But may be worth to grep the source to find what we do

orpiske: It seems we don't do that. Neither IntelliJ nor `egrep -inR "setFailureHandled" *` found any instance of us setting the flag to false. I believe we should be fine with this one.

davsclaus (approved):

PR #9432, opened 2023-02-27
CAMEL-19058: fix a type check scability issue for Message types

orpiske:
Using the type check agent, we can see that the Message type is hit pretty hard by the type check scalability issue and reports quite a lot of occurrences for these types: 

```
--------------------------
3:	org.apache.camel.support.DefaultMessage
Count:	197370941
Types:
	org.apache.camel.spi.DataTypeAware
	org.apache.camel.CamelContextAware
Traces:
	org.apache.camel.support.MessageSupport.copyFrom(MessageSupport.java:199)
		class: org.apache.camel.spi.DataTypeAware
		count: 98685471
	org.apache.camel.CamelContextAware.trySetCamelContext(CamelContextAware.java:28)
		class: org.apache.camel.CamelContextAware
		count: 98685470
```

This resolves most of them.

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsMessage.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsMessage.java
@@ -25,6 +25,7 @@ import jakarta.jms.Message;
 import jakarta.jms.Queue;
 import jakarta.jms.Session;
 import jakarta.jms.Topic;
+
 import org.apache.camel.Exchange;
 import org.apache.camel.RuntimeExchangeException;
 import org.apache.camel.support.DefaultMessage;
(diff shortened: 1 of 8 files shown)


orpiske:
Obs.: this one does not necessarily bring material changes to the throughput as the code is contended in other places, but does solve the reports.

orpiske:
Obs. 2: it uses the same pattern as in the other fixes, which replaces inheritance with delegation.

oscerd (approved):

davsclaus:
Hmm now we create a new object per message (the trait) only for fringe features like JMS external redelivery and dataType (not really much in use).

Also mind that PooledExchange has a reset that, see reset method on Message. You likely need to reset the trait or something.

orpiske:
> Hmm now we create a new object per message (the trait) only for fringe features like JMS external redelivery and dataType (not really much in use).

Yeah. There's that, indeed - though the memory overhead is probably small. One alternative I considered is to have the traits as a bit mask. It would make the code a little bit more complex, but may consume less memory overall. 

What do you think?

> 
> Also mind that PooledExchange has a reset that, see reset method on Message. You likely need to reset the trait or something.

Thanks, noted. I'll take a look.

orpiske:
@davsclaus actually, let me try the alternative approach on this one. Let's put it on hold for now.

oscerd (approved):

orpiske:
I am planning to go back to this work very soon. I had to divert my priorities to something else.

orpiske:
I'm closing this one. This was superseded by #9968

PR #9433, opened 2023-02-27
CAMEL-19060 + CAMEL-19058: minor improvements

orpiske:
Mostly minor fixes, but do seem to provide some improvements overall:

```
-- Patch:
Mean: 528232.0625000002
Geometric mean: 518388.324464431
Standard deviation: 97022.2449142631

-- 4.0.0-M1 
Mean: 451989.19642857154
Geometric mean: 451657.5975966994
Standard deviation: 17755.625310286025
```

### DIFF
--- a/core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
+++ b/core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
@@ -102,7 +102,9 @@ public class MainDurationEventNotifier extends EventNotifierSupport {
 
         if (maxMessages > 0 && complete) {
             boolean result = doneMessages.incrementAndGet() >= maxMessages;
-            LOG.trace("Duration max messages check {} >= {} -> {}", doneMessages.get(), maxMessages, result);
+            if (LOG.isTraceEnabled()) {
+                LOG.trace("Duration max messages check {} >= {} -> {}", doneMessages.get(), maxMessages, result);
+            }
 
             if (result && shutdownStrategy.isRunAllowed()) {
                 if ("shutdown".equalsIgnoreCase(action)) {
--- a/core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
+++ b/core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
@@ -23,10 +23,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.spi.CamelEvent;
-import org.apache.camel.spi.CamelEvent.ExchangeCompletedEvent;
-import org.apache.camel.spi.CamelEvent.ExchangeCreatedEvent;
-import org.apache.camel.spi.CamelEvent.ExchangeFailedEvent;
-import org.apache.camel.spi.CamelEvent.RouteReloadedEvent;
 import org.apache.camel.support.EventNotifierSupport;
 import org.apache.camel.util.StopWatch;
 import org.slf4j.Logger;
@@ -78,15 +74,13 @@ public class MainDurationEventNotifier extends EventNotifierSupport {
         }
     }
 
-    protected void doNotify(CamelEvent event) throws Exception {
+    protected void doNotify(CamelEvent event) {
         // ignore any event that is received if shutdown is in process
         if (!shutdownStrategy.isRunAllowed()) {
             return;
         }
 
-        boolean begin = event instanceof ExchangeCreatedEvent;
-        boolean complete = event instanceof ExchangeCompletedEvent || event instanceof ExchangeFailedEvent;
-        boolean reloaded = event instanceof RouteReloadedEvent;
+        final boolean reloaded = event.getType() == CamelEvent.Type.RouteReloaded;
 
         if (reloaded) {
             if (restartDuration) {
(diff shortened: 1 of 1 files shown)


core/camel-support/src/main/java/org/apache/camel/support/cache/DefaultProducerCache.java
```diff
@@ -173,7 +175,7 @@ public void releaseProducer(Endpoint endpoint, AsyncProducer producer) {
     @Override
     public Exchange send(Endpoint endpoint, Exchange exchange, Processor resultProcessor) {
         if (camelContext.isStopped()) {
-            exchange.setException(new RejectedExecutionException("CamelContext is stopped"));
+            exchange.setException(REJECTED_EXECUTION_EXCEPTION);
```
essobedo, line 178: Warning if you do so, the stack trace won't be correct anymore so it will make the debugging more complex

orpiske: That's a good point. I wonder if we could get away with it by using a better exception message. 

@davsclaus any thoughts about this?

orpiske: We could probably instantiate it here, instead of in the declaration. That way it would retain the stack trace.

orpiske: Or just call `fillInStackTrace` once.

oscerd (approved):

core/camel-support/src/main/java/org/apache/camel/support/cache/DefaultProducerCache.java
orpiske: I changed it to fill the stack trace once, so I guess this should do the trick.

davsclaus: Interresting optimization - however its ONLY when camel is stopped so I wonder what you see in those test that makes it worth to do this? in other words its only if the user tries to use the producer after camel has been stopped.

orpiske: This doesn't affect the hot part of the test, but I did notice that it was creating a lot of objects like this during the stop (under load, using main with `camel.main.durationMaxMessages`). 

So I guess this tends to occur when we have a lot of exchanges in transit and something has triggered the shutdown?

orpiske: Obs.: this one is pretty minor and I think we can get rid of it if you guys think this is not worth the effort. I don't think this one matters much in the big picture.

core/camel-support/src/main/java/org/apache/camel/support/cache/DefaultProducerCache.java
```diff
@@ -54,6 +54,10 @@
 public class DefaultProducerCache extends ServiceSupport implements ProducerCache {
 
     private static final Logger LOG = LoggerFactory.getLogger(DefaultProducerCache.class);
+    private static final RejectedExecutionException REJECTED_EXECUTION_EXCEPTION
+            = new RejectedExecutionException("CamelContext is stopped");
+    private volatile boolean filledStackTrace;
```
essobedo, line 59: You could rather use an `AtomicReference<RejectedExecutionException>` instead , then in `send` method you lazily create it with `exchange.setException(REJECTED_EXECUTION_EXCEPTION.updateAndGet(ex -> ex == null ? new RejectedExecutionException("CamelContext is stopped") : ex));`

orpiske: 🤔

core/camel-support/src/main/java/org/apache/camel/support/cache/DefaultProducerCache.java
```diff
@@ -173,7 +177,12 @@ public void releaseProducer(Endpoint endpoint, AsyncProducer producer) {
     @Override
     public Exchange send(Endpoint endpoint, Exchange exchange, Processor resultProcessor) {
         if (camelContext.isStopped()) {
-            exchange.setException(new RejectedExecutionException("CamelContext is stopped"));
+            if (!filledStackTrace) {
+                REJECTED_EXECUTION_EXCEPTION.fillInStackTrace();
+                filledStackTrace = true;
```
davsclaus, line 182: -1 to such strange code to happen during stopping.

We should only optimize for core routing engine (keep basic code clean and simple).

orpiske: Agreed. Let's rollback on this one as it's too much effort for something minor.

core/camel-support/src/main/java/org/apache/camel/support/cache/DefaultProducerCache.java
orpiske: I'm rolling back on this one for now (it's a minor thing - not important for the hot path), but I liked the idea. In the future, we should explore this in other areas of the code base.

essobedo: Yes that could be interesting in a place where we could create a new exception very frequently as building a stack trace is very slow

essobedo (approved):

core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
```diff
@@ -100,9 +94,12 @@ protected void doNotify(CamelEvent event) throws Exception {
             return;
         }
 
+        final boolean complete = event.getType() == CamelEvent.Type.ExchangeCompleted || event.getType() == CamelEvent.Type.ExchangeFailed;
```
davsclaus, line 97: I wonder if you can optimize this to only compute complete and begin if maxMessages > 0 and maxIdelSeconds > 0

orpiske: Good idea! I think I can. I will double check tomorrow.

davsclaus (approved):

oscerd (approved):

davsclaus (approved):

PR #9436, opened 2023-02-27
(chores) camel-base-engine: cleanup duplicated code

orpiske:


oscerd (approved):

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
```diff
@@ -813,10 +801,10 @@ public void after(Exchange exchange, UnitOfWork uow) throws Exception {
                 UnitOfWorkHelper.doneUow(uow, exchange);
             }
 
-            // after UoW is done lets pop the route context which must be done on every existing UoW
             if (route != null && existing != null) {
                 existing.popRoute();
             }
+            // after UoW is done lets pop the route context which must be done on every existing UoW
```
davsclaus, line 807: this comment better belongs where it was before

davsclaus (commented):
I guess there is no noticable performance hit with the extra method call? The method is too big for the JVM to inlining

orpiske:
> I guess there is no noticable performance hit with the extra method call? The method is too big for the JVM to inlining


Yeah, this one is probably too big for the inlining with the default JVM settings. The greatest benefit of consolidating this code is having nicer stack frames with the async profiler (and that makes it easier to rule out the parts that are performing nicely).

davsclaus (approved):

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
orpiske: Oh, alright. I'll merge and then move it back in place!

PR #9451, opened 2023-02-28
[CAMEL-19100] Fixes lost dataChangeFilterTrigger in camel-milo

rangoy:
# Description

Milo component does not use dataChangeFilterTrigger value from route, this was replaced by StatusValueTimestamp during setup.

Ref https://issues.apache.org/jira/browse/CAMEL-19100

# Target

Targets 3.x branch as camel-milo is disabled in 4.x branch, waiting for jakarata update of eclipse milo, ref https://github.com/eclipse/milo/issues/1098

# Tracking


oscerd (approved):

rangoy:
I see that I might have chosen the wrong target branch. Should I change it to 3.x?

essobedo:
> I see that I might have chosen the wrong target branch. Should I change it to 3.x?

3.20.x is to target 3.20.3, 3.x is to target 3.21.0 and main is to target 4.0-M2

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MonitorFilterConfiguration.java
```diff
@@ -80,7 +80,9 @@ public MonitoringFilter createMonitoringFilter() {
         switch (monitorFilterType) {
             default:
                 return new DataChangeFilter(
-                        DataChangeTrigger.StatusValueTimestamp, this.getDataChangeFilterDeadbandType(),
+                        this.dataChangeFilterTrigger != null
+                                ? this.dataChangeFilterTrigger : DataChangeTrigger.StatusValueTimestamp,
```
essobedo, line 84: Warning this is a behavior change, as far as I can see the current default value is [`DataChangeTrigger.StatusValue`](https://github.com/apache/camel/blob/c1a8be35640603c35160aba07d715c3cd3878af0/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MonitorFilterConfiguration.java#L41) whereas at runtime it used `DataChangeTrigger.StatusValueTimestamp`. Maybe should change the default value to `DataChangeTrigger.StatusValueTimestamp` and use `dataChangeFilterTrigger` here.

rangoy: Good point. I did not pay attenation to that the filter had a default different from what was actually applied here. 

I've updated the pullrequest to use the old effective default value (StatusValueTimestamp)

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MonitorFilterConfiguration.java
```diff
@@ -37,8 +37,8 @@ public class MonitorFilterConfiguration implements Cloneable {
     @UriParam(defaultValue = "0.0", description = "Deadband value for MonitorFilterType DataChangeFilter.")
     private Double dataChangeFilterDeadbandValue = 0.0;
 
-    @UriParam(description = "Daata change trigger for data change monitor filter type.")
-    private DataChangeTrigger dataChangeFilterTrigger = DataChangeTrigger.StatusValue;
+    @UriParam(description = "Data change trigger for data change monitor filter type.")
```
essobedo, line 40:
```suggestion
    @UriParam(defaultValue = "StatusValueTimestamp", description = "Data change trigger for data change monitor filter type.")

```
Let's add the default value to the metadata too

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MonitorFilterConfiguration.java
essobedo: For the sake of consistency with the rest of the code, don't check if it is `null` only call the corresponding getter

essobedo (approved):

essobedo:
@rangoy Could you please propose also a PR for the branch 3.x?

rangoy:
> @rangoy Could you please propose also a PR for the branch 3.x?

Sure. From my fork?

essobedo:
Yes please

PR #9454, opened 2023-02-28
[CAMEL-19101] Features for Camel-Micrometer

shmakovalexey:
# Description

Features for MicrometerRoutePolicy

Repair test
Allow users to use one MicrometerRoutePolicy instance for multiple routes. This is especially useful when using OSGI XML
Allow users to disable additional counters
Allow users to set up timers with advanced micrometer API settings
Written for all this tests

PR #9461, opened 2023-03-02
[Camel 19106] repair check style for 3.x

shmakovalexey:
check style was broken, I repair it

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -4260,7 +4260,7 @@
                                 </sourceDirectories>
                                 <headerLocation>header-java.txt</headerLocation>
                                 <includes>**/*.java,**/*.groovy,**/*.scala,**/*.properties,**/*.xml,**/*.xsd</includes>
-                                <excludes>${sourcecheckExcludesComma}**/archetype-resources/**/*.java,**/archetype-resources/**/*.groovy,**/archetype-resources/**/*.scala,**/archetype-resources/**/*.xml,**/org/json/simple/**/*.java</excludes>
+                                <excludes>${sourcecheckExcludesComma}**/archetype-resources/**/*.java,**/archetype-resources/**/*.groovy,**/archetype-resources/**/*.scala,**/archetype-resources/**/*.xml,**/org/json/simple/**/*.java,**/test/resources/**/*.xml</excludes>
                             </configuration>
                             <goals>
                                 <goal>check</goal>
--- a/tooling/maven/camel-package-maven-plugin/src/main/java/org/apache/camel/maven/packaging/PackageArchetypeCatalogMojo.java
+++ b/tooling/maven/camel-package-maven-plugin/src/main/java/org/apache/camel/maven/packaging/PackageArchetypeCatalogMojo.java
@@ -205,8 +205,26 @@ public class PackageArchetypeCatalogMojo extends AbstractMojo {
     }
 
     private static void writeTop(FileOutputStream fos) throws IOException {
-        String top = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<archetype-catalog>\n  <archetypes>";
-        fos.write(top.getBytes());
+        fos.write("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n".getBytes());
+        writeLicense(fos);
+        fos.write("<archetype-catalog>\n  <archetypes>".getBytes());
+    }
+
+    private static void writeLicense(FileOutputStream fos) throws IOException {
+        fos.write("<!--\n\n".getBytes());
+        fos.write("    Licensed to the Apache Software Foundation (ASF) under one or more\n".getBytes());
+        fos.write("    contributor license agreements.  See the NOTICE file distributed with\n".getBytes());
+        fos.write("    this work for additional information regarding copyright ownership.\n".getBytes());
+        fos.write("    The ASF licenses this file to You under the Apache License, Version 2.0\n".getBytes());
+        fos.write("    (the \"License\"); you may not use this file except in compliance with\n".getBytes());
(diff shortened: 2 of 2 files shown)


oscerd (approved):

shmakovalexey:
Error on checkstyle
[ERROR] Failed to execute goal on project camel-allcomponents: Could not resolve dependencies for project org.apache.camel:camel-allcomponents:pom:3.21.0-SNAPSHOT: Could not find artifact org.apache.camel:camel-zeebe:jar:3.21.0-SNAPSHOT in apache.snapshots (https://repository.apache.org/snapshots/)

I solved this problem on the local machine by installing the project in the local repository (mvn install). I don't know how to solve it in CI.

essobedo:
Please don't merge it, I need to check something first

oscerd:
Too late.

essobedo:
This PR sounds incorrect, why would we need those changes on this particular branch

oscerd:
Because checkstyle is failing.

essobedo:
If there is a problem somewhere it needs to be solved everywhere the same way

essobedo:
We were only supposed to cherry-pick some missing commits, here with those changes we have a different "fix" that sounds more like an XY problem

oscerd:
This doesn't make any sense: on main the checkstyle is not failing because the codebase is different and probably the checkstyle plugin has a different version. 3.x shouldn't be freeze as is, it's still in development

oscerd:
I don't have particular point on this: to me this fix is not incorrect. If you pretend to only cherry pick from main, you will never be able to do it. It won't work in the long term.

oscerd:
Revert it, if you wish to.

oscerd:
Reverted. Sorry @shmakovalexey , I hope you don't mind and keep contributing.

essobedo:
I've checked on the branches `3.20.x`, there is no checkstyle issue and this branch uses the exact same version of checkstyle as the branch `3.x`. Then, I checked on `3.x` with 0f9e74de42bdf8022aabaec9593ec935979847a4 as head which means before merging these changes, I don't get any checkstyle issue either.

The only checkstyle issue has been fixed 2 days ago by this https://github.com/apache/camel/commit/18383df5bbeca179b432257ed35d7aeef1bea848 which is more or less a cherry-pick of https://github.com/apache/camel/commit/44b02d6af3b93a05b4adbe9a6e377651738a0055 for main and https://github.com/apache/camel/commit/41cf79533d6c6778f4f3c2ed8d6afe29816ec97d for 3.20.x.

I mentioned it in the other PR https://github.com/apache/camel/pull/9454#issuecomment-1450382321

oscerd:
All of this rigidity doesn't make any sense to me. If you stop/veto harmless contributions like this one you won't improve the community and people will run away. This was completely harmless and being pedantic on this it's just a waste of time and energy.

oscerd:
All done btw, reverted and happy Friday!

shmakovalexey:
@essobedo @oscerd I spent 2 days of personal time and just won't leave it)
see https://issues.apache.org/jira/browse/CAMEL-19106 
Line length in file "_components\camel-spring-ws\src\test\resources\org\apache\camel\component\spring\ws\addresing\ProducerParamsBasicTest-context.xml:60_" is 278 characters
There are a lot of XML files in the test resources with lines longer than **250** characters

The file "_buildingtools/srcmain/resources/camel-checkstyle.xml:95_" contains the following rule


```xml
   <module name="LineLength">
        <property name="max" value="250"/>
    </module>
```
   

maven-checkstyle-plugin settings contain:
```xml
<includes>**/*.java,**/*.groovy,**/*.scala,**/*.properties,**/*.xml,**/*.xsd</includes>
```

**How can this rule for check XML not work???** Maybe something is cached on your PC and doesn't work?
Please, cd _components\camel-spring-ws_ and make clean and run checkstyle

I have Windows 10 and Java 11 on my PC. If everything works on your machine, the problem is deeper than it seemed - in the plugin itself  or even in Java 11, and depends on the platform. This will mean that the XML files on your machine are simply not processed by the maven-checkstyle-plugin and it does not check them.

Maybe it has something to do with the OS? Is something wrong with Windows? More precisely, on Windows everything works as it should, XML files are checked, but on Linux it doesn't.  Because on Windows, the check broke not only test XML files, also in the generated XML file, where forgoted to add a header with a Apache-License.

This can be dangerous - you expect XML files to be checked and it doesn't happen. If I'm right - then my solution will of course fix the style errors in the XML, but it will open up a strange issue with style validation on Linux, whose solution no one knows.

Well, or maybe everything works for everyone, and I'm just a schizophrenic from the Internet who invents errors that do not exist, and spends too much personal time on its solution. But I am absolutely sure that I am mentally healthy, and that I spent 2 days of my personal time not fighting illusions in my head, but trying to solve the problem of building the project.

There is another option to confirm the theory about the problem associated with Windows / Linux - find a PC or Virtual PC with Windows and conduct an experiment on Windows. 

I'm 99% sure I'm right - the problematic files and settings are the same in _main, camel-3.20, and camel-3.x_. I checked. And since everything works for everyone, it means that no one tried to build on Windows. And checking in XML files just doesn't work on other OS

PS - Sorry for my horrible English, I use Google Translate

shmakovalexey:
Sorry, I found all the broken rules in the supress file
```
<suppressionsLocation>camel-checkstyle-suppressions.xml</suppressionsLocation>
```
Probably the problem is that for some reason it is not processed on my PC. I will look into this more carefully.

shmakovalexey:
I found the cause of the style check problem on Windows and created a issue (CAMEL-19114), this PR is really useless and wrong. Sorry again for the previous emotional comment

PR #9471, opened 2023-03-07
deliveryTag -> For manual ack

mayurvaid-redvest:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-spring-rabbitmq/src/main/docs/spring-rabbitmq-component.adoc
+++ b/components/camel-spring-rabbitmq/src/main/docs/spring-rabbitmq-component.adoc
@@ -238,5 +238,22 @@ from("direct:order")
   .toD("spring-rabbit:order-${header.orderType}");
 ----
 
+== Manual Acknowledgement 
+
+If we need to manually acknowledge a message for some use case , we can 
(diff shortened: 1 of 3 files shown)


davsclaus:
Are you able to update the docs (see src/main/docs) about this delivery tag and how to do a manual commit

oscerd (changes requested):
I think deliveryTag String should be declared in the Constants class.

davsclaus:
> I think deliveryTag String should be declared in the Constants class.

Yes

mayurvaid-redvest:
> Are you able to update the docs (see src/main/docs) about this delivery tag and how to do a manual ack

Thanks for reviewing , Will add this documentation

davsclaus:
any update?

mayurvaid-redvest:
> any update?

Updated the docs and added constant as well

oscerd (approved):

mayurvaid-redvest:
> I wonder if the delivery tag only make sense to have as a header if ack mode is manual? Are there situations where you need this information elsewhere ?

Yes Even i am not aware of scenario where we would need this , But having and header would anyways help and gives flexibility to opt for it or not , What are your thoughts

davsclaus (approved):

PR #9481, opened 2023-03-07
CAMEL-18957: converted camel-jms to camel-test-infra-core

orpiske:
This one comes with a fix I want to backport, so, please don't squash if merging. 

Obs.: I am leaving on PTO (again 😄 ) so, I will be slow to reply (in case there's some issue w/ the PR).


orpiske:
Obs.: this one converts only the non-persistent and non-spring tests.

davsclaus (approved):

PR #9510, opened 2023-03-10
Convert expression when setting header / property on message / exchange.

ammachado:
# Description

Convert an expression to an object value when it is used on a `setHeader` method on a `Message` or on a `setProperty` on an `Exchange`

# Target

# Tracking

  Local build is failing with JAXB generation errors on `camel-spring-xml`

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/impl/DefaultExchangeTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/DefaultExchangeTest.java
@@ -29,6 +29,7 @@ import org.apache.camel.Message;
 import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.SafeCopyProperty;
 import org.apache.camel.TypeConversionException;
+import org.apache.camel.builder.Builder;
 import org.apache.camel.support.DefaultExchange;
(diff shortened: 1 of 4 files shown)


davsclaus (changes requested):
whats the use case for this? We should not change core code without a really good reason

ammachado:
I helped debugging a production issue where some code was refactored from a route definition (`from(...).setHeader("name", constant("value")`) to a processor (`exchange.getMessage().setHeader("name", constant("value")`).  Where previously the header was set to the correct value, after the refactor the header was being set to `constant{value}`.

The `setHeader` and `setProperty` methods on Java DSL have overloads for `Expression` and `Supplier` to resolve the values, where the methods on the message / exchange don't.

davsclaus:
okay but this PR is wrong as the Route DSL uses the mode/expression, and Java API uses the object as-is. So to set a constant value, set it directly.

The setHeader/setProperty are optimized for core use and we should not add overhead with this. Also when you set the value you evaluate it as well this is wrong, as some Camel components allow using an Expression as a header for dynamic computation that they evaluate later.

So the correct solution is to

exchange.getMessage().setHeader("name", "value")

PR #9546, opened 2023-03-15
CAMEL-19143: Added option to quote fields only when necessary

bricefrisco:
# Description

Implements [CAMEL-19143](https://issues.apache.org/jira/browse/CAMEL-19143)

The `quoting` option of the camel-bindy component enables quoting of all fields. This option provides the ability to quote fields only when it is necessary.

According to [RFC 4180 2.6](https://www.ietf.org/rfc/rfc4180.txt):
> 6.  Fields containing line breaks (CRLF), double quotes, and commas
>        should be enclosed in double-quotes.  For example:
>        "aaa","b CRLF
>        bb","ccc" CRLF
>        zzz,yyy,xxx

This option enables quoting a field **if and only if** that field:
contains the configured `crlf` character OR
contains the configured `quote` character OR
contains the configured `separator` character

# Target

# Tracking

### DIFF
--- a/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/BindyCsvFactory.java
+++ b/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/BindyCsvFactory.java
@@ -73,6 +73,7 @@ public class BindyCsvFactory extends BindyAbstractFactory implements BindyFactor
     private boolean autospanLine;
     private boolean allowEmptyStream;
     private boolean quotingEscaped;
+    private boolean quotingOnlyWhenNeeded;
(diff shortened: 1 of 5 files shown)


bricefrisco:
I am not sure what is wrong with `checkstyle` and `build`, these commands are passing for me:
```
mvn clean install -Psourcecheck
mvn clean install
```

If there is something I have missed please let me know :)

oscerd (approved):
I think this needs to go on main too

bricefrisco:
> I think this needs to go on main too

Ok, I have created https://github.com/apache/camel/pull/9552 as well.

PR #9566, opened 2023-03-16
[CAMEL-19160]add Object to String converter method to camel-jaxb

ffang:
# Description

# Target

# Tracking


davsclaus (changes requested):
Be careful with such general converter with Object -> String. Can you give more details in the JIRA why you look for this, and lets find a better way without the type converters.

PR #9568, opened 2023-03-17
CAMEL-18995:  Upgrade to HttpComponents 5.x

essobedo:
Fix for  https://issues.apache.org/jira/browse/CAMEL-18995

## Motivation

In order to get the latest improvements and bug fixes, we need to upgrade to HttpComponents 4.

## Modifications:

* Added the versions of HttpCore and HttpClient to the parent pom. Note that version 4 is still used to align the third-party libraries using v4 and for components not yet migrated.
*Removed HttpComponents from the dependencies of `camel-catalog-maven` and `camel-report-maven-plugin` as it doesn't seem to be used
* Upgared all components but `camel-as2`, `camel-olingo2` and `camel-olingo4` as they need to be done in a separate task
* Fixed some violations raised


oscerd (approved):
Very well done!Thanks!

essobedo:
Why the build is not launched?

davsclaus (approved):
Great work as this was one of the big migrations to do

PR #9585, opened 2023-03-20
Fix platform proxy test

Croway:


Croway:
Nice catch @essobedo, thanks for the headsup!

I thought the issue was with HTTP headers and http component.

PR #9592, opened 2023-03-21
Explicitly complete/abandon Azure Service Bus messages

dylanpiergies:
If an exception occurred during Exchange processing, the exception must be rethrown in the message receive and error handlers to trigger the `ServiceBusReceiverAsyncClient` to abandon the message so that the message is requeued.

Fixes [CAMEL-19155](https://issues.apache.org/jira/browse/CAMEL-19155).

### DIFF
--- a/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
+++ b/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
@@ -145,6 +145,9 @@ public class ServiceBusConsumer extends DefaultConsumer {
         // use default consumer callback
         AsyncCallback cb = defaultConsumerCallback(exchange, true);
         getAsyncProcessor().process(exchange, cb);
+        if (exchange.getException() != null) {
+            throw RuntimeCamelException.wrapRuntimeException(exchange.getException());
+        }
     }
 
     private Exchange createServiceBusExchange(final ServiceBusReceivedMessage receivedMessage) {
@@ -190,6 +193,7 @@ public class ServiceBusConsumer extends DefaultConsumer {
             getExceptionHandler().handleException("Error processing exchange", exchange,
                     exchange.getException());
         }
+        throw RuntimeCamelException.wrapRuntimeException(errorContext);
     }
 
     private Exchange createServiceBusExchange(final Throwable errorContext) {
--- a/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
+++ b/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
@@ -145,9 +145,6 @@ public class ServiceBusConsumer extends DefaultConsumer {
         // use default consumer callback
         AsyncCallback cb = defaultConsumerCallback(exchange, true);
         getAsyncProcessor().process(exchange, cb);
-        if (exchange.getException() != null) {
-            throw RuntimeCamelException.wrapRuntimeException(exchange.getException());
-        }
     }
 
     private Exchange createServiceBusExchange(final ServiceBusReceivedMessage receivedMessage) {
@@ -193,7 +190,6 @@ public class ServiceBusConsumer extends DefaultConsumer {
             getExceptionHandler().handleException("Error processing exchange", exchange,
                     exchange.getException());
         }
-        throw RuntimeCamelException.wrapRuntimeException(errorContext);
     }
 
     private Exchange createServiceBusExchange(final Throwable errorContext) {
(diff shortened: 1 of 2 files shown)


oscerd (approved):
Thanks. This needs to go on 3.x and main branches too. Thanks

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
```diff
@@ -145,6 +145,9 @@ private void onEventListener(final ServiceBusReceivedMessage message) {
         // use default consumer callback
         AsyncCallback cb = defaultConsumerCallback(exchange, true);
         getAsyncProcessor().process(exchange, cb);
+        if (exchange.getException() != null) {
```
davsclaus, line 148: Its actually a bit more complex if you use aync processing, as the "result" is only done when the callback is done, and not when the process is returned.

Is there some API in azure to rollback/abandon we can explicit call instead of rethrowing exception

davsclaus (changes requested):

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
dylanpiergies: There is, but it requires the `ServiceBusReceivedMessage` to be in scope and it returns a `Mono<Void>` to which we need to subscribe in order for it to execute. The way things are right now, an invocation to `ServiceBusReceiverAsyncClientWrapper#complete` or `ServiceBusReceiverAsyncClientWrapper#abandon` would return a `Mono<Void>` on which we'd then have to invoke `block()`.

I've looked into this problem in some depth, and I'm still looking for the "right" approach to addressing this. I kinda feel like using the async client isn't a very good fit, because we're then dealing with a component that is reactive from the outset and I don't think there's any framework within Camel for handling this in a clean and intuitive way (although correct me if I'm wrong).

I understand your point about the result not being valid until the callback completes, but while I was step debugging I did find that callback was always being executed before `getAsyncProcessor().process(exchange, cb);` returned. This probably isn't great either, because `onEventListener` is the processor for a `Flux` subscription which I believe places it within a reactive context and it therefore shouldn't block.

dylanpiergies: Went with (2). I _think_ this implementation is correct. Please check.

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/client/ServiceBusClientFactory.java
```diff
@@ -79,9 +79,7 @@ private static ServiceBusClientBuilder.ServiceBusReceiverClientBuilder createBas
             final ServiceBusClientBuilder busClientBuilder, final ServiceBusConfiguration configuration) {
         final ServiceBusClientBuilder.ServiceBusReceiverClientBuilder receiverClientBuilder = busClientBuilder.receiver();
 
-        if (configuration.isDisableAutoComplete()) {
-            receiverClientBuilder.disableAutoComplete();
-        }
+        receiverClientBuilder.disableAutoComplete();
```
davsclaus, line 82: Could you add a code comment here why this is disabled - eg that you take hands on commit / rollback in the consumer manually

davsclaus (approved):

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/client/ServiceBusClientFactory.java
dylanpiergies: Done.

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/client/ServiceBusClientFactory.java
```diff
@@ -79,9 +79,10 @@ private static ServiceBusClientBuilder.ServiceBusReceiverClientBuilder createBas
             final ServiceBusClientBuilder busClientBuilder, final ServiceBusConfiguration configuration) {
         final ServiceBusClientBuilder.ServiceBusReceiverClientBuilder receiverClientBuilder = busClientBuilder.receiver();
 
-        if (configuration.isDisableAutoComplete()) {
-            receiverClientBuilder.disableAutoComplete();
-        }
+        // We handle auto-complete in the producer, since we have no way to propagate errors back to the reactive
```
davsclaus, line 82: in the consumer ???

dylanpiergies: :facepalm: Corrected.

davsclaus:
Thanks, can you update the other PRs so they are all same code so we can merge this for the branches, thanks.

dylanpiergies:
> Thanks, can you update the other PRs so they are all same code so we can merge this for the branches, thanks.

Already done. :) Do you want me to squash them down as well?

davsclaus:
we can squash here from github, so that is not needed

PR #9597, opened 2023-03-22
Disable Infinispan tests depending on JTA

essobedo:
## Motivation

[Some tests](https://ci-builds.apache.org/job/Camel/job/Camel%20JDK17/job/main/692/testReport/) are failing on Jenkins because Infinispan doesn't support Jakarta Transactions

## Modifications:

* Disable the tests that depend on JTA


davsclaus (approved):

zhfeng:
@essobedo What are `JTA 1` and `JTA 2`? Are they related to transactions? I just curious about the context here.

essobedo:
> @essobedo What are `JTA 1` and `JTA 2`? Are they related to transactions? I just curious about the context here.

Yes, JTA stands for Java Transaction API, but you make me realize that it has been renamed to [Jakarta Transactions](https://en.wikipedia.org/wiki/Jakarta_Transactions)

zhfeng:
Good catch! So is there any issue tracking for Infinispan to support Jakarta Transaction?

oscerd (approved):

zhfeng (approved):

essobedo:
After a deeper investigation, the problem is not Infinispan but its test sources especially the class `MultipleCacheManagersTest` that we use for these tests and that is not Jakarta compatible, so the impact is really limited

essobedo:
I finally found the proper fix, I will create another PR for it

PR #9617, opened 2023-03-23
CAMEL-15105: move handling of internal services to a separate class

orpiske:
This helps further work to decouple and standardize the extension/plugin management

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
@@ -61,7 +61,6 @@ import org.apache.camel.Processor;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.ResolveEndpointFailedException;
 import org.apache.camel.Route;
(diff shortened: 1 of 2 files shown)


orpiske:
How about Proxy? Looking at what I have to do next, it will end up changing to a proxy instead of a facade.

davsclaus:
its not really a proxy, its some thing that helps with add a service internally in camel, to make sure its done in the correctly.

orpiske:
> Okay but what is the reason for introducing a proxy - that is something we should avoid, and leave objects as-is.

I may have confused you because I am talking about not about this change, but the upcoming one. The proxy I am talking about is the [Proxy design pattern](https://refactoring.guru/design-patterns/proxy).  

As in: 

Current state (tightly coupled) -> evolves to Facade (which I'll name as Helper) -> then evolves to a Proxy. 

The reason it evolves to a Proxy is that then, in the future, we can simplify / more easily abstract the service handling. Alternatively I have been thinking of using a Bridge, but I am thinking it won't be as flexible.

orpiske:
@davsclaus For this one, any other comments besides renaming it to Helper?

davsclaus (approved):

PR #9622, opened 2023-03-24
CAMEL-15105: decouple handling of internal services

orpiske:
Implement a handler for managing the internal services and their lifecycle.


orpiske:
I settled for a simpler design using a bridge instead of a proxy. This will allow moving/simplifying/etc the get/setExtension code in the `AbstractCamelContext` and making access to plugins more uniform and consistent.

PR #9632, opened 2023-03-24
CAMEL-19191: Use 5.1.x version to be compatible with SB

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-19191

## Motivation

SpringBoot still uses Http Components v 5.1 which is not fully compatible with v 5.2, we need to downgrade the version to prevent conflicts with the SpringBoot runtime.

## Modifications:

* Downgrade the version to  `5.1.4`
* Remove the option `socketTimeout` since it has been removed from the potential configuration of an `HttpClient` instance
* Adapt the documentation of `camel-http` to version 5
* Adapt the upgrade guide to version 5.1


parent/pom.xml
```diff
@@ -212,7 +212,7 @@
         <hsqldb-version>2.7.1</hsqldb-version>
         <httpunit-version>1.7</httpunit-version>
         <httpcore-version>5.2.1</httpcore-version>
```
davsclaus, line 214: httpcore should that not be 5.1.x also

essobedo: Clearly, my bad

PR #9634, opened 2023-03-24
CAMEL-18957 camel-paho: convert to use the JUnit 5 extension for the CamelContext

orpiske:


orpiske:
I may need to review this again on Monday, so marking as a draft to see how it works on Github

oscerd (approved):

components/camel-paho/pom.xml
```diff
@@ -48,15 +48,17 @@
 
         <dependency>
             <groupId>org.apache.camel</groupId>
-            <artifactId>camel-test-junit5</artifactId>
+            <artifactId>camel-core-catalog</artifactId>
```
davsclaus, line 51: I wonder why you need the catalog ?

orpiske: TBH, I don't know. It was [there previously](https://github.com/apache/camel/blob/main/components/camel-paho/pom.xml#L54-L58). It was just displayed here due to the text block movement.

orpiske: Looking more closely at this one, it seems it was added as part of [CAMEL-14499](https://issues.apache.org/jira/browse/CAMEL-14499). However, it appears to me the `camel-core-catalog` is not needed since the tests pass without it ... maybe we can clean this up (in paho and in other components on that list).

davsclaus: Yes it can be removed

orpiske: Great. I'll send a separate PR cleaning this up.

orpiske: Done: #9649.

davsclaus (approved):

orpiske:
> ### Components tested:
> 1 	1 	0 	1

I think this one is good to go. Let's merge.

PR #9648, opened 2023-03-27
provide a way to define global groovy variables

EvanMi:
# Description
When using groovy script, we sometime need to config some global variables without pollute the camelContext and exchange,  that say,  in every script we want to use not only the request, response, context, but also some other global variables, and don't want to put them in the header etc.
So we need a interface method to meet this request.
# Target

### DIFF
--- a/components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
+++ b/components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
@@ -19,11 +19,17 @@ package org.apache.camel.language.groovy;
 import groovy.lang.GroovyShell;
 import org.apache.camel.Exchange;
 
+import java.util.Map;
+
 public interface GroovyShellFactory {
 
     GroovyShell createGroovyShell(Exchange exchange);
 
     default String getFileName(Exchange exchange) {
+        return null;
+    }
+
+    default Map<String, Object> getVariables(Exchange exchange) {
         return null;
     }
 
--- a/components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
+++ b/components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
@@ -29,6 +29,12 @@ public interface GroovyShellFactory {
         return null;
     }
 
+    /**
+     * This method provide a way to define some global variables that will be applied to all the groovy script context.
+     *
+     * @param exchange the camel exchange in process.
+     * @return the global variables that will be applied to all the groovy script context.
+     */
     default Map<String, Object> getVariables(Exchange exchange) {
         return null;
     }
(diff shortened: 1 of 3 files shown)


oscerd (approved):

components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
```diff
@@ -27,4 +29,8 @@ default String getFileName(Exchange exchange) {
         return null;
     }
 
+    default Map<String, Object> getVariables(Exchange exchange) {
```
davsclaus, line 32: Can you add javadoc with details what is the purpose of this new method

EvanMi: Ok, by the way, if still need to push this commit to 3.x ?

davsclaus (approved):

davsclaus:
Can you create a JIRA so we can track this change, thanks.

EvanMi:
> Can you create a JIRA so we can track this change, thanks.

Sorry, i never used JIRA before, any guidelines?

essobedo:
There is a checkstyle issue raised by the CI:
```
[ERROR] /home/runner/work/camel/camel/components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java:22:1: Wrong order for 'java.util.Map' import. [ImportOrder]
Audit done.
```

components/camel-groovy/src/main/java/org/apache/camel/language/groovy/GroovyShellFactory.java
```diff
@@ -27,4 +29,14 @@ default String getFileName(Exchange exchange) {
...
     }
 
+    /**
+     * This method provide a way to define some global variables that will be applied to all the groovy script context.
+     *
+     * @param exchange the camel exchange in process.
+     * @return the global variables that will be applied to all the groovy script context.
+     */
+    default Map<String, Object> getVariables(Exchange exchange) {
+        return null;
```
essobedo, line 39: Maybe we could return an empty map instead, to avoid having to check if it is null and then have a code easier to read

davsclaus:
You can create an account for JIRA here
https://selfserve.apache.org/jira-account.html

EvanMi:
> You can create an account for JIRA here https://selfserve.apache.org/jira-account.html

Thanks for help, i have create account already, is there anything else to do? @davsclaus

davsclaus:
I created a JIRA
https://issues.apache.org/jira/browse/CAMEL-19212

EvanMi:
@davsclaus how can i merge this pull request to 3.x, i already saw this feature in 4.x jar file, but my team still use 3.x.

essobedo:
It has already been backported into 3.x branch https://github.com/apache/camel/commit/291d7365a46559ffc25909f41c5ac968d95b87f9 so it will be available in 3.21

EvanMi:
> It has already been backported into 3.x branch [291d736](https://github.com/apache/camel/commit/291d7365a46559ffc25909f41c5ac968d95b87f9) so it will be available in 3.21

clear, thanks.

PR #9659, opened 2023-03-27
CAMEL-19210 - Support for delayed delivery pulsar messages using deliverAt (for 3.x)

MrBlueMoo:
# Description

This functionality allows for delayed delivery options within Apache Pulsar to be used by specifying a time that you want the message to be delivered. 

This is simply is a passthrough for setting the deliverAt() on an Apache Pulsar message used used by the producer. The expected value is a Long representing the current timestamp in milliseconds (UTC) which is simply passed across without translation. 

The detail of the call used is as follows:
https://pulsar.apache.org/api/client/2.11.x/org/apache/pulsar/client/api/TypedMessageBuilder.html#deliverAt(long)

Uses the CamelPulsarProducerMessageDeliverAt header name for setting.

# Target

# Tracking
- [X] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue] https://issues.apache.org/jira/browse/CAMEL-19210


oscerd (approved):
Thanks. This should go on main too

davsclaus:
I have cherry picked from main

PR #9660, opened 2023-03-27
CAMEL-19210 - Support for delayed delivery pulsar messages using deliverAt (for main)

MrBlueMoo:
Duplicate of #9659 for main!

# Description

This functionality allows for delayed delivery options within Apache Pulsar to be used by specifying a time that you want the message to be delivered. 

This is simply is a passthrough for setting the deliverAt() on an Apache Pulsar message used used by the producer. The expected value is a Long representing the current timestamp in milliseconds (UTC) which is simply passed across without translation. 

The detail of the call used is as follows:
https://pulsar.apache.org/api/client/2.11.x/org/apache/pulsar/client/api/TypedMessageBuilder.html#deliverAt(long)

Uses the CamelPulsarProducerMessageDeliverAt header name for setting.

# Target

# Tracking
- [X] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue] https://issues.apache.org/jira/browse/CAMEL-19210


oscerd (approved):

davsclaus:
Can the description of the header be a bit more improved. The time is in millis since epoch and I assume its the local time where the client is running. And what happens if the time is in the past? Or if the time is too far in the future (eg +1 year or something) ?

davsclaus:
Ah you can copy their javadoc as its better

```
Deliver the message only at or after the specified absolute timestamp.
The timestamp is milliseconds and based on UTC (eg: [System.currentTimeMillis()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#currentTimeMillis()).

Note: messages are only delivered with delay when a consumer is consuming through a [SubscriptionType.Shared](https://pulsar.apache.org/api/client/2.11.x/org/apache/pulsar/client/api/SubscriptionType.html#Shared) subscription. With other subscription types, the messages will still be delivered immediately.
```

davsclaus (approved):

PR #9673, opened 2023-03-29
CAMEL-15105: initial plugin/extension rework

orpiske:
After all the other moves and refactorings, this finally gets to the point of CAMEL-15105: to work on a uniform interface for plugins/extensions. 

This provides only the first step, which delegates the plugin management to a PluginManager class. In subsequent PRs, these calls will be modified to provide uniform access to the plugins via specific setter/getters (thus closing the work on CAMEL-15105).


oscerd (approved):

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultContextPluginManager.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+import org.apache.camel.RuntimeCamelException;
+import org.apache.camel.health.HealthCheckRegistry;
+import org.apache.camel.spi.PluginManager;
+
+public class DefaultContextPluginManager implements PluginManager {
+    private final Map<Class<?>, Object> extensions = new ConcurrentHashMap<>();
+    private HealthCheckRegistry hcr;
```
davsclaus, line 30: is this in use?

davsclaus (approved):

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultContextPluginManager.java
orpiske: Oh man, I forgot to remove this. This variable is for the next step. Thanks for pointing it.

PR #9675, opened 2023-03-29
(chores) camel-cassandra: minor cleanups

orpiske:


oscerd (approved):

oscerd:
@orpiske i think the removal of System println should go on 3.x too

components/camel-cassandraql/src/main/java/org/apache/camel/component/cassandra/CassandraEndpoint.java
```diff
@@ -165,7 +169,8 @@ protected CqlSessionBuilder createSessionBuilder() {
 
         if (extraTypeCodecs != null) {
             String[] c = extraTypeCodecs.split(",");
-            System.err.println(c.toString());
+            LOG.error(Arrays.toString(c));
```
davsclaus, line 172: We should not log this at error level, instead at debug or trace level

orpiske: Fixed, thanks!

davsclaus (approved):

PR #9685, opened 2023-03-29
(chores) camel-base-engine: fix JMX enable/disable logic

orpiske:
I broke JMX enable/disable logic with one of the changes. This should fix the logic.


oscerd (approved):

orpiske:
I tested this one manually and it should be fine now. Merging earlier as I want to have some combined tests on Apache CI.

PR #9686, opened 2023-03-29
(chores) camel-base-engine: use class equality checks for accessible classes

orpiske:


oscerd (approved):

davsclaus:
I am not really sure about this as those classes can be loaded in various runtimes (also OSGi) and then you can't really rely on .class equals.

PR #9751, opened 2023-04-01
fixed results of inspection "HashSet can be replaced with EnumSet"

dk2k:
fixed results of inspection "HashSet can be replaced with EnumSet"

### DIFF
--- a/components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/spi/Method.java
+++ b/components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/spi/Method.java
@@ -16,7 +16,10 @@
  */
 package org.apache.camel.component.platform.http.spi;
 
-import java.util.*;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Locale;
+import java.util.Set;
 
 /**
  * An HTTP method.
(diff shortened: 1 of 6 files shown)


components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/spi/Method.java
```diff
@@ -16,11 +16,7 @@
  */
 package org.apache.camel.component.platform.http.spi;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Locale;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
davsclaus, line 19: We do not use wildcard imports

davsclaus (changes requested):
see prev comments

components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/spi/Method.java
dk2k: Fixed. It's IDE's optimization

dk2k: Sorry for accidentally merging branch apache:main in

davsclaus (approved):

PR #9755, opened 2023-04-01
replaced iterated addition to collection with bulk operation

dk2k:
replaced iterated addition to collection with bulk operation


orpiske (approved):

orpiske:
@dk2k please, can you rebase this code so it picks some fixes for our CI? Thanks.

PR #9770, opened 2023-04-02
CAMEL-18904:  Add functions to simple language to create empty map/li…

louisa-fr:
# Description
Ticket:  https://issues.apache.org/jira/browse/CAMEL-18904

This is one of my first OpenSource-Commits, so I hope I understood the task correctly. 
I wasn't quite sure, of which type the empty json should be or which other types would be useful, so this is still an open TODO.


davsclaus:
Thanks, we can skip json for now. And then make equalsIgnoreCase for all of them.

louisa-fr:
Thank you, I removed the open json-TODO for now and used equalsIgnoreCase for String, List and Map

davsclaus (approved):

PR #9771, opened 2023-04-02
Camel 19234: avoid ambiguity when looking up a bean 

louisa-fr:
Issue: https://issues.apache.org/jira/browse/CAMEL-19234

This code was a suggestion from the camel-zulip-chat, where it wasn't sure, whether it would have any side effects, so we thought we would open a PR with it to trigger discussion as it would help unblock an issue in the quarkus-artemis project :) 

Zulip-Chat: https://camel.zulipchat.com/#narrow/stream/257302-camel-quarkus/topic/Resolution.20of.20.40Identifier.20dependency

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/CamelContextHelper.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/CamelContextHelper.java
@@ -243,9 +243,12 @@ public final class CamelContextHelper {
      * it to the given type or throws NoSuchBeanException if not found.
      */
     public static <T> T mandatoryLookupAndConvert(CamelContext context, String name, Class<T> beanType) {
-        Object value = lookup(context, name);
+        Object value = lookup(context, name, beanType);
         if (value == null) {
-            throw new NoSuchBeanException(name, beanType.getName());
+            value = lookup(context, name);
+            if (value == null) {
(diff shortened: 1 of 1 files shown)


davsclaus (changes requested):
There is already mandatoryLookup that takes name and type you should use instead. This method is for converting to the type as it also states in the javadoc.

turing85:
@davsclaus I am not quite sure this is possible. We do not call this part directly; it is called through the runtime, especially in the `quarkus-camel` extension. See [the corresponding JIRA issue](https://issues.apache.org/jira/browse/CAMEL-19234).

If it is possible, can you give us a pointer where we have to adjust the corresponding call?

turing85:
No results. The stack trace in https://github.com/apache/camel-quarkus/issues/4725 indicates that it comes from some implementation of the `AbstractCamelContext` (-> `JmsComponent`) -> `DefaultComponent` -> `EndpointHelper` -> `CamelContextHelper`

davsclaus:
Here is a PR with a fix
https://github.com/apache/camel/pull/9811

Try this on CEQ

louisa-fr:
@turing85 has verified that https://github.com/apache/camel/pull/9811 fixes the issue, thanks for working on it ! I'll close this PR then.

PR #9791, opened 2023-04-03
changed multiplication by a pow of two to bit shift or to the result of multiplication

dk2k:
changed multiplication by a pow of two to bit shift or to the result of multiplication

### DIFF
--- a/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2AsynchronousMDNManager.java
+++ b/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2AsynchronousMDNManager.java
@@ -121,7 +121,7 @@ public class AS2AsynchronousMDNManager {
 
         String requestUri = buildRequestURI(uri);
 
-        int buffSize = 8 * 1024;
+        int buffSize = 8192; // 8 * 1024
 
         try (DefaultBHttpClientConnection httpConnection = new DefaultBHttpClientConnection(buffSize)) {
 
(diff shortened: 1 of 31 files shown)


dk2k:
@orpiske Hi, would you change your mind if I added into the code comments (sometimes I did already) as for what is multiplicated?

orpiske:
> @orpiske Hi, would you change your mind if I added into the code comments (sometimes I did already) as for what is multiplicated?

Yes. I think it would help to make it easier to understand.

Thanks

davsclaus:
Is the JDK compiler not optimising such things automatically ? Its a lot harder to read and understand as bit shifting operation only makes sense when you work with bit values / computations.

orpiske:
Yeah, the JDK is very likely to optimize this automatically.

dk2k:
@orpiske @davsclaus I found an article dated to 2020 https://programmer.help/blogs/the-optimization-and-application-of-bit-operation-in-java.html Left shift gives some performance gain.
My idea was to get as more performance as possible since Camel is a server software. Bit shifts are basis of computer science. If one doesn't know them, the sources of Camel can't help him/her.

orpiske:
> @orpiske @davsclaus I found an article dated to 2020 https://programmer.help/blogs/the-optimization-and-application-of-bit-operation-in-java.html Left shift gives some performance gain. My idea was to get as more performance as possible since Camel is a server software. Bit shifts are basis of computer science. If one doesn't know them, the sources of Camel can't help him/her.

While, indeed, even ourselves do micro-optimizations from time to time, we usually try to favor the ones that improve readability (i.e.: like the ones yourself did replacing the `""` checks for `isEmpty()`). These micro-optimizations hardly have any meaningful result on how Camel performs: bottlenecks are more often than not elsewhere in the code. 

Ultimately, the reduced readability ends up affecting the maintainers more - as we'll take longer to understand the code in 1, 2 ... 3 months from now. When you have to deal with such a large code base, balancing things like these become just as important.

davsclaus:
okay lets keep as-is to make maintaining code easier

PR #9797, opened 2023-04-03
static methods don't need final keyword, it's senseless

dk2k:
static methods don't need final keyword, it's senseless

### DIFF
--- a/components/camel-sql/src/generated/java/org/apache/camel/component/sql/stored/template/generated/TokenMgrError.java
+++ b/components/camel-sql/src/generated/java/org/apache/camel/component/sql/stored/template/generated/TokenMgrError.java
@@ -47,7 +47,7 @@ public class TokenMgrError extends Error
    * Replaces unprintable characters by their escaped (or unicode escaped)
    * equivalents in the given string
    */
-  protected static final String addEscapes(String str) {
+  protected static String addEscapes(String str) {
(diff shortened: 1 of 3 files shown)


components/camel-sql/src/generated/java/org/apache/camel/component/sql/stored/template/generated/TokenMgrError.java
```diff
@@ -47,7 +47,7 @@ public class TokenMgrError extends Error
    * Replaces unprintable characters by their escaped (or unicode escaped)
    * equivalents in the given string
    */
-  protected static final String addEscapes(String str) {
+  protected static String addEscapes(String str) {
```
essobedo, line 50: This class is generated so I'm afraid that this change won't have any effect

essobedo:
Please also use a different commit message, for something more clear

essobedo:
BTW, maybe it is not really applicable here, but telling that it is meanless, it is not completely true, please read this https://stackoverflow.com/a/1743748/1997376

orpiske (changes requested):
The commit message is not very clear, so requesting a change.

davsclaus (approved):

dk2k:
the topic seems to be controversial. I'm OK with keeping this PR unmerged

davsclaus:
merging anyway as generated class will re-gen

PR #9811, opened 2023-04-04
CAMEL-19234: Lookup beans by name+type first, and fallback to name-an…

davsclaus:
…d-convert, when resolving endpoint URI beans.

# Description

# Target

# Tracking


turing85 (approved):
The PR resolves the issue. I was able to cherry-pick this commit locally onto tag `4.0.0-M2`, build `camel-support` locally and verify that the corresponding integration tests in `quarkus-artemis` execute successfully (in JVM- and Native mode).

I left some comments, suggesting changes. The changes do not influence the behaviour but are only refactorings.

turing85:
Marker to get a cross reference: https://github.com/apache/camel-quarkus/issues/4725

davsclaus:
Thanks for all the suggestions, however the code is better to be as-is to be similar to the existing coding style, and make sure it keep working as before.

PR #9813, opened 2023-04-04
Changed Arrays comparison - was performed using ==, need Arrays.equals()

dk2k:
Changed Arrays comparison - was performed using ==, need Arrays.equals()


davsclaus:
This is a special optimization in camel internally

PR #9816, opened 2023-04-05
Removed empty List headerParts and the related loop

dk2k:
Removed empty List headerParts and the related loop. This loop does nothing.

### DIFF
--- a/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/SoapDataFormat.java
+++ b/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/SoapDataFormat.java
@@ -121,7 +121,6 @@ public class SoapDataFormat extends JaxbDataFormat {
             final Object inputObject, String soapAction,
             List<Object> headerElements) {
         List<Object> bodyParts = new ArrayList<>();
-        List<Object> headerParts = new ArrayList<>();
         bodyParts.add(inputObject);
 
         List<Object> bodyElements = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/SoapDataFormat.java
```diff
@@ -121,33 +121,18 @@ protected List<Object> createContentFromObject(
...
 
-        for (Object headerObj : headerParts) {
-            QName name = elementNameStrategy.findQNameForSoapActionOrType(soapAction, headerObj.getClass());
-            if (name == null) {
-                LOG.warn("Could not find QName for class {}", headerObj.getClass().getName());
-                continue;
-            } else {
-                JAXBElement<?> headerElem = getElement(headerObj, name);
-                if (null != headerElem) {
-                    headerElements.add(headerElem);
```
davsclaus, line 146: Headers is added to the incoming headerElements parameter in the method, so the old code looks correct

davsclaus (changes requested):

components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/SoapDataFormat.java
dk2k: headerParts is always an empty list. You refer to the code inside the loop, but this loop iterates through an empty list. Hence, the referred code is never reached. The entire loop can be removed

davsclaus (approved):

PR #9817, opened 2023-04-05
String.equalsTo("") replaced with String.isEmpty() in core and tooling

dk2k:
String.equalsTo("") replaced with String.isEmpty(). The latter is more efficient. Separate commit for core and tooling
In MavenDownloaderImpl combined two similar IF blocks and removed useless assignment

### DIFF
--- a/tooling/camel-tooling-maven/src/main/java/org/apache/camel/tooling/maven/MavenDownloaderImpl.java
+++ b/tooling/camel-tooling-maven/src/main/java/org/apache/camel/tooling/maven/MavenDownloaderImpl.java
@@ -731,10 +731,7 @@ public class MavenDownloaderImpl extends ServiceSupport implements MavenDownload
         // 3) ${user.home}/.m2/repository (if exists)
         // 4) /tmp/.m2/repository
         String localRepository = System.getProperty("maven.repo.local");
-        if (localRepository == null || "".equals(localRepository.trim())) {
-            localRepository = settings.getLocalRepository();
-        }
-        if (localRepository == null || "".equals(localRepository.trim())) {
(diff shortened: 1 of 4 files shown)


tooling/camel-tooling-maven/src/main/java/org/apache/camel/tooling/maven/MavenDownloaderImpl.java
```diff
@@ -731,10 +731,7 @@ Settings mavenConfiguration(
         // 3) ${user.home}/.m2/repository (if exists)
         // 4) /tmp/.m2/repository
         String localRepository = System.getProperty("maven.repo.local");
-        if (localRepository == null || "".equals(localRepository.trim())) {
-            localRepository = settings.getLocalRepository();
```
essobedo, line 735: Why remove this?

dk2k: The first step. Another IF with the same condition below. Both get combined into one block.
The second step. The assigned value isn't used before another value is assigned to localRepository => the firas assignment can be removed

essobedo: Sorry I don't get it, what about the use case where `settings.getLocalRepository()` returns a non-empty value?

dk2k: My bad. I understand now

essobedo (approved):
👍 LGTM, thx

davsclaus (approved):

PR #9825, opened 2023-04-06
fixed results of inspection "Single char arg in startsWith() and endsWith() can be substituted with charAt()"

dk2k:
fixed results of inspection "Single char arg in startsWith() and endsWith() can be substituted with charAt()". charAt() is more efficient
Added two helper methods into StringHelper in camel-util


oscerd (approved):

davsclaus:
Does the JDK not itself optimize for such use-cases, as its common to check starts with a single char.

orpiske:
> Does the JDK not itself optimize for such use-cases, as its common to check starts with a single char.

It probably does. AFAIK, there's a lot that could be in play here: from the JVM/compiler optimizing this, auto-vectorization, to how each platform handle the specific behaviors of the low-level operations that perform this change ... I'd probably leave this one as is.

orpiske (changes requested):
I'm not entirely sure if this is helpful. 

Maybe provide some benchmarks to support the change?

dk2k:
@orpiske I will provide it even if it's not in favour of the suggested code cganges ;)
https://gist.github.com/dk2k/5204c1a07d949647e277b7ceb03c266e
In the most cases native String startsWith() and endsWith() are faster
time for String.startsWith       49
time for StringHelper.startsWith 74
You can try it yourself.
So we should discard this PR, I believe.

orpiske:
I think we can close this, as there isn't really a performance problem.

PR #9826, opened 2023-04-06
Fixed results of inspection "redundant local variable".

dk2k:
Fixed results of inspection "redundant local variable". Inlinining of variables is the solution.


orpiske (changes requested):
Too many components have been changed. Please split this PR in core, catalog, components and dsl changes so it can be automatically tested.

PR #9827, opened 2023-04-07
CAMEL-19256: camel-jdbc - close stmt in the finnal block

zhfeng:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-jdbc/src/main/java/org/apache/camel/component/jdbc/JdbcProducer.java
+++ b/components/camel-jdbc/src/main/java/org/apache/camel/component/jdbc/JdbcProducer.java
@@ -31,7 +31,6 @@ import java.util.Map;
 import javax.sql.DataSource;
 
 import org.apache.camel.Exchange;
-import org.apache.camel.spi.Synchronization;
 import org.apache.camel.support.DefaultProducer;
 import org.apache.camel.support.PropertyBindingSupport;
 import org.apache.camel.support.SynchronizationAdapter;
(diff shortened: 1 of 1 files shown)


zhfeng:
The stmt and rs could be closed in `ResultSetIteratorCompletion` if the outputType is `JdbcOutputType.StreamList`.

oscerd (approved):

components/camel-jdbc/src/main/java/org/apache/camel/component/jdbc/JdbcProducer.java
```diff
@@ -194,21 +193,12 @@ private boolean doCreateAndExecuteSqlStatementWithHeaders(Exchange exchange, Str
 
     private boolean doCreateAndExecuteSqlStatement(Exchange exchange, String sql, Connection conn) throws Exception {
 
+        Statement stmt = null;
         ResultSet rs = null;
         boolean shouldCloseResources = true;
 
         try {
-            // We might need to leave it open to allow post-processing of the result set. This is why we
```
davsclaus, line 201: This is a bit wrong, in case you use JdbcOutputType.StreamList then we need to close it in the UoW, eg its when setResultSet returns false.

But I think we can keep the closing in the finally block if setResultSet returns true. Then its closed eager in finally, and then again in UoW done, but its already closed and isClosed will return true.

davsclaus (changes requested):

davsclaus:
Ah yeah its a bit of a mess as ResultSetIterator also has some close code, and some of the other methods does not seem to close PreparedStatement if you use StreamList as output type.

I guess we need to look deeper into this code, and ensure all that jdbc stuff is closed correctly in normal and streamlist mode (close in UoW).

components/camel-jdbc/src/main/java/org/apache/camel/component/jdbc/JdbcProducer.java
zhfeng: if `setResultSet` resturns `false`, it has added `exchange.getExchangeExtension().addOnCompletion(new ResultSetIteratorCompletion(iterator))`. So I think we don't need to close it in the UoW here.

zhfeng: Or it should be better to move these codes
```java
try {
...
shouldCloseResources = setResultSet(exchange, conn, rs);
...

} finally {
            if (shouldCloseResources) {
                closeQuietly(rs);
                closeQuietly(stmt);
            } else {
                @Override
                public void onDone(Exchange exchange) {
                    closeQuietly(rs);
                    closeQuietly(stmt);
                }
            });
           }
        }       

```

davsclaus: Yeah then the closing logic is at the same place and easier to ensure is correct

davsclaus (approved):

PR #9828, opened 2023-04-07
Optimization of loops with the only action of appending a string several times

dk2k:
Optimization of loops with the only action of appending a string several times. For this case String.repeat() method was introduced in java 11

### DIFF
--- a/catalog/camel-report-maven-plugin/src/main/java/org/apache/camel/maven/RouteCoverageMojo.java
+++ b/catalog/camel-report-maven-plugin/src/main/java/org/apache/camel/maven/RouteCoverageMojo.java
@@ -502,9 +502,7 @@ public class RouteCoverageMojo extends AbstractExecMojo {
     }
 
     private static String padString(int level) {
-        StringBuilder sb = new StringBuilder();
-        sb.append("  ".repeat(Math.max(0, level)));
-        return sb.toString();
+        return "  ".repeat(level);
     }
 
     private boolean matchFile(File file) {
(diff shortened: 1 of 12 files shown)


catalog/camel-report-maven-plugin/src/main/java/org/apache/camel/maven/RouteCoverageMojo.java
```diff
@@ -503,9 +503,7 @@ private static void gatherRouteCoverageSummary(
 
     private static String padString(int level) {
         StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < level; i++) {
-            sb.append("  ");
-        }
+        sb.append("  ".repeat(Math.max(0, level)));
```
davsclaus, line 506: Should it not just be .repeat(level). The level is not intended to be negative

dk2k: it should. And the StringBuilder isn't necessary after the loop gor removed. So it will be at least one more commit here.
It was just the automated substitution - to be on the safe side

davsclaus: Can you remove the Math.max code as repeat already check that the input must be a positive value

davsclaus (changes requested):

catalog/camel-report-maven-plugin/src/main/java/org/apache/camel/maven/RouteCoverageMojo.java
dk2k: fixed

davsclaus (approved):

PR #9831, opened 2023-04-08
Fixed results of inspection "redundant local variable" for tooling

dk2k:
Fixed results of inspection "redundant local variable". Inlining of variables is the solution. PR for tooling


tooling/maven/camel-package-maven-plugin/src/main/java/org/apache/camel/maven/packaging/EndpointSchemaGeneratorMojo.java
```diff
@@ -1051,9 +1051,8 @@ protected void findComponentClassProperties(
                             .filter(opt -> name.equals(opt.getName())).findAny();
                     if (prev.isPresent()) {
                         String prv = prev.get().getJavaType();
-                        String cur = fieldTypeName;
```
davsclaus, line 1054: using `cur` for current makes the code easier to read, so lets keep the code as-is

PR #9838, opened 2023-04-10
Fixed performance related issue "class types should be compared with == or !=,  and not with equals()". PR for core

dk2k:
Fixed performance related issue "class types should be compared with == or !=,  and not with equals()". PR for core

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/ComponentCustomizer.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/ComponentCustomizer.java
@@ -201,7 +201,7 @@ public interface ComponentCustomizer extends Ordered {
         }
 
         private BiPredicate<String, Component> condition() {
-            if (type.equals(Component.class)) {
+            if (type == Component.class) {
                 return this.condition != null
                         ? this.condition
                         : new BiPredicate<String, Component>() {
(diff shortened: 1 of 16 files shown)


orpiske (changes requested):
I am not entirely sure this is the case. Basically, this is the same thing done by `Object.equals` as is in the [OpenJDK source code](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/Object.java#L157-L159).

Please, can you provide some evidence ([using JMH](https://github.com/openjdk/jmh)) that there is a performance problem with the way we do this?

dk2k:
@orpiske I suspect some ambiguity here. It's Object.equals(Object) which I changed in this PR. And it Objects.equals(Object, Object) introduced in java 11 - I didn't change these calls. Oracle loves such questions in Java SE exams.
Dedicated new benchmark: https://gist.github.com/dk2k/505226619b28a693041c1197b427dea4

Sample timings:
time for Class.equals(Class)          52
time for Objects.equals(Class, Class) 11
time for Class == Class               8

time for Class.equals(Class)          90
time for Objects.equals(Class, Class) 9
time for Class == Class               8

time for Class.equals(Class)          68
time for Objects.equals(Class, Class) 9
time for Class == Class               6

time for Class.equals(Class)          89
time for Objects.equals(Class, Class) 9
time for Class == Class               7

So, I think the proposed changes are still relevant. Objects.equals() can be considered an option as well

orpiske:
Is this still relevant? I think we can close this one ...

davsclaus:
Yes I think we can close this. Also when using `==` then you may have classloading issues to consider when running on other runtimes like OSGi based and whatnot.

orpiske:
Alright, let's close this one.

PR #9851, opened 2023-04-12
JDK 8 support was dropped in the 3.15.x branch

benallard:
# Description

# Target

# Tracking

### DIFF
--- a/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
+++ b/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
@@ -1,7 +1,5 @@
 = What platforms are supported?
 
-- Camel 2.x supports JDK 8
-- Camel 3.x supports JDK 8 up to its 3.14.x branch
-- Camel 3.x supports JDK 11
-- Camel 4.x supports JDK 17
-
+- Java 8 Support until 3.14.x
+- Java 11 Support for all 3.x version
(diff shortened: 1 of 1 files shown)


docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
```diff
@@ -1,5 +1,7 @@
 = What platforms are supported?
 
 - Camel 2.x supports JDK 8
-- Camel 3.x supports JDK 8 (secondary) and 11 (primary)
+- Camel 3.x supports JDK 8 up to its 3.14.x branch
+- Camel 3.x supports JDK 11
```
davsclaus, line 5: 3.x supports JDK 17 from 3.18 onwards.

benallard: According to your table, it's 3.17, but 3.17 was not LTS.
What does that means, that you started testing with Java-17 back then? because I don't see any relevant change regarding java-17 in the pom.

benallard:
Should we better invert the table ?

```
- Java 8 Support until 3.14.x
- Java 11 Support for all 3.x version
- Java 17 Support starting with 3.17.0
```

davsclaus:
Yeah that is fine - whether a release is LTS or not does not matter in regards to when Camel supports a JDK version

PR #9906, opened 2023-04-20
Use jakarta API in ITs

Croway:
Generated DTOs will be generated without timestamp since swagger-codegen does not fully support jakarta yet.

Should we use openapi codegen maven plugin instead of swagger?

### DIFF
--- a/tooling/maven/camel-restdsl-openapi-plugin/src/main/java/org/apache/camel/maven/generator/openapi/AbstractGenerateMojo.java
+++ b/tooling/maven/camel-restdsl-openapi-plugin/src/main/java/org/apache/camel/maven/generator/openapi/AbstractGenerateMojo.java
@@ -126,7 +126,7 @@ abstract class AbstractGenerateMojo extends AbstractMojo {
     @Parameter
     String basePath;
 
-    @Parameter(defaultValue = "3.0.36")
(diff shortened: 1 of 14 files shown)


oscerd (approved):

oscerd:
I think we should move to openapi codegen

tooling/maven/camel-restdsl-openapi-plugin/src/main/java/org/apache/camel/maven/generator/openapi/AbstractGenerateMojo.java
```diff
@@ -214,13 +214,21 @@ void generateDto(final String language) throws MojoExecutionException {
...
+         * https://github.com/swagger-api/swagger-codegen-generators/pull/1131
+         */
+        configOptions.put("hideGenerationTimestamp", "true");
+        elements.add(new MojoExecutor.Element(
+                "configOptions", configOptions.entrySet().stream()
+                    .map(e -> new MojoExecutor.Element(e.getKey(), e.getValue()))
+                    .toArray(MojoExecutor.Element[]::new)));
+
+
+        System.out.println(elements.toArray(new MojoExecutor.Element[0]));
```
davsclaus, line 231: Should this system out be there

davsclaus (approved):

PR #9919, opened 2023-04-24
Backport to 3.18 the fix for CAMEL-19294

orpiske:


oscerd (changes requested):
I think this is wrong

orpiske:
WTH, let me check what happened on my backport process.

orpiske:
> I think this is wrong

It should be fine now. I used an wrong base for the PR command: 🤦

oscerd:
LGTM

PR #9926, opened 2023-04-25
CAMEL-19292: remove remaining logic tied to the LightweightContext

orpiske:


davsclaus:
Well this was a bit different than the LWCamelContext. This was a special mode to boot Camel where after startup it will null some model classes and whatnot, that allows JVM to GC and also clear some internal maps.

However this mode is not often in use, so I am okay removing it. We can come back later if we can do a flag with another name that could do something on the same page, if needed.

This is general Camel, as otherwise some of these optimizations may only be in CEQ.

davsclaus (approved):

PR #8896, opened 2022-12-13
Adding HTTPs support to AS2 component

shikhar97gupta:
Trying to add SSLContext option to AS2Configuration.

I added a new SSLContext field to AS2Configuration class and got "Unable to create mojo: Empty doc for option: sslContext, parent options: <null>" error while compiling. Following the contribution page, I have already added javadoc comment before the setter method (didn't see it before any getter method of other fields so skipped it for getter) and @UriParam for the field declaration.

(I had written code for adding the SSLContext configuration at component and endpoint level, but keeping the code minimal here. Also changed SSLContext to a String. Getting the same error though)

PR #9939, opened 2023-04-26
CAMEL-17946 Introducing HTTPS support for AS2 component

dk2k:
# Description

Introducing HTTPS for AS2 component.
I used this PR as a basis https://github.com/apache/camel/pull/8896 
The current PR is compatible with main and 4.x branches. Later I will try handling backporting to 3.x branch

@McIntozh your comment would be much appreciated. 
As for HTTPS support for AS2 for older versions. I used my middleware proxy application as a workaround - it accepts HTTP connections and sends them via HTTPS to remote server, works fine with mendelson server. The application https://github.com/dk2k/netty-https-proxy

# Target

`main` branch

# Tracking

Related to https://issues.apache.org/jira/browse/CAMEL-17946

# Testing

Updated the existing tests and added the test class MendelsonSslEndpointIT. In this class I send a message to HTTPS version of 3rd party Mendelson server - please see port 8444 in the settings. I do think it's good to provide this test - it documents the usage of AS2 client component with SSL. The test will be committed with annotation @Disabled due to it is strongly dependent on 3rd party remote server. The test in MendelsonSslEndpointIT works fine, I can see my messages in the web ui of Mendelson http://testas2.mendelson-e-c.com:8080/webas2/


components/camel-as2/camel-as2-component/src/test/resources/mendelson/key4.cer
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
orpiske, line 1: This file should be provided via parameters instead of being kept in the code base (i.e.: `mvn -Das2.http.key=/path/to/key`).

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonSslEndpointIT.java
```diff
@@ -0,0 +1,132 @@
...
+
+        AS2Endpoint endpoint = (AS2Endpoint) as2Component
+                .createEndpoint("as2://client/send?targetHostName=testas2.mendelson-e-c.com"
+                                + "&targetPortNumber=8444&inBody=ediMessage&requestUri=/as2/HttpReceiver" +
+                                "&ediMessageContentType=application/edifact" +
+                                "&signingAlgorithm=SHA3_256WITHRSA");
+
+        Assertions.assertEquals("mycompanyAS2", endpoint.getAs2From());
+        Assertions.assertEquals("mendelsontestAS2", endpoint.getAs2To());
+        Assertions.assertEquals("dk2kEdi", endpoint.getFrom());
```
orpiske, line 123: These values taken for the assertion expectations should be configurable so that they are not bound to this specific AS2 instance.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        try {
+            return (PrivateKey) ks.getKey(
+                    ks.aliases().nextElement(),
+                    password.toCharArray());
+        } catch (KeyStoreException e) {
+            e.printStackTrace();
+        } catch (UnrecoverableKeyException e) {
+            e.printStackTrace();
+        }
```
orpiske, line 168: I believe all these exceptions being caught in this block should be logged.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
```diff
@@ -0,0 +1,183 @@
...
+        throw new IllegalStateException("about to return null");
```
orpiske, line 169: Maybe provide a better error message here.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
```diff
@@ -0,0 +1,183 @@
...
+        } catch (NoSuchAlgorithmException e) {
+            String errMsg = "Error while trying to load certificate to the keyload. Requested algorithm isn't found.  " + e;
+            System.out.println(errMsg);
+        } catch (CertificateException e) {
+            String errMsg = "Error while trying to load certificate to the keyload. There is a certificate problem.  " + e;
+            System.out.println(errMsg);
+        } catch (InvalidKeySpecException e) {
+            String errMsg = "Can not init private key store  " + e;
+            System.out.println(errMsg);
+        }
```
orpiske, line 109: I believe all the messages being sent to the `stdout` here should be handled by a logger.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.net.ssl.SSLContext;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.http.conn.ssl.TrustAllStrategy;
+import org.apache.http.ssl.SSLContexts;
+
+/**
+ * That's a utility class for preparing Mendelson-specific certificate chain, private key, ssl context
+ */
+public class MendelsonCertLoader {
```
orpiske, line 29: I am not entirely sure we want this file as it is. This is tied to an specific instance of a service. Instead, I think it would be better to make it generic and configurable, so that anyone willing to test with a different instance could do so.

components/camel-as2/camel-as2-component/pom.xml
```diff
@@ -64,6 +64,12 @@
             <artifactId>camel-jetty</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.10.0</version>
+            <scope>test</scope>
+        </dependency>
```
orpiske, line 72: This likely needs to be correctly aligned w/ the `commons-io` version we use.

orpiske (changes requested):
Hi, thanks for your contribution. I have added some preliminary comments. 

In general: the tests should be made configurable and we should not have anything specific to a certain instance on the code base.

components/camel-as2/camel-as2-component/src/test/resources/mendelson/key4.cer
orpiske: That's a bad practice. Other reviewer may have missed.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonSslEndpointIT.java
dk2k: The entire idea of this test class is to bound it to a public AS2 service. It contains "Mendelson" in the class name

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
dk2k: fixed

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
dk2k: fixed

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
dk2k: I'm not sure that the requested changes are in the scope of the ticket CAMEL-17946. Maybe we need another ticket like "Provide versatile testing harness for AS2 component".
Let me explain the situation. All the Mendelson-related classes are in the test folder, they aren't part of AS2 JAR artifact. My idea was to provide a basic mechanism so that interested persons can verify my results for HTTPS AS2 connection - scientific approach. I din't aim to provide a versatile mechanism. Mendelson is almost the only public resource for testing, it was mentioned in https://github.com/apache/camel/pull/8896 and not by me. It doesn't bring something proprietary into the code base.
Important facts about Mendelson resource. It provides concrete type of private key and keystore. And its certificate chain is degenerated - consists of the only certificate. Furthermore, the same certificate chain is used for encryption and signing.
The only Mendelson test method is @Disabled by default.
It is important to test AS2 connection to non-localhost server due to it's the only way to test the behaviour of the new config param HostnameVerifier.
One other PRO: it was difficult to construct SSLContext from scratch for me. If I could choose, I would prefer to have real-life code examples in the code base.

orpiske: What I am considering here is: if this is disabled, it won't be tested often. If it won't be tested often, the code will rot and it will be upon us (maintainers/committers) to figure out - several months in the future - if it's still relevant or not (and, as things change, certificates get expired, etc) it's one hell of a pain to update and fix things on these obscure technologies. 

 So, by making it generic and configurable ... users themselves can handle this and put more burden on the maintainers.
 
 I am -1 on this, but let's see what others have to say.

orpiske: Since there was no feedback, let me provide more clarity: what I am asking here is to implement something like this: https://github.com/apache/camel/blob/main/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAuthManualTest.java

That is a manual test for connecting to an authenticated Kafka instance. I wrote to try it verify a problem an specific vendor of a managed Kafka service was having. As you can see in the linked test, there's no reference to any specific instance. I'd like this code to apply the same pattern.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonSslEndpointIT.java
orpiske: See my notes [here](https://github.com/apache/camel/pull/9939#discussion_r1180040542).

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
dk2k: I still hope to get more comments from other maintainers and contributors, I'm not in a rush. I think we need to calm down and don't take this PR personal.

I can agree that the test class's name for mendelson should be appended with "ManualTest" and not "IT" like now. But I don't understand why I should deliver a semi-prepared test class with a dependency on env variables instead of fully operational and friendly self-sustainable test class which matches all its requirements. Why a person would need to create env vars on his/her machine? They may intersect with existing vars, he/she may forget to remove them after all. That's an invasive approach.

I have another example for ManualTest class:  https://github.com/apache/camel/blob/main/components/camel-cometd/src/test/java/org/apache/camel/component/cometd/CometdProducerConsumerInteractiveAuthenticatedManualTest.java
Lines 73-74 were last edited by you (I don't say that you were the initial author). And in these lines the test class reads a keystore file from a similar resource folder jsse palced in test section of the sources. The test class CometdProducerConsumerInteractiveAuthenticatedManualTest isn't environment aware. All the config variables are part of the test class. Why is it OK to have it like this?

davsclaus:
Thanks for working on this. I dont have time to go deep in the review until its more ready. At that time, it may be easier to create a new clean PR if this one ends up with too many comments, changes etc.

orpiske:
> Thanks for working on this. I dont have time to go deep in the review until its more ready. At that time, it may be easier to create a new clean PR if this one ends up with too many comments, changes etc.

I think it's mostly OK, but I think we need to avoid keeping data related to specific service instances on the internet (even more so for things that we don't have deep knowledge about). These things tend to rot very quickly - and we are still cleaning up stuff like this that was added in the past (!!!!). Which is why I have a -1 on this.

So, as far as I am concerned, all it needs is to remove the service-specific bits and it should be ready to go.

dk2k:
@davsclaus sounds a bit like catch 22 because I won't be able to make it more ready without additional comments. For now the logic delivered in the JAR artifact is complete and we discuss with @orpiske the aspects of integration tests

dk2k:
@orpiske I agree that 3rd party stuff can rot and already stated that in the comment above the Mendelson test class. Please keep in mind that :

 * Unless required by applicable law or agreed to in writing, software
 * distributed under the License is distributed on an "AS IS" BASIS,
 * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 * See the License for the specific language governing permissions and
 * limitations under the License.
 
 This comment is in the head of each source file ;)
 By the way, should it be "expressed" instead of express?

davsclaus:
Yeah we cannot have an automated test connects to some commercial company on the internet.
The test needs to have its hostname / port etc externally configured and turned off by default.

And ideally we should have wiremock tests with captured network communication so the test could run in isolation.
Or if there are some docker images we can run via test-containers to run the test in isolation.

So would it be possible for you to externalize that test case, eg make it read from src/test/resources/test-server.properties (or some similar name). And leave it with TODO insert server name here. And we can have that Mendelson mentioned in the properties file but turned off. Then end users can choose to turn that on and use it if they desire.

components/camel-as2/camel-as2-component/src/test/java/org/apache/camel/component/as2/MendelsonCertLoader.java
oscerd: What happen if the Mendelson some day will not work anymore? We need a way to be not dependant on this kind of 3rd party stuff. We are trying to do this in multiple components, by adding containerized services to run when needed. This is really not in line with what we are doing. If this will be disabled for a long time, it will be something maintainers will have to deal with in the future. So I would prefer a different solution, so anyone could maintain this piece of software.

dk2k: I'm not sure that disabled by default test can be considered a dependency. It's just good manners: showing a way a person can establish HTTPS AS2 connection to non-localhost server.

If Mendelson goes down, just remove two classes with Mendelson in the name, the test/resource folder mendelson and the file test-server.properties. I will be able to do that myself. No hurt feelings on my side then :)

I think dockerization of AS2 server is beyond the scope of the ticket CAMEL-17946
Besides, what exactly are you going to dockerize? Camel AS2 server component? Both Camel AS2 server and client can have issues which will allow them to communicate, but not to other AS2-compliant server (what if the sources of both client and server suffer from the same issue?). Another point: it's important to test HTTPS connection to non-localhost server - test the new HostnameVerifier field. Will docker allow that?

I'm not sure you have seen my latest commit with moving Mendelson connection settings to a property file as per @davsclaus 's comment. Now a person needs to turn on 2 "knobs" to connect to Mendelson:
- remove/comment @Disabled annotation
- explicitly set Mendelson hostname in the properties.

Please let me know if the latest commit mitigates or even eliminates your concerns. If it dosen't, I will remove all Mendelson-related stuff.

davsclaus (approved):

davsclaus:
@orpiske LGTM

orpiske (approved):
There's a few issues, but LGTM. I'll fix those later.

orpiske:
Let's merge and have some tests running overnight.

PR #9968, opened 2023-04-29
CAMEL-19058: rework the Message to avoid hitting the type-check scalability issue

orpiske:
This introduces a trait that works similarly to the internal properties one in the Exchange so that we carry certain details about the message without relying on type specifications.

Among other things this:

- avoids unnecessarily recreating the DefaultMessage multiple times
- moved the data type aware checks out of the hot path
- reworks the isTransactedRedelivered to avoid costlier checks for specific message types

This superseds the proposed changes on #9432.

Additionally, this implements the same pattern of using an array for
internal properties as we already do on the Exchange. Subsequently, my
idea is to abstract this complex array manipulation in an specific class
so we can reuse code and simplify maintenance.

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.support;
 
+import java.util.Arrays;
+
 import org.apache.camel.CamelContext;
 import org.apache.camel.CamelContextAware;
 import org.apache.camel.Exchange;
@@ -24,6 +26,7 @@ import org.apache.camel.Message;
 import org.apache.camel.TypeConverter;
 import org.apache.camel.spi.DataType;
 import org.apache.camel.spi.DataTypeAware;
+import org.apache.camel.trait.message.MessageTrait;
 
 /**
  * A base class for implementation inheritance providing the core {@link Message} body handling features but letting the
@@ -33,19 +36,22 @@ import org.apache.camel.spi.DataTypeAware;
  * from {@link DefaultMessage}
  */
 public abstract class MessageSupport implements Message, CamelContextAware, DataTypeAware {
+    private static final int NUM_TRAITS = MessageTrait.values().length;
+
     CamelContext camelContext;
     TypeConverter typeConverter;
     private Exchange exchange;
     private Object body;
     private String messageId;
     private long messageTimestamp;
-    private DataType dataType;
+
+    private final Object[] traits = new Object[NUM_TRAITS];
 
     @Override
     public void reset() {
         body = null;
         messageId = null;
-        dataType = null;
+        Arrays.fill(traits, null);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


oscerd (approved):

core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
```diff
@@ -306,4 +310,20 @@ protected String createMessageId() {
...
+
+    @Override
+    public Object getPayloadForTrait(MessageTrait trait) {
+        return traits[trait.ordinal()];
+    }
+
+    @Override
+    public void setPayloadForTrait(MessageTrait trait, Object object) {
+        traits[trait.ordinal()] = object;
+    }
```
orpiske, line 328: This part here: if accepted, then I'll do create a separate class to abstract this. We already do a similar thing when handling the `internalProperties` on the [AbstractExchange](https://github.com/apache/camel/blob/main/core/camel-support/src/main/java/org/apache/camel/support/AbstractExchange.java#L252-L276).

davsclaus (approved):

davsclaus:
I wonder if we should move some of those trait out of the root package to avoid having to much stuff there (there is already too much) - as those are mostly for internal optimization use, then they better fit in .spi or in a new .trait package.

orpiske:
> I wonder if we should move some of those trait out of the root package to avoid having to much stuff there (there is already too much) - as those are mostly for internal optimization use, then they better fit in .spi or in a new .trait package.

I like this idea. Let me explore it in one shot with this PR. 

Also, I'm marking this one as a draft for a while. I found a bug in the JMS code and I'll need to review it.

Thanks!

orpiske:
Remarking as ready to review after fixing a bug in JMS changes. Also moved the traits to a separate `trait` package.

davsclaus (approved):

essobedo:
By chance do you have any benchmarks to share?

orpiske:
> By chance do you have any benchmarks to share?

Yes I do. I am preparing a blog post with the details, methodology, etc. But here's a preview for the disruptor (in just few of the many other scenarios):

Obs.: the notation is inverted and the baseline is the code is the code with this patch and the test code is the test with 4.0.0-M1.

*25 consumers + 25 producers*:
```
GEOMETRIC MEAN (RATE):
Test: 640260.6763745428 | Baseline: 712828.9356009845
Delta: -72568.25922644173
The geometric mean rate was lower for the test than for the baseline
```


*10 consumers + 9 producers*:
```
GEOMETRIC MEAN (RATE):
Test: 767544.7450252476 | Baseline: 841304.9248985515
Delta: -73760.17987330398
The geometric mean rate was lower for the test than for the baseline
```

*7 consumers + 3 producers*:
```
GEOMETRIC MEAN (RATE):
Test: 486716.1190278101 | Baseline: 610061.3909864418
Delta: -123345.27195863164
The geometric mean rate was lower for the test than for the baseline
```

All in all, the new code tends to be faster in around 95% of the cases. The results from JMH also indicate the code being faster (i.e.; having a higher score in ops/ms - IIRC), although I don't have them easily at hand right now.

PR #9970, opened 2023-04-30
CAMEL-19304: camel-jpa implement paging

jacekszymanski:
# Description
Camel JPA endpoint supports option maximumResults, but this is not enough to implement paging. Also, there is no way to specify offset apart from nativeQuery.

maximumResults is static on the endpoint; it may be specified dynamically only using toD, but this means creating dynamic endpoints.

This PR addresses the above:

- adds support for firstResult
- both maximumResults and firstResult may be specified in message headers, these override values specified on the endpoint, if any

# Target

# Tracking


components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaProducer.java
```diff
@@ -202,10 +203,18 @@ protected void processQuery(Exchange exchange, EntityManager entityManager) {
 
     @SuppressWarnings("unchecked")
     private void configureParameters(Query query, Exchange exchange) {
-        int maxResults = getEndpoint().getMaximumResults();
+        final int maxResults = Objects.requireNonNullElse(
+                exchange.getIn().getHeader(JpaConstants.JPA_MAXIMUM_RESULTS, Integer.class),
```
davsclaus, line 207: You can use the getHeader that takes a default value to void the Objects method. This is the standard we use in Camel.

components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaProducer.java
```diff
@@ -202,10 +203,18 @@ protected void processQuery(Exchange exchange, EntityManager entityManager) {
...
+                getEndpoint().getMaximumResults());
         if (maxResults > 0) {
             query.setMaxResults(maxResults);
         }
+        final int firstResult = Objects.requireNonNullElse(
+                exchange.getIn().getHeader(JpaConstants.JPA_FIRST_RESULT, Integer.class),
```
davsclaus, line 213: Same as before

jacekszymanski:
Thanks, corrected.

davsclaus (approved):

oscerd (approved):

Croway (approved):

PR #9972, opened 2023-05-01
CAMEL-19295 - Concurrency issues with dynamicMap in AbstractDynamicRegistry

rhuan080:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractDynamicRegistry.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractDynamicRegistry.java
@@ -16,7 +16,6 @@
  */
 package org.apache.camel.impl.engine;
 
-import java.util.AbstractMap;
 import java.util.AbstractSet;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -40,7 +39,7 @@ import org.apache.camel.support.service.ServiceHelper;
  * Base implementation for {@link org.apache.camel.spi.EndpointRegistry},
  * {@link org.apache.camel.spi.TransformerRegistry}, and {@link org.apache.camel.spi.ValidatorRegistry}.
  */
-public class AbstractDynamicRegistry<K, V> extends AbstractMap<K, V> implements StaticService {
+public class AbstractDynamicRegistry<K, V> extends ConcurrentHashMap<K, V> implements StaticService {
 
     protected final CamelContext context;
     protected final RouteController routeController;
@@ -208,6 +207,11 @@ public class AbstractDynamicRegistry<K, V> extends AbstractMap<K, V> implements 
             }
         }
         return Collections.unmodifiableCollection(answer);
+    }
+
+    @Override
+    public Collection<V> values() {
+        return Collections.unmodifiableCollection(super.values());
     }
 
     public Map<String, V> getReadOnlyMap() {
--- a/core/camel-core/src/test/java/org/apache/camel/impl/DefaultEndpointRegistryTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/DefaultEndpointRegistryTest.java
@@ -16,10 +16,19 @@
  */
 package org.apache.camel.impl;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import org.apache.camel.ProducerTemplate;
+import org.apache.camel.ServiceStatus;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.impl.engine.DefaultEndpointRegistry;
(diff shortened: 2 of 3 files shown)


rhuan080:
Hi, I sent this PR and I have seen the SimpleLRUCache [1] is not thread-safe, and we don`t have a ConcurrentLinkedHashMap or something like that. Thus, I think we need to review this class to turn it thread-safe.

[1] - https://github.com/apache/camel/blob/244b828a3718b4d5a6faf9e1b0585f9f34d580e8/core/camel-support/src/main/java/org/apache/camel/support/DefaultLRUCacheFactory.java#L141

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractDynamicRegistry.java
```diff
@@ -40,7 +39,7 @@
  * Base implementation for {@link org.apache.camel.spi.EndpointRegistry},
  * {@link org.apache.camel.spi.TransformerRegistry}, and {@link org.apache.camel.spi.ValidatorRegistry}.
  */
-public class AbstractDynamicRegistry<K, V> extends AbstractMap<K, V> implements StaticService {
+public class AbstractDynamicRegistry<K, V> extends ConcurrentHashMap<K, V> implements StaticService {
```
orpiske, line 42: TBH, I kinda dislike the fact that it extends a core Java class. I wonder if we should, maybe, re-evaluate how we do this to avoid tightly coupling here.

rhuan080: I agree with you. This PR is solving an issue. I think we can fix this issue with this code and open an improvement issue to review this point.

orpiske: To be clear: I am not saying you should do this now. I'm just thinking out loud ...

core/camel-core/src/test/java/org/apache/camel/impl/DefaultEndpointRegistryTest.java
```diff
@@ -68,4 +76,50 @@ public void configure() throws Exception {
...
+                    producerTemplate.requestBody("controlbus:route?routeId=route3&action=ACTION_STATUS&loggingLevel=off", null,
+                            ServiceStatus.class);
+                    producerTemplate.requestBody("controlbus:route?routeId=route4&action=ACTION_STATUS&loggingLevel=off", null,
+                            ServiceStatus.class);
+                    producerTemplate.requestBody("controlbus:route?routeId=route5&action=ACTION_STATUS&loggingLevel=off", null,
+                            ServiceStatus.class);
+
+                    allThreadCompletionSemaphore.countDown();
+
+                });
```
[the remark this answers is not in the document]
rhuan080: Hi @orpiske, 

Yes, I was able to reproduce the OOM using this test, every time I run it. It is based on the code shared in the issue.

davsclaus (approved):

davsclaus:
Yes we should look at SimpleLRUCache before 4.0, can you create a JIRA ticket

orpiske:
> Yes we should look at SimpleLRUCache before 4.0, can you create a JIRA ticket

I'm wondering if it would be possible to unify all these caches and pools that we have. For instance, we have the `ServicePool` which has overlapping features with the `LRUCache`.

davsclaus:
Okay had to revert as this causes many other tests to fail
https://github.com/apache/camel/pull/10155

@rhuan080 thanks for the PR, but when changing in core, then its best to run the tests in camel-core to ensure its not affected other parts

davsclaus:
> > Yes we should look at SimpleLRUCache before 4.0, can you create a JIRA ticket
> 
> I'm wondering if it would be possible to unify all these caches and pools that we have. For instance, we have the `ServicePool` which has overlapping features with the `LRUCache`.

Yeah but ServicePool is a bit complex and special for camel's producer/consumer caches.
Not really for end users. But if it can be made simpler then that would be good - its a bit tricky to maintain.

davsclaus:
We may just drop SimpleLRUCache and use camel-caffeine-lrucache by default, and maybe move it into root/core folder. This cache is concurrent and LRU based. However it has a slight overhead on startup, that Simple does not have.

@otavioprado something you may have fun with performance testing in your lab.

davsclaus:
I did a quick experiment with fully sync LRUCache and this unit test
https://github.com/apache/camel/pull/10157

It all works, but since its fully sync it may have an overhead.
To use caffeine, you can even today, just drop camel-caffeine-lrucache in the classpath and Camel uses this instead.
So end users can choose.

And we could make the simple just synchronized (but maybe slower).

@otavioprado you are welcome to test in lab

orpiske:
@davsclaus that's a good one to test. I'll write a test for it and will do it next week. Let's see how it goes!

PR #9982, opened 2023-05-03
CAMEL-19058: added a new wrapper for the enum array pattern used throughout the code base

orpiske:

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
@@ -1,0 +1,174 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.camel.support;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * This class provides an array-based structure used to store payloads associated with enums. This is
+ * used in the hot path of the core code to allow access to those payloads with constant time and low
+ * memory overhead.
+ *
+ * This data-structure is meant for internal usage of the Camel Core and is not meant for users.
+ * @param <T>
+ */
+public final class EnumArray<T extends Enum<?>> {
+    private final Object[] internalArray;
+    private final T[] values;
+
+    /**
+     * Creates a new EnumArray with a fixed size determined by the length of values
+     * @param values the Enum values (as in Enum.values())
+     */
+    public EnumArray(T[] values) {
+        this.internalArray = new Object[values.length];
+        this.values = values;
+    }
+
+    /**
+     * Whether this arrau contains a payload (value) for the given entry
+     * @param entry the entry to check for
+     * @return true if there is a payload or false otherwise
+     */
+    public boolean contains(T entry) {
+        Object payload = internalArray[entry.ordinal()];
+
+        return payload != null;
+    }
+
+    /**
+     * Gets the payload for the given entry
+     * @param entry the entry to get the payload (must not be null)
+     * @return the payload or false otherwise
+     */
+    public Object get(T entry) {
+        Objects.requireNonNull(entry, "The entry for a enum array cannot be null");
+
+        return internalArray[entry.ordinal()];
+    }
+
+    /**
+     * Gets the payload for the given entry
+     * @param entry the entry to get the payload (must not be null)
+     * @param clazz the class to cast the payload to
+     * @return the payload or false otherwise
+     * @param <K> the payload type
+     */
+    public <K> K get(T entry, Class<K> clazz) {
+        Object tmp = get(entry);
+
+        if (tmp != null) {
+            return clazz.cast(tmp);
+        }
+
+        return null;
+    }
+
+    /**
+     * Computes a new value for the entry if it's present on this instance
+     * @param entry the entry to compute the value for
+     * @param mappingFunction the mapping function that will provide the new value. It takes as a parameter the old value.
+     * @return the new value that was computed for this entry
+     */
+    public Object computeIfPresent(T entry, Function<Object, Object> mappingFunction) {
+        Object payload = get(entry);
+
+        if (payload != null) {
+            payload = mappingFunction.apply(payload);
+
+            set(entry, payload);
+        }
+
+        return payload;
+    }
+
+    /**
+     * Sets the payload for the given entry
+     * @param entry the entry to set the payload
+     * @param payload the payload
+     */
+    public void set(T entry, Object payload) {
+        internalArray[entry.ordinal()] = payload;
+    }
+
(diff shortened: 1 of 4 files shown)


core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * This class provides an array-based structure used to store payloads associated with enums. This is
+ * used in the hot path of the core code to allow access to those payloads with constant time and low
+ * memory overhead.
+ *
+ * This data-structure is meant for internal usage of the Camel Core and is not meant for users.
+ * @param <T>
+ */
+public final class EnumArray<T extends Enum<?>> {
```
essobedo, line 34:
```suggestion
public final class EnumArray<T extends Enum<T>> {
```

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+     * Creates a new EnumArray with a fixed size determined by the length of values
+     * @param values the Enum values (as in Enum.values())
+     */
+    public EnumArray(T[] values) {
+        this.internalArray = new Object[values.length];
+        this.values = values;
+    }
+
+    /**
+     * Whether this arrau contains a payload (value) for the given entry
```
essobedo, line 48:
```suggestion
     * Whether this array contains a payload (value) for the given entry
```

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
essobedo: Why not use `EnumMap` instead?

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public final class EnumArray<T extends Enum<?>> {
+    private final Object[] internalArray;
+    private final T[] values;
+
+    /**
+     * Creates a new EnumArray with a fixed size determined by the length of values
+     * @param values the Enum values (as in Enum.values())
+     */
+    public EnumArray(T[] values) {
```
essobedo, line 42:
```suggestion
    public EnumArray(Class<T> keyType) {
```

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+        this.internalArray = new Object[values.length];
+        this.values = values;
```
essobedo, line 44:
```suggestion
        this.values = keyType.getEnumConstants();
        this.internalArray = new Object[values.length];
```

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Whether this arrau contains a payload (value) for the given entry
+     * @param entry the entry to check for
+     * @return true if there is a payload or false otherwise
+     */
+    public boolean contains(T entry) {
+        Object payload = internalArray[entry.ordinal()];
+
+        return payload != null;
```
essobedo, line 55:
```suggestion
        return internalArray[entry.ordinal()] != null;
```

Let's inline to avoid adding a local variable

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Gets the payload for the given entry
+     * @param entry the entry to get the payload (must not be null)
+     * @return the payload or false otherwise
+     */
+    public Object get(T entry) {
+        Objects.requireNonNull(entry, "The entry for a enum array cannot be null");
+
+        return internalArray[entry.ordinal()];
```
essobedo, line 66:
```suggestion
        return internalArray[Objects.requireNonNull(entry, "The entry for a enum array cannot be null").ordinal()];
```

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param <K> the payload type
+     */
+    public <K> K get(T entry, Class<K> clazz) {
+        Object tmp = get(entry);
+
+        if (tmp != null) {
+            return clazz.cast(tmp);
+        }
+
+        return null;
```
essobedo, line 83:
```suggestion
        return clazz.cast(get(entry));
```
```

File: core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java [hunk 8]
```diff
@@ -0,0 +1,174 @@
...
+        return payload;
+    }
+
+    /**
+     * Sets the payload for the given entry
+     * @param entry the entry to set the payload
+     * @param payload the payload
+     */
+    public void set(T entry, Object payload) {
+        internalArray[entry.ordinal()] = payload;
```
essobedo, line 110:
```suggestion
        internalArray[Objects.requireNonNull(entry, "The entry for a enum array cannot be null").ordinal()] = payload;
```
To be homogenous, we should check if it is null here too

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
```diff
@@ -0,0 +1,174 @@
...
+        return payload != null;
+    }
+
+    /**
+     * Gets the payload for the given entry
+     * @param entry the entry to get the payload (must not be null)
+     * @return the payload or false otherwise
+     */
+    public Object get(T entry) {
+        Objects.requireNonNull(entry, "The entry for a enum array cannot be null");
```
davsclaus, line 64: This just adds unnessasary overhead on a hot path - the old code did not need this check

core/camel-support/src/main/java/org/apache/camel/support/EnumArray.java
orpiske: @essobedo 🤔 I like your idea. Let me give it a try and see if we can skip the ssc issue with it and it may be a better implementation.

orpiske:
@essobedo @davsclaus thanks for the great feedback folks!!! I kinda like Nicolas idea of using an EnumMap instead of a custom implementation. Let me go back to the drawing board, run some tests and try that instead. I'll mark this one as draft.

davsclaus:
-1  for enum map

A fixed Object[] array are much smaller - and the original optimization was for avoding Map and other types that are large, and take up extra memory per message/exchange.

And the lookup by ordinal value is a lot faster than a Map lookup by a hashed key.

orpiske:
@davsclaus yeah, there can be some drawbacks. It may also suffer from the type check issue (which is why I created this custom implementation) ... but it's a small change to check. Let me take a quick look and see how it goes.

essobedo:
> -1 for enum map
> 
> A fixed Object[] array are much smaller - and the original optimization was for avoding Map and other types that are large, and take up extra memory per message/exchange.
> 
> And the lookup by ordinal value is a lot faster than a Map lookup by key.

Internally an `EnumMap` uses an array, in fact it does more or less the same as this custom class

essobedo:
I quote the Javadoc https://docs.oracle.com/javase/8/docs/api/java/util/EnumMap.html:

> Enum maps are represented internally as arrays. This representation is extremely compact and efficient.

orpiske:
Folks, so, I tested using an `EnumMap` and it does not trigger the type check issue. The performance is roughly the same for both, and the internal implementation of `EnumMap` is pretty much the same one as we have here (with a much better interface). 

@davsclaus @essobedo I am not strongly in favor of any solution, but I feel the `EnumMap` one might be a better choice. 

Even if there is some performance impact, I think we can offset that by fixing other parts of the code that are still being affected by the JDK issue.

essobedo:
Without very good reasons, I personally tend to prefer solutions based on JDK classes as they can always be improved in future versions of the JDK but if you really prefer having a fork of it, it is fine with me.

orpiske:
Folks, I am closing this one. I believe we all converged on #9984 (EnumMap one) as the solution. Thanks for the reviews and comments!!!

PR #9984, opened 2023-05-03
CAMEL-19058: wrap some internal properties in an EnumMap

orpiske:
This should make it simpler to abstract their complexity while maintaining reasonable performance (as they are used in many parts within the hot path)


orpiske:
This is the alternative implementation for #9982

essobedo (approved):

davsclaus (approved):

davsclaus:
Yeah JDK classes are good to prefer to use when they fit the purpose - map optimized internally with quick object array, and lookup by ordinal fixed number (like the old code)

PR #9986, opened 2023-05-03
CAMEL-19102: cache the producer and endpoint locally to avoid hitting the type check scalability issue

orpiske:
This reworks the patch originally added on  907a9d0c08d40c5308bb440ec9ebaaa4d7544dc4 so that there is no side effect when handling singleton endpoints


oscerd (approved):

orpiske:
Folks, I am marking this one as a draft. I am thinking about linking a few related PRs, so it may be easier to look at the whole set of changes together.

orpiske:
I am back at looking at this one. The patch is (likely) fine, but I want to review the results one more time.

PR #9999, opened 2023-05-05
CAMEL-17495 Added support of media type "application/xml" for AS2 component

dk2k:
Introduced content-type application/xml for AS2.
Related to https://issues.apache.org/jira/browse/CAMEL-17495
Request for changes for "text/xml" was mentioned in the dialog in jira, but I can't see this type in RFC https://www.ietf.org/rfc/rfc4130.txt. So I didn't add support for "text/xml".

The class ApplicationEDIEntity was renamed to ApplicationEntity because now the common part of content type is "application/" after I added the type "application/xml". ApplicationEntity is parent for ApplicationXMLEntity among EDI specific classes.

Added a test for sender and for receiver. Tested against Mendelson, these changes in the property file aren't included.


davsclaus (approved):

davsclaus:
Thanks for the great work on improving and fixing camel-as2 and other parts in Camel

PR #10009, opened 2023-05-07
CAMEL-19314 - Fixing the connection pool shutdown when aws health checks are used

rhuan080:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-aws/camel-aws-cloudtrail/src/main/java/org/apache/camel/component/aws/cloudtrail/CloudtrailConsumerHealthCheck.java
+++ b/components/camel-aws/camel-aws-cloudtrail/src/main/java/org/apache/camel/component/aws/cloudtrail/CloudtrailConsumerHealthCheck.java
@@ -30,6 +30,7 @@ import software.amazon.awssdk.services.cloudtrail.model.ListChannelsRequest;
 public class CloudtrailConsumerHealthCheck extends AbstractHealthCheck {
 
     private final CloudtrailConsumer cloudtrailConsumer;
+    private CloudTrailClient client;
 
     public CloudtrailConsumerHealthCheck(CloudtrailConsumer cloudtrailConsumer, String routeId) {
         super("camel", "aws-cloudtrail-consumer-" + routeId);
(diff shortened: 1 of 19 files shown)


davsclaus:
Does this create a new client for each health check call?

davsclaus:
I would assume for health-checks we should reuse the existing client, and NOT close it when check is done. 
Creating a new client is "expensive" as you then setup a new connection to the cloud.

rhuan080:
I'll create a separated client for the health checker and reuse it. I'll send a commit to change it

simon-ras:
The reporter of https://issues.apache.org/jira/browse/CAMEL-19281 claimed that they saw a connection leak from the original implementation. Would it be possible to maybe verify this before making further changes?

We manage our aws clients as Spring beans and our client shouldn't not be shut down before the spring context is destroyed. We are heavily using the SNS and SQS camel components across many services and have not seen any connection leaks on version 3.20.2.

rhuan080:
Hi @simon-ras, 

I have reviewed the changes applied on https://issues.apache.org/jira/browse/CAMEL-19281 and I have seen a potential connection leak to KinesisClient only. I agree the client should not be shut down. However, this fix will solve the problem of closing connections. I think after applying this fix we can back to https://issues.apache.org/jira/browse/CAMEL-19281 and ask to the customer for a reproducer and further information.

rhuan080:
Now, after reviewing the https://issues.apache.org/jira/browse/CAMEL-19281, I'm not sure the KinesisClient is used by the reporter, then I'm not sure the issue is generated by the health checker. However, I think we can apply this fix to solve the close connection.

simon-ras:
The proposed PR does not work for injected clients. I would suggest reverting the previous "fix" instead.

rhuan080:
Thus, I will revert the connections and apply a fix just to KinesisClient, as it has an real scenario of connection leak.

davsclaus:
Is this PR ready to be merged?

rhuan080:
Hi, @davsclaus 

Yes, we can merge this.

simon-ras:
> This PR also need to be backported into 3.20.x as all AWS components are broken on that branch.

I just realized that this was already done!

PR #10032, opened 2023-05-09
CAMEL-19327: camel-jpa support singleResult option

jacekszymanski:
# Description

A common use case is that a query (or a find) should return exactly one object, and any other result is to be considered an error. camel-sql supports thehttps://issues.apache.org/jira/browse/CAMEL-19327# outputType=SelectOne option which will cause the component to throw an exception if the query does not return a single row. camel-jpa does not have such an option and thus user must manually check for body size.

This PR addresses that. It adds a boolean option ```singleResult```, which, if set to true:

- in case of a query will cause the component to put the single fetched object into the message body, not in a List, or throw an exception otherwise
- in case of a find, if no object is found the component will throw an exception rather than put null into the body

it has no effect on updates, and doesn't alter the consumer.

# Target

# Tracking


davsclaus (approved):

oscerd (approved):

davsclaus:
Thanks we then need to cherry-pick to main as well

PR #10037, opened 2023-05-09
(chores) Minor cleanups

orpiske:
- Avoid setting the context twice (and a type check as well)
- Use final where possible

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
@@ -204,8 +204,6 @@ public abstract class MessageSupport implements Message, CamelContextAware, Data
             return;
         }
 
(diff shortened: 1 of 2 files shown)


core/camel-support/src/main/java/org/apache/camel/support/MessageSupport.java
```diff
@@ -204,8 +204,6 @@ public void copyFromWithNewBody(Message that, Object newBody) {
             return;
         }
 
-        // must copy over CamelContext
-        CamelContextAware.trySetCamelContext(that, camelContext);
```
orpiske, line 208: The code is already setting up the context in the caller methods (either in the `copy` or in the caller methods prior to this one).

oscerd (approved):

davsclaus (approved):

PR #10041, opened 2023-05-09
Upgrade Jackson to 2.15.0

apupier:
# Description

Upgrade Jackson to 2.15.0

it aligns with Camel Kamelets catalog
https://github.com/apache/camel-kamelets/pull/1464

Note: there is an API break in transitive dependency with Snake Yaml which required one code change in Camel JBang.

backport from https://github.com/apache/camel/pull/10039

# Target

- [ ] I checked that the commit is targeting the correct branch (note that Camel 3 uses `camel-3.x`, whereas Camel 4 uses the `main` branch)

# Tracking
- [ ] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

# Apache Camel coding standards and style

- [ ] I checked that each commit in the pull request has a meaningful subject line and body.

- [ ] I formatted the code using `mvn -Pformat,fastinstall install && mvn -Psourcecheck`

PR #10042, opened 2023-05-09
Upgrade Jackson to 2.15.0

apupier:
# Description

Upgrade Jackson to 2.15.0

it aligns with Camel Kamelets catalog
https://github.com/apache/camel-kamelets/pull/1464

Note: there is an API break in transitive dependency with Snake Yaml which required one code change in Camel JBang.

backport from https://github.com/apache/camel/pull/10041

# Target

# Tracking


oscerd (changes requested):
I don't think we should change this in 3.20.x, what happens in camel-kamelets remains there. This is a breaking change for 3.20.x and I think we should stay on 2.14.x unless there is a critical CVE.

PR #10066, opened 2023-05-11
CAMEL-18963: upgrade openapi components to latest swagger versions

klease:
This allows support of all OpenAPI specification versions including OpenAPI 3.1.x in the components camel-openapi-java and camel-rest-openapi. The swagger.io dependencies are upgraded to the latest versions and the apicurio dependency is dropped.
Due to this change, in camel-openapi-java, the return type of the public method org.apache.camel.openapi.RestOpenApiReader.read() is now io.swagger.v3.oas.models.OpenAPI. When an OpenAPI 2.0 (swagger) specification is parsed, it is automatically upgraded to OpenAPI 3.0.x by the swagger parser.
In order to support the property openapi.version=2.0 in the openapi-java component, an OpenAPI3to2 converter has been added.
Tests are upgraded except RestOpenApiEndpointTest in camel-rest-openapi which tested apicurio-specific classes.
A new test for OpenAPI 3.1 specifications is added.

# Description

# Target

# Tracking


klease:
This turned out to be quite a big job, especially for camel-openapi-java. There will be a corresponding spring-boot commit to deal with the change to the return type public RestOpenApiReader.read() method.

components/camel-rest-openapi/pom.xml
```diff
@@ -48,17 +49,16 @@
...
+        
+         <dependency>
+            <groupId>io.swagger.core.v3</groupId>
+            <artifactId>swagger-models</artifactId>
+            <version>${swagger-openapi3-version}</version>
+        </dependency>
         <dependency>
-            <groupId>io.apicurio</groupId>
-            <artifactId>apicurio-data-models</artifactId>
-            <version>${apicurio-version}</version>
```
oscerd, line 55: I think the placeholder for apicurio-version could be removed from parent pom, since it won't be used anymore.

oscerd (changes requested):
Just a minor comment. Looks good.

pdumontel:
I'm sorry to use this medium to report a bug, but I don't have access to JIRA at the moment. The Basic and Bearer Token mappings are wrong -- 'basic' and 'bearer' are assigned to the name instead of the scheme property. This breaks swagger-ui when one tries to use the Authorize button. 
Lines 302 and 306 just need to use .scheme() instead of .name()

davsclaus:
Thanks

davsclaus:
PR with fix is welcome

pdumontel:
> PR with fix is welcome

Wow, thank you. I was just about to do that, checked out the source, but I see it's been fixed already. Thanks for the offer though! 
I enjoyed Camel in Action, is there hope for a 3rd edition?

PR #10076, opened 2023-05-12
CAMEL-19060: use an extension for the Synchronization

orpiske:
This helps the code a bit when under heavy contention.


oscerd (approved):

orpiske:
Here's a sample report for a test with 9 producers vs 9 consumers:

Patch vs Baseline (M3)(1)

Not so dramatic for 10x10, but the pattern repeats

Patch vs Baseline (M3)(2)

And the pattern repeats, with it usually beating the code a little more than 70% of the time (usually losing by a small margin on a few of the low contention scenarios):

chart

davsclaus (approved):

PR #10079, opened 2023-05-14
CAMEL-19350: outputTarget in camel-jpa (3.x)

jacekszymanski:
# Description
Unlike camel-sql, camel-jpa does not allow the user to put the result elsewhere than the message body; if this is needed the user must: preserve the body, use the component, put the result in the desired destination, restore the body.

This PR adds the possibility to put the result into a header or property without destroying the message body.

# Target

# Tracking


oscerd (approved):

davsclaus:
Can you use something else than a dot to store as property vs header, this is frankly a bit confusing.
Instead use a prefix such as `property:`, eg property:xxx

jacekszymanski:
OK, updated.

davsclaus (approved):

jacekszymanski:
Sure, I will, I should have it later today.

PR #10083, opened 2023-05-15
CAMEL-19352: improve mybatis doc

Croway:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-mybatis/src/main/docs/mybatis-component.adoc
+++ b/components/camel-mybatis/src/main/docs/mybatis-component.adoc
@@ -338,6 +338,25 @@ usual:
 </camelContext>
 ----
 
+== MyBatis Spring Boot Starter integration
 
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

davsclaus:
I wonder if 

> camel.component.mybatis.sql-session-factory=#sqlSessionFactory

We should make Camel mybatis component have autowried for SqlSessionFactory, then it would automatic detect the mybatis-spring-boot configured and use it OOTB. We do this for some of the other components.

zhfeng (approved):

Croway:
> I wonder if
> 
> > camel.component.mybatis.sql-session-factory=#sqlSessionFactory
> 
> We should make Camel mybatis component have autowried for SqlSessionFactory, then it would automatic detect the mybatis-spring-boot configured and use it OOTB. We do this for some of the other components.

Yes, this would be great, may you provide an example?

davsclaus:
its easy you just need to mark it in the annotation, such as
https://github.com/apache/camel/blob/main/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlComponent.java#L39

PR #10090, opened 2023-05-15
Autowire sqlSessionFactory

Croway:
Hello @davsclaus , is it enough? or the camel-spring-boot component has to be generated again?


davsclaus:
there is a bot for regen SB so if anything is needed it will send a PR automatic

PR #10103, opened 2023-05-16
Adds a null guard against parentObservation

marcingrzejszczak:
shouldn't happen but better be safe than sorry


oscerd (approved):

marcingrzejszczak:
I have no idea how to make this checkstyle pass in CI - I run the same command locally and the build passes :shrug:

davsclaus (approved):

PR #10131, opened 2023-05-17
CAMEL-19378: File Changed ReadLock Strategy with minAge only looks for lastModified

davidvoit:
If a file is copied with the original lastModified date, files are processes instantly. Also the file change detection is disabled if a copy takes more than minAge time. Instead of comparing startTime with lastModified on each iteration we are comparing it with now(). If files are created anew this should be the same behaviour, and files copied with preserved lastModifed are still correctly checked.

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/strategy/FileChangedExclusiveReadLockStrategy.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/strategy/FileChangedExclusiveReadLockStrategy.java
@@ -17,7 +17,6 @@
 package org.apache.camel.component.file.strategy;
 
 import java.io.File;
-import java.util.Date;
 
 import org.apache.camel.Exchange;
 import org.apache.camel.LoggingLevel;
@@ -58,7 +57,7 @@ public class FileChangedExclusiveReadLockStrategy extends MarkerFileExclusiveRea
         long lastModified = Long.MIN_VALUE;
         long length = Long.MIN_VALUE;
         StopWatch watch = new StopWatch();
-        long startTime = (new Date()).getTime();
+        long startTime = System.currentTimeMillis();
 
         while (!exclusive) {
             // timeout check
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davidvoit:
I know the patch is not optimal, it's a first draft to more cleary show the changes. Let's see this first draft as an RFC.
In the end we don't need to differentiate between minage != 0 and minage == 0. We can just always do the currentTime > minTriggerTime.

Just a short word why we are having some problem with the current implemntation. We are loading files from Windows clients. Windows client save files without modifing lastModifed. As our input files are quite big, the camel route get's triggered before the copying is finished.

orpiske:
I wonder if implementing a separate lock strategy would do the trick here? That would give users a choice ... IMHO

davidvoit:
The use case that a file needs to be in the folder for n seconds (to avoid
slow transfers which would not be noticed by the change lock) could also be
achieved with increasing the interval and not decreasing like its hinted in
the documentation. I will play with this stuff on monday

But non theless I would recommend to at least patch the filesize
comparision into the this strategy. This way we would have at least two
stat calls on the input file, instead of directly loading it

Otavio Rodolfo Piske ***@***.***> schrieb am Mi., 17. Mai
2023, 16:51:

davsclaus:
Yeah I can see the point with having length check in that 2nd OR.

davsclaus (approved):

davsclaus:
In light of this I think your patch is better than the current code.

For current time, you can likely use System.currentTimeMillis() instead

davsclaus:
I wonder if you could create a JIRA ticket for this issue also. This helps to backport to patch branches as we need to track changes in ticket system

davidvoit:
Sure. I will be back in the office on monday and then will modify the patch
and create a jira issue.

Claus Ibsen ***@***.***> schrieb am Do., 18. Mai 2023, 19:46:

davidvoit:
As discussed: Simplified the commit and created a jira issue

davsclaus (approved):

oscerd (approved):

davsclaus:
Just a heads up that one of the core tests is failing

FileChangedReadLockMinAgeShortCircuitTest.testChangedReadLockMinAge

davidvoit:
Sorry about that i look into this tomorrow

Claus Ibsen ***@***.***> schrieb am Mo., 22. Mai 2023, 19:06:

davsclaus:
The unit test is using the old behaviour, and with the new code, then the read lock cannot be acquired on first polll. I am changing the unit test.

davidvoit:
Sorry I only checked mvn test for the component, now I learned something new here and will also run mvn test on atleast core or a full mvn test (which takes ages on my setup :-( ) next time.

PR #10133, opened 2023-05-17
(chores) camel-aws2-kinesis: fixed readiness health check

orpiske:


orpiske:
Hey @oscerd, do you mind a look at this one, please? Apparently, the problem is that the ID we return is not the same as before:

```
boolean containsAws2S3HealthCheck = res2.stream()
                    .filter(result -> result.getCheck().getId().startsWith("aws2-kinesis-consumer"))
                    .findAny()
                    .isPresent();
```

I removed that check (and made a small cleanup to the test in the process).

oscerd (approved):

PR #10139, opened 2023-05-18
(chores) camel-opentelemetry: fix RouteConcurrentTest which needs to run in isolation

orpiske:

PR #10140, opened 2023-05-18
(chores) camel-observation: fix RouteConcurrentTest which needs to run in isolation

orpiske:
This also parallelizes the other tests, so it runs faster


orpiske:
Basically the same thing as #10139.

orpiske:
Apache archive seems offline. Merging and fixing (if any problem) later.

PR #10181, opened 2023-05-20
CAMEL-19375: Upgrade to angus-mail 2.0.1, test camel-mail against GreenMail instead of mock_javamail

ppalaga:
# Description

* Upgrade to angus-mail 2.0.1
* Test camel-mail against GreenMail instead of mock_javamail

# Tracking
- [x] https://issues.apache.org/jira/browse/CAMEL-19375


oscerd (approved):

components/camel-mail/src/generated/resources/META-INF/javamail.default.providers
```diff
@@ -1,9 +1,9 @@
 # Jakarta Mail IMAP provider Oracle
-protocol=imap; type=store; class=com.sun.mail.imap.IMAPStore; vendor=Oracle;
-protocol=imaps; type=store; class=com.sun.mail.imap.IMAPSSLStore; vendor=Oracle;
+protocol=imap; type=store; class=org.eclipse.angus.mail.imap.IMAPStore; vendor=Oracle;
```
davsclaus, line 2: I assume maybe vendor should be changed to Eclipse?

davsclaus (approved):

components/camel-mail/src/generated/resources/META-INF/javamail.default.providers
ppalaga: Thanks, let me change it.

ppalaga: It is copied verbatim from angus-mail, so there is not much I can do with it: https://github.com/eclipse-ee4j/angus-mail/blob/5622551fc384dc87b1cdd15bd309b3d4cc1f137d/providers/angus-mail/src/main/resources/META-INF/javamail.default.providers#L2

ppalaga:
faf3f1e:

* Reduced the public API of `org.apache.camel.component.mail.Mailbox` by not inheriting from `ArrayList`
* Made Mailbox a bit more resistant against concurrency issues.

ppalaga:
The angus-mail upgrade does not have any impact on camel-jms and when trying to reproduce locally, I see between 0 and 3 tests failing in camel-jms. Are the JMS tests known to be flaky?

davsclaus:
A little bit as they run concurrently - @otavioprado is improving the tests

ppalaga:
> they run concurrently

Interesting, I did not know some tests were executed in parallel. BTW, with the current changes, it is doable to make the tests in camel-mail ready for concurrent execution too. The key change in the current PR is that each mailbox user and mailbox is explicitly present in the test class as a field. What is missing is to change the creation of the users and mailboxes to random usernames, email addresses and passwords so that they cannot clash with other tests executed parallely. Plus the users and mailboxes should be removed after the test execution. This is doable e.g. by implementing a Junit5 extension.

ppalaga:
86e5586: Fixed formatting

I wonder why `-Pformat` is not active by default?

orpiske:
> > they run concurrently
> 
> Interesting, I did not know some tests were executed in parallel. BTW, with the current changes, it is doable to make the tests in camel-mail ready for concurrent execution too. The key change in the current PR is that each mailbox user and mailbox is explicitly present in the test class as a field. What is missing is to change the creation of the users and mailboxes to random usernames, email addresses and passwords so that they cannot clash with other tests executed parallely. Plus the users and mailboxes should be removed after the test execution. This is doable e.g. by implementing a Junit5 extension.

Yeah. We have been tweaking the JMS tests to reduce their flakiness. If you want to give it a try with the `camel-mail` ones, you need need to set the following properties on the pom:

```
<camel.surefire.reuseForks>true</camel.surefire.reuseForks>
<camel.surefire.forkCount>1</camel.surefire.forkCount>
<camel.surefire.forkTimeout>1200</camel.surefire.forkTimeout>
<camel.surefire.parallel>true</camel.surefire.parallel>
<camel.surefire.parallel.factor>0.5</camel.surefire.parallel.factor>
```

And then adjust the code to avoid using non-shareable resources when possible (or isolate the non-concurrent ones to a different execution). 

Ideally, converting the code to use the [JUnit 5 extension](https://issues.apache.org/jira/browse/CAMEL-18957) from `camel-test-infra-core`  instead of using `CamelTestSupport` from `camel-test-junit5` can also help making them simpler and easier to maintain.

PR #10209, opened 2023-05-25
CAMEL-19388: OSGI - Allow to use CXF 3.6+

essobedo:
## Motivation

Since Apache Camel 3.20 is an LTS version and CXF 3.5 should not be maintained anymore soon as 3.6 and 4.0 have been released, the goal of this task is to extend the version range of CXF to allow OSGI users to choose a higher version of CXF 3 that will remain backward compatible with CXF 3.5.

## Modifications:

* Extends the range of versions of CXF to `[3.5,4.0)`
* Upgrades CXF to the latest maintenance version of 3.5

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -150,8 +150,8 @@
     <corda-version>4.9.3</corda-version>
     <couchbase-client-version>3.4.1</couchbase-client-version>
     <curator-version>4.3.0</curator-version>
-    <cxf-codegen-plugin-version>3.5.5</cxf-codegen-plugin-version>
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus:
cherry pick to `camel-3.x` branch also

essobedo:
@davsclaus For 3.21+, I wanted to upgrade to CXF 3.6 directly, WDYT?

davsclaus:
> @davsclaus For 3.21+, I wanted to upgrade to CXF 3.6 directly, WDYT?

Yeah create a JIRA so it can be tracked

PR #10227, opened 2023-05-29
CAMEL-19399: camel-cxf - Don't add entry in Converter cache on error

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19399

## Motivation

If an error occurs while converting a value thanks to the CxfPayloadConverter, the value null is returned to let other fallback converters to try but if there are no such fallback converters, the cache of converters will keep the information that no converter exists which is not incorrect.

## Modifications

* Returns `MISS_VALUE` instead of `null` in case of an error to ensure that the result won't be stored in the cache.


essobedo:
@ffang Would you mind checking on your side if your use case (https://issues.apache.org/jira/browse/CAMEL-19160) is not broken with this fix?

PR #10234, opened 2023-05-29
CAMEL-19400: camel-elasticsearch - Use nested classes to speed up

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19400

## Motivation

The build on Jenkins takes between 7h and 8h which is much too long especially when we know that the max duration allowed is about 9 h.

So the idea of these changes is to reduce the duration of the integration tests of the elasticsearch component by merging all the integration tests into a single integration test class. The all integration test classes are converted into nested classes, this way `testcontainers` only starts a docker container of ES once for all instead of once for each integration test class.

## Modifications:

* Changes the behavior of `camel-test-infra-common` to only initialize and shutdown when it is called for outer classes
* Converts all integration test classes into nested classes of the same outer class

## Result:

In my local machine, with the old approach, the tests take about 200 seconds while it now takes about 40 seconds.


essobedo:
@orpiske as you are the code owner of the test-infra components, what do you think of this approach?

oscerd (approved):

orpiske:
> @orpiske as you are the code owner of the test-infra components, what do you think of this approach?

That's an interesting approach and I am OK with it. 

I am curious about something, though: have you tried implementing JUnit-based singleton services for Elasticsearch? This approach uses JUnit's [context extension store](https://github.com/apache/camel/blob/main/test-infra/camel-test-infra-common/src/test/java/org/apache/camel/test/infra/common/services/SingletonService.java#L40-L48) to retain the instance between test executions. 
We [do that for Kafka](https://github.com/apache/camel/blob/main/test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/KafkaServiceFactory.java#L25-L44) and it seems to improve the test times too (save for a few scenarios where we need to break stuff / or need broken stuff).

orpiske (approved):

essobedo:
@orpiske I was not aware of this approach, thx for sharing, let me use the same approach for ES to remain homogenous.

essobedo:
The follow up PR https://github.com/apache/camel/pull/10252

PR #10252, opened 2023-06-01
CAMEL-19400: camel-elasticsearch - Use singleton service to speed up

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19400

## Motivation

The build on Jenkins takes between 7h and 8h which is much too long especially when we know that the max duration allowed is about 9 h.

## Modifications:

* Adds a `SingletonService` to the `ElasticSearchServiceFactory`
* Adds the management of the authentication to the `ElasticSearchService`
* Adds the integration tests to use a singleton service to start only once an ES instance

## Result:

In my local machine, with the old approach, the tests take about 200 seconds while it now takes about 50 seconds.


oscerd (approved):

orpiske (approved):
Nice one!

PR #10257, opened 2023-06-02
CAMEL-19408: camel-swift - Restore the component

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19408

## Motivation

The component has been removed because the underlying library from prowide did not support the Jakarta namespace at the time but the latest version supports it such that the component can be restored.

## Modification

* Re-adds the code removed by the commit dfe25ea04a5e9ae3ddc35f5b7e488b57714c3c22
* Upgrades the version of prowide to `SRU2022-10.0.0`
* Configures surefire to allow reflection on the package `java.time`
* Adapts the expected JSON payload to the new datetime format
* Uses UTC as timezone to make sure that the tests pass the same way everywhere


davsclaus (approved):

oscerd (approved):

essobedo:
This PR doesn't affect core which means that the failures are not related to it, let's merge it

PR #10271, opened 2023-06-06
Camel 4.1 RFC: core converter code refactoring

orpiske:
This is a proposed refactoring of the core type conversion code. 

After analyzing the performance of a content-based router, I found out that the core conversion code has some bottlenecks, and a few significant performance problems (such as the one described in [CAMEL-19398](https://issues.apache.org/jira/browse/CAMEL-19398)).

This suggestion reworks the way the CoreTypeConverterRegistry works and: 
* Implements an internal cache of known type conversions to TypeConverters
* Cache conversions misses after they are resolved for the first time (original code did that too, but in a more complicated way)
* Improve the performance (initial estimates indicate something between 5%-10% improvement - but a lot more tests are needed). This seems to be corroborated by the improved resource usage (IPC increases from around ~0.95 to ~1.05 with this code). 

That said, this touches a few critical parts of the core code. Reviews and comments are more than welcome.

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/TypeConverter.java
+++ b/core/camel-api/src/main/java/org/apache/camel/TypeConverter.java
@@ -106,5 +106,4 @@ public interface TypeConverter {
      * @return          the converted value, or <tt>null</tt> if not possible to convert
      */
     <T> T tryConvertTo(Class<T> type, Exchange exchange, Object value);
-
 }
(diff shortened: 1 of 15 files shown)


orpiske:
Also, we probably want to discuss whether we want to have this on RC1 or we wait for 4.1 - if accepted.

orpiske:
Let's save this one for Camel 4.1.

orpiske:
Now that Camel 4 is released, I'll restart working on this.

orpiske:
Closing this one in favor of #11225.

PR #10272, opened 2023-06-06
CAMEL-19421 - Camel-Jira: Use Files.createTempFile in FileConverter i…

oscerd:
…nstead of creating File directly

# Description

# Target

# Tracking

cc @JLLeitschuh

### DIFF
--- a/components/camel-jira/src/main/java/org/apache/camel/component/jira/FileConverter.java
+++ b/components/camel-jira/src/main/java/org/apache/camel/component/jira/FileConverter.java
@@ -39,7 +39,7 @@ public final class FileConverter {
         Path path;
         if (body instanceof byte[]) {
             byte[] bos = (byte[]) body;
-            path = Files.createTempFile(genericFile.getFileName(), null, null);
+            path = Files.createTempFile(genericFile.getFileName(), null);
             Files.write(path, bos, StandardOpenOption.CREATE);
             path.toFile().deleteOnExit();
         } else {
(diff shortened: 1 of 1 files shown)


davsclaus:
I wonder why the camel-jira component has a file converter in the first place?

components/camel-jira/src/main/java/org/apache/camel/component/jira/FileConverter.java
```diff
@@ -34,20 +35,13 @@ private FileConverter() {
...
-                destDir += File.separator;
-            }
-            file = new File(destDir, genericFile.getFileName());
-            if (!file.getCanonicalPath().startsWith(destDir)) {
-                throw new IOException("File is not jailed to the destination directory");
-            }
-            Files.write(file.toPath(), bos, StandardOpenOption.CREATE);
-            // delete the temporary file on exit, as other routing may need the file for post processing
-            file.deleteOnExit();
+            path = Files.createTempFile(genericFile.getFileName(), null, null);
```
JLLeitschuh, line 42: Don't pass an argument to the `attrs` varargs. Without it, it will be implicitly the empty array, which is what you want. If you pass values, the default, secure settings won't be used.

```suggestion
            path = Files.createTempFile(genericFile.getFileName(), null);
```

oscerd: Thanks for the hint, didn't notice that.

oscerd:
> I wonder why the camel-jira component has a file converter in the first place?

I think we should review the reason why and eventually remove it.

JLLeitschuh (approved):
LGTM!

PR #10284, opened 2023-06-07
(chores) camel-core: isolate a few misbehaved tests

orpiske:


essobedo (approved):
Very interesting one 👍

essobedo:
> > Very interesting one 👍
> 
> Thanks. Hopefully will fix a few of the flaky ones on core 🙏 (there's still a few more though).

Yeah I hope so too, this makes me wonder if we have a ticket somewhere to address all the flaky tests that have been disabled over time?

PR #10285, opened 2023-06-07
(chores) camel-core: test isolation + parallelization fixes

orpiske:


orpiske:
These are particularly problematic on s390x.

oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/impl/ScheduledPollConsumerBackoffTest.java
```diff
@@ -22,10 +22,12 @@
 import org.apache.camel.spi.PollingConsumerPollStrategy;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
+import org.junit.jupiter.api.parallel.Isolated;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @DisabledIfSystemProperty(named = "ci.env.name", matches = "github.com", disabledReason = "Flaky on Github CI")
```
essobedo, line 29: Can it be removed?

orpiske: I don't think so, but it won't hurt to test. Let's give it a try.

orpiske:
> 🚫 There are (likely) no components to be tested in this PR

On a side note: we need to implement a tester for the core code: https://issues.apache.org/jira/browse/CAMEL-19427

davsclaus (approved):

PR #10286, opened 2023-06-07
CAMEL-19409: Allow clients to manually restart the subscription

jsight:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceConsumer.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceConsumer.java
@@ -108,6 +108,10 @@ public class SalesforceConsumer extends DefaultConsumer {
         return topicName;
     }
 
+    public SubscriptionHelper getSubscriptionHelper() {
+        return subscriptionHelper;
+    }
+
     @Override
     public void handleException(String message, Throwable t) {
         super.handleException(message, t);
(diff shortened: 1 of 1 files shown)


oscerd (approved):
This should go on main too i guess

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceConsumer.java
```diff
@@ -290,6 +294,24 @@ protected void doStart() throws Exception {
...
+     *
+     * @param alsoStopSubscription
+     */
+    public void stop(boolean alsoStopSubscription) {
+        stop();
+        if (alsoStopSubscription) {
+            try {
+                subscriptionHelper.stop();
+            } catch (Exception e) {
+                LOG.warn("Failed to stop subscription due to: " + e.getMessage());
```
orpiske, line 310: This should be: 

```
                LOG.warn("Failed to stop subscription due to: {}", e.getMessage(), e);
```

orpiske (changes requested):
Thanks for your contribution! I've left one small change request.

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceConsumer.java
orpiske: Thanks. Just a note: I meant without concatenating the log string at all.

Use log marker like this:

```
LOG.warn("Failed to stop subscription due to: {}", e.getMessage(), e);
```

Instead of concatenating the last part of the log message:
```
LOG.warn("Failed to stop subscription due to: " + e.getMessage(), e);
```

jsight: Oh, of course. Sorry I overlooked that part. I have updated the code.

davsclaus (approved):

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceConsumer.java
```diff
@@ -290,6 +294,24 @@ protected void doStart() throws Exception {
...
+     * If alsoStopSubscription is added, any underlying subscriptions will be stopped as well.
+     *
+     * SubscriptionHelper also logs out, so this will terminate the salesforce session as well.
+     *
+     * @param alsoStopSubscription
+     */
+    public void stop(boolean alsoStopSubscription) {
+        stop();
+        if (alsoStopSubscription) {
+            try {
```
davsclaus, line 307: We could add a LOG.info here about manually stopping invoked - this may be good to have in the logs so end users can see this was invoked on purpose

PR #10291, opened 2023-06-08
CAMEL-19307: camel-yaml-io - YAML route dumper

davsclaus:


oscerd (approved):

core/camel-yaml-io/src/test/java/org/apache/camel/yaml/out/XmlToYamlTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @MethodSource("routes")
+    @DisplayName("Test xml to yaml for <routes>")
+    void testRoutes(String xml) throws Exception {
+        try (InputStream is = new FileInputStream("../camel-xml-io/src/test/resources/" + xml)) {
+            RoutesDefinition expected = new ModelParser(is, NAMESPACE).parseRoutesDefinition().get();
+            StringWriter sw = new StringWriter();
+            new ModelWriter(sw).writeRoutesDefinition(expected);
+            LOG.info("xml={}\n{}\n", xml, sw);
+        }
+    }
```
orpiske, line 53: I think there is a problem with this test: it is missing an assertion (so it would always pass). Maybe wrap the code with `Assertions.assertDoesNotThrow` to ensure it fails when there's a problem?

orpiske (commented):
Just added some suggestions

PR #10315, opened 2023-06-09
CAMEL-19433: Handle error while consuming records from topic

devdeegaw:
# Description

In the event of an error occurring during record consumption from a topic, the Kafka consumer will persistently send LeaveGroup requests to the coordinator due to consumer poll timeout has expired.

The camel Kafka consumer silently ignores it and throws no exceptions when fetching data for the topics.
Handling the error safely unsubscribe the consumer, accompanied by an appropriate log message

# Target

# Tracking

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -358,6 +358,10 @@ public class KafkaFetchRecords implements Runnable {
             if (LOG.isTraceEnabled()) {
                 LOG.trace("The kafka consumer was woken up while polling on thread {} for {}", threadId, getPrintableTopic());
             }
+        } catch (Error e) { // NOSONAR - rethrown
+            LOG.error("Error {} while consuming from topic : {}", e.getMessage(), getPrintableTopic(), e);
+            safeUnsubscribe();
+            throw e;
         } catch (Exception e) {
             if (LOG.isDebugEnabled()) {
                 LOG.warn("Exception {} caught by thread {} while polling {} from kafka: {}",
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -358,10 +358,6 @@ public class KafkaFetchRecords implements Runnable {
             if (LOG.isTraceEnabled()) {
                 LOG.trace("The kafka consumer was woken up while polling on thread {} for {}", threadId, getPrintableTopic());
             }
-        } catch (Error e) { // NOSONAR - rethrown
-            LOG.error("Error {} while consuming from topic : {}", e.getMessage(), getPrintableTopic(), e);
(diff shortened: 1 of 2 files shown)


components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -358,6 +358,10 @@ protected void startPolling() {
             if (LOG.isTraceEnabled()) {
                 LOG.trace("The kafka consumer was woken up while polling on thread {} for {}", threadId, getPrintableTopic());
             }
+        } catch (Error e) { // NOSONAR - rethrown
```
davsclaus, line 361: I kinda forgot whether the order of the catch matters, eg I would assume this should be last, so regular exception can be caught for the normal exceptions

orpiske (changes requested):
~I am not entirely sure this is correct. We should not be catching [Error](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Error.html).

I think a more adequate way to do this is by [implementing](https://github.com/apache/camel/blob/2925e9694c521de3b78b7cb288037d321cf59717/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/KafkaErrorStrategies.java) a [PollExceptionStrategy](https://github.com/apache/camel/blob/2925e9694c521de3b78b7cb288037d321cf59717/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/PollExceptionStrategy.java) that instructs the record fetcher to unsubscribe. ~

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: Yeah, if we were to implement this ... it would have to be the last one, indeed.

devdeegaw: @orpiske Although it would not resolve the problem entirely, capturing the error and implementing appropriate logging would provide valuable insights to the developer regarding the occurrences of poll timeouts and the consumer leaving the group.

devdeegaw: @orpiske We found the root cause during our usage of the Spring Kafka framework (KafkaMessageListenerContainer as the listener for message consumption). However, thanks to the error handling and logging capabilities provided by the framework, we were able to capture and log the errors, despite not yet determining the root cause.

orpiske: The problem is: just because it worked on your particular scenario, there's no guarantee it would work reliably (or work at all) on all others that fall on the same category (i.e.: all other sub-classes of Error).

orpiske:
Closing this PR as it is stalled and the problem needs to be solved differently ([work-around described here](https://issues.apache.org/jira/browse/CAMEL-19433?focusedCommentId=17737789&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17737789))

PR #10325, opened 2023-06-10
CAMEL-19417: camel-spring-rabbitmq - convert all message properties by default

sdurrenmatt:
# Description

The goal is to support all the most important RabbitMQ message properties by default (without the need to provide a custom ``MessagePropertiesConverter``). Those include the removed RabbitMQ component's message properties. Can simplify migration of applications from RabbitMQ to Spring RabbitMQ, which is a necessary step for migrating to Camel 4.

I had to modify how content type and timestamp were already handled. I found it better to use new Spring RabbitMQ constants rather than exchange constants. This is a more consistent approach imo. Both are optional and application-specific, like most of the other message properties.

This PR also includes a copy-paste bug fix in the method ``appendInputHeader``.

# Target

# Tracking

### DIFF
--- a/components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
+++ b/components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
@@ -43,61 +43,61 @@ public class DefaultMessagePropertiesConverter implements MessagePropertiesConve
         MessageProperties answer = new MessageProperties();
         Message message = exchange.getMessage();
 
-        Object deliveryMode = message.removeHeader(SpringRabbitMQConstants.DELIVERY_MODE);
+        Object deliveryMode = message.getHeader(SpringRabbitMQConstants.DELIVERY_MODE);
         if (deliveryMode != null) {
             answer.setDeliveryMode(deliveryMode instanceof MessageDeliveryMode ? (MessageDeliveryMode) deliveryMode
                     : MessageDeliveryMode.fromInt(Integer.parseInt(deliveryMode.toString())));
         }
-        Object type = message.removeHeader(SpringRabbitMQConstants.TYPE);
+        Object type = message.getHeader(SpringRabbitMQConstants.TYPE);
         if (type != null) {
             answer.setType(type.toString());
         }
-        Object contentType = message.removeHeader(SpringRabbitMQConstants.CONTENT_TYPE);
+        Object contentType = message.getHeader(SpringRabbitMQConstants.CONTENT_TYPE);
         if (contentType != null) {
             answer.setContentType(contentType.toString());
         }
-        Object contentLength = message.removeHeader(SpringRabbitMQConstants.CONTENT_LENGTH);
+        Object contentLength = message.getHeader(SpringRabbitMQConstants.CONTENT_LENGTH);
         if (contentLength != null) {
             answer.setContentLength(Long.parseLong(contentLength.toString()));
         }
-        Object contentEncoding = message.removeHeader(SpringRabbitMQConstants.CONTENT_ENCODING);
+        Object contentEncoding = message.getHeader(SpringRabbitMQConstants.CONTENT_ENCODING);
         if (contentEncoding != null) {
             answer.setContentEncoding(contentEncoding.toString());
         }
-        Object messageId = message.removeHeader(SpringRabbitMQConstants.MESSAGE_ID);
+        Object messageId = message.getHeader(SpringRabbitMQConstants.MESSAGE_ID);
         if (messageId != null) {
             answer.setMessageId(messageId.toString());
         }
-        Object correlationId = message.removeHeader(SpringRabbitMQConstants.CORRELATION_ID);
+        Object correlationId = message.getHeader(SpringRabbitMQConstants.CORRELATION_ID);
         if (correlationId != null) {
             answer.setCorrelationId(correlationId.toString());
         }
-        Object replyTo = message.removeHeader(SpringRabbitMQConstants.REPLY_TO);
+        Object replyTo = message.getHeader(SpringRabbitMQConstants.REPLY_TO);
         if (replyTo != null) {
             answer.setReplyTo(replyTo.toString());
         }
-        Object expiration = message.removeHeader(SpringRabbitMQConstants.EXPIRATION);
+        Object expiration = message.getHeader(SpringRabbitMQConstants.EXPIRATION);
         if (expiration != null) {
             answer.setExpiration(expiration.toString());
         }
-        Object timestamp = message.removeHeader(SpringRabbitMQConstants.TIMESTAMP);
+        Object timestamp = message.getHeader(SpringRabbitMQConstants.TIMESTAMP);
         if (timestamp != null) {
             answer.setTimestamp(timestamp instanceof Date ? (Date) timestamp
                     : new Date(Long.parseLong(timestamp.toString())));
         }
-        Object userId = message.removeHeader(SpringRabbitMQConstants.USER_ID);
+        Object userId = message.getHeader(SpringRabbitMQConstants.USER_ID);
         if (userId != null) {
             answer.setUserId(userId.toString());
         }
-        Object appId = message.removeHeader(SpringRabbitMQConstants.APP_ID);
+        Object appId = message.getHeader(SpringRabbitMQConstants.APP_ID);
         if (appId != null) {
             answer.setAppId(appId.toString());
         }
-        Object priority = message.removeHeader(SpringRabbitMQConstants.PRIORITY);
+        Object priority = message.getHeader(SpringRabbitMQConstants.PRIORITY);
         if (priority != null) {
             answer.setPriority(Integer.parseInt(priority.toString()));
         }
-        Object clusterId = message.removeHeader(SpringRabbitMQConstants.CLUSTER_ID);
+        Object clusterId = message.getHeader(SpringRabbitMQConstants.CLUSTER_ID);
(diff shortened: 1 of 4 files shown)


components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
```diff
@@ -39,9 +40,64 @@ public DefaultMessagePropertiesConverter(CamelContext camelContext, HeaderFilter
     @Override
     public MessageProperties toMessageProperties(Exchange exchange) {
         MessageProperties answer = new MessageProperties();
-        String contentType = ExchangeHelper.getContentType(exchange);
+
+        Object deliveryMode = exchange.getMessage().removeHeader(SpringRabbitMQConstants.DELIVERY_MODE);
```
essobedo, line 44: Let's put `exchange.getMessage()` in a local variable instead of calling it several times

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
```diff
@@ -39,9 +40,64 @@ public DefaultMessagePropertiesConverter(CamelContext camelContext, HeaderFilter
...
+        if (deliveryMode != null) {
+            answer.setDeliveryMode(deliveryMode instanceof MessageDeliveryMode ? (MessageDeliveryMode) deliveryMode
+                    : MessageDeliveryMode.fromInt(Integer.parseInt(deliveryMode.toString())));
+        }
+        Object type = exchange.getMessage().removeHeader(SpringRabbitMQConstants.TYPE);
```
essobedo, line 49: Is-it necessary to remove the header?

sdurrenmatt: Yes we do not want it to be copied later on to the RabbitMQ message.

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQConstants.java
```diff
@@ -23,12 +23,47 @@ public final class SpringRabbitMQConstants {
...
 
     public static final String CHANNEL = "CamelSpringRabbitmqChannel";
-    @Metadata(description = "The exchange key.", javaType = "String")
+    @Metadata(description = "Producer: To override the endpoint configuration's routing key.", javaType = "String")
     public static final String ROUTING_OVERRIDE_KEY = "CamelSpringRabbitmqRoutingOverrideKey";
-    @Metadata(description = "The exchange name.", javaType = "String")
+    @Metadata(description = "Producer: To override the endpoint configuration's exchange name.", javaType = "String")
     public static final String EXCHANGE_OVERRIDE_NAME = "CamelSpringRabbitmqExchangeOverrideName";
-    @Metadata(description = "Delivery tag for manual acknowledge mode", javaType = "long")
+    @Metadata(description = "Consumer: Whether the message was previously delivered and requeued.", javaType = "Boolean")
```
essobedo, line 30: If it is only meant for the consumer, you could rather set the label to "consumer"

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQConstants.java
```diff
@@ -23,12 +23,47 @@ public final class SpringRabbitMQConstants {
     public static final String DEFAULT_EXCHANGE_NAME = "default";
 
     public static final String CHANNEL = "CamelSpringRabbitmqChannel";
-    @Metadata(description = "The exchange key.", javaType = "String")
+    @Metadata(description = "Producer: To override the endpoint configuration's routing key.", javaType = "String")
```
essobedo, line 26: ditto with "producer"

components/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerIT.java
```diff
@@ -118,15 +120,26 @@ public void testProducerContentType() throws Exception {
         admin.declareExchange(t);
         admin.declareBinding(BindingBuilder.bind(q).to(t).with("foo.bar.#"));
 
-        template.sendBodyAndHeader("direct:start", "<price>123</price>", Exchange.CONTENT_TYPE, "application/xml");
+        template.sendBodyAndHeaders("direct:start", "<price>123</price>",
+                ImmutableMap.<String, Object>builder()
```
essobedo, line 124: Let's use `Map.of` instead

essobedo (commented):
Hi, thx for the PR, here are some remarks/questions

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQConstants.java
sdurrenmatt: done

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQConstants.java
sdurrenmatt: done

components/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerIT.java
sdurrenmatt: done

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
sdurrenmatt: done

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
sdurrenmatt: on second thought maybe this is something that header filter strategy should take care of? we may want to re-use a header or two after the message has been published.

essobedo: What I can tell for sure is that it is quite unexpected that a converter alters the message. If they need to be removed it should be definitively done elsewhere IMHO

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
```diff
@@ -39,12 +41,68 @@ public DefaultMessagePropertiesConverter(CamelContext camelContext, HeaderFilter
     @Override
     public MessageProperties toMessageProperties(Exchange exchange) {
         MessageProperties answer = new MessageProperties();
-        String contentType = ExchangeHelper.getContentType(exchange);
+        Message message = exchange.getMessage();
+
+        Object deliveryMode = message.getHeader(SpringRabbitMQConstants.DELIVERY_MODE);
+        if (deliveryMode != null) {
+            answer.setDeliveryMode(deliveryMode instanceof MessageDeliveryMode ? (MessageDeliveryMode) deliveryMode
+                    : MessageDeliveryMode.fromInt(Integer.parseInt(deliveryMode.toString())));
```
essobedo, line 49: Let's see what others say but I'm not really a fan of calling `toString` without even knowing/checking the type

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
```diff
@@ -39,12 +41,68 @@ public DefaultMessagePropertiesConverter(CamelContext camelContext, HeaderFilter
...
+        }
+        Object type = message.getHeader(SpringRabbitMQConstants.TYPE);
+        if (type != null) {
+            answer.setType(type.toString());
```
essobedo, line 53: Same here, why don't you use the method `getHeader(String name, Class<T> type)` to trigger the type conversion if needed?

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
sdurrenmatt: right this is ambiguous especially since this is an enum... I simplifed that and added the enums for the doc.

components/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/DefaultMessagePropertiesConverter.java
sdurrenmatt: done everywhere

essobedo (approved):
Sounds much better thx 👍

oscerd (approved):

davsclaus (approved):

PR #10381, opened 2023-06-14
Backport to 3.18 the krb5 explanation for camel-kafka

orpiske:


davsclaus (approved):

orpiske:
> 3.18 is EOL - we have doe the last release (on VOTE now) but the docs are build from branch so doc changes can still be updated on the website

Yeah. I'm think it could help users that check on older versions (just to ensure a bit of consistency between the latest LTSes).


I think we need to wait for the vote to pass before merging this, right? Just to be on the safe side in case it has to be restarted.

davsclaus:
If it gets restarted then we can include this PR, so its okay to merge

PR #10400, opened 2023-06-15
should servlet and rest throw exceptions when duplicate and ambiguous paths occur

EvanMi:
1. throw exception when duplicate servlet path occur in create stage.
2. throw exception when ambiguous path occur in rest requesting stage.

### DIFF
--- a/components/camel-elasticsearch/src/main/java/org/apache/camel/component/es/converter/ElasticsearchActionRequestConverter.java
+++ b/components/camel-elasticsearch/src/main/java/org/apache/camel/component/es/converter/ElasticsearchActionRequestConverter.java
@@ -21,6 +21,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


davsclaus:
There is some code changes for camel-elasticsearch which should not be in this PR

EvanMi:
> There is some code changes for camel-elasticsearch which should not be in this PR

my mistake, thaks! @davsclaus

davsclaus (approved):

oscerd (approved):

davsclaus:
@EvanMi oh sorry we merged this too soon. There is test error in camel-serlvet in RestServletGetWildcardsTest. 

You changes seems to pick the wrong matching. Can you take another look and send a new PR.
Also can you create a JIRA ticket so we can better track this change.

EvanMi:
> @EvanMi oh sorry we merged this too soon. There is test error in camel-serlvet in RestServletGetWildcardsTest.
> 
> You changes seems to pick the wrong matching. Can you take another look and send a new PR. Also can you create a JIRA ticket so we can better track this change.

Sure,i'll take care of this.

PR #10416, opened 2023-06-16
Fix bug for https://github.com/apache/camel/pull/10400

EvanMi:
For detail see https://github.com/apache/camel/pull/10400, this PR fixed a bug.


oscerd (approved):

davsclaus (approved):

davsclaus:
@EvanMi can you check camel-undertow, it has failure with RestUndertowHttpGetCorsTest

EvanMi:
@davsclaus I found the problem out, the reason is that /users/{a}/b?restrickedMethod=GET&OPTIONS /users/{a}/b?restrickedMethod=POST&OPTIONS , these two routes a all match when options request coming in.Should i treat OPTIONS as special? Or an good advice?

图片

EvanMi:
> @davsclaus I found the problem out, the reason is that /users/{a}/b?restrickedMethod=GET&OPTIONS /users/{a}/b?restrickedMethod=POST&OPTIONS , these two routes a all match when options request coming in.Should i treat OPTIONS as special? Or an good advice?
> 图片

图片

EvanMi:
@davsclaus with this condition add, test passed
图片

davsclaus:
Yeah options is _special_

PR #10428, opened 2023-06-18
Escape OPTIONS from ambiguous path exception

EvanMi:
Escape OPTIONS from ambiguous path exception
for detials, see
https://github.com/apache/camel/pull/10416
https://github.com/apache/camel/pull/10400


core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
```diff
@@ -218,6 +218,16 @@ public static ConsumerPath matchBestPath(String requestMethod, String requestPat
         return answer;
     }
 
+    /**
+     *
+     * @param  requestMethod The request method
+     * @param  requestPath   The request path
+     * @return               if the request method and path can escape from the ambiguous exception
+     */
+    private static boolean shouldEscapeFromAmbiguous(String requestMethod, String requestPath) {
```
davsclaus, line 227: maybe we should rename this to `canBeAmbigious`

As escape you would think its about escaping characthers, eg s -> \s

EvanMi: sure, i'm not good at naming!

davsclaus (approved):

PR #10441, opened 2023-06-19
init outputTimestamp

hboutemy:
one character does not get updated at release time

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,7 @@
         <minimalJavaBuildVersion>${jdk.version}</minimalJavaBuildVersion>
 
         <!-- reproduceable builds: https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
-        <project.build.outputTimestamp>1</project.build.outputTimestamp>
+        <project.build.outputTimestamp>1687211006</project.build.outputTimestamp>
 
         <maven-compiler-plugin-version>3.11.0</maven-compiler-plugin-version>
         <maven-javadoc-plugin-version>3.5.0</maven-javadoc-plugin-version>


oscerd (approved):

pom.xml
```diff
@@ -117,7 +117,7 @@
         <minimalJavaBuildVersion>${jdk.version}</minimalJavaBuildVersion>
 
         <!-- reproduceable builds: https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
-        <project.build.outputTimestamp>1</project.build.outputTimestamp>
+        <project.build.outputTimestamp>1687211006</project.build.outputTimestamp>
```
davsclaus, line 120: Is this number just a random number you typed?

davsclaus:
@ppalaga any feedback on this PR ?

pom.xml
davsclaus: Can you point to more details on this practice? Do other open source projects do this or where can we find more about this

ppalaga: @hboutemy thanks for coming around! I was struggling a bit with project.build.outputTimestamp last week, because some zip creating plugin (perhaps jar plugin) was complaining about a negative timestamp. That's why I changed it to the current value `1980-01-01T00:00:02Z` and that's why the current PR has conflicts. 

Was that maybe also the reason why you have sent this PR?

> Can you point to more details on this practice?

I think https://maven.apache.org/guides/mini/guide-reproducible-builds.html is a nice quick resource about the topic of reproducible builds. (I guess it was written by you @hboutemy?) The part relevant for us is

> Q. Can the project.build.outputTimestamp property in pom.xml be updated automatically at release time?
> A. Yes.
>
> Details depend on your release process tooling:
>
> if you use [maven-release-plugin](https://maven.apache.org/plugins/maven-release-plugin/), you'll need version 3.0.0-M1 or later: it will automatically update the timestamp value in pom.xml during the release in the same commit that updates version,

We happen to use an ancient version 2.4.2, there is a comment referring to https://issues.apache.org/jira/browse/MRELEASE-812 which is fixed since 9+ years, so we could perhaps upgrade, @gzurowski?

ppalaga:
> @ppalaga any feedback on this PR ?

Thanks for the headsup, I commented inline.

pom.xml
gzurowski: Yes, the `maven-release-plugin` we are using is quite dated. We can try to upgrade. I am happy to test the release process with the upgraded plugin.

oscerd:
@hboutemy Can you rebase?

hboutemy:
given ppalaga changed the value and we upgraded the release plugin in #10572, this PR is now obsolete: I'm closing it, job done

PR #10467, opened 2023-06-21
CAMEL-19481: Upgrade to Jetty 10

essobedo:


davsclaus:
Is there anything from the Karaf community whether they will do a Karaf release that supports Jetty 10 ?

essobedo:
It is still under discussion but it is doable even in the next Karaf 4.4.x which will be released very soon but not before 3.21. So the question is do we make users wait 6 months to get this upgrade to have included in 3.22 or can we include it directly in 3.21 with a note indicating the limitation that will be removed once the next Karaf version will be released in a couple of weeks

davsclaus:
We keep 3.21 stable as-is - people dont casually upgrade karaf.
We are to build the release tomorrow.

davsclaus:
And we can do a 3.22.0 release sooner if we think its ready and community users would benefit from Karaf 4.5.x and Jetty 10 support

essobedo:
Ok let's merge it once a Karaf version supporting Jetty 10 is released

davsclaus:
camel 3.22 is being released now as-is, and there is no new 3.x releases planned

PR #10476, opened 2023-06-22
Newcomer friendly default build, faster fast profile

ppalaga:
https://issues.apache.org/jira/browse/CAMEL-19465
https://issues.apache.org/jira/browse/CAMEL-19461
https://issues.apache.org/jira/browse/CAMEL-19466


orpiske:
Overall LGTM. I'm going to do a few manual tests on different archs to double check if everything is alright on ARM and s390x - not concerned, it should be.

oscerd (approved):
I think this will need some manual tests, but LGTM. Thanks Peter.

orpiske (approved):
This is one of the patches that makes me anxious to see it merged. Nice one @ppalaga !

Here's the results on my systems (look at those numbers 🤩 ): 

| Arch | Tool | Patch Time | Baseline Time | 
| ----| ---- | ---- | ----| 
| x86 | mvnd | 08:59 | 15:32 | 
| ARM | mvnd | 01:43 | 08:21 | 
| s390x | mvn | 05:56 | 30:26  | 

All tests where done on a clean repo.

I haven't tested the build on power as I don't have the time to setup the machine, but it should be fine from what I can check now.

ppalaga:
9e53ff8: 

* Rabased
* Added "Workaround INFRA-24713 in the deploy script" https://github.com/apache/camel/pull/10476/commits/a3242d199bb3359253f6b010905701a4b6cfe8e3
* Added "Adapt CI scripts to new fast profile and..." https://github.com/apache/camel/pull/10476/commits/9e53ff8c3b76fef110b6f98b3282910e886bc2de

ppalaga:
Thanks for testing the alt. platforms, @orpiske !

essobedo:
Speaking of improving the build command any reason why not setting `clean install` as `defaultGoal`?

ppalaga:
> Speaking of improving the build command any reason why not setting `clean install` as `defaultGoal`?

I never use `defaultGoal` because I rarely know what is it for the current project if any at all. I wonder why `install` should be the default goal, rather than anything else? I am just curious how others, who see some value in it, think about this.

essobedo:
Because 99% of the time maven users usually launch the command `mvn clean install..." like it is describe in the contributing guide https://camel.apache.org/community/contributing/

essobedo:
For example, it is how it is done in Quarkus https://github.com/quarkusio/quarkus/blob/main/pom.xml#L200

oscerd:
mvnd clean install -Dquickly - 03:50 min - first time

mvnd clean install -Dquickly - 03:12 min - second time

mvnd install -Dquickly - 01:00 min - third time

Amazing.

Thanks.

Jenkinsfile.deploy
```diff
@@ -66,6 +66,10 @@ pipeline {
                 }
             }
             steps {
+                // Workaround https://issues.apache.org/jira/browse/INFRA-24713
+                def text = readFile ".mvn/wrapper/maven-wrapper.properties"
+                text.replaceAll("distributionUrl=.*", "https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.8/apache-maven-3.8.8-bin.zip")
+
```
orpiske, line 72: This approach did not work for our CI. The snapshot deployment is failing again: https://ci-builds.apache.org/job/Camel/job/Camel%20Daily%20Snapshot%20Deploy/job/main/363/

ppalaga:
> Because 99% of the time maven users usually launch the command `mvn clean install..." like it is described in the contributing guide https://camel.apache.org/community/contributing/

Sounds legit, thanks for explaining @essobedo.

Jenkinsfile.deploy
ppalaga: Sorry about that. Here is an attempt to fix it: https://github.com/apache/camel/pull/10486

orpiske:
Just bringing a separate discussion here, as more people are subscribed to this thread. 

Looking at the CI jobs we have: 
- Deployment job has progressed a bit. There seems to be something strange on `camel-xml-io` preventing it from [going further](https://ci-builds.apache.org/job/Camel/job/Camel%20Daily%20Snapshot%20Deploy/job/main/368/console)  and also with `camel-package-maven-plugin` ([same one causing problem on the main build](https://ci-builds.apache.org/job/Camel/job/Camel%20JDK17/job/main/951/console)). 

I do not have these problems on my own CI, which is very similar to the ASF one, so I think we are close to having the builds passing.

orpiske:
Another problem I found: `camel-kafka` is running the tests even though `-Dquickly` is being passed. I think that's because we configure the executions for testing backwards compatibility in a separate run.

orpiske:
It seems we are using a very old version of the invoker plugin (1.8), released in 2012. I managed to reproduce the problem on my own CI, so now trying a version with invoker 3.6.0.

orpiske:
> It seems we are using a very old version of the invoker plugin (1.8), released in 2012. I managed to reproduce the problem on my own CI, so now trying a version with invoker 3.6.0.

Upgrading seems to have helped. It now fails on `camel-util` because `maven-source-plugin` is defined twice when use the `deploy` profile. I think we can remove that, as attaching the sources can be done using the `source-jar` or `deploy` profiles (and, likely, not necessary during regular build).

orpiske:
So, current status for our builds is:
* Snapshot deploy: last build went as far as 95% of the build. I fixed it earlier today and a new build is in progress. Hopefully that will complete.
* JDK 17 build: currently hangs on the core tests. I don't think it's related to this patch, as this is a known issue (I am currently investigating this).

oscerd:
The SBOM generation action is not working anymore after this PR. I'll have a look.

PR #10479, opened 2023-06-23
CAMEL-19490: camel-elasticsearch - Upgrade to 8.8.x

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-19490

## Motivation

Elastic search 8.8 has just been released, we would like to upgrade to the latest version.

## Modifications:

* Updates the version of Elastic Search to `8.8.1`
* Updates the version of the docker image to `docker.elastic.co/elasticsearch/elasticsearch:8.8.1`
* Uses the default version of Jackson which is `2.15` as it is the version expected by the elastic search client

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -146,8 +146,8 @@
         <eddsa-version>0.3.0</eddsa-version>
         <egit-github-core-version>2.1.5</egit-github-core-version>
         <ehcache3-version>3.10.8</ehcache3-version>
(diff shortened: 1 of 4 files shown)


oscerd (changes requested):
Can you amend the commit message? It should be camel-elasticsearch. Nothing important.

essobedo:
> Can you amend the commit message? It should be camel-elasticsearch. Nothing important.

Oops, good catch, thx for the feedback 🙏

essobedo:
The failing tests are due to an incorrect path in the github action script, so let's merge it

PR #10480, opened 2023-06-23
(chores) code cleanups: move array designators to the type

orpiske:


orpiske:
GH staled. Let's merge to have some tests on ASF CI.

PR #10483, opened 2023-06-23
Handle -Pfastinstall as an alias of -Dquickly

ppalaga:
This PR addresses the requests on camel-dev ML to keep the old `fastinstall` profile in some form for backwards compatibility reasons. Hence this PR adds a small maven extension that does just that: when `-Pfastinstall` is present among maven command line arguments, the `full` profile is made inactive programmatically. Passing `-Dquickly` does the same.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,3 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <extensions>
+    <extension>
+        <groupId>org.l2x6.cq</groupId>
+        <artifactId>cq-alias-fastinstall-quickly-extension</artifactId>
+        <version>4.4.0</version>
+    </extension>
 </extensions>


orpiske (changes requested):
Hey @ppalaga, I am not entirely opposed to the suggestion, but I couldn't find the source of this plugin. It would be good to double check if complies with ASF guidelines.

oscerd (changes requested):
It's coming from here: https://github.com/l2x6/cq-maven-plugin/blob/050436107915d60747524ebf33cc835e479102ca/LICENSE.txt

Is there a specific reason for having all the cq stuff in terms of maven, in a separated repository and separated group id?

ppalaga:
> Is there a specific reason for having all the cq stuff in terms of maven, in a separated repository and separated group id?

It is not all, and it is not cq-only. They serve also other projects like quarkus-cxf. So the primary reason is reuse. Another important reason is build performance: not having to rebuild the tools with every build of the repo where those tools are used saves time and CO2. In case of this specific Maven extension, it actually has to be hosted outside of the camel git repo. Building and loading in the same build would not work. I am open to moving it elsewhere if you like?

oscerd:
> > Is there a specific reason for having all the cq stuff in terms of maven, in a separated repository and separated group id?
> 
> It is not all, and it is not cq-only. They serve also other projects like quarkus-cxf. So the primary reason is reuse. Another important reason is build performance: not having to rebuild the tools with every build of the repo where those tools are used saves time and CO2. In case of this specific Maven extension, it actually has to be hosted outside of the camel git repo. Building and loading in the same build would not work. I am open to moving it elsewhere if you like?

No no, I was just wondering why, but it makes totally sense and the license is fine. I think it's very specific so it makes sense to have a different release schedule and avoid being dependent on the ASF release policy.

orpiske (approved):

oscerd (approved):

oscerd:
I said is cq only because the README say so: https://github.com/l2x6/cq-maven-plugin#cq-maven-plugin

ppalaga:
> I said is cq only because the README

Oh, sorry, the README does not mirror the reality.

oscerd:
No worries, maybe it could help a newcomer.

PR #10505, opened 2023-06-27
Add possibility to override LRAClient

johbossle:
# Description

As a follow up on #10498 

The Saga EIP pattern is using rest based communication for interaction with the lra-coordinator (e.g. narayana). Although there are ways to secure the communication with the coordinator, this is not supported by the underlying code. Also, it is not possible to easily override some implementations, since they are created by internal constructors.

This change is targeting, that one can override (influence) the creation of the LRAClient in the LRASagaService. With this "extension", it is then possible to override some default behavior and for example you can set the HttpClient and use your own adjusted instance with an Authenticator or ssl-settings.

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [CAMEL 19508](https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-19508) filed for the change (usually before you start working on it).

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAClient.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAClient.java
@@ -116,9 +116,9 @@ public class LRAClient implements Closeable {
                 lraEndpoint = lraEndpoint + "?" + HEADER_TIME_LIMIT + "=" + step.getTimeoutInMilliseconds().get();
             }
             HttpRequest request = prepareRequest(URI.create(lraEndpoint))
-                    .header(HEADER_LINK, link.toString())
-                    .header(Exchange.SAGA_LONG_RUNNING_ACTION, lra.toString())
-                    .header("Content-Type", "text/plain")
+                    .setHeader(HEADER_LINK, link.toString())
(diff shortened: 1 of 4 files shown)


zhfeng (approved):

davsclaus (approved):

oscerd (approved):
Should be on main too.

zhfeng (approved):

davsclaus:
If this is a good idea to backport to 3.20.x/3.21.x etc then we need a JIRA to track each non-trivial change in LTS branches

johbossle:
> Yes @oscerd is correct we need also to remember to cherry pick this to main branch

@davsclaus Is there something where I can help to cherry-pick and port to main-branch?

davsclaus:
No its fine we could cherry pick this cleanly to main branch

johbossle:
Just want to end up here with an example usage pattern to supply custom authentication header:

```java
public class AuthenticatingLRASagaService extends LRASagaService {
  @Override
  protected LRAClient createLRAClient() {
    return new LRAClient(this, HttpClient.newBuilder().build()){
      @Override
      protected HttpRequest.Builder prepareRequest(URI uri) {
        HttpRequest.Builder builder = super.prepareRequest(uri);
        builder.header("Authentication", "MY_TOKEN");
        return builder;
      }
    };
  }
}
```

PR #10511, opened 2023-06-27
Fix test and use single kafka instance

Croway:
The test was successful because 2 kafka instances were running, therefore the failure on P/Z.

@Balavva-Mirji


oscerd (approved):

orpiske (approved):
Thanks!

PR #10512, opened 2023-06-27
CAMEL-19497: prevent unhandled exceptions from blocking indefinitely

orpiske:
This ensures that the done callback will be called (or, at least, tried to be called) in the case of unhandled exceptions (a few rare cases involving Throwables)


oscerd (approved):

davsclaus (approved):

davsclaus:
There is also a try catch block a bit further up, it could benefit from also call done in finally

orpiske:
> There is also a try catch block a bit further up, it could benefit from also call done in finally

Oh, nice. I'll add it there too. Thanks!

orpiske:
@davsclaus adding to the ones above cause side effects. This one is OK because the `callback.done(true);` is supposed to run regardless of a exception being thrown or not. On the one above, it's only called when catching an exception, so adding to the finally would also cause it to run when it shouldn't. I'm leaving as is for now - I can investigate later if you want.

davsclaus:
ok

PR #10517, opened 2023-06-27
Make plugin integration tests runnable against empty local Maven repository with Apache Snapshots repository disabled

ppalaga:


oscerd (approved):

ppalaga:
@essobedo `incremental-build.sh` puts `tooling/maven/camel-package-maven-plugin/src/it/HeaderSupport` into `-pl` if there was a change under it. Apparently, it cannot work like that. `findProjectRoot` should climb up to `tooling/maven/camel-package-maven-plugin`. I am not going to risk breaking it more with my sparse bash syntax knowledge.

PR #10539, opened 2023-06-28
CAMEL-19560 - Manage micrometer artifacts to build and test with a single deterministic set of versions

ppalaga:
https://issues.apache.org/jira/browse/CAMEL-19560

### DIFF
--- a/components/camel-micrometer/pom.xml
+++ b/components/camel-micrometer/pom.xml
@@ -104,6 +104,12 @@
             <version>${hamcrest-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>${assertj-version}</version>
+            <scope>test</scope>
+        </dependency>
 
     </dependencies>
 
--- a/components/camel-micrometer/pom.xml
+++ b/components/camel-micrometer/pom.xml
@@ -45,12 +45,10 @@
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-core</artifactId>
-            <version>${micrometer-version}</version>
         </dependency>
         <dependency>
             <groupId>io.micrometer</groupId>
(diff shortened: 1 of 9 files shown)


oscerd (approved):

ppalaga:
90a4298: rebased

ppalaga:
981e24c: Made the CurrentSpanTest more robust also in the observability component

ppalaga:
13c1a2e: one more attempt to make `CurrentSpanTest.testContextDoesNotLeak()` more robust

ppalaga:
https://github.com/apache/camel/pull/10567 shows that org.apache.camel.opentelemetry.CurrentSpanTest.testContextDoesNotLeak() and org.apache.camel.observation.CurrentSpanTest.testContextDoesNotLeak fail regardless of the management changes in this PR. 
I am not sure what is the problem I have filed https://issues.apache.org/jira/browse/CAMEL-19569 Would you perhaps like to have a look @marcingrzejszczak ?

marcingrzejszczak:
Sure, I'll look into that tomorrow

marcingrzejszczak:
When has this started failing? Apparently there's a flakey test (every now and then trace id after a couple of retries is being reused instead of recreated) and I can reproduce it sometimes on my laptop. I don't recall such things happen when I make the feature so I'm curious what has changed in the meantime

ppalaga:
I am not aware of any substantial change but my stand of knowledge might be rather inaccurate.

marcingrzejszczak:
If you add 

```
@AfterEach
    void noLeakingContext() {
        Assertions.assertSame(Context.root(), Context.current(), "There must be no leaking span after test");
    }
```

into `CamelOpenTelemetryTestSupport` and `CamelMicrometerObservationTestSupport` the `MulticastRouteTest` will start to fail sometimes. Apparently something is wrong in the Camel Tracing module that doesn't clear scopes properly. I'm still looking into that - it's very difficult to replicate and fix

PR #10541, opened 2023-06-29
CAMEL-19285: prevent Kafka client from entering an endless loop

orpiske:
When an authentication error is thrown by Kafka, this could cause an endless loop in the client. Potentially affecting the Kafka broker due to an excessive connection retries in a short period.

This prevents authentication related errors to cause this.


oscerd (approved):

orpiske:
I'll port this to other branches too.

PR #10544, opened 2023-06-29
CAMEL-19285: prevent Kafka client from entering an endless loop

orpiske:
When an authentication error is thrown by Kafka, this could cause an endless loop in the client. Potentially affecting the Kafka broker due to an excessive connection retries in a short period.

This prevents authentication related errors to cause this.


orpiske:
Marked this one as a draft to wait for the 3.21 vote results.

orpiske:
> Marked this one as a draft to wait for the 3.21 vote results.

3.21.0 vote is complete, let's merge.

PR #10569, opened 2023-07-03
CAMEL-19570: Change max.poll.interval.ms type from Long to Integer to…

alex-nt:
… be inline with Kafka config types

# Description

The option for [max.poll.interval.ms](https://camel.apache.org/components/3.20.x/kafka-component.html#_component_option_maxPollIntervalMs) from camel is not the same type as the one from [kafka](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#max-poll-interval-ms). With the latest version we see a nasty type incompatibility error.

# Target

# Tracking
- [x] [CAMEL-19570](https://issues.apache.org/jira/browse/CAMEL-19570)


alex-nt:
I have to admit that the generated code was changed by hand as I couldn't figure out how to generate it. Any guidance about that? Did I miss something along the way?

essobedo:
> I have to admit that the generated code was changed by hand as I couldn't figure out how to generate it. Any guidance about that? Did I miss something along the way?

From the root directory, launch `mvn clean install -pl components/camel-kafka`, it will launch the maven plugin that generates the source code. If it is not enough, you can force the code to be regenerated by removing the folder `components/camel-kafka/src/generated/java/org/apache/camel/`.

alex-nt:
@essobedo Thanks! I was able to confirm that I didn't forget anything generated-code-wise 😄

essobedo (approved):

oscerd (approved):

alex-nt:
What's up with `camel-componentdsl`, it seems to pickup older versions of packages, not the locally built ones. Am I wrong in this observation? It keeps picking up `Long` for my parameter from somewhere, I can't find where.

essobedo:
> What's up with `camel-componentdsl`, it seems to pickup older versions of packages, not the locally built ones. Am I wrong in this observation? It keeps picking up `Long` for my parameter from somewhere, I can't find where.

It needs to be re-generated too, so you need to the same steps as previously described but with `dsl/camel-componentdsl`

oscerd:
> I did try to do so, but it still picks up that thing from the wrong place (or another place). The steps I followed are:
> 
> 1. _camel-kafka_ `mvn clean install -DskipTest`
> 2. _root_ `mvn clean install -pl dsl/camel-componentdsl`
> 
> I (and the build agent) get a lot of weird errors like:
> 
> ```
> camel/dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/AzureFilesComponentBuilderFactory.java:[23,45] package org.apache.camel.component.file.azure does not exist
> 
> /camel/dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/OpensearchComponentBuilderFactory.java:[52,34] cannot find symbol
> [ERROR]   symbol:   class OpensearchComponent
> [ERROR]   location: interface org.apache.camel.builder.component.dsl.OpensearchComponentBuilderFactory
> ```
> 
> And if I look at the regenerated code, it is the old type there.

Try to run a full build.

components/camel-azure/pom.xml
essobedo, line 1: Please exclude this as it is not related

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
essobedo, line 1: Please exclude this as it is not related

oscerd (approved):

alex-nt:
For the failed test, I can either revert the change in the https://github.com/apache/camel/pull/10569/files#diff-7183c76097f802b4e7d5deb6f8ad104e5d3d9e27e6b511dbfa5d108f16d9261b file (I might have been a bit overzealous in changing it *everywhere*) or I can update the version to use `4.0.0-SNAPSHOT` (but that might cause circular dependency issues?). Which one is the preferred option?

alex-nt:
Also, how could I reproduce this locally? `mvn verify` passes for me in that module :(

davsclaus (approved):

PR #10594, opened 2023-07-05
[CAMEL-19575] Fixes camel-rabbitmq - RabbitMQConsumer keeps on consuming even when route shutdown is triggered.

nkg447:
Closing all consumers in RabbitMqConsumer concurrently

# Description

## What
* Suspend RabbitConsumer in RabbitMQConsumer when doSuspend is called. 
* RabbitConsumer cancels the channel on suspend.

## Why
When we have multiple consumers and we stop them one by one. If first consumer is processing a message, then the channel of that consumer will close completely once the message is consumed. If that takes time to complete, then other channels won't stop and continue consuming.

# Target

# Tracking


nkg447:
Found a bug with these changes. The interrupt from DefaultShutdownStrategy won't work coz parallelStream will make separate threads for closing the consumer.

Will fix this and raise a fresh review

nkg447:
found more issues. Will look for better fix. 
Closing this PR for now.

PR #10604, opened 2023-07-06
[CAMEL-19575] Fixes bug in camel-rabbitmq - RabbitMQConsumer keeps on consuming even when route shutdown is triggered.

nkg447:
# Description

# What
cancel all RabbitConsumers in RabbitMQConsumer before trying to close them

# Why
When we have multiple consumers and we stop them one by one. If first consumer is processing a message, then the channel of that consumer will close completely once the message is consumed. If that takes time to complete, then other channels won't stop and continue consuming.

# Target

# Tracking


nkg447:
https://issues.apache.org/jira/browse/CAMEL-19575

davsclaus (approved):

PR #10648, opened 2023-07-10
incremental-build - Replace quick-test with incremental test

essobedo:
## Motivation

With the addition of `quickly` build,  the incremental build can only be interesting to test the affected projects and the projects that depend on them to replace the quick test can only test the affected projects 

## Modifications:

* Replace the quick test with the incremental test
* Remove the `component-test` github action since it is no more needed


oscerd (approved):

essobedo:
@orpiske WDYT of it? As you are the code owner of the Quick Test, I would like to have your point of view before going any further. If you prefer to keep the Quick Test then let's simply remove the incremental build as it doesn't make sense anymore, it could only make sense for the tests.

essobedo:
> * Do we still retain the test summary, so we can view what failed with ease?
It could be improved to add more info but I personally believe that it is quite controversial. Indeed, TBH as a contributor, if the tests pass, I don't really need a comment indicating that the tests passed. For me, it is more like spam than real info because you get this comment for each existing PRs which means more emails received for no real value. IMHO the only potential value is to be notified when the tests ended if you want to merge it quickly but it should not happen so often. And when the tests fail, I know that I will need to check the log anyway, knowing which component failed only makes sense with the quick test as you have different log files IIRC.

But it is only my point of view if other committers find this info helpful, I can try to add it if needed in the incremental test

> * Do we still archive the logs, so we can investigate what has failed?
Yes of course
> * Can we adjust the incremental build/test to comment on the PR with failed status and so on, so both we and the contributors know what has failed?
See above

orpiske (approved):
Thanks for the clarifications, @essobedo. I think we should give it a try and see how it goes. Although we won't have the summaries and reports, we can add then again if we feel the new way is missing information. 

+1 from me.

essobedo:
@orpiske Fair enough, thx 👍

PR #10717, opened 2023-07-18
CAMEL-19568: disable header validation on camel-netty-http

orpiske:
Netty seems to complain about the content-type header when dealing with multipart messages. This disables the header validation so that those headers can go through.


orpiske:
> This will open us to CVE weakness. From http response splitting to other stuff. Validation should be always enabled

Oh, BTW, this applies to the validation of responses only. That is: the responses Camel is sending ... not those it is consuming.

oscerd:
> > This will open us to CVE weakness. From http response splitting to other stuff. Validation should be always enabled
> 
> Oh, BTW, this applies to the validation of responses only. That is: the responses Camel is sending ... not those it is consuming.

I didn't notice this

oscerd (approved):

orpiske:
The failure is unrelated to this change. Let's merge.

PR #10726, opened 2023-07-19
CAMEL-19615: fix chmodDirectory option to not try to change the directory permission for existing directory

michael-salzburg:
# Description

The current chmodDirectory option implementation try to change each directory part and fails if it cannot change the permission. The expected behavior is that only for new directory parts the permission is changed/set.

# Target

# Tracking


michael-salzburg:
typo on pull request. will create a new one

PR #10727, opened 2023-07-19
CAMEL-19615: Fix chmodDirectory option

michael-salzburg:
# Description

The current implementation steps down the directory and try to create each junk. If the chmodDirectory option is set then it will try for each (already existing junk) to change the directory permission and fails.
 
Expected Behaviour:
It should only try to change the permission if the directory does not yet exists

# Target

# Tracking


orpiske (changes requested):
Thanks for your contribution, but the PR looks broken or targeting an incorrect branch.

michael-salzburg:
I have set: apache:camel-3.x. Should I have set a dedicated e.g. 3.18.2, But which one?

oscerd (changes requested):
You need to apply your fix to a fresh branch from camel-3.x

michael-salzburg:
Wrong pull request

PR #10730, opened 2023-07-19
CAMEL-19615 chmodDirectory option try for each junk to change the dir…

michael-salzburg:
# Description

The current implementation steps down the directory and try to create each junk. If the chmodDirectory option is set then it will try for each (already existing junk) to change the directory permission and fails.
 
Expected Behaviour:
It should only try to change the permission if the directory does not yet exists

# Target

# Tracking


oscerd (approved):
LGTM, we should check on 3.20.x and on main too. Thanks @michael-salzburg

michael-salzburg:
> 

i can do. but I am not sure about the workflow (new branch / new PR / .....). (the issue is definitly the same - checked already)

davsclaus (approved):

michael-salzburg:
Hi @davsclaus , is there a reason why its not included/merged also in 3.18.x ?

oscerd:
3.18.x is EOL.

PR #10741, opened 2023-07-19
Upgrade zookeeper 3.8.2

davsclaus:
# Description

# Target

# Tracking


oscerd (approved):

orpiske:
> /component-test camel-zookeeper camel-zookeeper-master

I know @essobedo is on PTO, but leaving this as a hook for a future discussion: I think dependency upgrades are a good use case for us having the `/component-test` again. We haven't thought about that when disabling it ... so, maybe we should add that back.

PR #10774, opened 2023-07-21
(chores) camel-xml-dsl-io: run sequentially

orpiske:
The tests are no thread-safe


orpiske:
We can merge, this only makes these tests less unstable (but they are still broken in some environments).

PR #10821, opened 2023-07-25
CAMEL-19651: camel-kafka - Use default partitioning logic as the default

rgala:
Empty partitioner for default partitioning behavior


essobedo:
Thx for the PR, but could you please create a JIRA ticket first in https://issues.apache.org/jira/projects/CAMEL to describe the problem and to better track it

rgala:
Of course, there you go: [CAMEL-19651] (https://issues.apache.org/jira/browse/CAMEL-19651)

essobedo:
Please use the same commit message as the new PR title.

Regarding the build failure, there are some uncommitted files, so you will need to launch `mvn clean install -Dquickly` from the root of the camel project, see https://camel.apache.org/camel-core/contributing/#_building_camel_4 for more details

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConstants.java
```diff
@@ -61,7 +61,7 @@ public final class KafkaConstants {
 
     public static final String KAFKA_DEFAULT_SERIALIZER = "org.apache.kafka.common.serialization.StringSerializer";
     public static final String KAFKA_DEFAULT_DESERIALIZER = "org.apache.kafka.common.serialization.StringDeserializer";
-    public static final String KAFKA_DEFAULT_PARTITIONER = "org.apache.kafka.clients.producer.internals.DefaultPartitioner";
+    public static final String KAFKA_DEFAULT_PARTITIONER = "";
```
essobedo, line 64: About the change itself, I would rather prefer to have it removed instead of keeping a static field with an empty String as value, but we can wait for other points of views

essobedo:
Since a default behavior change, I believe you will need to update the migration guide too https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/camel-4-migration-guide.adoc

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConstants.java
orpiske: +1 to remove it. 

Just, please, beware that there are a couple of tests referring to this constant. They will need to be adjusted accordingly.

rgala: I have removed the constant

essobedo (approved):
Much better thank you 👍

essobedo:
There are failing tests https://github.com/apache/camel/actions/runs/5659712656/attempts/2#summary-15335168165

Try to launch `mvn clean install -pl components/camel-kafka` from the project root directory

oscerd (approved):

PR #10822, opened 2023-07-25
CAMEL-19539: Replace Thread.sleep in tests

nstefeko:
# Description

Replacing Thread.sleep in tests using Awaitility in camel-netty component. Using Threads.sleep is bug prone and can introduce flakiness when running on environments with different capacities.

# Tracking
https://issues.apache.org/jira/browse/CAMEL-19539


davsclaus:
Thanks, this pattern with awailability can be used in many other components that have hardcoded thread sleeps

PR #10832, opened 2023-07-26
fix multi-value property lost problem

EvanMi:
# Description

the multi-value property is lost

# Target

we should put the key from a multi-value (prefix) in the properties too, or the property may be lost


davsclaus:
Can you point to an example or use-case where you had a problem? And if we agree its a bug then we need a JIRA to track this and for backports too.

EvanMi:
> Can you point to an example or use-case where you had a problem? And if we agree its a bug then we need a JIRA to track this and for backports too.

yeah, 
blabla...
<toD uri="https://camel.apache.org?okStatusCodeRange=200-500&amp;httpClient.socketTimeout=60000&amp;bar=bar"></toD>
blabla...

on the above route, the "bar" param will be add to header, "okStatusCodeRange" will be suitable set, but "httpClient.socketTimeout" will lost.

> Can you point to an example or use-case where you had a problem? And if we agree its a bug then we need a JIRA to track this and for backports too.

yeah, 
blabla...
<toD uri="https://camel.apache.org?okStatusCodeRange=200-500&amp;httpClient.socketTimeout=60000&amp;bar=bar"></toD>
blabla...

on the above route, the "bar" param will be add to header, "okStatusCodeRange" will be suitable set, but "httpClient.socketTimeout" will lost.

davsclaus:
Thanks for that, can you create a JIRA and if possible add an unit test

EvanMi:
> Thanks for that, can you create a JIRA and if possible add an unit test

ok,i'll do both

davsclaus:
Thank you very much

davsclaus (approved):

EvanMi:
> we should put the key from a multi-value (prefix) in the properties too, or the property may be lost

i've add the unit test and JIRA https://issues.apache.org/jira/browse/CAMEL-19663

oscerd (approved):

davsclaus:
Thanks I had to merge this manually to resolve some regens

EvanMi:
> Thanks I had to merge this manually to resolve some regens

ok, but i found you missed the toD.json. or it's really should be "turn of" but not "turn off"

davsclaus:
> > Thanks I had to merge this manually to resolve some regens
> 
> ok, but i found you missed the toD.json. or it's really should be "turn of" but not "turn off"

You need to fix this in the java source code - the json file is auto generated

PR #10835, opened 2023-07-26
(chores) came-jms: avoid failing the test due to session being closed

orpiske:
This reuses the same approach from another commit and applies to another similar test (ref.: 2396b3c4b8d7f1df0b411330e5497fb16bc669d0)


davsclaus (approved):

orpiske:
There are other flakies, so the test is still reporting failures. But these ones are OK, so it's good to merge.

PR #10847, opened 2023-07-26
CAMEL-19491: aws2-sqs - use `ScheduledPollConsumerHealthCheck`

simon-ras:
The SQS consumer is a polling consumer, and thus we can use `ScheduledPollConsumerHealthCheck` where we observe the last poll rather than making a new operation against SQS. Previously `listQueues` was called, which is a different API than what the consumer actually uses (`receiveMessage`).

The `Sqs2ConsumerHealthCheck` used to check if configured region existed in the metadata of the client. That logic was copied over from the old `Sqs2ComponentVerifierExtension`. It is removed from this implementation as the call to `poll` will yield in the same outcome in case an incorrect Region is defined in the configuration. In this case, a failed poll, causing the health check to be DOWN.

# Target

# Tracking


components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
```diff
@@ -375,16 +371,6 @@ protected void doStart() throws Exception {
...
 
         super.doStart();
-
-        // health-check is optional so discover and resolve
-        healthCheckRepository = HealthCheckHelper.getHealthCheckRepository(getEndpoint().getCamelContext(), "components",
-                WritableHealthCheckRepository.class);
-
-        if (healthCheckRepository != null) {
-            consumerHealthCheck = new Sqs2ConsumerHealthCheck(this, getRouteId());
-            consumerHealthCheck.setEnabled(getEndpoint().getComponent().isHealthCheckEnabled() && getEndpoint().getComponent().isHealthCheckConsumerEnabled());
```
simon-ras, line 385: This solution does not honor these configuration options. Should they be removed, or should I override `doBuild()` in order to only add the health check in case it is configured?

davsclaus: Ah these new options are added in v4 - I am moving this to the scheduled poll consumer, so in your PR you can just remove these lines.

davsclaus: https://github.com/apache/camel/pull/10849

simon-ras:
Please be aware that there is a few more aws consumers which could benefit from the same solution e.g. AWS2S3Consumer

oscerd (approved):

davsclaus:
Yes @simon-ras yes can you update the PR with the other aws components that should have done the same as this, then we can have them all in the same PR

simon-ras:
> Yes @simon-ras yes can you update the PR with the other aws components that should have done the same as this, then we can have them all in the same PR

Done, added as separate commits, also updated original jira to explain that other components was affected.

davsclaus (approved):

oscerd (approved):

oscerd:
Thanks for this, having more people reporting and testing AWS/GCP/Azure components is really important.

davsclaus:
org.apache.camel.builder.endpoint.FileConsumeCharsetTest.testConsumeAndDelete | 0.199s | FAILURE
-- | -- | --

davsclaus:
test not related

PR #10867, opened 2023-07-27
CAMEL-19669: isolate heavy tests

orpiske:
- isolate test creating threads outside of the Pool
- isolate tests creating too many threads


oscerd (approved):

orpiske:
Seems to have worked well here, so I think we can merge.

PR #10880, opened 2023-07-28
support more MLLP charsets

plchldr:
# Description

We recently received HL7 messages with `MSH-18` = `UNICODE UTF-8`
This is currently not supported in camel-mllp, which we only became aware of because of a user complaint regarding garbled characters. This is because [`MllpCharsetHelper` falls back to a default charset](https://github.com/apache/camel/blob/camel-3.21.0/components/camel-mllp/src/main/java/org/apache/camel/component/mllp/MllpCharsetHelper.java#L56-L58) (ISO-8859-2).

We expected `MSH-18` to be respected if set, but the current behavior might just be according to spec ...
Anyway; for our use case this simple change suffices.

# Target

# Tracking

(this generated some unrelated changes that I could push, but would rather not)


oscerd (approved):

davsclaus (approved):

davsclaus:
Would be good to backport this to 3.x LTS branches also. And we appreciate a JIRA ticket about this change

davsclaus:
https://issues.apache.org/jira/browse/CAMEL-19677

PR #10885, opened 2023-07-28
(chores) camel-couchbase: disable ITs

orpiske:


orpiske:
These tests are way too heavy to run reliably. Even my build machine struggles with it (and, now, they are also causing problems with the latest testcontainers).

PR #10888, opened 2023-07-28
CAMEL-19675: fixed not copying safe copy properties

orpiske:
This could cause it to lose attachments when using multicast EIP


orpiske:
Waiting for full validation on my own CI.

oscerd (approved):

orpiske:
Passed on my CI. I think we're good to merge.

PR #10889, opened 2023-07-28
CAMEL-19676: camel-core: Do not log sensitive endpoint uri, if route …

davsclaus:
…failed to be created.

# Description

# Target

# Tracking

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
@@ -24,6 +24,7 @@ import jakarta.xml.bind.annotation.XmlTransient;
 import org.apache.camel.Endpoint;
 import org.apache.camel.builder.EndpointProducerBuilder;
 import org.apache.camel.spi.Metadata;
+import org.apache.camel.util.URISupport;
 
 /**
  * Sends the message to an endpoint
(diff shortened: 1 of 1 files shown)


core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
```diff
@@ -106,7 +109,18 @@ public String getPattern() {
 
     @Override
     public String getLabel() {
-        String uri = getEndpointUri();
+        if (endpointUriToString == null) {
+            String value = null;
+            try {
+                value = getEndpointUri();
+            } catch (RuntimeException e) {
+                // ignore any exception and use null for building the string value
```
essobedo, line 117: Why do we need to catch the exception now knowing it was not needed before?

essobedo (approved):

core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
davsclaus: Thats how the code is done elsewhere, its just for a lavel

davsclaus: See DefaultEndpoint I think it is

PR #10895, opened 2023-07-28
CAMEL-19675: added test case for aggregation strategy with attachment

rvanderhallen:
Test case for attachment being removed from the Exchange after AggregationStrategy


orpiske:
This one is gonna fail without my fixup patch on #10888. We are good to merge.

orpiske (approved):

PR #10928, opened 2023-07-31
Backport CAMEL-19675 to Camel 3.20.x

orpiske:


oscerd (approved):

components/camel-attachments/src/test/java/org/apache/camel/attachment/MulticastAggregationStrategyTest.java
```diff
@@ -0,0 +1,65 @@
+package org.apache.camel.attachment;
```
davsclaus, line 1: You need license header

orpiske: Thanks. The license header was added in a separate commit and the chery-pick missed. Let me add that one.

PR #10964, opened 2023-08-02
CAMEL-19695: camel-google-bigquery - Bump org.json to 20230618 (3.20)

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19695 for (3.20)

## Motivation

The component `camel-google-bigquery` indirectly depends on `org.json:json:jar:20200518:compile` which has a know CVE https://nvd.nist.gov/vuln/detail/CVE-2022-45688 that can be fixed by upgrading it to `20230227` or higher.

## Modifications:

* Add `org.json:json:20230618` to the dependencyManagement section
* Replace `org.json:json:20200518` with `org.json:json:20230618` in `camel-google-bigquery`

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -348,6 +348,7 @@
     <jslt-version>0.1.11</jslt-version>
     <jsmpp-version>2.3.11</jsmpp-version>
     <json-api>1.0</json-api>
+    <json-org-version>20230618</json-org-version>
     <json-patch-version>1.13</json-patch-version>
     <json-path-version>2.8.0</json-path-version>
     <json-schema-validator-version>2.2.14</json-schema-validator-version>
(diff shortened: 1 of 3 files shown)


oscerd (changes requested):
This always puzzled me, org.json json dependency has a category X license. We should avoid declaring it directly in our POM.

oscerd:
> This always puzzled me, org.json json dependency has a category X license. We should avoid declaring it directly in our POM.

But they changed to public domain from a while and this version is fine: https://github.com/stleary/JSON-java/blob/20230618/LICENSE

essobedo:
> > This always puzzled me, org.json json dependency has a category X license. We should avoid declaring it directly in our POM.
> 
> But they changed to public domain from a while and this version is fine: https://github.com/stleary/JSON-java/blob/20230618/LICENSE

A good remark and a good catch! 😅

PR #10975, opened 2023-08-03
(chores) camel-xmpp: fix most of the test log output to stdout

orpiske:


orpiske:
We are good to merge. The failure is unrelated to this change. There are problems in running the XMPP tests on GitHub, as we can see from this bogus PR: #10976.

PR #10981, opened 2023-08-03
 CAMEL-19647: use 503 for health check DOWN

LostArtist:
- Added unity test to verify changes

### DIFF
--- a/components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/MainHttpServerTest.java
+++ b/components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/MainHttpServerTest.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/MainHttpServerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void statusIsNotSatisfied() throws IOException, InterruptedException {
+        MainHttpServer server = new MainHttpServer();
+
+        camelContext = new DefaultCamelContext();
+        server.setCamelContext(camelContext);
+
+        camelContext.getRegistry().bind("fake", new MainHttpFakeHealthCheck());
+
+        server.setHost("0.0.0.0");
+        server.setPort(8080);
```
orpiske, line 46: An oversight on my part: we could probably have used the `AvailablePortFinder`.

components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/MainHttpServerTest.java
```diff
@@ -0,0 +1,62 @@
...
+        server.setPath("/");
+
+        server.setHealthCheckEnabled(true);
+        server.start();
+
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(URI.create("http://localhost:8080/q/health/ready"))
```
orpiske, line 53: And we'll need to adjust here once we use the `AvailablePortFinder`.

orpiske (changes requested):
Marking as request changes as we forgot to use the `AvailablePortFinder` on the test.

orpiske:
I'll fix the UT separately in this case - we need this patch for GA.

PR #11006, opened 2023-08-04
[CAMEL-19709] yaml dsl: include additional information such as description and title in the jscon schema

lburgazzoli:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-api/src/generated/java/org/apache/camel/spi/annotations/YamlProperty.java
+++ b/core/camel-api/src/generated/java/org/apache/camel/spi/annotations/YamlProperty.java
@@ -28,7 +28,11 @@ public @interface YamlProperty {
 
     String type();
 
+    String defaultValue() default "";
+
     String format() default "";
+
+    String displayName() default "";
 
     String description() default "";
 
--- a/core/camel-api/src/generated/java/org/apache/camel/spi/annotations/YamlType.java
+++ b/core/camel-api/src/generated/java/org/apache/camel/spi/annotations/YamlType.java
@@ -28,6 +28,12 @@ public @interface YamlType {
 
     String[] nodes() default {};
 
+    String displayName() default "";
+
+    String description() default "";
+
+    boolean deprecated() default false;
+
     int order() default 0;
 
     boolean inline() default false;
--- a/dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/src/generated/java/org/apache/camel/dsl/yaml/deserializers/ModelDeserializers.java
+++ b/dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/src/generated/java/org/apache/camel/dsl/yaml/deserializers/ModelDeserializers.java
@@ -253,10 +253,13 @@ public final class ModelDeserializers extends YamlDeserializerSupport {
             inline = true,
             types = org.apache.camel.model.dataformat.ASN1DataFormat.class,
             order = org.apache.camel.dsl.yaml.common.YamlDeserializerResolver.ORDER_LOWEST - 1,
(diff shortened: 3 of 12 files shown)


davsclaus:
@lburgazzoli you have some commits in this PR from seda and others that are not related. Maybe you can rebase and update this PR.

lburgazzoli:
oh damn, I thought I had removed them, will rebase tomorrow

lburgazzoli:
@davsclaus done

davsclaus:
Should 

    "deprecated" : "false",

Not be a json boolean type (according to spec) ???

    "deprecated" : false,

davsclaus:
The order of type, description, title, etc seems to not be strict/consistent. I wonder if we can make it the same order everywhere. Sometimes description is before title, and other times its after etc.

This file will be regen more often when we fix typo, add new option, change a default etc. So it would be good to ensure the generator is 100% on all OS (not like. that XML XSD that is a bit of a pain)

lburgazzoli:
@davsclaus let me have a look

lburgazzoli:
> Should
> 
> ```
> "deprecated" : "false",
> ```
> 
> Not be a json boolean type (according to spec) ???
> 
> ```
> "deprecated" : false,
> ```

done

lburgazzoli:
> The order of type, description, title, etc seems to not be strict/consistent. I wonder if we can make it the same order everywhere. Sometimes description is before title, and other times its after etc.

I have changed the json schema generation to be consistent, at least with what Jackson's ObjectMapper allows to configure. Let me know if you spot any  inconsistency

davsclaus:
Hmm jackson still dumps it in different order. They should use a linked map so the values are in the order they are added.

davsclaus:
Our own library `camel-util-json` has a basic JsonObject / JsonArray which is consistent. And `JSooner` has prettyPrint methods to dump

davsclaus (approved):

lburgazzoli:
> Hmm jackson still dumps it in different order. They should use a linked map so the values are in the order they are added.

can you give me an example ?

davsclaus:
https://github.com/apache/camel/blob/3907d9991010aa7cbda1a8bb1f647911458a6cac/core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogTracer.java#L350

davsclaus:
All the dev consoles uses it as well, see core/camel-console 

And mind that if you have some text then you can use the escape method to make it json safe (but you very likely do not need that, we use it to dump source code in json etc.)

davsclaus:
Its likely a bunch of work to change so maybe create a JIRA ticket and put it as v4.1.

lburgazzoli:
I mean, an example of a unordered property :)

But I think I got it so basically there were a discrepancy in the order the properties were set for types and related properties but Jackson did serialize them according to the insertion order. It should be fixed now. 

I'll try to use camel-util-json later as it would mean changing a number of things now :)

lburgazzoli:
> Its likely a bunch of work to change so maybe create a JIRA ticket and put it as v4.1.

https://issues.apache.org/jira/browse/CAMEL-19717

lburgazzoli:
ok, I think now the PR is now ready.

I did some investigation for https://issues.apache.org/jira/browse/CAMEL-19717 but I found that there is an issue with escaping https://issues.apache.org/jira/browse/CAMEL-19718 which seems not super easy to fix if we want to preserve the option to escapes forward slashes.

I will digg into that a little bit more in the coming days

davsclaus:
LGTM

@lburgazzoli we need to merge this today so it has time to get into 4.0 GA

PR #11026, opened 2023-08-07
Add apache-camel to release so we build source tarball

davsclaus:
# Description

# Target

# Tracking


components/camel-xmpp/src/test/java/org/apache/camel/component/xmpp/integration/XmppMultiUserChatIT.java
```diff
@@ -27,7 +27,7 @@
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
 @DisabledIfSystemProperty(named = "ci.env.name", matches = "github.com",
-        disabledReason = "Github environment has trouble running the XMPP test container and/or component")
+                          disabledReason = "Github environment has trouble running the XMPP test container and/or component")
```
orpiske, line 30: I think these files are not meant to be part of this PR, right?

davsclaus: Yes they are part of regen, which likely is already done on main, or if not will get merge together with this PR

PR #11027, opened 2023-08-07
CAMEL-19712: adapt to support Junit 5.10

jono-morris:
When calling JUnit's argument converter we now need to also pass a ParameterContext from which the converter retrieves a classloader that it passes to the resolved converter class:

https://github.com/junit-team/junit5/blob/r5.10.0/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java#L95

The Converter assumes it's converting parameters for Methods/constructors (retrieving the classloader from an Executable).  However, since we convert for @Parameter fields this change passes in an Executable from the ParameterizedTemplate constructor as a standin.


orpiske:
@jono-morris Thanks for the contribution!!!!

@davsclaus I'm wondering if we'll want to squeeze this one in 4.0 or if we hold a bit and push it to 4.1 ... any thoughts?

orpiske:
Also, c/c @oscerd for his thoughts.

orpiske (changes requested):
The PR is OK. Just marking as request changes to see if we want to push to 4.1 or not.

davsclaus:
Yeah lets wait for Camel 4.1 with this upgrade.

All other frameworks like SB, Quarkus and whatnot have not upgraded and released version with JUnit 5.10.x. Its too new.
And this has global affect on all tests of Camel and end users as well that heavily use JUnit

davsclaus:
In the mean time then here is a quick one to help with
https://issues.apache.org/jira/browse/CAMEL-19627

orpiske (approved):

orpiske:
As the `main` is open for 4.1 work, I think we can merge this.

PR #11053, opened 2023-08-08
CAMEL-12317: Fix naming convention for header constants

vignesh-manel:
# Description

Fixed naming convention

# Target

# Tracking


davsclaus:
We did you see that very shothand naming convention ?

vignesh-manel:
The aws sdk client itself uses shorthand convention https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sfn/SfnClient.html

So I thought header constants will have a consistent and short names, just a suggestion, if we could use this, as the previous names which were used in code were not consitent

vishalsingh2972:
is this issue still open @davsclaus ?

davsclaus:
The naming convention in Camel is for headers to have their component name as prefix.

So if you wanted to use `sfn` then the component should been named `camel-aws2-sfn`

PR #11061, opened 2023-08-09
CAMEL-19518: replaced Thread.sleep() in camel-disruptor tests

LostArtist:


components/camel-disruptor/src/test/java/org/apache/camel/component/disruptor/SedaDisruptorCompareTest.java
```diff
@@ -232,7 +233,7 @@ void speedTestDisruptor(
         System.out.println("Starting real test of: " + componentName);
 
         forceGC();
-        Thread.sleep(1000);
+        Awaitility.await().atMost(1, TimeUnit.SECONDS);
```
orpiske, line 236: This is missing the actual check to performed. Probably you need to put `untilAsserted` or something like that.

orpiske (changes requested):

orpiske (approved):

PR #11092, opened 2023-08-11
WIP: [CAMEL-19698] [CAMEL-19699] camel-yaml-dsl: schema: Restrict language…

igarashitm:
…s/dataformats to be mutually exclusive

* checkpoint: Added `oneOf` property on @YamlProperty annotation. This is to hold an oneOf grouping name to show those are mutually exclusive on deserializer annotation. The schema generator should then use these to generate oneOf restriction.

Expected change on YAML DSL schema (expression as an example):
https://github.com/igarashitm/misc/blob/main/camel/json-schema-yaml-validator/src/test/resources/schema/whenAndExpression.json
Validation Tests:
https://github.com/igarashitm/misc/blob/main/camel/json-schema-yaml-validator/src/test/groovy/com/github/igarashitm/camel/ExpressionTest.groovy

# TODO
 - [ ] Update GenerateYamlSchemaMojo to use `oneOf` and generate revised schema
 - [ ] Tests

# Target

# Tracking
https://issues.apache.org/jira/browse/CAMEL-19698
https://issues.apache.org/jira/browse/CAMEL-19699
Also see: https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-19700

Cc: @lburgazzoli


igarashitm:
Closing as I hit a major problem on proposed schema change, which has the one of inline or explicit expression required, but the fact is that they're optional. I couldn't find a way to express it with schema composition that works with current validator + draft4.

PR #11102, opened 2023-08-14
CAMEL-19656 batch visibility extender task

jono-morris:
Processes visibility extension entries of all exchanges submitted to the consumer using ChangeMessageVisibilityBatchRequest run as a single submission to changeMessageVisibilityBatch on the Amazon SqsClient. 

Added TimeoutExtender class that holds the extension entries and runs/shuts-down the TimeoutTask.

ChangeMessageVisibilityBatchRequest sample code from amazon:  https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/javav2/example_code/sqs/src/main/java/com/example/sqs/VisibilityTimeout.java#L102


davsclaus:
Thanks for helping with this. But I would like to do this in a more stable way without all the complexity of schedule/cancel a task.

Instead we just have a single background task that runs periodically every X interval. It has an internal queue with the current exchanges that are inflight. Then if the queue is empty it does nothing. Otherwise it sends a batch request to AWS.

Also how how long time to extend can just be a fixed amount, as the task is scheduled to run also in fixed interval.
Then we can have 2 options the end user can control.

- delay between extend task runs
- how long time to extend

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
```diff
@@ -404,14 +434,13 @@ public void cancel() {
         @Override
         public void run() {
             if (run.get()) {
-                ChangeMessageVisibilityRequest.Builder request
-                        = ChangeMessageVisibilityRequest.builder().queueUrl(getQueueUrl()).visibilityTimeout(repeatSeconds)
-                                .receiptHandle(exchange.getIn().getHeader(Sqs2Constants.RECEIPT_HANDLE, String.class));
+                ChangeMessageVisibilityBatchRequest.Builder request
+                        = ChangeMessageVisibilityBatchRequest.builder().queueUrl(getQueueUrl()).entries(entries);
```
simon-ras, line 438: Please be aware that the request can take at most 10 entries accordingly to the AWS documentation. `entries` at this point could be more than that.

davsclaus: Thanks then we should split the batch into group of 10, and then if there are 25, then we do 3 calls.

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
```diff
@@ -404,32 +410,44 @@ public void cancel() {
...
-                } catch (SqsException e) {
-                    if (e.getMessage().contains("Message does not exist or is not available for visibility timeout change")) {
+
+                List<ChangeMessageVisibilityBatchRequestEntry> entries = new LinkedList<>();
+
+                while (!requestQueue.isEmpty()) {
+
+                    // up to 10 requests can be sent with each ChangeMessageVisibilityBatch action
+                    while (!requestQueue.isEmpty() && entries.size() < MAX_REQUESTS) {
+                        entries.add(requestQueue.poll());
```
simon-ras, line 420: The entry should only be removed from this queue once it has been processed by the consumer.

The current solution will only extend the message a single time. Also I cannot see where you remove the entry from the extend queue once it has been fully processed.

I think it would be better if the consumer itself had a collection of in-flight exchanges - and then the extender would iterate that collection. The consumer then has to remove the exchanges once fully processed.

davsclaus: the UoW on the exchange should remove the exchange, then its done when the exchange is fully complete

simon-ras: Well, I appreciate the improvements :tophat:

davsclaus:
@simon-ras do you have any feedback on the updated code ?

simon-ras (approved):

PR #11106, opened 2023-08-14
CAMEL-19684: reduce flakiness in camel-rocketmq

orpiske:


orpiske:
Doesn't fully solve the problem, but helps a bit.

davsclaus (approved):

PR #11114, opened 2023-08-16
CAMEL-19751: disable misbehaved camel-xchange tests

orpiske:
The camel-xchange tests keep hitting the address https://fapi.binance.com/fapi/v1/exchangeInfo which both abuse the owner of the address and contribute to their flakiness.


orpiske:
Note: these tests are already disabled on ASF CI, but those running their own CI may notice flakiness on these.

davsclaus (approved):

PR #11122, opened 2023-08-16
CAMEL-19754: remove code checking for older Java versions

orpiske:


davsclaus (approved):

orpiske:
The failure is caused by a flaky. It should be good to merge.

PR #11134, opened 2023-08-17
(chores) camel-util: simplify StringHelper capitalize/decapitalize methods

orpiske:


orpiske:
Even faster than #11127

```
Benchmark                                          Mode  Cnt  Score    Error  Units
StringHelperTest.testCapitalize                    avgt   10  0.032 ±  0.001  us/op
StringHelperTest.testDashToCamelCaseNegative       avgt   10  0.019 ±  0.001  us/op
StringHelperTest.testDashToCamelCasePositive       avgt   10  0.023 ±  0.001  us/op
```

PR #11145, opened 2023-08-17
Camel-19557: added assertions to camel-core tests

LostArtist:


orpiske:
Regarding the first set of changes ... 

I think catching an `AssertionError` is not an adequate test practice. That deviates from the actual purpose of the AssertionError which is to indicate that the test has failed. 

For now, let's keep the original as is. Let's create a ticket on Apache Jira to note that this is a bad practice that must be fixed.

orpiske (changes requested):

orpiske (approved):

orpiske:
Test failure is unrelated to this.

PR #11153, opened 2023-08-20
CAMEL-17027 stream into buffer of configured size

jono-morris:
Read the message body in chunks instead of reading the entire body into memory to better handle large payloads.  

Introduced a new option 'buffersize' that controls how much of the message body is read for a single chunk with a default size set to 1000000, which is the same as the batchsize meaning that one chunk will be uploaded as one part.

Further PR to follow to address class thread-safety.


oscerd (changes requested):
Lgtm, but I would like to have a deeper look this week. So wait for me, before merging

oscerd (approved):
Looks good. No need to wait.

davsclaus (approved):

PR #11165, opened 2023-08-21
CAMEL-19760: netty-http:prevent the usage of proxy protocol in produc…

luigidemasi:
…er endpoint

### DIFF
--- a/components/camel-netty-http/src/main/java/org/apache/camel/component/netty/http/NettyHttpEndpoint.java
+++ b/components/camel-netty-http/src/main/java/org/apache/camel/component/netty/http/NettyHttpEndpoint.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.component.netty.http;
 
+import java.net.URI;
 import java.util.Map;
 
 import org.apache.camel.AsyncEndpoint;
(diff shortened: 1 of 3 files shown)


components/camel-netty-http/src/main/java/org/apache/camel/component/netty/http/NettyHttpProducer.java
```diff
@@ -51,6 +53,10 @@ public NettyHttpProducer(NettyHttpEndpoint nettyEndpoint, NettyConfiguration con
 
     @Override
     protected void doInit() throws Exception {
+        if (isProxyProtocol()) {
```
davsclaus, line 56: I wonder if this can this be checked in createProducer instead?

Also we do not log and throw exception, only throw exception

davsclaus (changes requested):

components/camel-netty-http/src/main/java/org/apache/camel/component/netty/http/NettyHttpProducer.java
luigidemasi: > I wonder if this can this be checked in createProducer instead?
yeah, it's better there. moved.
> Also we do not log and throw exception, only throw exception
ok, removed the log.

davsclaus (approved):

PR #11181, opened 2023-08-23
Camel19772

davsclaus:
# Description

# Target

# Tracking


oscerd (approved):

core/camel-xml-io/src/main/java/org/apache/camel/xml/LwModelToXMLDumper.java
```diff
@@ -258,4 +283,55 @@ private static NamespaceAware getNamespaceAwareFromExpression(ExpressionNode exp
...
+                for (Map.Entry<String, Object> entry : b.getProperties().entrySet()) {
+                    String key = entry.getKey();
+                    Object value = entry.getValue();
+                    buffer.write(String.format("            <property key=\"%s\" value=\"%s\"/>%n", key, value));
+                }
+                buffer.write(String.format("        </properties>%n"));
+            }
+            buffer.write(String.format("    </bean>%n"));
+        }
+    }
```
orpiske, line 335: @davsclaus this is a little bit off-topic but, I think it's something we should start thinking about for the future (maybe we'll need to create a ticket about it). 

We have a lot of places with code-to-be-generated that are built-in in the code base. IMHO, this is extremely hard to read, maintain and debug. At some point in 4.x, I believe, we should start investigating how to move this to some kind of template library.

orpiske: I created a ticket to take note and simplify the discussion: https://issues.apache.org/jira/browse/CAMEL-19779

PR #11184, opened 2023-08-23
(chores) camel-main: use a LongAdder to reduce contention

orpiske:


core/camel-main/src/main/java/org/apache/camel/main/MainDurationEventNotifier.java
```diff
@@ -75,6 +74,7 @@ public void notify(CamelEvent event) throws Exception {
     }
 
     protected void doNotify(CamelEvent event) {
+        LOG.info("Notify running by thread: {}", Thread.currentThread());
```
orpiske, line 77: This shouldn't be here.

davsclaus (approved):

PR #11188, opened 2023-08-23
CAMEL-19742: created base test class in camel-test-infra-core

LostArtist:
added dependencies to pom files


LostArtist:
Still working on it. Created PR to avoid loosing code

orpiske:
> Still working on it. Created PR to avoid loosing code

Thanks for kicking it off.

FYI, I marked it as a draft, as this is a WIP and we want to avoid this merging inadvertently.

PR #11206, opened 2023-08-25
CAMEL-19789: Fix handling of shardIterator so that all records are re…

klease:
…turned.

According to the AWS Kinesis client documentation, "use a GetShardIterator request to get the first shard iterator for use in your first GetRecords request and for subsequent reads use the shard iterator returned by the GetRecords request in NextShardIterator". Adapt the unit tests for this case and modify the KinesisConsumerIT test to verify that different records are returned.

# Description

# Target

# Tracking


klease:
I don't have a real-life use case with Kinesis to test so any comments are more than welcome, especially related to handling of closed shards, which according to the documentation, appears to only happen in the case of merging or splitting an existing shard.

oscerd (approved):
I think it should go on camel-4.0.x too

davsclaus:
Yes I have backported to 4.0.x

klease:
Thanks Claus.

PR #11225, opened 2023-08-29
CAMEL-19398: refactor the core type-converter

orpiske:
This is preliminary work that should simplify moving away from an Exception-based logic for type handling

It refactors the type conversion code so that it is preloaded to a Map, which store type conversion pairs. This map is checked whenever there is a type conversion, by matching the requested pair (from/to) with the known conversions on the Map.

Among other things, it reduces the incidence of runtime type detection and tries to let the compiler decide the method whenever it is possible

This builds on top of the work done on #10271.

### DIFF
--- a/components/camel-vertx/camel-vertx-common/src/generated/java/org/apache/camel/component/vertx/common/CamelVertxCommonBulkConverterLoader.java
+++ b/components/camel-vertx/camel-vertx-common/src/generated/java/org/apache/camel/component/vertx/common/CamelVertxCommonBulkConverterLoader.java
@@ -9,6 +9,7 @@ import org.apache.camel.Ordered;
 import org.apache.camel.TypeConversionException;
 import org.apache.camel.TypeConverterLoaderException;
 import org.apache.camel.TypeConverter;
+import org.apache.camel.converter.TypeConvertible;
 import org.apache.camel.spi.BulkTypeConverters;
 import org.apache.camel.spi.TypeConverterLoader;
 import org.apache.camel.spi.TypeConverterRegistry;
@@ -43,6 +44,7 @@ public final class CamelVertxCommonBulkConverterLoader implements TypeConverterL
     @Override
     public void load(TypeConverterRegistry registry) throws TypeConverterLoaderException {
         registry.addBulkTypeConverters(this);
+        doRegistration(registry);
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


orpiske:
Some performance details: 
- Testing with a content-based-router, routing based on the body of a message, this code was faster than 4.0.0 in 100% of the times 
- Testing with a routing-slip,  with a bean to dynamically route the exchanges, this code was faster than 4.0.0 in about 83% of the times
- For seda and controlbus, the performance from previous version seems approximately the same 
- There seems to be a small performance drop with the disruptor component (but I plan to investigate further)

orpiske:
Using the tests from camel-performance-tests repo:
| Test | 4.0.0 | This Patch | 
|------|-----|------|
| SlowTypeConverter | 35 ms/op | 7 ms/op |
| FastTypeConverterTest | 35 ms/op | 7 ms/op |

core/camel-api/src/main/java/org/apache/camel/converter/ConverterFunction.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.camel.converter;
+
+import org.apache.camel.Exchange;
+
+@FunctionalInterface
```
davsclaus, line 22: Add javadoc

components/camel-vertx/camel-vertx-common/src/generated/java/org/apache/camel/component/vertx/common/CamelVertxCommonBulkConverterLoader.java
```diff
@@ -165,6 +167,41 @@ private Object doConvertTo(Class<?> from, Class<?> to, Exchange exchange, Object
...
+        registry.addConverter(new TypeConvertible<>(io.vertx.core.json.JsonArray.class, java.lang.String.class), this);
+        registry.addConverter(new TypeConvertible<>(io.vertx.core.json.JsonObject.class, java.lang.String.class), this);
+        registry.addConverter(new TypeConvertible<>(io.vertx.core.json.JsonArray.class, java.util.List.class), this);
+        registry.addConverter(new TypeConvertible<>(io.vertx.core.json.JsonObject.class, java.util.Map.class), this);
+        registry.addConverter(new TypeConvertible<>(io.vertx.core.buffer.Buffer.class, org.apache.camel.StreamCache.class), this);
+        
+        
+    }
+
     public TypeConverter lookup(Class<?> to, Class<?> from) {
```
davsclaus, line 205: I assume the old code to lookup using switch statements then is no longer needed and will later be removed?

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
```diff
@@ -0,0 +1,192 @@
...
+     * mapping of all known converters)
+     *
+     * @param  toType     the type to convert to
+     * @param  fromType   the type to convert from
+     * @param  converters the map of all known converters
+     * @return            the type converter or null if unknown
+     */
+    static TypeConverter doLookup(
+            Class<?> toType, Class<?> fromType, Map<TypeConvertible<?, ?>, TypeConverter> converters) {
+        return doLookup(new TypeConvertible<>(fromType, toType), converters);
```
davsclaus, line 59: The downside of this is that a new object is created per lookup - the old code was not creating any new objects during convertion, which means no overhead on GC. But the new code is faster due to that JDK bug and other parts, so we take speed over GC

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
```diff
@@ -0,0 +1,192 @@
...
+        if (typeConverter != null) {
+            return typeConverter;
+        }
+
+        final TypeConverter superConverterTc = tryMatch(typeConvertible, converters);
+        if (superConverterTc != null) {
+            return superConverterTc;
+        }
+
+        final TypeConverter arrayConverterTc = tryObjectArrayConverters(typeConvertible, converters);
```
davsclaus, line 76: Maybe you can eager check if from.isArray to avoid calling a few methods to then do the check that late.

davsclaus: Converting to array is seldom in use

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Try to resolve the TypeConverter by forcing a costly and slow recursive check.
+     *
+     * @param  typeConvertible the type converter pair
+     * @param  converters      the map of all known converters
+     * @return                 the type converter or null if unknown
+     */
+    static TypeConverter tryMatch(
+            TypeConvertible<?, ?> typeConvertible, Map<TypeConvertible<?, ?>, TypeConverter> converters) {
```
davsclaus, line 166: All this stream filter stuff does have a bit of overhead in the JDK - sometimes code with just a for loop is faster.

components/camel-vertx/camel-vertx-common/src/generated/java/org/apache/camel/component/vertx/common/CamelVertxCommonBulkConverterLoader.java
orpiske: There's still a lot of code relying on the lookup and it seems that we have some features that require the lookup behavior (i.e.: fallback promotion [ seems to be used for beans], CXF converter, etc). 

For now, I am being conservative and I am aiming to be as compliant/compatible with the old code as possible to avoid introducing more changes than I can handle. 

So, I am thinking about leaving this one for this first patch. Then slowly start to replacing usages of the lookup (or clean it up - at least). 

What do you think?

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
orpiske: I'm going to give it a try with a plain for loop and see how it goes.

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
orpiske: Maybe we can remove that one? 🤔 I'll give it a try

core/camel-api/src/main/java/org/apache/camel/converter/ConverterFunction.java
orpiske: Thanks. This file was unused - leftover from the first version of this patch. I removed it.

components/camel-vertx/camel-vertx-common/src/generated/java/org/apache/camel/component/vertx/common/CamelVertxCommonBulkConverterLoader.java
orpiske: Saving this for the subsequent patch.

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
orpiske: It seems we can remove this, so this code is not longer valid.

core/camel-base/src/main/java/org/apache/camel/impl/converter/TypeResolverHelper.java
orpiske: There's indications that the plain loop is better as it creates less objects and all that. So, I changed to a loop as you suggested.

orpiske:
Marking as skip-tests because `camel-maven-package-plugin` is severely broken on GitHub.

davsclaus (approved):

PR #11228, opened 2023-08-29
Camel-Tracing: Removing NsqSpanDecorator since component is no more a…

oscerd:
…vailable

# Description

# Target

# Tracking


oscerd:
@marcingrzejszczak I'm looking at the span decorator available and I do think we could add some more. Do you have any suggestions on this particular task? Thanks.

marcingrzejszczak:
@oscerd I don't exactly follow. What is the concrete question? Are you trying to add more span decorators?

oscerd:
Yes, I meant to ask if you have idea of what component decorators make more sense in your experience

marcingrzejszczak:
I mean all components can be interesting to the users. Preferably we should ask the users what are they most interested in or which telemetry data they feel that they are currently missing.

oscerd:
@davsclaus I do think this is could be a good approach (asking users) for improving the decorators' coverage.

davsclaus:
Yes that would be good to have more of them

marcingrzejszczak:
Another option would be to go through https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/ to see which conventions are defined today (remember that these are not stable so let's not ever say that we are OTel semantic convention compatible because if we do we will have to keep on doing breaking changes) and try to double check if our instrumentations are setting similar meta data.

davsclaus:
@marcingrzejszczak that is a great idea - we should make these camel components use same convention if possible. I have created a JIRA: https://issues.apache.org/jira/browse/CAMEL-19812

davsclaus:
Yes its okay, we will add notes in our upgrade guides if tags are renamed, and can also do dual names to keep the deprecated for a number of releases.

oscerd:
I think it would be beneficial to have a look at what we have as decorators today and do a little example to what we want to have

PR #11229, opened 2023-08-29
CAMEL-19731 AS2: Fixed null pointer exception 

dk2k:
... when request's header dispositionNotificationTo isn't set

Added a unit test for this case. The existing tests are green

### DIFF
--- a/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/entity/EntityParser.java
+++ b/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/entity/EntityParser.java
@@ -444,52 +444,56 @@ public final class EntityParser {
             try {
                 // Determine Content Type of Message
                 String contentTypeStr = HttpMessageUtils.getHeaderValue(message, AS2Header.CONTENT_TYPE);
-                ContentType contentType = ContentType.parse(contentTypeStr);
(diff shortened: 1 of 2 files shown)


components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/entity/EntityParser.java
```diff
@@ -444,52 +444,56 @@ public static void parseAS2MessageEntity(HttpMessage message) throws HttpExcepti
...
                 String contentTypeStr = HttpMessageUtils.getHeaderValue(message, AS2Header.CONTENT_TYPE);
-                ContentType contentType = ContentType.parse(contentTypeStr);
-
-                // Determine Charset
-                String charsetName = StandardCharsets.US_ASCII.name();
-                Charset charset = contentType.getCharset();
-                if (charset != null) {
-                    charsetName = charset.name();
-                }
+                if (contentTypeStr != null) {
```
essobedo, line 447: Will it work too if we simply add `if (contentTypeStr == null) { return; }`?

dk2k: it works, I will add a commit with this change

oscerd (approved):

davsclaus (approved):

PR #11240, opened 2023-08-30
Blueprint

davsclaus:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-xml-io/src/main/java/org/apache/camel/xml/in/BaseParser.java
+++ b/core/camel-xml-io/src/main/java/org/apache/camel/xml/in/BaseParser.java
@@ -373,7 +373,7 @@ public class BaseParser {
         }
 
         String pn = parser.getName();
-        boolean match = Objects.equals(name, pn) || Objects.equals(name2, pn) || Objects.equals(name3, pn);;
+        boolean match = Objects.equals(name, pn) || Objects.equals(name2, pn) || Objects.equals(name3, pn);
         if (!match || !matchNamespace(namespace, parser.getNamespace(), null, false)) {
             return ""; // empty tag
         }
--- a/core/camel-xml-io/src/main/java/org/apache/camel/xml/in/BaseParser.java
+++ b/core/camel-xml-io/src/main/java/org/apache/camel/xml/in/BaseParser.java
@@ -16,29 +16,6 @@
  */
 package org.apache.camel.xml.in;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.Reader;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Base64;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.function.Consumer;
-import java.util.function.Supplier;
-
-import javax.xml.XMLConstants;
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
-
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.Text;
-
 import org.apache.camel.LineNumberAware;
 import org.apache.camel.model.language.ExpressionDefinition;
 import org.apache.camel.spi.NamespaceAware;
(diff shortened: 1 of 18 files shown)


essobedo:
Interesting feature, I'm just wondering in which case we could need both spring and blueprint beans? Maybe replacing spring beans with something more generic like "external beans" could be good enough, WDYT?

davsclaus:
they have different namespace so the model needs to be that naming, but the xml-io-dsl should moved the code into their own module (there is a JIRA) eg camel-xml-io-springxml camel-xml-io-blueprint

Also the point is not to run 100% but to assist users to migrate away via jbang

core/camel-xml-io/src/main/java/org/apache/camel/xml/in/BaseParser.java
```diff
@@ -368,6 +367,20 @@ protected String getNextTag(String name, String name2) throws XmlPullParserExcep
         return pn;
     }
 
+    protected String getNextTag(String name, String name2, String name3) throws XmlPullParserException, IOException {
+        if (parser.nextTag() != XmlPullParser.START_TAG) {
+            throw new XmlPullParserException("Expected starting tag");
+        }
+
+        String pn = parser.getName();
+        boolean match = Objects.equals(name, pn) || Objects.equals(name2, pn) || Objects.equals(name3, pn);;
```
essobedo, line 376:
```suggestion
        boolean match = Objects.equals(name, pn) || Objects.equals(name2, pn) || Objects.equals(name3, pn);
```

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/xml/blueprint/BlueprintXmlBeansHandler.java
```diff
@@ -0,0 +1,226 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Used for parsing and discovering legacy OSGi <blueprint> XML to make it runnable on camel-jbang, and for tooling to
+ * migrate this to modern Camel DSL in plain Camel XML or YAML DSL.
+ */
+public class BlueprintXmlBeansHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SpringXmlBeansHandler.class);
```
essobedo, line 50:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(BlueprintXmlBeansHandler.class);
```

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/xml/blueprint/BlueprintXmlBeansHandler.java
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+    }
+
+    protected String extractValue(CamelContext camelContext, String val, boolean resolve) {
+        // blueprint placeholder prefix
+        if (val != null && val.contains("${")) {
+            Matcher matcher = BLUEPRINT_PATTERN.matcher(val);
+            while (matcher.find()) {
+                String group = matcher.group(1);
```
essobedo, line 175: Unless I miss something the group is the same value as the match, so maybe the group is not even needed in the regular expression

essobedo: Or it could be replaced by `Pattern.compile("\\$\\{(.*?)}")` to avoid calling `substring`

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/xml/spring/SpringXmlBeansHandler.java
```diff
@@ -0,0 +1,318 @@
...
+ */
+public class SpringXmlBeansHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SpringXmlBeansHandler.class);
+    private static final Pattern SPRING_PATTERN = Pattern.compile("(\\$\\{.*?})"); // non-greedy mode
+
+    // when preparing spring-based beans, we may have problems loading classes which are provided with Java DSL
+    // that's why some beans should be processed later
+    private final List<String> delayedBeans = new LinkedList<>();
+    private Set<String> infraBeanNames;
```
essobedo, line 75: should be set and final otherwise line # 204 could throw an NPE

essobedo (approved):

orpiske (approved):

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/xml/blueprint/BlueprintXmlBeansHandler.java
davsclaus: Its the same code from elsewhere where we grab spring placeholder -> camel placeholder.

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/language/ExpressionReifier.java

davsclaus: Yes good idea to move the group so its the value as-is and avoid the substring

PR #11255, opened 2023-08-31
CAMEL-17027 make camel-aws-s3 threadsafe so it can be used concurrently

jono-morris:
Small exchanges are batched when individual exchanges are less than the batch-size, requiring processing state information to held between processing of each exchange. 

This state information has been moved to a State class that is guarded by synchronized blocks whenever it is access so that exchanges can be processed asynchronously.   

Asynchronous versions of existing unit tests have been added. 

Large files are still chunked down and sent as chunks, and state information is confined to the processing method.


oscerd (approved):

davsclaus (approved):

oscerd:
Do we need to backport to 4.0.x too?

PR #11279, opened 2023-09-01
CAMEL-19815: use a ConcurrentHashMap instead of a blanket synchronize

orpiske:
This should help offset the performance impacts of removing the cached QueueReference on d66469d6ac7243f0ae790c41a500965b885d3180

**4.1 (this patch)**
```
Benchmark                                         Mode  Cnt    Score    Error   Units
SedaRoundTripTest.sendBlocking                   thrpt   10  925.275 ±  1.146  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes  thrpt   10  229.778 ±  0.107  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes   avgt   10    0.004 ±  0.001   ms/op
SedaRoundTripTest.sendBlocking                      ss   10    0.163 ±  0.035   ms/op
SedaRoundTripTest.sendBlockingWithMultipleTypes     ss   10    0.570 ±  0.047   ms/op
```

**4.1 (without the patch)**
```
Benchmark                                         Mode  Cnt    Score    Error   Units
SedaRoundTripTest.sendBlocking                   thrpt   10  915.329 ±  0.320  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes  thrpt   10  225.853 ±  0.459  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes   avgt   10    0.005 ±  0.001   ms/op
SedaRoundTripTest.sendBlocking                      ss   10    0.167 ±  0.025   ms/op
SedaRoundTripTest.sendBlockingWithMultipleTypes     ss   10    0.626 ±  0.222   ms/op
```

**4.0 (baseline)**
```
Benchmark                                         Mode  Cnt    Score    Error   Units
SedaRoundTripTest.sendBlocking                   thrpt   10  922.456 ±  0.462  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes  thrpt   10  242.381 ±  0.107  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes   avgt   10    0.004 ±  0.001   ms/op
SedaRoundTripTest.sendBlocking                      ss   10    0.162 ±  0.022   ms/op
SedaRoundTripTest.sendBlockingWithMultipleTypes     ss   10    0.573 ±  0.044   ms/op
```

### DIFF
--- a/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaComponent.java
+++ b/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaComponent.java
@@ -19,6 +19,7 @@ package org.apache.camel.component.seda;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.camel.Component;
 import org.apache.camel.Endpoint;
@@ -53,7 +54,7 @@ public class SedaComponent extends DefaultComponent {
     @Metadata(label = "consumer,advanced", defaultValue = "1000")
     private int defaultPollTimeout = 1000;
 
-    private final Map<String, QueueReference> queues = new HashMap<>();
+    private final Map<String, QueueReference> queues = new ConcurrentHashMap<>();
     private final Map<String, Integer> customSize = new HashMap<>();
 
     public SedaComponent() {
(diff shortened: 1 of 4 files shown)


components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaComponent.java
```diff
@@ -189,10 +186,27 @@ public synchronized QueueReference getOrCreateQueue(
         log.debug("Created queue {} with size {}", key, size);
 
         // create and add a new reference queue
-        ref = new QueueReference(queue, size, multipleConsumers);
+        QueueReference ref = new QueueReference(queue, size, multipleConsumers);
         ref.addReference(endpoint);
         getQueues().put(key, ref);
```
essobedo, line 191: Maybe we should consider creating the queue and the queue reference in a computeIfAbsent to prevent concurrent creation for the same key

essobedo:
Maybe I misunderstood your results, but I don't see a real significant improvement, could you please clarify a bit your results?

orpiske:
Marking it as a draft as I want to go through @essobedo's comments next week.

essobedo:
Another question, Is there a real-life use case where we heavily create queues? otherwise, I don't really get the real need of improving it.

Do you have profiler reports indicating that this part of the code as an hot spot?

orpiske:
> Maybe I misunderstood your results, but I don't see a real significant improvement, could you please clarify a bit your results?

The key ones are the top two on each table. They are measured in terms of number of operations per millisecond. Even though the number seems small, when spread over the course of a few minutes/hours, this results in quite a lot more exchanges going through the core.

essobedo:
You mean that we need to compare `925 ops/ms` for 4.1 with `922 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlocking` and `229 ops/ms` for 4.1 with `242 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlockingWithMultipleTypes`, right?

components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaComponent.java
```diff
@@ -189,10 +186,27 @@ public synchronized QueueReference getOrCreateQueue(
...
+                                               + " does not match given queue size " + size);
+        }
+        // add the reference before returning queue
+        ref.addReference(endpoint);
+
+        if (log.isDebugEnabled()) {
+            log.debug("Reusing existing queue {} with size {} and reference count {}", key, size, ref.getCount());
+        }
         return ref;
     }
```
rhuan080, line 212: Can the 'synchronized' keyword in the 'registerQueue' method also be eliminated? I believe you can utilize 'computeIfAbsent,' as suggested by @essobedo  in the comment above.

orpiske:
> Another question, Is there a real-life use case where we heavily create queues? otherwise, I don't really get the real need of improving it.

The `getQueueReference` is in the hot path for the SedaProducer` `addToQueue`.

> Do you have profiler reports indicating that this part of the code as an hot spot?

Absolutely.

Screenshot 2023-09-01 at 17 06 51
Screenshot 2023-09-01 at 17 07 03

orpiske:
> You mean that we need to compare `925 ops/ms` for 4.1 with `922 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlocking` and `229 ops/ms` for 4.1 with `242 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlockingWithMultipleTypes`, right?

Exactly. This tries to bring the code closer to what we had in 4.0, as I had to revert one of the performance improvements we had there. 

BTW, answering your other question: "Is there a real-life use case where we heavily create queues". 

The answer is yes, we do. Basically, that patch for 4.0 was breaking a scenario for an user that was heavily creating queues.

essobedo:
Interesting, by chance, do you have the html version of the reports to allow me to zoom in?

orpiske:
Obs.: please, do feel free to review. I'll go through your review comments next week. For now just answering the non-review questions as my brain is basically a jelly after this week.

orpiske:
> Interesting, by chance, do you have the html version of the reports to allow me to zoom in?

Here's one: http://angusyoung.org/arquivos/tmp/profile-camel-19815-with-perf-v6-4.1.0-SNAPSHOT-2023-09-01-043702.html


Obs.: just keep in mind that this is just one of the many scenarios I run (and generated from the same JMH results I pasted - it's just the only one I have readily available). If you want a different one (i.e.: uncontended SEDA or etc), drop me a note next week and I can send it to you (or point to how you can generate your own).

essobedo:
> > You mean that we need to compare `925 ops/ms` for 4.1 with `922 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlocking` and `229 ops/ms` for 4.1 with `242 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlockingWithMultipleTypes`, right?
> 
> Exactly. This tries to bring the code closer to what we had in 4.0, as I had to revert one of the performance improvements we had there.
> 
> BTW, answering your other question: "Is there a real-life use case where we heavily create queues".
> 
> The answer is yes, we do. Basically, that patch for 4.0 was breaking a scenario for an user that was heavily creating queues.

If so if you check well the second case is better with Camel 4.0 and the first one is slightly better with Camel 4.1 but you have a higher Error `1.146` vs `0.462` which means that the results are more stable in Camel 4.0. 

I'm wondering if an improvement of less than 1 % is insignificant enough to be considered a real improvement as the results may vary a lot from one test to another. 

BTW, I've just realized that you [set forks to 1](https://github.com/apache/camel-performance-tests/blob/main/tests/camel-jmh/src/test/java/org/apache/camel/itest/jmh/SedaRoundTripTest.java#L55), maybe you should retry with 3 to get more reliable results?

orpiske:
> > > You mean that we need to compare `925 ops/ms` for 4.1 with `922 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlocking` and `229 ops/ms` for 4.1 with `242 ops/ms` for 4.0 in case of `SedaRoundTripTest.sendBlockingWithMultipleTypes`, right?
> > 
> > 
> > Exactly. This tries to bring the code closer to what we had in 4.0, as I had to revert one of the performance improvements we had there.
> > BTW, answering your other question: "Is there a real-life use case where we heavily create queues".
> > The answer is yes, we do. Basically, that patch for 4.0 was breaking a scenario for an user that was heavily creating queues.
> 
> If so if you check well the second case is better with Camel 4.0 and the first one is slightly better with Camel 4.1 but you have a higher Error `1.146` vs `0.462` which means that the results are more stable in Camel 4.0.
> 
> I'm wondering if an improvement of less than 1 % is insignificant enough to be considered a real improvement as the results may vary a lot from one test to another.
> 
> BTW, I've just realized that you [set forks to 1](https://github.com/apache/camel-performance-tests/blob/main/tests/camel-jmh/src/test/java/org/apache/camel/itest/jmh/SedaRoundTripTest.java#L55), maybe you should retry with 3 to get more reliable results?

Normally that would be the case (i.e.: if we were comparing the results from the same test class). However, these tests are already running on forked executions: the iterations for different versions run separately (as in, the automation runs the tests for one version and then for another - not on the same JVM, I mean).

essobedo:
> > Interesting, by chance, do you have the html version of the reports to allow me to zoom in?
> 
> Here's one: http://angusyoung.org/arquivos/tmp/profile-camel-19815-with-perf-v6-4.1.0-SNAPSHOT-2023-09-01-043702.html
> 
> Obs.: just keep in mind that this is just one of the many scenarios I run (and NOT generated from the same JMH results I pasted - it's just the only one I have readily available). If you want a different one (i.e.: uncontended SEDA or etc), drop me a note next week and I can send it to you (or point to how you can generate your own).

Thx for sharing do you have the same kind of report for 4.0 (the baseline)?

orpiske:
Let's put this one on pause for now. I tried with the suggestions, but the performance wasn't better:

*4.1.0-SNAPSHOT Updated*
```
SedaRoundTripTest.sendBlocking                   thrpt   10  914.608 ±  0.545  ops/ms
SedaRoundTripTest.sendBlockingWithMultipleTypes  thrpt   10  232.482 ±  0.419  ops/ms
```

I'll get back to this one with a clearer head in a couple of weeks, after I dispatch other higher priority items.

PR #11285, opened 2023-09-01
CAMEL-19783: Added reconnection, even if a message has not yet been received

jsight:
# Description

This makes the following changes:
 - Reconnect automatically with sensible defaults if we get an error before receiving the first message
 - Expose an API for updating the initialReplayId. At first, this might sound silly. "initialReplayId" is effectively a constant, right? But there's a notable exception in cases where the client restarts the route dynamically and may pull a new replay id from another source. This can happen in the case of clustering, where only one route is active at a time and failover may occur.
 - Expose an API to force plaintext via the consumer. This is for a similar reason to the existing method, but also makes it available to other test frameworks. We use a fake grpc endpoint to test the service and this will simplify that work.
 

# Target

- [x ] I checked that the commit is targeting the correct branch (note that Camel 3 uses `camel-3.x`, whereas Camel 4 uses the `main` branch)

# Tracking

- [ x] I checked that each commit in the pull request has a meaningful subject line and body.

- [ x] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes


oscerd (approved):

oscerd:
Can you open the same PR for 4.0.x @jsight ? Thank you.

jsight:
@oscerd Backported here: https://github.com/apache/camel/pull/11308

oscerd (approved):

PR #11305, opened 2023-09-05
CAMEL-19837: let the converter decide the appropriate datatype for conversion

orpiske:


orpiske:
@JiriOndrusek in cases like this, it seems more appropriate to let the converter decide in runtime what datatype is more suitable for conversion. 

Do you mind having a look, please?

oscerd (approved):

JiriOndrusek (approved):

PR #11307, opened 2023-09-05
CAMEL-19833: camel-bindy - race condition in BindyAbstractFactory.link()

luigidemasi:


components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/BindyAbstractFactory.java
```diff
@@ -133,7 +133,7 @@ public abstract void bind(CamelContext camelContext, List<String> data, Map<Stri
     /**
      * Link objects together
      */
-    public void link(Map<String, Object> model) {
+    public synchronized void link(Map<String, Object> model) {
```
essobedo, line 136: If we use a second `BindyCsvDataFormat` in our route on the same model but with different parameters will it be good enough to fix the problem? Generally speaking, I'm wondering if it should not be fixed at `ReflectionHelper` level by simply not restoring the flag. WDYT?

luigidemasi: It was my first idea to fix it at ReflectionHelper level but I wanted my changes to impact as less as possible on other components, but yeah the scenario you described, even uncommon, can be a problem. I will change the commit.

core/camel-util/src/main/java/org/apache/camel/util/ReflectionHelper.java
```diff
@@ -177,7 +177,7 @@ public static Method findMethod(Class<?> clazz, String name, Class<?>... paramTy
         return null;
     }
 
-    public static void setField(Field f, Object instance, Object value) {
+    public synchronized static void setField(Field f, Object instance, Object value) {
```
essobedo, line 180: It is not exactly what I had in mind, I was more thinking of removing the code that restores the old value of the flag this way we don't need any synchronization mechanism but we can wait for comments from others

essobedo:
@luigidemasi Did you see my last comment?

components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/fixed/link/BindyRaceConditionLinkTest.java
```diff
@@ -0,0 +1,95 @@
...
+                Path destination = Paths.get(tempDir.toString() + File.separator + source.getFileName());
+                Files.copy(source, destination, REPLACE_EXISTING);
+            }
+        }
+    }
+
+    @Override
+    public boolean isUseRouteBuilder() {
+        return true;
+    }
```
essobedo, line 74: Do we really need it? I think it is the default value

components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/fixed/link/BindyRaceConditionLinkTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private void copyFiles() throws IOException, URISyntaxException {
+        try (Stream<Path> stream = Files.list(Path.of(ClassLoader.getSystemResource(SOURCE_DIRECTORY_PATH).toURI()))) {
+            List<Path> paths = stream.filter(path -> path.toString().endsWith(".csv")).toList();
+            for (Path source : paths) {
+                Path destination = Paths.get(tempDir.toString() + File.separator + source.getFileName());
+                Files.copy(source, destination, REPLACE_EXISTING);
+            }
+        }
```
essobedo, line 68: Do we really need 10 big CSV files to reproduce? Maybe we could have only one file, split it into lines then parallelize the processing with `.split(body()).delimiter("\n").parallelProcessing()`, WDYT?

components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/model/fixed/link/MyModel.java
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+package org.apache.camel.dataformat.bindy.model.fixed.link;
+
+import org.apache.camel.dataformat.bindy.annotation.CsvRecord;
+import org.apache.camel.dataformat.bindy.annotation.DataField;
+import org.apache.camel.dataformat.bindy.annotation.Link;
+
+@CsvRecord(separator = ",", skipField = true)
+public class MyModel {
```
essobedo, line 25: To increase the chance of collision, maybe we should only have one single private field instead of many different fields, don't you agree?

core/camel-util/src/main/java/org/apache/camel/util/ReflectionHelper.java
```diff
@@ -179,20 +179,10 @@ public static Method findMethod(Class<?> clazz, String name, Class<?>... paramTy
...
         try {
-            boolean oldAccessible = false;
-            try {
-                oldAccessible = f.canAccess(instance);
-            } catch (Exception e) {
-                // ignore
-            }
-            boolean shouldSetAccessible = !Modifier.isPublic(f.getModifiers()) && !oldAccessible;
-            if (shouldSetAccessible) {
+            if (!Modifier.isPublic(f.getModifiers()) && !f.canAccess(instance)) {
```
essobedo, line 182: Yes, that sounds better. Maybe, we could do the same with `getField`, don't you agree?

components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/fixed/link/BindyRaceConditionLinkTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @RepeatedTest(3)
+    public void raceConditionTest() throws Exception {
+
+        Path filePath = Path.of(ClassLoader.getSystemResource(SOURCE_CSV_FILE_PATH).toURI());
+        String csv = Files.readString(filePath);
+        template.requestBody("direct:bindy-link-test", csv);
+
+        MockEndpoint end = getMockEndpoint("mock:end");
+        MockEndpoint fail = getMockEndpoint("mock:fail");
+        fail.expectedMessageCount(0);
```
essobedo, line 49: This makes me wonder if the test is properly written because if you expect no message, the condition is already met even before starting the Camel route. You should rather expect that `mock:end` receives all the messages. I think that MockEndpoint should not be needed

components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/fixed/link/BindyRaceConditionLinkTest.java
```diff
@@ -0,0 +1,100 @@
...
+            @Override
+            public void configure() {
+                BindyCsvDataFormat bindy = new BindyCsvDataFormat(MyModel.class);
+
+                ExecutorService executorService = new ThreadPoolExecutor(
+                        THREAD_POOL_CORE_SIZE,
+                        THREAD_POOL_MAX_SIZE,
+                        THREAD_POOL_KEEP_ALIVE_TIME,
+                        SECONDS,
+                        new ArrayBlockingQueue<>(THREAD_POOL_QUEUE_SIZE));
```
essobedo, line 83: Did you check if your test fails without your custom `ExecutorService` and the bug fix? If you really need it make sure that it is shutdown properly at the end of the test to free up the memory. I can't remember if is done out of the box by Camel.

essobedo (commented):
Just one last question/remark left, thank you for your patience

essobedo (approved):
Good Job thank you 👍

oscerd (approved):

PR #11313, opened 2023-09-06
CAMEL-19840: camel-core-api - Warn when the key store cannot be found

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19840 for 3.20

## Motivation

If `KeyStoreParameters` are configured with an incorrect path for the key store file, at runtime, we end up with an error that can be more or less easy to understand depending on the underlying component for which SSL has been configured.

## Modifications:

* Add a warning when the key store file cannot be found


oscerd (approved):
Does it works fine on 3.21.x, 3.x, 4.0.x and main? Thanks @essobedo

essobedo:
No, we have the same problem in other versions too. I will propose a PR for them too.

orpiske (approved):

PR #11316, opened 2023-09-06
CAMEL-19840: camel-core-api - Warn when the key store cannot be found

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19840 for 3.21

## Motivation

If `KeyStoreParameters` are configured with an incorrect path for the key store file, at runtime, we end up with an error that can be more or less easy to understand depending on the underlying component for which SSL has been configured.

## Modifications:

* Add a warning when the key store file cannot be found


oscerd (approved):

essobedo:
The failure is not related

PR #11321, opened 2023-09-06
CAMEL-19398: use Java's builtin custom boolean parser

orpiske:
Note: this changes the behavior for converting string to boolean when invalid data is passed. Previously the code would return `null`. With this, it defaults to `false`. 

Obs.: keeping as a draft as I would like @davsclaus opinion on this (and he is on PTO).

### DIFF
--- a/core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
+++ b/core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
@@ -175,32 +175,13 @@ public abstract class CoreTypeConverterRegistry extends ServiceSupport implement
         return (T) doConvertToAndStat(type, exchange, value, false);
     }
 
-    // must be 4 or 5 in length
     private static Boolean customParseBoolean(String str) {
-        int len = str.length();
-        // fast check the value as-is in lower case which is most common
-        if (len == 4) {
-            if ("true".equals(str)) {
-                return Boolean.TRUE;
-            }
-
-            if ("TRUE".equals(str.toUpperCase())) {
-                return Boolean.TRUE;
-            }
-
-            return null;
+        if ("true".equalsIgnoreCase(str)) {
+            return Boolean.TRUE;
         }
 
-        if (len == 5) {
-            if ("false".equals(str)) {
-                return Boolean.FALSE;
-            }
-
-            if ("FALSE".equals(str.toUpperCase())) {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/util/CamelContextHelperTest.java
```diff
@@ -45,7 +45,7 @@ public void testParsing() {
         eq(() -> CamelContextHelper.parseBoolean(context, "false"), false);
         eq(() -> CamelContextHelper.parseBoolean(context, "TrUe"), true);
         eq(() -> CamelContextHelper.parseBoolean(context, "FaLsE"), false);
-        fl(() -> CamelContextHelper.parseBoolean(context, "5"), IllegalArgumentException.class);
+        eq(() -> CamelContextHelper.parseBoolean(context, "5"), false);
```
davsclaus, line 48: Yeah this is bad - it should not be false as the input is invalid. We basically need to just check for true|false as only valid input - as the old code.

A lot of Camel use is configuration, so if you have a mistake there, then concerting to false would make it appear as working but something was not configured correctly.

davsclaus: If you want to use Boolean.parseBoolean, then you can do that first, and only check for invalid input if the result was false.

var a = Boolean.parseBoolean(s)
if a == false {
   .. // okay the input may be invalid, so check for that
}

davsclaus (changes requested):

core/camel-core/src/test/java/org/apache/camel/util/CamelContextHelperTest.java
davsclaus: Or maybe do what Boolean.parseBoolean does but manually, something ala

if ("true".equalsIgnoreCase(s)) return Boolean.TRUE
"false".equalsIgnoreCase(s) return Boolean.FALSE
return null

orpiske:
@davsclaus thanks for sharing your insights about this one! 

So, I think the way it is, this patch is too risky: it would break backwards compatibility and introduce shady behavior. 

I'll take a look at your proposed changes and modify it. I think it will be enough of a cleanup by doing just that.

orpiske:
This new changes use a simplified boolean conversion as suggested (while retaining the old behavior). I think this should be safer.

davsclaus (approved):

PR #11331, opened 2023-09-07
CAMEL-19842: fix not creating/refreshing seda queues

orpiske:
- Removing the cached queue references caused a regression in camel-zookeeper-master. This brings back the cached references (reverts: d66469d6ac7243f0ae790c41a500965b885d3180)
- Avoid keeping cached stalled references (ref: CAMEL-19815)


essobedo (approved):

oscerd (approved):
We still have the WireTap failing, otherwise the full build looks really good.

orpiske:
@oscerd yeah, I am on top of that one.

orpiske:
Obs.: we need to backport this to 4.0.1 was well. The same test is also failing there (for the same reasons).

PR #11342, opened 2023-09-08
Upgrade sshd 2.10.0 (#11338) (#11340)

coheigea:
Final backmerge to fix https://avd.aquasec.com/nvd/2023/cve-2023-35887/


essobedo:
@coheigea For the next backports like this one, maybe you should rather cherry-pick the commit of your branch instead of the merge commit in order to avoid having commit messages of type "... (#11338) (#11340)"

PR #11343, opened 2023-09-08
Fix incorrect URI format in camel-grpc docs

jamesnetherton:


essobedo:
Skipping the tests as it doesn't affect the code, thx for the fix

PR #11354, opened 2023-09-10
CAMEL-19812 use open telemetry naming conventions

jono-morris:
Updated camel-tracing to use open telemetry semantic conventions https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/

Skipped AzureServiceBusSpanDecorator as I couldn't find suitable conventions for the majority of the metadata tags using the current spec, e.g. enqueuedTime, expiresAt, partitionKey, replyToSessionId etc.


oscerd (approved):
@marcingrzejszczak

components/camel-tracing/src/main/java/org/apache/camel/tracing/SpanAdapter.java
```diff
@@ -23,24 +23,12 @@ public interface SpanAdapter {
 
     void setError(boolean error);
 
-    void setTag(Tag key, String value);
```
marcingrzejszczak, line 26: These are breaking changes, are we ok with that?

components/camel-tracing/src/main/java/org/apache/camel/tracing/SpanAdapter.java
```diff
@@ -23,24 +23,12 @@ public interface SpanAdapter {
...
-
-    void setTag(Tag key, Number value);
-
     void setTag(String key, String value);
 
     void setTag(String key, Number value);
 
     void setTag(String key, Boolean value);
 
-    default void setLowCardinalityTag(Tag key, String value) {
```
marcingrzejszczak, line 36: These are breaking changes, are we ok with that?

components/camel-tracing/src/main/java/org/apache/camel/tracing/TagConstants.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.tracing;
+
+public class TagConstants {
```
marcingrzejszczak, line 19: Are these actually different to what we had before? If that's the case that's a breaking change. IMO we should introduce new tags next to the current ones and deprecate the old ones. 

Also how will ensure the compatibility between this class and semantic conventions of OpenTelemetry? Are we to be compatible with all the versions? Which version is this compatible with, for which date? Semantic Conventions are not stable and they can be changed at any point in time. What's our strategy for such breakages?

oscerd (changes requested):
Let's wait for some more comments around what we want to do.

jono-morris:
We could associate new attributes with the existing Tag class.  This would avoid breaking changes and provide the option to deprecate tags in future. So, when we use a tag we could either use the Tag class name and/or the associated otel attribute. 

See: https://github.com/apache/camel/pull/11381

The attributes are from v1.21.0 semantic conventions, most of which were already mapped in OpenTelemetrySpanAdapter.

components/camel-tracing/src/main/java/org/apache/camel/tracing/SpanAdapter.java
davsclaus: Yeah Camel end users dont really work directly with this API. And we have migration docs where we can write the steps people need to migrate. And its better to get ahead of this early in Camel v4 than later.

components/camel-tracing/src/main/java/org/apache/camel/tracing/SpanAdapter.java
davsclaus: See next comment

components/camel-tracing/src/main/java/org/apache/camel/tracing/TagConstants.java
davsclaus: However these Tags are potentially more used by Camel end users and if possible we could have the old tags deprecated and if possible let the old tags delegate to the new tags.

components/camel-tracing/src/main/java/org/apache/camel/tracing/SpanAdapter.java
marcingrzejszczak: If that's fine for you that's fine for me too 😃

davsclaus:
@jono-morris thanks for working on this - we added some comments above. Not sure if you get a chance to continue working on this, it seems we may "just" need for a while to deprecate the old TagConstants (and if possible link them to the new ones, so users upgrading it would work). And then after the next LTS release we can remove those deprecations.

jono-morris:
@davsclaus thanks for that, it makes sense.  I'm hoping to finish this over the next day or two.

PR #11381, opened 2023-09-13
CAMEL-19812 add otel aligned attributes

jono-morris:
Following on from comments left on https://github.com/apache/camel/pull/11354

Demo adding otel attributes to the Tag Enum allowing either the Tag name to be used and/or the Tag attribute when setting a span attribute.

### DIFF
--- a/components/camel-observation/src/main/java/org/apache/camel/observation/MicrometerObservationSpanAdapter.java
+++ b/components/camel-observation/src/main/java/org/apache/camel/observation/MicrometerObservationSpanAdapter.java
@@ -55,6 +55,7 @@ public class MicrometerObservationSpanAdapter implements SpanAdapter {
     @Override
     public void setTag(Tag key, String value) {
         this.observation.highCardinalityKeyValue(key.toString(), value);
+        this.observation.highCardinalityKeyValue(key.getAttribute(), value);
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


components/camel-observation/src/main/java/org/apache/camel/observation/MicrometerObservationSpanAdapter.java
```diff
@@ -80,11 +81,13 @@ public void setTag(String key, Boolean value) {
     @Override
     public void setLowCardinalityTag(Tag key, String value) {
         observation.lowCardinalityKeyValue(key.toString(), value);
```
marcingrzejszczak, line 83: Do we want to deprecate this? If we want to migrate to the attributes version we could deprecate all of the `key.toString()` options and mark that this is scheduled for removal in e.g. 3 minor releases? WDYT?

components/camel-tracing/src/main/java/org/apache/camel/tracing/Tag.java
```diff
@@ -17,13 +17,39 @@
...
-    COMPONENT,
-    HTTP_STATUS,
-    HTTP_METHOD,
-    HTTP_URL,
-    MESSAGE_BUS_DESTINATION,
-    DB_TYPE,
-    DB_INSTANCE,
-    DB_STATEMENT,
-    ERROR
+    COMPONENT("component"),
```
marcingrzejszczak, line 20: This looks great!

marcingrzejszczak (approved):
I like this PR! I've added an additional comment

PR #11395, opened 2023-09-14
(chores) camel-components: use unmodifiable lists when possible

orpiske:


oscerd (approved):

orpiske:
The failure is unrelated, so it should be safe to merge.

PR #11396, opened 2023-09-14
(chores) camel-tooling: use unmodifiable lists when possible

orpiske:


orpiske:
The failure is unrelated, it should be safe to merge.

PR #11397, opened 2023-09-14
CAMEL-19859: camel-spring-rabbitmq set headerfilter not effective

TUCJVXCB:
# Description
if i use java code to config spring-rabbitmq component headerfilterstrategy, the doinit() method of SpringRabbitMQComponent will be executed before i config headerfilterstrategy, so the DefaultMessagePropertiesConverter may reference headerfilterStrategy the old value。

# Target

# Tracking


orpiske (changes requested):
Thanks for your contribution! I do have a request, though: can you please adjust the commit message so it's in the format `CAMEL-19859: <text>`. 

It makes it easier for us when browsing the commit history 😇

PR #11402, opened 2023-09-15
CAMEL-19859: camel-spring-rabbitmq set headerfilter not effective #11397

TUCJVXCB:
# Description

if i use java code to config spring-rabbitmq component headerfilterstrategy, the doinit() method of SpringRabbitMQComponent will be executed before i config headerfilterstrategy, so the DefaultMessagePropertiesConverter may reference headerfilterStrategy the old value。

# Target

# Tracking


oscerd (changes requested):
You re-opened the PR on 4.0.x, but it should be on main. Then we eventually backport.

TUCJVXCB:
> You re-opened the PR on 4.0.x, but it should be on main. Then we eventually backport.

ok, I re-PR

PR #11434, opened 2023-09-18
CAMEL-19870 Camel AS2: Should accept MDN field name Disposition as case insensitive

isakkirajan:
# Description

- The third party AS2 server returning the MDN response with disposition field case insensitive which is as per the spec (https://www.ietf.org/rfc/rfc4130.txt) but camel as2 validates this field with case sensitive hence it fails.
- This fix is already applied to main branch and this PR is a backport to camel 3.21.x branch.

# Target

# Tracking


oscerd (approved):

davsclaus:
We have backported this to the other branches, so you do not have to make more PRs

PR #11454, opened 2023-09-19
Fix google-sheets-stream component OAuth

christophd:
Stream component was using default scope https://www.googleapis.com/auth/spreadsheets when connecting. Apparently the scope alone is not sufficient to perform proper OAuth authentication and some more scopes are required. Instead of using this default scope in streams component do not specify any scope at all (same as google-sheets component is doing) in order to fix authentication. In case users need to manage scopes while connecting all required scopes need to be given.


christophd:
should be backported to 4.0.x

oscerd (approved):
Looks good, please open a PR on 4.0.x too. Thanks.

davsclaus (approved):

davsclaus:
The PR fails as there are changed files not in the commit.

christophd:
> The PR fails as there are changed files not in the commit.

yes, thanks. I have rebuilt the camel-catalog module and added the files to the PR

davsclaus:
Can you create a JIRA ticket so we can track this bug for 4.0.1 change-log.

christophd:
See https://issues.apache.org/jira/browse/CAMEL-19885

PR #11456, opened 2023-09-19
CAMEL-19398: cleanup type statistics

orpiske:
Move the statistics code to separate classes and interfaces


oscerd (approved):

davsclaus:
Hmm the type converter is a very critical path, and by moving this code, then the JVM needs to optimize to avoid the method call when statistics is not enabled. Which is 99% of the case.

orpiske:
> Hmm the type converter is a very critical path, and by moving this code, then the JVM needs to optimize to avoid the method call when statistics is not enabled. Which is 99% of the case.

Yes! My idea is to implement a no-op statistics in a subsequent PR, so that there's a better chance of it being removed by JVM's dead-code elimitation routine

davsclaus (approved):

PR #11457, opened 2023-09-19
CAMEL-19846: yaml and xml DSL - Add support for BeanLoader so they ca…

davsclaus:
…n define beans that can be used for advanced global configuration on camel context which the legacy spring/blueprint XML DSL can do.

# Description

# Target

# Tracking

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/BeanLoader.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/BeanLoader.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.spi;
+
+/**
+ * SPI for loading beans from a {@link Resource}.
+ *
+ * For example with YAML or XML DSL.
+ */
+public interface BeanLoader {
+
+    /**
+     * The loaded bean which typically has been bound to the {@link Registry}.
+     *
+     * @param name name of the bean
+     * @param bean the bean instance
+     */
+    void onLoadedBean(String name, Object bean);
+
+}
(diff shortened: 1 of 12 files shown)


oscerd (approved):

igarashitm:
@davsclaus so what I learned from @grgrzybek is that this is already supported by:

1. Register beans before DefaultConfigurationConfigurer searches
https://github.com/apache/camel/blob/main/dsl/camel-xml-io-dsl/src/main/java/org/apache/camel/dsl/xml/io/XmlRoutesBuilderLoader.java#L83-L102

3. DefaultConfigurationConfigurer search in bean registry, if CamelContext related bean is found, use them 
https://github.com/apache/camel/blob/main/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java#L894

It's already working for XML IO
 https://github.com/apache/camel-kamelets-examples/blob/main/jbang/app-routes-spring-beans/app.xml#L20-L22

So once YamlRoutesBuilderLoader#preParseRoute() supports loading beans into bean registry, this should work for YAML DSL as well.

@grgrzybek could you chime in if there's something to add?

grgrzybek:
See my comment under https://issues.apache.org/jira/browse/CAMEL-19846 ;)
Camel changes very fast, so I'd have to catch up a bit again. But when I was preparing https://github.com/apache/camel-kamelets-examples/blob/main/jbang/app-routes-spring-beans/app.xml, it worked ;)

davsclaus:
> @davsclaus so what I learned from @grgrzybek is that this is already supported by:
> 
> 1. Register beans before DefaultConfigurationConfigurer searches
>    https://github.com/apache/camel/blob/main/dsl/camel-xml-io-dsl/src/main/java/org/apache/camel/dsl/xml/io/XmlRoutesBuilderLoader.java#L83-L102
> 2. DefaultConfigurationConfigurer search in bean registry, if CamelContext related bean is found, use them
>    https://github.com/apache/camel/blob/main/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java#L894
> 
> It's already working for XML IO https://github.com/apache/camel-kamelets-examples/blob/main/jbang/app-routes-spring-beans/app.xml#L20-L22
> 

This is only when using spring-beans (eg real spring dependency injection). If you just have 100% camel xml files (routes and beans) then spring is NOT used at all, and this PR would be needed to have beans working for that - just like in YAML DSL.

> So once YamlRoutesBuilderLoader#preParseRoute() supports loading beans into bean registry, this should work for YAML DSL as well.
> 
> @grgrzybek could you chime in if there's something to add?

igarashitm:
I also looked at it with debugger on main branch and the preParseRoute() is still invoked before DefaultConfigurationConfigurer.afterConfigure(), so for me it looks @grgrzybek 's work is still valid.

igarashitm:
It's not the spring injection, XmlRoutesBuilderLoader.preParseRoute() parses beans and register them.

davsclaus:
Ah yes that is another possibility to move the "beans" parsing out of the regular phase, and make it part of preParse only with some special code in the route builder, ala we do for camel-k stuff.

Then the beans are eager and would be set during bootstrap

davsclaus:
Yeah I have a better PR in the works, so will close this - I just need to copy over some of the unit tests as we should have them for the new PR

davsclaus:
Replaced by https://github.com/apache/camel/pull/11476

PR #11475, opened 2023-09-19
(chores) use isBlank for testing blank Strings

orpiske:
- This avoids the overhead of creating a new String
- isBlank is usually a bit more efficient


davsclaus (approved):

oscerd (approved):

orpiske:
Failures seem unrelated. We should be good to merge.

PR #11498, opened 2023-09-20
CAMEL-19893: Migrate from jackson-module-jsonSchema to jackson-module-jsonSchema-jakarta

ppalaga:
https://issues.apache.org/jira/browse/CAMEL-19893


ppalaga:
@davsclaus is the impact on public API acceptable for a minor release?

oscerd (approved):

oscerd:
Well this would go to 4.1.0, so it's fine. We need to add a note to the migration guide

davsclaus (approved):

davsclaus:
There is also this PR
https://github.com/apache/camel/commit/3cae732be89d20e65d776ad72ef452f91654b1d2

I think this one fixes this by doing some pom.xml ant search/replace. But maybe that is no longer needed if this PR fixes this also ?

ppalaga:
> I think this one fixes this by doing some pom.xml ant search/replace. But maybe that is no longer needed if this PR fixes this also ?

Thanks, I was not aware of that one. I guess we need both.

PR #11528, opened 2023-09-22
CAMEL-19906 - Camel-Jbang: No need to have specific application prope…

oscerd:
…rties exports for Secret Refresh features

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
@@ -797,55 +797,6 @@ abstract class ExportBaseCommand extends CamelCommand {
         }
     }
 
-    protected void exportAwsSecretsRefreshProp(Properties properties) {
-        properties.setProperty("camel.vault.aws.accessKey", "<accessKey>");
-        properties.setProperty("camel.vault.aws.secretKey", "<secretKey>");
-        properties.setProperty("camel.vault.aws.region", "<region>");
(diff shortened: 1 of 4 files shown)


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
```diff
@@ -524,7 +524,6 @@ protected void copySettingsAndProfile(
             RuntimeUtil.loadProperties(prop2, profile);
         }
         prop2.putAll(prop);
-        prepareApplicationProperties(prop2);
```
davsclaus, line 527: This call still needs to be there

oscerd: It's used just here: https://github.com/apache/camel/pull/11528/files

For adding properties, which will be already exported by default from application.properties.. I introduced the method myself and I don't think it's needed. The prop2.putAll will do all the needed.

oscerd: I'll check again. I'm missing something I guess.

davsclaus: From the code then ExportQuarkus needs to have this method called as it does something special

oscerd: It should be fine now.

davsclaus (approved):

PR #11534, opened 2023-09-22
CAMEL-19864: fix default value for Univocity parsers line separator.

klease:
Use \n instead of \\n and adapt the utility class used to generate SpringBoot configuration file from the json files.

# Description

# Target

# Tracking


davsclaus:
Can you rebuild source and update PR as there are uncommitted files.

mvn clean install -Dquickly

oscerd (approved):
Lgtm. Just need a full build

klease:
I'm on it but due to a heavy schedule, not sure if I have time to commit during the day.

davsclaus:
No problem, one of us can likely do this and update the PR. Thanks for the status update.

klease:
I committed the generated catalog and camel-yaml-dsl files which look OK. Unfortunately the file catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd is not correct because it previously had the correct escape sequence (single backslash + n) and now it has an actual newline. So the generator needs to be reviewed. Also I didn't check the generation of the .adoc files which might be affected since currently they also have a single backslash.

davsclaus:
We could also remove that \n as default value, and just mention it in the docs / description.

klease:
> We could also remove that \n as default value, and just mention it in the docs / description.
True, but it seems OK now after the last commits.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
```diff
@@ -9382,7 +9382,7 @@ The line separator of the files The default value is to use the JVM platform lin
           <xs:annotation>
             <xs:documentation xml:lang="en">
 <![CDATA[
-The normalized line separator of the files The default value is a new line character. Default value: \n
+The normalized line separator of the files The default value is a new line character. Default value:
```
davsclaus, line 9385: Only here its not correct, but its something we can accept. Or you need to check the maven plugin that generates these comments: camel-eip-documentation-enricher-maven-plugin

klease: Yes, true; we could do this:  text += " Default value: " + defaultValueText.toString().replace("\n", "\\n");

klease:
I see the generated website files also have an issue where the newline appears literally. So I'm closing this pull request. Perhaps the easiest solution is what @davsclaus suggested which is to remove the default value. In fact these defaults are **not** present in the Camel Univocity component itself but in the underlying parser. Only in camel-spring-boot does the autoconfiguration set them on the Camel component itself.

PR #11599, opened 2023-09-28
Fix FileConsumerThreadsInProgressIssueTest on Windows

apupier:
On Windows, filenames cannot have `\r` and `\n`

# Description

# Target

# Tracking


core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerThreadsInProgressIssueTest.java
```diff
@@ -86,7 +86,7 @@ private void createManyFiles(int number) throws Exception {
         Path dir = testDirectory();
         for (int i = 0; i < number; i++) {
             String pad = String.format("%04d%n", i);
```
davsclaus, line 88: Its better to not generate file name using %n, so can you make a

    String name = String.format("%04d", i);

davsclaus (approved):

PR #11601, opened 2023-09-28
CAMEL-19644 - camel-jbang - Add command to generate SBOM report

oscerd:
# Description

# Target

# Tracking


davsclaus:
Okay this would work for an existing maven project if there is a pom.xml.

Can you look at how dependency copy does it, as it does a silent export in the background, and then after that executes that maven plugin.

This allows users to do

    camel sbom

On a jbang "project"

If you want to generate a sbom on an actual maven project, then you would do it the standard maven way with a <plugin> in the pom.xml etc.

oscerd:
> Okay this would work for an existing maven project if there is a pom.xml.
> 
> Can you look at how dependency copy does it, as it does a silent export in the background, and then after that executes that maven plugin.
> 
> This allows users to do
> 
> ```
> camel sbom
> ```
> 
> On a jbang "project"
> 
> If you want to generate a sbom on an actual maven project, then you would do it the standard maven way with a in the pom.xml etc.

Yeah, it makes sense, I removed that part. I didn't think about it. Thanks for pointing it out. I'll fix this accordingly.

davsclaus (approved):

PR #11638, opened 2023-10-04
CAMEL-8306: Add support for wildcards to match on prefix

TUCJVXCB:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
@@ -16,21 +16,16 @@
  */
 package org.apache.camel.support;
 
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.OptionalInt;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
  * A context path matcher when using rest-dsl that allows components to reuse the same matching logic.
  * <p/>
- * The component should use the {@link #matchBestPath(String, String, java.util.List)} with the request details and the
- * matcher returns the best matched, or <tt>null</tt> if none could be determined.
+ * The component should use the {@link #matchBestPath(String, String, List)} with the request details and the matcher
+ * returns the best matched, or <tt>null</tt> if none could be determined.
  * <p/>
  * The {@link ConsumerPath} is used for the components to provide the details to the matcher.
  */
@@ -156,9 +151,9 @@ public final class RestConsumerContextPathMatcher {
             }
         }
 
-        // if there are no wildcards, then select the matching with the longest path
-        boolean noWildcards = candidates.stream().allMatch(p -> countWildcards(p.getConsumerPath()) == 0);
-        if (noWildcards) {
+        // if there are no uri template, then select the matching with the longest path
+        boolean noCurlyBraces = candidates.stream().allMatch(p -> countCurlyBraces(p.getConsumerPath()) == 0);
+        if (noCurlyBraces) {
             // grab first which is the longest that matched the request path
             answer = candidates.stream()
                     .filter(c -> matchPath(requestPath, c.getConsumerPath(), c.isMatchOnUriPrefix()))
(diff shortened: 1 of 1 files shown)


core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
```diff
@@ -16,21 +16,16 @@
...
 
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.OptionalInt;
+import java.util.*;
```
oscerd, line 19: Please don't use * import.

oscerd (changes requested):
Minor finding.

core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
TUCJVXCB: > Please don't use * import.

got it

davsclaus:
It would be good to have an unit test for this as well - and also in one or more of the HTTP based components that support rest consumer

And then also find if possible where in the docs, to add some details about this

TUCJVXCB:
> It would be good to have an unit test for this as well - and also in one or more of the HTTP based components that support rest consumer
> 
> And then also find if possible where in the docs, to add some details about this

I look into other HTTP based components, they all use this class(RestConsumerContextPathMatcher).
image

later, I will add unit test and docs.

davsclaus:
Can you add tests where you have both wildcards and static paths, and some with curly placeholders

/camel/myapp/info
/camel/myapp/{id}
/camel/myapp/order/*

TUCJVXCB:
> Can you add tests where you have both wildcards and static paths, and some with curly placeholders
> 
> /camel/myapp/info /camel/myapp/{id} /camel/myapp/order/*

sure

core/camel-support/src/test/java/org/apache/camel/support/RestConsumerContextPathMatcherTest.java
```diff
@@ -83,19 +83,38 @@ public void testRestConsumerContextPathMatcherSuccess() {
...
+
+        RestConsumerContextPathMatcher.ConsumerPath<?> path1 = RestConsumerContextPathMatcher.matchBestPath("GET",
+                "/camel/foo", consumerPaths);
+
+        RestConsumerContextPathMatcher.ConsumerPath<?> path2 = RestConsumerContextPathMatcher.matchBestPath("GET",
                 "/camel/foo/bar", consumerPaths);
 
-        assertEquals(path1.getConsumerPath(), "/camel/*");
+        assertEquals(path1.getConsumerPath(), "/camel/foo");
         assertEquals(path2.getConsumerPath(), "/camel/*");
```
davsclaus, line 118: Should path2 not select /camel/foo/{id} because its the best candidate.

TUCJVXCB: The priority of wildcard is now higher than uri template . The correct order of priority is direct > URI template > wildcard, right?"

davsclaus: Correct. Wildcard should be last priority.

core/camel-support/src/main/java/org/apache/camel/support/RestConsumerContextPathMatcher.java
```diff
@@ -331,11 +345,31 @@ private static int countWildcards(String consumerPath) {
...
+        // Convert URI template to a regex pattern
+        String regex = consumerPath
+                .replace("/", "\\/")
+                .replace("{", "(?<")
+                .replace("}", ">[^\\/]+)");
+
+        // Add support for wildcard * as path suffix
+        regex = regex.replace("*", ".*");
+
+        // Match the provided path against the regex pattern
```
davsclaus, line 368: Is there a way to avoid regexp compile, as its slow and would impact high performance.
We need to do a simpler way of checking this.

TUCJVXCB: > Is there a way to avoid regexp compile, as its slow and would impact high performance. We need to do a simpler way of checking this.

Should be precompiled and saved, rather than compiling every request.

TUCJVXCB: > Is there a way to avoid regexp compile, as its slow and would impact high performance. We need to do a simpler way of checking this.

Trie tree is suitable for route matching, but it may require substantial code modifications.

davsclaus: Ah okay pre-compiled really helps, so its only calling the matcher per message

TUCJVXCB:
@davsclaus Could you please check if my changes look good

davsclaus:
Something is misisng in camel-platform-http I think it would also need this register/unregister.
And did you remove some tests by mistake ?

TUCJVXCB:
> Something is misisng in camel-platform-http I think it would also need this register/unregister. And did you remove some tests by mistake ?

The platform-http will use existing HTTP server to handle http request, so the routing match strategy involves using the existing HTTP server.

davsclaus:
org.apache.camel.component.undertow.ws.UndertowWsConsumerRouteTest.echo | 10.03s | FAILURE
-- | -- | --
org.apache.camel.http.common.CamelServletTest.testDuplicatedServletPath | 1.164s | ERROR

PR #11645, opened 2023-10-04
CAMEL-19949: camel-jbang - Allow to set a custom remote debugging port

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19949

## Motivation

The JVM debugging port is currently hard-coded to 4004, we would like to add the ability to set a custom port

## Modifications:

* Allow to set an integer as value of the argument ` --jvm-debug` to set a custom port


davsclaus (approved):

davsclaus:
Thanks this is much better

PR #11656, opened 2023-10-05
(chores) camel-core: avoid hitting the JVM slow path

orpiske:
In most cases, the type checks will be false, which causes the JVM to hit its slow path checking the type.

This already happens subsequently, so adjust it to check only for equality at first.

### DIFF
--- a/core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
+++ b/core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
@@ -107,7 +107,7 @@ public abstract class CoreTypeConverterRegistry extends ServiceSupport implement
             return null;
         }
 
-        if (type.isInstance(value)) {
+        if (type.equals(value.getClass())) {
             // same instance
             return (T) value;
         }


orpiske:
Waiting for my CI tests and perf results.

oscerd (approved):

core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
```diff
@@ -107,7 +107,7 @@ private <T> T fastConvertTo(Class<T> type, Exchange exchange, Object value) {
             return null;
         }
 
-        if (type.isInstance(value)) {
+        if (type.equals(value.getClass())) {
```
essobedo, line 110: This sounds really risky since `value` could be an instance of a subclass like in the next example:

Given:
```
    public interface Animal {
    }

    public static class Dog implements Animal {
    }
```
When:
```
        Dog pluto = new Dog();
        System.out.println("Is pluto an animal? Using isInstance " + (Animal.class.isInstance(pluto) ? "Yes" : "No"));
        System.out.println("Is pluto an animal? Using Equals " + (Animal.class.equals(pluto.getClass()) ? "Yes" : "No"));
```
Then
```
Is pluto an animal? Using isInstance Yes
Is pluto an animal? Using Equals No
```

orpiske: It's still gonna be checked for this later. This is just a duplicated one.

orpiske: More specifically: the code is doing this check twice. I'm just making the first check a bit lenient to avoid hitting a slow path in the JVM itself.

orpiske:
This one is good to go. No issues on my CI.

PR #11667, opened 2023-10-09
CAMEL-19945: camel-core - Add bean as property placeholder function

TUCJVXCB:
# Description

Add bean as property placeholder function

# Target

# Tracking

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/PropertiesComponent.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/PropertiesComponent.java
@@ -23,6 +23,7 @@ import java.util.Properties;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import org.apache.camel.NoSuchBeanException;
 import org.apache.camel.PropertiesLookupListener;
 import org.apache.camel.StaticService;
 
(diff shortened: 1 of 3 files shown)


davsclaus (changes requested):
This is not correct approach.

Take a look at how some of the other properties function has been implemented, and the code should be in camel-bean.

For example in camel-kuberntes
https://github.com/apache/camel/blob/main/components/camel-kubernetes/src/main/java/org/apache/camel/component/kubernetes/properties/SecretPropertiesFunction.java

TUCJVXCB:
> This is not correct approach.
> 
> Take a look at how some of the other properties function has been implemented, and the code should be in camel-bean.
> 
> For example in camel-kuberntes https://github.com/apache/camel/blob/main/components/camel-kubernetes/src/main/java/org/apache/camel/component/kubernetes/properties/SecretPropertiesFunction.java

thanks

TUCJVXCB:
> This is not correct approach.
> 
> Take a look at how some of the other properties function has been implemented, and the code should be in camel-bean.
> 
> For example in camel-kuberntes https://github.com/apache/camel/blob/main/components/camel-kubernetes/src/main/java/org/apache/camel/component/kubernetes/properties/SecretPropertiesFunction.java

Should we use a format like `from("kafka:{{bean:topicService.system}}").to("...")` ? ,it's similar to other placeholder

PR #11670, opened 2023-10-09
CAMEL-19536: replaced Thread.sleep() in camel-micrometer

LostArtist:
Replaced Thread.sleep() with Awaitility class. Don't know if it's the right way to do this, however it's the only working scenario I could find

### DIFF
--- a/components/camel-micrometer/pom.xml
+++ b/components/camel-micrometer/pom.xml
@@ -111,10 +111,9 @@
         <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
-            <version>4.2.0</version>
+            <version>${awaitility-version}</version>
             <scope>test</scope>
         </dependency>
-
 
     </dependencies>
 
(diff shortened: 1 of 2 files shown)


components/camel-micrometer/pom.xml
```diff
@@ -108,6 +108,13 @@
             <version>${assertj-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>4.2.0</version>
```
oscerd, line 114: This should come as property placeholder from Parent POM

oscerd (changes requested):
LGTM, just one minor finding.

components/camel-micrometer/pom.xml
LostArtist: thanks, fixed

oscerd (approved):

PR #11678, opened 2023-10-10
CAMEL-19970 - Avoid IllegalArgumentException for file without

apupier:
extension in same folder

when using JBang and that we let auto-detect routes, the files without extensions were added as potential routes and, as expected, no RouteLoader were found. An IllegalArgumentExeption was thrown at this time.

Note that the regression is a side effect. Previously it was part of Routes but there wasn't `pre-parsing` and so there was no error (at least when doing an export)

# Description

# Target

# Tracking


apupier:
not found a way to write a test easily.

apupier:
Given that it is a regression in Camel 3.21, I think we should backport it to 4.0.x and 3.21.x and 3.x if and when the PR is accepted

davsclaus (changes requested):
This should be done in the skipFile method instead, and I think its easier to just check if there is no file extension then return true to skip the file.

davsclaus (approved):

davsclaus:
Ah can you move that logic after the // skip dirs check as essentially almost all dirs are named without a dot / extension.

PR #11714, opened 2023-10-12
CAMEL-19982: camel-jbang - Allow --jvm-debug as last parameter

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19982 for 4.1

## Motivation

When `--jvm-debug` is used as last parameter in a run command like `camel run MyKafka.java --jvm-debug`, it fails with the next error `Error: missing required parameter`

## Modifications:

* Allow to prevent throwing exception when there are no arguments left in `ParameterConsumer`
* Add support of empty arguments in `DebugConsumer`


oscerd:
4.1.x is non LTS, so it won't be released anymore as "minor"

PR #11752, opened 2023-10-18
Remove deprecated usage of Exchange.getOut in StaxProcessor

apupier:
# Description

# Target

# Tracking


davsclaus (changes requested):
this is not correct, the old code is correct

Otherwise all the code should just be

exchange.getMessage().setBody

to only change the body, and not copy over headers etc.

apupier:
By what should it be replaced then? How to get rid of the deprecated method?

davsclaus:
> By what should it be replaced then? How to get rid of the deprecated method?

Leave it as-is. The deprecation is for end users to not use it.
We will later in Camel 5 remove and change internal inside camel

PR #11754, opened 2023-10-18
CAMEL-20008 - Improve assertion to compare xml and not string in XMLTokenExpressionIteratorTest

apupier:
* With Java 21, the order of attributes are different but it does not change the semantic
* factorize the xml-unit version in parent pom

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-stax/pom.xml
+++ b/components/camel-stax/pom.xml
@@ -74,6 +74,11 @@
             <artifactId>camel-jaxb</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-assertj3</artifactId>
+            <scope>test</scope>
+       </dependency>
 
     </dependencies>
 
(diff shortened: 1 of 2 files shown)


parent/pom.xml
```diff
@@ -2697,6 +2697,18 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
```
oscerd, line 2700: Unless there is a good reason I would avoid adding other dependencies in parent/pom

oscerd (changes requested):
just one minor thing.

parent/pom.xml
apupier: for `xmlunit-core`, it was already used in several projects.

for `xmlunit-assertj3`, it allows to write simpler code and have a better error message

apupier: I can update to use the same xmlunit-core in this PR, and provide a different PR to migrate from xmlunit-core to xmlunit-assertj3 to separate concerns and discuss if it is worthy or not to use the introduced one.

apupier: Second PR provided to replace unitcore with assertj https://github.com/apache/camel/pull/11774

apupier: the 2 other PRs have been merged.
Less code change in this PR

davsclaus (approved):

PR #11760, opened 2023-10-18
(chores) camel-core: reduce method size to force inlining

orpiske:
Reduce the size of a few hot methods, so that their byte code size is smaller than 325 bytes, to increase the chance that the JVM will inline them when possible.


oscerd (approved):

orpiske:
Looking good on CI. Ready to merge.

PR #11762, opened 2023-10-18
cleanups: replaced File#delete with Files#delete

LostArtist:


essobedo:
Are there any reasons why we should do that?
According to the Javadoc, the main interest in using `Files#delete` is for error reporting, to diagnose why a file cannot be deleted but in this PR you mostly don't care about the exception thrown so it is not really useful. Moreover, it causes [some known behavior changes](https://mail.openjdk.org/pipermail/core-libs-dev/2012-August/011314.html) and finally dealing with exceptions affects the performances so I tend to vote -1 for this change unless you have good reasons

orpiske (approved):

PR #11781, opened 2023-10-20
Added route naming example for YAML

ssalewski:
# Description

Added the same example used for naming a route as YAML DSL

# Target

# Tracking


davsclaus (approved):

essobedo (approved):
Thx for the contribution. Just FYI, you can inline `uri` with `to` in case you don't have other key/value pairs to provide to configure your producer, so it could be simplified with:
```
...
        - to: mock:bar
```

oscerd (approved):

PR #11812, opened 2023-10-23
Camel-15211: add props to configure SSL context parameters

jono-morris:
Added  `camel.main.ssl.` parameters allowing Camel's SSL security parameters to be configured.

Tests added to camel-jetty, camel-netty, and camel-vertx.


oscerd (approved):

core/camel-main/src/main/java/org/apache/camel/main/DefaultConfigurationProperties.java
```diff
@@ -159,6 +159,15 @@ public abstract class DefaultConfigurationProperties<T> {
     private String startupRecorderProfile = "default";
     private long startupRecorderDuration;
     private String startupRecorderDir;
+    private boolean sslEnabled;
+    private String sslKeyStore;
+    @Metadata(defaultValue = "changeit")
+    private String sslKeystorePassword = "changeit";
```
davsclaus, line 165: We should not have default passwords

davsclaus (changes requested):
Add a new camel.main.xxx - should also update in PrepareCamelMainMojo and there is also code changes needed in BaseMainSupport

jono-morris:
Thanks for the review.  Give me a day to make the changes.

jono-morris:
Added camel.main.ssl.xxx, and updated PrepareCamelMainMojo and BaseMainSupport. 
Thanks.

davsclaus:
Can we change `camel.main.ssl` to `camel.ssl` as this is not tied to main only.

jono-morris:
Yes of course. Let me do that.

jono-morris:
Changes committed changing `camel.main.ssl` to `camel.ssl`.  Thanks

davsclaus (approved):

davsclaus:
We also need some work for camel-spring-boot, to make camel.ssl configuration possible via its auto configuration.

Its something like threadpool but not exactly the same, but you can take a look
https://github.com/apache/camel-spring-boot/tree/main/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/threadpool

jono-morris:
Yes certainly, let me take a look.

PR #11837, opened 2023-10-25
CAMEL-19986: created throttling executor in camel-test

LostArtist:
example of using:

```
executeSlowly(size, 3, TimeUnit.MILLISECONDS, (i) -> template.sendBody(url, "Message " + i));
```

### DIFF
--- a/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/TestSupport.java
+++ b/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/TestSupport.java
@@ -23,6 +23,8 @@ import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+import java.util.function.IntConsumer;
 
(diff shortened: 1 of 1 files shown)


orpiske (changes requested):
Maybe use an [IntConsumer](https://docs.oracle.com/javase/8/docs/api/java/util/function/IntConsumer.html) instead of a custom interface.

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/TestSupport.java
```diff
@@ -581,4 +582,16 @@ public static String fileUri(Path testDirectory) {
     public static String fileUri(Path testDirectory, String query) {
         return "file:" + testDirectory + (query.startsWith("?") ? "" : "/") + query;
     }
+
+    @FunctionalInterface
+    public interface IntConsumer {
+        void execute(int i);
+    }
```
orpiske, line 589: This can be removed.

orpiske (changes requested):

orpiske (approved):

PR #11865, opened 2023-10-30
[3.21.x] CAMEL-20053: Fix watchUpdates consumer so that it sees issues created after route startup

jamesnetherton:


oscerd:
Component test comment doesn't work in 3.x

oscerd (approved):

PR #11868, opened 2023-10-30
(chores) camel-platform-http-vertx: break large methods

orpiske:
This should help obtain better information when profiling the code

### DIFF
--- a/components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/VertxPlatformHttpConsumer.java
+++ b/components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/VertxPlatformHttpConsumer.java
@@ -26,6 +26,7 @@ import java.util.regex.Pattern;
 
 import jakarta.activation.DataHandler;
 
+import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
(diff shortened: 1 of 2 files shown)


components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/VertxPlatformHttpConsumer.java
```diff
@@ -183,35 +184,44 @@ protected void handleRequest(RoutingContext ctx) {
...
+    }
+
+    private void handleFailure(RoutingContext ctx, Throwable failure) {
+        getExceptionHandler().handleException(
+                "Failed handling platform-http endpoint " + getEndpoint().getPath(),
+                failure);
+        ctx.fail(failure);
+    }
+
+    private Object processRequest(Exchange exchange) throws Exception {
```
jamesnetherton, line 221: Since `executeBlocking` expects `Callable<Void>` in this case, maybe return that instead of a `null` `Object`?

orpiske: We can evaluate that in a subsequent improvement, but, for now, I'd rather keep the same behavior as before to ensure the refactoring doesn't introduce side effects other than those potentially caused by moving things around.

PR #11870, opened 2023-10-30
CAMEL-19997: new camel-smb component

orpiske:
I'm working on a new component as part of CAMEL-19997. This is the very first iteration of it (only as consumer for now)  ... Kindly appreciate reviews and comments.


orpiske:
Note: also comes with a test-infra module for that.

orpiske:
Marking as draft again. There's a valid test error I need to fix when I get back.

orpiske:
Looking OK now.

parent/pom.xml
```diff
@@ -952,6 +953,11 @@
                 <artifactId>camel-chunk</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.camel</groupId>
+                <artifactId>camel-cifs</artifactId>
```
oscerd, line 958: Is thius really needed?

oscerd (changes requested):
I don't think camel-cifs is supposed to be listed

parent/pom.xml
orpiske: Man, that's not needed. That's very strange. That was the old name I started this component with. I'll erase my local cache and rebuild. Thanks for pointing it out!

components/camel-smb/src/main/resources/META-INF/services/org/apache/camel/component/smb
```diff
@@ -0,0 +1 @@
+class=org.apache.camel.component.cifs.SmbComponent
```
orpiske, line 1: This is also wrong (just noticed based on @oscerd's comment above)

orpiske:
Hey @oscerd, any other comments here? Otherwise I'll merge ... it would be good to have this one for 4.2.

oscerd:
For me it's +1

oscerd (approved):

orpiske:
Thanks! Let's get this one in.

PR #11879, opened 2023-10-31
(chore) ci: convert component-test parameters to env vars

essobedo:
## Modifications

* Use environment variables to avoid injections

## Result

The injection is no longer possible, the Github action simply fails


essobedo:
/component-test azure" true build.log && ls / && echo "

PR #11882, opened 2023-11-01
Converter lookup once again matches subclass converters like in Camel 3

fil512:
This is a fix to what looks like an accidental regression in Camel 3 -> Camel 4.

Fix for https://issues.apache.org/jira/browse/CAMEL-19828

Closes #CAMEL-19828

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
@@ -1,0 +1,76 @@
+package org.apache.camel.impl.converter;
+
+import org.apache.camel.TypeConverter;
+import org.apache.camel.converter.TypeConvertible;
+import org.apache.camel.support.SimpleTypeConverter;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class TypeResolverHelperTest {
+    private static final Map<TypeConvertible<?, ?>, TypeConverter> registeredConverters = new HashMap<>();
+
+    @BeforeAll
+    static void setUp() {
+        registeredConverters.put(new TypeConvertible<>(Source.class, Child.class), new SourceChildConverter());
+    }
+
+    @Test
+    public void testTryAssignableToChild() {
+        TypeConvertible<?, ?> requestedConverter = new TypeConvertible<>(Source.class, Child.class);
+        TypeConverter foundConverter = TypeResolverHelper.tryAssignableFrom(requestedConverter, registeredConverters);
+        assertNotNull(foundConverter);
+        Child result = foundConverter.tryConvertTo(Child.class, new Source("source"));
+        assertEquals("sourceP", result.parentField);
+        assertEquals("sourceC", result.childField);
+    }
+
+    // This general behaviour works in Camel 3 but stopped working in Camel 4 due to what looks like an accidental regression.
+    // See https://issues.apache.org/jira/browse/CAMEL-19828
+    @Test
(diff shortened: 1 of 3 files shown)


core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.camel.impl.converter;
```
oscerd, line 1: Missing license header

core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.camel.TypeConverter;
+import org.apache.camel.converter.TypeConvertible;
+import org.apache.camel.support.SimpleTypeConverter;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
```
oscerd, line 12: Don't use * imports

oscerd (changes requested):
Minor findings

core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
fil512: fixed

core/camel-core/src/test/java/org/apache/camel/impl/converter/TypeResolverHelperTest.java
fil512: fixed

orpiske (changes requested):
Just temporarily requesting changes to ensure it can be fully tested before merging.

essobedo:
Github action `component-test` doesn't work properly in this revision, a rebase on main is required to get the latest changes

orpiske:
@fil512 as it is, this patch breaks `camel-jackson`. It can be fixed by adding a strict converter on [JacksonTypeConverter](https://github.com/apache/camel/blob/main/components/camel-jackson/src/main/java/org/apache/camel/component/jackson/converter/JacksonTypeConverters.java#L172) that can match a BooleanNode type.

```
    @Converter
    public Boolean toBoolean(BooleanNode node, Exchange exchange) throws Exception {
        return node.asBoolean();
    }
```

fil512:
Fixed.

orpiske:
Something is still off with this one as I see a couple of tests failing: 

| Failed Test | Duration | Failure Type |
| --- | --- | --- |
| org.apache.camel.builder.xml.XPathTest.testXPathNotUsingExchangeEvaluate | 0.076s | ERROR |
| org.apache.camel.builder.NotifyBuilderTest.testFromFilterBuilderWhenExchangeDone | 0.135s | FAILURE |
| org.apache.camel.component.jackson.converter.JacksonJSonNodeConverterTest.convertToBoolean | 0.021s | ERROR |

fil512:
Those tests depended on the original behaviour.  I restored that behaviour.  Can you run the test suite again?

orpiske:
> Those tests depended on the original behaviour. I restored that behaviour. Can you run the test suite again?

I triggered a test on my CI and I'll check the results on Monday. I am not sure if that's the right approach, though ... but let's see how it works.

orpiske:
> > Those tests depended on the original behaviour. I restored that behaviour. Can you run the test suite again?
> 
> I triggered a test on my CI and I'll check the results on Monday. I am not sure if that's the right approach, though ... but let's see how it works.

There was an outage on the lab, so this job did not finish. I am running it again and will report once I have the results.

orpiske:
Worked OK on my own CI. Let's get this one in.

PR #11897, opened 2023-11-02
Remove hard coded username

cunningt:
PGEventEndpoint.java is failing SAST tests because of a hard coded username.     From my reading of the code, setting the user to postgres (9 years ago) predates the defaultValue of the URIParam for user, and I think it can be removed because of the default value being set in the URIParam annotation for user.

### DIFF
--- a/components/camel-pgevent/src/main/java/org/apache/camel/component/pgevent/PgEventEndpoint.java
+++ b/components/camel-pgevent/src/main/java/org/apache/camel/component/pgevent/PgEventEndpoint.java
@@ -63,7 +63,7 @@ public class PgEventEndpoint extends DefaultEndpoint {
     @Metadata(required = true)
     private String channel;
     @UriParam(defaultValue = "postgres", label = "security", secret = true)
-    private String user = "postgres";
+    private String user;
     @UriParam(label = "security", secret = true)
     private String pass;
     @UriParam


oscerd (approved):

davsclaus (changes requested):
Is the default value somewhere in postgres automatic `postgres` or what?

The uri param has default value set which otherwise would be wrong

davsclaus:
The UriParam does NOT magically set any kind of default value. You need to set. that in the code

oscerd (changes requested):
postgres is the default value just if you run this stuff from a CLI, in Camel you need to set that.

davsclaus (approved):

oscerd:
I don't think this PR is valid. This way we won't have a valid default defined.

orpiske:
> I don't think this PR is valid. This way we won't have a valid default defined.

I agree with @oscerd. I think SAST tests are wrong here: it's not Camel's fault (or job) to policy the default usernames used by other applications. IMHO, the fact that some tool uses an specific hard-coded user name is a problem of that application.

davsclaus:
Ah okay so is there some CLI tool that uses posgres as default value. But real world databases would surely not use posgtres as username and no password? So I wonder if maybe we can remove the default value all together in Camel. And if username/password is the only way to authenticate, we can mark these options as required = true

davsclaus:
Also I is the pasword option really named `pass` in posgress, or should it not be renamed to `password`

cunningt:
I'll try to find an alternative way of setting the username initially / solving the SAST test.    Loading credentials from a properties file and then setting it seems to solve it, but that seems a little bit clunky in this case.

PR #11920, opened 2023-11-06
CAMEL-20044 add extra logging for BreakOnFirstError

CodeSmell:
# Description

This PR has some extra logging related to CAMEL-20044 
This PR doesn't match the log files in the issue but is similar

This PR is meant as way to more readily see what is happening when Camel processes a message that fails when `breakOnFirstError` is set. That can be used as a discussion point to work on issues in `breakOnFirstError` processing

# Target

Camel-3.21.x

# Tracking
CAMEL-20044

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
@@ -136,8 +136,9 @@ public class KafkaRecordProcessor {
         // processing failed due to an unhandled exception, what should we do
         if (configuration.isBreakOnFirstError()) {
             
-            if (topicPartition.partition() != record.partition()) {
-                LOG.error("about to process an exception with UNEXPECTED partition. Got topic partition {}. " + 
+            if (lastResult.getPartition() != -1 &&
+                lastResult.getPartition() != record.partition()) {
+                LOG.error("about to process an exception with UNEXPECTED partition & offset. Got topic partition {}. " + 
                         " The last result was on partition {} with offset {} but was expecting partition {} with offset {}",
                         topicPartition.partition(), lastResult.getPartition(), lastResult.getPartitionLastOffset(), 
                         record.partition(), record.offset());
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
@@ -116,7 +116,7 @@ public class KafkaRecordProcessor {
         }
         if (exchange.getException() != null) {
             
-            LOG.debug("an exception was thrown for record at partition {} and offset {}",
+            LOG.debug("An exception was thrown for record at partition {} and offset {}",
                     record.partition(), record.offset());
             
             boolean breakOnErrorExit = processException(exchange, topicPartition, record, lastResult,
@@ -138,7 +138,7 @@ public class KafkaRecordProcessor {
             
             if (lastResult.getPartition() != -1 &&
                 lastResult.getPartition() != record.partition()) {
-                LOG.error("about to process an exception with UNEXPECTED partition & offset. Got topic partition {}. " + 
+                LOG.error("About to process an exception with UNEXPECTED partition & offset. Got topic partition {}. " + 
                         " The last result was on partition {} with offset {} but was expecting partition {} with offset {}",
                         topicPartition.partition(), lastResult.getPartition(), lastResult.getPartitionLastOffset(), 
                         record.partition(), record.offset());
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
@@ -62,6 +62,9 @@ public class KafkaRecordProcessor {
         if (record.key() != null) {
             message.setHeader(KafkaConstants.KEY, record.key());
         }
+        
+        LOG.debug("setting up the exchange for message from partition {} and offset {}",
+                record.partition(), record.offset());
 
         message.setBody(record.value());
     }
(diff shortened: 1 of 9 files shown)


CodeSmell:
Wasn't sure if should checkin these files under src/generated/resources

- kafka.json
- spring-main.json
- component.properties
- other.properties

davsclaus (approved):

oscerd (approved):
This looks good at first sight, but @orpiske could have a look too.

orpiske:
Looking at it ...

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -112,30 +115,48 @@ public ProcessingResult processExchange(
...
+                        " The last result was on partition {} with offset {} but was expecting partition {} with offset {}",
+                        topicPartition.partition(), lastResult.getPartition(), lastResult.getPartitionLastOffset(), 
+                        record.partition(), record.offset());
+            }
+            
             // we are failing and we should break out
             if (LOG.isWarnEnabled()) {
-                LOG.warn("Error during processing {} from topic: {}", exchange, partition.topic(), exchange.getException());
-                LOG.warn("Will seek consumer to offset {} and start polling again.", partitionLastOffset);
+                LOG.warn("Error during processing {} from topic: {}", exchange, topicPartition.topic(), exchange.getException());
```
orpiske, line 149: Maybe also log the exception message for more details? 

```
LOG.warn("Error during processing exchange {} from topic {}: {}", exchange, topicPartition.topic(), exchange.getException() ? null : exchange.getException().getMessage(), exchange.getException());
```

orpiske: Obs.: the exception message may also be present elsewhere in the log, but if the intention is to provide more data, I think it would be OK.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/ProcessingResult.java
```diff
@@ -21,14 +21,16 @@
 
 public final class ProcessingResult {
     private static final ProcessingResult UNPROCESSED_RESULT
-            = new ProcessingResult(false, AbstractCommitManager.START_OFFSET, false);
+            = new ProcessingResult(false, -1, AbstractCommitManager.START_OFFSET, false);
```
orpiske, line 24: Please, use a constant here and document it. That way it may be easier for us to understand the code later on.

orpiske (changes requested):
Thanks for the contribution! 

Overall it looks good to me and I am pointing just a few minor things.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -112,30 +115,48 @@ public ProcessingResult processExchange(
...
 
-            // force commit, so we resume on next poll where we failed except when the failure happened
-            // at the first message in a poll
-            if (partitionLastOffset != AbstractCommitManager.START_OFFSET) {
-                commitManager.forceCommit(partition, partitionLastOffset);
+            // force commit, so we resume on next poll where we failed 
+            // except when the failure happened at the first message in a poll
+            if (lastResult.getPartitionLastOffset() != AbstractCommitManager.START_OFFSET) {
+                // should we use record.offset ?
+                //commitManager.forceCommit(topicPartition, record.offset() - 1);
```
CodeSmell, line 158: this change that is commented out is what mostly addresses the issues found in CAMEL-20044 and CAMEL-19894

I can submit this as a separate PR 
I commented it out here b/c I had not added unit tests and b/c I wasn't sure if that would impact other ways that the Camel riders would want breakOnFirstError to work

I did see CAMEL-20089 but hope this fix can move forward as we it does fix 2 other issues

orpiske: In this case, it's better to remove the code. We can always refer to the git log for details about the history of the code.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
orpiske: Just one more note: the code as it is, is still not printing the exception message.

orpiske (changes requested):
Almost there. Just a few minor items.

And, as always, thanks for the contribution.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
CodeSmell: I am working on a separate PR for the fix
Is this worth pursuing?

orpiske: Yes it is. If you can fix it, we'll be happy to look at your proposed changes. 

Obs.: I am talking about being better to remove the commented code.

CodeSmell: Cool. Removed

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
orpiske: This is the only item remaining. As it is now, it is logging the message. Instead, I would like it to be exactly like this:

```
LOG.warn("Error during processing exchange {} from topic {}: {}", exchange, topicPartition.topic(), exchange.getException() ? null : exchange.getException().getMessage(), exchange.getException());
```

Just copy/paste the code above and it should be fine ⬆️

orpiske (changes requested):
Almost there. There is a small error on the log message for the exception which is printing the message in the exchange, instead of the exception message. Once that is fixed, we are ready to merge.

orpiske (approved):

orpiske:
This is looking OK.

CodeSmell:
> This is looking OK.

@orpiske 
I updated the logging here so that it each word is capitalized per other comment

orpiske:
Thank you @CodeSmell! I am now running manual tests on this code and will merge / report back soon.

orpiske:
This needs to be backported to 3.x

PR #11930, opened 2023-11-07
CAMEL-20061: SMPP interface version can now be configured

brunoriemenschneider:
Interface Version functionality as supported by jSMPP is adopted with @UriParam configuration parameter. The default value (interface version 3.4) remains unchanged.

The configuration value is parsed as String and converted to hex byte (which is used at protocol level) at a later stage when passed to jSMPP, as Camel does not support hexadecimal byte parsing from config values at this point. (I think the issue is in the Camel support component: the PropertyConfigurerSupport.class does not implement this case)

# Description

JIRA Ticket CAMEL-20061
The interface version of the SMPP protocol binding can now be specified via configuration parameter. The configuration paramter is simply passed to the underlying library jSMPP as input to the wrapper class BindParameter, which is used for the bind message to establish a connection to an SMSC.

The prior default value of SMPP v3.4 is retained as default value.

The SMPP protocol defines the bind parameter interface_version as hexadecimal byte. This implementation reads the 0xXX byte value as String, and converts it to byte when passed to jSMPP. This has been done because AFAIK the current implementation in the Camel support component the ProprtyConfigurerSupport.class seems not to support hexadecimal byte configurators.

# Target

# Tracking


components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConfiguration.java
```diff
@@ -733,6 +735,26 @@ public void setSingleDLR(boolean singleDLR) {
...
     }
 
+    public String getInterfaceVersion() {
+        return interfaceVersion;
+    }
+
+    /**
+     * Defines the interface version to be used in the binding request with the SMSC. The following values are allowed,
+     * as defined in the SMPP protocol:
+     * <ul>
```
oscerd, line 745: Since this will be used to generate documentation I would avoid ul and li, because for the Javadoc they will be good, but for generating the website they could be a mess.

oscerd (changes requested):
Minor findings, the logic is fine. We could have some troubles with docs generation

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConfiguration.java
brunoriemenschneider: I have used the syntax that a lot of the other configuration fields have used (look at this file and scroll up for a few cases). I can see though that they are not really useful in the docs (https://camel.apache.org/components/4.0.x/smpp-component.html). Will remove.

oscerd: Looks good then. I never noticed that, maybe because I didn't work on this specific component. All good.

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConfiguration.java
```diff
@@ -120,6 +120,8 @@ public class SmppConfiguration implements Cloneable {
     private Integer pduProcessorQueueCapacity = 100;
     @UriParam(label = "advanced", defaultValue = "false")
     private boolean singleDLR;
+    @UriParam(label = "advanced", enums = "0x00,0x33,0x34,0x50", defaultValue = "0x34")
+    private String interfaceVersion = "0x34";
```
davsclaus, line 124: Maybe just use a plain string to set the value in better human understandable as just a string with 3.3, 3.4, 5.0 etc. And then internally convert that to what jsnmp uses.

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConsumer.java
```diff
@@ -109,7 +110,8 @@ private SMPPSession createSession() throws IOException {
                         BindType.BIND_RX, this.configuration.getSystemId(),
                         this.configuration.getPassword(), this.configuration.getSystemType(),
                         TypeOfNumber.UNKNOWN, NumberingPlanIndicator.UNKNOWN,
-                        configuration.getAddressRange()));
+                        configuration.getAddressRange(),
+                        InterfaceVersion.valueOf(Byte.decode(configuration.getInterfaceVersion()))));
```
davsclaus, line 114: See prev comments

davsclaus (changes requested):

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConfiguration.java
brunoriemenschneider: The SMPP protocol uses these codes (e.g. se https://smpp.org/SMPP_v5.pdf - chapter 4.7.13) in its specification. Are you sure I should remove them? (Code is simpler/easier to understand like this)

pull_request_CAMEL-20061.md
```diff
@@ -0,0 +1,40 @@
+# Description
```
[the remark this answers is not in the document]
brunoriemenschneider: deleted the file

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConfiguration.java
brunoriemenschneider: I have introduced a new scheme using "legacy", "3.3", "3.4", "5.0"

components/camel-smpp/src/main/java/org/apache/camel/component/smpp/SmppConsumer.java
brunoriemenschneider: See above

oscerd (approved):

davsclaus (approved):

PR #11933, opened 2023-11-07
CAMEL-19828: avoid iterating twice over the converters' map

orpiske:


oscerd (approved):

orpiske:
Looking OK on own CI. Let's merge.

PR #11934, opened 2023-11-07
CAMEL-19835: camel-core: replace temporary dir logic in test code

anton-goncharov:
# Description

Making use of junit5 `@TempDir` in TestSupport to control lifecycle of a temporary directory when it's used in camel-core tests. @TempDir uses a system-specific path dedicated for temporary data.

- Removed `deleteDirectory()` methods and their usage 
- Changed `testDirectory()` and `fileUri()` and their overloads to use the temporary directory
- Removed `deleteDirectory()` from SpringTestSupport

Adapted tests where needed in `camel-core` and `camel-spring-xml`, which contains tests depending on `camel-core`

In some corner cases Camel context is initialized on class level (e.g. Spring tests with `@ContextConfiguration`), they're addressed in a way of supplying custom initializer that passes temp directory path through a system variable.

# Target

# Tracking


oscerd (approved):

anton-goncharov:
@oscerd I guess since it's my first contribution, it needs workflow approval each time. Can you please approve again?

oscerd:
It's something reserved to PMC and committer, so each time you push something we need to approve and run the pipeline again.

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
```diff
@@ -57,7 +57,7 @@ public void testLondon() throws Exception {
         mock.expectedBodiesReceived("London");
 
         template.sendBodyAndHeader(fileUri(), "London", Exchange.FILE_NAME, "london.txt");
-        mock.assertIsSatisfied(1000);
+        mock.assertIsSatisfied(3000);
```
orpiske, line 60: Why the timeout increase?

core/camel-core/src/test/java/org/apache/camel/TestSupport.java
```diff
@@ -445,43 +422,6 @@ public static Channel unwrapChannel(Processor processor) {
...
-    /**
-     * create the directory
-     *
-     * @param file the directory to be created
-     */
-    public static void createDirectory(String file) {
-        File dir = new File(file);
-        dir.mkdirs();
-    }
-
```
orpiske, line 484: We probably need to document the removal of these methods on the upgrade documentation. They may be used elsewhere / by users.

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
```diff
@@ -72,7 +72,7 @@ public void testDublin() throws Exception {
         mock.expectedBodiesReceived("Dublin");
 
         template.sendBodyAndHeader(fileUri(), "Dublin", Exchange.FILE_NAME, "dublin.txt");
-        mock.assertIsSatisfied(1000);
+        mock.assertIsSatisfied(3000);
```
orpiske, line 75: Same note about the timeout increase.

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
```diff
@@ -87,7 +87,7 @@ public void testMadrid() throws Exception {
         mock.expectedBodiesReceived("Madrid");
 
         template.sendBodyAndHeader(fileUri(), "Madrid", Exchange.FILE_NAME, "madrid.txt");
-        mock.assertIsSatisfied(1000);
+        mock.assertIsSatisfied(3000);
```
orpiske, line 90: Same note about the timeout increase.

orpiske (changes requested):
Thanks for your contribution! 

This is looking pretty cool. I did left a few comments about minor things I found.

core/camel-core/src/test/java/org/apache/camel/TestSupport.java
orpiske: Alternatively: wrapping them around the new `tempDir` stuff and marking them as deprecated would be acceptable as well (and, also, adding a note about their deprecation on the upgrade doc).

anton-goncharov: That's a good point about not removing them and deprecating instead. I'll take this approach, and will also update the upgrade doc. Thanks!

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
anton-goncharov: I couldn't find another explanation why this test has failed on a previous PR check run, as locally it passes when I run all camel-core tests. The same I see now but with other tests. After this change of timeout, this `FileConsumerFailureHandledTest` passes but two others fail: 
- `org.apache.camel.processor.MulticastParallelTimeoutStreamCachingTest.testCreateOutputStreamCacheBeforeTimeoutButWriteToOutputStreamCacheAfterTimeout`
- `org.apache.camel.processor.MulticastParallelStreamingTest.testMulticastParallel`

Both don't use temp directory anyhow and shouldn't be affected by my changes, so I'm wondering how stable such tests are. I'm very new to it so I'm trying to understand if it's something that can happen ocassionally due to slower test execution, or if it's related to my PR changes. I apologize for my confusion as someone new, but really appreciate your guidance and open to suggestions 🙂

orpiske: So, the challenge here is that we have to balance two things: 

1. Quick test execution 
2. Stability 


In some of the systems where these tests run, such as GitHub, the systems running them could be overloaded at the time of execution. Some tests use fixed sleeps and/or timeouts to wait for "something" to happen, but this process is not stable at all (hence, why we have been converting most of the sleeps with Awaitility). 

I suspect that's why you see these new failures: maybe if you run them again, you'll see a different set of failures. 

Normally, what we do is that we use the [ASF CI](https://ci-builds.apache.org/job/Camel/job/Camel%20JDK17/) as the baseline: as long as tests run fine there, we are OK with it (and we are a bit lenient with [most of] the core check failures on GitHub).

So, in this case, I'd leave it as is.

orpiske:
I think we won't have time to get this on in 4.2, but it's a good change. We might be able to add it to 4.3.

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
anton-goncharov: Thanks for the explanation, I will revert the timeouts then and set them back to 1000

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
anton-goncharov: Reverted back to 1000

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
anton-goncharov: Reverted back to 1000

core/camel-core/src/test/java/org/apache/camel/component/file/FileConsumerFailureHandledTest.java
anton-goncharov: Reverted back to 1000

core/camel-core/src/test/java/org/apache/camel/TestSupport.java
anton-goncharov: I added the deleted methods back and annotated as `@Deprecated`. About the doc, please let me know if I'm incorrect but this class `org.apache.camel.TestSupport` is from the `test` directory and shouldn't be visible to the end users of `camel-core`. Should I update 4.2→4.3 upgrade doc anyway?

orpiske: Ahh, good point. In this case we don't have to update the docs.

orpiske (approved):
Tested this one manually. It looks OK to me. The failures on GH seem unrelated.

anton-goncharov:
thanks @orpiske , you're a mind reader, I was about to write to ask if we can proceed with this one.
I'll be checking on ASF CI pipeline results

orpiske:
Thanks for your contribution @anton-goncharov !

PR #11935, opened 2023-11-07
CAMEL-20044 BreakOnFirstError can commit the wrong partition:offset

CodeSmell:
# Description

This PR offers a fix for breakOnFirstError where the incorrect partition:offset when a forceCommit is done 

# Target

This PR is off the camel-3.21.x branch

# Tracking
CAMEL-20044 
CAMEL-19894


CodeSmell:
Note this fix was done off my local branch that has the logging statements.
Those logging statements are also in this approved PR but it wasn't merged yet 
- https://github.com/apache/camel/pull/11920

CodeSmell:
I was adding integration tests for this fix
The two in the initial commit cover the basic behavior.
Still working on adding IT for the 2 issues 
(Note: need to create multiple partitions in the `BaseEmbeddedKafkaTestSupport` but wasn't sure how to do that)

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -112,30 +115,56 @@ public ProcessingResult processExchange(
             exchange.setException(e);
         }
         if (exchange.getException() != null) {
-            boolean breakOnErrorExit = processException(exchange, partition, lastResult.getPartitionLastOffset(),
+            
+            LOG.debug("an exception was thrown for record at partition {} and offset {}",
```
orpiske, line 119: Minor thing for consistency: capitalize the first letter of the log message. This is pattern we use throughout the code base.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -112,30 +115,56 @@ public ProcessingResult processExchange(
...
+            Exchange exchange, TopicPartition topicPartition, 
+            ConsumerRecord<Object, Object> record, ProcessingResult lastResult,
             ExceptionHandler exceptionHandler) {
 
         // processing failed due to an unhandled exception, what should we do
         if (configuration.isBreakOnFirstError()) {
+            
+            if (lastResult.getPartition() != -1 &&
+                lastResult.getPartition() != record.partition()) {
+                LOG.error("about to process an exception with UNEXPECTED partition & offset. Got topic partition {}. " +
```
orpiske, line 141: Same note.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessorFacade.java
```diff
@@ -77,6 +77,9 @@ public ProcessingResult processPolledRecords(
 
                 lastResult = processRecord(partition, partitionIterator.hasNext(), recordIterator.hasNext(), lastResult,
                         kafkaRecordProcessor, record);
+                
+                LOG.debug("processed record on partition {} and offset {} and got result for partition {} and offset {}",
```
orpiske, line 81: Same note.

orpiske (changes requested):
Just a few minor items

orpiske:
Thanks for your contribution!

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -112,30 +115,56 @@ public ProcessingResult processExchange(
...
+                // the record we are processing had the error 
+                // so we will force commit the offset prior 
+                // this will enable the current desired behavior to 
+                // retry the message 1 more time
+                //
+                // Note: without a more extensive look at handling of breakOnFirstError
+                // we will still need the lastResult so that we don't force 
+                // retrying this message over and over
+                //if (configuration.isBreakOnFirstErrorWithRetry()) {
+                    commitManager.forceCommit(topicPartition, record.offset() - 1);
```
CodeSmell, line 166: @orpiske 
some thoughts here on this

The `forceCommit` was what we had here so kept that for now. 
But based on some of the other conversations I think it may be worth changing that to `commit`.

That would defer how `breakOnFirstError` would work based on the `commitManager`

- no op would mean the route implementation would have to handle the offset using `KafkaManualCommit`
- using synch/asynch would mean that the record with the error would be retried 

the retry will vary still based on other configuration which is what I think contributes to this flag being so complex and why there are many opinions on how it should work (CAMEL-20089)

This may be a way to alter some of the behavior in 3.x until a refactor in 4.x

orpiske: Understood. I think that's fair. IMHO, it will be important to include that on the documentation as well, so users can refer to this behavior and understand what to expect. 

wdyt?

CodeSmell: I am good w/ what Camel team thinks is best.
The change here using `forceCommit` will fix the 2 reported issues so would like to see at least that move forward if you don't see any problems.

Looking at the code, if we did make some more changes, we may be able to remove the `lastResult` object that is passed around as well since we have access to the Kafka record. 

I can update the docs w/ whatever the new behavior is but may need a pointer as to where/how to do that

orpiske:
@CodeSmell now that we have that other ticket merged, I'll focus on this. I haven't had the time to trigger a test on this on my CI, but I'll be doing this today.

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaBreakOnFirstErrorSeekIssueIT.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+    
+    private void ifIsFifthRecordThrowException(Exchange e) {
+        if (e.getMessage().getBody().equals("5")) {
+            throw new RuntimeException("ERROR_TRIGGERED_BY_TEST");
+        }
+    }
+    
+    private void publishMessagesToKafka() {
+        final List<String> producedRecordsPartition0 = List.of("5", "6", "7", "8"); //, "9", "10", "11");
```
CodeSmell, line 132: This test isn't complete as I don't think increasing the partitions is working

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessor.java
```diff
@@ -149,14 +146,30 @@ private boolean processException(
...
+                // retrying this message over and over
+                // commitManager.forceCommit(topicPartition, record.offset() - 1);
+                
+                // we should just do a commit (vs the original forceCommit)
+                // when route uses NOOP Commit Manager it will rely
+                // on the route implementation to explicitly commit offset
+                // when route uses Synch/Asynch Commit Manager it will 
+                // ALWAYS commit the offset for the failing record
+                // and will ALWAYS retry it
+                commitManager.commit(topicPartition);
```
CodeSmell, line 172: This is the key change in behavior
It goes beyond a fix and gives route implementers ability to choose how breakOnFirstError should work using the commit manager

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaBreakOnFirstErrorWithBatchUsingKafkaManualCommitIT.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void kafkaBreakOnFirstErrorBasicCapabilityWithoutOnExcepton() throws Exception {
+        to.reset();
+        to.expectedMessageCount(7);
+        
+        // old behavior before the fix
+        // message-3 causes an error 
+        // and breakOnFirstError will cause it to be retried 1x
+        // then we move on
+        //to.expectedBodiesReceivedInAnyOrder("message-0", "message-1", "message-2", "message-3", "message-3", "message-4", "message-5");
```
CodeSmell, line 85: IMO this new behavior is what the route really wants

CodeSmell:
Moved changes to a separate PR

PR #11938, opened 2023-11-08
CAMEL-20085: aws2-sqs - catch any SdkException when deleting message

simon-ras:
# Description

Before this change, the exception would be logged in UnitOfWorkHelper  which catches any Throwable as the last option. The proper solution is  to catch it already in the consumer and let it be handled by the  exception handler.

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

oscerd:
@simon-ras can you backport on 4.0.x and check on 3.21.x and 3.x? Thanks

simon-ras:
> @simon-ras can you backport on 4.0.x and check on 3.21.x and 3.x? Thanks

I'm not actually sure that this is worth back-porting. There is nothing functionally changed from commit. What do you think?

oscerd:
ah yes, this is categorized as improvement. No need to backport. Thanks for checking.

PR #11945, opened 2023-11-08
CAMEL-19526: replaced the Thread.sleep with the best option or removed it.

rafamqrs:
# Description

Replaced Thread.sleep() in camel-jms.

# Target

# Tracking

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/async/MyAsyncProducer.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/async/MyAsyncProducer.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.component.jms.async;
 
+import java.time.Duration;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -23,6 +24,7 @@ import org.apache.camel.AsyncCallback;
 import org.apache.camel.CamelExchangeException;
 import org.apache.camel.Exchange;
 import org.apache.camel.support.DefaultAsyncProducer;
+import org.awaitility.Awaitility;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,7 +53,10 @@ public class MyAsyncProducer extends DefaultAsyncProducer {
         executor.submit(() -> {
 
             LOG.info("Simulating a task which takes {} millis to reply", getEndpoint().getDelay());
-            Thread.sleep(getEndpoint().getDelay());
+
+            Awaitility.await()
+                    .pollDelay(Duration.ofMillis(getEndpoint().getDelay()))
+                    .until(() -> true);
 
             int count = counter.incrementAndGet();
             if (getEndpoint().getFailFirstAttempts() >= count) {
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/async/MyAsyncProducer.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/async/MyAsyncProducer.java
@@ -16,7 +16,6 @@
  */
 package org.apache.camel.component.jms.async;
 
-import java.time.Duration;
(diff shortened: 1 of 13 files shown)


oscerd (approved):
cc @orpiske

orpiske:
Thanks for the contribution! I'll be looking closely at this one later today ... currently busy with a CI issue.

orpiske (changes requested):
Just marking like this 'cause it's pending further review.

components/camel-jms/src/test/java/org/apache/camel/component/jms/async/MyAsyncProducer.java
```diff
@@ -51,7 +53,10 @@ public boolean process(final Exchange exchange, final AsyncCallback callback) {
         executor.submit(() -> {
 
             LOG.info("Simulating a task which takes {} millis to reply", getEndpoint().getDelay());
-            Thread.sleep(getEndpoint().getDelay());
+
+            Awaitility.await()
+                    .pollDelay(Duration.ofMillis(getEndpoint().getDelay()))
+                    .until(() -> true);
```
orpiske, line 59: I think Thread.sleep is fine here.

components/camel-jms/src/test/java/org/apache/camel/component/jms/issues/TempReplyToIssueTest.java
```diff
@@ -75,7 +78,9 @@ public String handleMessage(
         producer.stop();
 
         // sleep a bit so Camel will send the reply a bit later
-        Thread.sleep(1000);
+        Awaitility.await()
+                .pollDelay(Duration.ofMillis(1000))
+                .until(() -> true);
```
orpiske, line 83: I think `Thread.sleep` is fine here.

components/camel-jms/src/test/java/org/apache/camel/component/jms/temp/JmsReconnectManualTest.java
```diff
@@ -102,7 +105,9 @@ public void configure() {
          * Wait long enough for the jms client to do a full reconnect. In the Tibco EMS case this means that a Temporary
          * Destination created before is invalid now
          */
-        Thread.sleep(5000);
+        Awaitility.await()
+                .pollDelay(Duration.ofMillis(5000))
+                .until(() -> true);
```
orpiske, line 110: I think `Thread.sleep` is fine here.

orpiske (changes requested):
Thanks for your contribution!

There are few places where I think it would be better to retain the `Thread.sleep` because the proposed solution with Awaitility makes the code harder to read (and, also, the adequate solution would be to reword the tests so that we can use adequate synchronization mechanisms - but that's beyond the point of this change).

Also, one more note about the commit message. Please make sure to write the commit message according to our standards. Trying doing something like this: 

```
CAMEL-19526: replaced the thread.sleep with the best option or removed
```

components/camel-jms/src/test/java/org/apache/camel/component/jms/temp/JmsReconnectManualTest.java
rafamqrs: Makes Sense, thank you!

rafamqrs:
> Thanks for your contribution!
> 
> There are few places where I think it would be better to retain the `Thread.sleep` because the proposed solution with Awaitility makes the code harder to read (and, also, the adequate solution would be to refactor the tests so that we can use adequate synchronization mechanisms - but that's beyond the point of this change).
> 
> Also, one more note about the commit message. Please make sure to write the commit message according to our standards. Trying doing something like this:
> 
> ```
> CAMEL-19526: replaced the thread.sleep with the best option or removed
> ```

My bad, I ended up forgetting to write the Jira task on the commit message and keep the commit message according to the standards.

orpiske (approved):
Thanks for the contribution! LGTM

PR #11959, opened 2023-11-09
CAMEL-20044 fix to handle commit on breakOnFirstError

CodeSmell:
# Description

This PR offers a fix for breakOnFirstError where the incorrect partition:offset when a forceCommit is done 

# Target

This PR is off the camel-3.21.x branch

# Tracking
CAMEL-20044 
CAMEL-19894

PR #11960, opened 2023-11-09
CAMEL-20044 Removing the lastResult used in breakOnFirstError 

CodeSmell:
# Description

This PR is adding onto what is in https://github.com/apache/camel/pull/11959
It shows that we can remove the `lastResult` from the code if we go with the changes proposed in this PR

# Target

camel-3.21.x

# Tracking
CAMEL-20044

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -307,7 +307,7 @@ public class KafkaFetchRecords implements Runnable {
 
             long pollTimeoutMs = kafkaConsumer.getEndpoint().getConfiguration().getPollTimeoutMs();
             Duration pollDuration = Duration.ofMillis(pollTimeoutMs);
-            
+
             if (LOG.isTraceEnabled()) {
                 LOG.trace("Polling {} from {} with timeout: {}", threadId, getPrintableTopic(), pollTimeoutMs);
             }
@@ -325,7 +325,7 @@ public class KafkaFetchRecords implements Runnable {
 
                 ProcessingResult result = recordProcessorFacade.processPolledRecords(allRecords);
                 updateTaskState();
-                
+
                 if (result != null && result.isBreakOnErrorHit() && !this.state.equals(State.PAUSED)) {
                     LOG.debug("We hit an error ... setting flags to force reconnect due to breakOnFirstError");
                     // force re-connect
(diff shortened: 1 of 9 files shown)


components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessorFacade.java
```diff
@@ -54,49 +54,53 @@ private boolean isStopping() {
...
+                    if (!consumerListener.afterProcess(result)) {
                         commitManager.commit(partition);
-                        return lastResult;
+                        return result;
                     }
                 }
             }
 
-            if (!lastResult.isBreakOnErrorHit()) {
+            if (!result.isBreakOnErrorHit()) {
```
CodeSmell, line 96: this part of the code is the only reason we need the offset  in `ProcessedResult`
may be worth trying to move it inside `createExchange`

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/KafkaRecordProcessorFacade.java
```diff
@@ -126,7 +129,7 @@ private ProcessingResult processRecord(
 
         ProcessingResult currentResult
                 = kafkaRecordProcessor.processExchange(exchange, partition, partitionHasNext,
-                        recordHasNext, record, lastResult, camelKafkaConsumer.getExceptionHandler());
+                        recordHasNext, record, camelKafkaConsumer.getExceptionHandler());
 
         if (!currentResult.isBreakOnErrorHit()) {
             commitManager.recordOffset(partition, currentResult.getPartitionLastOffset());
```
CodeSmell, line 135: this part of the code is one of 2 reasons I can see that we need the offset in `ProcessedResult`
may be worth trying to move it inside `createExchange`

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -297,6 +297,7 @@ private void subscribe() {
     }
 
     protected void startPolling() {
+        // not sure what we are using this for
         long partitionLastOffset = -1;
```
CodeSmell, line 301: saw this and it is used in Polling Strategy but I didn't see where we changed it

orpiske:
Attaching test logs from Github. 

[components-camel-kafka.log](https://github.com/apache/camel/files/13317865/components-camel-kafka.log)

CodeSmell:
Will add another PR with this for consideration

PR #11969, opened 2023-11-10
CAMEL-20062: left-over formatting fixes

orpiske:


oscerd (approved):

orpiske:
@oscerd @davsclaus Probably good to get this one on the 4.2 to avoid causing trouble during the release 

Obs.: not merging myself as I don't know if you are creating the branches or not, etc.

oscerd:
It's already on-going. The branches have been created and the process should be going

orpiske:
> It's already on-going. The branches have been created and the process should be going

Ahhh, no worries. I'll back port this later, in this case. Thanks for the follow up!

davsclaus (approved):

PR #11973, opened 2023-11-11
CAMEL-20101: camel-jolt - query parameter allowContextMapAll is ignored

sdurrenmatt:
# Description

Just excluding **allowContextMapAll** property as suggested. Thought to use **ExchangeHelper.createVariableMap** to populate the context map by default, but it would mean serialize the exchange headers and properties to valid JOLT context (similarly to JsltEndpoint). Not sure it's worth as we can already provide a custom context.

This option is a bit misleading since there is no context by default. Also one could think allowContextMapAll goes with CamelJoltContext, and allowTemplateFromHeader with CamelJoltResourceUri. In the end only allowTemplateFromHeader is necessary.

# Target

# Tracking


davsclaus:
Some files need to regen, let me do that

PR #12018, opened 2023-11-16
[WIP] Improve the documentation for httpHeaders parameter.

InfoSec812:
# Description

The documentation for the `httpHeaders` parameter of the `stream` component is ambiguous about how to format the parameter. It appears from the documentation that one would use `HEADER: VALUE`; but the code requires using `HEADER=VALUE`

# Target

# Tracking

Having issues getting this to run as there are other modules preventing the build.

### DIFF
--- a/components/camel-stream/src/main/java/org/apache/camel/component/stream/StreamEndpoint.java
+++ b/components/camel-stream/src/main/java/org/apache/camel/component/stream/StreamEndpoint.java
@@ -145,8 +145,8 @@ public class StreamEndpoint extends DefaultEndpoint {
 
     /**
      * When using stream:http format, this option specifies optional http headers, such as Accept: application/json.
-     * Multiple headers can be separated by comma. The format of headers should be {@code HEADER=VALUE}. An example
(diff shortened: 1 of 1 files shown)


oscerd (approved):

components/camel-stream/src/main/java/org/apache/camel/component/stream/StreamEndpoint.java
```diff
@@ -145,7 +145,8 @@ public String getHttpHeaders() {
 
     /**
      * When using stream:http format, this option specifies optional http headers, such as Accept: application/json.
-     * Multiple headers can be separated by comma.
+     * Multiple headers can be separated by comma. The format of headers should be {@code HEADER=VALUE}. An example
```
davsclaus, line 148: We usually do not use @code or other formatting as the text is being extracted for tooling as well and they use plain text. So it would be good to get that removed, thanks.

davsclaus (approved):

InfoSec812:
Just to be clear, I was unable to run the maven build to generate files due to errors in other modules, so there may need to be other changes.

PR #12041, opened 2023-11-17
(chores) camel-test-infra-etcd3: updated test image to the latest version

orpiske:
Updated to version 3.5.10

While looking at #12037 I noticed that our container image is outdated. This should bump to the latest version available (tested manually, it should be fine).

# Description

# Target

# Tracking


essobedo (approved):

oscerd (commented):
This is one the good reason of having dependabot running.. We could improve a lot: the downside is we could end up doing only deps upgrade/testing.

orpiske:
> This is one the good reason of having dependabot running.. We could improve a lot: the downside is we could end up doing only deps upgrade/testing.

It's a pity that dependabot can't pick the images on our test infra. It would make our life much easier.

PR #12056, opened 2023-11-17
(chores) camel-test-infra-elasticsearch: update ElasticSearch container to the latest version

orpiske:
Updates the container to 8.11.1


oscerd (approved):
LGTM, but I think SB will update to 8.11.x only with 3.2.x

orpiske:
> LGTM, but I think SB will update to 8.11.x only with 3.2.x

Should we wait on this one, them? We are already at the latest 8.10 ... so, not much harm IMHO.

orpiske:
Marking as draft to avoid merging this by mistake.

oscerd:
SB 3.2.0 is still using 8.10.4, but I think it's fine to move on.

orpiske:
Alright. Let's try and see how it goes ... worst case scenario, we rollback. Thanks @oscerd

PR #12066, opened 2023-11-17
CAMEL-19770: cleanup catches of Throwable in components

orpiske:


orpiske:
Pending further tests results on my CI.

orpiske:
Looking OK. Let's see how it goes on ASF CI.

PR #12067, opened 2023-11-17
CAMEL-20056: make it possible to set multiple headers at once.

klease:
Supported in Java, Yaml and Xml DSL.

# Description

# Target

# Tracking


oscerd (approved):

klease:
Failure in maven test due to org.apache.camel.support.jsse.TrustManagersParametersTest.testPropertyPlaceholders. I think this is due to parallel tests which are writing the same file in the target directory.

davsclaus:
> Failure in maven test due to org.apache.camel.support.jsse.TrustManagersParametersTest.testPropertyPlaceholders. I think this is due to parallel tests which are writing the same file in the target directory.

Yes it seems to be - we need to either isolate these tests or fix them so they have their private files

core/camel-core-model/src/main/java/org/apache/camel/model/SetHeadersDefinition.java
```diff
@@ -0,0 +1,118 @@
...
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlElementRef;
+import jakarta.xml.bind.annotation.XmlRootElement;
+
+import org.apache.camel.Expression;
+import org.apache.camel.builder.ExpressionBuilder;
+import org.apache.camel.spi.Metadata;
+
+/**
+ * A sequence of setHeader operations.
```
davsclaus, line 34: Can we update this to better tell the user what it does, something like:

Allows to set multiple headers at once on the message.

davsclaus (commented):
We need documentation in a file setHeaders-eip.adoc and in the camel-core-engine - copy the existing setHeader and modify this file.

davsclaus:
Much better description thanks - just noticed a little typo in "st the same time"

core/camel-core-model/src/main/java/org/apache/camel/model/SetHeadersDefinition.java
klease: setHeaders-eip.adoc added and description updated.

klease:
> Much better description thanks - just noticed a little typo in "st the same time"

Rats, silly me! That will change all the generated files again :-(.  By the way, I can't figure out how the nav.doc for the eips is generated since building docs does the others but not that one.

davsclaus:
> > Much better description thanks - just noticed a little typo in "st the same time"
> 
> Rats, silly me! That will change all the generated files again :-(. By the way, I can't figure out how the nav.doc for the eips is generated since building docs does the others but not that one.

You can try (if this works)

cd docs
mvn clean install

davsclaus (approved):

davsclaus:
Also it would be good to add a TIP in setHeader doc to link to this new one so users can more easily be aware of this variation

klease:
Thanks for committing Claus. It looks like the generation of the nav.adoc for the EIPs is broken. At least it's not done by running clean install in docs and it appears that that this is due to a not very recent change in gulpfile.js

davsclaus:
Okay maybe @zregvart can help out

zregvart:
@klease can you file a JIRA for the issue you're facing? I can run `./mvnw -pl :docs` successfully on my end.

klease:
@zregvart CAMEL-20135

PR #12089, opened 2023-11-20
CAMEL-20115: Support for Start Date and End Date in camel-quartz

prasanthrao:
# Description

Fix for [CAMEL-20115](https://issues.apache.org/jira/browse/CAMEL-20115) Support for Start Date and End Date in camel-quartz

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-20115) filed for the change (usually before you start working on it).

### DIFF
--- a/components/camel-quartz/src/main/java/org/apache/camel/component/quartz/CamelJob.java
+++ b/components/camel-quartz/src/main/java/org/apache/camel/component/quartz/CamelJob.java
@@ -16,8 +16,8 @@
  */
 package org.apache.camel.component.quartz;
 
-import java.util.Date;
 import java.util.Collection;
+import java.util.Date;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -59,7 +59,7 @@ public class CamelJob implements Job, InterruptableJob {
                         context.getTrigger().getStartTime(), context.getTrigger().getEndTime(), context);
                 return;
             }
-            
+
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Running CamelJob jobExecutionContext={}", context);
             }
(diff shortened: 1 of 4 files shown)


oscerd (commented):
It looks good, but I think it should go on main, since it is a new feature.

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/CamelJob.java
```diff
@@ -94,6 +101,18 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
         }
     }
 
+    private boolean hasTriggerExpired(JobExecutionContext context) {
+        Date fireTime = context.getFireTime();
+        if (context.getTrigger().getStartTime() != null && fireTime.before(context.getTrigger().getStartTime())) {
+            // Trigger invalid as Fire Time is before Start Time
+            return true;
+        } else if (context.getTrigger().getEndTime() != null && fireTime.after(context.getTrigger().getEndTime())) {
```
davsclaus, line 109: I think this should not be an else if, as you can have set BOTH start and end time, and then you need to check for both of them.

davsclaus (changes requested):

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/CamelJob.java
prasanthrao: Indeed I missed the case, and corrected it in the new commit.

oscerd:
Shouldn't this need to go on main first?

prasanthrao:
> Shouldn't this need to go on main first?

Should I also create a PR for the main branch?

davsclaus:
3.1.4 is EOL and we should only do this change in newer releases

PR #12098, opened 2023-11-20
CAMEL-20115: Support for Start Date and End Date in camel-quartz

prasanthrao:
# Description

Cherry-pick of https://github.com/apache/camel/pull/12089

# Target

# Tracking
- [ x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-20115) filed for the change (usually before you start working on it).

- [x ] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes

### DIFF
--- a/components/camel-quartz/src/main/docs/quartz-component.adoc
+++ b/components/camel-quartz/src/main/docs/quartz-component.adoc
@@ -166,18 +166,18 @@ The timeZone value is the values accepted by `java.util.TimeZone`.
 
 == Specifying start date
 
-The Quartz Scheduler allows you to configure start date per trigger. You can provide the start date as TimeInMillis as follows:
+The Quartz Scheduler allows you to configure start date per trigger. You can provide the start date in the date format yyyy-MM-dd'T'HH:mm:ssz.
 
 ----
-quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.startAt=1700452232554
+quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.startAt=2023-11-22T14:32:36UTC
 ----
 
 == Specifying end date
 
-The Quartz Scheduler allows you to configure end date per trigger. You can provide the end date as TimeInMillis as follows:
+The Quartz Scheduler allows you to configure end date per trigger. You can provide the end date in the date format yyyy-MM-dd'T'HH:mm:ssz.
 
 ----
-quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.endAt=1700452232554
+quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.endAt=2023-11-22T14:32:36UTC
 ----
 
 == Configuring misfire instructions
--- a/components/camel-quartz/src/main/docs/quartz-component.adoc
+++ b/components/camel-quartz/src/main/docs/quartz-component.adoc
@@ -166,7 +166,8 @@ The timeZone value is the values accepted by `java.util.TimeZone`.
 
 == Specifying start date
 
-The Quartz Scheduler allows you to configure start date per trigger. You can provide the start date in the date format yyyy-MM-dd'T'HH:mm:ssz.
+The Quartz Scheduler allows you to configure start date per trigger. You can provide the start date 
+in the date format yyyy-MM-dd'T'HH:mm:ssz.
 
 ----
 quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.startAt=2023-11-22T14:32:36UTC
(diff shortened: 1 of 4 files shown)


oscerd (approved):

davsclaus:
org.apache.camel.component.quartz.QuartzManuallyTriggerJobTest.testQuartzCronRoute | 10.54s | FAILURE
-- | -- | --
org.apache.camel.component.quartz.QuartzAddRoutesAfterCamelContextStartedTest.testAddRoutes | 10.96s | FAILURE
org.apache.camel.component.quartz.QuartzAddDynamicRouteTest.testAddDynamicRoute | 11.35s | FAILURE

davsclaus:
Do you have test errors locally ?

prasanthrao:
I might have found the issue. In [QuartzEndpoint](https://github.com/apache/camel/blob/99da2b85fbd897c5c52cff86eb285aedbdfb9b5e/components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java#L479) during createTrigger, the startAt is configured always to be few duration later than the current time. The scheduler misfires before the start time, and the misfire is skipped in CamelJob due to the new changes.

components/camel-quartz/src/main/docs/quartz-component.adoc
```diff
@@ -164,6 +164,22 @@ quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.timeZone=Europ
...
+----
+
+== Specifying end date
+
+The Quartz Scheduler allows you to configure end date per trigger. You can provide the end date as TimeInMillis as follows:
+
+----
+quartz://groupName/timerName?cron=0+0/5+12-18+?+*+MON-FRI&trigger.endAt=1700452232554
+----
+
```
orpiske, line 182: It's probably important do add note here stating that, when using a system clock such as `System.currentTimeMillis()`, users may be susceptible to problems caused by time drifts (such as when the system clock gets automatically adjusted after an inquiry to an NTP server) and unpredictable behavior during daylight saving time changes. 

Ideally, we should provide a duration that can be computed from a monotonic clock in complement to this.

davsclaus: Yes its also cumbersome to set a time as time millis - Its better its using date/timezone like you can do in simple language - also we should only do this in newer versions 4.3.0 and 3.22.0

prasanthrao: Is there any existing example for setting the time as date/timezone? I was inclined to use ISO-8601 but was thinking of something like "2001-07-04T12:08:56CET" for better readability, especially for DST based times.
I am presently using 3.14 due to some hard dependencies on Java 8. Hence I would urge you to also consider 3.14 as a candidate for shipping this feature.

orpiske: > Is there any existing example for setting the time as date/timezone? I was inclined to use ISO-8601 but was thinking of something like "2001-07-04T12:08:56CET" for better readability, especially for DST based times. I am presently using 3.14 due to some hard dependencies on Java 8. Hence I would urge you to also consider 3.14 as a candidate for shipping this feature.

I'd probably make this configurable.

prasanthrao: Can you share an example of the date format? I will look into it and submit the change

prasanthrao: I have replaced currentTimeMillis with date format "yyyy-MM-dd'T'HH:mm:ssz". Let me know if there is a better way to handle this case. Since cron does not support milliseconds I was inclined to the stated format.

prasanthrao: @orpiske @davsclaus can you please share your inputs/erviews?

components/camel-quartz/src/test/java/org/apache/camel/component/quartz/QuartzCronRouteWithStartDateEndDateTest.java
```diff
@@ -0,0 +1,65 @@
...
+            	Calendar calendar = Calendar.getInstance();
+            	calendar.setTimeZone(TimeZone.getTimeZone("UTC"));
+                calendar.add(Calendar.SECOND, 3);
+                Date startDate = calendar.getTime();
+                calendar.add(Calendar.SECOND, 2);
+                Date endDate = calendar.getTime();
+                
+                // triggers every 1th second at precise 00,01,02,03..59 with startAt and endAt exactly 2 second apart.
+                // configuration will create a maximum of three messages
+                from("quartz://myGroup/myTimerName?cron=0/1 * * * * ?&trigger.startAt=" + dateFormat.format(startDate)
```
orpiske, line 60: I hope I am not nitpicking too much but, maybe use `fromF` for more legibility. These concatenations makes it harder to read.

components/camel-quartz/src/main/docs/quartz-component.adoc
orpiske: Thanks. I am fine with the approach. I would still like to see a note about time possible time drifts and inconsistent behavior during daylight changes (because it's caused by the nature of using time specifications instead of duration).

prasanthrao: Thanks. I have updated the adoc with a brief note.

components/camel-quartz/src/test/java/org/apache/camel/component/quartz/QuartzCronRouteWithStartDateEndDateTest.java
prasanthrao: Thanks. Changed to `fromF`

orpiske (approved):
Thanks for your contribution, LGTM.

davsclaus (approved):

PR #12119, opened 2023-11-21
CAMEL-19997: camel-smb - Update first version

essobedo:
## Motivation

The first version of the SMB component is set to a snapshot version instead of a final version.

## Modifications:

* Change the first version to 4.3
* Remove the logging dependencies as they are inherited from the direct parent pom file
* Set the scope of `camel-test-infra-smb` to `test`

### DIFF
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/smb.json
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/smb.json
@@ -5,7 +5,7 @@
     "title": "SMB",
     "description": "Receive files from SMB (Server Message Block) shares.",
     "deprecated": false,
-    "firstVersion": "4.3.0-SNAPSHOT",
+    "firstVersion": "4.3.0",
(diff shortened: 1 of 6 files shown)


oscerd (approved):

davsclaus:
Claus Ibsen: @Nicolas Filotto camel-smb first version is a bit "wrong" as its added in 4.3.0 and is preview level

Claus Ibsen: the 3.22 is a "special case" and you can have it as first version 3.22 in this release only

Claus Ibsen: but keep main as 4.3

essobedo:
@davsclaus I understand, then it should rather be `4.3.0` instead of `4.3.0-SNAPSHOT`, WDYT?

davsclaus (approved):

davsclaus:
> @davsclaus I understand, then it should rather be `4.3.0` instead of `4.3.0-SNAPSHOT`, WDYT?

Yes that is correct

orpiske (approved):

PR #12138, opened 2023-11-22
Camel 16099: throttler EIP - throttle on concurrent requests

jono-morris:
Updated the Throttler to use the number of concurrent requests as the throttling measure instead of the number of requests per period. 

Now uses Semaphore to bound the rate with the bound of the semaphore set using the throttle rate. Semaphore 'fair' setting set to ensure fifo behaviour.

Maintains asyc resubmission when config allows with processes being submitted for immediate execution to the executor. 

Some tests removed that were based on testing the number of requests per period.


oscerd (approved):

davsclaus:
Can you add a note in the 4.3 upgrade guide about this change as end users may need to migrate

davsclaus (approved):

davsclaus:
There are some throttling unit test failures on CI after 2 build attempts. Not sure if its a bit flaky with this change ?

jono-morris:
Let me update the guide and take a look at those failing tests.

davsclaus:
@jono-morris in the migration guide, it would be good to add a before vs after example how to migrate the throttle EIP

jono-morris:
Added a note to the migration guide  https://github.com/apache/camel/pull/12200

pai911:
Thank you all for the hard work improving Apache Camel!

I have a question about this change that I would like to ask.

With this new change, how are we going to set the rate limit (10 req/sec) as specified by some API endpoints?

The new throttle EIP design doesn't seem to be practical since most APIs aren't designed with a rate limit of "max concurrent requests"?

orpiske:
> Thank you all for the hard work improving Apache Camel!
> 
> I have a question about this change that I would like to ask.
> 
> With this new change, how are we going to set the rate limit (10 req/sec) as specified by some API endpoints?
> 
> The new throttle EIP design doesn't seem to be practical since most APIs aren't designed with a rate limit of "max concurrent requests"?

I think we, somehow, missed this valid question. I was looking at one of the emails on our user list and someone raised the very same question. 


@jono-morris, I think, maybe, we should complement the migration guide with this information or a guidance for some scenarios (such as - what to do when there is a need to rate limit it when the time slot is greater than 1000msec and other use cases).

oscerd:
> > Thank you all for the hard work improving Apache Camel!
> > I have a question about this change that I would like to ask.
> > With this new change, how are we going to set the rate limit (10 req/sec) as specified by some API endpoints?
> > The new throttle EIP design doesn't seem to be practical since most APIs aren't designed with a rate limit of "max concurrent requests"?
> 
> I think we, somehow, missed this valid question. I was looking at one of the emails on our user list and someone raised the very same question.
> 
> @jono-morris, I think, maybe, we should complement the migration guide with this information or a guidance for some scenarios (such as - what to do when there is a need to rate limit it when the time slot is greater than 1000msec and other use cases).

There were other questions about throttling and about the removal of timeoutPeriodmillis.. I think we should not remove that option.

jono-morris:
> > > Thank you all for the hard work improving Apache Camel!
> > > I have a question about this change that I would like to ask.
> > > With this new change, how are we going to set the rate limit (10 req/sec) as specified by some API endpoints?
> > > The new throttle EIP design doesn't seem to be practical since most APIs aren't designed with a rate limit of "max concurrent requests"?
> > 
> > 
> > I think we, somehow, missed this valid question. I was looking at one of the emails on our user list and someone raised the very same question.
> > @jono-morris, I think, maybe, we should complement the migration guide with this information or a guidance for some scenarios (such as - what to do when there is a need to rate limit it when the time slot is greater than 1000msec and other use cases).
> 
> There were other questions about throttling and about the removal of timeoutPeriodmillis.. I think we should not remove that option.

Firstly, sorry about the delay in responding.  I've taken away some learnings.

About this change. 

Previously the Throttler employed a fixed window algorithm, allowing a fixed number of calls in a particular period, e.g. 2 requests per minute.  This is susceptible to bursts in requests. For example, if the limit is set at 1000 requests/hour, all 1000 requests might be made in the first minute of the window.

With this in mind the algorithm was shifted to more of a leaky bucket implementation using a semaphore with a fixed number of permits.  Callers try to acquire a permit, blocking if necessary when all available permits have been acquired.  When a caller completes processing the permit it holds is released.  This approach limits the number of concurrent requests, there is no notion of rate.

Let's discuss and find the best way forward.

jono-morris:
@pai911 See https://issues.apache.org/jira/browse/CAMEL-20355

pai911:
> @pai911 See https://issues.apache.org/jira/browse/CAMEL-20355

Thank you so much!

PR #12146, opened 2023-11-22
Properly generate EIP nav.adoc

zregvart:
# Description

Seems that the EIP nav.adoc, prior to CAMEL-20135, and it seems not entirely resolved by CAMEL-20135, was not generated properly. This attempts to generate it by including all the Asciidoc from the EIP module within the nav.adoc.

The notion of Asciidoc module without a source, i.e. one that doesn't need to be symlinked or manipulated in any way by Gulp is introduced. This is to provide the list of files to `createNav` correctly.

# Target

# Tracking


zregvart:
@klease can you have a look at this approach?

davsclaus (approved):

oscerd (approved):

zregvart:
> * I thought the :eips: in the xrefs was required. But maybe not in the nav file itself?

xref can contain only the name of the file if that file resides in the [same module and component version](https://docs.antora.org/antora/latest/page/page-links/#pages), and in this case `nav.adoc` and other `.adoc` are in the same `eips` module.

> * Not being a js expert, I'm not sure, but it looks to me like your change will add a createNav task both with and without a filter for all the sources which have both a json and an asciidoc with source property, and they all do except eips.

Yeah, I've added an [explicit](https://github.com/apache/camel/pull/12146/files#diff-a0e44ac534103410f3aa0fdacfc392140e8673f3454e0e842c3e6c82fb108810R451) `!asciidoc.source`, though Gulp maintains unique names of tasks, so the `nav:asciidoc:eips` task in the [`json` branch](https://github.com/apache/camel/pull/12146/files#diff-a0e44ac534103410f3aa0fdacfc392140e8673f3454e0e842c3e6c82fb108810R445) should override the task in the [`asciidoc`](https://github.com/apache/camel/pull/12146/files#diff-a0e44ac534103410f3aa0fdacfc392140e8673f3454e0e842c3e6c82fb108810R417) branch. I think having the explicit condition there helps though, at least when trying to grok the code.

klease:
@zregvart Thanks for the explanations and the fix!

zregvart:
> @zregvart Thanks for the explanations and the fix!

Thanks for having a look @klease :)

PR #12233, opened 2023-11-28
Camel 20158 fix throttle tests

jono-morris:
Use a semaphore in Throttle unit tests to verify that the throttle eip allows no more than the required number of concurrent requests, camel-spring-xml tests are still disabled.


jono-morris:
I ran the tests half a dozen times or so without issue locally with these changes, although I had trouble reliably reproducing the failures locally.  I'm keen to see how they work here.

davsclaus (approved):

davsclaus:
the test failure is not related to throttler

PR #12240, opened 2023-11-28
CAMEL-20152: camel-jetty - Allow to configure the multipart

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-20152

## Motivation

In the case of multipart, the content is kept in memory which can lead to an OOME, it is due to the fact that the `fileSizeThreshold` is set to `0` by default which should mean that the content is stored on disk according to the Jetty Javadoc but it is not what the Jetty code expects.

## Modifications:

* Proposes new properties to be able to configure the multipart
* Add a unit test to ensure that the file content is stored on disk


oscerd (approved):

mikadev (approved):

components/camel-jetty-common/src/main/java/org/apache/camel/component/jetty/JettyHttpComponent.java
```diff
@@ -1142,15 +1194,20 @@ protected CamelServlet createServletForConnector(
...
-        File file = File.createTempFile("camel", "");
-        boolean result = file.delete();
-        if (!result) {
-            LOG.error("failed to delete {}", file);
+        String location = endpoint.getFilesLocation();
+        if (location == null) {
+            File file = File.createTempFile("camel", "");
+            boolean result = file.delete();
+            if (!result) {
+                LOG.error("failed to delete {}", file);
```
davsclaus, line 1202: WARN - Also you can use FileUtil.deleteFile from camel-core

components/camel-jetty-common/src/main/java/org/apache/camel/component/jetty/JettyHttpComponent.java
```diff
@@ -1016,6 +1028,46 @@ public void setSendServerVersion(boolean sendServerVersion) {
...
+    public void setMaxFileSize(long maxFileSize) {
+        this.maxFileSize = maxFileSize;
+    }
+
+    public long getMaxRequestSize() {
+        return maxRequestSize;
+    }
+
+    @Metadata(description = "The maximum size allowed for multipart/form-data requests. -1 means no limit",
+              defaultValue = "-1", label = "consumer")
```
davsclaus, line 1046: Maybe these should also be advanced label like in the endpoint

components/camel-jetty-common/src/main/java/org/apache/camel/component/jetty/JettyHttpComponent.java
essobedo: Actually, if you check well it is not my code, this is the existing code that I simply moved into a if-block

components/camel-jetty-common/src/main/java/org/apache/camel/component/jetty/JettyHttpComponent.java
essobedo: It makes sense indeed

essobedo:
Due to https://github.com/jetty/jetty.project/issues/10936

davsclaus (approved):

joakime:
Is Camel using Jetty 12 with the ee10 environment?

These changes seem to be for Jetty 11, which is for Servlet 5.0.

This API spec behavior, where fileSizeThreshold value of 0 is documented, first shows up on Servlet 6.0.
That language is _not_ present in the API spec on Servlet 5.0. - https://github.com/jakartaee/servlet/blob/5.0.0-RELEASE/api/src/main/java/jakarta/servlet/annotation/MultipartConfig.java#L65-L70

Jetty 12 on the ee10 environment already supports this spec behavior.

davsclaus:
Yes its Jetty 11 in the source, however we are also working on upgrading to Jetty 12 for the upcoming Camel 4.3 release. But older Camel releases are using Jetty 11 and 9.

joakime:
Note, Jetty 9 is now at End of Community Support.

* https://github.com/jetty/jetty.project/issues/7958

essobedo:
Please note that even if it is mentioned only in EE 10, it is actually a clarification see https://github.com/jakartaee/servlet/issues/111, for example, in Glassfish, it is how it [has been implemented even in very old versions](https://github.com/eclipse-ee4j/glassfish/blob/5.1.0-RC1/appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/ThresholdingOutputStream.java#L222-L226) that is why I proposed it in version 11 but I can understand your point of view.

essobedo:
And IMHO, it makes more sense to consider that a threshold set to 0 is immediately reached instead of never being reached like it was set to Integer.MAX_VALUE which is somehow the current behavior in any version of Jetty

PR #12258, opened 2023-11-29
consumer dev console

davsclaus:
# Description

# Target

# Tracking


oscerd (approved):

oscerd:
I think we should collect an initial lists of components to supports, so we can create a console for each one.

davsclaus:
Yeah adding dev consoles makes it much easier to grab data for jbang and other tools / and for end users to see it when building integrations with /q/dev/xxxx

You can find a list of existing consoles, just by

git grep "@DevConsole"

PR #12269, opened 2023-11-30
CAMEL-20156: camel-openapi-java - attempt to convert allowValues by S…

zhfeng:
…chema::cast method

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
+++ b/components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.openapi;
 
+import java.text.SimpleDateFormat;
+
 import io.swagger.v3.core.util.Json;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.info.Info;
@@ -63,6 +65,7 @@ public class RestOpenApiReaderTest extends CamelTestSupport {
                         .dataType("array")
                         .arrayType("date-time")
                         .endParam()
+
                         .to("log:hi")
 
                         .get("/bye/{name}")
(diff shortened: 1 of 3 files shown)


components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
```diff
@@ -170,6 +251,14 @@ public void testReaderRead() throws Exception {
         assertTrue(json.contains("\"success\" : \"123\""));
         assertTrue(json.contains("\"error\" : \"-1\""));
         assertTrue(json.contains("\"type\" : \"array\""));
+        assertTrue(json.contains("\"enum\" : [ \"A\", \"B\", \"C\" ]"));
+        assertTrue(json.contains("\"enum\" : [ 1, 2, 3 ]"));
+        assertTrue(json.contains("\"enum\" : [ 1.0, 2.0, 3.0 ]"));
+        assertTrue(json.contains("\"enum\" : [ \"true\", \"false\" ]"));
```
zhfeng, line 257: It looks like swagger v2 does not convert `Boolean` value.  see

https://github.com/swagger-api/swagger-core/blob/v1.6.11/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java#L173-L192

So there is a String `true` and `false` . @jamesnetherton is this a problem?

oscerd (approved):

components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
jamesnetherton: I guess it's not a regression as such, right? The `else` clause for the type check would have been doing fallback to `StringSchema` in past releases anyway.

zhfeng: Yeah, I think so.

jamesnetherton (approved):

PR #12284, opened 2023-11-30
CAMEL-20163: cleanup parsing HTTP status range (camel-vertx-http)

orpiske:


components/camel-vertx/camel-vertx-http/src/main/java/org/apache/camel/component/vertx/http/VertxHttpEndpoint.java
```diff
@@ -60,18 +61,23 @@ public VertxHttpComponent getComponent() {
...
-                maxOkRange = Integer.parseInt(StringHelper.after(range, "-"));
-            } else {
+            if (!HttpUtil.parseStatusRange(range, this::setRanges)) {
                 minOkRange = Integer.parseInt(range);
                 maxOkRange = minOkRange;
             }
         }
     }
 
+    private void setRanges(int minOkRange, int maxOkRange) {
```
davsclaus, line 76: Is this method in use ?

oscerd (approved):

components/camel-vertx/camel-vertx-http/src/main/java/org/apache/camel/component/vertx/http/VertxHttpEndpoint.java
orpiske: Yes it is. It's referenced on line 69, by the `parseStatusRange`.

davsclaus (approved):

PR #12294, opened 2023-12-01
CAMEL-20174: camel-ssh: Provide support to configure algorithms

luigidemasi:


components/camel-ssh/src/main/java/org/apache/camel/component/ssh/SshConfiguration.java
```diff
@@ -62,6 +64,18 @@ public class SshConfiguration implements Cloneable {
...
+    private String kex;
+    @UriParam(label = "security")
+    private String ciphers;
+    @UriParam(label = "security")
+    private String macs;
+    @UriParam(label = "security")
+    private String signatures;
+    @UriParam(label = "advanced")
+    private String compressions;
+    @UriParam(label = "advanced,producer,consumer")
```
davsclaus, line 77: producer and consumer is default, (ie. common), so this can just be "advanced"

davsclaus (approved):

PR #12386, opened 2023-12-09
only extract the ClassPathXmlApplicationContext once

TestBoost:
# Description

This pull request is to make sure only extract the `ctx` once from the xml file `org/apache/camel/spring/xml/handler/ErrorHandlerDefinitionParser.xml`. There are four tests in this test class. Every test tries to call `getBean()` from ctx to get an instance with its name and type. However, `getBean()` is not changing contents in the `ctx`. In other words, given that all tests do not try to change the contents of this `ctx` context and they are just trying to read from it, we can just create the `ctx` once to speed up the this test class.

When run on our own machine, the test runtime can jump from `4.873 s` to `3.2581 s` after applying the change.

# Target

# Tracking

### DIFF
--- a/components/camel-spring-xml/pom.xml
+++ b/components/camel-spring-xml/pom.xml
@@ -102,6 +102,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-spring-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
             <artifactId>camel-management</artifactId>
             <scope>test</scope>
         </dependency>
--- a/components/camel-spring-xml/src/test/java/org/apache/camel/spring/xml/handler/ErrorHandlerDefinitionParserTest.java
+++ b/components/camel-spring-xml/src/test/java/org/apache/camel/spring/xml/handler/ErrorHandlerDefinitionParserTest.java
@@ -20,34 +20,26 @@ import org.apache.camel.Processor;
 import org.apache.camel.builder.LegacyDeadLetterChannelBuilder;
 import org.apache.camel.builder.LegacyDefaultErrorHandlerBuilder;
 import org.apache.camel.processor.errorhandler.RedeliveryPolicy;
+import org.apache.camel.spring.SpringTestSupport;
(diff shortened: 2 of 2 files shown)


davsclaus:
Thanks, the test should really extend CamelSpringTestSupport and be something like

```
public class ErrorHandlerDefinitionParserTest extends SpringTestSupport {

    @Override
    protected AbstractXmlApplicationContext createApplicationContext() {
        return new ClassPathXmlApplicationContext("org/apache/camel/spring/xml/handler/ErrorHandlerDefinitionParser.xml");
    }

```

And `ctx` should then be `applicationContext`.

Can you try to change the PR to do that, thanks.

TestBoost:
Thank you for the response. And I indeed find in the `CamelSpringTestSupport`, there is a place to only read the `context` once under certain conditions. I just refactored as you said but I am not sure if it looks OK.

Thanks a lot

davsclaus:
Oh we cannot depend on `camel-test-spring-junit5` as its a chicken and egg situation. As camel-spring-xml is a dependency on the test module, so you end up with circular dependency.

TestBoost:
Oh, I see and I did not notice this. So do we just extend `SpringTestSupport` rather than `CamelSpringTestSupport`? And it seems `SpringTestSupport` does not have the same way to check `isCreateCamelContextPerClass()`.

Thank you very much!

davsclaus:
Did you notice that this was the only test that did this, or did you find others in camel-spring-xml

TestBoost:
Thank you so much. I try to search in the test codes in camel-spring-xml and only find this test look like this as of now. However, there are indeed some tests like `ParentContextRegistryTest` (extends SpringTestSupport) and `FindSingleByTypeTest` (extends ContextTestSupport). For these two tests, maybe it's also better to choose `CamelSpringTestSupport` instead if there are no dependencies issues.

klease:
Most of the tests in camel-spring-xml extend `org.apache.camel.spring.SpringTestSupport`. That also provides the method `getMandatoryBean(Class<T> type, String name)` which would allow to simplify the code in your test.

davsclaus:
Sorry closing old PRs - the camel-spring-xml should ideally be built before the test modules. And its fine for us as is.

PR #12450, opened 2023-12-14
CAMEL-20225: move time-related information to a separate abstraction

orpiske:
This is the first part of the task. At this moment, it still relies on System.currentTimeMillis to compute the elapsed time in order to retain the current behavior and ensure the refactoring is solid. A subsequent PR will then add a new field to track the duration separatelly from the real clock part.


orpiske:
Previous tests were all good, but I'm waiting for one last test on my CI.

davsclaus:
The `Clock` API seems generic so it may not be tied only to Exchange - could be that its API doc should be made more generic.

oscerd (approved):

orpiske:
> The `Clock` API seems generic so it may not be tied only to Exchange - could be that its API doc should be made more generic.

Indeed! I'm starting small, but I think it could be used it on `Message` and a few other places in the future.

PR #12453, opened 2023-12-14
CAMEL-20113 camel/tests consolidating

kulagaIA:
# Description
Deleted tests/camel-partial-classtapth-test as there are many tests doing this check as part of their logic, including the type converter tests.
Deleted tests/camel-itest-standalone as it is duplicated by org.apache.camel.management.CamelContextDisableJmxTest .
Test from tests/camel-typeconverterscan was moved to processor tests. Processor converter tests were then moved to a separate package to avoid scanning the whole org.apache.camel.processor for custom converter. tests/camel-typeconverterscan was deleted as no relevant tests remained there.

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20113


orpiske (approved):
This is looking better and better! Thanks for your contribution!

PR #12467, opened 2023-12-17
CAMEL-20247: Rework of dynamic router EIP component.

Steve973:
# Description
Reworked the Dynamic Router EIP Component to fix URI subscribing, and to make the component simpler by using the recipient list processor, and having the component focus on managing subscriptions instead of reinventing wheels.

There will be two follow-on pull requests:
1. The spring boot starter counterpart to this component
2. The spring boot example project updated to reflect these changes

# Target

# Tracking


components/camel-dynamic-router/pom.xml
```diff
@@ -40,8 +41,24 @@
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-core-processor</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>jakarta.validation</groupId>
+      <artifactId>jakarta.validation-api</artifactId>
+      <version>3.0.2</version>
```
davsclaus, line 47: Do not use hardcoded versions

components/camel-dynamic-router/pom.xml
```diff
@@ -40,8 +41,24 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.16.0</version>
```
davsclaus, line 52: Do not use hardcoded versions
And why is jackson needed ?

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
```diff
@@ -0,0 +1,247 @@
...
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+		http://camel.apache.org/schema/spring http://camel.apache.org/schema/spring/camel-spring.xsd
+		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd ">
+
+    <camelContext id="camelContext" xmlns="http://camel.apache.org/schema/spring">
+        <route>
+            <from uri="direct:subscribe"/>
+            <toD uri="dynamic-router-control:subscribe
```
davsclaus, line 101: There seems to not be any dynamic in this, so it can maybe be a static to

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
```diff
@@ -0,0 +1,247 @@
...
+Above, because the control URI is dynamic, and since a `ProducerTemplate` does not have a built-in way to send to a
+dynamic URI, we have to send subscription parameters from a `ProducerTemplate` in a different way.  The dynamic-aware
+endpoint uses headers "under the hood", because the URI params are converted to headers, so we can set the headers
+deliberately.
+
+.Example XML URI Subscription
+[source,xml]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
```
davsclaus, line 92: We are not recommending old Spring <beans> today. Can you just make the example with Camel <route>

components/camel-dynamic-router/pom.xml
Steve973: I probably do not need the validation annotation that I was using, so I'll remove it.  For the future, do you prefer the use of a property for the version, or something else?

components/camel-dynamic-router/pom.xml
Steve973: I like using a jackson object mapper to get a string representation of an object, but I see how that is not really necessary, so I'll get rid of it.

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
Steve973: Did you see the whole URI?
```
            <toD uri="dynamic-router-control:subscribe
                      ?subscribeChannel=myChannel
                      &amp;subscriptionId=mySubId
                      &amp;destinationUri=direct:myDestination
                      &amp;priority=5
                      &amp;predicate=true
                      &amp;expressionLanguage=simple"/>
```
That's definitely dynamic, so I need it there for the documentation to be correct.

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
Steve973: I will make sure to convert all of the examples that use spring xml to something else (perhaps Spring Java config).  I can also do that for the integration tests that use old spring xml.

davsclaus (approved):
Its a massive PR so I dont have time for detailed review.

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
davsclaus: No just remove <camelContext> and only show generic <route> then it works anywhere with XML

Steve973: It's removed.

components/camel-dynamic-router/pom.xml
Steve973: It's gone.

components/camel-dynamic-router/pom.xml
Steve973: I got rid of the dependencies that had hard coded version so these are gone, too.

components/camel-dynamic-router/src/main/docs/dynamic-router-control-component.adoc
Steve973: The dynamic aware part takes a look at the query parameters and removes them while adding them to headers and the query parameters that you see here are the ones that it uses when it's converting to the plain endpoint without query parameters.

Steve973:
@davsclaus the previous build failed because I didn't re-generate the sources after the version got bumped to 4.4.0-SNAPSHOT.  I think it should be good now.

Steve973:
@davsclaus 

Thank you for taking the time to look at this, even if it was too much to examine thoroughly enough.  I would welcome a more thorough review when you have the time for this, if possible.  I enjoy contributing to Camel, and I want to make sure that I am contributing in a way that is most helpful for Camel and the committers.  Please let me know if there is some way that we can go through this so that I can learn the Camel way of doing this a bit better.

PR #12474, opened 2023-12-18
CAMEL-20251: Add Camel K commands to Camel JBang

christophd:
- Adds new subcommands to Camel JBang that allow to manage Camel K integrations
- run command to create Integrations on Kubernetes
- list command to list all Integrations on the cluster namespace
- delete command to remove Integrations
- logs command to print log output of a running Integration

# Description

Users are able to manage Camel K integrations running on Kubernetes.
The implementation uses the fabric8 Kubernetes client and the Camel K Integration CRDs to manage the integrations on the cluster.

Users are able to call
```
camel k run route.groovy
```

Where `k` provides a set of new subcommands related to running Camel routes on Kubernetes in the form of Camel K integrations.

# Follow-up

Not included in this PR is documentation of the new commands. I wanted to get feedback on this first. Also I am experimenting on a Camel JBang plugin mechanism that would allow us to enable such sub-commands on demand (e.g. `camel plugin add camel-k`.

# Target

# Tracking

Relates to CAMEL-20251

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/k/KubeCommand.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/k/KubeCommand.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.dsl.jbang.core.commands.k;
+
+import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "k",
+                     description = "Manage Camel integrations on Kubernetes (use config --help to see sub commands)")
+public class KubeCommand extends KubeBaseCommand {
+
+    public static final String OPERATOR_ID_LABEL = "camel.apache.org/operator.id";
+    public static final String INTEGRATION_LABEL = "camel.apache.org/integration";
+    public static final String INTEGRATION_CONTAINER_NAME = "integration";
+
+    public KubeCommand(CamelJBangMain main) {
+        super(main);
+    }
+
+    @Override
+    public Integer doCall() throws Exception {
+        // defaults to list integrations deployed on Kubernetes
+        new CommandLine(new IntegrationGet(getMain())).execute();
+        return 0;
+    }
+}
--- a/docs/user-manual/modules/ROOT/nav.adoc
+++ b/docs/user-manual/modules/ROOT/nav.adoc
@@ -6,6 +6,7 @@
 ** xref:building.adoc[Building]
 ** xref:camel-console.adoc[Camel Developer Console]
 ** xref:camel-jbang.adoc[Camel JBang]
+*** xref:camel-jbang-k.adoc[Camel Integration with Kubernetes]
 ** xref:camel-maven-plugin.adoc[Camel Maven Plugin]
 ** xref:camel-component-maven-plugin.adoc[Camel Component Maven Plugin]
 ** xref:camel-report-maven-plugin.adoc[Camel Maven Report Plugin]
--- a/docs/user-manual/modules/ROOT/pages/camel-jbang-k.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-jbang-k.adoc
@@ -1,0 +1,147 @@
+= Camel Integration with Kubernetes
+
+Please make sure to meet these prerequisites for running Camel integrations on Kubernetes:
+
+* Connect to namespace on a Kubernetes cluster where you want to run the integration
+* Camel K operator must be installed on the Kubernetes cluster (either installed on the same namespace or as global operator for the whole cluster)
+
+Running Camel routes on Kubernetes is quite simple with Camel JBang.
+In fact, you can develop and test your Camel route locally with Camel JBang and then promote the same source to running it as an integration on Kubernetes.
+Simply run the integration using the `k` subcommand in Camel JBang.
+
+[source,bash]
+----
+camel k run route.yaml
+----
+
+The command runs the Camel integration on Kubernetes.
+More precisely it creates a Camel K Integration custom resource in the current namespace.
+The Camel K operator makes sure to create a proper runtime image and run the integration (usually as a Pod).
+
+The Camel K operator will automatically manage and configure this integration.
+In particular the operator takes care on exposing services, configuring health endpoints, providing metrics, updating image streams and much more.
+
+By default, the run command will not wait for the integration to in state running.
+You need to add `-w` or `--wait` option in order to wait for the integration to become ready.
+
+The `--logs` option makes the command also print the integration output once the integration Pod is running.
+
+The run command offers a lot more options that you may use to configure the Camel K integration.
+
+[width="100%",cols="1m,3",options="header",]
+|=======================================================================
+|Option |Description
+
+|--name
+|The integration name. Use this when the name should not get derived from the source file name.
+
+|--image
+|An image built externally (for instance via CI/CD). Enabling it will skip the integration build phase.
+
+|--kit, -k
+|The kit used to run the integration.
+
+|--profile
+|The trait profile to use for the deployment.
+
+|--service-account
+|The service account used to run this Integration.
+
+|--pod-template
+|The path of the YAML file containing a PodSpec template to be used for the integration pods.
+
+|--operator-id
+|Operator id selected to manage this integration. (default=camel-k)
+
+|--dependency, -d
+|Adds dependency that should be included, use "camel:" prefix for a Camel component, "mvn:org.my:app:1.0" for a Maven dependency.
+
+|--property, -p
+|Add a runtime property or properties file from a path, a config map or a secret (syntax: [my-key=my-value,file:/path/to/my-conf.properties,[configmap,secret]:name]).
+
+|--build-property
+|Add a build time property or properties file from a path, a config map or a secret  (syntax: [my-key=my-value,file:/path/to/my-conf.properties,[configmap,secret]:name]]).
+
+|--config
+|Add a runtime configuration from a ConfigMap or a Secret (syntax: [configmap,secret]:name[/key], where name represents the configmap/secret name and key optionally represents the configmap/secret key to be filtered).
+
+|--resource
+|Add a runtime resource from a Configmap or a Secret (syntax: [configmap,secret]:name[/key][@path], where name represents the configmap/secret name, key optionally represents the configmap/secret key to be filtered and path represents the destination path).
+
+|--open-api
+|Add an OpenAPI spec (syntax: [configmap,file]:name).
+
+|--env, -e
+|Set an environment variable in the integration container, for instance "-e MY_VAR=my-value".
+
+|--volume, -v
+|Mount a volume into the integration container, for instance "-v pvcname:/container/path".
+
+|--connect, -c
+|A Service that the integration should bind to, specified as [[apigroup/]version:]kind:[namespace/]name.
+
+|--source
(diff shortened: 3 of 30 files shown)


oscerd (approved):

claudio4j:
Related to the command structure, wdyt to enable camel-k with an option `--runtime=camel-k` instead to have camel-k as subcommand `camel k <command>` ? Then it would be like `camel run MyAppjava --runtime=camel-k`

There is a feature request, which I don't know if that has evolved.
[CAMEL-19041 - Run with runtime](https://issues.apache.org/jira/browse/CAMEL-19041)

I see the `camel` has several sub-commands which could also be used for camel-k, but at the same time not all sub-commands may not target camel-k.
Just food for thought.

christophd:
@claudio4j yeah I was also thinking about it that way but then also realized that not all commands fit for all runtimes. and also running Camel routes on Kubernetes will be independent of the runtime (Spring Boot, Quarkus, ...) so we would need something like `--runtime=camel-k,quarkus` to use both Quarkus as a runtime and at the same time Camel K to run it on Kubernetes.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/k/KubeCommand.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.camel.dsl.jbang.core.commands.k;
+
+import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "k",
+                     description = "Manage Camel integrations on Kubernetes (use config --help to see sub commands)")
+public class KubeCommand extends KubeBaseCommand {
+
+    public static final String OPERATOR_ID_LABEL = "camel.apache.org/operator.id";
```
squakez, line 26: I wonder if these (and any other constant) should be instead moved into the CRD dependency. Something we might ponder eventually.

squakez (approved):
LGTM. However I did not properly understand how it works the test. Not sure if it exists the structure for it, but it would be nice to see some real execution (at least the dry run) such as `camel k run xyz -o yaml` and verify the output integration is the expected one.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/k/KubeCommand.java
christophd: yes, makes sense to me

gansheer:
I did some local test it works well. I still have some questions:
* I first tried the command without any kubernetes up. Is it standard in with jbang camel to end up with a full stacktrace in these kind of error cases ?
* Is the general plan to follow with the existing design from `kamel` CLI 2nd day operations commands ?

squakez:
> 
>     * Is the general plan to follow with the existing design from `kamel` CLI 2nd day operations commands ?

I don't think we should. We did a lot of development in the past on the CLI but the reality is that Kubernetes/Openshift already offer a lot of tooling to manage those aspects.

christophd:
> Is the general plan to follow with the existing design from kamel CLI 2nd day operations commands ?

What `kamel` CLI 2nd day operations commands do you have in mind?

gansheer:
> > Is the general plan to follow with the existing design from kamel CLI 2nd day operations commands ?
> 
> What `kamel` CLI 2nd day operations commands do you have in mind?

I don't know if the following constitutes 2nd day operations but I was thinking about the sub-commands `kit`, `rebuild`, `bind` and `promote`.
Also if we are not following the original design, I would rename `get` as `list` if there is no filter on integration name, but I recognize that's personal preference :relaxed:.

christophd:
@squakez I have added some test that uses the complete command line `camel k run route.yaml`

@gansheer renamed `get` command to `list`

squakez:
> @squakez I have added some test that uses the complete command line `camel k run route.yaml`
> 
> @gansheer renamed `get` command to `list`

Great. I think the entire testing is going to benefit a lot from this example.

claudio4j:
> @claudio4j yeah I was also thinking about it that way but then also realized that not all commands fit for all runtimes. 

Yeah, sure let's see how this evolves.

davsclaus:
1) It would be great to have jbang with a plugin for different command sets, as we end up with a big pile of dependencies and many commands that will get us in trouble in the future. Good to hear you do some experiment with this. Its not only the CK stuff that we can separate. It would also be for example the generator command, sbom, jolokia, hawtio etc

2) clever way to use k as sub command to separate this clearly

3) the get command has been called get for all its life in camel-k, and we use same name in camel-jbang, so I think its not a good idea to rename it to list (old habbits die hard)

4) docs, the jbang docs is a very long page, it needs to be broken up into sub pages, so a good start is to have the CK in its own sub page. And then we can use that as starting point to break up the existing big page also.

5) can the tests run in isolation or do they need a running k8s cluster. Just want to sure that we can build everyhing nicely without causing test failures due to k8s stuff as core project dont have any of that.

6) mind that jbang is for developers and not sys admins, or ops guys. I think the kubectl, oc and kamel go cli would be needed for those guys - I cant see them want to install java jbang to run a shell command.

7) the dependency on the camel-k fabric8 client to have those CRDs - is there in the future a better way, as its not good that the core project has a dependency on a Camel K release. Can it be the fabric8-client release instead.

christophd:
@davsclaus thanks for your input I'll go through it today

christophd:
> 3. the get command has been called get for all its life in camel-k, and we use same name in camel-jbang, so I think its not a good idea to rename it to list (old habbits die hard)

I see that the `camel get` provides many more details. I guess the `list` command I was doing here in this PR is more like the `camel ps` command. Should I use `camel k ps` then? I personally don't like `ps` very much so I am ok to use `camel k get` but just wanted to get some feedback. @davsclaus @oscerd @squakez

squakez:
> > 3. the get command has been called get for all its life in camel-k, and we use same name in camel-jbang, so I think its not a good idea to rename it to list (old habbits die hard)
> 
> I see that the `camel get` provides many more details. I guess the `list` command I was doing here in this PR is more like the `camel ps` command. Should I use `camel k ps` then? I personally don't like `ps` very much so I am ok to use `camel k get` but just wanted to get some feedback. @davsclaus @oscerd @squakez

IMO neither. The user should be able to list and operate via `kubectl` or any other Kubernetes tool as he normally does. The `camel k run` should be the only facility that we offer in my opinion. We may even not require to push the Integration to the cluster at all if we want. The ultimate feature it should offer is to create an Integration custom resource with the parameters provided by the user (here we should try to give the majority of parameters provided in the `kamel run`). We may have something like:
```
$ camel k run test.yaml
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  annotations:
    camel.apache.org/operator.id: camel-k
  creationTimestamp: null
  name: test
spec:
  flows:
  - from:
      parameters:
        period: "1000"
      steps:
      - setBody:
          simple: Hello Camel from ${routeId}
      - log: ${body}
      uri: timer:yaml
  traits: {}
```
and in order to apply on the cluster we may have something like a piped command in conjunction with kubectl/oc:
```
$ camel k run test.yaml | kubectl apply -f -
integration.camel.apache.org/test created
```

oscerd:
> 1. It would be great to have jbang with a plugin for different command sets, as we end up with a big pile of dependencies and many commands that will get us in trouble in the future. Good to hear you do some experiment with this. Its not only the CK stuff that we can separate. It would also be for example the generator command, sbom, jolokia, hawtio etc
> 2. clever way to use k as sub command to separate this clearly
> 3. the get command has been called get for all its life in camel-k, and we use same name in camel-jbang, so I think its not a good idea to rename it to list (old habbits die hard)
> 4. docs, the jbang docs is a very long page, it needs to be broken up into sub pages, so a good start is to have the CK in its own sub page. And then we can use that as starting point to break up the existing big page also.
> 5. can the tests run in isolation or do they need a running k8s cluster. Just want to sure that we can build everyhing nicely without causing test failures due to k8s stuff as core project dont have any of that.
> 6. mind that jbang is for developers and not sys admins, or ops guys. I think the kubectl, oc and kamel go cli would be needed for those guys - I cant see them want to install java jbang to run a shell command.
> 7. the dependency on the camel-k fabric8 client to have those CRDs - is there in the future a better way, as its not good that the core project has a dependency on a Camel K release. Can it be the fabric8-client release instead.

For the last question, the point 7, they moved back to camel-k the crd responsibility, so now we are releasing the CRDs as part of our release projects. Fabric8 Kubernetes Client started to have too much stuff to maintain.

squakez:
Also `kamel bind` and `kamel promote` may be useful commands with the same concept. You may just take the output of what's actually provided in the `-o yaml` of both.

squakez:
> For the last question, the point 7, they moved back to camel-k the crd responsibility, so now we are releasing the CRDs as part of our release projects. Fabric8 Kubernetes Client started to have too much stuff to maintain.

On top of that, consider that we do not necessarily depends on the Camel K release cycle. The CRDs definition should not change, or at least has to be backward compatible. So, if we come to add some new parameter, it could be available in future versions of Camel JBang.

christophd:
@squakez these are good thoughts, thank you.

I like the suggested way to use the pipe to `kubectl`. IMO providing the `camel k get` command in Camel JBang is not a huge burden for us though and this way users may decide on their own which path to take. 

The user is able to use the pipe to `kubectl` or just use the same `camel` CLI with the advantage to not have to switch tooling for basic operations.

The `camel k logs` command for instance also adds some simplification as the user does not have to select the Pod but only the integration name.

My take on this would be to keep some basic operations for managing integrations on Kubernetes in Camel JBang

christophd:
> 1. It would be great to have jbang with a plugin for different command sets, as we end up with a big pile of dependencies and many commands that will get us in trouble in the future. Good to hear you do some experiment with this. Its not only the CK stuff that we can separate. It would also be for example the generator command, sbom, jolokia, hawtio etc

Working and experimenting on it but it will take some more time to get results

> 3. the get command has been called get for all its life in camel-k, and we use same name in camel-jbang, so I think its not a good idea to rename it to list (old habbits die hard)

I'll go back to `camel k get` and push to this PR

> 4. docs, the jbang docs is a very long page, it needs to be broken up into sub pages, so a good start is to have the CK in its own sub page. And then we can use that as starting point to break up the existing big page also.

Working on it and will push to this PR

> 5. can the tests run in isolation or do they need a running k8s cluster. Just want to sure that we can build everyhing nicely without causing test failures due to k8s stuff as core project dont have any of that.

Yes, the Kubernetes mockserver used in the tests is just an arbitrary Http server that emulates the Kubernetes API. No need to have a Kubernetes cluster

> 6. mind that jbang is for developers and not sys admins, or ops guys. I think the kubectl, oc and kamel go cli would be needed for those guys - I cant see them want to install java jbang to run a shell command.

Right, and we should always provide the opportunity to use these standard tooling for instance by doing the pipe to `kubectl` as Pasquale suggested. At the same time I think Camel JBang is a great way to start with Camel and it should also be a great way to start with integrations running on Kubernetes.

davsclaus:
Thanks @christophd good answers.

christophd:
I am done with using `camel k get` again and starting to modularize the Camel JBang docs. Regarding the docs I hope I did it right as I was not able to build the pages locally as described in https://camel.apache.org/manual/improving-the-documentation.html#_local_build_instructions

I get these errors when running `./local-build.sh quick`

```
FATAL (antora): require(...) is not a function
    Cause: TypeError
        at Function.create (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/index.js:316:45)
        at Object.default_1 [as default] (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/cli/command.start.js:43:10)
        at handleCli (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/bin.js:145:25)
        at handleNoCommand (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/bin.js:135:12)
        at runFromCli (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/bin.js:71:16)
        at Object.<anonymous> (Projects/Camel/camel-website/.yarn/cache/browser-sync-npm-2.27.7-b8e58f9a91-7f51177473.zip/node_modules/browser-sync/dist/bin.js:38:5)
        at loadCJSModule (node:internal/modules/esm/translators:270:3)
        at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:305:7)
        at ModuleJob.runSync (node:internal/modules/esm/module_job:209:17)
        at require (node:internal/modules/esm/translators:255:9)
```

My env settings
```
yarn -v
3.2.3
node -v
v21.4.0
```

davsclaus:
that local doc stuff is some black magic. You dont need to use that, you can just create a new adoc file along camel-jbang, and link to it from that page. And then mvn clean install in the docs folder does sanity checking for xref and broken links.

And then we do have CI server doing the full build in camel-website

PR #12475, opened 2023-12-18
CAMEL-20225: use the clock API when possible for exchange duration calculations

orpiske:
This brings a few additional improvements:
1. Use a monotonic clock for calculating duration
2. Consolidate more code into the `elapsed` method 
3. Deprecate ad-hoc duration calculations


orpiske:
The PR is ready and very likely OK - I am running a second time on my CI and I'm waiting the results.

PR #12530, opened 2023-12-21
CAMEL-20019: demo session handler implementation

jono-morris:
# Description

Draft code demonstrating a Vertx server handler implementation that manages server-side sessions and session cookie handling.  

# Target

# Tracking

### DIFF
--- a/components/camel-platform-http-vertx/pom.xml
+++ b/components/camel-platform-http-vertx/pom.xml
@@ -85,11 +85,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.camel</groupId>
-            <artifactId>camel-vertx-http</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>io.vertx</groupId>
             <artifactId>vertx-auth-properties</artifactId>
             <version>${vertx-version}</version>
--- a/components/camel-platform-http-vertx/pom.xml
+++ b/components/camel-platform-http-vertx/pom.xml
@@ -85,6 +85,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-vertx-http</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.vertx</groupId>
             <artifactId>vertx-auth-properties</artifactId>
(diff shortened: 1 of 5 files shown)


components/camel-platform-http-vertx/pom.xml
```diff
@@ -84,6 +84,11 @@
             <artifactId>camel-log</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-vertx-http</artifactId>
+            <scope>test</scope>
+        </dependency>
```
orpiske, line 91: I'm a bit concerned that this may cause some weird cyclic dependency problems (or at least make the whole dependency chain more complex).

orpiske (commented):
Probably this needs further review

components/camel-platform-http-vertx/pom.xml
davsclaus: Yeah potentially - though it may be that camel-platform-http-vertx is built before all the vertx components where this one is from. But agree we had been bitten in the past with build cycles going full circle and breaking the build

davsclaus (approved):
Yeah a good start, this is vertx specific which of course works great as its vertx based. 
Adding that camel cookie stuff is something you can look at after getting this vertx stuff implemented. 
Also remember to add something into the docs, see src/main/docs folder

components/camel-platform-http-vertx/pom.xml
jono-morris: Updated the test to use `camel-http` instead.

orpiske: Thanks, but that still may create a very complex dependency chain as you have components within `components` with dependencies among themselves. 

The best way, I think, would be to create a `test-infra` module that is completely independent.

jono-morris: Thanks for that, let me look into it.

Merry Christmas!

jono-morris: Hey @orpiske, by using `camel-http` instead of `camel-vertx-http` it now means that no new deps have been introduced by this change. Would it be OK it we tackle the test dependency management improvements that you've proposed in a separate ticket/PR?

orpiske: Yes, that's OK to me. Thanks!

orpiske (approved):

PR #12580, opened 2023-12-22
﻿CAMEL-20276 fixing jms shared topic tests

kulagaIA:
# Description
 it seems that sometimes publication happens before artemis finishes creating consumers

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20276

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsTopicDurableSharedTest.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsTopicDurableSharedTest.java
@@ -18,18 +18,14 @@ package org.apache.camel.component.jms;
 
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Tags;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
 @Tags({ @Tag("not-parallel") })
-@DisabledIfSystemProperty(named = "ci.env.name", matches = "github.com", disabledReason = "Flaky on GitHub Actions")
-@Disabled("Disabled due to CAMEL-20276")
 public class JmsTopicDurableSharedTest extends AbstractPersistentJMSTest {
 
-    private static final String TEST_DESTINATION_NAME = "activemq:topic:in.only.topic.consumer.test";
+    private static final String TEST_DESTINATION_NAME = "activemq:topic:in.only.topic.shared.durable.test";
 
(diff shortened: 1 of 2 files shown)


kulagaIA:
One thing that probably could be done is polling artemis for the state of consumers, instead of just waiting. If it is needed i can dig into it

orpiske (approved):
LGTM, thanks!!!

components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsTopicDurableSharedTest.java
```diff
@@ -18,18 +18,19 @@
...
 
 @Tags({ @Tag("not-parallel") })
-@DisabledIfSystemProperty(named = "ci.env.name", matches = "github.com", disabledReason = "Flaky on GitHub Actions")
-@Disabled("Disabled due to CAMEL-20276")
 public class JmsTopicDurableSharedTest extends AbstractPersistentJMSTest {
 
-    private static final String TEST_DESTINATION_NAME = "activemq:topic:in.only.topic.consumer.test";
+    private static final String TEST_DESTINATION_NAME = "activemq:topic:in.only.topic.shared.durable.test";
+
+    private CountDownLatch artemisConsumersWereCreated = new CountDownLatch(1);
```
orpiske, line 33: Where are you counting down the latch?

components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsTopicDurableSharedTest.java
```diff
@@ -42,6 +43,7 @@ void testDurableSharedTopic() throws Exception {
         mock2.expectedMessageCount(1);
         mock2.expectedBodiesReceived("Hello World");
 
+        artemisConsumersWereCreated.await(1000, TimeUnit.MILLISECONDS);
```
orpiske, line 46: This latch is noop, as it is just timing out and it's not an adequate use of a latch.

kulagaIA: Yeah its the same as Thread.sleep if we dont poll artemis for the state of the consumers. Ok, I'll revert

orpiske (changes requested):
That's better but still not fully adequate. Instead of a blank `Thread.sleep(1000)` you actually want to wait for the context itself to be up for 1 second. 

I believe something like this should do the trick:

```
@BeforeEach
    void waitForConnections() {
        Awaitility.await().until(() -> context.getUptimeMillis() > 1000);
    }
```

orpiske (approved):
LGTM, thanks.

kulagaIA:
> LGTM, thanks.

Thanks for pointing out that i need to wait for the context to be up, instead of just noop!

PR #12583, opened 2023-12-22
CAMEL-20270: Introduce JBang plugins

christophd:
# Description

- Introduce plugin mechanism to modularize Camel JBang
- Add new plugin sub-commands (add, get, delete) to manage plugins
- Plugin configuration is stored to user home directory
- Load plugin commands on demand
- Create 1st plugin for Camel K and move commands into separate Maven module camel-jbang-plugin-k

PR introduces plugins for Camel JBang where each plugin is able to add sub-commands to the CLI. The advantage of having a plugin in a separate Maven module is to modularize the Camel JBang CLI and avoid stuffing all dependencies and functionality into `camel-jbang-core`.

The PR adds a 1st plugin module `camel-jbang-plugin-k` that manages integrations on Kubernetes (the commands have been added to Camel JBang recently).

Users need to enable/add plugins with the new command `camel plugin add`. The plugin usually gets a name, a command and a dependency (Maven GAV) that is added on demand when the user calls a plugin command. Once the plugin is added for the environment all commands of the plugin get listed in the CLI and users can just call the sub-commands like every other sub-command.

The plugin config holding all enabled plugins for the environment is stored in the `user.home` directory (`.camel-jbang-plugins.json`).

Users are able to list enabled plugins with `camel plugin get` (`--all` option to list all supported plugins in the Camel project).

```
camel plugin get --all

Supported plugins:

 NAME     COMMAND  DEPENDENCY                             DESCRIPTION                             
 camel-k  k        org.apache.camel:camel-jbang-plugin-k  Manage Camel integrations on Kubernetes
```

`camel plugin delete` will remove a plugin from the current environment so that the plugin commands are no longer listed in the CLI and users will not be able to call the plugin`s commands any more.

# Follow-up actions

- Update the documentation to describe new JBang plugin mechanism
- Extract more plugins from current Camel JBang sub-commands (e.g. sbom, hawtio, jolokia)

# Target

# Tracking

Relates to CAMEL-20270


davsclaus (approved):

davsclaus:
I didn't see whether there was some updates to the documentation - that is something we would need down the road when we have more plugins and users need to install them.

Also it may be possible to install multiple at once

camel plugin add foo bar cheese

or maybe you want all out of the box

camel plugin add --all

PR #12636, opened 2024-01-02
Increase max number of opened pull request for depdencies update from…

apupier:
… Dependabot

# Description

There is still potentially a long list of dependencies to update. It seems that the current default limit of 5 doesn't allow to catch up fast enough.
I'm proposing to increase to 20 to try to have a bit more opened at the same time without spamming too much and have too many "pages" of Pull Requests.
What do you think?

for reference https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

# Target

# Tracking

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,4 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 20


oscerd (approved):
Good work

PR #12647, opened 2024-01-04
Substituted string concatenation as argument of StringBuilder.append()

dk2k:
…  with chained append() calls - every concatenation is transformed by compiler to temp StringBuilder itself (additional garbage). Implemented templated var in instanceof check.

This issues are highlighted in static analysis check


orpiske (changes requested):
Just one nitpick comment. Please, can you adjust your commit message to be in the format: 
```
<short> 

<long description> 
```

Something like: 
```
(chores) Substituted string concatenations

Substituted string concatenations as argument of StringBuilder.append with chained append() calls - every concatenation is transformed to temp StringBuilder itself (garbage). Implemented templated var in instanceof check.
```

It makes a lot easier when reading with the `--oneline`. 

Thanks!

davsclaus:
Can you update the PR

dk2k:
I updated the PR

davsclaus (approved):

PR #12651, opened 2024-01-04
Adding jenkinsfile for ppc64le

sanskruti1098:


oscerd (commented):
As already said camel 4 is JDK 17 at least.

PR #12669, opened 2024-01-04
JmsDeliveryDelayTest fix flackiness

kulagaIA:
# Description
testInOnlyWithDelay was flacky because sometimes message is not received from artemis before .assertIsSatisfied(context) is called.
testInOutWithDelay was missing assertions of mock endpoint condition and of response contents.

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsDeliveryDelayTest.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/JmsDeliveryDelayTest.java
@@ -16,14 +16,20 @@
  */
 package org.apache.camel.component.jms;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.util.StopWatch;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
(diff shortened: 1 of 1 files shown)


kulagaIA:
> /component-test camel-jms
> 
> **Result** ❌ The tests failed please [check the logs](https://github.com/apache/camel/actions/runs/7414143507)

Apparently the problem is in how deliveryDelay is handled in jms component or jms implementation. I'll look into it.
If you have any ideas what else can cause this test to fail, please share :)

orpiske (changes requested):
A few problems:

1. In general, the StopWatch (or the standard Duration from Java) should be used for computing durations 

Looking at the comments on this test ... I think the right approach is to disable it on github and leave the logic as is. 

So, my .2 cents: just disable the test on github (search for `@DisabledIfSystemProperty` on the code base to see how we do it).

kulagaIA:
> In general, the `StopWatch` (or the standard `Duration` from Java) should be used for computing duration.
> 
> Looking at the comments on this test ... I think the right approach is to disable it on github and leave the logic as is.
> 
> So, my .2 cents: just disable the test on Github (search for `@DisabledIfSystemProperty` on the code base to see how we do it).

It seems that using StopWatch and measuring elapsed time after .assertIsSatisfied(context) was called, instead of calcualting elapsed time using exchange clock, helped to make this test at least less flacky. I think it can be merged, and if I see it failing again, I'll disable.

orpiske (approved):
LGTM. Let's try this approach.

PR #12708, opened 2024-01-09
CAMEL-20303 - Camel-Sql: Add ObjectInputFilter String pattern paramet…

oscerd:
…er in JdbcAggregationRepository to be used in unmarshall operations

# Description

# Target

# Tracking


ChayimZhang:
If I merge this commit into camel v3.14.10 to fix the same issue, what do I need to be aware of, are there any incompatibilities? @oscerd

oscerd:
The only thing you need to consider is just that this has never been compiled with JDK 8. For 3.x it has been compiled with JDK 11 and for 4.x with JDK 17. So it could be possible it doesn't work with JDK 8.

PR #12777, opened 2024-01-12
CAMEL-20218: Use a single Kafka consumer with assign() call

tashoyan:
Use a single Kafka consumer with `assign()` call to consume all cached entries from the beginning of the Kafka topic.

# Description

The existing implementation of `KafkaIdempotentRepository` may fail to restore the cached entries on application restart. This implementation is based on `KafkaConsumer.subscribe()` method and KafkaConsumer rebalancing. It navigates the Kafka consumer to the beginning of the Kafka topic using `ConsumerRebalanceListener.onPartitionsAssigned()` callback. However, the rebalancing occurs asynchronously. Therefore there is a chance, that first few calls to `KafkaConsumer.poll()` will be done before the consumer gets navigated to the beginning of the KafkaTopic. Hence, KafkaConsumer gets zero records from the Kafka topic and `KafkaIdempotentRepository` gets initialized with empty cache. With empty cache, `KafkaIdempotentRepository` re-consumes the entire input without any deduplication.

The proposed implementation does not rely on Kafka consumer groups at all. It has the only KafkaConsumer and explicitly assigns topic partitions using `KafkaConsumer.assign()` method. Actually `KafkaIdempotentRepository` needs only one KafkaConsumer - no need to balance the load across many consumers.

The proposed implementation does not need Kafka consumer groupId. The parameter `groupId` is still accepted by the public API, but ignored. All constructors and methods dealing with groupId are marked as deprecated.

There is no automated test for this change, because the problem can be reproduced only when restarting an application, and the Kafka topic should already have some cached entries. I have tested this fix manually on a real environment. The existing automated tests run successfully.

# Target

main

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20218

- [ v] I checked that each commit in the pull request has a meaningful subject line and body.

- [ v] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/processor/idempotent/kafka/KafkaIdempotentRepository.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/processor/idempotent/kafka/KafkaIdempotentRepository.java
@@ -18,15 +18,15 @@ package org.apache.camel.processor.idempotent.kafka;
 
 import java.time.Duration;
 import java.util.Collection;
-import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 
 import org.apache.camel.CamelContext;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

orpiske (approved):

tashoyan:
Same modification, but for the 3.22.x branch: https://github.com/apache/camel/pull/12797

valdar:
Thanks for your contribution @tashoyan

PR #12821, opened 2024-01-17
fix binary operator match problem

EvanMi:
# Description

The binary operator is recognized in some places that should not be recognized, just like the “in” keyword in test case, which is recognized from “minValue”。

# Target

# Tracking


davsclaus (changes requested):
Can you avoid adding new JAR for testing, and also the big test class avoid implementing an empty dummy camel context extension.

EvanMi:
> Can you avoid adding new JAR for testing, and also the big test class avoid implementing an empty dummy camel context extension.

Any advice ? there has no test in this modual, so i import juit. And to avoid import other modual, i create the dummy camel context.  So, if move the testcase to other modual to avoid this? any suggested modual?

davsclaus:
tests are in core/camel-core

davsclaus:
This PR causes many new test failures in camel-core

EvanMi:
> This PR causes many new test failures in camel-core

Okay, I'll test it locally

EvanMi:
> This PR causes many new test failures in camel-core

All test cases in core have passed

davsclaus (approved):

PR #12848, opened 2024-01-19
CAMEL-20253: camel-jbang - Add support for jolokia 2.x

kulagaIA:
# Description
Migration guide https://jolokia.org/migration.html does not specify api change, so org.apache.camel.dsl.jbang.core.commands.process.Jolokia was left as it is.
JolokiaHttpRequestHandlerSupport is expected to be made more pluggable in the future.

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20253


davsclaus:
Can you remove the `@DevConsole` as its not needed

kulagaIA:
> Can you remove the `@DevConsole` as its not needed

Indeed it is not needed, thanks, removed it

PR #12857, opened 2024-01-22
CAMEL-20336: Add a WebAssembly component and language

lburgazzoli:
# Description

# Target

# Tracking


lburgazzoli:
> LGTM, I have just a question:
> 
> Is wee_alloc changeable?
> 

not needed anymore, will remove it

davsclaus:
It may be that the docs for language is a mostly a copy of the component doc. I think I didnt see an example how to use it as a language. Also what is the benefit of a language vs a component. Can you have very little wasm function scripts that you can inlined in a camel language, or would you always have these in external files. 

If the component does not do any kind of remote connection (I assume it calls always a local process), then you can set remote = false in the @UriEndpoint.

davsclaus (approved):

lburgazzoli:
> It may be that the docs for language is a mostly a copy of the component doc. I think I didnt see an example how to use it as a language. Also what is the benefit of a language vs a component. 

yeah at this very early stage, a language and a component are pretty much similar, the only difference is that at this stage the component acts on the whole exchange, so it can change headers & what not, whereas the language can only change the payload or be used as a predicate. 

I expect things to be improved over time, also depending on the maturity of the runtime.

About the documents, do we have a place where we can put some common docs ?

> Can you have very little wasm function scripts that you can inlined in a camel language, or would you always have these in external files.

It will always be an external resource since what the wasm runtime runs is a Wasm bitecode.

lburgazzoli:
> > LGTM, I have just a question:
> > Is wee_alloc changeable?
> 
> not needed anymore, will remove it

done

components/camel-wasm/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    </parent>
+
+    <artifactId>camel-wasm</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Camel :: Wasm</name>
+    <description>Camel Wasm component</description>
+
+    <properties>
+        <camel.surefire.fork.additional-vmargs>-Dwasm.resources.path=${project.basedir}/src/test/resources</camel.surefire.fork.additional-vmargs>
```
orpiske, line 36: Is this component portable? I think we should probably include the skips for `ppc64le` and `s390x`. 

```
<skipITs.ppc64le>true</skipITs.ppc64le>
<skipITs.s390x>true</skipITs.s390x>
```

I can give it a try later on these platforms.

orpiske: Trying it right now ... I'll let you know in about 45 minutes 😇

components/camel-wasm/src/test/java/org/apache/camel/language/wasm/WasmLanguageTest.java
```diff
@@ -0,0 +1,98 @@
...
+            Exchange out = pt.to("direct:in")
+                    .withHeader("foo", "bar")
+                    .withBody("hello")
+                    .request(Exchange.class);
+
+            assertThat(out.getException())
+                    .isNotNull()
+                    .hasCauseInstanceOf(RuntimeException.class)
+                    .extracting(Throwable::getCause, as(InstanceOfAssertFactories.THROWABLE))
+                    .hasMessage("this is an error");
```
orpiske, line 95: One thing I noticed: we need to add a `log4j2.properties` file to the `test/resources` directory. This is causing the test to leak to stdout. 

Something like this should do the trick:

```
appender.out.type = File
appender.out.name = out
appender.out.fileName = target/camel-wasm-test.log
appender.out.layout.type = PatternLayout
appender.out.layout.pattern = %d [%-15.15t] %-5p %-30.30c{1} - %m%n
appender.stdout.type = Console
appender.stdout.name = stdout
appender.stdout.layout.type = PatternLayout
appender.stdout.layout.pattern = %d [%-15.15t] %-5p %-30.30c{1} - %m%n

rootLogger.level = WARN
rootLogger.appenderRef.out.ref = out

logger.camel.name=org.apache.camel
logger.camel.level=INFO

logger.camelWasm.name=org.apache.camel.component.wasm
logger.camelWasm.level=DEBUG
```

lburgazzoli: done

orpiske (approved):
✅ s390x -> OK
✅  ppc64le -> OK

Nice one! Thanks @lburgazzoli

lburgazzoli:
> @lburgazzoli can your resolve conflicts so we could merge? Thanks!

done

PR #12879, opened 2024-01-23
CAMEL-19241: implement support for auto-commits with batching in camel-kafka

orpiske:


oscerd (approved):

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/batching/KafkaRecordBatchingProcessor.java
```diff
@@ -43,6 +44,28 @@ final class KafkaRecordBatchingProcessor extends KafkaRecordProcessor {
...
+                return;
+            }
+
+            LOG.debug("Calling commit on {} exchanges using {}", exchanges.size(), commitManager.getClass().getSimpleName());
+            commitManager.commit();
+        }
+
+        @Override
+        public void onFailure(Exchange exchange) {
+            LOG.warn("Skipping auto-commit on the batch because processing the exchanged has failed");
```
davsclaus, line 65: This may be noisy if you have a recurring error and kafka batching will keep poll the same messages as offset is not moved forward.

orpiske: Indeed! 

However, it will only happen if the user is not handling the error correctly (as described on the documentation). 

We could add a flag, so that it logs only once ... but I also feel that it may be hard for the users to notice the problem. Maybe we could do that + also increase the log level to error, so that the users can find with more easily with `grep` or their observability tools. 

Do you have any suggestions how we could this more elegantly? (Or, if we have solved this problem in some other component, just let me know and I'll do the same).

orpiske: > The other components would use consumer.getExceptionHandler().handleException(...) which are usually if there was a caused unhandled exception, and the default implementation will just log a WARN.

Understood. I think I got your point. 

> Is kafka not a lot faster, or how "slow is that rate" ?

In theory the rate (of the message) should be something similar to `(producer rate / batch size) - consumption delay`. IOW, the larger the batch size, the lower the incidence.

orpiske: BTW, based on your suggestion ... I think I should also adjust the code below, where there was the error handling used in streaming mode.

PR #12943, opened 2024-01-28
CAMEL-20237: camel-platform-http-main - Make it possible to configure http auth

kulagaIA:
# Description
Authentication is configured similar to session management in camel-platform-http-vertx.

In vert.x order of adding AuthenticationHandler matters (see https://groups.google.com/g/vertx/c/6lSKBWzK2pE).
Handlers for routes farther in the url path should be added first, that's why we are sorting AuthenticationConfigEntries by path length.

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20237


kulagaIA:
/component-test camel-platform-http-vertx camel-platform-http-main

kulagaIA:
/component-test camel-platform-http-vertx camel-platform-http-main

oscerd (approved):

davsclaus (approved):

davsclaus:
Can you add a bit of docs into the doc file inside src/main/docs folder

components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/auth/AuthenticationConfig.java
```diff
@@ -0,0 +1,104 @@
...
+import io.vertx.ext.auth.authentication.AuthenticationProvider;
+import io.vertx.ext.auth.properties.PropertyFileAuthentication;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.handler.AuthenticationHandler;
+import io.vertx.ext.web.handler.BasicAuthHandler;
+
+public class AuthenticationConfig {
+    public static final String DEFAULT_VERTX_PROPERTIES_FILE = "camel-platform-http-vertx-auth.properties";
+    private boolean authenticationEnabled;
+    private List<AuthenticationConfigEntry> entries;
```
orpiske, line 33: Minor nitpick, but ... couldn't these be adjusted to be made final? I think it would be cleaner.

orpiske (commented):
Overall it LGTM, but I left a comment about something I noticed.

components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/auth/AuthenticationConfig.java
kulagaIA: Big thanks for reading through my code! :)

Agree, but I think it will be inconvenient to have a final authentication field, because sometimes we want to create AuthenticationConfig first, and then populate it later. For example, this is how it is done in MainHttpServer.

orpiske: Understood, thanks for the explanation!

orpiske (approved):

kulagaIA:
> Can you add a bit of docs into the doc file inside src/main/docs folder

added some description of how default auth works and how to set up custom auth

davsclaus (approved):

PR #12949, opened 2024-01-29
CAMEL-20372: Change max.poll.interval.ms type from Long to Integer

aniketjadhav840:
# Description

Changed max.poll.interval.ms type in KafkaConfiguration.java from Long to Integer to be inline with Kafka config types.
Also added auto-generated changes.

# Target

- [x] I checked that the commit is targeting the correct branch 3.22.x

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20372


orpiske (changes requested):
Folks, just marking as request for changes so I have time to test it manually.

orpiske (approved):
LGTM, thanks @aniketjadhav840

PR #12953, opened 2024-01-30
CAMEL-20202: Read Azure BinaryData as stream

jono-morris:
# Description

Prefer the toStream method over the toString method on the BinaryData

# Target

# Tracking

### DIFF
--- a/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/QueueConsumer.java
+++ b/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/QueueConsumer.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.component.azure.storage.queue;
 
+import java.io.InputStream;
 import java.time.Duration;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus (changes requested):
you should not create uow and release uow when creating the exchange. Also this code seems to read the entire stream into memory, which is not much improved over the old code.

oscerd:
Maybe look at AWS SQS

oscerd (approved):

jono-morris:
Thanks for that. Let me investigate further.

jono-morris:
Updated to pass the InputStream from the Azure BinaryData object to the exchange body.  Also added a `SynchronizationAdapter` to close the stream when the exchange is complete similar to what is done in `AWS2S3Consumer`.  That method also has an `autocloseBody` option, would it make sense to also include that option here?

davsclaus (approved):

davsclaus:
The `autocloseBody` is related to `includeBody` option in AWS. I dont think we have those in Azure. And would that make sense. If you have includeBody=false, then you just get the metadata about the file. But nevermind that would be another JIRA and another PR.

PR #2887, opened 2019-04-26
Fixed invalid markdown links to ActiveMQ website

orpiske:
This PR fixes minor documentation problems I found when working on a separate PR.

PR #12966, opened 2024-01-31
[Camel 20377] *IT tests should be run with failsafe instead of surefire

cziesman:
# Description

There are many tests that use TestContainers and Docker in *IT.java classes that are currently being run by the `surefire` plugin. These tests should be run by `failsafe`, as they meet the definition of integration tests, not unit tests, and should not cause the whole build to fail if any of them fail.

Use the command `mvn verify -Pintegration-test -Dquickly=true` to run the integration tests.

Using any other Maven commands with the `default` or `full` profiles that invoke the test phase should only run unit tests. 

Existing profiles that have their own `failsafe` configuration for integration tests should not be affected.

# Target

# Tracking

- [ X] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes


parent/pom.xml
```diff
@@ -260,7 +260,7 @@
         <jedis-client-version>5.1.0</jedis-client-version>
         <jetcd-version>0.7.7</jetcd-version>
         <jetty-version>12.0.5</jetty-version>
-        <jetty-plugin-version>${jetty-version}</jetty-plugin-version>
+        <jetty-maven-plugin-version>11.0.19</jetty-maven-plugin-version>
```
davsclaus, line 263: Is there not a v12 of the jetty maven plugin ?

davsclaus (approved):

orpiske (approved):
Those tests are, indeed, in need of cleanup, fixes and de-duplication. Many of those don't follow the current patterns and standards we use elsewhere. Some have been moved/de-duplicated as part of [CAMEL-20113](https://issues.apache.org/jira/browse/CAMEL-20113), but there's still a lot TODO in this part of the code. 

Thanks for your contribution. Anything we can do to clean these up is a great thing.

orpiske:
Oh and just one minor nitpick stuff. 

If you write your commit messages like this `CAMEL-20377: <message>` the PR request gets automatically associated with the ticket, which makes it a lot easier for us to track the changes and what is going on.

oscerd (approved):

parent/pom.xml
cziesman: No, the Maven plugin version does not track with the Jetty version. Highest plugin version is 11.0.20, released 31 Jan, 2024.

davsclaus:
There is a merge conflict I wonder if you can fix that

davsclaus (approved):

orpiske:
Hey @davsclaus and @oscerd , please hold on just a bit on this one. There is something else I need to check.

davsclaus:
ok

orpiske (changes requested):
Just to temporarily check something I noticed.

pom.xml
```diff
@@ -915,5 +925,37 @@
...
+                                <include>**/*IntegrationTest.java</include>
+                            </includes>
+                            <excludes>
+                                <exclude>**/*Test.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
```
orpiske, line 959: I am truly sorry for having missed this, but I would us to avoid piling up the build with profiles that are very rarely used (by ourselves or by our CI or by the automation that we have). 

We (myself included) spent a lot of time cleaning up a several profiles that existed on Camel 3 but we never used and complicated the build.

For profiles, the less and simpler, the better.  

@davsclaus @oscerd wdyt?

pom.xml
```diff
@@ -409,13 +410,22 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maves-surefire-plugin</artifactId>
+                    <artifactId>maven-surefire-plugin</artifactId>
```
orpiske, line 413: This is OK.

pom.xml
```diff
@@ -409,13 +410,22 @@
...
                     <version>${maven-surefire-plugin-version}</version>
+                    <configuration>
+                        <includes>
+                            <include>**/*Test.java</include>
+                        </includes>
+                        <excludes>
+                            <exclude>**/*IT.java</exclude>
+                            <exclude>**/*IntegrationTest.java</exclude>
+                        </excludes>
+                    </configuration>
```
orpiske, line 423: We don't really need this. Surefire [already does it by default](https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html).

pom.xml
```diff
@@ -409,13 +410,22 @@
...
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
-                    <version>${maven-surefire-plugin-version}</version>
+                    <version>${maven-failsafe-plugin-version}</version>
```
orpiske, line 428: This is OK.

pom.xml
oscerd: I think it's better to avoid more profiles, we have already enough

orpiske (changes requested):
I am very sorry my first review missed several things, which I only realized now. I made a new review pointing the details that need to be fixed.

pom.xml
orpiske: On more note is that any configuration to the surefire configuration needs to be done on the parent's [pom.xml file](https://github.com/apache/camel/blob/main/parent/pom.xml). If needed, it's better to keep it in a single place.

cziesman: Surefire doesn't automatically exclude the pattern `*IntegrationTest.java`. I added all of the combinations just for completeness.

pom.xml
cziesman: The versions are the same today, but that may not always be the case. Or, there may be a bug in one of the plugins that requires reverting to a previous version. Finally, just for clarity, the version property name should match the artifact ID.

orpiske: Yeah, I think this is OK.

pom.xml
cziesman: Has anybody else tried to run integration tests lately? When I run `mvn failsafe:verify` against `main`, I get dozens of these errors:

```
[WARNING] The POM for org.apache.maven.plugins:maves-surefire-plugin:jar:3.2.5 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.apache.maven.plugins:maves-surefire-plugin:3.2.5: Plugin org.apache.maven.plugins:maves-surefire-plugin:3.2.5 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.apache.maven.plugins:maves-surefire-plugin:jar:3.2.5 (absent): org.apache.maven.plugins:maves-surefire-plugin:jar:3.2.5 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced
```

It turns out that this line in `pom.xml`

                    `<artifactId>maves-surefire-plugin</artifactId>`

should be this:

                    `<artifactId>maven-surefire-plugin</artifactId>`

When I fix that, no integration tests are executed. Maybe there is a magic combination of properties and profiles that enables integration tests to run, but I couldn't figure it out.

You make a very good point about reducing complexity, but I think maybe there needs to be a separate effort to refactor the POM structure. Things seem to have gotten so complicated that tests can't even run reliably.  In the meantime, after days of trying, the profile that I created is the only way I could find to run integration tests.

orpiske: You should use: `mvn verify` to run the tests (unit + integration).

orpiske: My (and I guess, the same for other committers) is just: 

```
cd path/to/project 
# Build with Maven Daemon, for quicker build
mvnd -Dquickly clean install
# Go to whatever you are working on (i.e.; camel-kafka)
cd components/camel-kafka 
# Run the tests
mvn verify 
```

Done.

pom.xml
orpiske: This change, if necessary, should be done on the parent pom.

davsclaus:
What is the status/consensus on this PR? 

We will cut 4.4 LTS start of next week so we should get this done before, or move to 4.5

orpiske:
> What is the status/consensus on this PR? 
> 
> 
> 
> We will cut 4.4 LTS start of next week so we should get this done before, or move to 4.5

There's only a typo to be fixed here. The tests were not running for the contributor because he was using the incorrect command to do so.

It would still be good to fix the typo, though.

cziesman:
Due to the default activation rules, the full profile seems to interfere with the ability to run integration tests as a standalone task. The only way I could run them was via the profile that I created. If there is a combination  of properties and profiles that enables integration tests to run, please let me know what that is, as it is not documented as far as I can tell. 

I can try to move the profile to `parent/pom.xml` but it may not work, if my memory is correct. In any case, how would a contributor know what belongs in `pom.xml` vs `parent/pom.xml`?

orpiske:
> Due to the default activation rules, the full profile seems to interfere with the ability to run integration tests as a standalone task. The only way I could run them was via the profile that I created. If there is a combination of properties and profiles that enables integration tests to run, please let me know what that is, as it is not documented as far as I can tell.

If you are running [as you described above](https://github.com/apache/camel/pull/12966#discussion_r1476489604) (where you said you were running the command `mvn failsafe:verify`), then you are **doing it wrong**. That's why the tests are not running for you. You should use `mvn verify` [as explained in the failsafe documentation](https://maven.apache.org/surefire/maven-failsafe-plugin/).

> 
> I can try to move the profile to `parent/pom.xml` but it may not work, if my memory is correct. 

The only item to fix is the typo on the project's `pom.xml` file.

> In any case, how would a contributor know what belongs in `pom.xml` vs `parent/pom.xml`?

Typically, learning the details about a project happen organically: via reviews, discussions (on Zulip, here or on the mailing list, etc), etc. This is specially true a in project as large and as complex as Camel, with a rather small number of core contributors. 

We try our best to document how users can get started with [contributions in our Contributing Guide](https://camel.apache.org/camel-core/contributing/). In there, we link to a number of issues that are [good for first timers](https://issues.apache.org/jira/browse/CAMEL-20384?filter=12348073): issues and improvements that don't involve the project build, CI, core code, architecture and other items that may require a greater understanding of the project needs and requirements. [That's how many of us started](https://github.com/apache/camel/pull/2887). In many cases, we rely on convention over configuration (such as the case of running the unit tests using `mvn test` and the integration tests using `mvn verify`), so that contributors jumping from another project can reuse what they've learned.

orpiske:
We can close this. I merged the typo fix on #13069 (which was what we needed). Unfortunately I couldn't cherry-pick the specific change, but credits were given in the commit message. 

Thanks for your contribution!

PR #12985, opened 2024-02-03
CAMEL-20361: camel-jbang - Make jolokia pluggable for camel-platform-http-main

kulagaIA:
# Description
`camel-jolokia` is now a context plugin with `@DevConsole` annotation.

I was not able to generate file `core/camel-jolokia/src/generated/resources/META-INF/services/org/apache/camel/dev-console/jolokia` with the `camel-package-maven-plugin`. This file was added by hands, and it is possible that something is wrong with `camel-jolokia` maven configuration, or something is wrong with the maven plugin (see https://camel.zulipchat.com/#narrow/stream/364655-camel-core-dev/topic/how.20to.20add.20context.20plugin/near/419592914 ).

Also added jolokia to example repo: https://github.com/apache/camel-examples/pull/133

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20361


core/camel-jolokia/src/main/java/org/apache/camel/impl/jolokia/DefaultJolokiaHttpRequestHandlerFactory.java
```diff
@@ -36,17 +37,21 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class JolokiaHttpRequestHandlerSupport extends ServiceSupport implements StaticService {
+/**
+ * Default {@link org.apache.camel.console.DevConsoleRegistry}.
+ */
+@DevConsole("jolokia")
```
davsclaus, line 43: This is wrong. @DevConsole is only for camel-console type of consoles.
You need to come up with another annotation such as JdkFactory.

core/pom.xml
```diff
@@ -59,6 +59,7 @@
         <module>camel-cloud</module>
         <module>camel-health</module>
         <module>camel-console</module>
+        <module>camel-jolokia</module>
```
davsclaus, line 62: jolokia should not be a core module, move this under components

core/camel-api/pom.xml
```diff
@@ -56,6 +56,13 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
```
davsclaus, line 59: camel-api should NOT have any other dependencies, this must be removed

core/camel-api/src/main/java/org/apache/camel/spi/JolokiaHttpRequestHandlerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.apache.camel.spi;
+
+import org.apache.camel.StaticService;
+import org.jolokia.server.core.http.HttpRequestHandler;
+
+/**
+ * Factory to abstract the creation of the Jolokia HttpRequestHandler.
+ */
+public interface JolokiaHttpRequestHandlerFactory extends StaticService {
```
davsclaus, line 25: We should make a generic factory that just returns Object or / Class<T> T so we can use this for other stuff we can plugin into http-server in the future.

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
```diff
@@ -370,6 +371,7 @@ protected void initPlugins() {
         camelContextExtension.lazyAddContextPlugin(AnnotationBasedProcessorFactory.class,
                 this::createAnnotationBasedProcessorFactory);
         camelContextExtension.lazyAddContextPlugin(DumpRoutesStrategy.class, this::createDumpRoutesStrategy);
+        camelContextExtension.lazyAddContextPlugin(JolokiaHttpRequestHandlerFactory.class, this::createJolokiaHttpRequestHandlerFactory);
```
davsclaus, line 374: Jolokia should only be discovered if enabled, eg that jolokiaEnabled option in camel-main

davsclaus (changes requested):

davsclaus:
jolokia should be option and not in core at all. Instead there should be a more generic factory for platform-http, that can load jolokia and whatelse we add in the future. 

It should be platform-http that when it startup (doInit etc) will check if jolokia is configured to be enabled, and then use the factory to find the JAR (if present) and if present then create the factory and does what it need to do. If the JAR is not present then some exception is thrown. We do this elsewhere in Camel.

davsclaus:
@kulagaIA I wonder if you will get more time for this before start of next week, as that is the deadline where we will cut the next 4.4 LTS release.

kulagaIA:
> @kulagaIA I wonder if you will get more time for this before start of next week, as that is the deadline where we will cut the next 4.4 LTS release.
Inteding to finish before 12.07.24, is it ok?

kulagaIA:
/component-test camel-platform-http-main camel-platform-http camel-platform-http-jolokia

kulagaIA:
/component-test camel-platform-http-main camel-platform-http camel-platform-http-jolokia

kulagaIA:
Made PlatformHttpPluginRegistry in camel-platform-http to be a generic factory for platform-http, that can load jolokia and whatelse we add in the future.
Also, apologies for messing up with the commit dates, wanted to finish before 4.4 LTS but got sick

components/camel-platform-http-main/src/main/java/org/apache/camel/component/platform/http/main/MainHttpServer.java
```diff
@@ -465,17 +479,22 @@ protected void setupJolokia() {
...
-        try (JolokiaHttpRequestHandlerSupport jolokiaService = new JolokiaHttpRequestHandlerSupport()) {
-            jolokiaService.start();
-            requestHandler = jolokiaService.getHttpRequestHandler();
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
-        return requestHandler;
+        PlatformHttpPluginRegistry registry
+                = camelContext.getCamelContextExtension().getContextPlugin(PlatformHttpPluginRegistry.class);
+        JolokiaPlatformHttpPlugin jolokia
```
davsclaus, line 493: Jolokia should only be discovered if jolokiaEnabled

kulagaIA: this is only called inside 
if (jolokiaEnabled) {
    setupJolokia();
}

davsclaus: ok

davsclaus (approved):

davsclaus:
Sorry we have now updated main to 4.5.0-SNAPSHOT so you may need to update this PR a little bit, thanks

PR #13024, opened 2024-02-06
Keep DHIS2 component backwards compatible

claudemamo:
# Description

Keep DHIS2 component backwards compatible

# Target

# Tracking

### DIFF
--- a/components/camel-dhis2/pom.xml
+++ b/components/camel-dhis2/pom.xml
@@ -60,6 +60,12 @@
             <classifier>v40.2.2</classifier>
         </dependency>
         <dependency>
+            <groupId>org.hisp.dhis.integration.sdk</groupId>
+            <artifactId>jackson-resource-model</artifactId>
+            <version>${dhis2-java-sdk.version}</version>
+            <classifier>v2.39.1</classifier>
+        </dependency>
+        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-test-junit5</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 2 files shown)


components/camel-dhis2/pom.xml
```diff
@@ -59,6 +59,12 @@
             <version>${dhis2-java-sdk.version}</version>
             <classifier>v40.2.2</classifier>
         </dependency>
+        <dependency>
+            <groupId>org.hisp.dhis.integration.sdk</groupId>
+            <artifactId>jackson-resource-model</artifactId>
+            <version>${dhis2-java-sdk.version}</version>
+            <classifier>v2.39.1</classifier>
```
oscerd, line 66: This should be configurable maybe through a parameter in parent pom.

oscerd (changes requested):
Minor finding.

components/camel-dhis2/pom.xml
claudemamo: You mean the classifier? We won't likely change the classifier but add more dependencies like this in the POM with different classifiers. For example, when DHIS2 v41 comes out, we'll declare another dependency like the one above:

```xml
<groupId>org.hisp.dhis.integration.sdk</groupId>
<artifactId>jackson-resource-model</artifactId>
<version>${dhis2-java-sdk.version}</version>
<classifier>v41</classifier>
```

But the `v2.39.1` dependency will remain to avoid breaking any apps.

oscerd: Thanks for the explanation.

davsclaus (approved):

PR #13037, opened 2024-02-07
(RFC) CAMEL-20355: enable support for 2 types of throttler (option 2)

orpiske:

PR #13038, opened 2024-02-07
(RFC) CAMEL-20355: enable support for 2 types of throttler (option 1)

orpiske:
Alternative implementation to #13037.  

This PR keeps the common options as part of the ThrottleDefinition and uses a separate set of configuration classes (`ThrottlerConfig`, `ConcurrentThrottlerConfig` and `TotalRequestsThrottlerConfig`) with only the **minimal set of differences** configured via them. That means: 

* `TotalRequestsThrottlerConfig`: has only the `timePeriodMillis` setting
* `ConcurrentRequestsThrottlerConfig`: has only the `maximumConcurrentRequests` setting.

This one seems to require less changes on YAML files, but, IMHO requires uglier code and a bit more duplication.


davsclaus:
I am against EIPs have different config classes (the only one that does that is resequencer and its a bad EIP and we should fix this in the future). It makes everything ugly and complex in DSL and tooling when this happens.

component and endpoint also have all the options in one class etc. And users select what they configure.

I would suggest to
- mode option (configure if X or Y) can be an enum, 
- or you can also just add a boolean option that must be set=true to use the new behavior (but finding a good name can be tricky)
- add back the old timePeriodMillis option that was removed.

orpiske:
I've just created another draft based on the feedback (not reusing the PR to avoid messing up with the tests on my own CI).

Please see #13040

PR #13040, opened 2024-02-07
CAMEL-20355: switch throttler using Enum/flags

orpiske:
Here's another try.

This one uses an Enum to switch between throttling modes. 

- Known problems: the ManagedThrottlerMBean is not being registered and is not working at the moment. 

I'm still testing this on my own CI, but given it's not so radically different than previous ones, I'm hopeful it will be OK.  

Obs.: the commits are dirty (I know).


catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models.properties
```diff
@@ -204,6 +205,7 @@ to
 toD
 tokenize
 topic
+totalRequestsConfig
```
davsclaus, line 208: I assume these are left over from the old code

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models.properties
```diff
@@ -18,6 +18,7 @@ circuitBreaker
 claimCheck
 combinedServiceDiscovery
 combinedServiceFilter
+concurrentConfig
```
davsclaus, line 21: I assume these are left over from the old code

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/concurrentConfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
davsclaus, line 1: I assume these are left over from the old code

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/totalRequestsConfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
davsclaus, line 1: I assume these are left over from the old code

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
```diff
@@ -57,29 +57,60 @@ public class ThrottleDefinition extends ExpressionNode implements ExecutorServic
     @XmlAttribute
     @Metadata(label = "advanced", javaType = "java.lang.Boolean")
     private String rejectExecution;
+    @XmlAttribute
+    @Metadata(defaultValue = "1000", javaType = "java.time.Duration")
+    private String timePeriodMillis;
+
+    @XmlAttribute
+    @Metadata(label = "advanced", javaType = "org.apache.camel.model.ThrottlingMode", defaultValue = "TotalRequests",
```
davsclaus, line 65: I dont think it should be advanced, and I think this should be the first option, eg move to top so user select mode first.

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
```diff
@@ -89,35 +120,132 @@ public String getShortName() {
...
     }
 
+    /**
+     * Sets the maximum number of concurrent requests
+     *
+     * @param      maximumConcurrentRequests the maximum number of concurrent requests
+     * @deprecated                           Use {@link #maximumRequests(long)}
+     * @return                               the builder
+     */
+    @Deprecated(since = "4.4.0")
```
davsclaus, line 182: The new throttler code was added in 4.3 (non LTS) so IMHO its fine to remove/change the code as upgrade from LTS to LTS is what matters.

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
```diff
@@ -89,35 +120,132 @@ public String getShortName() {
...
+
+    /**
+     * Sets the number of concurrent requests
+     *
+     * @param      maximumConcurrentRequests the maximum number of concurrent requests
+     * @deprecated                           Use {@link #maximumRequests(long)}
+     * @return                               the builder
+     */
+    @Deprecated(since = "4.4.0")
+    public ThrottleDefinition maximumConcurrentRequests(String maximumConcurrentRequests) {
```
davsclaus, line 203: The new throttler code was added in 4.3 (non LTS) so IMHO its fine to remove/change the code as upgrade from LTS to LTS is what matters.

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
```diff
@@ -247,6 +375,19 @@ public ThrottleDefinition executorService(String executorService) {
         return this;
     }
 
+    /**
+     * Sets the throttling mode
```
davsclaus, line 379: We may want a little bit more doc here. Maybe talk a bit what each mode does.

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/ThrottleReifier.java
```diff
@@ -33,30 +35,45 @@ public ThrottleReifier(Route route, ProcessorDefinition<?> definition) {
...
         Expression maxRequestsExpression = createMaxRequestsPerPeriodExpression();
         if (maxRequestsExpression == null) {
             throw new IllegalArgumentException("MaxRequestsPerPeriod expression must be provided on " + this);
         }
 
-        Expression correlation = null;
-        if (definition.getCorrelationExpression() != null) {
-            correlation = createExpression(definition.getCorrelationExpression());
-        }
+        if (ThrottlingMode.toMode(definition.getMode()) == ThrottlingMode.ConcurrentRequests) {
```
davsclaus, line 55: Add parseString(definition...) as that does property placeholders

core/camel-management-api/src/main/java/org/apache/camel/api/management/mbean/ManagedConcurrentRequestsThrottlerMBean.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.api.management.mbean;
+
+import org.apache.camel.api.management.ManagedAttribute;
+
+public interface ManagedConcurrentRequestsThrottlerMBean extends ManagedProcessorMBean {
```
davsclaus, line 21: Can we avoid having different mbeans - as there is only 1 EIP and ideally we should match that with the mbean too.

Keep the old mbean name, and then add attribute for mode, so you can see there which mode it uses.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models.properties
orpiske: Yes it is. I'm not sure why the build did not remove those. It seems to have removed others. I'll check.

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
orpiske: Before or after `correlationExpression`. It seems like we enforce a particular order via `@XmlType`:

`@XmlType(propOrder = { "expression", "correlationExpression" })` 

I'm wondering if I should do something like:

`@XmlType(propOrder = { "expression", "mode", "correlationExpression" })`

core/camel-core-model/src/main/java/org/apache/camel/model/ThrottleDefinition.java
orpiske: Great. I was trying not to be super disruptive, but it makes sense. It's even better if we can get rid of this.

core/camel-management-api/src/main/java/org/apache/camel/api/management/mbean/ManagedConcurrentRequestsThrottlerMBean.java
orpiske: I am not sure, TBH. 

It seems that the bean works on top of the throttler itself. So, in some way, we would need to rework the code to make the throttler generic. 

I think we can do it, but I believe I might need to tackle the more critical issues first and then this.

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/ThrottleReifier.java
orpiske: I may need a clarification here. I'll ping you on Zulip / Slack.

core/camel-management-api/src/main/java/org/apache/camel/api/management/mbean/ManagedConcurrentRequestsThrottlerMBean.java
orpiske: Done. Also implemented in a way that we could add other types of throttlers in the future.

orpiske:
> I still need to update the documentation and include an entry on the migration guide. I will do this today.

Done.

davsclaus (approved):

PR #13041, opened 2024-02-07
Officially add Splunk HEC URL and token as query parameters

Nephery:
# Description

Make providing `splunkURL` and `token` through query parameters an officially documented feature of the Splunk HEC component.

**Background**

The current method (i.e. passing tokens via path parameters) leaks sensitive values into the logs, since Camel itself does not expect path parameters to contain sensitive info. This PR provides an alternative means of securely providing credentials to this component.

**PR Change Specifics**

* Marked `token` and `splunkURL` as query parameters.
* Ensured that these properties are also validated when they're set through query parameters.
  * Prevented syntax validation of the `splunkURL` and/or `token` path parameters when their respective query parameters are already set.

# Target

# Tracking


Nephery:
@oscerd wouldn't that be breaking backwards compatibility for a minor version upgrade? Or is that okay in Camel?

From a simplicity & code perspective, I agree that your suggestion would be a lot nicer.

oscerd:
Camel is too big to follow semantic versioning. We usually allow breaking changes between minor versions. When we release an LTS all the patch releases will avoid breaking changes

Nephery:
Okay, sounds good. In that case, I'll update the PR to do what you suggested.

Nephery:
> In my opinion it doesn't make sense to have both the parameters as path and query. I would leave just the url as path and add the token as mandatory query parameter. The proposed solution will confuse end user and since we are going to release 4.4.0 soon, I think we could remove the token as path parameter, add it only as mandatory query parameter and document it in the migration guide.

I adjusted the PR according to @oscerd's proposal.

oscerd (approved):

oscerd:
LGTM, if you have time it would be nice to have an example of usage of the component in camel-examples, camel-spring-boot-examples or camel-quarkus-examples. Let's wait for more reviews :-)

davsclaus:
@Nephery if you want to work more on splunk then we have this ticket
https://issues.apache.org/jira/browse/CAMEL-20393

PR #13060, opened 2024-02-08
CAMEL-20399: String to short type conversion fails

valdar:
# Description
String to short type conversion fails with:

`NoTypeConversionAvailable No type converter available to convert from type: java.lang.String to the required type: short`

to reproduce it just test:

`assertEquals(Short.parseShort("1"), context.getTypeConverter().mandatoryConvertTo(short.class, "1"));`

# Target

# Tracking


davsclaus:
why do you need a short, any real use-case for using it

davsclaus (approved):

oscerd:
I think there is an option in hdfs changed from int to short in 4.3.0.. we removed hdfs in 4.4.0 but I guess is useful anyway

davsclaus:
ok thanks. Yeah I was trying to say since we have not had this reported before then I wondered how you found out eg what did you did = what is the use-case.

But its a simple and good fix and indeed missing

valdar:
> why do you need a short, any real use-case for using it

I found out because of a component option that is set to short.

PR #13082, opened 2024-02-09
[CAMEL-20405] renamed classes from *IntegrationTest.java to *IT.java

cziesman:
# Description

Several test classes are named *IntegrationTest.java. These have been renamed to *IT.java so that they match the default naming pattern for the failsafe plugin.

# Target

# Tracking

- [X ] I checked that each commit in the pull request has a meaningful subject line and body.

- [X ] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes


orpiske (changes requested):
I think you also need to review the failsafe plugin configuration on the component's pom file.

davsclaus:
> failsafe

Otavio do you mean that some of the component have customized configuration for `failsafe-plugin` in their pom? It seems so as if you do a `git grep failsafe-plugin` then there are about 10 components.

But even so what changes are we looking for to do? Would it be if they do some naming matching where they expected names of xxxIntegrationTest style, and this PR renames the names back into xxxIT style.

orpiske:
> > I think you also need to review the failsafe plugin configuration on the component's pom file.
> 
> Otavio do you mean that some of the component have customized configuration for `failsafe-plugin` in their pom? It seems so as if you do a `git grep failsafe-plugin` then there are about 10 components.
> 
> But even so what changes are we looking for to do? Would it be if they do some naming matching where they expected names of xxxIntegrationTest style, and this PR renames the names back into xxxIT style.

No, I just mean that (I think) the failsafe plugin configuration for `camel-safesforce` also needs to be updated because [we do seems to adjust it for the `*IntegrationTest.java` pattern](https://github.com/apache/camel/blob/main/components/camel-salesforce/camel-salesforce-component/pom.xml#L449-L487).

orpiske:
> Otavio do you mean that some of the component have customized configuration for failsafe-plugin in their pom? It seems so as if you do a git grep failsafe-plugin then there are about 10 components.

I don't mean those, but just as a matter of explanation, there are some components that have custom configuration to support a few cases: 

* Enable testing on "alternative" platforms (`ppc64le`, `s390x`, etc). For instance, `camel-consul` and `camel-zookeeper`. 
* Add backwards/forwards compatibility testing. For instance, `camel-kafka` has extra goals to test the code against Kafka 2.x
* Other component-specific testing tweaks/configuration. For instance `camel-infinispan` and `camel-salesforce`. 

That's why I said mentioned that the configuration for `camel-salesforce` needed to be adjusted.

cziesman:
@orpiske Good catch on the pom.xml changes! I have updated them.

davsclaus (approved):

orpiske (approved):
LGTM. Thanks @cziesman!

davsclaus:
You cannot run the salesforce integrations tests automatically. You need to setup a real account and more, see the readme file.

davsclaus (changes requested):
see prev comment

cziesman:
@davsclaus There are many integration tests that require external resources, not just the Salesforce tests. This will obviously cause test failures because most people won't have the environments needed to run any of the integration tests.

Should integration tests run as part of `mvn verify`? They are enabled by default but can be skipped with `mvn verify -DskipITs` or with a property definition in `pom.xml` or `parent/pom.xml`.

cziesman:
I looked at the POM for the Salesforce component and the incremental build log. I don't think the name change was responsible for the build failures. I think that any code change would have caused the integration tests to fail, as they are executed automatically as part of running 'mvn verify' regardless of how they are named, because there is an explicit '<includes> element in the 'failsafe` configuration.

cziesman:
I renamed the Salesforce integration tests to *ManualIT.java to match the naming convention for other Camel manual integration tests. I also disabled execution by `mvn verify` unless the `integration` profile was being used.

davsclaus (approved):

davsclaus:
LGTM

PR #13087, opened 2024-02-11
camel-knative-http: Avoid Nullpointer on KnativeHttpProducer/Consumer factory

christophd:
# Description

- Lookup router and vertx instance when not explicitly set on the factory
- Avoids Nullpointer exception when for instance running the Knative component with Camel JBang

# Target

# Tracking

### DIFF
--- a/components/camel-knative/camel-knative-http/src/main/java/org/apache/camel/component/knative/http/KnativeHttpConsumerFactory.java
+++ b/components/camel-knative/camel-knative-http/src/main/java/org/apache/camel/component/knative/http/KnativeHttpConsumerFactory.java
@@ -53,6 +53,13 @@ public class KnativeHttpConsumerFactory extends ServiceSupport implements CamelC
     }
 
     @Override
+    protected void doInit() throws Exception {
+        if (router == null) {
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (changes requested):
the factory should NOT have null camel context. Why and how do you get a NPE

christophd:
@davsclaus in fact the vertx and router instance setter never gets called in the Camel core project. When running with Camel Quarkus the camel-quarkus-knative extension sets the instance like here: https://github.com/apache/camel-quarkus/blob/main/extensions/knative/runtime/src/main/java/org/apache/camel/quarkus/component/knative/KnativeProducerRecorder.java#L33

but when run without Quarkus e.g. in Camel JBang the setter never gets called so we may lookup from the Camel context or create a new Vert.x instance.

christophd:
Force pushed, assuming now that camel context is not null in the factory

davsclaus:
Can you try with implementing doStart or doInit and setup. Those ought to be called on startup which should allow to setup vertx and whatnot.

davsclaus (approved):

PR #13094, opened 2024-02-12
CAMEL-20409: camel-core - ModelReifierFactory should detect custom on…

davsclaus:
… classpath

# Description

# Target

# Tracking

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/ModelReifierFactory.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/ModelReifierFactory.java
@@ -28,22 +28,54 @@ import org.apache.camel.Route;
  */
 public interface ModelReifierFactory {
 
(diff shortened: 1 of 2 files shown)


core/camel-api/src/main/java/org/apache/camel/spi/ModelReifierFactory.java
```diff
@@ -28,22 +28,54 @@
  */
 public interface ModelReifierFactory {
 
+    /**
+     * Service factory key for custom factories.
+     */
+    String FACTORY = "model-reifier-factory";
```
davsclaus, line 34: @lburgazzoli so a custom JAR with a class that extends the default factory and has @JdkService(FACTORY) as annotation should make camel detect your class instead of the default

oscerd (approved):

PR #13095, opened 2024-02-12
camel-jbang - Automatic include http server when using knative as it …

davsclaus:
…needs this.

# Description

# Target

# Tracking


oscerd (approved):

christophd:
I have added an example to the kamelets-examples repository https://github.com/apache/camel-kamelets-examples/pull/56

This one should reproduce an issue that we have regarding the setup of the http-platform server and the Knative event consumer. The problem is that the consumer tries to resolve the http-platform service from the Camel context before this service has been setup. So the result is that the Knative consumer is not able to leverage the http-platform and incoming events are not consumed properly.

davsclaus:
Yeah thanks. This is likely easier if we bake in http transport directly in camel-knative then the complexity is lower
https://issues.apache.org/jira/browse/CAMEL-20412

christophd:
Potential successor PR for this one is #13120. It uses the same approach as given in this PR but also adds some fixes on how the KnativeHttp consumer references the platform http service in `doStart()` phase.

Solving https://issues.apache.org/jira/browse/CAMEL-20412 is still a good idea to simplify the Knative component  though

PR #13109, opened 2024-02-13
CAMEL-20404: Create a component for Qdrant Vector Database

lburgazzoli:
# Description

# Target

# Tracking


lburgazzoli:
This is a test PR to check an odd behavior where a number of file are slightly amended (import order, java doc) hwne building locally

orpiske:
This is looking pretty cool @lburgazzoli !

components/camel-qdrant/src/test/java/org/apache/camel/component/qdrant/QdrantTestSupport.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.camel.test.junit5.CamelTestSupport;
+import org.junit.jupiter.api.TestInstance;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class QdrantTestSupport extends CamelTestSupport {
+    @Container
+    public static QdrantContainer QDRANT = new QdrantContainer();
```
orpiske, line 29: We should probably create a `test-infra` component for this, as it simplifies maintaining it (including watching the code for updated container images for CVEs). 

I don't think it needs to be done right now, but we should probably log an issue and do it as soon as the PR is merged.

lburgazzoli: @orpiske test-infra module added

test-infra/camel-test-infra-qdrant/src/test/java/org/apache/camel/test/infra/qdrant/services/QdrantLocalContainerService.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public QdrantLocalContainerService(QdrantContainer container) {
+        this.container = container;
+    }
+
+    @Override
+    public void registerProperties() {
+        System.setProperty(QdrantProperties.QDRANT_HTTP_HOST, getHttpHost());
+        System.setProperty(QdrantProperties.QDRANT_HTTP_PORT, "" + getHttpPort());
```
orpiske, line 49: Just a nitpick: `String.valueOf`, otherwise it keeps adding noise to SonarSource analysis.

test-infra/camel-test-infra-qdrant/src/test/java/org/apache/camel/test/infra/qdrant/services/QdrantContainer.java
```diff
@@ -0,0 +1,84 @@
...
+    public String getHttpHost() {
+        return getHost();
+    }
+
+    public int getHttpPort() {
+        return getMappedPort(HTTP_PORT);
+    }
+
+    public HttpResponse<byte[]> put(String path, Map<Object, Object> body) throws Exception {
+        final String reqPath = !path.startsWith("/") ? "/" + path : path;
```
orpiske, line 69: I think you can use `StringHelper.after(path, "/", path)` here.

test-infra/camel-test-infra-qdrant/src/test/java/org/apache/camel/test/infra/qdrant/common/QdrantProperties.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.apache.camel.test.infra.qdrant.common;
+
+public class QdrantProperties {
+    public static final String INFRA_TYPE = "qdrant";
+    public static final String QDRANT_HTTP_HOST = "qdrant.http.host";
+    public static final String QDRANT_HTTP_PORT = "qdrant.http.port";
+    public static final String QDRANT_GRPC_HOST = "qdrant.grpc.host";
+    public static final String QDRANT_GRPC_PORT = "qdrant.grpc.port";
+    public static final String QDRANT_API_KEy = "qdrant.apiKey";
```
orpiske, line 25: There's a small typo here `KEy` instead of `KEY`.

orpiske (commented):
Looking great! I only find a typo and other small things to nitpick about 😇

components/camel-qdrant/src/main/java/org/apache/camel/component/qdrant/QdrantProducer.java
```diff
@@ -0,0 +1,264 @@
...
+    // Helpers
+    //
+    // ***************************************
+
+    private CamelContext getCamelContext() {
+        return getEndpoint().getCamelContext();
+    }
+
+    private <T> void call(ListenableFuture<T> future, BiConsumer<T, Throwable> consumer) {
+        Futures.addCallback(
```
orpiske, line 249: One thing I've been thinking is whether we could avoid using Guava. It's always a source of alignment problems and whatnot. It would be great if we could minimize it (IMHO).

oscerd: They're are improving from this point of view. The releases are much more stable now.

orpiske: > They're are improving from this point of view. The releases are much more stable now.

Oh, that's great to know. I have acquired a bad impression of it because of the problems of the past, but I happy to know it's not problematic anymore (as I do recognize it's useful). In this case, I think it's safe to ignore my suggestion above.

test-infra/camel-test-infra-qdrant/src/test/java/org/apache/camel/test/infra/qdrant/services/QdrantContainer.java
lburgazzoli: My understading is that `StringHelper.after`, returns the string after the given topic, here instead I was to add a '/' if it is not set, so I'm not sure to understand the suggestion

components/camel-qdrant/src/main/java/org/apache/camel/component/qdrant/QdrantProducer.java
lburgazzoli: No we can't as that is a fundamental part of grpc-java i.e. the `ListenableFuture` is what the various methods of the client to return ... I know it is a pain but there is not much we can do as long a grpc-java is involved

orpiske: Marking this as resolved, as @oscerd clarified this is not a problem anymore. Thanks!

test-infra/camel-test-infra-qdrant/src/test/java/org/apache/camel/test/infra/qdrant/services/QdrantContainer.java
orpiske: Oh, you're write. Not sure what I was thinking 🤦

orpiske (approved):
Nice one!

components/camel-qdrant/src/main/java/org/apache/camel/component/qdrant/QdrantEndpoint.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    public void doStart() throws Exception {
+        super.doStart();
+    }
+
+    @Override
+    public void doStop() throws Exception {
+        super.doStart();
```
aldettinger, line 119: Might be doStop() ?

lburgazzoli: done

components/camel-qdrant/src/test/java/org/apache/camel/component/qdrant/QdrantDeleteTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static io.qdrant.client.ConditionFactory.matchKeyword;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class QdrantDeleteTest extends QdrantTestSupport {
+
+    @DisplayName("Tests that trying to delete without passing the action name triggers a failure")
+    @Test
+    public void deleteWithoutRequiredParameters() {
+        Exchange result = fluentTemplate.to("qdrant:upsert")
```
aldettinger, line 34: Maybe we use qdrant:delete like other tests ?

aldettinger (approved):
Small neat, beyond looks good.
The converter part is quite handy :+1:

components/camel-qdrant/src/test/java/org/apache/camel/component/qdrant/QdrantDeleteTest.java
lburgazzoli: done

valdar (approved):

PR #13120, opened 2024-02-14
camel-jbang: Automatically load main http server for Knative component

christophd:
# Description

- KnativeHttpConsumer needs the main http server in order to provide a Http endpoint route that is supposed to be called by the Knative broker subscription
- Initialize platform main http server when Knative component is used in Camel JBang
- Use default port 8080 only if camel.jbang.platform-http.port setting has not been set
- Add KnativeHttpConsumer and KnativeHttpProducer factory as a service to the Camel context so init() and start() methods are called properly

Closes #13095

# Target

# Tracking

### DIFF
--- a/components/camel-knative/camel-knative-component/src/main/java/org/apache/camel/component/knative/KnativeComponent.java
+++ b/components/camel-knative/camel-knative-component/src/main/java/org/apache/camel/component/knative/KnativeComponent.java
@@ -20,7 +20,6 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.apache.camel.CamelContext;
-import org.apache.camel.CamelContextAware;
 import org.apache.camel.Endpoint;
 import org.apache.camel.component.knative.spi.Knative;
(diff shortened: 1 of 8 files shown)


christophd:
@davsclaus I was able to solve the issues related to the KnativeHttpConsumer not being able to leverage the MainHttpServer in Camel JBang. It also solves the Knative example provided in https://github.com/apache/camel-kamelets-examples/tree/main/jbang/knative

The only thing that bugs me is the fact that the MainHttpServer instance now also is started when using the Knative component as producer only in Camel JBang (e.g. `to("knative:event/my-event")` The Knative producer does not need the Http server instance as it connects to the Knative broker as a client. Camel JBang starts the Http server regardless of that because the Knative component is loaded

davsclaus (approved):

davsclaus:
Yes we can later see if we can optimize knative to only ask for http server for its consumer. You are welcome to create a JIRA about that

christophd:
Relates to CAMEL-20421

PR #13134, opened 2024-02-15
Refactor camel jbang dependency list to not depend on the export command

claudio4j:
The intention is two-fold, first to have the `dependency list` not have to depend on the Export sub-command. Second the calculate dependencies functionality will be used in an upcoming update of camel-k camel-jbang plugin to calculate the dependencies of the `IntegrationRun` sub-command.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
@@ -16,146 +16,86 @@
  */
 package org.apache.camel.dsl.jbang.core.commands;
 
-import java.io.File;
+import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
-import java.util.Properties;
+import java.util.Set;
+import java.util.TreeSet;
 
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.NodeList;
-
-import org.apache.camel.dsl.jbang.core.common.CommandLineHelper;
-import org.apache.camel.dsl.jbang.core.common.RuntimeUtil;
-import org.apache.camel.dsl.jbang.core.common.XmlHelper;
+import org.apache.camel.dsl.jbang.core.common.MavenGavComparator;
+import org.apache.camel.main.KameletMain;
+import org.apache.camel.main.download.DownloadListener;
 import org.apache.camel.tooling.maven.MavenGav;
-import org.apache.camel.util.CamelCaseOrderedProperties;
-import org.apache.camel.util.FileUtil;
 import picocli.CommandLine;
 
-@CommandLine.Command(name = "list",
-                     description = "Displays all Camel dependencies required to run")
-public class DependencyList extends Export {
+@CommandLine.Command(name = "list", description = "Displays all Camel dependencies required to run")
+public class DependencyList extends CamelCommand {
 
-    protected static final String EXPORT_DIR = CommandLineHelper.CAMEL_JBANG_WORK_DIR + "/export";
+    // ignored list of dependencies, can be either groupId or artifactId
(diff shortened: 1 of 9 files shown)


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
```diff
@@ -16,146 +16,89 @@
...
-@CommandLine.Command(name = "list",
-                     description = "Displays all Camel dependencies required to run")
-public class DependencyList extends Export {
+@CommandLine.Command(name = "list", description = "Displays all Camel dependencies required to run")
+public class DependencyList extends CamelCommand {
+
+    // ignored list of dependencies, can be either groupId or artifactId
+    private static final String[] SKIP_DEPS = new String[] {
+            "org.fusesource.jansi", "org.apache.logging.log4j", "camel-core-languages", "camel-endpointdsl",
+            "camel-java-joor-dsl" };
```
apupier, line 38: Why are there ignored dependencies?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
```diff
@@ -16,146 +16,89 @@
...
+        RunDownloadListener downloadListener = new RunDownloadListener();
+        main.setDownloadListener(downloadListener);
+        main.setSilent(!verbose);
+        // enable stub in silent mode so we do not use real components
+        main.setStubPattern("*");
+        // do not run for very long in silent run
+        main.addInitialProperty("camel.main.autoStartup", "false");
+        main.addInitialProperty("camel.main.durationMaxSeconds", "1");
+        main.addInitialProperty("camel.jbang.verbose", Boolean.toString(verbose));
+        main.addInitialProperty("camel.jbang.ignoreLoadingError", Boolean.toString(!verbose));
```
apupier, line 84: Why `verbose` parameter is used to `ignoreLoadingError`?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
claudio4j: The org.fusesource.jansi, org.apache.logging.log4j were already there before and I didn't want to disrupt it. The other ignored dependencies are mostly from camel core modules, I think these could be removed from the list.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
apupier: I fear it will break `dependency copy `on Camel routes which are in-progress. Which is a major use case when using in IDE integration (for reference https://github.com/camel-tooling/vscode-camelk/blob/b625bd537a52ca603edb0a92173dcd2292e48a37/src/JavaDependenciesManager.ts#L52 )

claudio4j: It was a leftover from previous test. Just removed now.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyList.java
claudio4j: I removed the `camel-*` artifacts from the ignored list.

davsclaus:
And the dependency copy command still works after this change ?

claudio4j:
> And the dependency copy command still works after this change ?

Sorry, no. I will put the PR in draft and work on the the copy and update subcommands.

oscerd (commented):
Optimization is a good choice, like refactoring sometimes, but is this something really needed? I have troubles to see what it is going to add

claudio4j:
This proposal was mostly driven by having the dependency list we are going to need in `camel-k run` command, so I thought why not have this in the `dependency list` itself instead of camel-k, but I failed to look for the impact on the copy and update sub-commands. I will close this PR and will update only the camel-k run sub-command. Sorry for the noise.

PR #13137, opened 2024-02-16
(chores) camel-core: added the ability to provide a cause to ObjectHelper.notNull

orpiske:


orpiske:
Just to support something we are working on for #13109

PR #13189, opened 2024-02-19
CAMEL-20381: camel-core - EIP model should include which exchange pro…

davsclaus:
…perties they set when using the EIP.

# Description

# Target

# Tracking


oscerd (approved):

davsclaus:
@jono-morris here is what I am working on.

If you want to help then we need to markup more EIPs on the Exchange, like what is done for split and loop. The descriptions can likely be found in the doc files, eg loop-eip.adoc etc. And make sure to set its java type etc.

We should also find out afterwards how we can take the options from the generated json file in the model, and automatic have that in the EIP docs (like its done for the EIP options itself). But this requires some changes in camel-website.

davsclaus:
The website docs does this

// eip options: START
include::partial$eip-options.adoc[]
// eip options: END

We need something ala

// eip options: START
include::partial$eip-exchangeProperties.adoc[]
// eip options: END

jono-morris:
Thanks, let me start by marking up some more of the EIPs

davsclaus:
I think this PR has all the EIPs - there are only a few that uses exchange properties

davsclaus:
> Yeah there didn't seem that many of them. `SagaProcessor` used `SAGA_LONG_RUNNING_ACTION` and a number of them used Exchange properties via `ExchangePropertyKey`, e.g. Enricher, InterceptSendToEndpointProcessor etc. not sure if you wanted to include those.

Not all of them as some of for internal use. Enricher uses dynamic processor (toD) that sets some but I may have forgotten to add it to the label

PR #13235, opened 2024-02-21
CAMEL-20410: more fixes

orpiske:


orpiske:
Note for Jetty: this one probably needs a second pass, as there seems to be outdated items there (and also CAMEL-20443 fixes).

PR #13270, opened 2024-02-22
Handle quote escape in bindy

Croway:
# Description

# Target

# Tracking


oscerd (approved):

davsclaus:
Nice magic with the scanner.

We have StringQuoteHelper for safe splitting when using quotes as double or single quote chars. But I guess the scanner and bindy can in theory use other quote chars - but who would not use " ?

davsclaus (approved):

davsclaus:
If we need to backport this, then its good to have a JIRA ticket for this to track this

PR #13301, opened 2024-02-26
CAMEL-20019: add platform http vertx cookie handler

jono-morris:
# Description

A simple cookie handler that allows cookie addition, retrieval and expiry/removal. 

Using the cookieJar that is available via the Vertx routing-context seemed to the be simplest implementation, and add/remove/get cookie operations delegate to the request/response Vertx objects.

Get/Remove operations on the CookieHandler return String values as opposed to Cookies to abstract away from specific Cookie implementations

The cookie handler is made available to routes as the 'COOKIE_HANDLER'  exchange property.

# Target

# Tracking


jono-morris:
@davsclaus, were you thinking of something along these lines for the Vertx cookie hander?

davsclaus (approved):

davsclaus:
Yes that is really good.

components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/cookie/CookieConfiguration.java
```diff
@@ -0,0 +1,167 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.platform.http.cookie;
+
+/**
+ * Attributes that are set when creating Cookies.
+ */
+public class CookieConfiguration {
```
davsclaus, line 22: You can add `@Configurer` to let camel source generate code to configure this without reflection.

Then you can configure this also via, something ala:

camel.component.platform-http.cookie-configuration.max-age=2000

jono-morris: Added the `@Configurer` annotation and changed the properties to `cookiexxx` to prevent duplicate names.  https://github.com/apache/camel/pull/13301/commits/c025fdb0fef8ce7cbfae4536be30e8f9b8be86fb

davsclaus (approved):

PR #13326, opened 2024-02-27
align output log

Croway:
```
[l.example.MyApplication.main()] MainHttpServer                 INFO      http://0.0.0.0:8080/   (POST) (accept:application/json produce:application/json)
[l.example.MyApplication.main()] MainHttpServer                 INFO      http://0.0.0.0:8080/q/dev    (produce:text/plain,application/json)
[l.example.MyApplication.main()] MainHttpServer                 INFO      http://0.0.0.0:8080/q/health    (produce:application/json)
[l.example.MyApplication.main()] MainHttpServer                 INFO      http://0.0.0.0:8080/q/jolokia    (produce:text/plain,application/json)
```

### DIFF
--- a/components/camel-platform-http-main/src/main/java/org/apache/camel/component/platform/http/main/MainHttpServer.java
+++ b/components/camel-platform-http-main/src/main/java/org/apache/camel/component/platform/http/main/MainHttpServer.java
@@ -327,27 +327,39 @@ public class MainHttpServer extends ServiceSupport implements CamelContextAware,
                 // log only if changed
                 if (last == null || last.size() != endpoints.size() || !last.containsAll(endpoints)) {
                     LOG.info("HTTP endpoints summary");
+                    int longestEndpoint = 0;
                     for (HttpEndpointModel u : endpoints) {
-                        StringBuilder sb = new StringBuilder();
-                        sb.append("http://0.0.0.0:" + (server != null ? server.getPort() : getPort()) + u.getUri());
-                        if (u.getVerbs() != null || u.getProduces() != null || u.getConsumes() != null) {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus:
You can maybe do so '%s:30.30` or whatever syntax to make the text more aligned as 3 columns in a spreadsheet.

You can also loop the lines and calculate the longest and then adjust the `%sN..N` syntax accordingly.

orpiske:
I wonder if there's a project that we could use that could take a bean as input and would pretty print the information as the output. 
Quite frankly, this is very hard to read.

orpiske:
Obs.: no talking about your proposal. Just venting my frustration that we have to keep concatenating strings and doing a lot of ifs and elses on the code. 

IMHO, the view should always be separated from the other layers on the code.

Croway:
@orpiske I agree, I was thinking about using some kind of template, but seems a bit overkill for this use case, the latest commit uses format with `%-Ns` and seems a bit more readable, I am open to using some kind of template though.

orpiske:
> @orpiske I agree, I was thinking about using some kind of template, but seems a bit overkill for this use case, the latest commit uses format with `%-Ns` and seems a bit more readable, I am open to using some kind of template though.

Yeah. I think the proposed change is OK given the constraints we have right now. 

Longer term, I think we need to find something that could be used globally on the code and would allow us to separate the (usually messier) presentation code from the rest.

davsclaus:
we use a table printer library in jbang

orpiske:
> we use a table printer library in jbang

That one looks pretty cool

davsclaus (approved):

PR #13350, opened 2024-02-29
Openapi3

davsclaus:
# Description

# Target

# Tracking


components/camel-openapi-java/pom.xml
```diff
@@ -56,57 +54,89 @@
...
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.swagger.core.v3</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
```
PascalSchumacher, line 117: Hi Claus, I think these exclusions are unnecessary, because Jackson does not have any Swagger dependencies.

davsclaus: Thanks

PR #13364, opened 2024-03-01
CAMEL-20495: camel-jsonpath - If resultType is List then automatic wr…

davsclaus:
…ap single element into a List.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-jsonpath/src/test/java/org/apache/camel/jsonpath/JsonPathSplitSingleListTest.java
+++ b/components/camel-jsonpath/src/test/java/org/apache/camel/jsonpath/JsonPathSplitSingleListTest.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.jsonpath;
+
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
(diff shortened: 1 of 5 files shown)


oscerd (approved):

components/camel-jsonpath/src/test/java/org/apache/camel/jsonpath/JsonPathSplitSingleListTest.java
```diff
@@ -0,0 +1,69 @@
...
+                        // it will be wrapped inside a List object.
+                        .split().jsonpath("$.store.book[0]", List.class)
+                        .to("mock:authors")
+                        .convertBodyTo(String.class);
+            }
+        };
+    }
+
+    @Test
+    public void testSplit() throws Exception {
```
rhuan080, line 50: Can we add more tests to make sure the others patterns will work as expected? 
```
from("direct:start2")
                        // we select first book, but since we split after wards then ensure
                        // it will be wrapped inside a List object.
                        .split().jsonpath("$.store.book", List.class)
                        .to("mock:authors2")
                        .convertBodyTo(String.class);

                from("direct:start3")
                        // we select first book, but since we split after wards then ensure
                        // it will be wrapped inside a List object.
                        .split().jsonpath("$.store.book[*]", List.class)
                        .to("mock:authors3")
                        .convertBodyTo(String.class);
```

davsclaus: Those 2 examples above will select all books and be a List already.
There is an online evaluator here: https://jsonpath.com/

rhuan080: Yes, I'm suggesting it because in my tests with the 3.18.x base I use the `expr.setOptions(new Option[]{Option.ALWAYS_RETURN_LIST});` it resolve the $.store.book[0], but is breaking the $.store.book. Thus, I would like to make sure it will solve one scenario without changing the behavior of the other. Make sense?

davsclaus: Yes I added your tests into class `JsonPathSplitMultipleListTest`

davsclaus: And see `JsonPathSplitSingleListOptionTest` for an example with ALWAYS_RETURN_LIST

rhuan080: Thank you! Can I backport to 3.18.x?

davsclaus: No 3.18 is EOL. Upstream only 3.21 and 3.22 is supported.

davsclaus:
I am backporting to supported 3.x versions

PR #13392, opened 2024-03-06
CAMEL-20522: camel-test - CamelTestSupport does not call tearDown for…

davsclaus:
… each test method in some test classes

# Description

# Target

# Tracking


components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
```diff
@@ -140,11 +138,13 @@ public long timeTaken() {
     public void beforeEach(ExtensionContext context) throws Exception {
         currentTestName = context.getDisplayName();
         globalStore = context.getStore(ExtensionContext.Namespace.GLOBAL);
+        setUp();
```
orpiske, line 141: I think we should rename this base class. IMHO, it can be very confusing to have this method named as `setup` and not running automatically as part of the `BeforeEach`. 

I base this on the reasoning that the name may imply the former way of setting up tests from JUnit 4. 

Wdyt?

PR #13343, opened 2024-02-29
[CAMEL-19779] Refactor code generators to use velocity templates (work in progress)

gnodet:
- **Add missing generated annotations and message**
- **Regen**
- **Switch to velocity for converter loaders**
- **Regen**

PR #13394, opened 2024-03-06
CAMEL-20518: replace invalid quote breaking the catalog generation on Windows

orpiske:


oscerd (approved):

davsclaus (approved):
Oh nice spotted. 

I wonder if the camel-package-maven-plugin can detect this. There is some code that grabs javadoc and "sanitize" the javadoc. Could be that you can search for this char and replace it with the other quote. Or do a WARN if detected or even fail etc.

orpiske:
> Oh nice spotted.
> 
> I wonder if the camel-package-maven-plugin can detect this. There is some code that grabs javadoc and "sanitize" the javadoc. Could be that you can search for this char and replace it with the other quote. Or do a WARN if detected or even fail etc.

Thanks! 

+1 
I think we should implement this sanitization, indeed. I think we could start looking at that after #13343, so not to cause too much rebasing work for that cleanup.

orpiske:
Created https://issues.apache.org/jira/browse/CAMEL-20527 based on @davsclaus suggestion

PR #13397, opened 2024-03-06
CAMEL-20517 camel-jbang-plugin-k doesn't recognize command

claudio4j:
# Description

Instead to spawn a new process, this fix downloads the camel-jbang plugin and run it as part of the same camel-jbang OS process.

The change in `DefaultClassResolver.java` was motivated as when searching for a resource file from the downloaded maven artifact it's not available from the classloader in `getApplicationContextClassLoader()` , so it has to search in the list of `classloaders`

I will add some tests, but in the meantime I would like some preliminary review.

# Target

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20517


davsclaus:
I would like if you can try make this without changing code in the core modules in that classloader. So instead can you add a custom classloader into camel-jbang and use it there

claudio4j:
@davsclaus @christophd ready for review.

davsclaus (approved):

PR #13403, opened 2024-03-07
CAMEL-20521: camel-amqp - AMQP publisher application is losing messages with local JMS transaction enabled

luigidemasi:


oscerd (approved):

components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsConfiguration.java
```diff
@@ -620,7 +620,13 @@ private Object doSendToDestination(
                 // Check commit - avoid commit call within a JTA transaction.
                 if (session.getTransacted() && isSessionLocallyTransacted(session)) {
                     // Transacted session created by this template -> commit.
-                    JmsUtils.commitIfNecessary(session);
+                    Assert.notNull(session, "Session must not be null");
+                    try {
+                        session.commit();
```
davsclaus, line 625: Why are you not calling JmsUtils.commitIfNessasary ?

davsclaus (changes requested):

components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsConfiguration.java
luigidemasi: @davsclaus take a look at jira description, it's all there.
https://issues.apache.org/jira/browse/CAMEL-20521
I'll probably open a PR on spring as well.

davsclaus: No camel-jms is for JMS and should be as-is that is battle tested with spring-jms and all those kind of JMS brokers.

This ticket is about AMQP which is not JMS.
If there is a special need for camel-amqp in AMQP mode, then that must be done specially in camel-amqp and not in general camel-jms.

Spring likely dont care about your AMQP as they use rabbitmq and its AMQP version.

orpiske: IMHO, a more appropriate fix for this would involve creating an specific JMS template for AMQP.

At a quick glace, it doesn't seem like the code easily allows this, so probably would involve some refactoring on the process.

luigidemasi: Ok, I'll try to reproduce the problem with plain jms component, if it's only a AMQP/QPid problem, I'll create a JmsTemplate for AMQP.

components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPConfiguration.java
```diff
@@ -51,4 +64,28 @@ public void setIncludeAmqpAnnotations(boolean includeAmqpAnnotations) {
...
+    public static class CamelAmqpTemplate extends CamelJmsTemplate {
+
+        public CamelAmqpTemplate(AMQPConfiguration config, ConnectionFactory connectionFactory) {
+            super(config, connectionFactory);
+        }
+
+        @Override
+        protected void commitIfNecessary(Session session) throws JMSException {
+            //# CAMEL-20521 - we can't use JmsUtils.commitIfNecessary because it catches
+            // jakarta.jms.IllegalStateException that can be thrown if the session is closed
```
zhfeng, line 76: I'm not sure why this can fix CAMEL-20521 when session is closed during the commit? It will lead to the re-connecting?

luigidemasi: It leads to a redelivery if configured, this fix allows the exception to bubble up and can be handled by the error handler.
At the moment the exception is stopped at   `org.springframework.jms.support.JmsUtils.commitIfNecessary`, not even logged at trace level, so the camel route continue on the normal flow, like the commit has ended successfully.

zhfeng: OK, thanks for the explaination!

components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsConfiguration.java
luigidemasi: @davsclaus @orpiske, not able to reproduce with plain jms, it seems that this happens only with amqp/Qpid. I moved the changes to camel-amqp component.

zhfeng (approved):

components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPConfiguration.java
```diff
@@ -51,4 +64,28 @@ public void setIncludeAmqpAnnotations(boolean includeAmqpAnnotations) {
...
+
+            // Check commit - avoid commit call within a JTA transaction.
+            if (session.getTransacted() && isSessionLocallyTransacted(session)) {
+                // Transacted session created by this template -> commit.
+                Assert.notNull(session, "Session must not be null");
```
davsclaus, line 81: This assert seems not to make sense as you will have NPE above

davsclaus (approved):

davsclaus:
Can you update this PR as 
```
diff --git a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/releases/camel-releases.json b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/releases/camel-releases.json
index 7f20c95b7c3..b32e2ae444c 100644
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/releases/camel-releases.json
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/releases/camel-releases.json
@@ -859,6 +859,7 @@
     {
         "version": "4.4.0",
         "date": "[20](https://github.com/apache/camel/actions/runs/8210131412/job/22481316768?pr=13403#step:8:21)24-02-17",
+        "eol": "2025-02-17",
         "kind": "lts",
         "jdk": "17,[21](https://github.com/apache/camel/actions/runs/8210131412/job/22481316768?pr=13403#step:8:22)"
     }
```

luigidemasi:
@davsclaus @zhfeng I ran my test this morning and I was able to reproduce the issue also with plain jms component. I'll continue to run tests but it seems that the issue is not limited to AMQP/Qpid unfortunately...

zhfeng:
@luigidemasi is it possible to share the reproducer?

orpiske:
> @davsclaus @zhfeng @orpiske I ran my test this morning and I was able to reproduce the issue also with plain jms component and Apache Artemis 2.28.0 Broker. I'll continue to run tests but it seems that the issue is not limited to AMQP/Qpid unfortunately...

Given the scope and criticality of this (as losing messages is truly a bad thing), I think it would be important to consider adding actual integration tests, instead of relying on mocks. This is particularly important if this can happen in other circumstances / other brokers, IMHO

luigidemasi:
> Given the scope and criticality of this (as losing messages is truly a bad thing), I think it would be important to consider adding actual integration tests, instead of relying on mocks. This is particularly important if this can happen in other circumstances / other brokers, IMHO

@orpiske The issue is not limited to the amqp component, I can reproduce it with plain camel-jms too, I added a test that reproduces the issue using a real Artemis broker running in a docker container.

orpiske (approved):
LGTM, thanks!

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/langchain-embeddings.json
zhfeng, line 1: Not sure why it is here but I think all of the `langchain-embeddings` changes should be excluded from this commit.

zhfeng (changes requested):

zhfeng (approved):

luigidemasi:
I also opened an issue on spring that has already been fixed:
https://github.com/spring-projects/spring-framework/issues/32473

davsclaus:
org.apache.camel.component.jms.JmsDeliveryDelayTest.testInOnlyWithDelay | 5.123s | FAILURE
-- | -- | --

luigidemasi:
The new spring version, 6.1.6, that contains the fix for this issue, has been released. Should I close this issue or just modify the commit removing everything except tests?

davsclaus:
@luigidemasi it would be great if we could have the tests (great work). And then upgrade spring (that can be in another PR)

davsclaus:
spring upgrade here: https://github.com/apache/camel/pull/13778

PR #13423, opened 2024-03-08
CAMEL-16557: add a free form key value map on components, dataformats, languages, etc

lburgazzoli:
# Description

# Target

# Tracking


lburgazzoli:
@davsclaus here a quick POC about addin arbitrary annotations to the catalog entries

davsclaus:
I dont think we need to have org.apache.camel as prefix for protocol, autoscaling etc. 
Also is there anything that is autoscaling = false. It seems its just everything that is true

lburgazzoli:
> I dont think we need to have org.apache.camel as prefix for protocol, autoscaling etc. 

At this stage it was a port 1-1 of what exists in camel-k where the `org.apache.camel` was meant as a namespace for camel specific things, opening the door to add other, non camel specific annotation, like `dev.knative, io.strimzi, software.amazon`. 

Now I think we can remove the namespace since we are moving to the camel catalog, I'll do so on Monday.

> Also is there anything that is autoscaling = false. It seems its just everything that is true

This was just a marker were the absence means autoscaling=false

lburgazzoli:
@davsclaus, I've removed the `org.apache.camel` prefix

davsclaus:
A bit more is still there

"metadata": { "camel.apache.org\/routes.language": "kotlin" },

lburgazzoli:
> A bit more is still there
> 
> "metadata": { "camel.apache.org/routes.language": "kotlin" },

done, hopefully

davsclaus:
Did you really put 

```
@Metadata(annotations = {
        "autoscaling=true",
})
```

in all the components ?

I think CK should just infer that it can scale, unless a very special component has set autoscaling=false
And then only put this in the json if its false. The protocol=http and those language stuff is fine

lburgazzoli:
ok, `autoscaling=true` has been removed. I guess we can use the protocol as initial way ot infer the autoscaling capabilities and then add specific annotation for special components as suggested.

davsclaus (approved):

oscerd (approved):

PR #13424, opened 2024-03-08
Update what-platforms-are-supported.adoc

nrkkalyan:
Added some additional clarification on the java version support.

# Description

# Target

# Tracking

### DIFF
--- a/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
+++ b/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
@@ -1,5 +1,5 @@
 = What platforms are supported?
 
-- Java 8 Support until 3.14.x
-- Java 11 Support for all 3.x version
-- Java 17 Support starting with 3.17.0
+- Java 8+ Support until 3.14.x
+- Java 11+ Support for all 3.x version
+- Java 17+ Support starting with 3.17.0
--- a/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
+++ b/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
@@ -1,6 +1,9 @@
 = What platforms are supported?
 
-- Java 8+ Support until 3.14.x
-- Java 11+ Support for all 3.x version
-- Java 17+ Support starting with 3.17.0
-- Java 21+ Support starting with 4.2
+|===
+| Camel Version | Compile Version | JRE Version
+| Until 3.14.x | Java 8 | Java 8+
+| Starting 3.x | Java 11 | Java 11+
+| Starting 3.17.0 | Java 17 | Java 17+
+| Starting 4.2 | Java 21 | Java 21+
+|===
--- a/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
+++ b/docs/user-manual/modules/faq/pages/what-platforms-are-supported.adoc
@@ -3,3 +3,4 @@
 - Java 8+ Support until 3.14.x
 - Java 11+ Support for all 3.x version
 - Java 17+ Support starting with 3.17.0
+- Java 21+ Support starting with 4.2


nrkkalyan:
Added additional clarity with respect to supported java versions for each version

davsclaus:
Java 8 supported from 3.0 to 3.14
Java 11 supported for all of 3.x
Java 17 supported from 3.17
Java 21 supported from 4.2

davsclaus:
Can you use this information to update the PR with better information, thanks

nrkkalyan:
Hi Claus

Thank you for your comments, but sorry to say, I don't understand what you mean "update the PR with better information,"

As my understanding is, one can use latest Java version (Java11 or above) for 3.x Apache camel.
And 3.x is not restricted to work only with Java11. 
Like wise 3.17 is compatible to compile and run with Java 17 and higher java version.
Please clarify?

davsclaus:
Remove the `+` sign

nrkkalyan:
The + sign I mean for java 11 and above.
Is it not the case that one can use Java 17 to compile and run applications with apache 3.x?

davsclaus:
> The + sign I mean for java 11 and above. Is it not the case that one can use Java 17 to compile and run applications with apache 3.x?

Only in Camel 3.17 onwards, as I listed in previous comments. The JDK versions change during the years.

nrkkalyan:
Ok I have added changes that would clearly state the different java versions one can use to Compile and as Runtime environments.
Basically this states that if an application is using any of the apache-camel version, say 3.16.0 then that application can use JRE 21. Please let me know if one cannot use JRE-21 for 3.16.0?

davsclaus:
No we dont support official any kind of JRE version - We only support JDK11, 17, 21 and so on. If you use JDK12, 14, 18 or whatever then its not supported.

nrkkalyan:
I mean running my own application that uses camel say 3.16.0. 
So in that case still I cannot use  Java 21 to compile and run my application just because my application has a dependency on camel 3.16.0?  This is very strange. Please clarify if I can use JDK 21 to compile and run my application. 

If you mean to say official support then do you mean, if my application is using camel 3.16.0 then you won't support if the error occurred in my application due to running in Java 21.

davsclaus:
Yes you can see the supported version on download page
https://camel.apache.org/download/

For example Camel 4.4 says
Supports Java 17 and 21

davsclaus:
> If you mean to say official support then do you mean, if my application is using camel 3.16.0 then you won't support if the error occurred in my application due to running in Java 21.

Yes correct. Also ASF is only community support.

nrkkalyan:
> Yes you can see the supported version on download page https://camel.apache.org/download/
> 
> For example Camel 4.4 says Supports Java 17 and 21

I am not sure if I was able to convey my question properly when I say JRE.
My case is I have a batch application deployed on AWS platform, the application process some files from S3 bucket. 
My application stack is: Java-21, Spring 2.7.x, Camel 3.16.0.

I understand the support part. But other than that,
If I use camel 3.16.0 in my application compiled and running on JDK-21 on AWS platform will there be any problem?

davsclaus:
> > Yes you can see the supported version on download page https://camel.apache.org/download/
> > For example Camel 4.4 says Supports Java 17 and 21
> 
> Sorry for extending the discussion, but it is very important for me and possibly others to understand the JRE part of camel.
> 
> I am not sure if I was able to convey my question properly when I say JRE. My case is I have a batch application deployed on AWS platform, the application process some files from S3 bucket. My application stack is: Java-21, Spring 2.7.x, Camel 3.16.0.
> 
> I understand the support part. But other than that, If I use camel 3.16.0 in my application compiled and running on JDK-21 on AWS platform will there be any problem?

I dont know try it out - its very likely it will as JDK have great backwards compatability, but Camel also uses 3rd party JARs and they may not. 

The ASF Camel project **only** support the JDK and Camel combinations I already mentioned. We have 25.000+ tests executed on CI servers for the JDK + Camel combinations that we use to help verify.

davsclaus:
If you need more help then ask on the chat room or mailing list

PR #13456, opened 2024-03-12
chore: minor camel-catalog cleanup

lburgazzoli:
# Description

# Target

# Tracking

### DIFF
--- a/catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
+++ b/catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
@@ -59,16 +59,34 @@ public class DefaultCamelCatalog extends AbstractCamelCatalog implements CamelCa
     private static final String SCHEMAS_XML = "org/apache/camel/catalog/schemas";
     private static final String MAIN_DIR = "org/apache/camel/catalog/main";
     private static final String BASE_RESOURCE_DIR = "org/apache/camel/catalog";
+
     public static final String FIND_COMPONENT_NAMES = "findComponentNames";
+    public static final String FIND_COMPONENT_LABELS = "findComponentLabels";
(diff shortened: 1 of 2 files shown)


catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
```diff
@@ -113,13 +131,32 @@ public void setRuntimeProvider(RuntimeProvider runtimeProvider) {
...
+        cache.remove(LIST_OTHERS_AS_JSON);
+
+        cache.remove(FIND_MODEL_NAMES);
+        cache.remove(FIND_MODEL_LABELS);
+        cache.remove(LIST_MODELS_AS_JSON);
+
+        cache.remove(FIND_TRANSFORMER_NAMES);
+        cache.remove(LIST_TRANSFORMERS_AS_JSON);
+
+        cache.remove(SUMMARY_AS_JSON);
```
lburgazzoli, line 159: @davsclaus I don't know if this was intentional but saw a number of element of the cache not being removed when a new runtime provider is set so I removed them. Maybe cleaning up the whole cache and re-cache some of the element upfront would be better ?

oscerd (approved):

catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
davsclaus: Yeah that is safer, and its no problem - also for changing provider, you would just create the provider up front and not change it runtime wise.

davsclaus: You are welcome to clear all cache, or create a JIRA so we wont forget

lburgazzoli: done

PR #13507, opened 2024-03-15
CAMEL-20565 - Azure servicebus dead-lettering

romain-pfund:
# Description

Add a enableDeadLettering consumer configuration to allow deadlettering failed message to the deadletter sub-queue

According to [pplication-level-dead-lettering](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues#application-level-dead-lettering). The type of the exception is in the DeadLetterReason and the stack trace of the exception in the DeadLetterDescription.

# Target

# Tracking


components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
```diff
@@ -228,8 +232,21 @@ public void onFailure(Exchange exchange) {
             if (cause != null) {
                 getExceptionHandler().handleException("Error during processing exchange.", exchange, cause);
             }
+
             if (!getConfiguration().isDisableAutoComplete()) {
-                clientWrapper.abandon(message).subscribeOn(Schedulers.boundedElastic()).subscribe();
+                if (getConfiguration().isEnableDeadLettering()) {
```
oscerd, line 237: I think you need also to double check if there is actually a subqueue in your client instance.

oscerd (changes requested):
Overall looks good, we need to do a full build and I think there is one thing missing about checking the subqueue

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
romain-pfund: Yes your're right, thanks for the feedback.

romain-pfund: I've just removed my precedent comment as it was not correct.

According to the [Azure ServiceBus documentation](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues)
"Azure Service Bus queues and topic subscriptions provide a secondary subqueue, called a dead-letter queue"
So every queue or subscription has a deadletter.

but you are right, is the consumer is already consuming subqueue, we cannot deadletter the message. So I need to check if the subQueue option is set

oscerd: Yes, that was what I meant.

romain-pfund: I've added the check
`Objects.isNull(getConfiguration().getSubQueue()) || getConfiguration().getSubQueue().equals(SubQueue.NONE)`

romain-pfund:
Is there a way to write a test to cover this "onFailure" part ?

for now I'm manully testing with route like:
```
  public static final String AZURE_SERVICE_BUS_COMPONENT = "azure-servicebus";
  @Override
  public void configure() {
    String subscription = "test-scr";
    String inputTopic = "test";
    fromF("%s:%s?subscriptionName=%s&enableDeadLettering=true", AZURE_SERVICE_BUS_COMPONENT, inputTopic, subscription)
        .routeId("routeTestId")
        .throwException(new Exception("test failure"))
        .log("${body}");

  }
```

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
```diff
@@ -228,8 +233,23 @@ public void onFailure(Exchange exchange) {
             if (cause != null) {
                 getExceptionHandler().handleException("Error during processing exchange.", exchange, cause);
             }
+
             if (!getConfiguration().isDisableAutoComplete()) {
-                clientWrapper.abandon(message).subscribeOn(Schedulers.boundedElastic()).subscribe();
+                if (getConfiguration().isEnableDeadLettering() && (Objects.isNull(getConfiguration().getSubQueue()) ||
```
oscerd, line 238: You can use also ObjectHelper from camel-utils to be aligned with the codebase

romain-pfund: ok thanks I'll check.
First time here so happy discover new things

oscerd:
> Is there a way to write a test to cover this "onFailure" part ?
> 
> for now I'm manully testing with route like:
> 
> ```
>   public static final String AZURE_SERVICE_BUS_COMPONENT = "azure-servicebus";
>   @Override
>   public void configure() {
>     String subscription = "test-scr";
>     String inputTopic = "test";
>     fromF("%s:%s?subscriptionName=%s&enableDeadLettering=true", AZURE_SERVICE_BUS_COMPONENT, inputTopic, subscription)
>         .routeId("routeTestId")
>         .throwException(new Exception("test failure"))
>         .log("${body}");
> 
>   }
> ```

In this component we are only using integration tests, so you could have a look at ServiceBusConsumerTest and write a similar tests with this option enabled, but you'll need a real Azure Servicebus queue/topic.

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
romain-pfund: so something like
`(ObjectHelper.isEmpty(getConfiguration().getSubQueue()) ||ObjectHelper.equal(getConfiguration().getSubQueue(), SubQueue.NONE)` ?

oscerd: Yes, something like that.

components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
romain-pfund: it's pushed

romain-pfund:
just a question, it seams that the PR build is looping, is it normal

oscerd:
It is not looping, it is just that you need to run a full build to regenerate the metadata of Azure Service bus component. So you should run mvn clean install -DskipTests from the root folder of the project and update the PR with a new commit.

oscerd (approved):
This looks good by the way. We can merge and regenerate after merging

romain-pfund:
I want just do the integration test. I should have time to do it tomorrow or monday

oscerd:
No rush. Thanks for your time!

romain-pfund:
Didn't had time yesterday, Maybe we can look to merge it and I'll do the intégration test later in another PR.
I'm a bit full this week, may have more time next week

oscerd:
Sure, please make the PR ready to review.

romain-pfund:
shoudn't I do the mvn clean install -DskipTests and commit first ?

oscerd:
Yes, if you have time.

romain-pfund:
got trouble building it (i'm on a windows os right now), you said we can merge then rebuild ?

oscerd:
Let's merge this and regen

PR #13545, opened 2024-03-20
(chores) camel-mllp: parallelize unit tests

orpiske:


oscerd (approved):

davsclaus (approved):

davsclaus:
fyi I think there is one test that always fail on jdk21 - there is a JIRA about it

orpiske:
> fyi I think there is one test that always fail on jdk21 - there is a JIRA about it

Yeah, I noticed that one when I was looking at this. I marked it as `help-wanted` and good start contribution.

PR #13593, opened 2024-03-22
CAMEL-20602: Support user properties on Camel JBang bind command

christophd:
# Description

CAMEL-20602

- Allow to specify properties on source, sink, step in a Pipe
- User properties can be set as additional command options targeting either the source or the sink or a specific step

This PR adds a new option to the Camel JBang bind command `--property`. Users can set properties on source, sink or steps with this syntax:

```
camel bind timer-to-log --source timer-source --sink log-sink --property source.message=Hello --property sink.showHeaders=false
```

This will result in the Pipe using the properties on the source and sink.

```yaml
apiVersion: camel.apache.org/v1
kind: Pipe
metadata:
  name: timer-to-log
spec:
  source:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1
      name: timer-source
    properties:
      message: Hello
  sink:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1
      name: log-sink
    properties:
      showHeaders: true
```

# Target

# Tracking

See CAMEL-20602


oscerd (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Bind.java
```diff
@@ -64,21 +71,51 @@ public class Bind extends CamelCommand {
                         required = true)
     String sink;
 
+    @CommandLine.Option(names = { "-p", "--prop", "--property" },
```
davsclaus, line 74: We only use a single and double slash for the commands in camel-jbang

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Bind.java
```diff
@@ -64,21 +71,51 @@ public class Bind extends CamelCommand {
...
+                        description = "Adds a pipe property in the form of [source|sink|step-<n>].<key>=<value> where <n> is the step number starting from 1",
+                        arity = "0")
+    String[] properties;
+
+    @CommandLine.Option(names = { "--output", "-o" },
```
davsclaus, line 79: same here

davsclaus (changes requested):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Bind.java
christophd: I was copying this from Run command: https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java#L235

davsclaus: Ah yeah we should have that changed now after the 4.4 LTS so it can be consistent. You are welcome to fix that as well. Maybe in another PR and add a note in the upgrade guide 4.x

christophd: so to be sure it is about removing the `-p` option, right? I can do that for the new options in this PR already, np

davsclaus: yeah

christophd: done

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Bind.java
christophd: done, thx

davsclaus (approved):

PR #13626, opened 2024-03-27
Reword some configuration properties

gnodet:
I used LLM to reword some properties, but a review from a native english speaker would be welcomed.


core/camel-api/src/main/java/org/apache/camel/spi/StreamCachingStrategy.java
```diff
@@ -117,38 +117,44 @@ interface SpoolRule {
     boolean isEnabled();
 
     /**
-     * To filter stream caching of a given set of allowed/denied classes. By default, all classes that are
-     * {@link java.io.InputStream} is allowed.
+     * To filter stream caching of a given set of allowed/denied classes. Classes inheriting from any of the specified
+     * classes will be cached.
+     * By default, all classes that inherit from {@link java.io.InputStream} are allowed.
```
apupier, line 122: it is not only rewording, it is adding more information on which class are cached. Given that you mention that it is a "reword", is it expected?

core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json
```diff
@@ -19,123 +19,123 @@
...
+    { "name": "camel.main.autoStartup", "description": "Specifies if the object should start automatically when Camel is initiated. It is important to note that only routes can be disabled at the moment, as CamelContexts are always started. If auto startup is set to false on CamelContext, it will take precedence and no routes will be started. In this case, CamelContext must be started explicitly using the org.apache.camel.CamelContext.start() method, followed by manual starting of routes using CamelContext.getRouteController().startRoute(String) . By default, the object will start up automatically.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "boolean", "javaType": "boolean", "defaultValue": true },
```
apupier, line 28: `By default, the object will start up automatically.`

What is the object?

core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json
```diff
@@ -19,123 +19,123 @@
...
-    { "name": "camel.main.compileWorkDir", "description": "Work directory for compiler. Can be used to write compiled classes or other resources.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "string", "javaType": "java.lang.String" },
+    { "name": "camel.main.beanIntrospectionLoggingLevel", "description": "Configures the logging level for bean introspection, logging its usage activity. The default logging level is {link LoggingLevel#TRACE TRACE} .", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "object", "javaType": "org.apache.camel.LoggingLevel", "enum": [ "ERROR", "WARN", "INFO", "DEBUG", "TRACE", "OFF" ] },
```
apupier, line 33: the introduced link is not specifying the full classname, i do not understand how ti can work

core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json
```diff
@@ -19,123 +19,123 @@
...
+    { "name": "camel.main.jmxEnabled", "description": "Enables JMX in your Camel application.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "boolean", "javaType": "boolean", "defaultValue": true },
+    { "name": "camel.main.jmxManagementMBeansLevel", "description": "Sets the mbeans registration level. The default value is ManagementMBeansLevel#Default .", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "object", "javaType": "org.apache.camel.ManagementMBeansLevel", "defaultValue": "Default" },
```
apupier, line 69: `ManagementMBeansLevel#Default `

I think the full name should be provided or none

core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json
```diff
@@ -19,123 +19,123 @@
...
     { "name": "camel.main.messageHistory", "description": "Sets whether message history is enabled or not. Default is false.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "boolean", "javaType": "boolean", "defaultValue": "false" },
-    { "name": "camel.main.modeline", "description": "Whether camel-k style modeline is also enabled when not using camel-k. Enabling this allows to use a camel-k like experience by being able to configure various settings using modeline directly in your route source code.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "boolean", "javaType": "boolean", "defaultValue": "false" },
+    { "name": "camel.main.modeline", "description": "Specifies whether camel-k style modeling is also enabled when not using camel-k. Enabling this allows you to use a camel-k like experience by being able to configure various settings using modeline directly in your route's source code.", "sourceType": "org.apache.camel.main.DefaultConfigurationProperties", "type": "boolean", "javaType": "boolean", "defaultValue": "false" },
```
apupier, line 84: this is wrong, modeline is a specific terminology

apupier (changes requested):
I reviewed only few files and it introduces several errors. I think it is better to not do a bulk change. it is better to have correct information with medium english than wrong one with good english.

orpiske (approved):
Nice one!

orpiske (commented):
On a second thought, I think @apupier is right on a few points. I think it would be better to go for a smaller set of changes as there are, indeed, some errors. 

That said, some of the changes I saw, do make the documentation clearer ... so I think the approach is interesting.

orpiske:
@gnodet on a side note: what are you using for these changes? In the previous doc work (for components and EIPs) I have used Grazie Pro plugin for IntelliJ and it is very competent at finding typos and grammar issues.

PR #13686, opened 2024-04-03
CAMEL-20639: Support Kafka topic references in Pipes

christophd:
# Description

CAMEL-20639

- Enhance Camel JBang bind command to support Kafka topic resources
- Add Strimzi Kafka topic endpoint binding provider

# Target

# Tracking

### DIFF
--- a/dsl/camel-yaml-dsl/camel-yaml-dsl/src/main/java/org/apache/camel/dsl/yaml/YamlRoutesBuilderLoader.java
+++ b/dsl/camel-yaml-dsl/camel-yaml-dsl/src/main/java/org/apache/camel/dsl/yaml/YamlRoutesBuilderLoader.java
@@ -112,7 +112,7 @@ public class YamlRoutesBuilderLoader extends YamlRoutesBuilderLoaderSupport {
     @Deprecated
     private static final String BINDING_VERSION = "camel.apache.org/v1alpha1";
     private static final String PIPE_VERSION = "camel.apache.org/v1";
-    private static final String STRIMZI_VERSION = "kafka.strimzi.io/v1";
+    private static final String STRIMZI_VERSION = "kafka.strimzi.io/v1beta2";
     private static final String KNATIVE_MESSAGING_VERSION = "messaging.knative.dev/v1";
     private static final String KNATIVE_EVENTING_VERSION = "eventing.knative.dev/v1";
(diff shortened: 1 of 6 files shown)


dsl/camel-yaml-dsl/camel-yaml-dsl/src/main/java/org/apache/camel/dsl/yaml/YamlRoutesBuilderLoader.java
```diff
@@ -112,7 +112,7 @@ public class YamlRoutesBuilderLoader extends YamlRoutesBuilderLoaderSupport {
     @Deprecated
     private static final String BINDING_VERSION = "camel.apache.org/v1alpha1";
     private static final String PIPE_VERSION = "camel.apache.org/v1";
-    private static final String STRIMZI_VERSION = "kafka.strimzi.io/v1";
+    private static final String STRIMZI_VERSION = "kafka.strimzi.io/v1beta2";
```
davsclaus, line 115: Must it be exactly v1beta2, what happens when there is a v1beta3 and when it goes v1 GA.
I wonder if we need to be exact or can be more open, and just match on startsWith /v1beta2 /v1beta3 /v1 or whatever

davsclaus (approved):

dsl/camel-yaml-dsl/camel-yaml-dsl/src/main/java/org/apache/camel/dsl/yaml/YamlRoutesBuilderLoader.java
christophd: In theory Strimzi team could rename/delete the custom resource from KafkaTopic to something different in v1 but that is not very likely to happen. So the idea to match just `kafka.strimzi.io` sounds good to me (I will work on it on a separate PR).

I think for the bind command where the Pipe is created by Camel we still need to stick to a specific version that is used by default. Users can always bind to a fully custom object reference providing the full resource identifier such as `kafka.strimzi.io/v1:KafkaTopic:my-topic`.

davsclaus: ok

oscerd (approved):

PR #13690, opened 2024-04-03
CAMEL-20643: Add option to propagate Opentelemetry Context even when t…

johnpoth:
…racing is disabled for a Camel Processor

https://issues.apache.org/jira/browse/CAMEL-20643

Adds `propagateContext` option to `OpenTelemetryTracingStrategy` that propagates the Opentelemetry Context even when the Processor is excluded. This is useful when custom Spans are created with @WithSpan for fine grained tracing. It is turned off by default for backward compatibility.

Thanks !

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/OpenTelemetryTracingStrategy.java
+++ b/components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/OpenTelemetryTracingStrategy.java
@@ -36,6 +36,8 @@ public class OpenTelemetryTracingStrategy implements InterceptStrategy {
 
     private final OpenTelemetryTracer tracer;
 
+    private boolean propagateContext;
+
     public OpenTelemetryTracingStrategy(OpenTelemetryTracer tracer) {
         this.tracer = tracer;
     }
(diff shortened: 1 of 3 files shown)


components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/OpenTelemetryTracingStrategy.java
```diff
@@ -36,6 +36,8 @@ public class OpenTelemetryTracingStrategy implements InterceptStrategy {
 
     private final OpenTelemetryTracer tracer;
 
+    private boolean propagateContext = false;
```
oscerd, line 39: This is false by default.

johnpoth:
Thanks @oscerd  for reviewing! I changed the default to `true`. In that case we might want to consider this as a bug ? I'm not sure why someone wouldn't want the OpenTelemetry Context propagated ... WDYT ? Thanks !

oscerd:
What I meant it's just that set that to false is not needed. It could be false by default, but a primitive boolean is false when you declare the variable

johnpoth:
:-) I see ! Updated, should be good, thanks !

oscerd (approved):

johnpoth:
Thanks !

PR #13707, opened 2024-04-05
Fix camel-vertx-websocket endpoint instantiation when connection to r…

alexisgra:
…emote websocket server

# Description

While trying to migrate from `camel-ahc-ws` to `camel-vertx-websocket`, I encountered a problem with the `consumeAsClient` option. The remote server I was trying to interact with refused to upgrade to websocket because of invalid query parameters.

Previously, endpoint properties were set after configuring the `websocketURI`. As a result, when using the endpoint with `consumeAsClient` option, endpoint properties were passed as query param (`consumeAsClient` for example). Some APIs reject upgrade to websocket when unknown parameters are passed in the call. 

The purpose of this commit is to set endpoint properties before configuring `websocketURI`. So properties are binded and removed from the parameters map before configuring the `websocketURI`. 

This is inspired by the way `VertxHttpEndpoint` is instantiated.

I modified the endpoint instanciation and I added a test to verify that endpoint properties are correctly removed from the `websocketURI` query params.

# Target

# Tracking

### DIFF
--- a/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketComponent.java
+++ b/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketComponent.java
@@ -76,6 +76,13 @@ public class VertxWebsocketComponent extends DefaultComponent implements SSLCont
             }
         }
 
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus:
Thanks I created a JIRA
https://issues.apache.org/jira/browse/CAMEL-20656

PR #13708, opened 2024-04-06
CAMEL-18017: return body part fields copied from parsed MDN entity to avoid corruption

jono-morris:
# Description

The body part fields of a parsed MDN are written out to an output stream when a signature needs to be verified.  Any (ordering/formatting) changes introduced when the entity is parsed and written mean that the signature verification will fail.  One approach is to write out a direct copy of the body part fields of the received entity instead of trying to parse and recreate each field.

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

davsclaus:
great work, you are welcome to keep digging into other JIRAs related to camel-as2.

PR #13723, opened 2024-04-09
Camel 19613: Convert components to use the camel-test-infra-core :camel-asterik

anjaliInfy1822:
Convert components to use the camel-test-infra-core :camel-asterik


components/camel-asterisk/pom.xml
```diff
@@ -17,7 +17,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
orpiske, line 22: This does not seem related to the change, so I'd suggest to leave this one untouched.

orpiske (changes requested):
I am not sure if we want to proceed with these conversions without solving CAMEL-19742 first. 

Most of the changes look OK to me, but I'll mark this as "Request Changes" to hear what others have to say about this (c/c: @davsclaus, @oscerd).

components/camel-asterisk/pom.xml
anjaliInfy1822: Thanks @orpiske will remove whatever autoformatting took place here.


> This does not seem related to the change, so I'd suggest to leave this one untouched.

orpiske: Thanks

orpiske:
> Unless there is agreement of doing everything like this everywhere, I would avoid introducing different approaches.

+1. I agree with you.

Last time I talked with @davsclaus, we agreed that we need to have a way to convert the code in a way that is cleaner, simpler and easier to use. That's part of what we would like to do in CAMEL-19742 ... But we don't have that yet. 

So, as much as I would like to see the code using a modern test approach, I leaning towards a -1 on this change too (and will add a reinforcement note on the ticket that this effort is on hold).

orpiske:
@anjaliInfy1822 I **truly** appreciate your help with proposing this change. 

However, at the moment, we are blocked waiting for a change that would make converting easier and less disruptive. Unfortunately we don't have that yet - and it's going to take a while until we are able to finish it. 

Because of that, I am closing your PR without merging.

PR #13795, opened 2024-04-15
hazelcast-seda: Nested transactions are not allowed!

zhelev:
# Description

If **transacted** is enabled for a **hazelcast-seda** queue, the transaction never gets committed if body is _null_. This is often the case when polling from an empty queue. This leads to "Camel Exception cause:  java.lang.IllegalStateException: Nested transactions are not allowed!". We found this issue as our Hazelcast txBackupLogs consumed a large amount of memory.

# Target
Camel 4. 

Older versions are also affected. It would be nice to merge down to Camel 3 LTS versions.


davsclaus:
Can you create a JIRA ticket as backports of bug fixes would need to have this tracked

davsclaus:
I created a ticket
https://issues.apache.org/jira/browse/CAMEL-20677

PR #13823, opened 2024-04-17
CAMEL-20614: deep-copy output processors during instantiation of a route template

bartoszpop:
CAMEL-20614: deep-copy output processors during instantiation of a route template

# Description

When multiple threads try to instantiate and send an exchange to the same kamelet in parallel, org.apache.camel.component.kamelet.KameletConsumerNotAvailableException may be thrown because the underlying RouteTemplateDefinition is shallow-copied and changes to the RouteDefinition are reflected in the RouteTemplateDefinition.

# Target

# Tracking

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/model/NoOutputDefinition.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/model/NoOutputDefinition.java
@@ -32,7 +32,7 @@ public abstract class NoOutputDefinition<Type extends ProcessorDefinition<Type>>
     public NoOutputDefinition() {
     }
 
-    NoOutputDefinition(NoOutputDefinition source) {
+    protected NoOutputDefinition(NoOutputDefinition source) {
         super(source);
     }
 }
--- a/core/camel-core-model/src/main/java/org/apache/camel/model/RouteTemplateDefinition.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/model/RouteTemplateDefinition.java
@@ -38,10 +38,6 @@ import org.apache.camel.spi.Metadata;
 import org.apache.camel.spi.Resource;
 import org.apache.camel.spi.ResourceAware;
 
-import static java.util.Collections.emptyList;
-import static java.util.Collections.emptyMap;
-import static java.util.Objects.requireNonNullElse;
-
 /**
  * Defines a route template (parameterized routes)
  */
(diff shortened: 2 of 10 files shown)


core/camel-core-model/src/main/java/org/apache/camel/model/RouteTemplateDefinition.java
```diff
@@ -416,10 +421,10 @@ public RouteDefinition asRouteDefinition() {
         copy.setLogMask(route.getLogMask());
         copy.setMessageHistory(route.getMessageHistory());
         copy.setOutputType(route.getOutputType());
-        copy.setOutputs(route.getOutputs());
-        copy.setRoutePolicies(route.getRoutePolicies());
+        copy.setOutputs(copy(route.getOutputs()));
+        copy.setRoutePolicies(new ArrayList<>(requireNonNullElse(route.getRoutePolicies(), emptyList())));
```
gnodet, line 425: While this would be equivalent, I wonder if adding a method such as 
```
List<T> shallowCopy(List<T> list) {
    return (list != null) ? new ArrayList<>(list) : null;
}
```
same for maps...

core/camel-core-model/src/main/java/org/apache/camel/model/NoOutputDefinition.java
```diff
@@ -29,4 +29,10 @@ public List<ProcessorDefinition<?>> getOutputs() {
         return Collections.emptyList();
     }
 
+    public NoOutputDefinition() {
+    }
+
+    NoOutputDefinition(NoOutputDefinition source) {
```
gnodet, line 35: Should all the copy constructors be  `protected` rather than package private ? This won't change much as afaik all the models are in the same package, but the idea of this constructor is that it will be called by children classes, so protected makes more sense here.

gnodet:
In addition to the two minor comments above, I think it would be really nice if the test case you wrote could be slightly reworked as a unit test and added to this PR.

orpiske (changes requested):
We might also want to do a full test on this one. I'll trigger and report back (likely tomorrow as the build queue is quite large at the moment).

bartoszpop:
@gnodet comments addressed, please review

gnodet (approved):

davsclaus (approved):

orpiske (approved):
LGTM.

davsclaus:
Can you add copy to Enrich / PollEnrich as well in another PR (PR for main is fine)

bartoszpop:
> Can you add copy to Enrich / PollEnrich as well in another PR (PR for main is fine)

@davsclaus I did prepare for both - [main](https://github.com/apache/camel/pull/13869) and [camel-4.4.x](https://github.com/apache/camel/pull/13869). Please review.

PR #13837, opened 2024-04-17
CAMEL-20388: Restore the old cookieStore Strategy

Croway:
The issue is related to https://github.com/cometd/cometd/issues/1600, salesforce server replies with Set-Cookie values that are wrong or deprecated, and the latest setCookieParser implementation in Jetty 12 `RFC6265SetCookieParser` ignore some of the cookies, therefore, when the /connect endpoint is invoked, some cookies are missing and we get `403::Unknown client`.

This PR simply restores the old setCookie and cookieParser strategy.


zhfeng (approved):

davsclaus (approved):

davsclaus:
I assume this needs to be backported to 4.4.x branch also

PR #13866, opened 2024-04-19
Add handshakeHeaders uri param in vertx-websocket component

alexisgra:
# Description

Some websocket apis require authentication and various headers during the handshake process in order to upgrade to websocket. It was not supported by the vertx-component. So I added an uri param to set a map of headers that will be passed as header in the HTTP handshake request. It is important to note that it only works when the endpoint is a producer, or a consumer of a remote host (consumeAsClient true).

I added three tests : 
- One to test that handshakeHeaders are correctly binded in the configuration.
- One to test that handshakeHeaders are correctly passed in the HTTP handshake request when the endpoint is a producer.
- One to test that handshakeHeaders are correctly passed in the HTTP handshake request when the endpoint is a consumer of a remote host.

# Target

# Tracking

No JIRA issue found.


oscerd (approved):

davsclaus:
Using `Map` as UriParam is not soo tooling friendly. I wonder if it can be made to just a plain `String` where entires are separated by comma. 

Also I wonder should it match by exact value or would the presence of a header only be enough to match (ie match value = *)

alexisgra:
Thanks for the review! @davsclaus 
I also wondered about the Map. I'd seen maps in Uriparam before but it was quite rare. No problem, I'll make the change! It will be easier to use. 

For your comment about the match, I'm not sure I understand. Are you talking about matches in the tests? 

Thanks!

davsclaus:
This test
```
Map.of("Authorization", "Bearer", "ApiSign", "12345");
```

What if you have many ApiSign, do you need to add each of them, or would you just match if ApiSign header was present or match by regexp or something.

alexisgra:
Hi @davsclaus ,

When I changed the map to comma-separated `String` entries, I noticed a problem.
If you look at the http specification, you'll see that formats such as `header=value,header2=value2` or `header=value;header2=value2` don't work. Commas, semi-colons, colons and equals can be used in header values. 

Here are a few examples: 
`Date: Tue, 15 Nov 1994 08:12:31 GMT`
`Accept-Language: en-US,en;q=0.5`


So I think we have to choose between several possibilities: use `String` and don't support all header types (list, parameters...), use message headers as is the case in other HTTP components or keep a `Map`.

What do you think? Thanks for your help

alexisgra:
No, I don't have any real world example. We can make the choice to not support comma in handshake headers. I will add it in the doc.
However, I have real world example with `:`, `=` with the APIs I'm trying to integrate using `camel-vertx-websocket`. Typically on an `ApiTimestamp` and an `ApiSign`. 

Do I update with this format: `headerName,headerValue,headerName2,headerValue2`?

Thanks.

davsclaus:
Ah yeah lets use another approach with `multiValue = true` which allows to keep a `Map` type and users can provide entries for the map one at a time with a given prefix.

See for example QuartzEndpoint.java file how it does it

PR #13878, opened 2024-04-22
(chores) camel-core: removed redundant throws declarations (in tests)

orpiske:
It's a large one, but basically getting rid of unused or redundant throws declarations in tests, so we can get higher quality reports when analyzing the code (and reduce clutter in general).


oscerd (approved):

orpiske:
Looking good on my own CI. Let's merge.

PR #13884, opened 2024-04-22
CAMEL-20704 duplicate case blocks

dk2k:
I can agree that it's really a minor issue


davsclaus:
This is a generated source file so you should not edit it manually

// Generated by camel-yaml-dsl-maven-plugin - do NOT edit this file!

PR #13899, opened 2024-04-23
CAMEL-20707 Implemented isPresent() check for 2 optionals

dk2k:
CAMEL-20707 Implemented isPresent() check for 2 optionals instead of null check


orpiske (approved):
LGTM.

PR #13900, opened 2024-04-23
CAMEL-20708 added volatile keyword for double locking in JiraEndpoint

dk2k:
CAMEL-20708 added volatile keyword for double locking in JiraEndpoint


orpiske:
You could also remove the transient keyword.

orpiske (changes requested):
I don't think this is needed. The Java memory guarantees should ensure that all changes made after the lock are visible to all threads affected by it.

PR #13921, opened 2024-04-25
Reduced Cognitive complexity of the code.

anjaliInfy1822:
Reduced the cognitive complexity of the code, by reducing the if else and for loop nesting multiple times.

### DIFF
--- a/catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
+++ b/catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
@@ -232,8 +232,7 @@ public class DefaultCamelCatalog extends AbstractCachingCamelCatalog implements 
 
     @Override
     public List<String> findComponentNames() {
-        return cache(FIND_COMPONENT_NAMES,
-                () -> Stream.of(runtimeProvider.findComponentNames(), extraComponents.keySet())
+        return cache(FIND_COMPONENT_NAMES, () -> Stream.of(runtimeProvider.findComponentNames(), extraComponents.keySet())
                         .flatMap(Collection::stream)
                         .sorted()
                         .toList());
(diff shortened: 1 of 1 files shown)


catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
```diff
@@ -659,4 +654,15 @@ private String loadResource(String file) {
         }
     }
 
+    private ArtifactModel matchArtifactModelFromList(List<String> artifactModelList, String groupId, String artifactId,
+            String version) {
+        for (String name : artifactModelList) {
+            ArtifactModel<?> am = componentModel(name);
```
davsclaus, line 660: this is wrong, here is only component, the old had per different type

davsclaus (changes requested):

catalog/camel-catalog/src/main/java/org/apache/camel/catalog/DefaultCamelCatalog.java
anjaliInfy1822: Hi @davsclaus , sorry I missed that part. Would it be okay to have different function for each case?

davsclaus: This PR does not improve anything, the old code is fine, and also avoids backport problems.

If you want to help with Camel, then help with better things, than this stuff. Also it helps to get involved with Camel first than open your first PR out of "the blue".

anjaliInfy1822: @davsclaus  sure Thanks

orpiske: @anjaliInfy1822 My .2 cents: focus on the "easy" tickets and "help wanted" first to gain some experience with the code base before proceeding with things like this. 
As Claus said, get involved with the community (specially on the Zulip chat) to keep an eye on things we need help with.
Once you have a good grasp of the code base and know what to expect, then it's OK to focus on cleaning up things (we do have a lot of things to cleanup!!! But this one is not one of them).

PR #13965, opened 2024-04-26
CAMEL-20717 Throw created exception in TcpSocketConsumerRunnable

dk2k:
CAMEL-20717 Throw created exception in TcpSocketConsumerRunnable


davsclaus:
Ah I think the exception should actually not be thrown but just removed, as the idea is that in case of timeout the handle method is triggered, but the run method should keep running. So instead can you just remove the exception.

davsclaus (changes requested):
Ah I think the exception should actually not be thrown but just removed, as the idea is that in case of timeout the handle method is triggered, but the run method should keep running. So instead can you just remove the exception.

davsclaus (approved):

davsclaus:
You need to remove this line too

```
import org.apache.camel.component.mllp.MllpInvalidMessageException;
```

dk2k:
removed the import

oscerd (approved):

orpiske (approved):

PR #13967, opened 2024-04-28
CAMEL-19522: camel-http replace Thread.sleep in tests

3keNL:
Replaced occurences of Thread.sleep() with Awaitility. Also checked camel-http-common but it had no occurences


davsclaus (changes requested):
this is however wrong, as the sleep is not waiting for any condition, but wait a fix time

PR #14008, opened 2024-05-01
CAMEL-20680: reduced test execution time by 400sec by removing unnecessary producer.send 

jvsingh:
# Description

Test KafkaBreakOnFirstErrorSeekIssueIT consistently took over 460 seconds (most other integration tests in camel-kafka took around 50 seconds on an average). 
(These tests have a dependency on docker, so if its test images are not present on the build machine, the first time these are downloaded in about 200 seconds. i.e. the first time this test took 700 seconds to execute. But subsequent runs took 460 secons.)
In this change, I removed the unnecessary loop to push messages to partition 1. Each of 7 iterations of this loop was taking 60 seconds

Note that times reported above are for my development machine and could vary for other build machines. However, the proportional reduction in test execution time should be similar (around 80% reduction in execution time for repeat tests when docker images are already present locally) 

# Target

# Tracking


jvsingh:
To verify, run `mvn test -Dtest=org.apache.camel.component.kafka.integration.KafkaBreakOnFirstErrorSeekIssueIT`
before and after this change

oscerd (commented):
At first sight this looks good, but I'd like to have the feedback from @orpiske

orpiske:
Thanks and LGTM, but I need to test this manually, as these tests have been disabled on CI envs.
I will look at it today.

orpiske (approved):

PR #14063, opened 2024-05-06
Updated pom to skip build on Windows running on ARM

cziesman:
# Description

The `camel-salesforce` component cannot build on Windows running on ARM because there is no gRPC library available. In order to build successfully on Windows/ARM, this PR disables the build on that environment.

In general, adding profiles is discouraged, but since there is no way to disable a module via properties, adding profiles to the `camel-salesforce` parent pom was the only way to make this work.

# Target

# Tracking

- [ x] I checked that each commit in the pull request has a meaningful subject line and body.

- [ x] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes

`mvn clean install -DskipTests` fails on Windows/ARM due to an error when building the documentation. That error probably requires a separate JIRA issue.

### DIFF
--- a/components/camel-salesforce/pom.xml
+++ b/components/camel-salesforce/pom.xml
@@ -32,12 +32,6 @@
     <name>Camel :: Salesforce :: Parent</name>
     <description>Camel Salesforce parent</description>
 
-    <modules>
-        <module>camel-salesforce-component</module>
-        <module>camel-salesforce-codegen</module>
-        <module>camel-salesforce-maven-plugin</module>
-    </modules>
-    
     <properties>
         <salesforce.component.root>${project.basedir}</salesforce.component.root>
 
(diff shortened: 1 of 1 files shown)


gnodet:
@cziesman why the fact that there's no native library breaks the build ?
It should at most break when running tests, but the native library isn't really used at build time, is it ?

cziesman:
@gnodet This does currently fail at build time, as was noted in Jira (https://issues.apache.org/jira/browse/CAMEL-20706):

Protobuf compilation fails on Windows 11 on ARM because there is no protoc executable available for ARM.

```
Failed to execute goal org.xolstice.maven.plugins:protobuf-maven-plugin:0.6.1:compile (default) on project camel-salesforce: Unable to resolve artifact: Missing:
[ERROR] ----------
[ERROR] 1) com.google.protobuf:protoc:exe:windows-aarch_64:3.25.3

<snipped/>

[ERROR] 1 required artifact is missing.
```

davsclaus (approved):

orpiske (approved):
I think we might need this one. We already do similar things to other components whenever they have unfulfilled requirements on a few platforms (i.e: dlj on ppc64 and s390x, etc).

davsclaus:
@cziesman this PR did not work - it causes a normal build on osx and other systems to somehow skip salesforce also and affects the total build with missing files in catalog from salesforce etc. Reverted

cziesman:
@davsclaus That's very odd. I ran the build on my x86 MacBook and the Salesforce artifacts were generated. I wouldn't have submitted the PR otherwise. I am on PTO through Tuesday but I will revisit this when I return.

cziesman:
@davsclaus How are other components that are not built in specific environments excluded from the tooling? I ran this again on OSX on Intel and ARM, and the build passed. I ran it on Fedora 38 on ARM and it failed trying to build `camel-endpointdsl`.

PR #14065, opened 2024-05-07
Add @Serial annotation

gnodet:


davsclaus (approved):

orpiske:
@gnodet @davsclaus @oscerd sorry for my ignorance here ... serialization has never been my thing, so: is this still something we need?

orpiske:
@gnodet @davsclaus I wonder if we should just go ahead and remove all this serialization stuff. If it's something old, unused (and very likely unsafe), I don't see the point of having this in the code base.

orpiske:
> not as much anymore - we do not recommend serializing java objects and sending them over the wire - a few old components still have an option to do that like JMS with transferExchange=true|false

Oh, ok. Now I got it. I think we can't remove these without breaking backwards compatibility. 

Thanks for the explanation @gnodet and @davsclaus

PR #14066, opened 2024-05-07
Make EndpointRegistry extends Map<NormalizedEndpointUri, Endpoint>

gnodet:
Change the `CamelContext.getEndpointRegistry()` to return an `EndpointRegistry<NormalizedEndpointUri>` instead of an `EndpointRegistry<? extends ValueHolder<String>>`.  

This is a small incompatible change.

### DIFF
--- a/components/camel-rest/src/main/java/org/apache/camel/component/rest/DefaultRestRegistry.java
+++ b/components/camel-rest/src/main/java/org/apache/camel/component/rest/DefaultRestRegistry.java
@@ -33,7 +33,7 @@ import org.apache.camel.Service;
 import org.apache.camel.ServiceStatus;
 import org.apache.camel.StatefulService;
 import org.apache.camel.StaticService;
-import org.apache.camel.ValueHolder;
+import org.apache.camel.spi.NormalizedEndpointUri;
 import org.apache.camel.spi.RestConfiguration;
 import org.apache.camel.spi.RestRegistry;
 import org.apache.camel.support.LifecycleStrategySupport;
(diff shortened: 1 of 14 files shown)


davsclaus (approved):

davsclaus:
You can add a note about this API change in the Camel 4.7 upgrade page

davsclaus:
You likely need to update code at places like

camel-rest/src/main/java/org/apache/camel/component/rest/DefaultRestRegistry.java

davsclaus:
You can do a 

cd components
git grep getEndpointRegistry

davsclaus:
this PR needs to fix in the components as commented above

gnodet:
> this PR needs to fix in the components as commented above

In theory yes, in practice, no component is affected.  The reason is that the `K` generic type in the `EnpointRegistry` is only used on `keySet()` and `entrySet()` methods, which are not used, and components, when they need to access the registry do so without using an intermediate variable which could cause a cast compilation error.

I'm actually wondering if that parameter is needed at all, and while we're changing this method, it may be simpler to just make `EndpointRegistry` non parameterised and extends `Map<NormalizedEndpointUri, Endpoint>`. @davsclaus thoughts ?

davsclaus:
> > this PR needs to fix in the components as commented above
> 
> In theory yes, in practice, no component is affected. The reason is that the `K` generic type in the `EnpointRegistry` is only used on `keySet()` and `entrySet()` methods, which are not used, and components, when they need to access the registry do so without using an intermediate variable which could cause a cast compilation error.
> 
> I'm actually wondering if that parameter is needed at all, and while we're changing this method, it may be simpler to just make `EndpointRegistry` non parameterised and extends `Map<NormalizedEndpointUri, Endpoint>`. @davsclaus thoughts ?

Yes that is better as we dont need other kind of parameters at all

davsclaus (approved):

davsclaus:
org.apache.camel.issues.RecipientListUseOriginalMessageEndpointExceptionIssueTest.testRecipientListUseOriginalMessageIssue | 1.822s | FAILURE
-- | -- | --
org.apache.camel.builder.xml.XPathTest.testXPathSplitConcurrent | 1.763s | FAILURE

PR #14067, opened 2024-05-07
Do not use Optional as parameters or fields

gnodet:


orpiske (approved):

davsclaus (approved):

davsclaus:
maybe camel-spring-boot also needs to be updated (I dont know)

PR #14083, opened 2024-05-08
CAMEL-20722 - disabled on ppc64le due to consistent failure

jvsingh:
CAMEL-20722 - disabled on ppc64le due to consistent failure on ppc64le

# Description

CAMEL-20722 - disabled 6 tests on ppc64le due to consistent failure on ppc64le

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [CAMEL-20722](https://issues.apache.org/jira/browse/CAMEL-20722) filed for the change (usually before you start working on it).


davsclaus (approved):

davsclaus:
I assume there is no @DisabledOnOs so you need to do as you do

jvsingh:
> I assume there is no @DisabledOnOs so you need to do as you do

Actually there is a @DisabledOnOs - I just wasn't sure if 
`@DisabledOnOs(OS.AIX)` 
or ` @DisabledOnOs(value = {OS.AIX}, architectures = {"ppc64le"} )` 
 is precisely equivalent to disabling it on `ppc64le `
Whereas the approach I used has been used elsewhere in camel-kafka precisely to exclude `ppc64le `so I felt using the same approach might be safer

davsclaus:
ok

PR #14100, opened 2024-05-10
Revert "Updated pom to skip build on Windows running on ARM"

davsclaus:
Reverts apache/camel#14063


davsclaus:
This caused salesforce to be skipped on my osx and causes some source code build files to be missing and catalog and endpointdsl is changed. 

We need another kind of fix for this

PR #14144, opened 2024-05-14
Rerun failing tests

gnodet:


parent/pom.xml
```diff
@@ -3535,8 +3535,8 @@
                             <failIfNoTests>false</failIfNoTests>
                             <failIfNoSpecifiedTests>false</failIfNoSpecifiedTests>
                             <runOrder>alphabetical</runOrder>
-                            <!-- lets re-run the failed test one more time, just to be sure -->
-                            <rerunFailingTestsCount>0</rerunFailingTestsCount>
+                            <!-- lets re-run the failed test, just to be sure -->
+                            <rerunFailingTestsCount>2</rerunFailingTestsCount>
```
orpiske, line 3539: Just for the record: both GH actions and ASF already do this by default (2 for unit tests and 1 for integration tests).

gnodet: I don't see tests being rerun in the following failed builds however: https://github.com/apache/camel/actions/runs/8925211751/job/24513370969?pr=14042

gnodet: I don't think the `surefire.rerunFailingTestsCount` system property is used, because the plugin has a specific configuration set to 0 and the property is only used as a default.

If we want the system property to take precedence, we need to remove the line or use something like:
```
<rerunFailingTestsCount>${surefire.rerunFailingTestsCount}</rerunFailingTestsCount>
```
and eventually define it with a default value:
```
<properties>
  <surefire.rerunFailingTestsCount>2</surefire.rerunFailingTestsCount>
</properties>
```

orpiske: @gnodet oh, so you mean this one is overriding the property we set manually? OMG.

orpiske: BTW, it may be necessary to do the same in a few more places (and also for the failsafe plugin). I looked briefly at the code and there's a few more places where this is set. 

Good catch!!!

gnodet: > @gnodet oh, so you mean this one is overriding the property we set manually? OMG.

Yes, the properties used as default values for plugin configs are... defaults ;-)
So if the plugin configuration specifically sets a value, the property won't be used.

gnodet: > BTW, it may be necessary to do the same in a few more places (and also for the failsafe plugin). I looked briefly at the code and there's a few more places where this is set.
> 
> Good catch!!!

FWIW, the `rerunFailingTestsCount` is not explicitely set in the failsafe plugin config, so that one is not a problem and the `failsafe.rerunFailingTestsCount` system property should be honored.

davsclaus (approved):

parent/pom.xml
gnodet: @orpiske I don' really mind the default value, but I need the rerun to happen in GitHub.  So if you prefer, I can revert the default value to 0 but keep it in the properties.  Or just remove the line. It will have the same effect.
Or use a `camel.surefire.rerunFailingTestsCount` as we have for some other properties... but that will need to adjust all the scripts / GitHub actions.
If you're fine with the current state, I can simply merge :-)

orpiske: @gnodet I am fine with the current approach ... it will be great to have this flag working! Let's go ahead and merge it.

oscerd (approved):

PR #14178, opened 2024-05-17
CamelTestSupport: removal of wrong annotation + addition of method for @Override

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-20784

I remoced the annotation (`CamelQuarkusTest`) which does not exist and I moved the validation into a protected method. Child class is able to override the behavior if needed.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
+++ b/components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
@@ -343,7 +343,7 @@ public abstract class CamelTestSupport
     public void setUp() throws Exception {
         testStartHeader(getClass(), currentTestName);
 
-        ExtensionHelper.hasUnsupported(getClass());
+        assertUnsupported();
 
         if (isCreateCamelContextPerClass()) {
             createCamelContextPerClass();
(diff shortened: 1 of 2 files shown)


JiriOndrusek:
FYI @orpiske
Ideally I'd use 'claszz.isAssignableFrom(CamelQuarkusTestSupport.class' instead of moving the validation  into protected method. But because I can use only String name of the class, I'd need to find class by name and handle ClassNotFoundException. I see solution in the PR clearer for the users.
In case you disagree I'll refactor it differently.

orpiske (changes requested):
So, I'd like us to start decoupling this class. 

I am sorry to say it, but this class is horribly tightly couple and full of problems (i.e.: like mixing up being both an JUnit 5 Extension as well as a test case in itself. 

So, because of that, I would like us to not add more methods to its interface and, instead, work to slowly migrate those away (while giving time for the community to update code). 

That's why I'd marking this as request changes.

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
```diff
@@ -359,6 +359,10 @@ public void setUp() throws Exception {
         watch.restart();
     }
 
+    protected void assertUnsupported() {
+        ExtensionHelper.hasUnsupported(getClass());
+    }
+
```
orpiske, line 365: This is the problematic part I talk about on my comment.

orpiske:
@JiriOndrusek I think we can close this one, right?

orpiske:
This has become invalid in the context of CAMEL-20785 changes. Let's close it.

PR #14222, opened 2024-05-22
Another set of code related warnings fixes

gnodet:
* Fix visibility issues
* Fix busy-wait loop issues

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/aggregate/AggregateProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/aggregate/AggregateProcessor.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.AbstractQueuedSynchronizer;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -124,6 +125,7 @@ public class AggregateProcessor extends AsyncProcessorSupport
     private Map<String, String> closedCorrelationKeys;
     private final Set<String> batchConsumerCorrelationKeys = new ConcurrentSkipListSet<>();
     private final Set<String> inProgressCompleteExchanges = ConcurrentHashMap.newKeySet();
+    private final WaitableInteger inProgressCount = new WaitableInteger();
     private final Set<String> unconfirmedCompleteExchanges = ConcurrentHashMap.newKeySet();
     private final Set<String> inProgressCompleteExchangesForRecoveryTask = ConcurrentHashMap.newKeySet();
     private final Map<String, RedeliveryData> redeliveryState = new ConcurrentHashMap<>();
@@ -836,6 +838,7 @@ public class AggregateProcessor extends AsyncProcessorSupport
         LOG.debug("Aggregation complete for correlation key {} sending aggregated exchange: {}", key, exchange);
 
         // add this as in progress before we submit the task
+        inProgressCount.increment();
         inProgressCompleteExchanges.add(exchange.getExchangeId());
         if (recoveryInProgress.get()) {
             inProgressCompleteExchangesForRecoveryTask.add(exchange.getExchangeId());
@@ -1219,6 +1222,7 @@ public class AggregateProcessor extends AsyncProcessorSupport
 
             // must remember to remove in progress when we failed
             inProgressCompleteExchanges.remove(exchangeId);
+            inProgressCount.decrement();
             // do not remove redelivery state as we need it when we redeliver again later
         }
 
@@ -1244,6 +1248,7 @@ public class AggregateProcessor extends AsyncProcessorSupport
             } finally {
                 // must remember to remove in progress when we are complete
                 inProgressCompleteExchanges.remove(exchangeId);
+                inProgressCount.decrement();
             }
         }
 
@@ -1698,15 +1703,14 @@ public class AggregateProcessor extends AsyncProcessorSupport
         int expected = forceCompletionOfAllGroups();
 
         StopWatch watch = new StopWatch();
-        while (!inProgressCompleteExchanges.isEmpty()) {
+        if (!inProgressCompleteExchanges.isEmpty()) {
             LOG.trace("Waiting for {} inflight exchanges to complete", getInProgressCompleteExchanges());
             try {
-                Thread.sleep(100);
+                inProgressCount.await();
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
                 // break out as we got interrupted such as the JVM terminating
                 LOG.warn("Interrupted while waiting for {} inflight exchanges to complete.", getInProgressCompleteExchanges());
-                break;
             }
         }
 
@@ -1723,6 +1727,7 @@ public class AggregateProcessor extends AsyncProcessorSupport
 
         // cleanup when shutting down
         inProgressCompleteExchanges.clear();
+        inProgressCount.reset();
 
         if (shutdownExecutorService) {
             camelContext.getExecutorServiceManager().shutdownNow(executorService);
@@ -1875,6 +1880,47 @@ public class AggregateProcessor extends AsyncProcessorSupport
             LOG.debug("Forcing discarding of all groups with {} exchanges", total);
         }
         return total;
+    }
+
+    /**
+     * Synchronization class to avoid busy-loop when waiting for exchanges to be processed during shutdown.
+     */
+    protected static final class WaitableInteger extends AbstractQueuedSynchronizer {
+
+        // await for this integer to be equal to zero
+        public void await() throws InterruptedException {
+            acquireSharedInterruptibly(0); // the arg is not used, see below
+        }
+
+        // decrement the integer
+        public void decrement() {
+            releaseShared(-1);
+        }
+
+        // increment the integer
+        public void increment() {
+            releaseShared(1);
+        }
+
+        // reset the integer to zero, this call won't trigger threads blocked on an await() call
+        public void reset() {
+            setState(0);
+        }
+
+        // the arg is passed through from acquireSharedInterruptibly, but not used
+        protected int tryAcquireShared(int unused) {
+            return (getState() == 0) ? 1 : -1;
+        }
+
+        // called from releaseShared, i.e. from increment() and decrement()
+        protected boolean tryReleaseShared(int releases) {
+            for (;;) {
+                int c = getState();
+                int nextc = c + releases;
+                if (compareAndSetState(c, nextc))
+                    return nextc == 0;
+            }
+        }
     }
 
 }
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java
@@ -16,7 +16,11 @@
  */
 package org.apache.camel.processor.resequencer;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Timer;
+import java.util.concurrent.CountDownLatch;
+import java.util.function.Predicate;
 
 import org.apache.camel.util.concurrent.ThreadHelper;
 
@@ -79,6 +83,12 @@ public class ResequencerEngine<E> {
     private Boolean rejectOld;
 
     /**
+     * List containing wait conditions to be evaluated whenever the sequence is modified. Access to this field should be
+     * done inside a {@code synchronized(this)} block.
+     */
+    private Map<CountDownLatch, Predicate<Sequence<?>>> waitConditions = new HashMap<>();
+
+    /**
      * Creates a new resequencer instance with a default timeout of 2000 milliseconds.
      *
      * @param comparator a sequence element comparator.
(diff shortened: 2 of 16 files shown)


oscerd (approved):

gnodet:
There's another busy-wait loop issue in `org.apache.camel.support.Task`.
This is usually not a problem when performing non trivial tasks and re-attempting again in case of a failure, such as trying to connect to a remote server every 5 seconds.  However, this package is also used to wait for trivial conditions, such as in the `ServicePool` where [a foreground task is used to wait for a service to be started every 5 milliseconds](https://github.com/apache/camel/blob/740c3b1a0235356346e8ee9ca38d0596513ca425/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java#L139).  I think this is an abuse of the framework.  For such cases,  I think an alternative strategy should be provided, using an implementation of `java.util.concurrent.Condition` that can be waited for.

core/camel-core-processor/src/main/java/org/apache/camel/processor/aggregate/AggregateProcessor.java
```diff
@@ -1877,4 +1882,44 @@ public int forceDiscardingOfAllGroups() {
...
+        public void increment() {
+            releaseShared(1);
+        }
+
+        // reset the integer to zero, this call won't trigger threads blocked on an await() call
+        public void reset() {
+            setState(0);
+        }
+
+        protected int tryAcquireShared(int acquires) {
```
lburgazzoli, line 1910: the `acquires` parameters seems to be unused, don't know if that's expected but is so a small comment could help understanding why.

core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java
```diff
@@ -110,6 +120,30 @@ public synchronized int size() {
         return sequence.size();
     }
 
+    public void waitUntil(Predicate<Sequence<?>> pred) throws InterruptedException {
+        CountDownLatch latch;
+        synchronized (this) {
```
lburgazzoli, line 125: _nitpick_

I don't know how exactly this class is used so this may no be an issue, but it is usually a good idea not to synchronize on the instance as you may get some side effects (got bitten in the past, so I noticed it), like the following code would end up in a deadlock:

```java
 var re = enw ResequencerEngine<>()
 syncrronized(re) {
     re.waitUntil(s -> { ... })
 }
 ```

**Comment by gnodet** [comment]
> There's another busy-wait loop issue in `org.apache.camel.support.Task`. This is usually not a problem when performing non trivial tasks and re-attempting again in case of a failure, such as trying to connect to a remote server every 5 seconds. However, this package is also used to wait for trivial conditions, such as in the `ServicePool` where [a foreground task is used to wait for a service to be started every 5 milliseconds](https://github.com/apache/camel/blob/740c3b1a0235356346e8ee9ca38d0596513ca425/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java#L139). I think this is an abuse of the framework. For such cases, I think an alternative strategy should be provided, using an implementation of `java.util.concurrent.Condition` that can be waited for.

Upon further investigation, the potential problem comes from this commit:
 https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a
 I'm not sure how this is supposed to work: services usually inherit from `ServiceSupport` class where starting the service is a blocking operation.  When the service is created, we call `addService` with `forceStart` set to true.  My understanding is that returning from this call, the service will either be started, or the start will be deferred, but I don't really see how it can be in the `STARTING` state.  @orpiske ?

File: core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java (hunk 2 above)
Comment (gnodet) [reply core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java:134]: You're right.  I've added some comments.  However, I'm not sure there's a work around.  The conditions need to be evaluated while holding a lock on the resequencer engine, so if that lock is held by another thread, it WILL deadlock for sure.

Comment (gnodet) [reply core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java:134]: The conditions are evaluated when `insert()` or `deliverNext()` is called.  Those are synchronised methods.  We do synchronize inside `evaluateConditions`, but this is not really required, as the caller already holds the lock (it's more to be coherent with the `waitUntil`).  So the whole thing already deadlocks if the lock is held forever by another thread.  This won't change the behaviour.

**Comment by orpiske** [comment]
> > There's another busy-wait loop issue in `org.apache.camel.support.Task`. This is usually not a problem when performing non trivial tasks and re-attempting again in case of a failure, such as trying to connect to a remote server every 5 seconds. However, this package is also used to wait for trivial conditions, such as in the `ServicePool` where [a foreground task is used to wait for a service to be started every 5 milliseconds](https://github.com/apache/camel/blob/740c3b1a0235356346e8ee9ca38d0596513ca425/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java#L139). I think this is an abuse of the framework. For such cases, I think an alternative strategy should be provided, using an implementation of `java.util.concurrent.Condition` that can be waited for.
> 
> Upon further investigation, the potential problem comes from this commit: [07b9c68](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a) I'm not sure how this is supposed to work: services usually inherit from `ServiceSupport` class where starting the service is a blocking operation. When the service is created, we call `addService` with `forceStart` set to true. My understanding is that returning from this call, the service will either be started, or the start will be deferred, but I don't really see how it can be in the `STARTING` state. @orpiske ?

The key problem I was avoiding on [07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a), was the  `if (producer instanceof StatefulService) {` which caused a [type check miss](https://redhatperf.github.io/post/type-check-scalability-issue/) under high load. 

The STARTING state check had been [there before](https://github.com/apache/camel/commit/e653b815839be4fb5b1146a9fd8312a15d997318). Looking at the history of this change, it seems that the STARTING state check is related to [CAMEL-14048](https://issues.apache.org/jira/browse/CAMEL-14048) and a problem with Netty.

File: core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java (hunk 2 above)
Comment (lburgazzoli) [reply core/camel-core-processor/src/main/java/org/apache/camel/processor/resequencer/ResequencerEngine.java:134]: I don't know if applicable but a common patter is to have a _lock_ instance and lock on that

```java
private final Object lock = new Object()

public void waitUntil(Predicate<Sequence<?>> pred) throws InterruptedException {
    synchronized(lock) { ... }
}
```

but this is really a nitpick, so as you explained the issue may be also happen somewhere else

gnodet: Yes, I agree to the general rule you outline.  However, I did not want to rewrite the whole resequenced engine synchronisation, so this PR aims to integrate with the current way.  I'd suggest to use another PR to rework the synchronisation of the `ResequencerEngine` if needed.

orpiske:
@gnodet it would probably be good to run a full test on the code before merging. I've triggered a build on my own CI to see how it goes.

gnodet:
> > > There's another busy-wait loop issue in `org.apache.camel.support.Task`. This is usually not a problem when performing non trivial tasks and re-attempting again in case of a failure, such as trying to connect to a remote server every 5 seconds. However, this package is also used to wait for trivial conditions, such as in the `ServicePool` where [a foreground task is used to wait for a service to be started every 5 milliseconds](https://github.com/apache/camel/blob/740c3b1a0235356346e8ee9ca38d0596513ca425/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java#L139). I think this is an abuse of the framework. For such cases, I think an alternative strategy should be provided, using an implementation of `java.util.concurrent.Condition` that can be waited for.
> > 
> > 
> > Upon further investigation, the potential problem comes from this commit: [07b9c68](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a) I'm not sure how this is supposed to work: services usually inherit from `ServiceSupport` class where starting the service is a blocking operation. When the service is created, we call `addService` with `forceStart` set to true. My understanding is that returning from this call, the service will either be started, or the start will be deferred, but I don't really see how it can be in the `STARTING` state. @orpiske ?
> 
> The key problem I was avoiding on [07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a), was the `if (producer instanceof StatefulService) {` which caused a [type check miss](https://redhatperf.github.io/post/type-check-scalability-issue/) under high load.
> 
> The STARTING state check had been [there before](https://github.com/apache/camel/commit/e653b815839be4fb5b1146a9fd8312a15d997318). Looking at the history of this change, it seems that the STARTING state check is related to [CAMEL-14048](https://issues.apache.org/jira/browse/CAMEL-14048) [and a problem with Netty](https://github.com/apache/camel/commit/0f9f8790cccf3e45af4fe9c46ec030c5e63d2083).

Thx for the pointers.  I've had another look, I think the fix https://github.com/apache/camel/commit/0f9f8790cccf3e45af4fe9c46ec030c5e63d2083 is wrong and the problem was actually fixed correctly by https://github.com/apache/camel/commit/4c8c071fba493dbf41f09103712036805133a5a8.  This commit makes sure that the double-check locking only see a fully started service.

I'll add a commit to remove that part of the code.

gnodet:
> @gnodet it would probably be good to run a full test on the code before merging. I've triggered a build on my own CI to see how it goes.

Agreed.   The PR template mentions `build-all` but that's not an existing label,  I've seen `incremental-build-all` however. Is that the correct one ? We need to fix the template btw...

gnodet:
> > > > There's another busy-wait loop issue in `org.apache.camel.support.Task`. This is usually not a problem when performing non trivial tasks and re-attempting again in case of a failure, such as trying to connect to a remote server every 5 seconds. However, this package is also used to wait for trivial conditions, such as in the `ServicePool` where [a foreground task is used to wait for a service to be started every 5 milliseconds](https://github.com/apache/camel/blob/740c3b1a0235356346e8ee9ca38d0596513ca425/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java#L139). I think this is an abuse of the framework. For such cases, I think an alternative strategy should be provided, using an implementation of `java.util.concurrent.Condition` that can be waited for.
> > > 
> > > 
> > > Upon further investigation, the potential problem comes from this commit: [07b9c68](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a) I'm not sure how this is supposed to work: services usually inherit from `ServiceSupport` class where starting the service is a blocking operation. When the service is created, we call `addService` with `forceStart` set to true. My understanding is that returning from this call, the service will either be started, or the start will be deferred, but I don't really see how it can be in the `STARTING` state. @orpiske ?
> > 
> > 
> > The key problem I was avoiding on [07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a](https://github.com/apache/camel/commit/07b9c68ea2a9a70cf3a3f42c1045fb9fd0acbe6a), was the `if (producer instanceof StatefulService) {` which caused a [type check miss](https://redhatperf.github.io/post/type-check-scalability-issue/) under high load.
> > The STARTING state check had been [there before](https://github.com/apache/camel/commit/e653b815839be4fb5b1146a9fd8312a15d997318). Looking at the history of this change, it seems that the STARTING state check is related to [CAMEL-14048](https://issues.apache.org/jira/browse/CAMEL-14048) [and a problem with Netty](https://github.com/apache/camel/commit/0f9f8790cccf3e45af4fe9c46ec030c5e63d2083).
> 
> Thx for the pointers. I've had another look, I think the fix [0f9f879](https://github.com/apache/camel/commit/0f9f8790cccf3e45af4fe9c46ec030c5e63d2083) is wrong and the problem was actually fixed correctly by [4c8c071](https://github.com/apache/camel/commit/4c8c071fba493dbf41f09103712036805133a5a8). This commit makes sure that the double-check locking only see a fully started service.
> 
> I'll add a commit to remove that part of the code.

Pushed a commit.  However I think the `ServicePool` still has some synchronization issues.  We use a `BlockingQueue`, but we're holding a lock to access it, so that's quite unexpected...  That's for another PR though, maybe with the `ResequencerEngine` concurrency to avoid locking on the object itself.

orpiske:
> > @gnodet it would probably be good to run a full test on the code before merging. I've triggered a build on my own CI to see how it goes.
> 
> Agreed. The PR template mentions `build-all` but that's not an existing label, I've seen `incremental-build-all` however. Is that the correct one ? We need to fix the template btw...

I think these labels are outdated. IIRC, we never managed to run full builds on GH because it takes too long and causes the action runner to abort after some time.

For now, the only way to run a full test is by running the tests manually (or, semi automatically, as I do on my environment).

orpiske:
First run (before the last commit pushed) had only 1 failure on ` org.apache.camel.component.infinispan.remote.InfinispanRemoteProducerIT.replaceAValueByKeyAsyncWithOldValue`. 

I have a new build running with the latest commit.

orpiske (approved):
LGTM. Tested on my own CI and there were only 2 failures which are known flakies:

* org.apache.camel.component.file.FileSortByExpressionTest.testSortFiles
* org.apache.camel.component.rocketmq.RocketMQRouteIT.testSimpleRoute

PR #14238, opened 2024-05-24

fanyang:
# Description

# Target

# Tracking


components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
```diff
@@ -47,11 +53,28 @@ public Kinesis2Endpoint getEndpoint() {
...
+            this.exceptionThrownByPutRecord = null;
+            throw exceptionThrownByPutRecord;
+        }
+
         PutRecordRequest request = createRequest(exchange);
-        PutRecordResponse putRecordResult = connection.getClient(getEndpoint()).putRecord(request);
-        Message message = getMessageForResponse(exchange);
-        message.setHeader(Kinesis2Constants.SEQUENCE_NUMBER, putRecordResult.sequenceNumber());
-        message.setHeader(Kinesis2Constants.SHARD_ID, putRecordResult.shardId());
+        if (getEndpoint().getConfiguration().isAsyncClient()) {
```
davsclaus, line 63: This is wrong, as you cannot use async client in a camel producer that is fully synchronous. Camel needs to be aware that this is async and route accordingly - this kind of work is complex and takes time to understand how to do it.

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
```diff
@@ -47,11 +53,28 @@ public Kinesis2Endpoint getEndpoint() {
 
     @Override
     public void process(Exchange exchange) throws Exception {
+        Exception exceptionThrownByPutRecord = this.exceptionThrownByPutRecord;
```
davsclaus, line 56: This is not as good as concurrent calls could lead to thread safety issues - Use AtomicInstance or what the name is.

Also its not understandable why a previous exception should cause the next to fail. There is no batch stuff here. 

I think you should seperate the items you try to help with

- batching (make this work with current client)
- async (much harder - wait with this as its not needed asap)

davsclaus (changes requested):

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
fanyang: Thanks for the comments. Keep using synchronous client.

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
fanyang: Hi Claus, thanks for the comments. This PR keeps using current client and supports sending batch. 
Since sending batch already has a good throughput, and as you said async process is complex, It seems that we don't really need an async client.

PR #14241, opened 2024-05-26

fanyang:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
@@ -75,7 +75,6 @@ public class Kinesis2Producer extends DefaultProducer {
     private void sendSingleRecord(Exchange exchange) {
         PutRecordRequest request = createRequest(exchange);
         PutRecordResponse putRecordResult = connection.getClient(getEndpoint()).putRecord(request);
-        LOG.trace("Sent 1 record.");
         Message message = exchange.getMessage();
         message.setHeader(Kinesis2Constants.SEQUENCE_NUMBER, putRecordResult.sequenceNumber());
         message.setHeader(Kinesis2Constants.SHARD_ID, putRecordResult.shardId());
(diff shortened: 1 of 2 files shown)


components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
```diff
@@ -69,8 +96,37 @@ private PutRecordRequest createRequest(Exchange exchange) {
         return putRecordRequest.build();
     }
 
-    public static Message getMessageForResponse(final Exchange exchange) {
-        return exchange.getMessage();
+    private PutRecordsRequestEntry createRequestEntry(Exchange exchange) {
```
oscerd, line 99: There is no security about the presence of the header.

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
```diff
@@ -47,9 +59,24 @@ public Kinesis2Endpoint getEndpoint() {
...
+        if (batchComplete) {
+            flushRequestBatch();
+        }
+    }
+
+    private void sendSingleRecord(Exchange exchange) {
         PutRecordRequest request = createRequest(exchange);
         PutRecordResponse putRecordResult = connection.getClient(getEndpoint()).putRecord(request);
-        Message message = getMessageForResponse(exchange);
+        LOG.trace("Sent 1 record.");
```
oscerd, line 78: I think this is useless without details.

oscerd (changes requested):
I think the batch support could be done really better than this. Maybe we could think about different methods or at least better architecture.

davsclaus:
This will not work with concurrent batch

from a to azure
from b to azure

the array list should likely be a concurrent list, and when you flush then you would need some kind of synchronization to avoid a race condition where the list is clear and a message is added at the same time, could lead to lost message.

Also this batching is okay for mixing concurrent messages in the same batch as they are also individual records being send together to azure - to speedup this.

There is no upper size on azure, what if the batch has 50.000 elemenents ?

fanyang:
Create another PR to support concurrency. https://github.com/apache/camel/pull/14606
Close this one.

PR #14298, opened 2024-06-02
CAMEL-20680 - fix root cause of slow publish to partition 1 + CAMEL-20722 re-enable

jvsingh:
CAMEL-20680 - fixed root cause of slow publish to partition 1. 
- Added a wait in `setupTopics ()` to ensure that required number of partitions are actually created
- Restored the loop in  `publishMessagesToKafka()` that publishes to partition 1
- Change the position of errored record from 5 to 6 _(to ensure that something is actually read from partition 1 as well)_
After these, the test continues to execute in under 45 sec even after adding back the loop that sends to partition 1  _(the long delay in `publishMessagesToKafka` when CAMEL-20680  was created was because partition 1 was being used before it was even created!)_

CAMEL-20722 - re-enabled test for CI, as the symptom from build 96 **might** go away as a side effect of this

# Description

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

jvsingh:
I have no local changes after mvn verify. 
Only 1 source file changed in this commit

jvsingh:
Could the error be related to this? https://github.com/apache/camel/actions/runs/9332185585?
I have no change in the source file flagged up in the error

davsclaus:
If you rebase on latest code from main then it should work

jvsingh:
> If you rebase on latest code from main then it should work

Done

PR #14300, opened 2024-06-03
CAMEL-20826: camel-package-maven-plugin:generate not creating reproducible jandex.idx files.

chirino:
# Description

We needed to sort input list before feeding it into jandex.

# Target

# Tracking
- [X] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue]

https://issues.apache.org/jira/browse/CAMEL-20826


chirino:
The `camel-servicenow-maven-plugin` is the only jar that contains a jandex.idx that does not reproduce with this change.  Wonder if we can disable the package plugin for that jar?

chirino:
The camel-servicenow-maven-plugin now also skips execution of the camel-package pluign.

davsclaus (approved):

PR #14313, opened 2024-06-03
CAMEL-20785: create a new transient context manager

orpiske:
- adjust the code to use the new transient context manager
- split responsibilities from the test support class


orpiske:
Continues to split responsibilities ... laying the foundation for transforming these in a full-blown JUnit 5 extension on 4.8.0 or 4.9.0

PR #14314, opened 2024-06-04
CAMEL-20660: Use synchronous Azure ServiceBusSenderClient

dylanpiergies:
# Description

The implementation of CAMEL-20266 introduced an issue whereby the async client would attempt to execute the `AsyncCallback#done` method. When the consumer is a `ServiceBusConsumer`, this would cause the callback invocation to fail because it results in `Mono#block` being invoked on a Reactor thread.

Switching the `ServiceBusProducer` to synchronous operation seems to be the most straightforward way to address this issue. It is unknown whether this issue can also impact other consumers.

# Target

# Tracking


dylanpiergies:
Looks like there's some inconsistency with source generation. Executing `./mvnw clean install -DskipTests` does not leave any uncommitted files. Trying with `./etc/scripts/regen.sh`.

davsclaus (approved):

PR #14324, opened 2024-06-04
chore: Use actions/upload-artifact v4

apupier:
it will avoid this message:

Deprecation notice: v1, v2, and v3 of the artifact actions The following artifacts were uploaded using a version of actions/upload-artifact that is scheduled for deprecation: "build.log", "tests.log".
Please update your workflow to use v4 of the artifact actions. Learn more:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

It wasn't updated automatically by dependabot because it is inside reusable actions.
It doesn't seem that this action are used in matrix jobs that can upload several times the same artifact so we should be fine just by upgrading the number

# Description

# Target

# Tracking

### DIFF
--- a/.github/actions/component-test/action.yaml
+++ b/.github/actions/component-test/action.yaml
@@ -44,7 +44,7 @@ runs:
         FAST_BUILD: "true"
         LOG_FILE: build.log
     - name: archive logs
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       if: always()
       with:
         name: build.log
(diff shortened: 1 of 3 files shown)


apupier:
/component-test jetty

to test that classic case is still working with component-test build

davsclaus (approved):

apupier:
argh seems I missed some matrix job which are using these reusable actions (and which means that when playing a PR, it is no tusing the reusable github action from this same PR) :

https://github.com/apache/camel/actions/runs/9364653659/job/25778098025?pr=14327#step:9:73

Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

PR #14344, opened 2024-06-04
CAMEL-20722 new log entry to log the exception

jvsingh:
# Description

This particular exception is not getting logged because another exception arises in the CATCH block. 
So, this log entry will be useful to determine the particular symptom seen in builds 96, 113 (only occurs in CI but not in local)
The main test case for KafkaBreakOnFirstErrorSeekIssueIT actually works fine except for this random symptom seen only on CI. 

# Target

# Tracking


test-infra/camel-test-infra-kafka/src/test/java/org/apache/camel/test/infra/kafka/services/KafkaService.java
```diff
@@ -42,6 +42,7 @@ default void beforeAll(ExtensionContext extensionContext) {
             initialize();
         } catch (Exception e) {
             Logger log = LoggerFactory.getLogger(KafkaService.class);
+            log.error(" Unexpected error in {}: {} {} ", this.getClass().getSimpleName(), e.getMessage(), e.getCause());
```
davsclaus, line 45: Should whitespace be deleted

orpiske (approved):

orpiske:
Functionality-wise, looks OK. Just needs the fixing that @davsclaus mentioned and it should be ready to merge.

davsclaus (approved):

PR #14384, opened 2024-06-05
CAMEL-20785: provide a way to create context managers when needed

orpiske:
This is to make CEQ's life easier with these refactorings.


orpiske:
c/c: @JiriOndrusek

PR #14408, opened 2024-06-06
CAMEL-20843: further split the CamelTestSupport responsibilities

orpiske:
- Move the legacy test/context setup methods to the AbstractTestSupport
- Moved most of the helper logic to the TestSupport
- Cleaned up one affected test
- Added documentation and comments
- Deprecated a few more methods


orpiske:
Passed without problems on the CI. Let's merge.

PR #14424, opened 2024-06-07
CAMEL-20850: camel-core - compress changes in SimpleLRUCache

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-20850

## Motivation

When the queue of changes is full, some entries can be evicted while the cache is not, which is not expected.

## Modifications:

* Compress the changes by removing duplicates before the eviction
* Add a minimum size for the queue of changes

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/support/cache/SimpleLRUCacheTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/support/cache/SimpleLRUCacheTest.java
@@ -19,11 +19,17 @@ package org.apache.camel.support.cache;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
+import static org.apache.camel.support.cache.SimpleLRUCache.MINIMUM_QUEUE_SIZE;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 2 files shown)


orpiske (approved):

oscerd (approved):

core/camel-support/src/main/java/org/apache/camel/support/cache/SimpleLRUCache.java
```diff
@@ -248,6 +270,24 @@ private Entry<K, V> nextOldestChange() {
         return oldest;
     }
 
+    /**
+     * Removes duplicates from the queue of changes.
+     */
+    private void compressChanges() {
```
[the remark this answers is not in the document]
essobedo: Indeed, it looks like there is a time window during which changes could be lost, let me try to improve this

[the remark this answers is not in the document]
essobedo: I added 2 more tests and it seems to behave as expected

essobedo:
@gnodet Hi, any remarks/comments before I merge it?

core/camel-core/src/test/java/org/apache/camel/support/cache/SimpleLRUCacheTest.java
```diff
@@ -287,4 +280,102 @@ void replaceWithOldValue() {
...
+                try {
+                    for (int j = 0; j < totalKeysPerThread; j++) {
+                        cache.put(threadId + "-" + j, Integer.toString(j));
+                    }
+                } finally {
+                    latch.countDown();
+                }
+            }).start();
+        }
+        latch.await();
```
davsclaus, line 346: Maybe add a timeout on the latch if something bad happens then it would wait forever

essobedo: No, it cannot wait forever because `latch.countDown()` is called in a finally block

davsclaus: ok

davsclaus (approved):

PR #14449, opened 2024-06-10
ParserCommon tests

AntonHrbnts:
# Description

# Target

# Tracking

### DIFF
--- a/catalog/camel-route-parser/src/main/java/org/apache/camel/parser/helper/ParserCommon.java
+++ b/catalog/camel-route-parser/src/main/java/org/apache/camel/parser/helper/ParserCommon.java
@@ -118,8 +118,8 @@ public final class ParserCommon {
     public static boolean isNumericOperator(JavaClassSource clazz, Block block, Expression expression) {
         if (expression instanceof NumberLiteral) {
             return true;
-        } else if (expression instanceof SimpleName) {
-            FieldSource<JavaClassSource> field = getField(clazz, block, (SimpleName) expression);
+        } else if (expression instanceof SimpleName simpleName) {
+            FieldSource<JavaClassSource> field = getField(clazz, block, simpleName);
             if (field != null) {
                 return field.getType().isType("int") || field.getType().isType("long")
                         || field.getType().isType("Integer") || field.getType().isType("Long");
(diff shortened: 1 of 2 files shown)


davsclaus (changes requested):
Sorry but this test is complex to understand and would just cause pain for us to maintain in the future.

orpiske (changes requested):
As @davsclaus said: the test code is a bit too complex. 

My suggestion: try to break the test in smaller pieces. For instance, separate the test setup part (i.e.: setting up mocks, expectations, etc) from the test result verification (asserting the result, logging when necessary, etc). 

Additionally, try to document to use JUnit's `@DisplayName` annotation to describe what the test does and what is the expected result (i.e.; `Tests that the code can consume null values and return an empty Box as expected`).

catalog/camel-route-parser/src/main/java/org/apache/camel/parser/helper/ParserCommon.java
```diff
@@ -118,8 +118,8 @@ static FieldSource<JavaClassSource> getField(JavaClassSource clazz, Block block,
     public static boolean isNumericOperator(JavaClassSource clazz, Block block, Expression expression) {
         if (expression instanceof NumberLiteral) {
             return true;
-        } else if (expression instanceof SimpleName) {
-            FieldSource<JavaClassSource> field = getField(clazz, block, (SimpleName) expression);
+        } else if (expression instanceof SimpleName simpleName) {
+            FieldSource<JavaClassSource> field = getField(clazz, block, simpleName);
```
orpiske, line 122: Also, this part here implies that there is something more other than tests. Maybe adjust the PR description, commit message and/or reference a Jira ticket if you are providing a fix.

catalog/camel-route-parser/src/test/java/org/apache/camel/parser/helper/ParserCommonTest.java
```diff
@@ -0,0 +1,118 @@
...
+            parserCommonMockedStatic.when(() -> ParserCommon.getField(clazz, block, expression)).thenReturn(fieldSource);
+            parserCommonMockedStatic.when(() -> ParserCommon.isNumericOperator(clazz, block, expression)).thenCallRealMethod();
+
+            assertFalse(ParserCommon.isNumericOperator(clazz, block, expression));
+
+            parserCommonMockedStatic.verify(() -> ParserCommon.getField(clazz, block, expression));
+            parserCommonMockedStatic.verify(() -> ParserCommon.isNumericOperator(clazz, block, expression));
+            parserCommonMockedStatic.verifyNoMoreInteractions();
+        }
+    }
```
orpiske, line 117: I think I would probably move these 2 tests to a separate file. Then, I'd separate the test setup from the verification part and would use the `@DisplayName` annotation to make the expectations clearer.

davsclaus:
closing inactive PR

PR #14458, opened 2024-06-11
(chores) ci: redirect outputs in script

essobedo:
## Motivation

The redirection of standard outputs is not supported by GitHub jobs so we don't get the log of builds.

## Modifications:

* Move the redirection of the standard outputs in the build script


oscerd (approved):

apupier (approved):

essobedo:
The log files are now prevent which validates this PR so let's merge it

PR #14480, opened 2024-06-11
fix(sql-component): Fix sql component adoc

lordrip:
# Description
Currently, the `body` reference in the sql-component `.adoc` file is breaking the camel-website build.

Related PR: https://github.com/apache/camel-website/pull/1190

### DIFF
--- a/components/camel-sql/src/main/docs/sql-component.adoc
+++ b/components/camel-sql/src/main/docs/sql-component.adoc
@@ -357,7 +357,7 @@ from("direct:query")
     .to("mock:query");
 ----
 
-If the dynamic list of values is stored in the message body, you can use `(:#in:$+{body}+)` to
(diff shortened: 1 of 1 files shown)


components/camel-sql/src/main/docs/sql-component.adoc
```diff
@@ -357,7 +357,7 @@ from("direct:query")
     .to("mock:query");
 ----
 
-If the dynamic list of values is stored in the message body, you can use `(:#in:$+{body}+)` to
```
lhein, line 360: If I understand the AsciiDoc (https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#escaping-substitutions) docs correctly, it should be:

```
`(:#in:$++{body}++)`
```

No?

PR #14482, opened 2024-06-11
CAMEL-20850: camel-core - compress changes in SimpleLRUCache

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-20850 in Camel 4.4

## Motivation

When the queue of changes is full, some entries can be evicted while the cache is not, which is not expected.

## Modifications:

* Compress the changes by removing duplicates before the eviction
* Add a minimum size for the queue of changes


essobedo:
The failure is not related so let's merge it

PR #14489, opened 2024-06-12
CAMEL-20838: pre-work (avoid using APIs soon-to-be restricted)

orpiske:


orpiske:
Getting very close to complete this one.

PR #14531, opened 2024-06-14
CAMEL-20800: Use cloudant instead of LightCouch

Croway:
# Description

Remove LightCouch since the library is no longer maintained in favour of cloudant. The code can be improved (a lot), I tried my best to keep the component backward-compatible (Gson usage instead of cloudant Document mostly), I think it is worth to deprecate the actual API and refactor the component to better use cloudant.

# Target

# Tracking


components/camel-couchdb/pom.xml
```diff
@@ -47,16 +47,9 @@
...
-        </dependency>
-        <!-- prefer to use the httpclient version used by Camel components to align
-            the version -->
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>${httpclient4-version}</version>
+            <groupId>com.ibm.cloud</groupId>
+            <artifactId>cloudant</artifactId>
+            <version>0.8.6</version>
```
davsclaus, line 52: Add version to parent/pom.xml

davsclaus:
Yeah it would be good to update and improve the component as you suggest also.

davsclaus:
Oh you need to build camel-catalog and dsl as well to have changed files included in this PR

davsclaus (approved):

PR #14556, opened 2024-06-17
CAMEL-20722 - ensure concurrent processing from multiple kafka partitions

jvsingh:
Since this test is still flaky on **ppc64le**, made the below changes: 
-1) Add consumerCount > 1 in the poller
-2) Append random uuid to topic & route name for uniqueness 
-3) Remove fixed duration wait & replaced it with an "up to" duration in assertIsSatisfied(long millisec). 
_This further reduces a few seconds from test execution time._
-4) Ensure both partitions get read and asserted (even if the kafka poller stops reading from partition 0 due to breakOnFirstError, it will still read from partition 1).
(Some of the 'flakiness' could be reproduced locally by running the test repeatedly: because there is no guarantee which partition gets read by a Kafka consumer first. 
The only guarantee is: consumer should receive messages "from a particular partition" in the "same order in which they were produced".

Must create a blog post with a compilation of all changes, once this test is stable.

# Description

# Target

# Tracking


davsclaus (approved):

orpiske:
Nice one!

PR #14586, opened 2024-06-19
CAMEL-20884 - improve Velocity usage in Camel package Maven plugin

apupier:
- use a single RuntimeInstance of Velocity given that it is initialized always with the same parameters It was retrieved at least 2 times per component.
- use a cache to retrieve Velocity Template. It was retrieved at least 2 times per component. So now there is only 15 calls.
- Initialize the VelocityContext with the map of parameter in the constructor instead of iterating on each element of the map and adding them one by one

it was using around 10% of the time spent in org.apache.camel.** classes during a `mvn clean install -D quickly`. Locally went from around 10'05" to around 9'20" to execute previously mentioned command

# Description

# Target

# Tracking


apupier:
Generating timestamps to analyze logs, locally I got:

Before:
Screenshot from 2024-06-17 15-23-06

After:
image

PR #14606, opened 2024-06-20

fanyang:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Producer.java
@@ -23,10 +23,21 @@ import org.apache.camel.util.ObjectHelper;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.kinesis.model.PutRecordRequest;
 import software.amazon.awssdk.services.kinesis.model.PutRecordResponse;
+import software.amazon.awssdk.services.kinesis.model.PutRecordsRequest;
+import software.amazon.awssdk.services.kinesis.model.PutRecordsRequestEntry;
+import software.amazon.awssdk.services.kinesis.model.PutRecordsResponse;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


davsclaus:
Would you be able to add a new option on the endpoint (and component) to turn on this functionality. As this kind of batching only works for camel batching component that has the BATCH_COMPLETE flag. So it will only be applicate under some situations. And in case you shutdown Camel then if there are any pending messages in-memory then the batch should be auto complete and send during this process to avoid loosing those messages.

fanyang:
Hi Clause, do you mean adding an option to turn on/off sending batch? I don't think we need such option. If someone don't want to send batch, just not setting the flag. If someone want to send batch, need to set the flag in order to tell producer whether the current batch is complete.
For the second issue, I override doStop() to flush the remaining when stop.

oscerd:
> Hi Clause, do you mean adding an option to turn on/off sending batch? I don't think we need such option. If someone don't want to send batch, just not setting the flag. If someone want to send batch, need to set the flag in order to tell producer whether the current batch is complete. For the second issue, I override doStop() to flush the remaining when stop.

No, this is not the way we usually enable the behavior. If I want to use batch I need to be able to configure this at configuration level.

davsclaus:
> Hi Clause, do you mean adding an option to turn on/off sending batch? I don't think we need such option. If someone don't want to send batch, just not setting the flag. If someone want to send batch, need to set the flag in order to tell producer whether the current batch is complete. For the second issue, I override doStop() to flush the remaining when stop.

No there are components that always set these batch flags (such as file, sql, aws-sqs, azure-storage) and the problem would be that this component would not be thread-safe for mixed, and that if something fails and the completion flag is not reached to this component then the messages stays in memory etc. 

The user must explicit turn on batch to be aware of different behaviour in this component

fanyang:
Well, got it. It seems that setting BATCH_COMPLETE flag is not good. 
Another solution is to check the type of exchange body. If it's Iterable, then will send batch, if it's byte[], then will send single record. What do you think?

davsclaus:
I think you need to step back and find out what is your use-cases and where batching makes sense.

What is it that you are doing with Camel and azure where you think batching can help? What are your sources for the messages that you want to send to azure

PR #14609, opened 2024-06-20
CAMEL-20838: use a JUnit extension for collecting the test name when necessary

orpiske:


apupier:
requires a rebase on main branch (regen done in a separated commit) to have green PR

PR #14618, opened 2024-06-23

fanyang:
# Description

# Target

# Tracking


davsclaus (approved):

davsclaus:
It would be good to update the docs as well, do you mind creating another PR to update src/main/docs (see this folder) and add some information about sending a batch if the payload is iterarable such as a List

PR #14627, opened 2024-06-24
CAMEL-20901: Add an option to provide a custom Vertx instance to ArangoDB

jamesnetherton:


components/camel-arangodb/src/main/java/org/apache/camel/component/arangodb/ArangoDbComponent.java
```diff
@@ -31,6 +32,8 @@ public class ArangoDbComponent extends DefaultComponent {
 
     @Metadata(label = "advanced", autowired = true)
     private ArangoDB arangoDB;
+    @Metadata(label = "advanced")
```
davsclaus, line 35: I think it should be autowired = true, then its automatic uses the one from Quarkus out of the box

davsclaus (approved):

PR #14640, opened 2024-06-25
CAMEL-20917: avoid using hard-coded groups and versions

orpiske:
Also comes with a minor code cleanup


orpiske:
@davsclaus Please, would you mind a second look when you get a chance? Thanks!

orpiske:
> There is also `MavenGav` class you can use to parse and extract

Oh, I think that one will do the trick in a much better way. Let's hold this one for now and I'll change it to use that. Thanks for the pointer!

orpiske:
Adjusted to use a `MavenGav` instance.

PR #14641, opened 2024-06-25
Avoid NPE on Camel Tests

santam85:
# Description

Affects camel-test component. Related to https://issues.apache.org/jira/browse/CAMEL-20838

Fixes errors like:
```
java.lang.NullPointerException: Cannot invoke "org.apache.camel.test.junit5.CamelContextManager.setGlobalStore(org.junit.jupiter.api.extension.ExtensionContext$Store)" because "this.contextManager" is null
        at org.apache.camel.test.junit5.CamelTestSupport.beforeAll(CamelTestSupport.java:148)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        Suppressed: java.lang.NullPointerException: Cannot invoke "org.apache.camel.test.junit5.CamelContextManager.stop()" because "this.contextManager" is null
                at org.apache.camel.test.junit5.CamelTestSupport.afterAll(CamelTestSupport.java:153)
                ... 1 more
```
when running camel tests.

# Target

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20838

### DIFF
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
@@ -75,7 +75,7 @@ Configures batch-processing resequence eip.
     <xs:annotation>
       <xs:documentation xml:lang="en">
 <![CDATA[
-Calls a Java bean
+Define custom beans that can be used in your Camel routes and in general.
 ]]>
       </xs:documentation>
     </xs:annotation>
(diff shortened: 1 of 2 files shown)


orpiske (changes requested):
I need to look more closely at this one. We don't want the context to be created in the `beforeAll` unless the test has the correct annotation. 

It's possible that the test may be abusing one of the soon-to-be restricted APIs.

orpiske:
@santam85 please, can you try with this one instead: https://github.com/apache/camel/pull/14652

santam85:
Updated the PR to avoid creating the contextManager in beforeAll everytime

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
```diff
@@ -144,15 +145,16 @@ public void beforeAll(ExtensionContext context) {
             testConfigurationBuilder.withCreateCamelContextPerClass(perClassPresent);
             contextManager = contextManagerFactory.createContextManager(ContextManagerFactory.Type.BEFORE_ALL,
                     testConfigurationBuilder, camelContextConfiguration);
+            ExtensionContext.Store globalStore = context.getStore(ExtensionContext.Namespace.GLOBAL);
+            contextManager.setGlobalStore(globalStore);
```
orpiske, line 149: You don't need to change anything right now, but I think may cause some problems for nested tests. In those cases, JUnit may run the before* triggers more than once (and on each of those, we might need to update the global store). 

For now, keep it as a is, but we'll see how it goes after I run a full test of this PR.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
```diff
@@ -3973,50 +3973,10 @@ Enables random backoff. Default value: false
...
-as either singleton or prototype scope. So when using prototype scope then this depends on the bean registry
-implementation. Default value: Singleton
-]]>
-            </xs:documentation>
-          </xs:annotation>
-        </xs:attribute>
+        <xs:attribute name="ref" type="xs:string"/>
+        <xs:attribute name="method" type="xs:string"/>
+        <xs:attribute name="beanType" type="xs:string"/>
+        <xs:attribute name="scope" type="xs:string"/>
```
orpiske, line 3979: All of this needs to be removed from the PR. Please rebase the code and make sure this is not included.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
```diff
@@ -75,7 +75,7 @@ Configures batch-processing resequence eip.
     <xs:annotation>
       <xs:documentation xml:lang="en">
 <![CDATA[
-Calls a Java bean
+Define custom beans that can be used in your Camel routes and in general.
```
orpiske, line 78: This should be removed from the PR.

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
```diff
@@ -144,15 +145,16 @@ public void beforeAll(ExtensionContext context) {
...
-        ExtensionContext.Store globalStore = context.getStore(ExtensionContext.Namespace.GLOBAL);
-        contextManager.setGlobalStore(globalStore);
     }
 
     @Override
     public void afterAll(ExtensionContext context) {
-        contextManager.stop();
+        if(Objects.nonNull(contextManager)) {
+            contextManager.stop();
+        }
```
orpiske, line 157: You can ignore if the contextManager is null here. So, just add a check for nullity and, if present, then stop. 

There may be cases when the after* hooks could be called multiple times and/or without the contextManager having been created (I'm thinking when using `@Nested` tests, but we'll see after I run a full build).

orpiske (changes requested):
Thanks for your contribution. I'd like to favor your contribution over my POC on #14652, but please take a look at my review comments for things that need to be fixed before merging.

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
santam85: not sure I get your request, the non-null check is there?

orpiske: For instance:
```
 if (contextManager != null) {
            // It may be null in some occasion, such as when failing to initialize the context
            contextManager.stop();
        }
```

santam85:
Addressed all comments

components/camel-test/camel-test-junit5/src/main/java/org/apache/camel/test/junit5/CamelTestSupport.java
orpiske: Ran fine on the CI. Not a problem.

orpiske (approved):

PR #14652, opened 2024-06-26
CAMEL-20838: avoid causing an NPE during shutdown or initialization if the context manager is not available

orpiske:


oscerd (approved):

santam85:
@orpiske This works for me, but I obviously get the warning "Unable to set the global store for [...] because the context manager is not available" for every test. Isn't an idea to just move lines 149-150 inside the if?

orpiske:
> @orpiske ... As the execution is always `beforeAll` -> `beforeEach`, the contextManager will always be null unless the perClass annotation is present, or am I missing something?
> 

That's correct. 

> Updated #14641 with this change.

Thanks. I will also review that, run a full test on it and we'll go with either one of them (likely yours - after I review it).

davsclaus (approved):

orpiske:
Merged #14641 instead. Let's close.

PR #14674, opened 2024-06-28
CAMEL-19897: another shot at removing isStatisticsEnabled

orpiske:


oscerd (approved):

core/camel-base/src/main/java/org/apache/camel/impl/converter/CoreTypeConverterRegistry.java
```diff
@@ -616,10 +624,9 @@ protected void doBuild() throws Exception {
     protected void doStop() throws Exception {
         super.doStop();
         // log utilization statistics when stopping, including mappings
-        if (statistics.isStatisticsEnabled()) {
-            final String info = generateMappingStatisticsMessage();
-            LOG.info(info);
-        }
+
+        final String info = generateMappingStatisticsMessage();
+        LOG.info(info);
```
davsclaus, line 629: The noop statistic should not log anything

orpiske:
Obs.: the check is likely to fail because the build seems to be generating a new `camel-spring.xsd` file.

davsclaus:
try to rebase as there were some saga failing test that I need to look at on monday

davsclaus (approved):

PR #14718, opened 2024-07-03
BugFix removed write only object

dk2k:
BugFix removed write only object


davsclaus (approved):

davsclaus:
You need to remove the import also
```
-import java.util.concurrent.atomic.AtomicBoolean;
```

dk2k:
I removed the import

PR #14745, opened 2024-07-05
CAMEL-20227: camel-kafka - Kafka offset advances when using Pausable …

davsclaus:
…EIP and messages are lost when resumed

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-kafka/src/main/docs/kafka-component.adoc
+++ b/components/camel-kafka/src/main/docs/kafka-component.adoc
@@ -441,8 +441,6 @@ In this example, consuming messages can pause (by calling the Kafka's Consumer p
 The pausable EIP is meant to be used as a support mechanism when there is an exception somewhere in the route that prevents the exchange from being processed. More specifically,
 the check called by the `pausable` EIP should be used to test for transient conditions preventing the exchange from being processed.
 
-NOTE: If you use manual commit with Pauseable to only commit offset when the consumer is not paused, then also set `autoOffsetReset=true`.
-
 == Kafka Headers propagation
 
 When consuming messages from Kafka, headers will be propagated to camel exchange headers automatically.
--- a/components/camel-kafka/src/main/docs/kafka-component.adoc
+++ b/components/camel-kafka/src/main/docs/kafka-component.adoc
@@ -441,6 +441,8 @@ In this example, consuming messages can pause (by calling the Kafka's Consumer p
 The pausable EIP is meant to be used as a support mechanism when there is an exception somewhere in the route that prevents the exchange from being processed. More specifically,
 the check called by the `pausable` EIP should be used to test for transient conditions preventing the exchange from being processed.
 
+NOTE: If you use manual commit with Pauseable to only commit offset when the consumer is not paused, then also set `autoOffsetReset=true`.
+
 == Kafka Headers propagation
 
 When consuming messages from Kafka, headers will be propagated to camel exchange headers automatically.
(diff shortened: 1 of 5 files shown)


oscerd (approved):

davsclaus:
Argh yeah this test is a bit wrong as you want to be able to consume all 15 messages, but in the middle the consumer is paused

davsclaus:
okay I improved this so the pause and resume will not lose messages and that the pause is waiting a poll cycle, otherwise it keeps looping asap which is not desireable

orpiske:
Cool. It's a holiday here, but I will try to take a look later today

orpiske:
At first it wasn't very clear to me how the user would be losing messages after the pause. 

The KafkaConsumer documentation says that  [`pause`](https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#pause-java.util.Collection-) _"Suspend fetching from the requested partitions. Future calls to [poll(Duration)](https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll-java.time.Duration-) will not return any records from these partitions until they have been resumed using [resume(Collection)](https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#resume-java.util.Collection-)"_.

But looking more closely to the reproducer, I think I kinda understand the problem ... It's in part because of a misuse of the API (and, poor documentation for this particular feature). 

In particular, our documentation states: 

_"The pausable EIP is meant to be used as a support mechanism when there is an **exception** somewhere in the route that prevents the exchange from being processed. More specifically, the check called by the pausable EIP should be used to test for transient conditions preventing the exchange from being processed."_ 

Specifically, the trigger for those transient problems should be pulled by an exception, so that [we have a failure in the processing result](https://github.com/apache/camel/blob/main/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/streaming/KafkaRecordStreamingProcessorFacade.java#L97-L117). Otherwise, [we never cause the pause](https://github.com/apache/camel/blob/main/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/KafkaConsumerListener.java#L71-L88), thus leading to them losing the messages.

So, to make it short: the reproducer from the user is wrong.

orpiske:
That said, I think we do have a room for improvement on the API. I am not sure if we should be [allowing the code to resume the assignment if pause hasn't been called](https://github.com/apache/camel/blob/main/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/KafkaConsumerListener.java#L58-L68). 

Maybe we could throw an exception explaining that this is a misuse of the API, and trying to force a resume without and underlying transient problem is not supported.

davsclaus:
Yeah okay that can be better explain in the docs / javadoc / and the API is using a predicate so an exception is not thrown.

If the intend is to pause/resume kafka via some logic then use `RoutePolicy` that is intended for that. There is a throttling policy out of the box, but you can make your own as well.

orpiske:
@davsclaus actually, this is different from pausing the route per se. The idea for the `pausable` API is to pause the Kafka consumer (so as not to cause a rebalance when there is a transient issue). 

Nonetheless, let's see if we can brainstorm some improvements to this once you are back. There are some specific requirements for pausing the KafkaConsumer (such as: it must be done from the same thread as the one calling 'poll'), but if we could do this using the `RoutePolicy`, we could retire this API and make it simpler for the users.

davsclaus:
closing old PR

PR #14768, opened 2024-07-09
Kafka dev console and jbang command

davsclaus:
# Description

# Target

# Tracking


oscerd (approved):

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaDevConsole.java
```diff
@@ -0,0 +1,181 @@
...
+                                sb.append(String.format("\n        Commit Partition: %s", r.partition()));
+                                sb.append(String.format("\n        Commit Offset: %s", r.offset()));
+                                if (r.epoch() > 0) {
+                                    long delta = System.currentTimeMillis() - r.epoch();
+                                    sb.append(String.format("\n        Commit Offset Since: %s",
+                                            TimeUtils.printDuration(delta, true)));
+                                }
+                            }
+                        }
+                    }
```
orpiske, line 93: IMHO, this would be much better done by using a script language such as Velocity. Formatting the strings builtin are terribly hard to maintain.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
```diff
@@ -343,6 +371,30 @@ protected void startPolling() {
...
+                        if (count != null) {
+                            count.countDown();
+                        }
+                    } catch (Exception e) {
+                        // ignore cannot get last commit details
+                        LOG.debug("Cannot get last offset committed from Kafka brokers due to: {}. This exception is ignored.",
+                                e.getMessage(), e);
+                    }
+                }
+
```
orpiske, line 397: IMHO, it would be better to do this using a decorator. The biggest problem here is that we are attaching an edge case behavior right into the hot path of the component.

orpiske (commented):
I've left some suggestions. I think this is adding a lot of complexity to a component that is already very complex. IMHO, a decorator would be able to isolate that complexity and reduce the performance impacts on the hot path.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaDevConsole.java
davsclaus: No this is how the dev consoles are already done

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
davsclaus: No its only in use on special request

davsclaus:
> I've left some suggestions. I think this is adding a lot of complexity to a component that is already very complex. IMHO, a decorator would be able to isolate that complexity and reduce the performance impacts on the hot path.

The dev consoles are basic string and json object - its pure easy code - overkill to have template language as it adds unwanted dependency on some library. The dev consoles are used during development and not any hot-path at all.

Its a tremendous help for users of Camel & tooling and something we must add alot more to components.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
orpiske: > No its only in use on special request

So yeah, that's my point. We are calling `compareAndSet` here which, even though is an Intrinsic method, may actually cause a memory overhead due to need to perform CAS. 

So, even though an integration doesn't need the console, it would still be subject to this, which could hurt performance.

orpiske: And the reason I suggested a decorator, is because if we have a NOOP one and a Console one, then we could give the JVM a chance to use DCE to eliminate the code without hurting the performance.

orpiske:
> > I've left some suggestions. I think this is adding a lot of complexity to a component that is already very complex. IMHO, a decorator would be able to isolate that complexity and reduce the performance impacts on the hot path.
> 
> The dev consoles are basic string and json object - its pure easy code - overkill to have template language as it adds unwanted dependency on some library. The dev consoles are used during development 

OK.


> ... and not any hot-path at all.

Actually, the code is indeed right in the hot path of the consumer, as it's within the loop that calls `poll`. That's my main concern with this change.

components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
davsclaus: added a check if dev console enabled to skip the CAS

davsclaus:
> > > I've left some suggestions. I think this is adding a lot of complexity to a component that is already very complex. IMHO, a decorator would be able to isolate that complexity and reduce the performance impacts on the hot path.
> > 
> > 
> > The dev consoles are basic string and json object - its pure easy code - overkill to have template language as it adds unwanted dependency on some library. The dev consoles are used during development
> 
> OK.
> 
> > ... and not any hot-path at all.
> 
> Actually, the code is indeed right in the hot path of the consumer, as it's within the loop that calls `poll`. That's my main concern with this change.

Only doing the boolean check if dev-console is enabled now

orpiske (approved):

PR #14769, opened 2024-07-10
CAMEL-20227: avoid calling the consumer test unless Kafka consumer has been paused

orpiske:
Hey @davsclaus, this is more or less what I was thinking about in my comment on https://github.com/apache/camel/pull/14745#issuecomment-2212350669. 

I'm wondering if we should add a log on to inform that we are not calling the `afterConsumeEval.test(null)` because it has not been paused (I have not done so yet, as I think it would flood the log with those messages, because it would be called after every `poll`).

### DIFF
--- a/components/camel-kafka/src/main/docs/kafka-component.adoc
+++ b/components/camel-kafka/src/main/docs/kafka-component.adoc
@@ -439,8 +439,10 @@ from("kafka:topic")
 
 In this example, consuming messages can pause (by calling the Kafka's Consumer pause method) if the result from `canContinue` is false.
 
-The pausable EIP is meant to be used as a support mechanism when there is an exception somewhere in the route that prevents the exchange from being processed. More specifically,
+IMPORTANT: The pausable EIP is meant to be used as a support mechanism when *there is an exception* somewhere in the route that prevents the exchange from being processed. More specifically,
 the check called by the `pausable` EIP should be used to test for transient conditions preventing the exchange from being processed.
+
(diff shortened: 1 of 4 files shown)


orpiske:
BTW, this is just a remedy. Longer term I think I should consider augmenting the `RoutePolicy` API - as you suggested - and checking if this could be moved there in some way.

davsclaus (approved):

davsclaus:
> Hey @davsclaus, this is more or less what I was thinking about in my comment on [#14745 (comment)](https://github.com/apache/camel/pull/14745#issuecomment-2212350669).
> 
> I'm wondering if we should add a log on to inform that we are not calling the `afterConsumeEval.test(null)` because it has not been paused (I have not done so yet, as I think it would flood the log with those messages, because it would be called after every `poll`).

Yeah that is not desireable. If any logging then it can be TRACE level.
Also you can add in javadoc of the API that eval is not called if paused etc.

davsclaus:
> BTW, this is just a remedy. Longer term I think I should consider augmenting the `RoutePolicy` API - as you suggested - and checking if this could be moved there in some way.

We can add a note in the kafka doc to point users to route policy as well.
And maybe add an OOTB policy that can use a predicate or something as logic to suspend/resume, so users can use that.
And with the new label=bean configurer we can make it configurable and tooling friendly, and for low-code users as well.

orpiske:
Thanks. I'll include these documentation improvements into this PR.

orpiske:
Created https://issues.apache.org/jira/browse/CAMEL-20957 for re-implementing this as a `RoutePolicy`.

orpiske:
Looking good on the CI. Let's merge.

PR #14789, opened 2024-07-11
[CAMEL-20969] k8s plugin may generate conflicting docker build depend…

tdiesler:
…encies

# Description

# Target

# Tracking


davsclaus (approved):

christophd (commented):
Yes, great to consolidate the dependency options. 👍  

Can we add `--dependency` as an alias to `--dep` and `--deps` in export command? This would enable the user to use the same option for `run` and `export` commands

christophd:
Some commands use the comma delimited array, some use the String array option. We should aim to consolidate this, too.

I personally like the String array option much better, because it gives you the opportunity to have `--dependency ABC --dependency XYZ`  compared to `--dependencies ABC,XYZ`

christophd:
Let's merge this one and iterate on improvements

PR #14859, opened 2024-07-17
[CAMEL-20962] camel-jbang: SpringBoot support in Kubernetes plugin

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-20990

To test this, you can use the sample project [here](https://github.com/tdiesler/camel-cloud-examples/tree/main/spring-boot/timer-log).

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
@@ -29,6 +29,7 @@ import java.util.stream.Collectors;
 
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
 import org.apache.camel.dsl.jbang.core.commands.Export;
+import org.apache.camel.dsl.jbang.core.commands.ExportBaseCommand;
 import org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.ContainerTrait;
 import org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.TraitCatalog;
 import org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.TraitContext;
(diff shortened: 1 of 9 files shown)


oscerd (approved):

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
```diff
@@ -195,12 +197,16 @@ public Integer export() throws Exception {
...
-                    .collect(Collectors.toMap(parts -> parts[0], parts -> parts[1])));
-        }
+        labels = Optional.ofNullable(labels).orElse(new String[0]);
+        context.addLabels(Arrays.stream(labels)
+                .map(item -> item.split("="))
+                .filter(parts -> parts.length == 2)
+                .collect(Collectors.toMap(parts -> parts[0], parts -> parts[1])));
+
+        // Always provide these labels
+        // if (!context.getLabels().containsKey("app.kubernetes.io/name")) {
```
christophd, line 207: consider removing commented code

christophd (approved):
Awesome work! I love that we use the same property names just with different prefix according to the runtime provider

Just a nitpick about commented code fragment and there are some conflicts that need to be resolved. Many thanks!

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
tdiesler: Was being removed by https://github.com/apache/camel/pull/14864
Done, anyway

tdiesler:
Please note, the KubernetesExportTest now uses a combination of cmd line args and direct property access like this ...

```
        KubernetesExport command = createCommand(new String[] { "classpath:route.yaml" },
                "--gav=camel-test:route:1.0.0", "--runtime=" + rt.runtime());
        command.traits = new String[] { "container.port=8088", "container.image-pull-policy=IfNotPresent" };
        Assertions.assertEquals(0, command.doCall());
```

these args are then used by ...

```
        CommandLine.populateCommand(command, argsLst.toArray(new String[0]));
```

Although I like configuring these cmd objects explicitly, it only works from subclasses (i.e. a test cannot set properties on ExportBase). There are other places in the code where these access restrictions (i.e. non public) cause issues when the k8s plugin calls into csb plugin.

How do you feel about making these cmd properties public? I could get rid of `CommandLine.populateCommand`, tests could set stuff in the base class and stuff in the base class could get set before delegating to a cmd.

christophd:
I think `CommandLine.populateCommand()` is fine. I suppose we are using this primarily in tests and it is the way piccoli handles the command arguments so I would prefer to keep it as making all members public is also not an ideal solution IMO.

oscerd (approved):

PR #14884, opened 2024-07-19
CAMEL-20998: Prioritize objectName configuration over headers

bricefrisco:
# Description

Fixes https://issues.apache.org/jira/browse/CAMEL-20998

The `GoogleCloudStorageConstants.OBJECT_NAME` was previously prioritized over the endpoint configuration. This led to unexpected behavior, particularly when using the producer and consumer in the same route (I have documented more details about this in the Jira ticket).

I fixed this by checking for the `objectName` in the configuration first, and only if it is not specified there does it fall back to the header value.

# Target

# Tracking


orpiske (approved):
LGTM, thanks!

oscerd (approved):

PR #14926, opened 2024-07-24
Add support for polling more than 10 messages from AWS SQS server

jakubmalek:
# Description

Hi,
I'm using [camel-aws-sqs-source-kafka-connector](https://camel.apache.org/camel-kafka-connector/4.0.x/reference/connectors/camel-aws-sqs-source-kafka-source-connector.html), and recently I've found that I'm unable to poll all the messages from SQS queue in time. 
At first I wanted to increase the maximum number of messages polled in a single request. 
By default the client is configured to receive only `1` message and the maximum messages allowed by AWS is `10`. 
Here I had to add `maxMessagesPerPoll` parameter to the [Kamelet](https://github.com/apache/camel-kamelets/issues/2118) in order to configure it for the connector. 
_I'm mentioning the Kamelet because the parameter there has `maximum: 10` but this change renders that constraint obsolete._

Without the Kamelet there is a workaround to configure the `queueNameOrArn` with `?maxMessagesPerPoll=10&` suffix, to sneak in the parameter to the endpoint. 
But even with that and enabled `greedy` polling I was unable to keep up, and poll all messages in time. 
Increasing the number of concurrent polling threads is an option here, but I'm interested in processing the messages in the correct order. 

Looking at the Spring implementation, I've found that they are invoking receive requests with multiple threads and then merging the results together:
https://github.com/awspring/spring-cloud-aws/blob/a81b879d1a6b46f8a26794b98a601f7450d84907/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/source/AbstractSqsMessageSource.java#L194-L198

I think a similar approach could be applied in Apache Camel, where the number of concurrent requests is calculated from `ceil(maxMessagesPerPoll / 10D)` formula.

## Changes

* I've moved the logic for invoking the SQS requests into a dedicated class `Sqs2PollingClient`
* If the `maxMessagesPerPoll` is greater than 10, it executes multiple requests asynchronously and merge the results together
* To maintain the right ordering, the messages are ordered by the sequence number attribute if available
* In case of queue auto-creation I made sure that only a single request handling missing queue error will create the queue
* In case of recently deleted queue, I've removed the `Thread.sleep(30_000)` in favor of postponed creation with next polling call after `30_000` back-off time has passed
* I also noticed that `ReceiveMessageRequest.Builder#attributeNamesWithStrings` is now deprecated, so I've replaced it with `ReceiveMessageRequest.Builder#messageSystemAttributeNamesWithStrings`

# Target

# Tracking

_Do I need create a ticket for that ? I don't have the account in JIRA :)_


oscerd (changes requested):
The road could be good, but we should use the standard way of Camel of doing this stuff. You could have a look at how we are using executors in https://github.com/apache/camel/blob/main/components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java#L35 

We should not encapsulate everything into the client final class, but doing that the Camel way

jakubmalek:
I've changed the executor to be created via `ExecutorServiceManager` bean.
I wasn't sure about `java.time.Clock` bean, there is a `org.apache.camel.clock.Clock` but it provides uptime since exchange creation instead of current system time. In general I could use this time as reference point to measure time elapsed but I wasn't sure if this the right bean for that.

I'm not sure about the point of not encapsulating client. Do you mean that the logic should be part of `Sqs2Consumer` class ? 
In general I wanted to have a separated class as its tested more easily and the consumer class is quite large already. 
But if its the preferred way I can adjust.

oscerd:
In my opinion, the client logic should be inside a Task, like

https://github.com/apache/camel/blob/main/components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java#L113

oscerd:
@orpiske could you have a look too? Thanks.

oscerd:
I need to test this and also I need to look at the code from an IDE and not from the GitHub UI.

orpiske:
> @orpiske could you have a look too? Thanks.

Sure thing. I'll take a look at it today.

jakubmalek:
Thanks for the comments, I will try to adapt to the task approach, although I'm not sure if you I should refactor the consumer code to schedule the task in fixed period, or to re-use existing abstraction of `ScheduledBatchPollingConsumer` and simply move the logic into the existing `poll` method.

orpiske:
> Thanks for the comments, I will try to adapt to the task approach, although I'm not sure if I should refactor the consumer code to schedule the task in fixed period, or to re-use existing abstraction of `ScheduledBatchPollingConsumer` and simply move the logic into the existing `poll` method.

Personally, I'd prefer the `ScheduledBatchPollingConsumer` approach because I think it makes the code more consistent. However, I do realize it doesn't fit all cases (i.e.; it would be tricky to get that one for Kafka), so it's OK too if that doesn't suit the change. 

Ultimately, I think it depends on who handles the batching (Camel or the implementation library) and also what modes are available (streaming, batching, reactive, etc).

jakubmalek:
I've made some changes:
* First I've replaced the `Sqs2PollingClientTest` with `Sqs2ConsumerTest` where I test the logic using `Consumer` API
  * This way its irrelevant how the implementation is done inside the consumer, so I think its better this way
  * In terms of the testing I've also decided to switch to the `AmazonSQSClientMock` instead of Mockito, so I extended it a bit
* Second, I've moved the `Sqs2PollingClient` to `Sqs2Consumer.PollingTask` class 
  * In general it's not really a scheduled task, as `ScheduledBatchPollingConsumer` already does the scheduling, so I'm simply calling the "task" on the poll method inside the consumer
  * I was considering option to simply get rid of the dedicated task class, and simply put the implementation in the consumer class but I liked that its separated, but if you think its too artificial, I'm open to suggestions

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
```diff
@@ -473,4 +437,339 @@ private void logException(Exception e, List<ChangeMessageVisibilityBatchRequestE
...
+    private static class PollingTask implements Callable<List<software.amazon.awssdk.services.sqs.model.Message>>, Closeable {
+        /**
+         * The maximum number of messages that can be requested in a single request to AWS SQS.
+         */
+        private static final int MAX_NUMBER_OF_MESSAGES_PER_REQUEST = 10;
+
+        /**
+         * The maximum number of threads used for receive requests.
+         */
+        private static final int MAX_NUMBER_OF_CONCURRENT_REQUESTS = 50;
```
jakubmalek, line 475: Doubt: I guess this parameter should be configurable.

orpiske:
> I've made some changes:
> 
>     * First I've replaced the `Sqs2PollingClientTest` with `Sqs2ConsumerTest` where I test the logic using `Consumer` API
>       
>       * This way its irrelevant how the implementation is done inside the consumer, so I think its better this way
>       * In terms of the testing I've also decided to switch to the `AmazonSQSClientMock` instead of Mockito, so I extended it a bit
> 
>     * Second, I've moved the `Sqs2PollingClient` to `Sqs2Consumer.PollingTask` class
>       
>       * In general it's not really a scheduled task, as `ScheduledBatchPollingConsumer` already does the scheduling, so I'm simply calling the "task" on the poll method inside the consumer
>       * I was considering option to simply get rid of the dedicated task class, and simply put the implementation in the consumer class but I liked that its separated, but if you think its too artificial, I'm open to suggestions

Thanks! I'm a bit busy today, so I don't think I'll be able to re-review it until Monday ... but I wanted to ACK that I've seen the changes.

oscerd (approved):

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
```diff
@@ -473,4 +438,370 @@ private void logException(Exception e, List<ChangeMessageVisibilityBatchRequestE
...
+        private final Integer waitTimeSeconds;
+        private final Collection<MessageSystemAttributeName> attributeNames;
+        private final Collection<String> messageAttributeNames;
+        private final int numberOfRequestsPerPoll;
+        private final boolean queueAutoCreationEnabled;
+        private final MessageSystemAttributeName sortAttributeName;
+
+        @SuppressWarnings("resource")
+        private PollingTask(Sqs2Endpoint endpoint) {
+            clock = endpoint.getCamelContext().getClock();
```
orpiske, line 509: I think it would be better to avoid the CamelContext clock here. Instead, create a new one: 
`clock = new MonotonicClock()`. 

There's 2 reasons:

1. The `endpoint.getCamelContext().getClock()` gets the CamelContext's clock, which is used to track events for the camel context. 
2. It's slower, because the Context's clock has a more complex implementation (i.e.: need to be able to track multiple events, etc).

orpiske (changes requested):
Thanks @jakubmalek! This is looking great. 

I only have one small comment regarding the type of clock, which needs to be changed (I think it's just 1 line change). 

Everything else looks great to me.

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
jakubmalek: Done

orpiske (approved):
LGTM, thanks!

components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
orpiske: Thank you

jakubmalek:
I don't have the write access to merge the PR, so if there is nothing to fix, please merge it for me :)

orpiske:
Looking good. Let's merge.

PR #14960, opened 2024-07-29
CAMEL-20986: implement LangChain4j document splitters as a DSL

orpiske:


core/camel-api/src/main/java/org/apache/camel/spi/Tokenizer.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    Configuration newConfiguration();
+
+    void configure(Configuration configuration);
+
+    String name();
+
+    String[] tokenize(Exchange exchange);
+}
```
orpiske, line 37: TODO: this needs to be documented.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models.properties
```diff
@@ -83,6 +83,8 @@ jsonpath
 jtaTransactionErrorHandler
 kamelet
 kubernetesServiceDiscovery
+langChain4j
+langchain4j
```
oscerd, line 87: Is this right?

oscerd (approved):

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models.properties
orpiske: Hm, that looks suspicious. Let me take a look.

orpiske: I had named it differently in two occasions and there were leftovers. It's fixed now, thanks!

orpiske:
FYI, 

The checks are failing because the change is spread all over the code base:

```
There are too many projects to test so only the affected projects are tested
Jul 31, 2024 4:22:12 AM org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
[main] WARN org.jline - Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
Processing surefire and failsafe reports to create the summary
Error: Process completed with exit code 1.
```

So the incremental tester seems unable to run them.

However, looking at the logs, they all pass. This part is fine.

davsclaus:
LGTM

It would be good to add javadoc to the new code in the core/camel-api as its the public API for end users.

Also it would be good to see a standalone example with camel-spring-boot and camel-jbang - but that surely can be done later.

davsclaus (approved):

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/langChain4j.json
```diff
@@ -0,0 +1,23 @@
+{
+  "model": {
+    "kind": "model",
+    "name": "langChain4j",
+    "title": "LangChain4J Tokenizer",
+    "description": "Represents a tokenizer that uses LangChain4j for tokenization",
+    "deprecated": false,
+    "firstVersion": "4.8.0",
+    "label": "tokenizer,transformation,langchain4j",
```
davsclaus, line 9: Can we use `ai` instead of `langchain4j` so the label matches better that its for ai stuff. It has langchain in the name.

orpiske: Absolutely. I have adjusted it.

orpiske:
> LGTM
> 
> It would be good to add javadoc to the new code in the core/camel-api as its the public API for end users.
> 

Done!

> Also it would be good to see a standalone example with camel-spring-boot and camel-jbang - but that surely can be done later.

Tokenization + Camel JBang ... you are giving me some ideas here. I'll put that on my TODO, I think it would be cool to show this in other runtimes / ways.

Thanks!

orpiske:
All review points resolved. Tests passed on my CI. Let's merge and see how it goes. Thanks!

PR #14972, opened 2024-07-30
CAMEL-20199: camel-core - Remove synchronized blocks

essobedo:

# Description

# Target

# Tracking

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/CamelContextTracker.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/CamelContextTracker.java
@@ -19,6 +19,8 @@ package org.apache.camel.spi;
 import java.io.Closeable;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
 
 import org.apache.camel.CamelContext;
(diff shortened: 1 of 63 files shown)


essobedo:
Hi, is there anybody that could review this PR? Thank you in advance

orpiske:
Hi @essobedo, I'll take a quick look at it and I'll trigger a full test on my CI. I won't be able to post the results today, though.

essobedo:
> Hi @essobedo, I'll take a quick look at it and I'll trigger a full test on my CI. I won't be able to post the results today, though.

Hi @orpiske, it sounds great thank you

orpiske (changes requested):
I am fine with the approach. As far as I can tell, it's implemented as suggested.

Unfortunately, the way Java implemented VTs causes us to have these excess of Locks (which I kind dislike a bit). 

As such, there's an open question about how much `synchronized` blocks we want to remove. My previous understanding was that this was a problem only on the hot path (however, the code would still be susceptible to thread pinning problems if that code was inadvertently used). 

So, in short: LGTM, but we may still discuss how much `synchronized` we might want to adjust. 

I am only marking as request changes to hear some feedback from @oscerd and @davsclaus

essobedo:
I just wanted to add that if we have to choose between synchronized blocks and locks according to the situation, we need to be very careful during the review of each PR that adds new synchronized blocks as we have to evaluate the locking duration to know if it is an acceptable choice. Having a clear strategy at least allows to ease the review process but there is no bad strategy or good strategy both make sense and have drawbacks so I am waiting for your feedback before starting other PRs for the rest of the code

orpiske:
FYI, I've looked at the test results and they are looking great.

essobedo:
Good to know, thank you @orpiske

oscerd (approved):

oscerd:
This looks good to me, we just need to apply the same logic everywhere. Thanks.

davsclaus (approved):

orpiske (approved):

PR #14988, opened 2024-07-31
CAMEL-21038 Filter out app/bean.json so that EIPDocumentationEnricher produces consistent results

cunningt:
# Description

camel-spring.xsd is being updated quite a bit when devs do builds and devs are having to remove it from their commits quite a bit.     The changes that keep sneaking in look like they are because there are two bean.json's - we need to filter the app/bean.json out when doing documentation enrichment and then the camel-spring.xsd generation will be consistent.

CAMEL-21038 Filter out app/bean.json so that EIPDocumentationEnricher produces consistent results

# Target

# Tracking

- [x ] I have run `mvn clean install -DskipTests` locally and I have committed all auto-generated changes


oscerd (approved):

orpiske (approved):
Thanks @cunningt

PR #14996, opened 2024-08-01
CAMEL-20199: Remove synchronized block from Api Collections

essobedo:

## Motivation

## Modifications:

* Uses the [initialization-on-demand holder idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom) to avoid using synchronized blocks in Api Collections


essobedo:
@orpiske Hello 👋 , this one should be less controversial I think 😃

orpiske:
@essobedo looks promising! I'll trigger a full test on my CI.

oscerd (approved):

essobedo:
@orpiske why do you need a full build? Only a couple of components are affected and the incremental build seems to have launched the tests of the affected components with success but ok if it is needed

orpiske:
> @orpiske why do you need a full build? Only a couple of components are affected and the incremental build seems to have launched the tests of the affected components with success but ok if it is needed

I like running some tests like this so I can catch up discrepancies between GH and my own CI (on top of adding more coverage to some PRs).

orpiske (approved):

PR #14997, opened 2024-08-01
[CAMEL-21039] Align quarkus generated k8s resources with other runtimes

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21039

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
@@ -188,6 +188,21 @@
     </build>
 
     <profiles>
+        <profile>
+            <id>camel.debug</id>
+            <activation>
+                <property>
+                    <name>camel.debug</name>
+                    <value>true</value>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.apache.camel</groupId>
+                    <artifactId>camel-debug</artifactId>
+                </dependency>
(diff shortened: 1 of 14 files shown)


dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
```diff
@@ -94,39 +94,10 @@
                     </execution>
                 </executions>
             </plugin>
-            <!-- package as runner jar -->
```
davsclaus, line 97: Why is this removed? the camel-main as runtime should be a standard runtime that camel jbang can run and export. I just want to ensure that this still works and that its isolated from SB or Quarkus and that its 100% standalone.

davsclaus (changes requested):
I have been on PTO but want to ensure that camel-jbang does not break the camel-main runtime with all the code changes going in.

tdiesler:
Currently, we make a copy of the respective *-pom.tmpl for each runtime. i.e. you will find `*-kubernetes-pom.tmpl` for each runtime. Then, it is up to JKube to package the executable jar for use in the Docker container, build the image and finally generate the k8s manifest.

Hence, when you do `camel kubernetes export`, you'll get an executable jar, which you can run as documented [here](https://github.com/tdiesler/camel-cloud-examples/tree/main/main/timer-log). It is however not the same as you would get from plain `camel export`.

The idea was to implement `camel kubernetes export`, with as little as possible changes to the plain export/run.

PS: Indeed, the `package as runner jar` step was delete by mistake.

PR #15017, opened 2024-08-05
[CAMEL-21039] Align quarkus generated k8s resources with other runtimes

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21039

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
@@ -126,10 +126,9 @@
             <plugin>
                 <groupId>org.eclipse.jkube</groupId>
                 <artifactId>kubernetes-maven-plugin</artifactId>
-                <version>${camel.main.jkube.version}</version>
+                <version>${jkube.version}</version>
                 <configuration>
                     <imagePullPolicy>${camel.main.kubernetes.image-pull-policy}</imagePullPolicy>
-                    <buildStrategy>docker</buildStrategy>
                     <images>
                         <image>
                             <name>${camel.main.kubernetes.image-name}</name>
@@ -166,6 +165,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
+                <version>3.3.1</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
@@ -109,7 +109,7 @@ public class KubernetesExport extends Export {
                         description = "The image registry group used to push images to.")
     protected String imageGroup;
 
-    @CommandLine.Option(names = { "--image-builder" },
+    @CommandLine.Option(names = { "--image-builder" }, defaultValue = "docker",
                         description = "The image builder used to build the container image (e.g. docker, jib, podman, s2i).")
     protected String imageBuilder;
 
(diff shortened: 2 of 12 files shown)


tdiesler:
Currently, we make a copy of the respective *-pom.tmpl for each runtime. i.e. you will find *-kubernetes-pom.tmpl for each runtime. Then, it is up to JKube to package the executable jar for use in the Docker container, build the image and finally generate the k8s manifest.

Hence, when you do camel kubernetes export, you'll get an executable jar, which you can run as documented [here](https://github.com/tdiesler/camel-cloud-examples/tree/main/main/timer-log). It is however not the same as you would get from plain camel export.

The idea was to implement camel kubernetes export, with as little as possible changes to the plain export/run.

PS: Indeed, the package as runner jar step was delete by mistake.

dsl/camel-jbang/camel-jbang-core/src/main/resources/templates/main-kubernetes-pom.tmpl
```diff
@@ -105,7 +105,7 @@
                     </descriptors>
                     <archive>
                         <manifest>
-                            <mainClass>{{ .MainClassname }}</mainClass>
+                            <mainClass>examples.timerlog.CamelApplication</mainClass>
```
davsclaus, line 108: This seems like a mistake

oscerd (approved):

davsclaus (approved):

davsclaus:
@christophd do you have any feedback on this PR ?

gansheer (approved):

davsclaus (approved):

tdiesler:
Yes, otherwise it would be in draft or had unresolved comments

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
```diff
@@ -260,51 +260,30 @@ public Integer export() throws Exception {
...
             // Quarkus specific dependencies
+            // [TODO] migrate to JKube
             if (clusterType != null && clusterType.equals("openshift")) {
                 addDependencies("io.quarkus:quarkus-openshift");
                 if (imageBuilder == null) {
                     // use s2i image builder as a default on OpenShift
                     imageBuilder = "s2i";
                 }
-            } else {
-                addDependencies("io.quarkus:quarkus-kubernetes");
```
christophd, line 280: why did you remove this?

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
```diff
@@ -260,51 +260,30 @@ public Integer export() throws Exception {
...
-                imagePullPolicy = StringHelper.camelCaseToDash(imagePullPolicy);
-            }
             buildProperties.add("%s.kubernetes.image-pull-policy=%s".formatted(propPrefix, imagePullPolicy));
         }
 
         // Quarkus Runtime specific
         if (runtime == RuntimeType.quarkus) {
 
             // Quarkus specific dependencies
+            // [TODO] migrate to JKube
```
christophd, line 270: Is this still a TODO?

christophd (approved):
Just some clarification questions

christophd:
@tdiesler sorry, I am late to the review party. I only have two questions to be sure that I understand correctly. many thanks

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
tdiesler: It is part of https://issues.apache.org/jira/browse/CAMEL-21013 which I'm also working on

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
tdiesler: because, as we discussed, this is now uniformly handled by jkube

davsclaus: Hmm we should surely allow users to use the _quarkus standard_ for running on k8s. And would that not be that you should use their quarkus-kubernetes extension? So you can do all their configurations as well according to their guides.
https://quarkus.io/guides/deploying-to-kubernetes

davsclaus (changes requested):
need to be cliarified more

davsclaus:
ups did my review change cause the comments from christopher and me dissapear ?

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
christophd: @tdiesler it confuses me that the `io.quarkus:quarkus-openshift` is being added then when cluster-type is "openshift". how is that going to work when we remove all the quarkus settings?

christophd: @davsclaus the downside of this is that the export for runtime `quarkus` would be different compared to the other runtimes `main` and `spring-boot`. the idea of this PR is to find a common ground for all runtimes so that all exports use pure jKube plugin functionalities. but yes this also means users cannot use the quarkus Kubernetes extension configuration settings OOTB after export @tdiesler I got that right?

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
christophd: should we the wait for CAMEL-21013 to be resolved first? Otherwise this PR might break the deployment to OpenShift right?

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
tdiesler: Yes, the idea was that independent of runtime there is just one way on configuring k8s/openshift resources. The Openshift related code that is currently marked as TODO is untested legacy (which could also be removed in this PR). Support for Openshift will come back with [CAMEL-21013](https://issues.apache.org/jira/browse/CAMEL-21013), also uniformly through JKube.

davsclaus:
Okay we can all these runtimes. But later we should really use the _powers of the runtimes_ and let all this k8s configuration be done how these runtimes do it - Q and SB may not be similar, but our users want to use the standard way, instead of some custom way.

And the export is just a way to move on from prototype to a real project where the developer can then be on a good start to finish the project

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
tdiesler: Sure, we could wait until openshift is done as well. If we're not going to release any time soon, I'd prefer to do this iteratively and leave openshift out for now. I can continue to look at it next week - today I couldn't get cluster through the chat bot.

tdiesler:
Needs to get revisited. We want to stick with Quarkus for image building and k8s resource generation and use JKube for main and spring-boot

PR #15044, opened 2024-08-07
CAMEL-21049: Kamelets in Camel core

christophd:
# Description

Kamelets are still kind of separated from the Camel core project. The Kamelets live in a separate GitHub project repository (https://github.com/apache/camel-kamelets) and follow a separate release lifecycle which creates cyclic module dependencies.

In addition to that, writing and maintaining Kamelets is no fun as it boils down to writing YAML files manually.

This PR shows a potential way to move Kamelets into the Camel core project. The Kamelets are written in Java instead of YAML and use of a new service provider annotation `@KameletSpec`. The Kamelets are discoverable via service loader discovery mechanism. The metadata as well as the service loader information is auto-generated by the Camel Maven plugin tooling. The Kamelets also become part of the Camel catalog.

The PR implements a few Kamelets as an example:
-  set-body-action
- data-type-action
- timer-source
- log-sink

# Target

# Tracking

See CAMEL-21049


oscerd (commented):
This looks promising. Thanks @christophd. A few points:

1. I think rewriting the Kamelets as KameletSpec annotated is another effort in the path to move them to the core. I'm not sure if we want to translate 350 yaml files to KameletSpec annotated class..
2. Where do we place the documentation for the catalog? I think it's still worth to have a separated documentation.
3. We need also to think about all the bean we are using from camel-kamelets-utils because they are used in some of the kamelets.
4. The catalog got some more methods to deal with some features of Kamelets, I think we should review what we want to port.

@davsclaus please have a look too.

christophd:
going back to the drawing board. thanks for all comments and feedback

PR #15063, opened 2024-08-09
feat(core): auto cloud native properties

squakez:
Introduced camel.main.cloudPropertiesLocation which, if exists, would look into the related csv directories to find properties as possibly set in cloud native environments such as Kubernetes.

Closes CAMEL-21036

# Description

# Target

# Tracking


core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -433,6 +448,43 @@ private Properties tryLoadProperties(
...
+                if (Files.exists(confPath) && Files.isDirectory(confPath)) {
+                    Files.walkFileTree(confPath, new SimpleFileVisitor<Path>() {
+                        @Override
+                        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
+                            if (!Files.isDirectory(file)) {
+                                try {
+                                    String val = new String(Files.readAllBytes(file));
+                                    cp.put(file.getFileName().toString(), val);
+                                } catch (IOException e) {
+                                    // Do nothing or should we throw a RuntimeException?
```
squakez, line 469: I am not sure if we should throw an exception here. This may happen for instance if the user is setting a binary configuration.

davsclaus: Yeah I am not sure if its possible to know if its binary vs text (I guess there is no good naming convention).

Also I wonder if they file name should have extension stripped as they key should be without the extension should it not?

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -415,6 +421,15 @@ protected void configurePropertiesService(CamelContext camelContext) throws Exce
         if (op != null) {
             pc.setOverrideProperties(op);
         }
+
+        Optional<String> cloudLocations = pc.resolveProperty(MainConstants.CLOUD_PROPERTIES_LOCATION);
+        if (cloudLocations.isPresent()) {
+            LOG.info("Cloud properties location: {}", cloudLocations);
+            final Properties kp = tryLoadCloudProperties(op, cloudLocations.get());
+            if (kp != null) {
+                pc.setOverrideProperties(kp);
```
davsclaus, line 430: There is some code above that also set some OverrideProperties - so you would override what may have been added before. So I think you need to check if something exists, and then putAll in that case.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -433,6 +448,43 @@ private Properties tryLoadProperties(
...
+                    });
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        if (overridProperties == null) {
+            return cp;
+        }
+        Properties mergedProperties = new Properties(overridProperties);
```
davsclaus, line 483: Camel has a Location based properties so we know where they are from that we use for reporting to users. See `OrderedLocationProperties`

core/camel-main/src/main/java/org/apache/camel/main/DefaultConfigurationProperties.java
```diff
@@ -2727,4 +2728,16 @@ public T withStartupRecorderDir(String startupRecorderDir) {
...
+    }
+
+    /**
+     * Sets the locations (comma separated values) where to find properties configuration as defined for cloud native
+     * environments such as Kubernetes.
+     */
+    public void setCloudPropertiesLocation(String cloudPropertiesLocation) {
+        this.cloudPropertiesLocation = cloudPropertiesLocation;
+    }
+
```
davsclaus, line 2742: You should also add the `withXXX` method for this new option.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
squakez: No, there is no easy way to check that. However, given the user is setting this configuration on purpose, should be aware of what he/she is doing. As it's an auto configuration I'd be more inclined to just skip the failing configuration, maybe adding a log line to report that.

The file name is not expected. Typically, the file name is straight from the secret/configmap key. In the unit test I had to add a .txt as there was some maven process running validation and looking for licenses. However, it should work correctly without file extension.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
squakez: I'll have a look at this then, thanks.

davsclaus:
Users may have binary certificates as configmaps so I wonder if there should be a xxxBinary option so you can use another folder for binary files.

Or some other option where you can list the file names that are binary

squakez:
I made some test with some binary file, and it does not throw any exception. The presence of any binary file would be parsed normally, although it would not be probably something useful to have into app properties. The exception throwing is a generic one provided in all file IO signatures.

However, this is something that should not be expected. In order to enable this feature, the op team needs to mount any given secret/configmap in the application pod in a given folder (ie, /etc/camel/conf.d/_configmaps) and enable the parsing via setting `camel.main.cloud-properties-location= /etc/camel/conf.d/_configmaps` setting. I'd expect they mount the binary certificate under any different location if they don't want the property to be scanned as an application properties. Anything they are eventually mounting in those folders is because they want the runtime to scan as a property.

davsclaus:
Okay we can document this more clearly that the scan is intended for text based properties

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -433,6 +448,44 @@ private Properties tryLoadProperties(
...
+                        @Override
+                        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
+                            if (!Files.isDirectory(file)) {
+                                try {
+                                    String val = new String(Files.readAllBytes(file));
+                                    cp.put(loc, file.getFileName().toString(), val);
+                                } catch (IOException e) {
+                                    LOG.warn("Some error happened while reading property from cloud configuration file {}",
+                                            file, e);
+                                    e.printStackTrace();
```
davsclaus, line 470: Remove this the stacktrace is already logged in the WARN

davsclaus (approved):

squakez:
Not sure to understand this error:
```
There are uncommitted changes
HEAD detached at pull/15063/merge
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/main/camel-main-configuration-metadata.json

no changes added to commit (use "git add" and/or "git commit -a")
```
Is that file something I have to skip to pass the validation by any chance?

davsclaus:
You need to include the generated files in the catalog folder, you can run from root folder

```
camel install -P fastinstall
```

And then include updated files such as in the catalog folder in this PR.

PR #15083, opened 2024-08-09
CAMEL-21067: marked camel-univocity as deprecated

orpiske:


oscerd (approved):

apupier:
Shouldn't it be added to the migration guide?

davsclaus (approved):

apupier:
Also shouldn't we mark all public classes as Deprecated inside this module?

davsclaus:
No its fine just to mark the component as deprecated as Otavio do. No need to do more.

davsclaus:
Ah okay its a data format, so its a good idea to add `@Deprecated` to its model definition classes, so its also marked as deprecated in the DSL and in catalog and so on. So there is a bit more work when its a dataformat / language vs a regular component.

orpiske:
Let's mark this one as a draft. Later I'll take a look at the public classes we should deprecate.

orpiske:
Changes done. Let's merge.

PR #15086, opened 2024-08-09
[CAMEL-21063] Uniformly provide a heath endpoint for exported cloud projects

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21063

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
@@ -271,12 +271,31 @@ public class KubernetesExport extends Export {
         boolean allowPush = imageRegistry != null;
         buildProperties.add("jkube.%s.push=%b".formatted(imageBuilder, allowPush));
 
+        // Runtime specific for Main
+        if (runtime == RuntimeType.main) {
+            addDependencies("org.apache.camel:camel-health",
+                    "org.apache.camel:camel-platform-http-main",
+                    "org.apache.camel:camel-rest",
+                    "org.apache.camel:camel-rest-openapi");
+        }
+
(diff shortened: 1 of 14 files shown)


davsclaus (changes requested):
TODO: We need to look at the export main as it seems to be broken during my PTO.

The main export was working fine before and uses jkube and jib.

We need to look at this again and make sure it works as before in the 4.7.0 release.

tdiesler:
This PR restores the health endpoints in the k8s export layer (they used to be there by default).

To build with jib, you can use the `--image-builder=jib` property, 'docker' is the default. Building with jib requires `--image-registry` - currently this doesn't push automatically. The work with jkube/openshift is in progress.

Could you pls be a little more specific about 'seems to be broken' and 'works as before'? Apparently, improving the test coverage is needed too.

orpiske:
> TODO: We need to look at the export main as it seems to be broken during my PTO.
> 
> The main export was working fine before and uses jkube and jib.
> 
> We need to look at this again and make sure it works as before in the 4.7.0 release.

BTW, this raises the alarm that with all these new code getting into Camel JBang we might need to reinforce our test coverage on this component. Being a tool, it's probably a lot harder to test than regular components, but I think we might come up with something if we brainstorm it.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
```diff
@@ -109,7 +109,7 @@ public class KubernetesExport extends Export {
                         description = "The image registry group used to push images to.")
     protected String imageGroup;
 
-    @CommandLine.Option(names = { "--image-builder" },
+    @CommandLine.Option(names = { "--image-builder" }, defaultValue = "docker",
```
christophd, line 112: why not keep jib as the default image builder? Jib does not require users to install a Docker deamon

tdiesler:
Needs to get revisited. We want health endpoint with `camel kubernetes export` but not necessarily for plain `camel export`

PR #15103, opened 2024-08-12
CAMEL-21072: use the Ollama container from testcontainers

orpiske:
- Also allow adjusting the model used for the test


orpiske:
Check failure is unrelated. Let's merge.

PR #15145, opened 2024-08-13
CAMEL-20638: camel-platform-http - Do not return http request headers

kulagaIA:
# Description
- added platfrom-http endpoint option that makes proxy of HeaderFilterStrategy in PlatformHttpEndpoint to filter common http request headers so that they are not sent in the response

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20638


davsclaus (approved):

components/camel-platform-http/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
```diff
@@ -132,9 +169,30 @@ protected PlatformHttpConsumer createPlatformHttpConsumer(Processor processor) t
...
+        }
         return headerFilterStrategy;
     }
 
+    private HeaderFilterStrategy enhanceHeaderFilterStrategyToSkipHttpRequestHeaders(
+            HeaderFilterStrategy headerFilterStrategy) {
+        return new HeaderFilterStrategy() {
+            @Override
+            public boolean applyFilterToCamelHeaders(String headerName, Object headerValue, Exchange exchange) {
+                if (COMMON_HTTP_REQUEST_HEADERS.contains(headerName)) {
```
sean-hernon, line 183: According to the HTTP spec, headers are case-insensitive, but here they are being compared in a case-sensitive way. `COMMON_HTTP_REQUEST_HEADERS` should contain a set of normalised (e.g. lowercase) headers, then the header being compared should also be normalised (e.g. `COMMON_HTTP_REQUEST_HEADERS.contains(headerName.toLowerCase())`).

Sorry, I realise I'm a bit late to the party, as this has been merged, but it's problematic for my use case. What is the right way about going about getting this changed? Thank you.

davsclaus: Send another PR for review on top of main branch

PR #15163, opened 2024-08-14
Refactoring tracing interfaces to `camel-api` and `camel-support`

ammachado:
This allows moving the span decorators closer to the components.

# Description

Move tracing span interfaces to `camel-api` and abstract implementations to `camel-support`. Refactoring the current span property as a `ExchangePropertyKey`/internal property.

# Target

# Tracking


core/camel-api/src/main/java/org/apache/camel/Exchange.java
```diff
@@ -67,6 +67,7 @@
 @ConstantProvider("org.apache.camel.ExchangeConstantProvider")
 public interface Exchange extends VariableAware {
 
+    String ACTIVE_SPAN = "OpenTracing.activeSpan";
```
orpiske, line 70: qq: this is not following the pattern we use. Is there a reason for that?

orpiske (changes requested):
I'm not entirely sure I understand the motivations for the change, so marking as requested changes to wait for more feedback.

core/camel-api/src/main/java/org/apache/camel/Exchange.java
ammachado: This was copied from `camel-trace`, but I agree that it should be renamed.

ammachado:
> I'm not entirely sure I understand the motivations for the change, so marking as requested changes to wait for more feedback.

The idea of this pull request is to move functionality from `camel-trace` to the core libraries, and move all span decorators to their respective components.

core/camel-api/src/main/java/org/apache/camel/Exchange.java
davsclaus: This key is also used in `core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java`

davsclaus:
This requires a JIRA for tracking something like this happening so users can see this in changelogs

davsclaus:
The decorates can be moved to each component - however so far it has not been needed and having them in one place allows to easier identify which decorators we have. Also have you had a need for a decorator to rely directly on its component APIs vs as it is now.

oscerd (changes requested):
I'm not sure what would be the advantages of having decorators directly in the components.. If we place all of them in a single point, we could easily know what we support and what we don't, even for catalog purpose. If we move them in the single components it would be much more complicated.

PR #15182, opened 2024-08-19
OpenTelemetry testing improvements

ammachado:
# Description

This PR adds leaking spans detection, OTEL log reporting and prints a tree with the generated traces from the test run to help identify issues in user code.

# Target

# Tracking


davsclaus:
When you work on PRs in the future can you avoid doing code changes in the same PRs that change comments, coding standard etc in other codes. The PRs should be easier to review and ensure other code stays the same without causing unforseen problems. You can send other PRs for "cleanup" stuff.

core/camel-api/src/main/java/org/apache/camel/ExchangePropertyKey.java
```diff
@@ -23,6 +23,7 @@
  */
 public enum ExchangePropertyKey {
 
+    ACTIVE_SPAN("OpenTracing.activeSpan"),
```
davsclaus, line 26: Can you avoid doing changes to core in this PR and leave this where it was before.
Adding new to Exchange should be done in another PR

parent/pom.xml
```diff
@@ -388,6 +388,8 @@
         <opensearch-testcontainers-version>2.1.0</opensearch-testcontainers-version>
         <openstack4j-version>3.12</openstack4j-version>
         <opentelemetry-version>1.41.0</opentelemetry-version>
+        <opentelemetry-log4j2-version>2.7.0-alpha</opentelemetry-log4j2-version>
```
davsclaus, line 391: Hmm this is alpha software so we may have trouble maintaining it since this code change change a lot as it would not be stable yet.

davsclaus (changes requested):
Thanks for all the work on this. See my prev comments.

There are 2 concerns
- the exchange property key should not be in core (yet)
- alpha software versions

parent/pom.xml
ammachado: All versions available for this component are marked as `alpha`, unfortunately. This is used for testing only.

core/camel-api/src/main/java/org/apache/camel/ExchangePropertyKey.java
ammachado: I will revert this change.

parent/pom.xml
davsclaus: okay are they released regulay ie how old is those versions you use now ?

davsclaus (approved):

davsclaus:
This PR on CI reports test failures such as


Failed Test | Duration | Failure Type
-- | -- | --
org.apache.camel.opentelemetry.ClientRecipientListRouteTest.testRoute | 5.013s | FAILURE
org.apache.camel.opentelemetry.MulticastParallelRouteTest.testRoute | 5.023s | FAILURE
org.apache.camel.opentelemetry.MulticastRouteTest.testRoute | 4.547s | FAILURE
org.apache.camel.opentelemetry.TwoServiceWithExcludeTest.testRoute | 0.001s | ERROR
org.apache.camel.opentelemetry.OpenTelemetryPropagateContextTest.testTracingOfProcessors | 1.247s | FAILURE
org.apache.camel.opentelemetry.ABCRouteTest.testRoute | 6.668s | FAILURE
org.apache.camel.opentelemetry.CustomComponentNameRouteTest.testRoute | 6.090s | FAILURE

ammachado:
Thanks @davsclaus for the help. The tests are fixed.

orpiske (changes requested):
I'm not sure why the log4j configuration file is being replaced.

components/camel-opentelemetry/src/test/resources/log4j2.properties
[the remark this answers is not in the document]
davsclaus: Yes we must keep the existing log4j properties that is what we use

orpiske (approved):
LGTM, thanks!

davsclaus:
We need to do something about double closing

2024-08-21 06:24:25,911 [Pool-2-worker-1] ERROR ActiveSpanManager              - Tracing: scope already closed: io.opentelemetry.context.ThreadLocalContextStorage$ScopeImpl@21264a6a

These kind of ERROR should not happen usually and it may be that we should ignore double closing instead.

orpiske:
I'm seeing a couple of failures in the `camel-opentelemetry`. We might need to fix this or rollback the changes.

PR #15239, opened 2024-08-20
(chores) Simplify stream copies

orpiske:
2 things: 
- Use `transferTo` for simple copies, which should perform a bit better: https://bugs.openjdk.org/browse/JDK-8299336
- Use the same [buffer size as the one used by OpenJDK](https://github.com/openjdk/jdk/pull/11783/files).


oscerd (approved):

davsclaus (approved):

orpiske:
I need to review some of the test results.

orpiske:
> I need to review some of the test results.

The changes are looking good. Let's merge.

PR #15261, opened 2024-08-21
(chores) camel-langchain4j-chat: use human-readable names for the tools

orpiske:


oscerd (approved):

orpiske:
I've just got positive feedback about this change. So, let's merge.

PR #15298, opened 2024-08-23
Ban com.sun.xml.bind:jaxb-impl, replace it with jaxb-runtime where needed

ppalaga:
# Description

org.glassfish.jaxb:jaxb-runtime is the new Jakarta implementation of JAXB. It should be preferred to the old com.sun.xml.bind:jaxb-impl

# Target

# Tracking

No Jira, because this is just housekeeping. But I can create one, if needed.

### DIFF
--- a/apache-camel/pom.xml
+++ b/apache-camel/pom.xml
@@ -39,6 +39,7 @@
         <!-- we do not need remote resources in assembly -->
         <remoteresources.skip>true</remoteresources.skip>
         <tarLongFileMode>gnu</tarLongFileMode>
+        <enforcer.phase>none</enforcer.phase><!-- Save some time -->
     </properties>
 
     <repositories>
--- a/catalog/camel-allcomponents/pom.xml
+++ b/catalog/camel-allcomponents/pom.xml
@@ -32,6 +32,10 @@
     <name>Camel :: All Components Sync point</name>
     <description>Depends on all components to ensure correct build ordering</description>
 
+    <properties>
+        <enforcer.phase>none</enforcer.phase><!-- We enforce all dependencies individuall, so there is no point enforcing here again -->
+    </properties>
+
     <repositories>
         <repository>
             <id>atlassian</id>
--- a/components/camel-jaxb/pom.xml
+++ b/components/camel-jaxb/pom.xml
@@ -47,18 +47,6 @@
             <artifactId>camel-xml-jaxp</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>com.sun.xml.bind</groupId>
-            <artifactId>jaxb-core</artifactId>
-            <version>${jaxb-core-version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.sun.xml.bind</groupId>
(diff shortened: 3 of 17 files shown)


davsclaus:
Where is this information. Do we have a page that states this.

At this page the front page says
https://eclipse-ee4j.github.io/jaxb-ri/

```
        <dependencies>
            <dependency>
                <groupId>jakarta.xml.bind</groupId>
                <artifactId>jakarta.xml.bind-api</artifactId>
                <version>4.0.2</version>
            </dependency>
        </dependencies>

        <dependencies>
            <dependency>
                <groupId>com.sun.xml.bind</groupId>
                <artifactId>jaxb-impl</artifactId>
                <version>4.0.5</version>
                <scope>runtime</scope>
            </dependency>
```

ppalaga:
Thanks for the pointer, @davsclaus. Well, the authoritative resources about JAXB implementations are rather sparse and confusing. 
[This SO answer](https://stackoverflow.com/a/72151763) seems to describe the situation in the most complete way: 
* com.sun.xml.bind:jaxb-impl is marked as "old" in its own pom description: https://repo1.maven.org/maven2/com/sun/xml/bind/jaxb-impl/4.0.5/jaxb-impl-4.0.5.pom
* org.glassfish.jaxb:jaxb-runtime and com.sun.xml.bind:jaxb-impl classes overlap and differ only in whether they include or depend on some transitives (istack/txw2)
* The SO answer proposes to use newest versions of org.glassfish.jaxb:jaxb-runtime for projects after Jakarta migration and com.sun.xml.bind:jaxb-impl (versions before v3) for projects that still require javax.xml.bind. They can even co-exist in that way. 

Mainly the last point makes great sense to me. 

What are other's thoughts? Maybe I am missing something?

davsclaus:
And it would be good to add a note in the 4.8 upgrade docs about this change (so users would be aware) but hopefully they dont need to worried.

ppalaga:
> So this would need to check that camel and CSB and CEQ still works aftwards.

It is fine for Camel Quarkus, where we already merged a similar change https://github.com/apache/camel-quarkus/pull/6382

> And it would be good to add a note in the 4.8 upgrade docs 

Where can I do that?

davsclaus:
upg doc here
https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_8.adoc

davsclaus:
I wonder if CSB need some kind of update as well since CEQ had that ?

ppalaga:
> I wonder if CSB need some kind of update as well since CEQ had that ?

Yes, having an enforcer check might be a good idea to make sure that `com.sun.xml.bind:jaxb-impl` is not smuggled back by SB deps.

ppalaga:
> upg doc here https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_8.adoc

Thanks, done in 57f07d3. Please review.

davsclaus (approved):

PR #15318, opened 2024-08-26
CAMEL-21117: camel-jbang - Detect transacted dsl and add camel-jta au…

zhfeng:
…tomatically

# Description

# Target

# Tracking


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
```diff
@@ -570,6 +584,8 @@ private int run() throws Exception {
             // do not run for very long in silent run
             main.addInitialProperty("camel.main.autoStartup", "false");
             main.addInitialProperty("camel.main.durationMaxSeconds", "1");
+            TransactedPolicy policy = new DummyTransactedPolicy();
```
davsclaus, line 587: Its a bit hacky to do this in jbang run command directly.

I wonder if you can do this in the kamelet-main downloader, where you detect transacted DSL and then bind those dummy policies

davsclaus (approved):

PR #15381, opened 2024-08-30
CAMEL-21001: camel-jbang Kubernetes - Add Route trait

gansheer:
# Description

* Implement Route trait that is capable of configuring the Openshift Route resource
* Fix parsing of trait profile

# Target

# Tracking


dsl/camel-jbang/camel-jbang-plugin-kubernetes/pom.xml
```diff
@@ -54,6 +54,13 @@
             <version>${kubernetes-client-version}</version>
         </dependency>
 
+        <!-- Kubernetes -->
+        <dependency>
```
davsclaus, line 58: Why is this JAR added ?

oscerd (approved):

dsl/camel-jbang/camel-jbang-plugin-kubernetes/pom.xml
davsclaus: I mean did this JAR bring any code that this plugin needs or was it added for some other reason? Those JARs are a bit big so we should avoid if not needed.

gansheer: It brings the Route classes for the Openshift Route ressource, but I see what you means. Let me check if the model jar is enough in our case.

gansheer: I replaced the client jar with the model jar.

davsclaus (approved):

PR #15382, opened 2024-08-31
Spring WS docs and code cleanup.

ammachado:
# Description

This PR adds small changes to the Spring-WS component documentation, as well as it applies the `instanceof pattern` cleanup a few times.

# Target

# Tracking


components/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/SpringWebserviceComponent.java
```diff
@@ -75,20 +75,19 @@ private void addConsumerConfiguration(
         if (type != null) {
             LOG.debug("Building Spring Web Services consumer of type {}", type);
             String lookupKey = getLookupKey(remaining, type);
+            configuration.setEndpointMappingType(type);
+            configuration.setEndpointMappingLookupKey(lookupKey);
             if (EndpointMappingType.BEANNAME.equals(type)) {
                 addEndpointDispatcherToConfiguration(configuration, lookupKey);
             } else {
                 addEndpointMappingToConfiguration(parameters, configuration);
+                if (type.equals(EndpointMappingType.XPATHRESULT)) {
```
davsclaus, line 84: it seem this code is moved into this else block which was not the case before.

ammachado: You are correct, but this block will only execute if the branch where it is now is executed anyway, so I moved them together.

davsclaus:
modified:   components/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/bean/WSACamelEndpointMapping.java
	modified:   components/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/converter/StringSourceConverter.java

components/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/SpringWebserviceComponent.java
orpiske: A bit of a nitpick, but ideally it should be on a separate commit, as it prevents us from cherry-picking the doc changes to 4.4.x if we would like to.

orpiske (commented):
Those are great changes. 

IMHO it would be better to have them on separate commits, so we can cherry-pick changes if we want to.

davsclaus:
@ammachado there are 2 modified files. If you run mvn clean install in the folder then it would update those files.

ammachado:
My local build has some failing tests on `camel-openapi-java`, that's why I missed these changes. I'll open a JIRA for them.

ammachado:
I added a PR on Spring-WS to upgrade WSS4J as well. https://github.com/spring-projects/spring-ws/pull/1429 and match the version used in Camel.

davsclaus (approved):

PR #15396, opened 2024-09-02
CAMEL-21153: rework the function calling features

orpiske:
Create a new component specific for function calling, based on the same functionality already present on the camel-langchain4j-chat component.


oscerd (approved):

components/camel-ai/camel-langchain4j-tools/src/main/docs/langchain4j-tools-component.adoc
```diff
@@ -0,0 +1,187 @@
...
+The parameters can be found as headers in the consumer route, in particular, if you define `parameter.userId=5`,
+in the consumer route `${header.userId}` can be used.
+
+.Producer and consumer example:
+[source, java]
+----
+from("direct:test")
+    .to("langchain4j-tools:test1?tags=users");
+
+from("langchain4j-chat:test1?tags=users&description=Query user database by number&parameter.number=integer")
```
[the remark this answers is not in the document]
orpiske: Thanks for pointing this!

This should work as well, but I think it's better to use the same parameter / header name described above. I'll adjust the documentation.

components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsEndpoint.java
```diff
@@ -0,0 +1,198 @@
...
+    @UriParam
+    private LangChain4jToolsConfiguration configuration;
+
+    @Metadata(label = "consumer")
+    @UriParam(description = "Tool description")
+    private String description;
+
+    @Metadata(label = "consumer")
+    @UriParam(description = "List of Tool parameters in the form of parameter.<name>=<type>", prefix = "parameter.",
+              multiValue = true, enums = "string,integer,number,object,array,boolean,null")
```
davsclaus, line 63: Hmm so the enums is referring to the possible values in the Map for the multi-value. Yeah I dont recall seeing this before, but if we get this into the metadata json then tooling can use it to know this and provide a better UX experience.

components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsEndpoint.java
```diff
@@ -0,0 +1,198 @@
...
+        CamelToolSpecification camelToolSpecification
+                = new CamelToolSpecification(toolSpecification, new LangChain4jToolsConsumer(this, processor));
+        final CamelToolExecutorCache executorCache = CamelToolExecutorCache.getInstance();
+
+        String[] splitTags = TagsHelper.splitTags(tags);
+        for (String tag : splitTags) {
+            executorCache.put(tag, camelToolSpecification);
+        }
+
+        return camelToolSpecification.getConsumer();
```
davsclaus, line 124: When creating a consumer, you also need to call `configureConsumer(consumer)` before returning the object.

orpiske: Ahh, good one. BTW, I've just noticed that we have the same problem on the `camel-langchain4j-chat` component. I'll fix here first and later there. Thanks!

components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsEndpoint.java
orpiske: Thanks! IMHO, it would be very important to have better UX for this component, as it would allow users to design intelligent routes with ease. 

But, I am not sure I understand how exactly I should change it. Can you elaborate, please?

orpiske:
Changing to draft again to address all comments and make sure we have them addressed before merge.

components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsEndpoint.java
davsclaus: No changes needed. Just that the UI folks should be aware for multi value = true, and enum, then the enum is for the values.

orpiske: Understood, thanks!

orpiske:
Ready to go. Merging!

zbendhiba:
Nice !

PR #15421, opened 2024-09-04
CAMEL-21163: unwrap the synthetic ClientProxy beans

zhfeng:
# Description

# Target

# Tracking

### DIFF
--- a/core/camel-util/src/main/java/org/apache/camel/util/UnwrapHelper.java
+++ b/core/camel-util/src/main/java/org/apache/camel/util/UnwrapHelper.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


core/camel-util/src/main/java/org/apache/camel/util/UnwrapHelper.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.camel.util;
+
+import java.lang.reflect.Method;
+
+/**
+ * Helper for unwrapping from the proxy objects
+ */
+public final class UnwrapHelper {
+
+    public static <T> T unwrapFromQuarkus(T obj) {
```
davsclaus, line 26: We can add other runtimes here - so the name could be `unwrapClientProxy` and allow to be more generic in the future.

oscerd (approved):

davsclaus (approved):

core/camel-util/src/main/java/org/apache/camel/util/UnwrapHelper.java
zhfeng: Yeah, I rename it.

jamesnetherton (approved):

PR #15434, opened 2024-09-04
CAMEL-21040 + CAMEL-21066

orpiske:


orpiske:
I adjusted the release documentation here, so it would be good to double check @davsclaus / @oscerd

oscerd (approved):

davsclaus (approved):

PR #15445, opened 2024-09-05
chore(camel-jbang): Kubernetes export of current directory

christophd:
# Description

- Allow Kubernetes export to use all files in the current directory

# Target

# Tracking


oscerd (approved):

davsclaus:
Beware when you only do changes to kubernetes export. The regular export should be similar.

So with this change then a readme.md file is exported to where?

christophd:
@davsclaus are you referring to the changes in `/jbang/core/commands/kubernetes/MetadataHelper.java` I will check if this is being used by arbitrary export but I think it is only being used by Kubernetes plugin as it inspects the source code for trait configs. This is why non-source code files like `readme.md` are ignored here

davsclaus (approved):

PR #15460, opened 2024-09-05
CAMEL-21164: camel-rest - Code first should use actual values for pro…

avi5kdonrh:
…perty placeholders in dumped API spec

Changes: camel-openapi-java:

          org.apache.camel.openapi RestOpenApiReader
          org.apache.camel.openapi.SpringRestOpenApiReaderModelApiSecurityPlaceholderTest
          org/apache/camel/openapi/SpringRestOpenApiReaderModelApiSecurityPlaceholderTest.xml

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
+++ b/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
@@ -272,7 +272,7 @@ public class RestOpenApiReader {
                         ? new String[] { getValue(camelContext, rest.getPath()) }
                 : new String[0];
 
-        parseOas30(openApi, rest, pathAsTags);
+        parseOas30(camelContext, openApi, rest, pathAsTags);
 
         // gather all types in use
         Set<String> types = new LinkedHashSet<>();
@@ -341,7 +341,7 @@ public class RestOpenApiReader {
         });
     }
 
-    private void parseOas30(OpenAPI openApi, RestDefinition rest, String[] pathAsTags) {
+    private void parseOas30(CamelContext camelContext, OpenAPI openApi, RestDefinition rest, String[] pathAsTags) {
         String summary = rest.getDescriptionText();
 
         for (String tag : pathAsTags) {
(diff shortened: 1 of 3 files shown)


davsclaus:
This is theoretical a bit wrong as the model classes should not have their values changed.

So can you see if you can find another solution where you replace the values in the dumper

avi5kdonrh:
Thanks for taking a look @davsclaus.  I have replaced the value in the dumper with the new commit. Please review.

components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
```diff
@@ -358,36 +358,40 @@ private void parseOas30(OpenAPI openApi, RestDefinition rest, String[] pathAsTag
...
+                    openApi.getComponents().addSecuritySchemes(CamelContextHelper.parseText(camelContext, def.getKey()), auth);
                 } else if (def instanceof ApiKeyDefinition) {
                     ApiKeyDefinition rs = (ApiKeyDefinition) def;
                     SecurityScheme auth = new SecurityScheme().type(SecurityScheme.Type.APIKEY)
-                            .name(rs.getName()).description(def.getDescription());
-
-                    if (rs.getInHeader() != null && Boolean.parseBoolean(rs.getInHeader())) {
+                            .name(CamelContextHelper.parseText(camelContext, rs.getName()))
+                            .description(CamelContextHelper.parseText(camelContext, def.getDescription()));
+                    String inHeader = CamelContextHelper.parseText(camelContext, rs.getInHeader());
```
davsclaus, line 374: You can use parseBoolean instead of parseText here

davsclaus:
Thanks this is much better - just a few places where it may be better to use `parseBoolean` than `parseText`

components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
avi5kdonrh: Thanks for the review, I have updated the PR.

davsclaus (approved):

PR #15515, opened 2024-09-11
[CAMEL-21199] Camel-jackson not properly marshalling 4-byte characters

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-21199

Jackson doesn't handle 4-byte characters well. Marshalling a 4-byte Japanese kanji character results in two UTF-16 escapes to be written instead of the character itself. While this is ok for emoji an such, it's not for natural languages. 

Jackson issue: https://github.com/FasterXML/jackson-core/issues/223

Reproducer:
```
from("file:data?file-name=input.txt&noop=true")
    .log("${body}")
    .unmarshal().json(JsonLibrary.Jackson)
    .log("${body[0]['name']}")
    .marshal().json(JsonLibrary.Jackson, true)
    .log("${body}");
```

with the file input.txt containing:
`[{"name": "システム𩸽"}]`

Solution:
While it's an error in Jackson-core, it could be avoided by forcing Jackson to use _WriterBasedJsonGenerator_ instead of _UTF8JsonGenerator_ (see the original Jackson issue reproducer in the [topmost comment](https://github.com/FasterXML/jackson-core/issues/223#issue-110880547))

### DIFF
--- a/components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
+++ b/components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
@@ -19,6 +19,7 @@ package org.apache.camel.component.jackson;
 import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -157,7 +158,11 @@ public abstract class AbstractJacksonDataFormat extends ServiceSupport
         if (this.schemaResolver != null) {
             schema = this.schemaResolver.resolve(exchange);
         }
-        this.objectMapper.writerWithView(jsonView).with(schema).writeValue(stream, graph);
+        // using OutputStreamWriter because of Jackson not handling 4-byte characters properly with only a stream
+        // https://github.com/FasterXML/jackson-core/issues/223
+        try (OutputStreamWriter writer = new OutputStreamWriter(stream)) {
+            this.objectMapper.writerWithView(jsonView).with(schema).writeValue(writer, graph);
+        }
 
         if (contentTypeHeader) {
             exchange.getMessage().setHeader(Exchange.CONTENT_TYPE, getDefaultContentType());
--- a/components/camel-jackson/src/test/java/org/apache/camel/component/jackson/JacksonMarshalMultibyteCharactersTest.java
+++ b/components/camel-jackson/src/test/java/org/apache/camel/component/jackson/JacksonMarshalMultibyteCharactersTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.jackson;
+
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
+import org.apache.camel.test.junit5.CamelTestSupport;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+public class JacksonMarshalMultibyteCharactersTest extends CamelTestSupport {
+
+    @Test
+    public void testMarshal4ByteCharacter() throws Exception {
+        Map<String, Object> in = new HashMap<>();
+        in.put("name", "システム\uD867\uDE3D");
+
+        MockEndpoint mock = getMockEndpoint("mock:reverse");
+        mock.expectedMessageCount(1);
+        mock.message(0).body().isInstanceOf(Map.class);
+        mock.message(0).body().isEqualTo(in);
+
+        Object marshalled = template.requestBody("direct:in", in);
+        String marshalledAsString = context.getTypeConverter().convertTo(String.class, marshalled);
+
+        // prior to fixing CAMEL-21199, this was the result returned by Jackson
+        assertNotEquals("{\"name\":\"システム\\uD867\\uDE3D\"}", marshalledAsString);
+
+        assertEquals("{\"name\":\"システム\uD867\uDE3D\"}", marshalledAsString);
+
+        template.sendBody("direct:back", marshalled);
+        mock.assertIsSatisfied();
+    }
+
+    @Override
+    protected RouteBuilder createRouteBuilder() {
+        return new RouteBuilder() {
+
+            @Override
+            public void configure() {
+                JacksonDataFormat format = new JacksonDataFormat();
+                from("direct:in").marshal(format);
+                from("direct:back").unmarshal(format).to("mock:reverse");
+            }
+        };
+    }
+
+}


davsclaus:
This should be fixed in jackson instead of workaround in camel that can affect others.

davsclaus:
Yes and if its dangerous for jackson, then why is it not dangerous for camel ?

What you can/should do is to create a JIRA ticket, and add a new option to the dataformat, where you can configure this with a boolean (default false) and then only use this new code if enabled. Then existing is not affected at all.

rnetuka:
> Yes and if its dangerous for jackson, then why is it not dangerous for camel ?
It was claimed dangerous for Jackson to dabble with _UTF8JsonGenerator_ which is erronous. This solution forces Jackson to use _WriterBasedJsonGenerator_ inside, which seems to work fine.

> What you can/should do is to create a JIRA ticket, and add a new option to the dataformat, where you can configure this with a boolean (default false) and then only use this new code if enabled. Then existing is not affected at all.
Seems like an overkill to me, but if you like, I can surely add the option

davsclaus:
> > Yes and if its dangerous for jackson, then why is it not dangerous for camel ?
> 
> It was claimed dangerous for Jackson to dabble with _UTF8JsonGenerator_ which is erronous. This solution forces Jackson to use _WriterBasedJsonGenerator_ inside, which seems to work fine.
> 
> > What you can/should do is to create a JIRA ticket, and add a new option to the dataformat, where you can configure this with a boolean (default false) and then only use this new code if enabled. Then existing is not affected at all.
> 
> Seems like an overkill to me, but if you like, I can surely add the option

If its so non dangerous why has jackson not fixed it already. Please respect the large existing user-base of Camel that dont want hacks in commonly used components.

Also we need unit tests to go with that thanks.

rnetuka:
> If its so non dangerous why has jackson not fixed it already

Don't ask me. Ask them :)

rnetuka:
@davsclaus I agree that it should be fixed in Jackson. Sadly, it hasn't been fixed for a long time and they don't consider it much of an issue. This _hack_ as you call it just tries to overcome the error by invoking different string-writing mechanism (on their side) and should be same as soon as calling Jackson marshal method ends.

I see no point in providing an option. It's hardly an option for Camel users. It's either fixed, or not.

If you don't want this fix (which I can understand), feel free to close it. I'll than try to create a fix for Jackson itself. But I doubt they will accept it...

davsclaus:
> @davsclaus I agree that it should be fixed in Jackson. Sadly, it hasn't been fixed for a long time and they don't consider it much of an issue. This _hack_ as you call it just tries to overcome the error by invoking different string-writing mechanism (on their side) and things should be same as soon as calling Jackson _marshal_ method ends.
> 
> I see no point in providing an option. It's hardly an option for Camel users. It's either fixed, or not.
> 

I am sorry but we cannot risk affecting everyone else with some code change that Jackson itself does have out of the box, and its been outstanding since 2015. Would such a change cause unforseen changes to existing users, performance degration or what else. 

Are there other reported tickets at Jackson than this single instance?



> > Also we need unit tests to go with that thanks.
> 
> It cannot be unit-tested. It's integration issue.
> 

What we ask is that there is a junit test class in the PR that tests this code change.


> If you don't want this fix (which I can understand), feel free to close it. I'll then try to create a fix for Jackson itself. But I doubt they will accept it...

Maybe take more time to investigate if there are other reports on Jackson. But in the mean time there is only a solution you can do and that is on the Camel side. But that requires more work which you then must step up and do. 

And it would require a JIRA ticket in the Apache Camel JIRA at
https://issues.apache.org/jira/projects/CAMEL/

rnetuka:
> I am sorry but we cannot risk affecting everyone else with some code change that Jackson itself does have out of the box, and its been outstanding since 2015. Would such a change cause unforseen changes to existing users, performance degration or what else.

I'm not sure if we talk about the same thing here. Jackson offers two string generators - _UTF8JsonGenerator_ (which we are using at the moment and which contains the error) and _WriterBasedJsonGenerator_ (which seems to handle 4-byte characters fine). I don't see switching from one to another risky for Camel users, since they should behave exactly the same (not speaking about the error itself).

rnetuka:
@davsclaus I've added a test for the fix. Also, the Camel JIRA issue has been created.

components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
```diff
@@ -157,7 +158,9 @@ public void marshal(Exchange exchange, Object graph, OutputStream stream) throws
         if (this.schemaResolver != null) {
             schema = this.schemaResolver.resolve(exchange);
         }
-        this.objectMapper.writerWithView(jsonView).with(schema).writeValue(stream, graph);
+        // using OutputStreamWriter because of Jackson not handling 4-byte characters properly with only a stream
+        // https://github.com/FasterXML/jackson-core/issues/223
+        this.objectMapper.writerWithView(jsonView).with(schema).writeValue(new OutputStreamWriter(stream), graph);
```
orpiske, line 163: Shouldn't you close the `OutputStreamWriter` here?

orpiske (changes requested):
Likely need to close the `OutputStreamWriter` ... waiting for clarifications.

components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
rnetuka: Thanks Otavio. Please check the updated changes.

components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
```diff
@@ -157,7 +158,11 @@ public void marshal(Exchange exchange, Object graph, OutputStream stream) throws
         if (this.schemaResolver != null) {
             schema = this.schemaResolver.resolve(exchange);
         }
-        this.objectMapper.writerWithView(jsonView).with(schema).writeValue(stream, graph);
+        // using OutputStreamWriter because of Jackson not handling 4-byte characters properly with only a stream
+        // https://github.com/FasterXML/jackson-core/issues/223
+        OutputStreamWriter writer = new OutputStreamWriter(stream);
+        this.objectMapper.writerWithView(jsonView).with(schema).writeValue(writer, graph);
+        writer.close();
```
orpiske, line 165: Use try-with-resources here. It's cleaner.

orpiske (changes requested):

orpiske:
BTW, I am only reviewing the technicalities here. @davsclaus points are a lot more important than my notes.

components/camel-jackson/src/main/java/org/apache/camel/component/jackson/AbstractJacksonDataFormat.java
rnetuka: The time I spent doing fixes instead of developing finally starts to show up :) Thanks for the tip!

components/camel-jackson/src/test/java/org/apache/camel/component/jackson/JacksonMarshalMultibyteCharactersTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+public class JacksonMarshalMultibyteCharactersTest extends CamelTestSupport {
+
+    @Test
+    public void testMarshalAndUnmarshalMap() throws Exception {
+        Map<String, Object> in = new HashMap<>();
+        in.put("name", "システム𩸽");
```
davsclaus, line 35: We do not store non-ascii characters directly in the source code. You can use Java unicode syntax to build up those chars. 

See this unit test
https://github.com/apache/camel/blob/main/core/camel-core/src/test/java/org/apache/camel/component/file/FileConvertBodyToUTF8Test.java

It uses the thai elephant symbol
https://www.compart.com/en/unicode/U+0E08

I assume you can find the uni code for those symbols above and use the java string syntax for these values.

rnetuka:
@orpiske That's fine. Thak you very much for your review!

oscerd (approved):

davsclaus:
We use unicode escape to have foregin symbols in the camel source code.

See for example
https://github.com/apache/camel/blob/main/core/camel-core/src/test/java/org/apache/camel/component/file/FileConvertBodyToUTF8Test.java

Can you see if you can replace those to escaped values in this PR unit test.

components/camel-jackson/src/test/java/org/apache/camel/component/jackson/JacksonMarshalMultibyteCharactersTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+public class JacksonMarshalMultibyteCharactersTest extends CamelTestSupport {
+
+    @Test
+    public void testMarshal4ByteCharacter() throws Exception {
+        Map<String, Object> in = new HashMap<>();
+        in.put("name", "システム\uD867\uDE3D");
```
davsclaus, line 35: We use unicode escape to have foregin symbols in the camel source code.

See for example
https://github.com/apache/camel/blob/main/core/camel-core/src/test/java/org/apache/camel/component/file/FileConvertBodyToUTF8Test.java

Can you see if you can replace those to escaped values in this PR unit test.

davsclaus (changes requested):

davsclaus:
Added option `useWriter` in this PR
https://github.com/apache/camel/pull/15538

davsclaus:
On main branch we have the new option `useWriter` that can be enabled for this use-cases.

davsclaus:
@rnetuka what is the status of this ?

oscerd (approved):

rnetuka:
PR for Jackson issue 223 has been merged: https://github.com/FasterXML/jackson-core/pull/1335. There will be a new Jackson option to combine utf-16 surrogate characters into a one unicode char in their _UTF8JsonGenerator_.

 I'm closing this PR and will create a replacement with the new option being supported.

PR #15516, opened 2024-09-11
CAMEL-21195: resolved flaky tests

LostArtist:
resolved flaky tests


orpiske (changes requested):
Minio seems flaky only on ppc64le.

orpiske:
This is not correct, therefore closing.

PR #15558, opened 2024-09-13
CAMEL-21216: Apache Camel LRA does not work with Oracle MicroTX LRA coordinator

DFiedler:
Enable Oracle MicroTX lra-coordinator compatibility (https://issues.apache.org/jira/browse/CAMEL-21216).

1) While joining an LRA Transaction, the payload sent by Apache Camel is URL encoded (fix in: LRAUrlBuilder.java)
2) The Query parameter sent by Oracle LRA coordinator are url encoded, but will not be decoded by Apache Camal LRA (see also: CAMEL-21197) (workaround in: LRASagaRoutes.java)

The changes are backwards compatible to Narayana lra-coordinator.

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
@@ -16,15 +16,15 @@
  */
 package org.apache.camel.service.lra;
 
-import org.apache.camel.Exchange;
-import org.apache.camel.builder.RouteBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.stream.Collectors;
+
+import org.apache.camel.Exchange;
+import org.apache.camel.builder.RouteBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.apache.camel.service.lra.LRAConstants.*;
 
@@ -32,7 +32,7 @@ public class LRASagaRoutes extends RouteBuilder {
 
     private static final Logger LOG = LoggerFactory.getLogger(LRASagaRoutes.class);
 
-    private LRASagaService sagaService;
+    private final LRASagaService sagaService;
 
     public LRASagaRoutes(LRASagaService sagaService) {
         this.sagaService = sagaService;
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
@@ -16,23 +16,21 @@
  */
 package org.apache.camel.service.lra;
 
-import java.net.URISyntaxException;
-import java.nio.charset.StandardCharsets;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-
 import org.apache.camel.Exchange;
-import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.builder.RouteBuilder;
-import org.apache.camel.util.URISupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import static org.apache.camel.service.lra.LRAConstants.PARTICIPANT_PATH_COMPENSATE;
-import static org.apache.camel.service.lra.LRAConstants.PARTICIPANT_PATH_COMPLETE;
-import static org.apache.camel.service.lra.LRAConstants.URL_COMPENSATION_KEY;
-import static org.apache.camel.service.lra.LRAConstants.URL_COMPLETION_KEY;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static org.apache.camel.service.lra.LRAConstants.*;
 
 public class LRASagaRoutes extends RouteBuilder {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LRASagaRoutes.class);
 
     private LRASagaService sagaService;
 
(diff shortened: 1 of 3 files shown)


zhfeng:
It seems that when seding the LRA Join request, there is a wrong context-text header, maybe it need to change to `application/x-www-form-urlencoded` since the payload is encodeded. See https://github.com/apache/camel/blob/2e59dce41273cc83bd0f2a79d6b1215bce9f5e67/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAClient.java#L130-L135

Can you share the Oracle LRA coordinator configurations? and I could use to reproduce this issue locally.

zhfeng:
Hmm, it looks like the payload is redundant  since there is a link header. So can you try to use
```java
.PUT(HttpRequest.BodyPublishers.ofString(""))
```

to see if it works with Oracle LRA coordinator?

DFiedler:
this is our current oracle lra-coordinator config .. 
.. i will try your suggestion on Monday ;o)

have a nice weekend!
[configmap-tcs-config.yaml.txt](https://github.com/user-attachments/files/17001712/configmap-tcs-config.yaml.txt)
[statefulset-otmm-tcs.yaml.txt](https://github.com/user-attachments/files/17001713/statefulset-otmm-tcs.yaml.txt)

DFiedler:
Hi,

when i am changing the the code to:

`.PUT(HttpRequest.BodyPublishers.ofString(""))`

then i am still getting this error message from LRA-Coordinator:

`error while joining LRA: 400 / error: URL is double encoded`

The same error (still) occurs, if i change the header:

`.setHeader(CONTENT_TYPE, "application/x-www-form-urlencoded")`

best regards
Dirk

zhfeng:
@DFiedler sorry for the late reply since there are the public holidy on Mon and Tue in China.

Well, how can I run the Oracle LRA coordinator on my local machine? is there any docker image? It would be very helpful if you can share a simple reproducer. I'd like to take a deep look.

DFiedler:
@zhfeng  we are using OpenShift as the preferred runtime, u can also use Docker: https://docs.oracle.com/en/database/oracle/transaction-manager-for-microservices/24.2/tmmdg/supported-operating-systems.html

because oracle also implements the Eclipse Microprofile LRA  specification, it should work out of the box instead of Narayana (i guess u are using it for lra-coordinator) ... 

In our case this is the lra-coordinator base path (oracle microtx url): http://otmm-tcs.qa-servicebuilder-dev.svc.cluster.local:9000
and this the lra-coordinator context path: api/v1/lra-coordinator

zhfeng:
Thanks @DFiedler - I can start a oracle coordinator locally and reproduce the issue.

I'd like to discuss this issue with the Narayana team to see if there is any feedback. So let's hold the PR.

components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAUrlBuilder.java
```diff
@@ -91,18 +91,12 @@ public LRAUrlBuilder options(Map<String, ?> options) {
...
-            }
-            copy.query += key + "=" + value;
-        } catch (UnsupportedEncodingException e) {
-            throw new IllegalStateException(e);
+        if (copy.query.isEmpty()) {
+            copy.query += "?";
+        } else {
+            copy.query += "&";
         }
+        copy.query += key + "=" + value;
```
zhfeng, line 99:
```suggestion
        copy.query += toNonnullString(key) + "=" + toNonnullString(value);
```

zhfeng:
Hi @DFiedler 

After meeting the narayana team, it looks like

- The communications between the LRA client and coordinator is not a part of the MP LRA specification
- The `Link` header format is private used in Narayana LRA internally
- Oracle MicroTx LRA seems following the similar way with Narayana but we can not expect it can be OOTB
- The changes to **NOT** encoding the query paramters in `Link` is OK

davsclaus:
@zhfeng can you help get the last bits in place so this can be merged so it works with both oracle and narayana

zhfeng:
@davsclaus sure, I can.

@DFiedler Do you mind I open a new PR to supersede this one?

DFiedler:
@zhfeng I want to ask, what is the next step here .. can i do anything to finalize this PR?

zhfeng:
@DFiedler 

- Please take a look at my above suggestion on `LRAUrlBuilder.java` 
- Have a new method to replace `URISupport.parseQuery` to get the parmeters from HTTP and decode them
- Maybe run `./mvnw -f components/camel-lra/pom.xml process-sources` to format the files and make the CI happy.

components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAUrlBuilder.java
zhfeng: @DFiedler please take a look this suggestion.

DFiedler: hi @zhfeng .. i followed your suggestion and adjusted the code

components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
```diff
@@ -87,12 +88,18 @@ private void verifyRequest(Exchange exchange) {
                 if (!queryParams.isEmpty()) {
                     if (queryParams.get(URL_COMPENSATION_KEY) != null) {
                         compensationURI = queryParams.get(URL_COMPENSATION_KEY).toString();
+                        // CAMEL-21216: the call from the lra-coordinator is not correctly url-decoded
+                        // as long as 'CAMEL-21197' is not solved, this workaround is needed
+                        compensationURI = java.net.URLDecoder.decode(compensationURI, StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
DFiedler: @zhfeng 
i implemented a complete new parse method (parseQuery) in the local class to avoid a new dependency to other libs like 'org.apache.http.client.utils.URLEncodedUtils', i also set up a small junit test for it .. 

What do u think?

zhfeng:
@DFiedler Good job!

Did you check these changes work with Oracle MicroTx as well?

I wonder if we can have such a integration test with the MicroTx and it can be in a seperate PR.

zhfeng:
Maybe we can have a similar MircoTx test-infra like https://github.com/apache/camel/tree/main/test-infra/camel-test-infra-microprofile-lra

zhfeng (approved):

DFiedler:
@zhfeng  thx for approving .. 
One additional question, what are the release cycles for 4.8 and for next 4.9 release .. because 4.8 is only two weeks ago ..

zhfeng:
I will backport this fix to `camel-4.8.x` and it should be in the next release of `4.8.x`. I'm not very sure about the relase cycle and @davsclaus might know more.

PR #15610, opened 2024-09-18
Use the newly maintained univocity parsers

gnodet:
* remove locking for lazy initialized fields, as those are not very expensive to create, so worst case, we have double initiatization, which is not a problem
* move the TLS into the HeaderRowProcessor
* remove 'or not' in 'whether or not'


davsclaus:
You need to remove the deprecation in the pom.xml

>     <name>Camel :: UniVocity Parsers (deprecated)</name>

And then do a rebuild as other files are changed - such as in dsl / catalog etc

davsclaus (approved):

PR #15637, opened 2024-09-20
CAMEL-21244: camel-clickup (webhooks only)

nickshoe:
# Description

The pull request contains the code for a new component (`camel-clickup`) that is capable of receiving webhook events from [ClickUp](https://clickup.com/). The component is a delegate for the `camel-webhook` meta-component.

The current version of the component only support webhooks related functionality; in particular, it only supports the `taskTimeTrackedUpdated` event, but it can be easily further exteded by implementing the other events DTOs. It already register/unregister webhooks automatically on ClickUp by using ClickUp APIs.

# Target

# Tracking


catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/clickup.json
```diff
@@ -0,0 +1,41 @@
+{
+  "component": {
+    "kind": "component",
+    "name": "clickup",
+    "title": "ClickUp",
+    "description": "Receives events from ClickUp webhooks.",
+    "deprecated": false,
+    "firstVersion": "1.0.0-SNAPSHOT",
```
oscerd, line 8: This should be 4.9.0-SNAPSHOT

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/ClickUpWebhookProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.clickup;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.camel.*;
```
oscerd, line 23: Please don't use * imports.

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/model/Events.java
```diff
@@ -0,0 +1,96 @@
...
+ * <p>
+ * Space webhooks - spaceCreated: Triggered when a new Space is created. - spaceUpdated: Triggered when an existing
+ * Space is updated. - spaceDeleted: Triggered when a Space is deleted.
+ * <p>
+ * Goal and Target (key result) webhooks - goalCreated: Triggered when a new Goal is created. - goalUpdated: Triggered
+ * when an existing Goal is updated. - goalDeleted: Triggered when a Goal is deleted. - keyResultCreated: Triggered when
+ * a new Target is created. - keyResultUpdated: Triggered when an existing Target is updated. - keyResultDeleted:
+ * Triggered when a Target is deleted.
+ */
+public final class Events {
```
oscerd, line 52: This is something nice to have, but it will be very hard to maintain, if they deprecate/add/update events. Something to think about for the future is autogenerating the events.

oscerd (approved):
Just some minor findings, but it's good for me. Waiting for others to review.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/clickup.json
nickshoe: It's strange, since that code resulted from the `mvn clean install -DskipTests` command execution.
I'll try to re-execute it once more and see if it gets generated the right way.

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/model/Events.java
nickshoe: Ok, I see your point. Public APIs such as ClickUp's should never change contract with breaking changes.. ClickUp's APIs are versioned (current version is `v2`). Anyways, should I remove webhooks events deserialization altogether? In that case, shall I pass `String` objects to the downstream processors?

nickshoe:
@oscerd I've added some tests. They conver the main webhook functionalities (register, unregister, call).

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/model/Events.java
nickshoe: I like the autogeneration approach, that would be possible by leveraging ClickUp's OpenAPI spec. In any case, though, the problem with unexpected APIs changes would persist. What was your idea?

oscerd: This could be an improvement for the future, it's not mandatory now. OpenAPI spec could be a good way.

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/clickup.json
nickshoe: @oscerd it's not generating the proper version number... maybe it's due to some automatic mechanism for calculating the first version.. could it be that it uses `1.0.0-SNAPSHOT` because the `4.9.0-SNAPSHOT` it's not yet released? Should I edit all the occurrences the version manually? Thanks

components/camel-clickup/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+        </dependency>
+
+        <!-- Google Guava -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>33.3.0-jre</version>
```
tadayosi, line 64: Shouldn't this refer to a version defined in the parent pom?

tadayosi (approved):
LGTM

components/camel-clickup/pom.xml
nickshoe: @tadayosi I'll check it right away!

nickshoe: Addressed in https://github.com/apache/camel/pull/15637/commits/fb812bd01012453de9d1b4c5bd091e3f9e304aad

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/clickup.json
davsclaus: Set first version in `@UriEndpoint`

nickshoe: @davsclaus it's already there... 

https://github.com/nickshoe/camel/blob/472cc0ecbbefff39e0cfb7b96781dbd079102a97/components/camel-clickup/src/main/java/org/apache/camel/component/clickup/ClickUpEndpoint.java#L41

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/model/Events.java
nickshoe: @oscerd another tought regarding event deserialization: even though we'd use plain strings instead of typed obejcts, the problem with breaking changes in the event payloads still exists; there's no guarantee that the ClickUp API (or whatever 3rd party vendor's API) will provide a retro-compatible data structure.. A precaution I've put in place in the DTOs is the `@JsonIgnoreProperties(ignoreUnknown = true)` Jackson option, that should at least make the deserialization process to work even if the vendor adds new keys to an existing schema.

nickshoe: Idea: the typed deserialization could be opted-out with a component/endpoint configuration param..

davsclaus (approved):

components/camel-clickup/src/main/java/org/apache/camel/component/clickup/ClickUpEndpoint.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.camel.spi.UriEndpoint;
+import org.apache.camel.spi.UriParam;
+import org.apache.camel.support.DefaultEndpoint;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Receives events from ClickUp webhooks.
+ */
+@UriEndpoint(firstVersion = "4.9.0-SNAPSHOT", scheme = "clickup", title = "ClickUp", syntax = "clickup:workspaceId",
```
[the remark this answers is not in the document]
davsclaus: Yeah its a bit weird - though you need to set consumerOnly = true as this component is only for consumers. Also you can set firstVersion in pom.xml under <properties>.

davsclaus:
This PR is getting big, can we finish up the last bits. And then any more work do this in new PRs.

nickshoe:
@davsclaus ok, I'll just fix the `firstVersion` thing and leave the PR as it is.

nickshoe:
@davsclaus OK, it's all sorted now. 

It seems that, having set `consumerOnly = true`, the `mvn clean install -DskipTests` command regenerated the files with the wrong version number (maybe Maven wasn't regenerating them because of some caching mechanism... ?!).

I will not make any more changes unless I am asked.

oscerd:
I guess we can merge and then iterate if needed.

PR #15638, opened 2024-09-20
Zipsplitter null body

thomas-gantenbein-tga:
# Description
See https://camel.zulipchat.com/#narrow/stream/257298-camel/topic/Idiomatic.20way.20to.20pollEnrich.20until.20there.20are.20no.20more.20items: In case one of the incoming bodies is null, this change should avoid failure of ZipAggregationStrategy.

I also addressed SonarQube warnings about cognitive complexity and removed deprecated annotations from test code while I was at it.

Viewing the three commits individually is probably best to see what changed in terms of functionality and what was just refactoring.


davsclaus (approved):

davsclaus:
I wonder if you would check `camel-tarfile` if the agg strategy there has similar issue

PR #15651, opened 2024-09-23
feat(jbang): make export reproducible

squakez:
Added a build propertues required as specified in https://maven.apache.org/guides/mini/guide-reproducible-builds.html

# Description

# Target

# Tracking


davsclaus (approved):

davsclaus:
I wonder if adding a boolean option so users can turn this off of not desired

squakez:
> I wonder if adding a boolean option so users can turn this off of not desired

I can do that but it would be complicating the export subcommand with yet another option. IMO this should be a best practice we enforce since prototype as it neither have any side effect on the final build.

davsclaus:
> > I wonder if adding a boolean option so users can turn this off of not desired
> 
> I can do that but it would be complicating the export subcommand with yet another option. IMO this should be a best practice we enforce since prototype as it neither have any side effect on the final build.

Yeah its something that can always be added later, so its good as-is

PR #15653, opened 2024-09-23
CAMEL-21252 Add checking for non .*-versions to sync-properties-maven-plugin

cunningt:
# Description

Noticed today that sync-properties-maven-plugin will not sync a property where the format is `<project>.<version>`, but only syncs `<project>-<version>`.      Added a check to sync-properties-maven-plugin that errors if a property is added to camel-parent of the "project.version" format with an explanation so we can catch these, and changed the two properties that are not being synced from camel-parent to camel-dependencies to match the correct version.

### DIFF
--- a/camel-dependencies/pom.xml
+++ b/camel-dependencies/pom.xml
@@ -63,6 +63,9 @@
                         <goals>
                             <goal>sync-properties</goal>
                         </goals>
+                        <configuration>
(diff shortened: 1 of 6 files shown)


oscerd (approved):

davsclaus (approved):

oscerd:
I guess we need to backport to 4.8.x and 4.4.x

PR #15672, opened 2024-09-24
[CAMEL-21238] Reloading in k8s run does not change route behaviour

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21238


davsclaus (approved):

davsclaus:
The CI server takes a long time to test, 90 min so cancelling that job

davsclaus:
Hi @tdiesler your PR actually caused other tests to fail, as it seems like the export append `-000` inside `.camel-jbang-run` folder or some number to the name, so for example kubernetes delete test cannot find the project - because it does not know about the `-000` stuff.

Can you help fix these tests

PR #15678, opened 2024-09-24
CAMEL-21256: camel-github startingSha=last does not work with large history

electricsam:
# Description

I encountered an issue using startingSha=last with a repository with a large history.  The last sha is not correctly resolved.   [CAMEL-17473](https://issues.apache.org/jira/browse/CAMEL-17473) did not resolve this issue.  

The following issues were discovered from real-world testing and a unit test:

1. it appears that the first commit returned from the GitHub API is the most recent and the sha from that should be used as the lastSha variable.
2. There is a threading issue where CommitConsumer.poll() can be called concurrently with CommitConsumer.doStart().  In some situations lastSha can be null in the poll() method, producing the same behavior as startingSha=beginning, which is undesired.

# Target

# Tracking

### DIFF
--- a/components/camel-github/src/main/java/org/apache/camel/component/github/consumer/CommitConsumer.java
+++ b/components/camel-github/src/main/java/org/apache/camel/component/github/consumer/CommitConsumer.java
@@ -46,7 +46,7 @@ public class CommitConsumer extends AbstractGitHubConsumer {
     private boolean started = false;
 
     public CommitConsumer(GitHubEndpoint endpoint, Processor processor, String branchName,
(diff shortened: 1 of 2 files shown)


davsclaus:
Can you update the PR as there are formatting changes

	modified:   components/camel-github/src/main/java/org/apache/camel/component/github/consumer/CommitConsumer.java
	modified:   components/camel-github/src/test/java/org/apache/camel/component/github/consumer/CommitConsumerLastTest.java

oscerd (approved):

PR #15697, opened 2024-09-25
CAMEL-21323: upgrade camel-http to 5.3

jonomorris:
# Description

1. We needed to set the uri-authority on the request via the `RequestValidateHost` request interceptor for the request-router that's been added to `Serverbootstrap`, and also the 'CanonicalHostName'. 

2. The default charset in `EntityUtils` has changed to UTF-8 from iso8859-1, and if we want to use iso8859-1 encoding we need to set the 'content-type' header, e.g. "text/plain;charset=iso8859-1".

3. HTTP 1.1 is now the default version in `Http1Config`, and this is set on http-requests.

# Target

- [x ] I checked that the commit is targeting the correct branch (note that Camel 3 uses `camel-3.x`, whereas Camel 4 uses the `main` branch)

# Tracking

### DIFF
--- a/components/camel-http/src/test/java/org/apache/camel/component/http/CamelComponentVerifierExtensionTest.java
+++ b/components/camel-http/src/test/java/org/apache/camel/component/http/CamelComponentVerifierExtensionTest.java
@@ -17,7 +17,6 @@
 package org.apache.camel.component.http;
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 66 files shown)


davsclaus (approved):

davsclaus:
There is even a 5.4 release I think. You are welcome to take a look at that after this PR is merged.

jonomorris:
Looks like I don't have access to merge.

davsclaus:
Hmm not sure what that requires, if you are in the ASF GH group which it seems if you hover mouse on your name. Maybe @oscerd can remember what it would need.

PR #15700, opened 2024-09-25
CAMEL-21263 + CAMEL-21264

orpiske:


components/camel-smb/src/main/java/org/apache/camel/component/smb/converter/SmbConverter.java
```diff
@@ -0,0 +1,41 @@
...
+@Converter(generateLoader = true)
+public class SmbConverter {
+
+    @Converter
+    public InputStream toInputStream(com.hierynomus.smbj.share.File smbFile) {
+        return smbFile.getInputStream();
+    }
+
+    @Converter
+    public byte[] toByteArray(com.hierynomus.smbj.share.File smbFile) {
```
davsclaus, line 34: I wonder if there should also be a toString converter that reads all the bytes and adds into a string. The charset can be taken from exchange (see maybe some of the other converters in core)

orpiske: I think there's "risk" here in the sense that the files are not text ones. So, it could make it easy for the users to confuse things. I think the `byte[]` one kinda forces them to deal with that appropriately.

wdyt?

davsclaus: ok

orpiske:
BTW, I think this is missing doc updates. I will also include this here ... just sending the code earlier for a quicker review.

davsclaus (approved):

components/camel-smb/src/test/java/org/apache/camel/component/smb/SmbComponentConnectionIT.java
```diff
@@ -0,0 +1,79 @@
...
+
+        mock.assertIsSatisfied();
+    }
+
+    @Override
+    protected RouteBuilder createRouteBuilder() throws Exception {
+        return new RouteBuilder() {
+            private void process(Exchange exchange) throws IOException {
+                final byte[] data = exchange.getMessage().getBody(byte[].class);
+                final String filePath = exchange.getMessage().getHeader(SmbConstants.SMB_FILE_PATH, String.class);
```
llowinge, line 57: @orpiske Could you add this to the example in the docs ? I have a feeling it still fits under the "absolute basic" scenario and user would like to see which file is being consumed.

orpiske: Good point. Let's include this!

PR #15707, opened 2024-09-25
CAMEL-21286 - Add a cache for JAXB ObjectFactory lookup

ammachado:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-jaxb/src/main/java/org/apache/camel/converter/jaxb/JaxbHelper.java
+++ b/components/camel-jaxb/src/main/java/org/apache/camel/converter/jaxb/JaxbHelper.java
@@ -17,12 +17,16 @@
 package org.apache.camel.converter.jaxb;
 
 import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


oscerd (changes requested):
It would make sense to make the dimension selectable

ammachado:
@oscerd and how you'd configure that? There are other options on `FallbackConverter`, but they're implemented using Camel's global context options. Should I use a similar pattern?

oscerd:
I think there is no value in adding an arbitrary dimension here. It could work for your case but it could be useless for others. So we need to find a good way @davsclaus

davsclaus (approved):

davsclaus:
This feature is only relevant if you turn on a special option so it will only affect a small set of users. And the cache will speedup in those cases. Setting a dimension is likely only needed in rare cases if you have a lot of different JAXB classes that you process by Camel, causing the cache to run out of space.

PR #15712, opened 2024-09-26
Add breaking change note for Camel JBang repos parameter in Migration

apupier:
guide

for reference, it was changed in this commit
https://github.com/apache/camel/commit/23da1eb48096fed69fbf8becfc72b3541e33c9dd

# Description

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

apupier:
Just discovered that it wasn't changed for all commands, so need to be more precise

for instance https://github.com/apache/camel/blob/dfa55cc0574f8af576723a4969167af63c24fc9a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/catalog/CatalogBaseCommand.java#L63

I'm wondering if we should not support both --repos and --repository for 4.8.1 and 4.9 to let time to untangle everything

PR #15716, opened 2024-09-26
CAMEL-21152: Remove camel k dependency from camel-jbang kubernetes plugin

gansheer:
# Description

* Add Traits model classes specific to camel-jbang kubernetes plugin instead of camel-k one
* Modify camel-jbang kubernetes plugin commands and helpers
* Adapt camel-jbang k plugin that use camel-jbang kubernetes plugin common code

# Target

# Tracking


christophd (approved):
A bit sad to loose the model builders. maybe we can bring those back in a follow-up?

gansheer:
> A bit sad to loose the model builders. maybe we can bring those back in a follow-up?

Good idea. I also know there is some possible useless code (some trait parameters) that could probably be deleted soon.
I am planning a follow up to work more on what should really be present only in the kubernetes plugin.

gansheer:
Follow-up created : https://issues.apache.org/jira/browse/CAMEL-21278

davsclaus (approved):

PR #15717, opened 2024-09-26
CAMEL-21283 - Switch back to repos parameter in Camel JBang run and export commands

apupier:
To stay coherent with other commands and keep working when using --camel-version using a previous version.

Removed the --repository option which would require a bunch of other changes to support it correctly

# Description

# Target

# Tracking

### DIFF
--- a/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_8.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_8.adoc
@@ -32,7 +32,7 @@ if the parameter `includeProperties` is set to true.
 
 The `camel trace` command has changed to show tracing status (by default). To dump traced messages use `camel trace --action=dump`.
 
-*Breaking changes* The parameter `--repos` has been renamed `--repository` for `run` and `export` commands. It requires to be updated when using from command-line and in `application.properties`.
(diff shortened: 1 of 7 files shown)


docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_8.adoc
```diff
@@ -32,7 +32,7 @@ if the parameter `includeProperties` is set to true.
 
 The `camel trace` command has changed to show tracing status (by default). To dump traced messages use `camel trace --action=dump`.
 
-*Breaking changes* The parameter `--repos` has been renamed `--repository` for `run` and `export` commands. It requires to be updated when using from command-line and in `application.properties`.
+*Breaking changes only in 4.8.0* The parameter `--repos` has been renamed `--repository` for `run` and `export` commands. It requires to be updated when using from command-line and in `application.properties`. In 4.8.1, the `--repos` has been set back and  `--repository` removed.
```
apupier, line 35: It implies the removal of an option from a LTS version  but given that:
* it was introduced just for this version
* it was broken for several cases
* changes were not applied to other commands than run and export, meaning it was not coherent across commands
* to apply allow changes
* it was initially done in an effort to be able to provide parameter using style `--repository=xxx --repository=yyy` instead of `--repository=xxx,yyy` but it was already reverted during 4.8 development
I hope it can be acceptable to break again the Camel JBang command-line API and get back to previous state.

davsclaus (approved):

PR #15729, opened 2024-09-26
CAMEL-21269: camel-langchain4j-tokenizer fixes for the YAML DSL

orpiske:
This is a more more comprehensive fix for the `tokenizer` support for the YAML DSL.  There were quite a few problems in the model files that prevented it from being used with YAML (thus affecting any sub-project or project relying on this). 

Although it's relatively big due to the regen of a few files, all the changes are backwards compatible and anyone using it with the Java DSL should not have to change any code.


oscerd (approved):

dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl.json
```diff
@@ -7267,6 +7267,100 @@
...
+            "type" : "boolean",
+            "title" : "Disabled",
+            "description" : "Whether to disable this EIP from the route during build time. Once an EIP has been disabled then it cannot be enabled later at runtime."
+          },
+          "id" : {
+            "type" : "string",
+            "title" : "Id",
+            "description" : "Sets the id of this node"
+          },
+          "langChain4jCharacterTokenizer" : { },
```
davsclaus, line 7291: This looks a bit suspesious that these are here and empty

orpiske: Hm ... interesting (maybe that's why we are not seeing them on the dropdown, @lordrip). 

Thanks @davsclaus, let me investigate this.

orpiske: @davsclaus upon closer inspection this seems fine. We have the same thing for the Marshal definition (which was what I based myself for adjusting this): https://github.com/apache/camel/blob/db7a2bb33eb9d2501134ca9b393d7281d7a1c56f/dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl.json#L3251-L3257

davsclaus: okay thanks

davsclaus (approved):

davsclaus:
You should also ping @mgubaidullin as it can affect Camel Karavan

orpiske:
> You should also ping @mgubaidullin as it can affect Camel Karavan

Absolutely. I couldn't reproduce w/ the existing version of Karavan, but even if it got everything correctly, the generated YAML would still be wrong.

It's on my TODO list for next week to figure out if/how Karavan is affected so I can pinpoint exactly where/how we need to fix any code relying on the incorrect model.

orpiske:
I've just confirmed that the YAML itself is functional. Now I'll work with the users / tooling creators to make sure they can leverage the new EIP. 

Let's merge this to make it easily available for others.

PR #15745, opened 2024-09-28
Cleanup dependency management

gnodet:
# Description

# Target

# Tracking


davsclaus:
This may have caused camel-jbang-core to not compile

[ERROR] /Users/davsclaus/workspace/camel/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/RunHelper.java:[47,47] cannot access org.codehaus.plexus.util.xml.pull.XmlPullParserException
[ERROR]   class file for org.codehaus.plexus.util.xml.pull.XmlPullParserException not found

davsclaus:
okay it seems it causes more problems - we should consider reverting this.
tests in camel-jbang-core is failing, and the jbang is not working correctly.

I am trying again tomorrow to be sure, as I want to do a full rebuild with and without this commit

gnodet:
> okay it seems it causes more problems - we should consider reverting this. tests in camel-jbang-core is failing, and the jbang is not working correctly.
> 
> I am trying again tomorrow to be sure, as I want to do a full rebuild with and without this commit

That could definitely be the cause.  I did a full successful rebuild, but without running tests...

davsclaus:
Okay I did 2 full rebuilds this morning

- main (as is) = fails
- reverted = ok

I am going to revert this, and @gnodet maybe you can find out the problem or maybe do less of a cleanup to narrow down where it starts to fail

PR #15789, opened 2024-10-01
fix(jbang): include platform-http dependency

squakez:
only when necessary

Closes CAMEL-21294

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportQuarkus.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportQuarkus.java
@@ -426,7 +426,7 @@ class ExportQuarkus extends Export {
         StringBuilder sb = new StringBuilder();
         for (MavenGav gav : gavs) {
             //Special case, quarkus-pom.tmpl already have them included.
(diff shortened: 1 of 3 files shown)


davsclaus:
if you have a rest app then you need platform-http. I wonder if you could try export these kinds before merging.

davsclaus (approved):

davsclaus:
Also the openapi examples would be using platform-http such as
https://github.com/apache/camel-kamelets-examples/tree/main/jbang/open-api-contract-first
https://github.com/apache/camel-kamelets-examples/tree/main/jbang/open-api

squakez:
> Also the openapi examples would be using platform-http such as https://github.com/apache/camel-kamelets-examples/tree/main/jbang/open-api-contract-first https://github.com/apache/camel-kamelets-examples/tree/main/jbang/open-api

I tested with and without the platform. Sure, let me try testing with rest and openapi as well. I feel it should work because it is normally working with main, but let's make sure with some test.

davsclaus:
I am wonder about whether export to Q and SB works for these examples

squakez:
I've tested REST and Openapi examples against all runtimes. The openapi contract first is suffering this problem https://issues.apache.org/jira/browse/CAMEL-21301 - but it seems it was an existing regression as I've tested with released 4.8.0 and it is getting the very same result. I'm going to merge this and solve the other separately as they does not seem to be related.

PR #15810, opened 2024-10-02
CAMEL-21277: use WrappedFile in camel-smb for greater consistency

orpiske:


davsclaus (approved):

davsclaus:
I wonder if the file name should be stored in Exchange.FILE_NAME also - And do we know the length of the file then that can be in the wrapped file also. See how camel-ftp does some of that.

orpiske:
> I wonder if the file name should be stored in Exchange.FILE_NAME also - And do we know the length of the file then that can be in the wrapped file also. See how camel-ftp does some of that.

+1 for the File Name (maybe even back-port this one to 4.8 to keep consistent). 

Let me take a look at providing the file size as well.

orpiske:
@davsclaus I included both the name and the size. Later I'll backport the name part only to 4.8.

PR #15816, opened 2024-10-03
CAMEL-21309: camel-cxf - Force using sync client when using tracing/opentelemetry as otherwise spans are not working correctly.

davsclaus:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
@@ -59,6 +59,8 @@ import org.apache.cxf.jaxrs.client.WebClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.camel.Exchange.ACTIVE_SPAN;
+
 /**
  * CxfRsProducer binds a Camel exchange to a CXF exchange, acts as a CXF JAXRS client, it will turn the normal Object
  * invocation to a RESTful request according to resource annotation. Any response will be bound to Camel exchange.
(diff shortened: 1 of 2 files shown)


components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
johnpoth, line 1: It seems `exchange.getProperty(ACTIVE_SPAN)` always returns null. 

`exchange.getProperty(ExchangePropertyKey.ACTIVE_SPAN)` seems to be what we need

components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfProducer.java
johnpoth, line 1: Same, It seems `exchange.getProperty(ACTIVE_SPAN)` always returns null. 

`exchange.getProperty(ExchangePropertyKey.ACTIVE_SPAN)` seems to be what we need

johnpoth (changes requested):

components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
davsclaus: It was lacking a mapping
https://github.com/apache/camel/commit/a290693bd4aadcf188b1c0947a531a26e6c1f73c

But we can optimize to use the key constant as suggested

davsclaus: ok changed

johnpoth (approved):

johnpoth:
LGTM !

PR #15830, opened 2024-10-03
RFC (Camel 4.10) CAMEL-21080: removing legacy methods

orpiske:


squakez (commented):
I wonder if we can have a deprecation notice time and make the legacy coexists while the new methods are working. In this way we guarantee backward compatibility and warn the user to move their test to the new method.

orpiske:
> I wonder if we can have a deprecation notice time and make the legacy coexists while the new methods are working. In this way we guarantee backward compatibility and warn the user to move their test to the new method.

Unfortunately the `@deprecated` doesn't have a field to add that, but maybe we could include the `forRemoval=true` on the `@deprecated` annotations we currently have on `CamelTestSupport.

orpiske:
I've also been reflecting a bit on this and I'm starting to think that 4.10 is way too short of a notice for users. We are currently reaching a maturity level on 4.x where many users will be looking more seriously to migrate. 

It may not be a good idea to drop such another task on top of the other things they have to validate when moving from 2.x and 3.x all the way to 4.x. We always intended 4.x to be as close as a drop-in replacement as possible, so I am a bit conflicted here.

orpiske:
After some consideration, I have decided to halt this change. What we have now is enough of a cleanup and, IMHO, removing these methods would unnecessarily violate one of the goals we had for Camel 4 (to be as close as possible a drop-in replaced for 3.x). Additionally, I think there are other areas in and around Camel that need more focus right now. 
I might revive this some point in the future when/if it becomes safe to remove these methods.

Therefore, closing this one for now.

PR #15878, opened 2024-10-08
Use latest released version

Croway:
Mimic main behaviour


oscerd (approved):

davsclaus:
We were trying to make it ready for the next release so when gregor builds 4.4.4 it was hardcoded with correct version, so when you install camel jbang via some github url where you say @4.4.4 (or something) it would use that version.

But I frankly think we just need to make a better CLI command to set the jbang version in this fille, and not rely on weird/complex installation via jbang.

PR #15913, opened 2024-10-10
CAMEL-21249 - Camel-Kamelets: Move Kamelets utils in Camel Kamelets c…

oscerd:
…omponent

# Description

# Target

# Tracking


components/camel-kamelet/pom.xml
```diff
@@ -43,6 +43,54 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-core-model</artifactId>
         </dependency>
+        <dependency>
```
oscerd, line 46: I choose to set both, components and 3rd party dependency as provided, because they should be declared in the Kamelet YAML using those actions. If you have other idea let me know.

davsclaus (approved):

davsclaus:
Yeah we can move things later if we come up with a better plan. 

I wonder if that aws ddb data transformer can be moved to camel-aws instead - it belongs there and not really here in kamelets. This can be done in another PR.

Also we need to check a bit with jbang as camel-kamelets have extra dependencies now, although they are provided scope. So just to make sure it does not cause that jbang downloads all of these, when they are not in use.
However I think it would already skip provided scope, but we need to check after merging this PR.

oscerd:
Ok, I'm going to merge this and then move the AWS part.

PR #15915, opened 2024-10-10
CAMEL-21249 - Camel-Kamelets: Move Kamelets utils in Camel Kamelets component - Move DDB2 part to AWS DDB

oscerd:
# Description

# Target

# Tracking


davsclaus (approved):

components/camel-aws/camel-aws2-ddb/pom.xml
```diff
@@ -62,6 +62,12 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-gson</artifactId>
+            <scope>provided</scope>
```
davsclaus, line 68: I wonder if the other components with this data type transformer has optional for some JARs?
If we dont really do that, then its likely better to just have it normal so everything can work out of the box. And instead of camel-gson, it can dependet on just google-gson

oscerd: I'll fix that. Thanks.

oscerd:
Done. For the Jackson dependency I need to check if I can remove the component

PR #15925, opened 2024-10-11
[CAMEL-21310] camel-jackson - Upgrade to 2.18

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-21310


oscerd (approved):

davsclaus:
It would also be a good to have another PR to add doc etc about that new 4-byte feature option you helped get into jackson. And we could consider making that easier to turn on that currently

PR #15929, opened 2024-10-11
CAMEL-21343: camel-seda - Allow to mimic the behavior of camel-vm

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-21343

## Motivation

The component camel-vm is still frequently used by the community so even if it has been decided to remove it from camel, we need to provide a way to mimic its behavior to ease the migration to recent camel versions.

## Modifications:

* Add methods meant to be overridden to mimic the behavior of camel-vm

### DIFF
--- a/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
+++ b/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.camel.AsyncCallback;
 import org.apache.camel.AsyncProcessor;
 import org.apache.camel.Exchange;
+import org.apache.camel.ExchangeExtension;
 import org.apache.camel.Processor;
 import org.apache.camel.ShutdownRunningTask;
 import org.apache.camel.Suspendable;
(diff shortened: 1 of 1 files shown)


essobedo:
@davsclaus Hi 👋, I'm not really familiar with the camel-vm component, would you mind reviewing my PR to ensure that it makes sense?

components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
```diff
@@ -215,15 +210,38 @@ protected void doRun() {
...
+    /**
+     * Strategy to invoke when the exchange is done being processed.
+     * <p/>
+     * This method is meant to be overridden by subclasses to be able to mimic the behavior of the legacy component
+     * camel-vm, that is why the parameter {@code prepared} is not used by default.
+     *
+     * @param original the exchange before being processed
+     * @param prepared the exchange processed
+     */
+    protected void onProcessingDone(Exchange original, Exchange prepared) {
```
davsclaus, line 222: You do not need to have 2 parameters here, it can just be named exchange, and it should be the `prepared` exchange that is being called from the callback.

essobedo: Interesting feedback, I was not sure TBH, I just wanted to allow to call `copyResults` like before the change https://github.com/apache/camel/commit/ed5cc8a6c9f54a4fdf25a2eed03fa528c8b0f9a1#diff-495571c4ab51f8270c10e0c9d149c351014465883804a481440403ebe4e4fc4dL190, it is not needed anymore?

davsclaus: ah yeah that is actually needed since this logic is now in the callback (async compatible) instead of afterwards sendToConsumers.

adessaigne:
Thank you very much @essobedo and @davsclaus 👍

PR #15955, opened 2024-10-15
Fix CVE

Croway:
@davsclaus the latest released jackson-dataformat-avro:2.17.2 is not compatible with avro:1.12.0.

Do we really need avro:1.12.0 in Camel? would it be possible to downgrade it to 1.11.4?

### DIFF
--- a/.github/actions/install-packages/action.yml
+++ b/.github/actions/install-packages/action.yml
@@ -22,5 +22,5 @@ runs:
   steps:
     - run: sudo apt-get update
       shell: bash
-    - run: sudo apt-get install -qqy --no-install-recommends libtinfo5
+    - run: sudo apt-get install -qqy --no-install-recommends libtinfo6
       shell: bash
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (approved):

davsclaus:
ah yeah we should use 1.11.x that works better with jackson. Camel 4.9 is jackson 2.8.x and that seems to be working with the newer avro.

davsclaus:
So can we try to downgrade to 1.11.4 and then have this override fix in jackson-avro so it also use that same version - would that not be the best solution for 4.8.x ?

Croway:
> So can we try to downgrade to 1.11.4 and then have this override fix in jackson-avro so it also use that same version - would that not be the best solution for 4.8.x ?

Yes, I agree, I'll update avro version in this PR

Croway:
I've downgraded avro to 1.11.4 and all tests are successful

davsclaus (approved):

PR #15958, opened 2024-10-15
[CAMEL-21199] Camel-jackson not properly marshalling 4-byte characters

rnetuka:
Issue: https://issues.apache.org/jira/browse/CAMEL-21199

In Jackson 2.18-, when using marshalling with UTF-8 generator (which is used undercover by Camel), they return 4-byte characters as a surrogate pair, escaped. This PR uses the new feature in 2.18 which, when turned on, combines the surrogates into a one character when before flushing the output.

Switching the UTF-8 generator with writer, which was providing the same job, but might have performace impact, is no longer needed.

I'd love to hear your opinion on these changes.

If you confirm, I'd suggest to:
* remove JacksonMarshalUseWriterTest, since I believe it was added only for the temporary solution with the writer and is no longer needed
* enabling combineUnicodeSurrogates by default since I believe most Camel users working with 4-byte characters because of their native language would prefer this behavior

Also, I'd like to ask you to provide me with a docs link where I can describe this behavior.

Thank you!


davsclaus:
This requires changes to core/camel-core-model as useWriter is an option there. So you can change the option `useWriter` to be `combineUnicodeSurrogates`. And then there are some metadata files updated due to this, and this requires a regen of the source code - camel-catalog etc.

davsclaus:
Whether the option is default enabled - that is something we can do afterwards. The majority of the world do not use 4 byte chars and its not default in jackson

davsclaus (approved):

davsclaus:
Great work, glad you found out the core dataformat stuff, that is a bit special and also inherited from the early days of Camel.

rnetuka:
@davsclaus Thank you!

davsclaus:
The CSB project may need to rebuild that -starter due to the rename of this option

PR #15961, opened 2024-10-15
Use specific tag of minio container for Integration tests

apupier:
Using latest is making the build potentially break on external changes. In this case, the latest push is missing s390x architecture. See https://github.com/minio/minio/issues/20560

it allows to fix tests on main branch

# Description

# Target

# Tracking


apupier:
for information, s390x is deprecated for minio https://github.com/minio/minio/issues/20560#issuecomment-2414388163

So etiher we start skipping IT tests on s390x or we pin to latest available with s390x for a hile and upgrade to latest tag for other architectures. (can be done in another PR)

PR #15998, opened 2024-10-17
[CAMEL-21362] camel k8s run cannot find pod on openshift

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21362


davsclaus:
So it seems it needs 2 labels now.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/traits/BaseTrait.java
```diff
@@ -20,6 +20,7 @@
 public abstract class BaseTrait implements Trait {
 
     public static final String INTEGRATION_LABEL = "camel.apache.org/integration";
```
squakez, line 22: Probably you need to remove this then.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/traits/ContainerTrait.java
```diff
@@ -82,6 +82,7 @@ public void apply(Traits traitConfig, TraitContext context) {
                 .editOrNewTemplate()
                 .editOrNewMetadata()
                 .addToLabels(INTEGRATION_LABEL, context.getName())
```
squakez, line 84: ditto

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/traits/ContainerTrait.java
```diff
@@ -93,6 +94,7 @@ public void apply(Traits traitConfig, TraitContext context) {
                 .editOrNewJobTemplate()
                 .editOrNewMetadata()
                 .addToLabels(INTEGRATION_LABEL, context.getName())
```
squakez, line 96: ditto

squakez (commented):
I guess you may need to remove the Integration label at all which appears to be a Camel K leftover. Not sure if that would impact any other existing feature though.

davsclaus:
thanks for the review @squakez

tdiesler:
Should this be removed in 4.x as well? It'd be a breaking change.

davsclaus:
> Should this be removed in 4.x as well? It'd be a breaking change.

yeah fine to remove in 4.8.x as its a brand new release and 4.8.0 is a bit buggy here

davsclaus (approved):

gansheer:
LGTM.
A little out of scope but I saw this comment in the code : 
https://github.com/apache/camel/blob/f23f02531518c8a2b34a0b5d582a96e81fcd7c42/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java#L232-L237
Do you know if there is a lot of diff between jkube, quarkus and generated manifest for labels ?

squakez:
> Should this be removed in 4.x as well? It'd be a breaking change.

It is a valid concern. However, as this is a prototyping tool, I think we should not worry that much as we don't expect production workflows leveraging this directly.

christophd:
I think the root cause of all of this is that when using OpenShift as a `cluster-type` we switch from Kubernetes Quarkus extension to the OpenShift Quarkus extension and unfortunately these two seem to behave differently. 

The Quarkus Kubernetes extension properly merges the Kubernetes manifest coming from the Camel JBang plugin (in `src/main/kubernetes`) with its own generated Kubernetes manifest. The Quarkus OpenShift extension presumably is not doing the same merger so things like the INTEGRATION_LABEL may not get properly added to the final Kubernetes manifest.

Switching from INTEGRATION_LABEL to KUBERNETES_NAME_LABEL as in this PR fixes it because KUBERNETES_NAME_LABEL is what the Quarkus extension uses. I am fine with the fix in this PR and I also think that is does not break existing behavior since the users coming from 4.8.0 will not face breaking changes because of this as it is an internal detail in the manifest.

In general we need to analyze and find out why the Kubernetes manifest is not merged properly when using the Quarkus OpenShift extension. In 4.8.0 the focus was to support plain Kubernetes where all of this is working properly.

christophd (approved):

davsclaus:
@christophd can you create a JIRA with your last comment so we wont forget

christophd:
@davsclaus see https://issues.apache.org/jira/browse/CAMEL-21365

davsclaus:
@tdiesler can you update the PR
	modified:   dsl/camel-jbang/camel-jbang-plugin-k/src/test/java/org/apache/camel/dsl/jbang/core/commands/k/IntegrationExportTest.java

PR #16008, opened 2024-10-18
Camel 21302

ffang:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-tracing/src/main/java/org/apache/camel/tracing/ActiveSpanManager.java
+++ b/components/camel-tracing/src/main/java/org/apache/camel/tracing/ActiveSpanManager.java
@@ -56,6 +56,13 @@ public final class ActiveSpanManager {
      * @param span     The span
      */
     public static void activate(Exchange exchange, SpanAdapter span) {
+        if (exchange.getProperty(ExchangePropertyKey.CLOSE_CLIENT_SCOPE, Boolean.FALSE, Boolean.class)) {
+            //Check if we need to close the CLIENT scope created by
+            //DirectProducer in async mode before we create a new INTERNAL scope
+            //for the next DirectConsumer
+            endScope(exchange);
+            exchange.setProperty(ExchangePropertyKey.CLOSE_CLIENT_SCOPE, Boolean.FALSE);
+        }
         exchange.setProperty(ExchangePropertyKey.ACTIVE_SPAN,
                 new Holder(exchange.getProperty(ExchangePropertyKey.ACTIVE_SPAN, Holder.class), span));
         if (Boolean.TRUE.equals(exchange.getContext().isUseMDCLogging())) {
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
@@ -42,7 +42,6 @@ import jakarta.ws.rs.core.Response;
 import org.apache.camel.AsyncCallback;
 import org.apache.camel.CamelExchangeException;
 import org.apache.camel.Exchange;
-import org.apache.camel.ExchangePropertyKey;
 import org.apache.camel.Message;
 import org.apache.camel.component.cxf.common.CxfOperationException;
 import org.apache.camel.component.cxf.common.message.CxfConstants;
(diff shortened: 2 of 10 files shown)


ffang:
This PR includes @johnpoth 's  AsyncCxfTest, a proposal fix(and testcase to expose this issue without cxf endpoint involved) for CAMEL-21302, also revert commits introduced by CAMEL-21309, as if I read the camel-tracing code correctly, the problem is from camel-direct component but not camel-cxf.

Thanks for review!

Freeman

johnpoth (approved):

johnpoth:
Great job @ffang ! Tested locally and everything seems to be working fine!

davsclaus (approved):

components/camel-tracing/src/main/java/org/apache/camel/tracing/ActiveSpanManager.java
```diff
@@ -56,6 +56,13 @@ public static SpanAdapter getSpan(Exchange exchange) {
      * @param span     The span
      */
     public static void activate(Exchange exchange, SpanAdapter span) {
+        if (exchange.getProperty(ExchangePropertyKey.CLOSE_CLIENT_SCOPE, Boolean.FALSE, Boolean.class)) {
+            //Check if we need to close the CLIENT scope created by
+            //DirectProducer in async mode before we create a new INTERNAL scope
+            //for the next DirectConsumer
+            endScope(exchange);
+            exchange.setProperty(ExchangePropertyKey.CLOSE_CLIENT_SCOPE, Boolean.FALSE);
```
davsclaus, line 64: Its likely not needed to set this as the default value above is already false

components/camel-tracing/src/main/java/org/apache/camel/tracing/ActiveSpanManager.java
```diff
@@ -56,6 +56,13 @@ public static SpanAdapter getSpan(Exchange exchange) {
      * @param span     The span
      */
     public static void activate(Exchange exchange, SpanAdapter span) {
+        if (exchange.getProperty(ExchangePropertyKey.CLOSE_CLIENT_SCOPE, Boolean.FALSE, Boolean.class)) {
```
davsclaus, line 59: We can use removeProperty to not keep it around anymore as we have ended the scope

davsclaus:
the otel/tracing code is complex and even more those otel frameworks.
its a bit pita that this fix needs to have code scattered in camel-direct / camel-tracing / camel-core that is complex to remember how is working together.

Hope that these otel frameworks becomes more friendly and dont have this hard thread local issue

components/camel-tracing/src/main/java/org/apache/camel/tracing/ActiveSpanManager.java
ffang: Hi @davsclaus ,

We run into this piece of code only when property ExchangePropertyKey.CLOSE_CLIENT_SCOPE is true, and we need to switch it as false to avoid doing endScope(exchange) again.

Thanks for review!
Freemna

ffang:
> Great job @ffang ! Tested locally and everything seems to be working fine! I guess we might want to remove the `AsyncCxfTest` changes at some point since it turned out to be unrelated ?

Thanks @johnpoth  for the feedback!

I'd keep your AsyncCxfTest here if it's OK to have camel-cxf as test scope dependency in camel-opentelemetry. 

Freeman

ffang:
> the otel/tracing code is complex and even more those otel frameworks. its a bit pita that this fix needs to have code scattered in camel-direct / camel-tracing / camel-core that is complex to remember how is working together.
> 
> Hope that these otel frameworks becomes more friendly and dont have this hard thread local issue

Totally agree! Especially when async gets involved!

johnpoth:
I did a small PoC to rewrite [1] camel-opentelemetry by removing the thread local stuff and I also removed the dependency on camel-tracing. The problem is that a lot of other libraries rely on the thread local context [2] to be set. To mitigate that, we could set and unset the threadlocal in a InterceptStrategy so that libraries would be able to retrieve it. I do think we can leave it as it is now and maybe come back to this idea if things get too complicated.

[1] https://github.com/johnpoth/camel/tree/opentelemetry-rewrite/components/camel-opentelemetry
[2] https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/bootstrap/Java8BytecodeBridge.java#L22
[2] https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation

PR #16041, opened 2024-10-22
chore: stabilize FtpRecursiveDepth2IT

apupier:
this test failed several times in a row on Jenkins Ci with slightly different errors.
```
mock://result Message with body a2 was expected but not found in [b2,
b3, b2, a3]
```
```
mock://result Message with body a3 was expected but not found in [b3,
b2]
```

When playing locally with @RepeatedTest(100), I do not reproduce the error but noticed that 4 or 5 tests took more than 5 seconds although others took 1 second. When I tried to set a low timeout locally, it is failing the test with similar kind of errors.
So trying with an higher timeout, even though it seems strange that there can be that much variation in route execution.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpRecursiveDepth2IT.java
+++ b/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpRecursiveDepth2IT.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.file.remote.integration;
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus:
the mock endpoint should have 10 sec default timeout

apupier:
> the mock endpoint should have 10 sec default timeout

upgraded to 20 seconds then

davsclaus (approved):

PR #16048, opened 2024-10-22
CAMEL-20367: Adds Jolokia and Hawtio tests

jvrubel:
Adds tests for [Jolokia and Hawtio](https://camel.apache.org/manual/camel-jbang.html#_using_jolokia_and_hawtio).

The `hawtio` command does not support a `--background` option and capturing the output of the command is not really possible until it terminates, so in the hawtio test the assertion is made on the expected output of `curl` and Jolokia being attached.


oscerd (approved):

davsclaus (approved):

jvrubel:
I just realised that the output of `curl http://localhost:8888/hawtio/` also contains `content="Hawtio Management Console"`, so for the sake of readability it will be better to assert on this instead of `You need to enable JavaScript`. I'll quickly change that.

PR #16053, opened 2024-10-22
[CAMEL-21361] fabric8 kubernetes-client fails to connect to openshift

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21361

The root cause was a convoluted TCCL issue. In general, when you work with TCCL, you SHOULD reset it to its original value, like so ...

```
        var tccLoader = Thread.currentThread().getContextClassLoader();
        try {
            ClassLoader cl = createClassLoader();
            Thread.currentThread().setContextClassLoader(cl);

            // do stuff

        } finally {
            Thread.currentThread().setContextClassLoader(tccLoader);
        }
```

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/pom.xml
@@ -80,6 +80,12 @@
             <version>${kubernetes-client-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>openshift-client</artifactId>
+            <version>${kubernetes-client-version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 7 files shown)


dsl/camel-jbang/camel-jbang-plugin-kubernetes/pom.xml
```diff
@@ -55,6 +55,11 @@
         </dependency>
 
         <!-- OpenShift -->
+        <dependency>
```
davsclaus, line 58: Is this only used for testing then set test scope

tdiesler: It's use run `run` as well

davsclaus: where is it needed - it was not there before

tdiesler: I looked again ... and actually no, the run code uses the k8s API. Let me see if it actually works with scope test ...

tdiesler: done

PR #16084, opened 2024-10-24
[CAMEL-21381] camel k8s cannot run spring-boot on openshift

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21381


oscerd (approved):

davsclaus (approved):

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/test/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/PodLogsTest.java
```diff
@@ -33,7 +33,8 @@ public void shouldHandlePodNotFound() throws Exception {
         PodLogs command = createCommand();
         command.name = "mickey-mouse";
         command.maxWaitAttempts = 2; // total timeout of 4 seconds
-        command.doCall();
+        int exit = command.doCall();
+        Assertions.assertEquals(1, exit);
```
davsclaus, line 37: Just wonder why exit code 1 is expected, is the command supposed to fail somehow ? If so maybe add a code comment to tell this reason so we wont get confused later looking back at this code

tdiesler: It is debatable, wether the `jbang camel kubernetes logs ...` command should "fail" when the Pod is not there. User feedback (in print the output) is clear. We might as well succeed the command and reserve failure to unexpected server failures. Would that be better?

davsclaus: Yes I think this is better

tdiesler: Done

tdiesler:
Thanks for merging. Could you perhaps leave the cherry-pick to 4.8.x to me - I'd like to run some (manual) tests before we commit to that otherwise stable branch?

PR #16092, opened 2024-10-25
CAMEL-21355: refine no tools called scenario

orpiske:
- stop throwing an exception
- set a header if no tool was called


orpiske:
This should provide a middle ground between throwing an exception (as it is now) and failing silently (as it was before in `-chat` component).

PR #16095, opened 2024-10-25
(follow-up) remove dependency management from individual poms

gnodet:


davsclaus:
With and without this PR do you see some changes in the `mvn dependency:tree` or something ?

PR #16106, opened 2024-10-28
CAMEL-21395: camel-debezium - Use dedicated package for each db type

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-21395

## Motivation

The project camel-debezium-common shares the same package with the other camel-debezium sub-projects, which causes issues on environments like OSGI.

## Modifications:

* Update the camel debezium maven plugin to be able to provide the package where the class will be generated
* Use a dedicated package for each db type
* Log the errors that could occur while starting the Debezium engine.


davsclaus (approved):

essobedo:
@davsclaus Hi, thanks for the review. Would it be acceptable to backport it into branch 4.8.x?

davsclaus:
Yes if you add a note to the 4.8 upgrade guide (4.8.1 -> 4.8.2) about that these class names has been changed. However as a camel end user you would use these components in the standard camel way and very likely not be affected of this change at all.

davsclaus:
also mind that camel spring boot may need to be regenerated for its -starter - so can you fix/look at that first

essobedo:
@davsclaus good catch, here is the PR for SB https://github.com/apache/camel-spring-boot/pull/1270

PR #16107, opened 2024-10-28
CAMEL-21397 - camel-file - fix autoCreateStepwise for Windows environments

s-clauw:
# Description

When autoCreateStepwise = true and no permissions are set it still tries to set the permissions, which fail on Windows environments.
This is due to the fact that the 'permissions' object is never null, it's initialised as EnumSet.noneOf(PosixFilePermission.class);

We would need this fix in 4.8.2 if possible (blocking our prod deployment) :)

# Target

# Tracking


davsclaus (approved):

s-clauw:
Actually another fix is needed as we tested it last night. The path split does not work on Windows for fileshares ('\\<servername>\...' as it doesn't take into account the server name for SMB URI's. It creates the file on the local machine instead of on a fileshare.
We will propose a fix in this PR.

davsclaus:
@s-clauw can you create JIRA tickets about these findings as we need this for 4.8.x changelogs and release notes.

s-clauw:
> @s-clauw can you create JIRA tickets about these findings as we need this for 4.8.x changelogs and release notes.

I created it as Critical as it will fail on Windows environments when autoCreateStepwise is enabled. Feel free to adjust it: [CAMEL-21397](https://issues.apache.org/jira/browse/CAMEL-21397)

components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
```diff
@@ -121,7 +127,14 @@ protected boolean buildDirectory(File dir, Set<PosixFilePermission> permissions,
             File base;
             // reusing absolute flag to handle relative and absolute paths
             if (absolute) {
-                base = new File("");
+                if (isWindows() && dir.getPath().startsWith(File.separator + File.separator)) {
+                    List<String> uncParts = Arrays.stream(parts).filter(s -> !s.isEmpty()).collect(Collectors.toList());
```
davsclaus, line 131: this code is a bit hacky. 
Also can you show an UNC path example that currently fails.

Also mind there is camel-smb for SMB file systems.

s-clauw: commit is reversed

PR #16114, opened 2024-10-29
CAMEL-21395: camel-debezium - Use dedicated package for each db type - 4.8

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-21395 (v 4.8)

## Motivation

The project camel-debezium-common shares the same package with the other camel-debezium sub-projects, which causes issues on environments like OSGI.

## Modifications:

* Update the camel debezium maven plugin to be able to provide the package where the class will be generated
* Use a dedicated package for each db type
* Log the errors that could occur while starting the Debezium engine.


essobedo:
The failure is not related to this fix https://github.com/apache/camel/pull/16122

PR #16127, opened 2024-10-30
CAMEL-21402 Added means to verify expressions in MockEndpoint

dk2k:
Yes, I know there is a way to verify expressions like:
mock.message(0).header("foo").matches().constant(123);

But still I can't verify a regex this way
There are more possible usages. Say, XpathExpression against a header. I don't think it's possible with the current API

### DIFF
--- a/components/camel-mock/src/main/java/org/apache/camel/component/mock/MockEndpoint.java
+++ b/components/camel-mock/src/main/java/org/apache/camel/component/mock/MockEndpoint.java
@@ -118,6 +118,8 @@ public class MockEndpoint extends DefaultEndpoint implements BrowsableEndpoint, 
     private volatile Map<String, Object> expectedPropertyValues;
     private volatile Map<String, Object> expectedVariableValues;
 
+    private volatile List<ExpressionWithDescription> expectedExpressions;
+
     private final AtomicInteger counter = new AtomicInteger();
 
     @UriPath(description = "Name of mock endpoint")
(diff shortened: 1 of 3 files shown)


davsclaus:
Here is how you can already do this using the existing API
https://github.com/apache/camel/commit/98ab77779a3f24642ddf412ce75e08357080cd6b

dk2k:
Hi Claus @davsclaus ,
your PR gives a good hint on the capabilities, which I missed. I like this functional Stream-like API, but in some cases you need something less rigid.
I don't insist, that my proposed changes in API cover something previously completely uncovered.

My point is:
1. For each type of processing like regex("value[2,3]") you need to add a method into MockValueBuilder. What happens if I want to use an expression with xpath()? I need to add a method. Another type of processing? Another method. Sounds like poor extendability/maintainability.
 If you ask me how somebody may want to process a header with xpath, I have this link https://stackoverflow.com/questions/49381455/extract-xml-using-xpath-in-camel
So, people send XMLs in HTTP headers :)
I check options for xpath and I can see that currently it can be applied only to request's body, not to headers.
 
2. MockEndpoint already provides means to test a header value against a constant:
resultEndpoint.expectedHeaderReceived("header", "value");
I believe the same expectation can be checked with:
mock.message(4).header("header").regex("value"); 
or
mock.message(4).header("header").asExpression("value"); 
(More cumbersome, but still it's a duplication of capabilities of method expectedHeaderReceived() - I can understand that it may be preserved for the sake of backward compatibility)

That's why I decided to extend the API. New method expectedExpressionMatches() is designed very similar to expectedHeaderReceived()
expectedExpressionMatches() provides extendability. My plan was to create XpathExpression(Expression expr, String xpath) - maybe in a separate PR.

The more testing capabilities, the better. Like ThereIsMoreThanOneWayToDoIt principle in Perl

davsclaus:
You can already use any of the languages, see this xpath test
https://github.com/apache/camel/commit/5308773e385397159c19175cfe591da56d3a434c

We could improve its API a little bit to allow spefiying the xpath directly without having to use .expression(...) to set its literal value.

dk2k:
I can see that this chain calls probably cover all the testing needs. Still: what if I want to test a custom Expression, created on the fly? It's about extensibility without needing to alter existing classes in the core.
I'll take a look at built-in languages

davsclaus:
Yeah a custom `Expression` method on the fly, where you may want to have some lambda style, ala

```
message(0).header("cheese").expression(e ->  // code goes here).isEqualTo("gauda")
```

You are welcome to take a fresh look and see if something non intrusive can be improved without too much complexity or yet another way of doing something. We need to be more consistent and lean on the built-in functionality today.

Closing this PR

PR #16160, opened 2024-11-05
fix(components/plc4x): fix NPE when Camel can not establish connection

jmeierhofer:
# Description
When Camel can not establish the connection to the S7 PLC, the component is throwing follow-up exceptions which are causing the application start up to fail and which hide the real connection issue.
- Added isConnected() to endpoint which checks if connection is null. This method is public to allow external logic to check if endpoint is really connected.
- Moved canWrite() call to be checked only when connected.
- Added additional canWrite() check after reconnectIfNeeded().

Steps to reproduce:
- Add Camel route with S7 driver
- PLC is not available
- Connection failure
- Additional NullPointerExceptions are thrown

# Target

# Tracking


oscerd (approved):

jmeierhofer:
@oscerd I think the maven download failed during the JDK17 build. Can you please trigger the 2 builds again?

PR #16182, opened 2024-11-06
CAMEL-21443: Camel JBang shell command

gnodet:
A simple `shell` command which will enter a REPL loop with nice completion and help.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-core/pom.xml
@@ -88,6 +88,11 @@
             <version>${picocli-version}</version>
         </dependency>
         <dependency>
+            <groupId>info.picocli</groupId>
+            <artifactId>picocli-shell-jline3</artifactId>
+            <version>${picocli-version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.github.freva</groupId>
             <artifactId>ascii-table</artifactId>
             <version>${ascii-table-version}</version>
--- a/dsl/camel-jbang/camel-jbang-core/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-core/pom.xml
@@ -141,6 +141,11 @@
             <version>${jansi-version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-xml</artifactId>
+        </dependency>
+
         <!-- test dependencies -->
         <dependency>
             <groupId>org.apache.camel</groupId>
--- a/dsl/camel-jbang/camel-jbang-core/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-core/pom.xml
@@ -93,6 +93,11 @@
             <version>${picocli-version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


davsclaus:
How much other dependencies does jline bring in?

I wonder as camel run should ideally be without too much clutter. And maybe we need to find a way for jbang to be able to dynamic select DEPS based on the command. The DEPS is a bit too static for us.

davsclaus:
Also can you update the jbang docs, as this seems like a great feature for users to learn and play with camel-jbang and its worth to have in the docs. See under docs root folder there is a camel-jbang.adoc file.

gnodet:
> How much other dependencies does jline bring in?

There's no required dependency at all in JLine.  They are all either optional or test deps:
https://repo1.maven.org/maven2/org/jline/jline/3.27.1/jline-3.27.1.pom

> I wonder as camel run should ideally be without too much clutter. And maybe we need to find a way for jbang to be able to dynamic select DEPS based on the command. The DEPS is a bit too static for us.

dsl/camel-jbang/camel-jbang-core/pom.xml
```diff
@@ -136,6 +141,11 @@
             <version>${jansi-version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-xml</artifactId>
+        </dependency>
```
davsclaus, line 147: What is this used for ?

davsclaus (approved):

dsl/camel-jbang/camel-jbang-core/pom.xml
gnodet: This is slightly unrelated. The latest maven 4 builds include a fix in the way transitive dependency management is computed. This dependency is actually included in the dep tree with maven 3, but not with Maven 4, so I just included it at the same time. But this could be splitter into a different PR if needed.

dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java
```diff
@@ -22,6 +22,7 @@
 //DEPS org.apache.camel:camel-bom:${camel.jbang.version:4.8.1}@pom
 //DEPS org.apache.camel:camel-jbang-core:${camel.jbang.version:4.8.1}
 //DEPS org.apache.camel.kamelets:camel-kamelets:${camel-kamelets.version:4.8.0}
+//DEPS org.jline:jline:${camel-jline.version:3.27.1}
```
davsclaus, line 25: Can this be in the pom.xml instead so we keep this file as before

davsclaus (changes requested):

davsclaus:
@gnodet did you get time to look at this - can we avoid DEPS to keep the file as before - there are also some camel jbang code that runs a custom camel version that creates a .java source that would break if this file now have extra deps instead of just in the pom.xml

gnodet:
> @gnodet did you get time to look at this - can we avoid DEPS to keep the file as before - there are also some camel jbang code that runs a custom camel version that creates a .java source that would break if this file now have extra deps instead of just in the pom.xml

I'll have a look at it today.

PR #16214, opened 2024-11-11
camel-fury: Improve to allow auto discovery Fury from the registry

zhfeng:
* replace to use BaseFury
* add an option with allowAutoWiredFury
* add FuryDataFormatReifier

# Description

# Target

# Tracking


core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/FuryDataFormat.java
```diff
@@ -40,6 +40,19 @@ public class FuryDataFormat extends DataFormatDefinition {
     @Metadata(description = "Class of the java type to use when unmarshalling")
     private String unmarshalTypeName;
 
+    @XmlAttribute
+    @Metadata(description = "Whether to require register classes", defaultValue = "True", javaType = "java.lang.Boolean")
```
davsclaus, line 44: true should be lower case

components/camel-fury/src/main/java/org/apache/camel/component/fury/FuryDataFormat.java
```diff
@@ -85,7 +91,21 @@ protected void doInit() throws Exception {
             unmarshalType = camelContext.getClassResolver().resolveClass(unmarshalTypeName);
         }
 
-        fury = Fury.builder().withLanguage(Language.JAVA).requireClassRegistration(true).build();
+        if (fury == null && isAllowAutoWiredFury()) {
+            Set<BaseFury> beans = getCamelContext().getRegistry().findByType(BaseFury.class);
```
davsclaus, line 95: Use `findSingleByType` instead

core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/FuryDataFormat.java
```diff
@@ -40,6 +40,19 @@ public class FuryDataFormat extends DataFormatDefinition {
     @Metadata(description = "Class of the java type to use when unmarshalling")
     private String unmarshalTypeName;
 
+    @XmlAttribute
+    @Metadata(description = "Whether to require register classes", defaultValue = "true", javaType = "java.lang.Boolean")
+    private String requireClassRegistration;
+
+    @XmlAttribute
```
davsclaus, line 47: this thread safe and autowired options could be labelled as advanced

core/camel-core-model/src/main/java/org/apache/camel/model/MarshalDefinition.java
```diff
@@ -88,6 +89,7 @@ public class MarshalDefinition extends NoOutputDefinition<MarshalDefinition> imp
             @XmlElement(name = "custom", type = CustomDataFormat.class),
             @XmlElement(name = "fhirJson", type = FhirJsonDataFormat.class),
             @XmlElement(name = "fhirXml", type = FhirXmlDataFormat.class),
+            @XmlElement(name = "fury", type = FuryDataFormat.class),
```
davsclaus, line 92: You should add this in other places, just grep for one of the existing ie in Unmarshal and DataFormats definition classes

davsclaus (changes requested):

core/camel-core-model/src/main/java/org/apache/camel/model/MarshalDefinition.java
zhfeng: Thanks @davsclaus for pointing out. Just wonder if there is any docs for develop a new `DataFormat`?

davsclaus: No specific docs - the best is looking at any of the others and see how they are done.

davsclaus (approved):

PR #16220, opened 2024-11-11
Fix for RestOpenApiProcessor for parsing path variables from the http request uri succesfully. 

egekaraosmanoglu:
Fix/Update org.apache.camel.component.rest.openapi.RestOpenApiProcessor.java

When RestOpenApiProcessor tries to process the path variables in the request,  HttpHelper.evalPlaceholders  throws an ArrayIndexOutOfBoundsException because consumerPath  from the openApi spec starts with a "/" but the uri of the request does not start with a "/".

# Target

# Tracking

### DIFF
--- a/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiProcessor.java
+++ b/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiProcessor.java
@@ -112,13 +112,12 @@ public class RestOpenApiProcessor extends DelegateAsyncProcessor implements Came
             String consumerPath = rcp.getConsumerPath();
 
             //if uri is not starting with slash then remove the slash in the consumerPath from the openApi spec
-            if(consumerPath.startsWith("/") && !uri.startsWith("/"))
-            {
+            if (consumerPath.startsWith("/") && uri != null && !uri.startsWith("/")) {
                 consumerPath = consumerPath.substring(1);
             }
-            
+
             // map path-parameters from operation to camel headers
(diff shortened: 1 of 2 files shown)


components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiProcessor.java
```diff
@@ -109,6 +109,14 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
         if (m instanceof RestOpenApiConsumerPath rcp) {
             Operation o = rcp.getConsumer();
 
+            String consumerPath = rcp.getConsumerPath();
+
+            //if uri is not starting with slash then remove the slash in the consumerPath from the openApi spec
+            if(consumerPath.startsWith("/") && !uri.startsWith("/"))
```
davsclaus, line 115: Add space after if

davsclaus: And use same code style as elsewhere, ie curly bracket on same line etc

davsclaus (changes requested):
code style changes needed for PR to build

components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiProcessor.java
egekaraosmanoglu: sure thanks, also I see that I forgot to use the variable  consumerPath while calling HttpHelper.evalPlaceholders.

egekaraosmanoglu:
Hi, applied the code style fixes and added a null check. I hope it is fine now! Thanks.

davsclaus (approved):

PR #16227, opened 2024-11-11
CAMEL-20503: camel-http OAuth2 support for caching / refreshing tokens

kulagaIA:
# Description

- added caching oauth2 tokens for http component, and configuration parameters for caching
- token is cached for combination of credentials and host: if username/password or authentication host changes, token will be queried again.

# Tracking
https://issues.apache.org/jira/browse/CAMEL-20503


davsclaus (approved):

oscerd (approved):

davsclaus:
Oh @kulagaIA you need to add the new option to the exclude list in `WebsocketEndpoint` and rebuild the source code, as this should only be for camel-http only.

kulagaIA:
> Oh @kulagaIA you need to add the new option to the exclude list in `WebsocketEndpoint` and rebuild the source code, as this should only be for camel-http only.

done

davsclaus (approved):

components/camel-http/src/main/java/org/apache/camel/component/http/OAuth2ClientConfigurer.java
```diff
@@ -37,55 +45,143 @@ public class OAuth2ClientConfigurer implements HttpClientConfigurer {
     private final String clientSecret;
     private final String tokenEndpoint;
     private final String scope;
+    private final boolean cacheTokens;
+    private final Long cachedTokensDefaultExpirySeconds;
+    private final Long cachedTokensExpirationMarginSeconds;
+    private final static Map<OAuth2URIAndCredentials, TokenCache> tokenCache = new HashMap<>();
```
PascalSchumacher, line 51: It is great to see this feature added, but this should be made thread safe (e.g. by using `ConcurrentHashMap`) otherwise different threads may not see each others updates and in the worst case there can be an infinite loop during concurrent modification.

davsclaus: yeah thanks

PR #16319, opened 2024-11-19
Fix since attribute for Flowable Component

roumn:
# Description

Fix a typo in the documentation for flowable component

# Target


oscerd:
Thanks, but the fix needs to be done at endpoint level: https://github.com/apache/camel/commit/a0392951c17bb5cb997a1123ee8fa2c5efdbe08e

PR #16346, opened 2024-11-21
[Backport 4.8.x] Fix for RestOpenApiProcessor for parsing path variables from the http…

squakez:
… request uri succesfully.  (#16220)

* Update RestOpenApiProcessor.java

if uri is not starting with slash then remove the slash in the consumerPath from the openApi spec, because if there is any path variables (/../{id}) in the consumerPath and the uri does not start with "/" then  HttpHelper.evalPlaceholders method throws an ArrayIndexOutOfBoundsException.

* code style fixes and added null check the uri.

# Description

# Target

# Tracking


oscerd (approved):

squakez:
There are tests failure in Camel JBang plugins which are not related to this PR. @oscerd is it fine to merge anyway given the problems are not within this PR? not sure the policy we're adapting here in the core. Thanks.

PR #16422, opened 2024-12-03
CAMEL-21019: Create a Camel TensorFlow Serving component

tadayosi:
# Description

https://issues.apache.org/jira/browse/CAMEL-21019

# Target

# Tracking


tadayosi:
@davsclaus @oscerd Is it still possible to make it in the 4.9.0 release?  It'd be great if we can have it in the 4.9.0 train. Otherwise, I'm fine with holding it for 4.10.0.

davsclaus:
No its too late for 4.9, so it will be in 4.10

tadayosi:
OK, then I'll keep it until 4.10 development starts.

davsclaus (approved):

gansheer:
@tadayosi @davsclaus  How does declaring Apache and Google license works in practice for the filescomponents/camel-ai/camel-tensorflow-serving/src/main/proto/google/protobuf/any.proto and components/camel-ai/camel-tensorflow-serving/src/main/proto/google/protobuf/wrappers.proto ?

tadayosi:
@gansheer Camel's build tool requires and automatically add the license header to every text file in the project. But the actual file `any.proto` and the other proto files are originally copied from the TensorFlow repository as-is, that's why it has double license headers. My understanding is however that APL-2.0 is so permissive that there's no problem with doing so and there should be no practical implications with it.

gansheer (approved):

tadayosi:
Rebased to 4.10.0-SNAPSHOT. I'll merge it once CI passes.

PR #16497, opened 2024-12-09
Fixed NPE when no response body in DefaultRawClient

baconberry:
# Description

When performing raw requests using the `raw` endpoint in the salesforce component, if there's an error without body the `DefaultRawClient` is unable to create a `SalesforceException` because it throws a `NullPointerException`, I see other internal clients implementations, and there the null response body is handled correctly.

[This following branch can be used to see the current wrong behavior.](https://github.com/baconberry/camel/blob/reproduce-salesforce-raw-client-npe/components/camel-salesforce/camel-salesforce-component/src/test/java/org/apache/camel/component/salesforce/RawPayloadTest.java)

# Target

# Tracking

Is this trivial enough?


oscerd (approved):

davsclaus (approved):

oscerd:
This needs to be backported to 4.8.x too

PR #16610, opened 2024-12-19
Allow propagating -version properties from the top pom to camel-dependencies

ppalaga:
# Description

sync-properties-maven-plugin currently considers only version properties defined in camel-parent with the sole exception of `license-maven-plugin-version` that comes from `org.apache.camel:camel`. Because `org.apache.camel:camel` is the parent of camel-parent, this change enables the propagation of non-plugin version properties from the there too.

# Target

# Tracking
- [x] This is a small build time detail that does not deserve a Jira issue.


oscerd (approved):

ppalaga:
abeabec: fixed the formatting

gansheer (approved):

PR #16625, opened 2024-12-20
CAMEL-21562. It's not a problem if the client is not allowed to do HeadBucket

mihxil:
# Description

 The AWS2S3Endpoint tries to do a HeadBucket request, which may not be permitted. This is only a problem if the endpoint is also configured to create the bucket if it does not exist.

Otherwise it may simply be that other actions on the bucket will succeed, and there is not reason to fail if just the HeadBucket request is not permitted.

I though this would be CAMEL-18022, but  I think it is different. 
Now:
https://issues.apache.org/jira/browse/CAMEL-21562


oscerd (approved):

oscerd:
The build error is not related to this PR.

PR #16661, opened 2024-12-27
CAMEL-21484: mp-config should not be loadable as it has overhead at r…

davsclaus:
…untime

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-microprofile/camel-microprofile-config/src/main/java/org/apache/camel/component/microprofile/config/CamelMicroProfilePropertiesSource.java
+++ b/components/camel-microprofile/camel-microprofile-config/src/main/java/org/apache/camel/component/microprofile/config/CamelMicroProfilePropertiesSource.java
@@ -16,42 +16,16 @@
  */
 package org.apache.camel.component.microprofile.config;
 
-import java.util.Collections;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


gansheer (approved):

Paul-Blanchaert:
@davsclaus 
I believe we still require the method isValidForActiveProfiles(String name). This filters the "active" properties for the selected runtime profile(s). That might be the reason why the "loadable" was chosen in the first place.
[CAMEL-21480](https://issues.apache.org/jira/browse/CAMEL-21480) is expected to enhance this isValidForActiveProfiles(String name) method to support for multiple profiles.

Paul-Blanchaert:
On 2nd thought: probably that CAMEL-21480 is probably also sorted by using the service as it delegates that function to the service...

oscerd (approved):

oscerd (approved):

jamesnetherton:
> I believe we still require the method isValidForActiveProfiles(String name). This filters the "active" properties for the selected runtime profile(s

I don't think that's needed with this change. If we're just resolving properties at runtime via `ConfigProvider`, then it'll factor in whatever active profiles are enabled.

jamesnetherton (approved):

PR #16686, opened 2025-01-02
CAMEL-21352: extend camel smb

jonomorris:
# Description

Expand the options available to the SMB consumer and producer by extending `GenericFile` classes. 

A number of additional consumer and producer options are now supported including  'move', 'preMove', 'moveFailed', 'readLock', 'readLockTimeout', 'delete', 'include', 'exclude', 'includeExt', 'excludeExt', 'filter', 'localWorkDirectory', 'sortBy', 'maxMessagesPerPoll', 'maxDepth', 'minDepth', 'sorter', 'maxMessagesPerPoll', 'eagerMaxMessagesPerPoll', 'idempotentKey',  and more.

# Target

# Tracking


components/camel-smb/src/main/java/org/apache/camel/component/smb2/Smb2Configuration.java
```diff
@@ -44,18 +51,26 @@ public class Smb2Configuration extends GenericFileConfiguration {
     private String password;
     @UriParam(label = "security", description = "The user domain")
     private String domain;
+    @UriParam(defaultValue = "*.txt", description = "The search pattern used to list the files")
```
davsclaus, line 54: I dont think we should only include txt files, it should be all files.

jonomorris: Makes sense, now includes all files.

davsclaus (approved):

gansheer (approved):

orpiske:
I am curious about why adding an entirely new component instead of refactoring the existing one.

orpiske (changes requested):
Overall, LGTM, but I am curious about the point I added above (i.e: adding a new component). 

IMHO, Camel is already far too bloated and, if we could adjust what we have, that would be better.

davsclaus:
This is not the final work. We will delete the old and rename to smb when its ready.
If this is already the case then @jonomorris you are welcome to do that.

orpiske:
> This is not the final work. We will delete the old and rename to smb when its ready. If this is already the case then @jonomorris you are welcome to do that.

Ah, got it, thanks for the explanation!

That sounds OK to me, but let's make sure we do it in one step so not to mess the git history (i.e.: so we can easily trace back to its origins, changes, and bisect things).

orpiske:
> > This is not the final work. We will delete the old and rename to smb when its ready. If this is already the case then @jonomorris you are welcome to do that.
> 
> Ah, got it, thanks for the explanation!
> 
> That sounds OK to me, but let's make sure we do it in one step so not to mess the git history (i.e.: so we can easily trace back to its origins, changes, and bisect things).

Actually, forget that. It's not that old and so not much to lose. Feel free to do it as you seem fit.

orpiske (approved):

jonomorris:
@davsclaus I'd be happy to delete the old and rename to smb.  Would you like that done in this PR?

davsclaus:
> @davsclaus I'd be happy to delete the old and rename to smb. Would you like that done in this PR?

Yes that would be good

jonomorris:
@davsclaus I've moved smb2 to smb now.  Thanks.

davsclaus:
okay so its ready to merge

davsclaus:
okay I am going to merge this over to main, and regen CSB

davsclaus:
@jonomorris we need to update the 4.10 upgrade guide about camel-smb is refactored to allow users to be aware.

davsclaus:
@jonomorris its merged to main now, great work, and a lot of tests.

I have these errors locally, not sure if you see that too
```
[ERROR] Failures:
[ERROR]   SmbConsumerPartialReadNoPathIT.testSmbSimpleConsumeAbsolute mock://result Body of message: 0. Expected: <Hello World> but was: <24483
>
[ERROR]   SmbRecursiveMaxDepthIT.testDirectoryTraversalDepth mock://received_send Body of message: 0. Expected: <Goodday> but was: <World>
[ERROR]   SmbRecursiveMinDepthIT.testDirectoryTraversalDepth mock://received_send Body of message: 0. Expected: <Hello> but was: <Goodday>
[INFO]
[ERROR] Tests run: 58, Failures: 3, Errors: 0, Skipped: 0
```

jonomorris:
@davsclaus  that's great, thanks.  I'll review those test cases,  they pass locally.  Will also add some notes to the upgrade guide.

PR #16689, opened 2025-01-02
fixes CAMEL-21582, makes microprofiles properties loadable

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-21582

partially reverts some changes from https://github.com/apache/camel/pull/16661/

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-microprofile/camel-microprofile-config/src/main/java/org/apache/camel/component/microprofile/config/CamelMicroProfilePropertiesSource.java
+++ b/components/camel-microprofile/camel-microprofile-config/src/main/java/org/apache/camel/component/microprofile/config/CamelMicroProfilePropertiesSource.java
@@ -16,8 +16,12 @@
  */
 package org.apache.camel.component.microprofile.config;
 
-import org.apache.camel.spi.PropertiesSource;
+import java.util.Properties;
+import java.util.function.Predicate;
+
+import org.apache.camel.spi.LoadablePropertiesSource;
 import org.apache.camel.spi.annotations.JdkService;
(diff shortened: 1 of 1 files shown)


JiriOndrusek:
@davsclaus I'm not sure whether this fix is the correct one. It is quite simple, no more API interfaces are required and makes sense. On the other hand there might be possible different fixes. WDYT?

jamesnetherton:
I have an alternate working impl to set 'initial properties' on camel main.

I'm not sure if that's best way forwards though as I mentioned on Zulip:

https://camel.zulipchat.com/#narrow/channel/257302-camel-quarkus/topic/CamelMicroProfilePropertiesSource

JiriOndrusek:
Thanks for the information. I used a little bit different approach in this PR, as the properties are grabbed dynamically only when they are required. I agree that a different solution might be better.

davsclaus:
CEQ should support `camel.xxx` configuration in `application.properties` out of the box

So if this test starts failing then we should correct the mp config code.https://github.com/apache/camel-quarkus/blob/camel-main/extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelAutowiredDisabledTest.java

CSB works differently as it has Spring AutoConfiguration classes that configures this "the spring way".

davsclaus (approved):

davsclaus:
@jamesnetherton I think we should merge this so CEQ works again

jamesnetherton (approved):

PR #16699, opened 2025-01-03
CAMEL-21586: Create a OpenAI Embedding Model builder acting as Java B…

zbendhiba:
…ean for a Kamelet

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-kamelet/pom.xml
+++ b/components/camel-kamelet/pom.xml
@@ -44,6 +44,12 @@
             <artifactId>camel-core-model</artifactId>
         </dependency>
         <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-open-ai</artifactId>
+            <version>${langchain4j-version}</version>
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

davsclaus (changes requested):
Oh we cannot have camel-kamelet use 3rd party JARs. This should be in some of the camel-ai JARs instead.

davsclaus:
Oh the dependency can be set to provided like some of the others. So when doing this then its okay.

zbendhiba:
Yes, we got a discussion with @oscerd 
I think we can add it like that for the first round, as some other kamelets also have these kind of dependencies. 
I can open a JIRA ticket to figure out how to better deal with these cases in some future. The JIRA can at least track the need to revise the architecture.

WDYT @oscerd @davsclaus ?

zbendhiba:
The `<scope>provided</scope>` is added to the dependency

davsclaus (approved):

PR #16700, opened 2025-01-03
CAMEL-21560: Create a HugginFace Embedding Model builder acting as Ja…

zbendhiba:
…va Bean for a Kamelet

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-kamelet/pom.xml
+++ b/components/camel-kamelet/pom.xml
@@ -34,7 +34,6 @@
     <description>To call Kamelets</description>
 
     <dependencies>
-
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-core-reifier</artifactId>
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

davsclaus:
Oh the dependency can be set to provided like some of the others. So when doing this then its okay.

components/camel-kamelet/pom.xml
```diff
@@ -34,7 +34,11 @@
     <description>To call Kamelets</description>
 
     <dependencies>
-
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-hugging-face</artifactId>
```
oscerd, line 39: Needs to be provided

oscerd (changes requested):

components/camel-kamelet/pom.xml
zbendhiba: okay

davsclaus (approved):

PR #16718, opened 2025-01-06
CAMEL-21595: avoid an NPE if the model decides not to call any tool

orpiske:


orpiske:
Failures are unrelated.

PR #16732, opened 2025-01-07
CAMEL-21596 : Camel Qdrant:  Add similariry search feature

zbendhiba:
# Description

# Target

# Tracking


oscerd (approved):
Looks good, it would be nice if the filter could be a simple string to be transformed on the POJO while using it, that way it would be easier to use inline in the endpoint.

zbendhiba:
> Looks good, it would be nice if the filter could be a simple string to be transformed on the POJO while using it, that way it would be easier to use inline in the endpoint.

@oscerd  Sounds like a better option. However, 'filter' is many things.

Example of a JSON format filter
```
"filter": {
    "must": [
      {
        "key": "category",
        "match": { "value": "laptop" }
      },
      {
        "key": "price",
        "range": {
          "gt": null,
          "gte": null,
          "lt": null,
          "lte": 1000
        }
      }
    ]
  },
```

This is what I can think of:
- Add a simple key/value strings in case we want to use only a single match.
- Add two headers, to deal with multiple matches and numeric values: one expecting a map for the keys/values, and one expecting a range for the values that are numeric.

WDYT ?

oscerd:
> > Looks good, it would be nice if the filter could be a simple string to be transformed on the POJO while using it, that way it would be easier to use inline in the endpoint.
> 
> @oscerd Sounds like a better option. However, 'filter' is many things.
> 
> Example of a JSON format filter
> 
> ```
> "filter": {
>     "must": [
>       {
>         "key": "category",
>         "match": { "value": "laptop" }
>       },
>       {
>         "key": "price",
>         "range": {
>           "gt": null,
>           "gte": null,
>           "lt": null,
>           "lte": 1000
>         }
>       }
>     ]
>   },
> ```
> 
> This is what I can think of:
> 
> * Add a simple key/value strings in case we want to use only a single match.
> * Add two headers, to deal with multiple matches and numeric values: one expecting a map for the keys/values, and one expecting a range for the values that are numeric.
> 
> WDYT ?

Probably it makes less sense, it's too complex to inline in the endpoint. Better to stick with the POJO. Thanks.

orpiske (approved):

davsclaus (approved):

PR #16820, opened 2025-01-15
CAMEL-21614: camel-core - Prevent cache change miss on queue swap - 4.8

essobedo:
Fix for https://issues.apache.org/jira/browse/CAMEL-21614 (4.8)

## Motivation

In some specific use cases, the eviction of the entries never ends

## Modifications:

* Improve the way to swap the queue of change to avoid the race condition that could cause cache changes miss by leveraging a `ReadWriteLock` to only prevent change additions during the swap


oscerd (approved):

oscerd:
Don't worry about the failure. It's because it must be rebuilt with the new SNAPSHOT version.

essobedo:
I know, I'm currently doing the Regen on the branch, thx anyway for the info

essobedo:
Regen done, it should be good now

PR #16826, opened 2025-01-15
[CAMEL-18276] : azure-service-bus component support for session enabled entities

Pulin412:
# Description

Session support for azure-service-bus component. Existing component is updated to handle both non-session (existing functionality) and session based entities.

New properties are added - 

- `sessionId` - For `producer`, sessionId is required to send with message while publishing message for session based entities.
- `sessionEnabled` - For `consumer`, boolean flag to let the consumer create Session aware clients to consume messages from session aware entities.

JIRA Issue - https://issues.apache.org/jira/browse/CAMEL-18276

# Target

# Tracking


davsclaus:
I wonder if `sessionEnabled` should be an option on component level, so you can configure this once if you always want to use this. And if so you are welcome to work on that for another PR

dylanpiergies:
Seems OK to me. I've double checked the integration tests are passing, since they don't run during builds. All good. Aside from that, I haven't really delved into the nuts and bolts of the session functionality with Service Bus. I'd have to spend some time looking through the Azure documentation to see if there are any additions that need to be made for full-featured session support.

PR #16848, opened 2025-01-17
CAMEL-20199: Remove synchronized blocks from components T to Z

essobedo:

## Motivation

## Modifications:

* Replace mutex with locks
* Use locks instead of synchronized blocks
* Leverage ConcurrentMap methods to get rid of synchronized blocks


oscerd (approved):

davsclaus (approved):

davsclaus:
The build error is a know flake on java 21

org.apache.camel.component.mllp.MllpTcpServerConsumerOptionalEndOfDataWithValidationTest.testMessageWithoutEndOfDataByte

essobedo:
> The build error is a know flake on java 21
> 
> org.apache.camel.component.mllp.MllpTcpServerConsumerOptionalEndOfDataWithValidationTest.testMessageWithoutEndOfDataByte

If I revert my changes on the affected class the tests pass so it looks like there is a regression due to my changes so for now,  I prefer to revert my changes

PR #16851, opened 2025-01-17
Log container to STDOUT + List and Stop running services

Croway:
- Use ~/.camel to handle running infra service processes, the infra process start with infra- and have the following structure infra-{serviceName}-{pid}.log
- Add camel-test-infra-common to camel-jbang-core, this will add testcontainer as a dependency, @davsclaus is it doable? otherwise, I'll refactor the code to use reflection heavily. If you do not see an issue having testcontainer as dependency in camel-jbang-core, I'd move CamelLogConsumer into camel-jbang-core
- By default the log file will contain the container log, I added a flag --log that output the log to the stdout.
- Added a stop command, can be used like `camel infra stop minio` that will delete the file and kill the process (@davsclaus I used a slightly different approach compared to StopProcess).
- Added a --running parameter to the list command that shows only the running services

```
$ camel infra run minio
Starting service minio
{
  "accessKey" : "testAccessKey",
  "host" : "localhost",
  "port" : 32835,
  "secretKey" : "testSecretKey"
}
Press any key to stop the execution

$ camel infra list --running
 ALIAS             IMPLEMENTATION  DESCRIPTION
 minio                             MinIO Object Storage, S3 compatible
 
$ camel infra stop minio
Shutting down service minio (PID: 28844)
```

I'd remove the log `Press any key to stop the execution` since infra stop is implemented, wdyt?


oscerd (approved):

davsclaus:
Maybe use `ps` or `get` as we do for camel

```
camel infra ps
```

Show just some basic information

```
camel infra get
```

Should more details

Then its the same as for camel run

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraStop.java
```diff
@@ -0,0 +1,63 @@
...
+        for (File pidFile : CommandLineHelper.getCamelDir().listFiles(
+                (dir, name) -> name.startsWith("infra-" + serviceToStop))) {
+
+            String name = pidFile.getName();
+            String pid = name.substring(name.lastIndexOf("-") + 1, name.lastIndexOf('.'));
+
+            printer().println("Shutting down service " + serviceToStop + " (PID: " + pid + ")");
+            FileUtil.deleteFile(pidFile);
+
+            ProcessHandle.of(Long.valueOf(pid)).ifPresent(ph -> ph.destroyForcibly());
```
davsclaus, line 52: Does this give test container time to stop the service cleanly, as if you destroy the pid by force then would that not be something like kill -9 ?

dsl/camel-jbang/camel-jbang-core/pom.xml
```diff
@@ -151,6 +151,12 @@
             <artifactId>plexus-xml</artifactId>
         </dependency>
 
+        <dependency>
```
davsclaus, line 154: Yeah we need to not have this dependency. 

You can dynamic download this JAR via that camel maven thing, and we can maybe try set its scope = provided so camel run will not use it normally. And then you can still compile and use its API just that you have downloaded the JAR first. 

Otherwise we can also move the command to its own camel jbnag plugin then the dependency can be there (just k8s has kube client JAR etc). But then users need to install this plugin first before they can use it.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraRun.java
```diff
@@ -150,14 +137,43 @@ private void run(String testService, String testServiceImplementation) throws Ex
...
-            actualService.getClass().getMethod("shutdown").invoke(actualService);
-            sc.close();
+        while (!sc.hasNext()) {
         }
+
+        ((InfrastructureService) actualService).shutdown();
+        sc.close();
+    }
+
+    private static @NotNull DependencyDownloaderClassLoader getDependencyDownloaderClassLoader(
```
orpiske, line 161: Just a heads up that there is a `@NotNull` annotation here (IntelliJ keeps adding those from time to time).

oscerd (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraRun.java
Croway: thanks, I removed that

dsl/camel-jbang/camel-jbang-core/pom.xml
Croway: I removed the dependency, it is still possible to run everything as expected via reflection (thanks to the jar download), I'll add some smoke tests for the reflection part.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraStop.java
Croway: Yeah, Forcibly is too much, I used ph.destroy() instead and I didn't find any issue, luckily, testcontainers, handles well the kill of the main process, I think that the sidecar image ryuk takes care of it.

Croway:
> Maybe use `ps` or `get` as we do for camel
> 
> ```
> camel infra ps
> ```
> 
> Show just some basic information
> 
> ```
> camel infra get
> ```
> 
> Should more details
> 
> Then its the same as for camel run

Thanks a lot for the input, I added
```
$ camel infra ps

 ALIAS             IMPLEMENTATION  DESCRIPTION
 arangodb                          ArangoDB is a multi-model database for high-performance applications.
```
that prints the running services, and

```
$ camel infra get arangodb

{
  "getHost" : "localhost",
  "getPort" : 8529,
  "getServiceAddress" : "localhost:8529"
}
```
that displays the information of the running service.

I was thinking about adding a --log to the `camel infra get` that simply tails the containers log dumped into ~/.camel/infra-arangodb-1111.log, wdyt? would it make sense?

davsclaus:
yeah a `camel infra log` would be good. If only 1 service is running it can log that, if there are 2+ then it can maybe fail, or log both of them together. You can then do `camel infra log arangodb` to only log this if you have 2+ services.

Croway:
> yeah a `camel infra log` would be good. If only 1 service is running it can log that, if there are 2+ then it can maybe fail, or log both of them together. You can then do `camel infra log arangodb` to only log this if you have 2+ services.

Thanks a lot for the input, I added the command `camel infra log`, as per documentation, if no service is provided, all the services log will be displayed and tailed in the same CLI in a "docker-compose" fashion
```[service1] ...
[service2] ...
```

otherwise, `camel infra log service` will log only the service.

In case of stop or deletion of the .log file in the .camel folder, the command will stop automatically

davsclaus (approved):

davsclaus:
LGTM

orpiske (approved):

PR #16901, opened 2025-01-23
feat(dsl): camel jbang export property

squakez:
You can now use `camel export --property xyz=abc`

Closes CAMEL-21645

# Description

# Target

# Tracking


squakez:
> I wonder if --property should override any existing value that may already be when writing the application.properties
> 
> so if you already have `hello = peter` and then export with --property=hello=jack --- then the written file will a single property with jack as the value. Otherwise you have duplicates.

The fact is that, at that stage the application.properties is completely blank as we're exporting an app from scratch. The presence of duplicates can only happen if the user forcefully provide twice the same parameter key.

davsclaus:
You can export foo.java application.properties and other files together. Either explicit name the files, or use `*` etc

squakez:
> You can export foo.java application.properties and other files together. Either explicit name the files, or use `*` etc

I've tested against an export with `application.properties` and I think this is working correctly as it overrides the existing value since the parameter should have priority over the existing values. I created a test to validate this behavior.

davsclaus (approved):

PR #16907, opened 2025-01-24
Accept Quarkus and SpringBoot versions as system properties

bartoszpop:
Quarkus and SpringBoot versions can't be defined in jbang-catalog.json because of [jbangdev#1910](https://github.com/jbangdev/jbang/issues/1910). As a fix Camel JBang should accept the version defined as a system property

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
@@ -63,6 +64,7 @@ import org.apache.camel.util.FileUtil;
 import org.apache.camel.util.IOHelper;
 import org.apache.camel.util.StringHelper;
 import picocli.CommandLine;
+import picocli.CommandLine.Option;
 
 public abstract class ExportBaseCommand extends CamelCommand {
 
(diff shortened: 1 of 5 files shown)


davsclaus:
what are you trying to do. The option can be defined as CLI argument for a CLI tool - its not intended as a JVM system property

bartoszpop:
@davsclaus we need to have control over Quarkus/SpringBoot version via the jbang-catalog.json file. Currently, the only option to set the version is to use "--quarkusVersion" but it won't work with jbang-catalog.json because of [jbangdev#1910](https://github.com/jbangdev/jbang/issues/1910). JBang team advised to use system property, similarly as Camel JBang does for Camel and Kamelets Catalog version.

davsclaus:
Who are we - the came-catalog file is for ASF Camel only. You can make your own catalog file

davsclaus:
You should use camel config set where you can configure any options you want to have special

bartoszpop:
@davsclaus _camel config set_ is not sufficient as the user needs invoke this command explicitly. The requirement we have is to deliver a ready to use jbang-catalog.json. What is your concern about having it configurable with system property? It would be consistent with how the camel and kamelet versions work (see [CamelJBang.java](https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-main/src/main/jbang/main/CamelJBang.java)).

bartoszpop:
[@davsclaus](https://github.com/davsclaus) camel config set is not sufficient as the user needs invoke this command explicitly. The requirement we have is to deliver a ready to use jbang-catalog.json. What is the concern about having it configurable with system property? It would be consistent with how the camel and kamelet versions are configurable now (see [CamelJBang.java](https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-main/src/main/jbang/main/CamelJBang.java)). Because Quarkus and SpringBoot versions are hardcoded in [RuntimeType.java](https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/RuntimeType.java), hence we can't get a new Camel Quarkus version until a new Camel version is released.

PR #16945, opened 2025-01-28
CAMEL-21659: introduce deadline concept to TimeoutExtender

simon-ras:
# Description

Add a `deadline` to each entry to be extended via `TimeoutExtender`. This timeout is set to now plus 80% duration of the visibility timeout. Any message that would reach this deadline by the next execution of the TimeoutExtender is extended.

This cuts down the number of extend requests dramatically as the previous implementation would extend ALL messages, including those polled shortly before the execution of the extender.

This PR also changes the logging behavior in case the message could not be extended due to `ReceiptHandleIsInvalid`. This effectively aligns the logic to how it worked prior to using the batched operation where it would catch the exception `ReceiptHandleIsInvalidException`. The batch operation however does not throw this exception, instead the collection of failed entries has to be checked.

# Target

# Tracking
- [x] CAMEL-21659

### DIFF
--- a/components/camel-aws/camel-aws2-sqs/src/test/java/org/apache/camel/component/aws2/sqs/integration/SqsComponentLocalstackIT.java
+++ b/components/camel-aws/camel-aws2-sqs/src/test/java/org/apache/camel/component/aws2/sqs/integration/SqsComponentLocalstackIT.java
@@ -24,6 +24,7 @@ import org.apache.camel.ProducerTemplate;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.aws2.sqs.Sqs2Constants;
 import org.apache.camel.component.mock.MockEndpoint;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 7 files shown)


components/camel-aws/camel-aws2-sqs/src/test/java/org/apache/camel/component/aws2/sqs/integration/SqsComponentLocalstackIT.java
```diff
@@ -38,15 +39,16 @@ public class SqsComponentLocalstackIT extends Aws2SQSBaseTest {
     private MockEndpoint result;
 
     @Test
+    @Disabled("Seem to depend on order, executed by itself, then it passes")
```
simon-ras, line 42: This test fails when running on fresh `main` and does not have anything to do with this PR.

I cannot however, figure out why it is not failing on CI. Would be grateful if anyone could checkout what is going on with this test.

simon-ras:
This change has had about 24hours of testing in a production environment where it has received both high and low load with DEBUG enabled of `Sqs2Consumer`. Tested with `visibilityTimeout` of 20 seconds with a delay of 10s between `TimeoutExtender` executions.

davsclaus (approved):

oscerd (approved):

components/camel-aws/camel-aws2-sqs/src/test/java/org/apache/camel/component/aws2/sqs/integration/SqsComponentLocalstackIT.java
simon-ras: Done

davsclaus:
Can you send a PR for 4.8.x branch

PR #16947, opened 2025-01-28
CAMEL-21505: corrupt binary files

jonomorris:
# Description

Updated AS2 entities so that message content is held in a byte array allowing entities to support non-text content, e.g. a zip file.   

If the `Content-Transfer-Encoding` header is set to 'base64' the message content is base64 encoded for transport and decoded back to a byte array when received.  If using base64 encoding we assume binary-data message content, and AS2 entities provide this to the AS2Consumer as an `InputStream`.

# Target

# Tracking


davsclaus (approved):

davsclaus:
great work thanks

PR #16973, opened 2025-01-29
CAMEL-21683: Configure jax-rs consumer via CxfRsEndpoint bean

Croway:
This fixes https://camel.zulipchat.com/#narrow/channel/257301-camel-spring-boot/topic/How.20to.20implement.20rest.20and.20soap.20routes.20in.20the.20same.20project/near/496322324


davsclaus (approved):

ffang:
Hi @Croway ,
LGTM! Can we have a testcase for this usage?

Thanks!
Freeman

Croway:
> Hi @Croway , LGTM! Can we have a testcase for this usage?
> 
> Thanks! Freeman

Good point, I added the test case

PR #17028, opened 2025-02-03
Upgrade to kafka 3.9.0 container images

davsclaus:
# Description

# Target

# Tracking

### DIFF
--- a/test-infra/camel-test-infra-kafka/src/main/resources/org/apache/camel/test/infra/kafka/services/container.properties
+++ b/test-infra/camel-test-infra-kafka/src/main/resources/org/apache/camel/test/infra/kafka/services/container.properties
@@ -14,6 +14,6 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
-kafka3.container=mirror.gcr.io/apache/kafka:3.8.1
+kafka3.container=mirror.gcr.io/apache/kafka:3.9.0
 redpanda.container.image=mirror.gcr.io/redpandadata/redpanda:v24.1.16
-strimzi.container.image=quay.io/strimzi/kafka:latest-kafka-3.8.1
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus:
something weird is happening - anyone welcome to try locally if they can have 3.9.0 containers working.

claudio4j:
I'm looking at why the 3.9.0 is failing to start.

claudio4j:
The error when starting the kafka container, from `target/camel-kafka-test.log`

```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: advertised.listeners cannot use the nonroutable meta-address 0.0.0.0. Use a routable IP address. 
```

There is an open bug to fix it [KAFKA-18281](https://issues.apache.org/jira/browse/KAFKA-18281) and fix is ongoing.

apupier:
> There is an open bug to fix it [KAFKA-18281](https://issues.apache.org/jira/browse/KAFKA-18281) and fix is ongoing.

issue marked as fixed in 3.9.1 and 4.0.0
3.9.1 not yet released.
4.0.0 released (but maybe a too big version jump?)

claudio4j:
I did local tests with the 4.0.0 container image and all tests passed. It should be ok to update container image and kafka api lib to 4.x.

davsclaus:
Even if kafka 4 is out now, then IMHO it would be better to use 3.x container images for testing so it stays together with the camel-kafka client versions. 

The big v4 upgrade will come later. CSB and CEQ need to be confortable with kafka v4 as well before we upgrade as default in camel too.

davsclaus:
Anyone may want to help with this on a linux or an older mac system to give this a go. The M4 seems to not work with these images for me when I resumed this work a while back

davsclaus:
replaced by #18334 18334

PR #17034, opened 2025-02-04
KnownDependenciesResolver to support packages

bartoszpop:
This pull request enhances KnownDependenciesResolver to match the well known dependencies against packages up to the root package, so that a single entry in the camel-main-known-dependencies.properties file is sufficient to load any class from this jar, instead of having a separate entry for each class.


davsclaus:
we are not changing the LTS branch

PR #17035, opened 2025-02-04
KnownDependenciesResolver to support packages

bartoszpop:
This pull request enhances KnownDependenciesResolver to match the well known dependencies against packages up to the root package, so that a single entry in the camel-main-known-dependencies.properties file is sufficient to load any class from this jar, instead of having a separate entry for each class.

This is a cherry-pick from #17034

### DIFF
--- a/dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/KnownDependenciesResolver.java
+++ b/dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/KnownDependenciesResolver.java
@@ -66,7 +66,7 @@ public final class KnownDependenciesResolver {
 
     public MavenGav mavenGavForClass(String className) {
         MavenGav answer = null;
-        String gav = mappings.get(className);
+        String gav = findGav(className);
         if (gav != null) {
             answer = MavenGav.parseGav(gav, camelContext.getVersion());
         }
@@ -79,5 +79,14 @@ public final class KnownDependenciesResolver {
             }
         }
         return answer;
+    }
+
+    private String findGav(String prefix) {
+        String gav = mappings.get(prefix);
+        while (gav == null && prefix.lastIndexOf(".") != -1) {
+            prefix = prefix.substring(0, prefix.lastIndexOf("."));
+            gav = mappings.get(prefix);
+        }
+        return gav;
     }
 }
--- a/dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KnownDependenciesResolverTest.java
+++ b/dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KnownDependenciesResolverTest.java
@@ -1,0 +1,41 @@
+package org.apache.camel.main.download;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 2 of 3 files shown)


davsclaus:
what is the reason for this, this file is maintained by camel project and not be end users. It is an implementation detail.

davsclaus:
okay do you have an example of a real world situation this was beneficial - just want to understand better.

bartoszpop:
Hi @davsclaus, thank you for your feedback. We utilize this mechanism to inject predefined bean and AggregationStrategy classes to JBang runtime, meaning that we pass the parameter `--dep=com.example:custom-catalog:1.0.0` and is contains `camel-main-known-dependencies.properties` with a list of custom classes pointing to custom jars, e.g.
```
org.example.some.BeanClass=com.example:some-jar:1.0.0
org.example.some.AggregationStrategyClass=com.example:some-jar:1.0.0
org.example.other.BeanClass=com.example:other-jar:1.0.0
```
Instead of having separate entries for each class it would be reasonable to have them merged, because all classes in the same jar will share the same package, e.g.
```
org.example.some=com.example:some-jar:1.0.0
org.example.other=com.example:other-jar:1.0.0
```

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/KnownDependenciesResolver.java
```diff
@@ -80,4 +80,13 @@ public MavenGav mavenGavForClass(String className) {
         }
         return answer;
     }
+
+    private String findGav(String prefix) {
+        String gav = mappings.get(prefix);
+        while (gav == null && prefix.lastIndexOf(".") != -1) {
```
davsclaus, line 86: You cannot keep dropping sub package names, which can end up into picking the wrong JAR. 
It should only drop the classname

davsclaus (changes requested):

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/KnownDependenciesResolver.java
bartoszpop: But the class may be in a subpackage, e.g. consider com.example:somejar:1.0.0 and two classes com.example.somejar.subpackage.SomeClass
com.example.somejar.otherpackage.OtherClass

davsclaus: Hmm its an use-case outside camel-jbang, but it may not cause trouble. But mind if anything is pain then its removed.

The unit test files need to have ASF license headers. And also the properties file.

bartoszpop: I added the missing headers. Pls review

davsclaus:
modified:   dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KnownDependenciesResolverTest.java

davsclaus:
can you update this file as its auto formatted as your code style is not standard camel

davsclaus:
okay i can rebuild afterwards to update the file

bartoszpop:
Thank you @davsclaus

PR #17060, opened 2025-02-05
extends fix of  CAMEL-21713: endpoint url override to aws-secret-manager

JiriOndrusek:
follows https://github.com/apache/camel/pull/17052

The override endpoint url is missing also for `CloudTrailClient`.
The localstack cloudtrail lookup is not implemented yet, therefore endpoint url override cannot be  used now in this case. But I think that it is better to have the code prepared for the future.

# Description

# Target

# Tracking


oscerd (approved):

davsclaus (approved):

davsclaus:
is this something that is good to backport to 4.8.x branch ?

JiriOndrusek:
> is this something that is good to backport to 4.8.x branch ?

It's not required for beckport to 4.8.x from my POV. Very simple workaround is to add parameters with credentials with "fake" values.
On the other hand the fix is so small, that it should be save to backport and it should make users happier when using azure-key-vault with identity.

PR #17070, opened 2025-02-05
Add more headers in platform-http multipart uploads

Croway:


core/camel-api/src/main/java/org/apache/camel/Exchange.java
```diff
@@ -117,6 +117,7 @@ public interface Exchange extends VariableAware {
     String CONTENT_TYPE = "Content-Type";
     String COOKIE_HANDLER = "CamelCookieHandler";
     String CORRELATION_ID = "CamelCorrelationId";
+    String ATTACHMENTS_SIZE = "CamelAttachmentsSize";
```
davsclaus, line 120: I think there is somewhere you need to document this header, try look for some of the other headers.

davsclaus: Also put it up to A headers so its sorted

davsclaus: Ah yeah add it to `ExchangePropertyKey` also and doc via `@Metadata`

components/camel-platform-http/src/main/docs/platform-http-component.adoc
```diff
@@ -91,6 +91,9 @@ Since Apache Camel 4.10, multipart file uploads are easier and harmonized across
 * The Apache Camel message contains:
 ** The uploaded file in the message body
 ** The file name in the "CamelFileName" message header
-** The file's content type in the "Content-Type" message header
+** The file's content type in the "CamelFileContentType" message header
+** The file's size in the "CamelFileLength" message header
```
jamesnetherton, line 95: Nitpick

```suggestion
** The file content type in the "CamelFileContentType" message header
** The file size in the "CamelFileLength" message header
```

core/camel-api/src/main/java/org/apache/camel/Exchange.java
Croway: mmm, I cannot find much documentation for other headers, should I annotate it with Metadata and add a description?

davsclaus: Yes

Croway: https://github.com/apache/camel/pull/17070/files#diff-c4832886119fc6dd6a4a54e68c5e2740ea9ae86b8699a469648284189d51e839 like this? I do not fully understand how `ExchangePropertyKey` works, for example, FILE* related headers are not there.

davsclaus: Its the most commonly used headers - and since HTTP with attachments is fairly common then IMHO it would be okay to add here

davsclaus:
the size header should likely only be set if its > 0,  as then it may be always there, also for just plain REST GET etc

Croway:
> the size header should likely only be set if its > 0, as then it may be always there, also for just plain REST GET etc

The ATTACHMENTS_SIZE header is added only during a multipart request. But I do not fully understand your comment.

davsclaus:
ah okay there was a if multipart check first - so this is fine. Just make sure on CSB that its also only added if really multipart data so its not there for anything else

PR #17071, opened 2025-02-05
CAMEL-21715: Fix test-jar dependencies

christophd:
# Description

- Since we have been moving some classes in the test-infra-* modules from src/test/java to src/main/java we need to explicitly depend on both test-jar and normal typed dependencies
- Maven will figure it out but IDEs such as IntelliJ raise compile errors
- Explicitly add dependency to test-infra-* modules to avoid compile errors in IntelliJ due to test-jar typed dependencies

# Target

Fixes CAMEL-21715

# Tracking

### DIFF
--- a/components/camel-activemq/pom.xml
+++ b/components/camel-activemq/pom.xml
@@ -59,7 +59,12 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
-
+       <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-infra-artemis</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-test-infra-artemis</artifactId>
@@ -67,7 +72,6 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
-
        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-test-infra-messaging-common</artifactId>
--- a/components/camel-activemq6/pom.xml
+++ b/components/camel-activemq6/pom.xml
@@ -52,7 +52,7 @@
             <version>${activemq6-version}</version>
         </dependency>
 
-       <dependency>
+        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-test-infra-core</artifactId>
             <version>${project.version}</version>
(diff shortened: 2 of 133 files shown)


davsclaus:
so we need both test and test-jar dependencies ?

christophd:
this is only for IDEs like IntelliJ because Maven itself will figure out the transitive dependency when the `test-jar` is given in the POM as a dependency. IntelliJ complains with compile errors. I think this is because the IDE strictly separates test and main dependency tree

Croway:
```<dependency>
            <artifactId>camel-test-infra-artemis</artifactId>
            <scope>test</scope>
            <type>test-jar</type>
        </dependency>
        ```
        The test jar one is enough since the following plugin https://github.com/apache/camel/blob/main/test-infra/pom.xml#L108 will assembly both test and main classes into the test-jar

christophd:
@Croway ah was not ware of this. but why is IntelliJ showing those compile errors then? I the assembly disabled when using the `-Dquickly` option?

Croway:
> this is only for IDEs like IntelliJ because Maven itself will figure out the transitive dependency when the `test-jar` is given in the POM as a dependency. IntelliJ complains with compile errors. I think this is because the IDE strictly separates test and main dependency tree

oh, this is strange, actually I didn't notice it in intellij

Croway:
> @Croway ah was not ware of this. but why is IntelliJ showing those compile errors then? I the assembly disabled when using the `-Dquickly` option?

no, it is not disabled with -Dquickly option

christophd:
am I the only one seeing the compile errors in the IDE? for instance when opening `org.apache.camel.component.amqp.AMQPSAuthRouteTest`

Croway:
This is the only issue I see on the ide

image

oscerd (approved):

christophd:
@oscerd @davsclaus do you see compile errors in the IDE, too? I don't want to merge this just because of an issue that is only on my side

davsclaus:
I only opened camel-amqp project (not entire camel) and IDEA has no compile errors when I click "rebuild project"

davsclaus:
IntelliJ IDEA 2024.3.2.2 (Community Edition)
Build #IC-243.23654.189, built on January 29, 2025
Runtime version: 21.0.5+8-b631.30 aarch64 (JCEF 122.1.9)
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.lwawt.macosx.LWCToolkit
macOS 15.3
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
Memory: 8192M
Cores: 14
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
  llm.show.ai.promotion.window.on.start=false
  ide.images.show.chessboard=true
Non-Bundled Plugins:
  com.intellij.ideolog (243.23654.177)
  Docker (243.23654.177)
  DevKit (243.23654.189)
  org.asciidoctor.intellij.asciidoc (0.43.6)
Kotlin: 243.23654.189-IJ

christophd:
okay, then I guess this is not a common issue. I'd suggest to close this PR and the CAMEL-21715 then

oscerd:
I don't see errors in the IDE, on intellij

davsclaus:
okay we can re-open if we somehow this code is needed again

PR #17082, opened 2025-02-06
CAMEL-21730: JBang export cannot validate route that depends on copie…

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21730


tdiesler:
The prosed change, fixes it for export, but not run - changing to draft

oscerd (approved):

gansheer (approved):

gansheer:
@tdiesler Does it need to be backported to 4.10.0 ?

tdiesler:
@davsclaus mentioned on jira, that copying files in deep structures is not something we would like to support. For symetry, I'd backport to 4.10 but we might also want to roll back this change - I'll take another look at this and comment later today.

PR #17114, opened 2025-02-11
feat(components): add camel-telemetry-simple component

squakez:
# Description

ref CAMEL-21670

We're adding a simple telemetry component which can result useful to prove the internal logic of `camel-telemetry` component works correctly. Moreover it is useful for debugging or experimental workflows where a user want to quickly measure the performance of any given Camel route without the need to involve more complex telemetry systems.

# Target

# Tracking


davsclaus:
Can you use some other name than simple, as it would be associated with the camel simple language.

Also the options you add to camel-main should be in the OtelConfigurationProperties

squakez:
> Can you use some other name than simple, as it would be associated with the camel simple language.
> 

Yes sure, let me come with a better naming.

> Also the options you add to camel-main should be in the OtelConfigurationProperties

Given the impossibility to maintain backward compatibility I was thinking to create a new Otel component (let's call it `camel-opentelemetry-2` just for the sake of this discussion). The idea is that we can mature the adoption of this new component and eventually deprecated the actual `camel-opentelemetry`. The new component will implement those options accordingly. In any case, it would be a separate PR and we can kick off a separate thread to understand what is the best strategy to enable adoption and limit potential backward compatibility problems.

davsclaus (approved):

davsclaus:
LGTM

Also an idea could be to add a `dev-console` to this component, then jbang and other tooling would be able to access this and show spans/traces in a nicer way - such as a new camel jbang command.

davsclaus:
See
https://issues.apache.org/jira/browse/CAMEL-20045

PR #17157, opened 2025-02-14
CAMEL-21731: Adding reference to external Open API docs

Croway:
Make it possible to reference external Open API docs via `.apiProperty(..)`

```
restConfiguration()
    .apiProperty("externalDocs.url", "https://openweathermap.org/api")
    .apiProperty("externalDocs.description", "API Documentation"); 
```


davsclaus (approved):

davsclaus:
Yeah that list in JIRA is that all that is potentilally missing? This way with apiProperty makes it flexible and adding to DSL is verbose for features that hardly anybody is using. So lets do as this PR at first. And you can add some docs to the rest-dsl doc page.

Croway:
Sure, thanks for the feedback!

I got that list from https://swagger.io/specification/#fixed-fields

PR #17160, opened 2025-02-14
Removes second call to  DefaultConfigurationConfigurer.configureVaultRefresh

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-21746

# Description

# Target

# Tracking


davsclaus:
This needs a bit more investigation as camel-main and CEQ runs differently on startup - Check git blame, and I need to recall why its added - but I suspected its due to detect vault used in beans on startup.

davsclaus:
okay I had a look and yeah it looks like this 2nd call should be removed.

PR #17161, opened 2025-02-15
INFRA-26486: re-enable develocity

orpiske:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ node_modules/
 mvnd.zip*
 .camel-jbang
 .camel-jbang-run
+.mvn/.develocity/develocity-workspace-id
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (approved):

davsclaus:
A nitpick is that the license header in the xml file could be similar to what we use in pom.xml (without the ~ chars)

orpiske:
> A nitpick is that the license header in the xml file could be similar to what we use in pom.xml (without the ~ chars)

Ahhh, thanks for pointing this. Let's adjust it to make it consistent.

orpiske:
Maybe today ... my CI is unstable lately.

orpiske:
Looking good. Let's merge

PR #17197, opened 2025-02-19
camel-jbang-plugin-kubernetes - jkube to generate Deployment instead of DeploymentConfig

claudio4j:
when the target is an openshift cluster

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
@@ -292,6 +292,9 @@ public class KubernetesExport extends Export {
                 printer().printf("OpenShift forcing --image-builder=docker%n");
                 imageBuilder = "docker";
             }
+            // the deployment trait already generates the src/main/jkube/deployment.yml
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

gansheer (commented):
@claudio4j I was testing the main branch and this change. One issue here is that when deploying on OCP we loose the reference to the created `ImagStream` when switching from `DeploymentConfig` to `Deployment| when using a basic command: `camel kubernetes run TimerLog.java --cluster-type=openshift` and end up with `ErrImagePull` trying to pull the image `"image" : "timer-log:1.0-SNAPSHOT"` :
```
  Normal   Pulling         43s (x4 over 2m17s)  kubelet            Pulling image "http-log:1.0-SNAPSHOT"
  Warning  Failed          42s (x4 over 2m15s)  kubelet            Failed to pull image "http-log:1.0-SNAPSHOT": reading manifest 1.0-SNAPSHOT in docker.io/library/http-log: requested access to the resource is denied
  Warning  Failed          42s (x4 over 2m15s)  kubelet            Error: ErrImagePull
  Warning  Failed          29s (x6 over 2m14s)  kubelet            Error: ImagePullBackOff
  Normal   BackOff         18s (x7 over 2m14s)  kubelet            Back-off pulling image "http-log:1.0-SNAPSHOT"
```
If there is no way to have it deploy corectly on default configuration, I would suggest to wait for your work on automatic cluster detection before this change.

claudio4j:
I will look at this, thanks for reporting.

PR #17216, opened 2025-02-20
CAMEL-21710 camel-jbang-plugin-kubernetes - automatic cluster detection

claudio4j:
Set parameters when connected to specific cluster
https://issues.apache.org/jira/browse/CAMEL-21710

# Description

# Target

# Tracking

### DIFF
--- a/docs/user-manual/modules/ROOT/pages/camel-jbang-kubernetes.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-jbang-kubernetes.adoc
@@ -1296,7 +1296,9 @@ camel kubernetes run route.yaml --image-registry=kind
 ----
 
 When connecting to a local Kubernetes cluster, you may need to specify the image registry where the application container image gets pushed to.
-The run command is able to automatically configure the local registry when using predefined names such as `kind` or `minikube`.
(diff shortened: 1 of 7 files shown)


claudio4j:
As there are other cluster types like "kind", minikube with podman and many other flavors, if there are enough interest, they can be supported too.

davsclaus (approved):

gansheer (commented):
I'm wondering if the delete command should also have this automatic detection. In case of Openshift we have to add the `cluster-type` parameter for the delete to work correctly (see documentation https://camel.apache.org/manual/camel-jbang-kubernetes.html#_delete_deployments). I think it would make sense to have the same if we apply a cluster-type automatically on `run` command we do the same on `delete` command.
WDYT ?

claudio4j:
I looked at the delete command, which I think should be redesigned, see [discussion on zulip](https://camel.zulipchat.com/#narrow/channel/352237-camel-jbang/topic/k8s.20delete.20command.20relies.20on.20temp.20files/near/500906805).

gansheer:
> I looked at the delete command, which I think should be redesigned, see [discussion on zulip](https://camel.zulipchat.com/#narrow/channel/352237-camel-jbang/topic/k8s.20delete.20command.20relies.20on.20temp.20files/near/500906805).

Ok, then adding a warning about the auto detection on delete command in the doc will do for now.

claudio4j:
@gansheer I'm working on the delete command which I will push for review on a different PR.

gansheer (approved):

PR #17262, opened 2025-02-25
CAMEL-21769: camel-aws-sqs - Error is causing the sqs message to be extended forever

narsinallamilli:
# Description

Change will stop exchange from extending after an error occurred in processing.

# Target

# Tracking


davsclaus (changes requested):
do not catch Error - if you have an out of memory error then you cannot avoid the JVM not working correctly afterwards and you need to let the JVM crash

narsinallamilli:
> do not catch Error - if you have an out of memory error then you cannot avoid the JVM not working correctly afterwards and you need to let the JVM crash

Yes, I am rethrowing it.
https://github.com/apache/camel/pull/17262/commits/f150570c41c70bd84cc5d6f148c4811a53902440

davsclaus (approved):

oscerd (approved):

PR #17265, opened 2025-02-25
[CAMEL-21740] Check initial replay id validity on consumer startup

a-jubar:
# Description
I have added a new checkInitialReplayIdValidity method that gets executed once during the consumer startup if the provided ReplayPreset is CUSTOM.
If the provided intial replay id is corrupt/invalid an exception is thrown and Camel shuts down. (The same behaviour as when a null initialReplayId is provided.)
This workaround is just suggestion. It feels a bit hacky to me, but seems to be the best solution.

I think the error handling of this commit https://github.com/apache/camel/commit/546a5016f05ce7bfc5a663bfd6ae6a0bf1d4d3e0 will not work like intended because the method resubscribeOnError() will always set the ReplayPreset to CUSTOM if replayId is not null.
Sorry for being a bit late :)

# Target

# Tracking

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
@@ -24,8 +24,11 @@ import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.protobuf.ByteString;
 import com.salesforce.eventbus.protobuf.ConsumerEvent;
 import com.salesforce.eventbus.protobuf.FetchRequest;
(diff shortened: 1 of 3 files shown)


components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
```diff
@@ -174,6 +181,58 @@ public void subscribe(PubSubApiConsumer consumer, ReplayPreset replayPreset, Str
...
+                    if (trailers != null && PUBSUB_ERROR_CORRUPTED_REPLAY_ID
+                            .equals(trailers.get(Metadata.Key.of("error-code", Metadata.ASCII_STRING_MARSHALLER)))) {
+                        error.set(t);
+                    }
+                }
+                latch.countDown();
+            }
+
+            @Override
+            public void onCompleted() {
```
davsclaus, line 210: Just to be sure, so if the reply id is valid then onNext is invoked, so you dont need to count down the latch here also

davsclaus (approved):

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
a-jubar: Yes, if the replay id is valid the requested event is returned by the server and onNext gets invoked.

PR #17335, opened 2025-03-03
camel-jbang-plugin-kubernetes: make the delete op to delete from the k8s resources

claudio4j:
Do not use the camel-jbang temporary directory
The "run" op sets labels used in the "del" op
Use a single KubernetesClient instance in the KubernetesHelper

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
@@ -17,93 +17,68 @@
 
 package org.apache.camel.dsl.jbang.core.commands.kubernetes;
 
-import java.io.File;
-import java.io.FileInputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import io.fabric8.kubernetes.api.model.StatusDetails;
+import io.fabric8.openshift.client.OpenShiftClient;
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
-import org.apache.camel.util.ObjectHelper;
+import org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.BaseTrait;
 import org.apache.camel.util.StringHelper;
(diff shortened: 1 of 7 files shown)


davsclaus (approved):

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
```diff
@@ -17,93 +17,68 @@
...
-            if (candidateDir.isDirectory()) {
-                resolvedManifestDir = candidateDir;
+        printer().printf("Deleting all resources from app: %s%n", appName);
+        Map<String, String> labels = new HashMap<>();
+        // this label is set in KubernetesRun command
+        labels.put(BaseTrait.KUBERNETES_LABEL_MANAGED_BY, "camel-jbang");
+        labels.put("app", appName);
+        List<StatusDetails> deleteStatuses = new ArrayList<>();
+        try {
+            // delete the most common resources
```
christophd, line 55: What about resources like Secret, ConfigMap and other resources that have been created as part of the Kubernetes export/run manifest?

christophd (commented):
This would now only delete Deployment and Service resources right? I wonder if there are other resources that are left on the cluster after the delete operation. I think the Camel JBang Kubernetes export/run command is able to create other resources as part of the traits, too.

gansheer (approved):

claudio4j:
ok, I'm looking at the configmap/secrets and knative resources.

christophd:
The idea of the previous implementation was to delete whatever resource has been added to the Kubernetes export manifest. This is why the delete command has been looking for the temp export directory in order to retrieve the manifest file.

I guess we should make sure no resources are left on the cluster after the delete operation.

claudio4j:
While working to deploy a knative-service to be able to delete it I had found numerous problems with knative-service not working at all, see [CAMEL-21841](https://issues.apache.org/jira/browse/CAMEL-21841), [CAMEL-21835](https://issues.apache.org/jira/browse/CAMEL-21835), which requires investigation. I had already fixed other problems, but will go in a different PR.
That said, the delete op to handle the knative and knative-service will be handled in a different PR once these traits would be working ok.

davsclaus:
LGTM

PR #17370, opened 2025-03-06
CAMEL-21839: there may be a body even if the tools are not called

orpiske:


oscerd (approved):

davsclaus:
@orpiske is this good or do you need more work on this ?

orpiske:
I am going back to this tomorrow. I have a few things still on the queue that I need to dispatch before reviewing this results.

orpiske:
This should be ready to merge once the checks are complete.

PR #17373, opened 2025-03-06
CAMEL-21847 Implemented try-with-resources for HttpClient. 

dk2k:
Extricated repeating string literal to a constant


davsclaus (changes requested):
This is wrong, the client should not be closed here.

The problem is that the configureHttpClient should have the HttpClient provided as a parameter that it can use inside the request interceptor.

So we need a different solution to this.

davsclaus:
okay here is a fix as the http client should only be closed when the oauth is no longer in use, when the endpoint is stopped
https://github.com/apache/camel/commit/1662b2c5cf90e21dac8a69a683d192a66ca6090d

PR #17389, opened 2025-03-07
chore(doc): MDC documentation and proposal

squakez:
# Description

Ref CAMEL-21827

* Add documentation for MDC usage
* Add a proposal for enhancing the MDC design

# Target

# Tracking

### DIFF
--- a/docs/user-manual/modules/ROOT/pages/mdc.adoc
+++ b/docs/user-manual/modules/ROOT/pages/mdc.adoc
@@ -1,0 +1,36 @@
+= Mapped Diagnostic Context (MDC)
+
+The Mapped Diagnostic Context is a technology used in Java to provide a set of customized information into each log trace. The major logging frameworks implements it, and, although it may have certain limitations, this technology is used to enhance the logging and monitoring of a Java application (Camel applications included).
+
+The main limitation of this technology is the fact that it stores values on a context that is available at thread level. Since Camel is an application that manages multiple thread, when it deals with asynchronous calls, the context propagation may not work correctly.
+
+NOTE: the framework should generally handle MDC correctly. However, there could be components (eg, tracing components) and other asynchronous parts of the system that still require the implementation of the context propagation: please report if you notice anything wrong.
+
+== How to configure in Camel application
+
+The first thing you need to do is to enable the `camel.main.useMdcLogging=true`. This flag will automatically include in the MDC context the following Exchange information:
+
+* camel.breadcrumbId
+* camel.exchangeId
+* camel.messageId
+* camel.correlationId
+* camel.routeId
+* camel.stepId
+* camel.contextId
+* camel.transactionKey
+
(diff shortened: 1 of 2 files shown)


squakez:
@lsergio this may be solving the issues you're facing with the MDC usage. I'd like to have your opinion from a UX perspective, thanks!

lsergio:
@squakez let me see if I got it correctly.... The developer would implement a listener that would be called for every log eip or log 
component call in order to set the mdc. If so, the developer should be aware that if getting the MDC data is an expensive operation, it should be cached. Usually with MDC we set the parameters only once per thread.

One question: would this approach work for logs generated by any component? Like when a http call fails, and the http component logs the error with the stack trace, having the MDC on those logs as well would be important to have the full understanding of what happened during the transaction.

lsergio:
I was also wondering if the listener idea could be applied to thread creation instead. 

Camel would use a thread factory that creates an specific kind of thread which would invoke the listeners. Something like:

```java
interface ThreadListener {
	
	void onThreadStarted(Thread t);
	
	void onThreadCompleted(Thread t);
}


class CamelThread extends Thread {
	
	private Runnable task;
	private ThreadListener listener;

	public CamelThread(Runnable task, ThreadListener listener) {
		this.task = task;
		this.listener = listener;
	}
	
	@Override
	public void run() {
		
		listener.onThreadStarted(this);
		try {
			task.run();
		} finally {
			listener.onThreadCompleted(this);
		}
	}
}


public class Threads {

	
	public static void main(String[] args) throws InterruptedException {
		
		Runnable r = () -> { System.out.println("This is where the async code runs");};
		
		ThreadListener listener = new ThreadListener() {
			
			@Override
			public void onThreadStarted(Thread t) {
				System.out.println("Sets the MDC info");
				
			}
			
			@Override
			public void onThreadCompleted(Thread t) {
				System.out.println("Clears the MDC info");
				
			}	
		};
		
		CamelThread thread = new CamelThread(r, listener);
		thread.start();
                thread.join();
		
	}
}
```	

Of course the core codebase is really large and complex and I may not be taking into consideration a lot of aspects.

lsergio:
Another point that may not be solved by the solutions, although it is not a problem I'm facing.

If I had two routes in the same yaml dsl file, would I be able to use different listeners for each one, for being able to set different mdc properties?

squakez:
> @squakez let me see if I got it correctly.... The developer would implement a listener that would be called for every log eip or log
> component call in order to set the mdc. If so, the developer should be aware that if getting the MDC data is an expensive operation, it should be cached. Usually with MDC we set the parameters only once per thread.

Hello @lsergio. No, the developer is not required to do any low level development. The idea is quite the opposite. Basically, if we agree on this design, the developer will just need to include any information he/she wants to include into the Exchange (ie, an header). The framework will take care to include this into the context, hence, any logging trace will contain the MDC variables. The framework will be in charge to clear it as well.

Again, the idea of this (or any alternative approach) is to give the user a consistent way of using the feature in every DSL and without requiring to know the low details of the core or even the Java language.

lsergio:
@squakez I see it. yep, that sounds good and way easier than today.

davsclaus (approved):

PR #17402, opened 2025-03-10
CAMEL-21739: camel-ftp - Fix leading / in some cases

graben:
Hi @davsclaus,

follow-up bug fix to CAMEL-21739 which could cause leading slashes to get produced accidentally. If a new issue is needed, I'll open it.

Regards,
Benjamin


davsclaus (changes requested):
Can you provide an unit test or at least something that better show this problem as the last PR caused test failures and was not correct. So we would be more cautions to merge PRs this time

graben:
Sure, but I think the test framework still has problems with my windows client to "really" proper work. So I have to build tests with a bit of uncertainty. :(

davsclaus:
okay if you can do windows testing, and then we can do linux testing on this PR afterwards via this CI and other ENGs that has linux/mac computers

graben:
I added a new test case that should have a leading / without my patch and should work in the CI pipe, hopefully.

davsclaus (approved):

PR #17415, opened 2025-03-11
CAMEL-21811: camel-oauth - Prefer Ingress over TLS configured on Keyc…

tdiesler:
…loak directly

https://issues.apache.org/jira/browse/CAMEL-21811

These changes affect keycloak & kafka infrastructure setup through helm charts - they are not java code related.
It is important to access Keycloak over SSL from any OAuth client. That's why we need KC behind an SSL terminating proxy (e.g. Traefik)


oscerd (approved):

davsclaus (approved):

tdiesler:
yes

PR #17416, opened 2025-03-11
CAMEL-21812: camel-jbang - Provide an option in camel k8s to trust a …

tdiesler:
…given cert

https://issues.apache.org/jira/browse/CAMEL-21812

With this change, containers now use an enhanced version of run-java.sh (which we have had in jbang-core templates) as entrypoint for all runtimes.

The entrypoint (run-java.sh) now supports SSL_TRUSTSTORE_CERTIFICATES which you may already know from the Strimzi Kafka image.
When using `jbang camel kubernetes` you add certificates that you want the resulting java process to trust the set of input files and either use the SSL_TRUSTSTORE_CERTIFICATES envar directly or use `--property=ssl.truststore.certificates=...` which will then map to that envar.

For eaxmple like this

```
	jbang camel kubernetes export platform-http-files/* tls/* \
	--dep=org.apache.camel:camel-oauth:4.11.0-SNAPSHOT \
	--gav=examples:platform-http-oauth:1.0.0 \
	--property=camel.oauth.baseUri=https://keycloak.local/kc/realms/camel \
	--property=camel.oauth.redirectUri=http://127.0.0.1:8080/auth \
	--property=camel.oauth.logout.redirectUri=http://127.0.0.1:8080/ \
	--property=camel.oauth.clientId=camel-client \
	--property=camel.oauth.clientSecret=camel-client-secret \
	--property=ssl.truststore.certificates=tls/cluster.crt \
	--trait container.image-pull-policy=IfNotPresent \
	--ignore-loading-error=true \
	--image-builder=docker \
	--image-push=false \
	--runtime=camel-main
```
	
run-java.sh adds the given certificates to the truststore on startup.

Beyond camel-oauth, this generally gives our jbang generated images the ability to connect to arbitrary endpoints with TLS. When needed this could also move a layer down, such that it is generally supported in jbang camel (not limited to containers)


oscerd (approved):

davsclaus (approved):

gansheer (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
```diff
@@ -63,6 +64,7 @@
 import org.apache.camel.util.FileUtil;
 import org.apache.camel.util.IOHelper;
 import org.apache.camel.util.StringHelper;
+import org.apache.commons.io.IOUtils;
```
davsclaus, line 67: Can you avoid using this library and use what we have in Camel already - There is something likely already on IOHelper or elsewhere in Camel to copy an InputStream

davsclaus (changes requested):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportBaseCommand.java
tdiesler: Done

davsclaus (approved):

PR #17418, opened 2025-03-11
CAMEL-21842 camel-jbang-plugin-kubernetes pod startup probe fails because of wrong path

claudio4j:
# Description

# Target

# Tracking


docs/user-manual/modules/ROOT/pages/camel-jbang-kubernetes.adoc
```diff
@@ -180,6 +180,31 @@ The Kubernetes plugin export command also inherits all options from the arbitrar
 
 TIP: See the possible options by running: `camel kubernetes export --help` for more details.
 
+=== Kubernetes Container Health Probes
+
+The https://camel.apache.org/components/next/others/observability-services.html[Observability Services] component is set by default in the project created by the export command, then the health path probes are prefixed with the `/observe` endpoint, regardless if they are Quarkus or Spring Boot.
```
squakez, line 185: The link may fail in `camel-website`. Add the link to the component instead.

squakez: Also, camel-main runtime would be the same.

docs/user-manual/modules/ROOT/pages/camel-jbang-kubernetes.adoc
```diff
@@ -180,6 +180,31 @@ The Kubernetes plugin export command also inherits all options from the arbitrar
...
+----
+/observe/health/live
+/observe/health/ready
+/observe/health/started
+----
+
+For Spring Boot (there is no startup probe)
+[source]
+----
+/observe/health/liveness
```
squakez, line 200: These should be the same of the ones we have for the other runtimes. It is possible that the springboot also exposes those as an alias, but we should make use of the harmonized ones.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
```diff
@@ -331,6 +331,8 @@ public Integer export() throws Exception {
         var jkubeVersion = jkubeMavenPluginVersion(settings, mapBuildProperties());
         buildProperties.add("jkube.version=%s".formatted(jkubeVersion));
 
+        setContainerHealthPaths();
```
squakez, line 334: Why do you need to do this? It defeats the purpose of the component. Ideally you need to include those probes on the Deployment without worrying which is the runtime used.

squakez (commented):
I think there was some misunderstanding on the usage of the component. The goal is to provide the dependency in the pom and any probe on the Deployment if that is required.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesExport.java
claudio4j: jkube relies on quarkus and sb properties to generate the various probes, look for jkube-healthcheck-spring-boot and jkube-healthcheck-quarkus in jkube maven plugin doc. That's why I used it so jkube generates the container probes manifests accordingly.

claudio4j:
The failed checks report the following, but don't show the cause:
```
Processing surefire and failsafe reports to create the summary
Error: Process completed with exit code 1.
```
The build-21.log.zip shows the build finished successfully.

gansheer (approved):

davsclaus (approved):

PR #17419, opened 2025-03-11
CAMEL-21855: camel-jbang - k8s delete may not work any more

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-21855

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
@@ -21,65 +21,61 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import io.fabric8.kubernetes.api.model.StatusDetails;
 import io.fabric8.kubernetes.client.dsl.base.ResourceDefinitionContext;
 import io.fabric8.openshift.client.OpenShiftClient;
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
-import org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.BaseTrait;
-import org.apache.camel.util.ObjectHelper;
 import org.apache.camel.util.StringHelper;
 import org.codehaus.plexus.util.ExceptionUtils;
 import picocli.CommandLine;
 
 import static org.apache.camel.dsl.jbang.core.commands.kubernetes.KubernetesHelper.getKubernetesClient;
+import static org.apache.camel.dsl.jbang.core.commands.kubernetes.traits.BaseTrait.KUBERNETES_LABEL_NAME;
 
 @CommandLine.Command(name = "delete",
-                     description = "Delete Camel application from Kubernetes. This operation will delete all resources associated to this app, such as: Deployment, Routes, Services, etc. filtering by labels \"app.kubernetes.io/managed-by=camel-jbang\" and \"app=<app name>\".",
+                     description = "Delete Camel application from Kubernetes. This operation will delete all resources associated to this app, such as: Deployment, Routes, Services, etc. filtering by label \"app.kubernetes.io/name=<name>\".",
                      sortOptions = false)
 public class KubernetesDelete extends KubernetesBaseCommand {
-
-    @CommandLine.Parameters(description = "The deployed application name", arity = "1", paramLabel = "<app name>")
-    String appName;
 
     public KubernetesDelete(CamelJBangMain main) {
         super(main);
     }
 
     public Integer doCall() throws Exception {
-        printer().printf("Deleting all resources from app: %s%n", appName);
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

oscerd (approved):

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
```diff
@@ -25,41 +25,36 @@
...
         // this label is set in KubernetesRun command
-        labels.put(BaseTrait.KUBERNETES_LABEL_MANAGED_BY, "camel-jbang");
-        labels.put("app", appName);
+        labels.put(KUBERNETES_LABEL_NAME, name);
         List<StatusDetails> deleteStatuses = new ArrayList<>();
         try {
             // delete the most common resources
-            // delete Deployment cascades to pod
-            deleteStatuses.addAll(getKubernetesClient().apps().deployments().withLabels(labels).delete());
+            deleteStatuses.addAll(getKubernetesClient().apps().deployments().withLabels(labels).withLabels(labels).delete());
```
claudio4j, line 52: The `withLabels(labels)` is repeated.

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
```diff
@@ -25,41 +25,36 @@
...
     }
 
     public Integer doCall() throws Exception {
-        printer().printf("Deleting all resources from app: %s%n", appName);
+        printer().printf("Deleting all resources from app: %s%n", name);
         Map<String, String> labels = new HashMap<>();
         // this label is set in KubernetesRun command
-        labels.put(BaseTrait.KUBERNETES_LABEL_MANAGED_BY, "camel-jbang");
-        labels.put("app", appName);
+        labels.put(KUBERNETES_LABEL_NAME, name);
```
claudio4j, line 48: The `KUBERNETES_LABEL_MANAGED_BY` is there, as the `run` command sets this label.
Also, the `manage-by=camel-jbang` label is to make sure these custom resources are set by camel-jbang and not to delete any other deployment by a deployment that sets the same name.

tdiesler: `run` is an extension of `export + package + deploy`, for which manage-by is not camel-jbang, but jkube. There is a hack in the code that silently overrides manage-by=jkube with manage-by=camel-jbang, but only for the run case.

We could filter by manage-by, but only it works for all cases. Accidentally deleting (other) resources with the given name in the same namespace is a non-issue IMHO - we delete deployment + service (and some openshift specific stuff).

Do you have an actual use case where not filtering by manage-by is a problem?

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
tdiesler: Well spotted - thanks

dsl/camel-jbang/camel-jbang-plugin-kubernetes/src/main/java/org/apache/camel/dsl/jbang/core/commands/kubernetes/KubernetesDelete.java
claudio4j: I added the managed-by=camel-jbang since camel-jbang deploys it only for the `run` command. As for the export command, the user may want to use different tools to deploy and manage, so the managed-by label was not changed. The `managed-by` label was added as additional filter for `delete` command to be on the safe side, to delete only the resources run by camel-jbang.

tdiesler: If we must have a [app.kubernetes.io/managed-by](http://app.kubernetes.io/managed-by) filter, it would be 'jkube' which present for the export & run case. imho, it is however quite unlikely that a user deploys a resource 'foo' to a given namespace only to have it mistakenly deleted by camel k8s because that also generates + deploys resource 'foo' of the same type in the same ns. Hence, I'd say [managed-by](http://app.kubernetes.io/managed-by) filter is not necessary.

claudio4j: ok, sounds good to me to drop the managed-by label.

claudio4j (approved):

gansheer (approved):

tdiesler:
Is this still waiting on anything?

PR #17448, opened 2025-03-13
CAMEL-21858: Checking the httpMethod parameter when decorating the sp…

lsergio:
…an for http components

# Description

The code checks if the endpoint uri for http components has the httpMethod parameter. If so, this is used as the http method in the tracing span. If not, we keep the current logic.

# Target

- [X ] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking
- [X ] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

- [ X] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

### DIFF
--- a/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/decorators/AbstractHttpSpanDecorator.java
+++ b/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/decorators/AbstractHttpSpanDecorator.java
@@ -27,8 +27,14 @@ public abstract class AbstractHttpSpanDecorator extends AbstractSpanDecorator {
     public static final String POST_METHOD = "POST";
     public static final String GET_METHOD = "GET";
 
-    public static String getHttpMethod(Exchange exchange, Endpoint endpoint) {
-        // 1. Use method provided in header.
+    public String getHttpMethod(Exchange exchange, Endpoint endpoint) {
+        //  1. Use method from httpMethod param, if present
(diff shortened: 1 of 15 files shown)


squakez (commented):
Thanks for the PR. I think the fix could be fine, but I'd avoid to apply it on the `AbstractHttpSpanDecorator` since this is something specific to how Camel manage the `camel-http` instead. IMO, the abstraction is fine as it is, as the idea of having a request parameter for http method is not something part of the http specification, but it is specific to `camel-http` component instead. If you move the logic into the `HttpSpanDecorator` I think it would be good though.

lsergio:
Thanks for the feedback, @squakez . I implemented the change in the camel-telemetry component as well, and also moved the httpMethod handling  to the HttpSpanDecorator.

davsclaus:
Can you add this for camel-vertx-http which has similar `httpMethod` option and same logic for determine method that was used

lsergio:
@davsclaus done!

squakez (approved):

lsergio:
I see the builds have failed, but it does not seem to be related to my changes. I rebased my branch, just in case.

davsclaus (approved):

davsclaus:
yeah it seems the build tests passed but somehow the job returned an error code in the shell.

davsclaus:
ah it was a doc build

PR #17528, opened 2025-03-21
CAMEL-21884: camel-platform-http - Introduce useBodyHandler option

zhfeng:
# Description

# Target

# Tracking


zhfeng:
component-test is suck ?

components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/DefaultHttpRequestBodyHandler.java
```diff
@@ -27,21 +27,27 @@
...
 /**
- * Default {@link HttpRequestBodyHandler} that will read to read the entire HTTP request body into memory.
+ * Default {@link HttpRequestBodyHandler} that will read the entire HTTP request body into memory if useBodyHandler is
+ * enabled.
  */
 class DefaultHttpRequestBodyHandler extends HttpRequestBodyHandler {
-    DefaultHttpRequestBodyHandler(Handler<RoutingContext> delegate) {
+    final boolean useBodyHandler;
+
+    DefaultHttpRequestBodyHandler(Handler<RoutingContext> delegate, boolean useBodyHandler) {
```
jamesnetherton, line 36: Nitpick - I think it'd be better to have a dedicated class, like:

```java
class NoOpHttpRequestBodyHandler extends HttpRequestBodyHandler {
    NoOpHttpRequestBodyHandler(Handler<RoutingContext> delegate) {
        super(delegate);
    }

    @Override
    void configureRoute(Route route) {
    }

    @Override
    Future<Void> handle(RoutingContext routingContext, Message message) {
        return Future.succeededFuture();
    }
}
```

That way you don't pollute `DefaultHttpRequestBodyHandler` with `useBodyHandler` checks. WDYT?

jamesnetherton (commented):
I left a comment as a potential improvement.

Also we need tests for this.

oscerd:
> component-test is suck ?

It seems it cannot find camel-buildtools 4.11.0-SNAPSHOT, but it's there https://repository.apache.org/content/repositories/public/org/apache/camel/camel-buildtools/4.11.0-SNAPSHOT/

components/camel-platform-http-vertx/src/main/java/org/apache/camel/component/platform/http/vertx/DefaultHttpRequestBodyHandler.java
zhfeng: Hi @jamesnetherton ,

I create such a test 
```java
public class VertxPlatformHttpNoBodyHandlerTest {
    private final int port = AvailablePortFinder.getNextAvailable();
    private final WireMockServer wireMockServer = new WireMockServer(options().port(port));

    @BeforeEach
    void before() {
        wireMockServer.stubFor(post(urlPathEqualTo("/test"))
                .withRequestBody(containing("Hello World"))
                .willReturn(aResponse()
                        .withBody("This is a test")));

        wireMockServer.start();
    }

    @AfterEach
    void after() {
        if (wireMockServer != null) {
            wireMockServer.stop();
        }
    }

    @Test
    void testNoBodyHandler() throws Exception {
        final CamelContext context = VertxPlatformHttpEngineTest.createCamelContext();
        final var mockUrl = "http://localhost:" + wireMockServer.port();

        try {
            context.addRoutes(new RouteBuilder() {
                @Override
                public void configure() {
                    from("platform-http:/camel?matchOnUriPrefix=true&useBodyHandler=false")
                            .process(exchange -> {
                                HttpMessage message = exchange.getMessage(HttpMessage.class);
                                message.setBody(message.getRequest());
                            })
                            .removeHeader("CamelHttpUri")
                            .setHeader("OrgCamelHttpUri", simple(mockUrl + "${header.CamelHttpPath}"))
                            .setHeader("CamelHttpPath", simple(""))
                            .toD("${bean:" + PathCreator.class.getName()
                                 + "?method=createNewUri(${header.OrgCamelHttpUri})}?bridgeEndpoint=true");
                }
            });

            context.start();

            given()
                    .body("Hello World")
                    .post("/camel/test")
                    .then()
                    .statusCode(200)
                    .body(is("This is a test"));
        } finally {
            context.stop();
        }
    }
}
```

But the test is hanging and can not get any response. When I remove the line `message.setBody(message.getRequest());`, it starts to work. I'm struggling for a while. Can you give me some lights?

zhfeng: Thanks @jamesnetherton - I just push my commits with the above test on https://github.com/zhfeng/apache-camel/tree/CAMEL-21884-tests

jamesnetherton: I am confused as to what's making the request hang. It's maybe related to the setup and configuration of `VertxPlatformHttpServer` used in the tests.

If I run a similar test on Camel Quarkus, it seems to work. Likewise If I create just a plain Vert.x Server -> Client test (E.g no Camel), it also works.

zhfeng: Hmm, this is WARN message in the log
```
[worker-thread-0] WARN  VertxImpl - You're already on a Vert.x context, are you sure you want to create a new Vertx instance?
```

any hint?

jamesnetherton: > any hint?

It just means that the platform-http component and the vertx-http component created their own Vert.x instances. You can avoid the waning by binding a Vertx instance to the registry.

zhfeng: The root case is that we are running the toD in the `vertx-worker-thread` , and there is a `HttpConnetion` which is  running in `vertx-event-loop-thread`  still reads the content from the network. So if the vertx-http forwards the body with `HttpServerRequest` after `vertx-event-loop-thread`, it can not read any byte and is hanging. Then the fix is to pause the http request in the `NoOpHttpRequestBodyHandler`.

jamesnetherton (approved):

PR #17593, opened 2025-03-27
CAMEL-21896: camel-jbang - Allow configuration preset per directory

Croway:
Added the following options:

* `--mergeConfigurations` loads the configurations default one _~/.camel-jbang-user.properties_ and the one in the current directory (if exists) _./.camel-jbang-user.properties_, the latter is prioritized. This config is a bit hacky, it is used to preload the configuration before the actual picocli execution, if the argument mergeConfigurations exists, the configurations are merged, and the argument removed before passing the args to Picocli.
* `--local` Use the configuration in current directory (to be used with `camel config` commands)

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigSet.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigSet.java
@@ -29,13 +29,16 @@ public class ConfigSet extends CamelCommand {
     @CommandLine.Parameters(description = "Configuration parameter (ex. key=value)", arity = "1")
     String configuration;
 
+    @CommandLine.Option(names = { "--local" }, description = "Retrieve configurations from current directory")
+    boolean local;
+
     public ConfigSet(CamelJBangMain main) {
         super(main);
     }
 
     @Override
     public Integer doCall() throws Exception {
-        CommandLineHelper.createPropertyFile();
+        CommandLineHelper.createPropertyFile(local);
 
         if (configuration.split("=").length == 1) {
             printer().println("Configuration parameter not in key=value format");
@@ -46,8 +49,8 @@ public class ConfigSet extends CamelCommand {
             String key = StringHelper.before(configuration, "=").trim();
             String value = StringHelper.after(configuration, "=").trim();
             properties.put(key, value);
-            CommandLineHelper.storeProperties(properties, printer());
-        });
+            CommandLineHelper.storeProperties(properties, printer(), local);
+        }, local);
 
         return 0;
     }
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigSet.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigSet.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.dsl.jbang.core.commands.config;
 
+import java.io.IOException;
+
 import org.apache.camel.dsl.jbang.core.commands.CamelCommand;
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
 import org.apache.camel.dsl.jbang.core.common.CommandLineHelper;
@@ -29,8 +31,8 @@ public class ConfigSet extends CamelCommand {
     @CommandLine.Parameters(description = "Configuration parameter (ex. key=value)", arity = "1")
     String configuration;
 
-    @CommandLine.Option(names = { "--local" }, description = "Retrieve configurations from current directory")
-    boolean local;
+    @CommandLine.Option(names = { "--global" }, description = "Use global or local configuration")
+    boolean global = true;
 
     public ConfigSet(CamelJBangMain main) {
         super(main);
@@ -38,6 +40,10 @@ public class ConfigSet extends CamelCommand {
 
     @Override
     public Integer doCall() throws Exception {
+        return setConfiguration(!global);
+    }
+
+    private int setConfiguration(boolean local) throws IOException {
         CommandLineHelper.createPropertyFile(local);
 
         if (configuration.split("=").length == 1) {
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigUnset.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigUnset.java
@@ -28,8 +28,8 @@ public class ConfigUnset extends CamelCommand {
     @CommandLine.Parameters(description = "Configuration key", arity = "1")
     String key;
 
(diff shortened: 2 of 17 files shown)


oscerd (approved):

davsclaus:
I wonder if we need any new options at all. At first glance it seems more complicated and when would you use these options?

I would assume that if there is a local file then its always used and take precedence over the default configuration. That's it.

Are there other CLi tools that has these kind of options and are they also named similar ?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigSet.java
```diff
@@ -29,13 +29,16 @@ public class ConfigSet extends CamelCommand {
     @CommandLine.Parameters(description = "Configuration parameter (ex. key=value)", arity = "1")
     String configuration;
 
+    @CommandLine.Option(names = { "--local" }, description = "Retrieve configurations from current directory")
```
tadayosi, line 32: The wording should be "Store ... to" instead of "Retrieve ... from" here.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/config/ConfigUnset.java
```diff
@@ -28,6 +28,9 @@ public class ConfigUnset extends CamelCommand {
     @CommandLine.Parameters(description = "Configuration key", arity = "1")
     String key;
 
+    @CommandLine.Option(names = { "--local" }, description = "Retrieve configurations from current directory")
```
tadayosi, line 31: "Unset" instead of "Retrieve"

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/version/VersionSet.java
```diff
@@ -43,13 +43,16 @@ public class VersionSet extends CamelCommand {
     @CommandLine.Option(names = { "--reset" }, description = "Reset by removing any custom version settings")
     boolean reset;
 
+    @CommandLine.Option(names = { "--local" }, description = "Retrieve configurations from current directory")
```
tadayosi, line 46: "Store ... to" over "Retrieve ... from"

tadayosi: Or if we want the same wording for all options, we should consider a more generic wording. Examples:
"Apply configrations to current directoyr", or
"Use local configurations instead of global"?

tadayosi (commented):
Thanks, looks great besides the discussion about the usage of `--mergeConfigurations`. My reply on the topic follows.

tadayosi:
@davsclaus @Croway 
> I wonder if we need any new options at all. At first glance it seems more complicated and when would you use these options?

If we start heavily using `camel config set/get/list` commands, we'll soon want this feature. For example, some camel script might require additional deps such as `dependency=dev.langchain4j:langchain4j-ollama`, but I'm sure we don't want to apply it to every route. The same should apply to, say, `max-idle-seconds=1`, and `logging-category=org.apache.camel=WARN`. Those are not something that should be stored globally, but at the same time I'm sure we don't want to repeat all of these for every run for some specific routes. That's when this new feature works out. You can create a specific working directory, set up a local config, and just run `camel run <script>`.

> I would assume that if there is a local file then its always used and take precedence over the default configuration. That's it.

For `--mergeConfigurations`, I agree with Claus. it's not critical that we merge local and global configs with this use case. It's not a big deal, we can always use local one or global one. If we want to enable global again, simply `rm .camel-jbang-user.properties`.

On the other hand, if we accept the new feature, `--local` should be still required, as otherwise there's no authentic way to store those configs locally. It's not cool that we must first define them with `camel config set ...` and then manually `cp ~/.camel-jbang-user.properties .`.

> Are there other CLi tools that has these kind of options and are they also named similar ?

direnv uses the `source_up` instruction to refer to the parent config. Similar function but naming is different.

davsclaus:
If the goal is to have per project "camel jbang" configuration file then I think it should be

- not a hidden file, so its should be named something ala `camel-jbang.properties` or `camel-jbang-local.properties` or `camel-jbang-user.properties` or some other better name. 

- the CLI tool able to modify both local and global options with a `--global` flag etc. 

- make the local option take precedence (if exists) and fallback to global option (if exist)

- make the CLI tool able to show a break-down of the options so you can see whether a option is local or global

tadayosi:
@davsclaus @Croway Thanks Claus. Now I agree with all of these.

> * not a hidden file, so its should be named something ala `camel-jbang.properties` or `camel-jbang-local.properties` or `camel-jbang-user.properties` or some other better name.

Looks like `camel-jbang-user.properties` is most straightforward.

> * the CLI tool able to modify both local and global options with a `--global` flag etc.

So, `--global` is `true` by default (for backward compatibility) and use `--global=false` to save configs to local, correct?  I like it.

> * make the local option take precedence (if exists) and fallback to global option (if exist)

Initially I thought it not taking any precedence but I can accept it. All we need will be just carefully choose which options should be stored globally.

> * make the CLI tool able to show a break-down of the options so you can see whether a option is local or global

That's actually very cool feature. Indeed direnv has `direnv status` option that shows which upstream `.envrc` it refers.

Croway:
Thanks for the suggestions @tadayosi @davsclaus , I implemented what requested, this is a demo:
```
➜  camel init test.yaml
➜  ll
total 8
-rw-r--r--@ 1 fmariani  wheel   185B 31 Mar 11:11 test.yaml
➜  camel config list
----- Global -----
runtime = quarkus
➜ camel config set runtime=spring-boot --global=false
➜ camel config list
----- Local -----
runtime = spring-boot
----- Global -----
runtime = quarkus
➜ l
total 16
drwxr-xr-x@  5 fmariani  wheel   160B 31 Mar 11:12 .
drwxrwxrwt  10 root      wheel   320B 31 Mar 11:11 ..
drwxr-xr-x@  2 fmariani  wheel    64B 31 Mar 11:11 .camel-jbang
-rw-r--r--@  1 fmariani  wheel    51B 31 Mar 11:12 camel-jbang-user.properties
-rw-r--r--@  1 fmariani  wheel   185B 31 Mar 11:11 test.yaml
➜ camel export test.yaml --gav=test:test:1
Exporting integration with the following configuration:
    --runtime=springBoot
Generating fresh run data
Exporting as Spring Boot project to: .
➜ rm -rf camel-jbang-user.properties
➜ camel export test.yaml --gav=test:test:1
Exporting integration with the following configuration:
    --runtime=quarkus
Generating fresh run data
Exporting as Quarkus project to: .
```

I modified the `config list` command to print both local and global configurations (if exists), by default the local configuration is used if existing.

davsclaus (approved):

davsclaus:
Also if not done then add some docs to camel-jbang.adoc file (yeah its a long file) but its good to have some docs on the cool features we have. And then we may break up this big file into seperate sub docs in the future.

Croway:
@davsclaus I added some documentation, to do so, I edited the Configuration section, is it ok? or the jbang doc file is "append only"?

davsclaus:
its okay thanks

PR #17645, opened 2025-04-02
CAMEL-21926 camel-attachment fix inconsistent behaviour when last attachment is removed

anirudh-04:
…

# Description
Updated the remove attachment logic: remove ATTACHMENTS trait when last attachment is removed.

# Target

# Tracking

### DIFF
--- a/components/camel-attachments/src/main/java/org/apache/camel/attachment/DefaultAttachmentMessage.java
+++ b/components/camel-attachments/src/main/java/org/apache/camel/attachment/DefaultAttachmentMessage.java
@@ -212,7 +212,11 @@ public final class DefaultAttachmentMessage implements AttachmentMessage {
 
     @Override
     public void removeAttachment(String id) {
-        getAttachmentsMap().remove(id);
+        Map<String, Object> attachmentsMap = getAttachmentsMap();
+        attachmentsMap.remove(id);
+        if (attachmentsMap.isEmpty()) {
+            removeTrait(MessageTrait.ATTACHMENTS);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


components/camel-attachments/src/main/java/org/apache/camel/attachment/DefaultAttachmentMessage.java
```diff
@@ -212,7 +212,9 @@ public Set<String> getAttachmentNames() {
 
     @Override
     public void removeAttachment(String id) {
-        getAttachmentsMap().remove(id);
+        Map<String, Object> attachmentsMap = getAttachmentsMap();
+        if (attachmentsMap.remove(id) != null && attachmentsMap.isEmpty())
```
turing85, line 216: Why check for `attachmentsMap.remove(id) != null`? The check for `attachmentsMap.isEmpty()` should suffice and I am not sure how much performance we actually safe by the `null`-check.

anirudh-04: Understood, I had added the check to ensure an actual removal occurred. If it makes the code less readable I can simplify the logic.

turing85: From my point of view, this are two different requirements (removing the requested attachment, and checking whether the attachments are empty).  I tend on the side of clarity, and having `attachmentsMap.remove(id)` separated from the condition seems clearer to me. But I am not a maintainer; let's see what @davsclaus and/or @jamesnetherton say.

jamesnetherton: >  I tend on the side of clarity, and having attachmentsMap.remove(id) separated from the condition seems clearer to me

I agree. Please also add braces (`{` `}`) to the if statement.

oscerd (approved):

jamesnetherton (approved):

oscerd:
@turing85 can you review again?

turing85 (approved):

oscerd:
Can you create a PR to backport to 4.10.x?

anirudh-04:
> Can you create a PR to backport to 4.10.x?

https://github.com/apache/camel/pull/17657

PR #17655, opened 2025-04-03
CAMEL-21736 google-bigquery-sql: Support fetching query results for S…

anirudh-04:
…ELECT statements

# Description
Opening this as a draft to check if the approach is correct. The update allows SELECT queries to return actual results instead of just the row count. Let me know if this aligns with expectations or if any changes are needed. Thanks!

# Target

# Tracking


oscerd:
@lsergio FYI

anirudh-04:
> Looks good. In the issue there was a user aiming to work on this, next time please first ask if a contributor is working on that or not, to avoid waste of time and work of others. Thanks.

Apologies, I thought sufficient time had passed. Would ask from next time 🙌

davsclaus (approved):

davsclaus:
Is there any kind of limit on the number of returned rows, what can happen if there is a lot of rows then we build up an im-memory list of all the rows.

anirudh-04:
@davsclaus it does seem like the maximum response size can be quite large - 10GB. There are ways to paginate the response which could help avoid OOME. Please suggest an approach here. 

Limit documentation - https://cloud.google.com/bigquery/quotas

oscerd:
I think we should return a paginated result, or at least we need to verify the number of rows returned and eventually choose to paginate.

anirudh-04:
@oscerd to confirm, by paginated result - do you mean that the first response should return only a single page of results along with a nextPageToken (in Header) that clients can use in subsequent requests to fetch additional pages?

oscerd:
Yes, so the end user can iterate.

anirudh-04:
Should I be committing the auto-generated google-bigquery-sql.json, GoogleBigQuerySQLEndpointUriFactory.java, GoogleBigQuerySQLEndpointConfigurer.java changes as well?

oscerd (commented):
Yes, you should. Thanks.

PR #17694, opened 2025-04-08
CAMEL-21937: stop routing messages if one of the tools invoked stop

orpiske:


oscerd (approved):

orpiske:
This needs further discussion, so I'm not going to merge it just yet (and it may be open for some days as my PTO is approaching quickly).

PR #17731, opened 2025-04-11
CAMEL-21947: camel-file/camel-ftp: If read lock is not acquired then …

davsclaus:
…file is not removed from idempotent repo if eager=true. WIP patch.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpReadLockNotStartedIT.java
+++ b/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpReadLockNotStartedIT.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


oscerd (approved):

components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpReadLockNotStartedIT.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @TempDir
+    Path testDirectory;
+
+    @BindToRegistry("myLock")
+    private final GenericFileProcessStrategy lock = new MyReadLock();
+
+    protected String getFtpUrl() {
+        return "ftp://admin@localhost:{{ftp.server.port}}"
+               + "/notstarted?password=admin&processStrategy=#myLock&idempotent=true";
```
lantzen, line 49: Should this not also be "&idempotentEager=true"?

davsclaus: eager is true by default

components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/integration/FtpReadLockNotStartedIT.java
lantzen, line 1: This unit-test was elegant and take a lot less time to execute than the on in my PR, with a lot of take aways :)

davsclaus: Yes but testing with the real FTP server is better but can be hard - this test do indeed end up calling started = false, and trigger that part of the code you identified where the bug was. So both the unit test and also debugging this unit test ensured that this code part was called. And the repo is also checked to ensure the file is deleted from the repo so it can be picked up again later.

PR #17734, opened 2025-04-11
CAMEL-21845: camel-sql - Only commit/rollback if auto commit has been true

graben:
Add-on to the optimization from CAMEL-21845. Commit or rollback should only be called if auto commit has not been set to false before. This might be intended to control the transaction for whatever reason! In cases of XA transactions, for example, calling commit or rollback would cause exceptions.

### DIFF
--- a/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
@@ -159,7 +159,7 @@ public class SqlProducer extends DefaultProducer {
                     boolean restoreAutoCommit = true;
 
                     if (batch) {
-                        if (manualCommit) {
+                        if (!exchange.isTransacted() && manualCommit) {
                             // optimize batch by turning off auto-commit
                             restoreAutoCommit = ps.getConnection().getAutoCommit();
                             ps.getConnection().setAutoCommit(false);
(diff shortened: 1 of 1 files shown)


oscerd (approved):

zhfeng (approved):

components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
```diff
@@ -171,12 +171,12 @@ public Object doInPreparedStatement(PreparedStatement ps) throws SQLException {
                                 total += count;
                             }
                             exchange.getIn().setHeader(SqlConstants.SQL_UPDATE_COUNT, total);
-                            if (manualCommit) {
+                            if (manualCommit && !restoreAutoCommit) {
```
zhfeng, line 174: I wonder if there is any better way to check if we are running in a JTA transaction?

zhfeng: IIRC
>Commit or rollback should only be called if auto commit has not been set to false before.

So it should check with `if (manualCommit && restoreAutoCommit)` ?

graben: Ups, seems your right. I'll change it.

graben: > I wonder if there is any better way to check if we are running in a JTA transaction?

You're right that there are better ways to identify JTA, but it's only one option why a connection has been set to auto commit before calling sql producer.

zhfeng: @davsclaus can we use `exchange.IsTransacted()` ?

zhfeng: I think it should be `true` if `sql` is using in a `transacted()` like
```java
from("direct:foo")
    .transacted()
    .to("sql:INSERT INTO ...");
```

davsclaus: Yes if using JTA/XA then you need to use a TX manager and setup this via spring or camel-jta, and this requires enabling the transacted DSL in the routes, so we then know `exchange.isTransacted`. 

So if this is the case then this component can do something different.

graben: Well, the question is if only JTA is a valid case that autocommit has been set to false before SqlProducer is called. If yes exchange.isTransacted fits better.

davsclaus:
I wonder if exchange.isTransacted then it should do as before CAMEL-21845 as when you use JTA/XA et all then the TX manager handles all the commit/rollback and you should not change auto commit and other settings.

And then this optimization is only for normal use-cases

graben:
No, because the commit is called from SqlProducer. I ran into it already.

zhfeng (approved):

davsclaus (approved):

PR #17735, opened 2025-04-12
CAMEL-21871: camel-azure-serviebus - TokenCredential is autowired, an…

davsclaus:
…d ensure if explicit configured on component then that is used.

# Description

# Target

# Tracking


oscerd (approved):

oscerd:
I think there are others components doing that in azure space

davsclaus:
> I think there are others components doing that in azure space

Ah yeah it seems they all should ideally be made to use autowired - its a bit more work to do them all

anirudh-04:
@davsclaus wouldn't createServiceBusSenderClient invocation give npe because of the switch case - 

```
        switch (configuration.getCredentialType()) {
            case CONNECTION_STRING -> builder.connectionString(configuration.getConnectionString());
            case TOKEN_CREDENTIAL -> builder.credential(fullyQualifiedNamespace, credential);
            case AZURE_IDENTITY -> builder.credential(fullyQualifiedNamespace, new DefaultAzureCredentialBuilder().build());
        }

```

Test to replicate, do we need to handle this?

```
    @Test
    void testCreateBaseServiceBusClientWithNoCredentialType() throws Exception {
        ServiceBusConfiguration configuration = new ServiceBusConfiguration();
        ServiceBusClientFactory factory = new ServiceBusClientFactory();
        final ServiceBusSenderClient senderClient = factory.createServiceBusSenderClient(configuration);

        assertNotNull(senderClient);
    }
```

davsclaus:
the component sets the credential type on the configuration so it works - your unit test is wrong as it does not use the camel component but creating the objects manually.

davsclaus:
You can make the switch default to be connection string if you want

            default -> builder.connectionString(configuration.getConnectionString());

davsclaus:
okay I added so that if its null then its using connection string as default (as before) and added your unit test. Thanks

PR #17757, opened 2025-04-15
Camel-Console is not needed in production, make it provided

Croway:
I think that the best practice here, would be that users create a `dev` maven profile with camel-console dependency.

I do not think we need documentation here since in camel-console documentation there is already
```
== Using Camel Console

The `camel-console` must be added to the classpath, and enabled either via...
...
```

But if this PR makes sense, and is merged, I'll open a jira issue for the release notes

### DIFF
--- a/components/camel-kafka/pom.xml
+++ b/components/camel-kafka/pom.xml
@@ -45,6 +45,7 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-console</artifactId>
+            <scope>provided</scope>
         </dependency>
 
         <!-- kafka java client -->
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus (changes requested):
the console must be included in camel-kamelet-main as its needed by the camel-jbang CLI itself

davsclaus:
> The `camel-console` must be added to the classpath, and enabled either via...

This is only if you use CSB / CEQ or run camel yourself and can include the console or not.

for jbang it must be included

davsclaus (approved):

PR #17780, opened 2025-04-16
Close stream resources

Croway:
Close potential streams that are not closed.

All the core tests are successful

### DIFF
--- a/core/camel-main/src/main/java/org/apache/camel/main/MainHelper.java
+++ b/core/camel-main/src/main/java/org/apache/camel/main/MainHelper.java
@@ -497,22 +497,16 @@ public final class MainHelper {
     private String doGetVersion() {
         String version = null;
 
-        InputStream is = null;
         // try to load from maven properties first
-        try {
-            Properties p = new Properties();
-            is = MainHelper.class
-                    .getResourceAsStream("/META-INF/maven/org.apache.camel/camel-main/pom.properties");
+        try (InputStream is = MainHelper.class
+                .getResourceAsStream("/META-INF/maven/org.apache.camel/camel-main/pom.properties")) {
             if (is != null) {
+                Properties p = new Properties();
                 p.load(is);
                 version = p.getProperty("version", "");
             }
         } catch (Exception e) {
-            // ignore
-        } finally {
-            if (is != null) {
-                IOHelper.close(is);
-            }
+            LOG.warn("Error loading version information from pom.properties: {}", e.getMessage(), e);
         }
 
         // fallback to using Java API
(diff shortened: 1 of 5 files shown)


core/camel-main/src/main/java/org/apache/camel/main/MainHelper.java
```diff
@@ -497,22 +497,16 @@ private static void loadLines(InputStream in, Set<String> lines, Function<String
...
                 p.load(is);
                 version = p.getProperty("version", "");
             }
         } catch (Exception e) {
-            // ignore
-        } finally {
-            if (is != null) {
-                IOHelper.close(is);
-            }
+            LOG.warn("Error loading version information from pom.properties: {}", e.getMessage(), e);
```
davsclaus, line 509: The old code ignores this so make it the same

core/camel-support/src/main/java/org/apache/camel/converter/stream/StreamCacheConverter.java
```diff
@@ -50,9 +50,11 @@ public static StreamCache convertToStreamCache(ByteArrayInputStream stream, Exch
     public static StreamCache convertToStreamCache(InputStream stream, Exchange exchange) throws IOException {
         // transfer the input stream to a cached output stream, and then creates a new stream cache view
         // of the data, which ensures the input stream is cached and re-readable.
-        CachedOutputStream cos = new CachedOutputStream(exchange);
-        IOHelper.copyAndCloseInput(stream, cos);
-        return cos.newStreamCache();
+        // Use try-with-resources to ensure the CachedOutputStream is closed automatically
+        try (CachedOutputStream cos = new CachedOutputStream(exchange)) {
```
davsclaus, line 54: This is wrong as COS should not be closed - there is some special code for stream cahcing when its based on file disk and whatnot.

core/camel-support/src/main/java/org/apache/camel/support/InputStreamIterator.java
```diff
@@ -65,11 +67,18 @@ public int available() throws IOException {
...
             try {
-                byte[] buf = converter.mandatoryConvertTo(byte[].class, it.next());
+                next = it.next();
+                byte[] buf = converter.mandatoryConvertTo(byte[].class, next);
                 return new ByteArrayInputStream(buf);
             } catch (NoTypeConversionAvailableException e) {
                 throw new IOException(e);
+            } finally {
+                // ensure the original object is closed if it was closeable
+                if (next != null && next instanceof Closeable closeableNext) {
```
davsclaus, line 79: This is also wrong as it.next can be part of some object that should only be closed when its fully done.

davsclaus (changes requested):
Sorry but this is risky to just blindly close everything some report tool spit out.

PR #17829, opened 2025-04-22
Removal of duplicate conditional branches in docs

rnetuka:
IDENTICAL_BRANCHES (CWE-398)

The same code is executed regardless of whether encoding === "utf8" is true, because the "then" and "else" branches are identical.


davsclaus (changes requested):
this is an auto generated file as part of some node stuff - do not make changes to it

PR #17833, opened 2025-04-22
Resource leak in PackageHelper

rnetuka:
RESOURCE_LEAK (CWE-404)

Input stream is not closed anywhere. @davsclaus is this solution ok, or do you prefer closing with _IOHelper_?


davsclaus:
no when you can use java try stuff then that is okay

PR #17916, opened 2025-04-29
CAMEL-22028: camel-jbang - Regression with missing routes file

tdiesler:
https://issues.apache.org/jira/browse/CAMEL-22028


tdiesler:
I was wondering about http, https - let me add those. Any other schemes?
Also, opted to ignore classpath verification - is that ok for now?

davsclaus:
yeah the export does not download these files but when you package and run the app via java -jar then the files are downloaded from github.

So I just wanted to point out that this code does not fail when you use these http/github - so maybe the code should skip these. 

I am not aware of others than http/https/github out of the box

Custom resolvers can be made so you could have a 'clippy' and it did something special.

So maybe this check is for "file" and "classpath" only instead. And anything else is ignored.

PR #18042, opened 2025-05-13
Cleanup based on static analysis

dk2k:


davsclaus:
the sql is actually wrong as we forgot to implement the service interface, I am fixing this. Then the other bits are okay. Maybe you can update this PR and remove the sql code

dk2k:
@davsclaus I reverted changes for SQL-related class

components/camel-aws/camel-aws-secrets-manager/src/main/java/org/apache/camel/component/aws/secretsmanager/vault/CloudTrailReloadTriggerTask.java
```diff
@@ -349,7 +349,6 @@ public void run() {
                                     if (isReloadEnabled()) {
                                         LOG.info("Update for AWS secret: {} detected, triggering CamelContext reload", name);
                                         triggerReloading = true;
-                                        message.receiptHandle();
```
davsclaus, line 352: @oscerd just wonder if this method did some that was needed or it can be safely removed

davsclaus (approved):

components/camel-aws/camel-aws-secrets-manager/src/main/java/org/apache/camel/component/aws/secretsmanager/vault/CloudTrailReloadTriggerTask.java
davsclaus: ah its a noop

PR #18131, opened 2025-05-20
CAMEL-22087: Make name and port camel CLI option working on CSB and CEQ

smongiar:
# Description

This should fix https://issues.apache.org/jira/browse/CAMEL-22087

`--verbose` option should be fixed too

# Target

- [main]


davsclaus:
This looks like it can be backported to 4.10.x. You are welcome to create a PR for that branch

davsclaus:
I was able to backport cleanly to 4.10.x this PR

davsclaus:
I added a comment in the ticket - does it work with --name ?

smongiar:
Sorry for the late response, so I tried with  _--name_ and _--port_ option. It seems working good.
I even tried with configuration properties (solved in camel-jbang-run.properties files) with

`camel config set name=ciao`
`camel config port=8083`

and it's working too

PR #18146, opened 2025-05-21
CAMEL-22096: upgrades the tokenizer to use LangChain4j 1.0.0

orpiske:


orpiske:
This is just a draft for now as I'd like to wait for @zbendhiba's initial work to upgrade LangChain4j to 1.0.0.  This should make the same for our tokenizer code.

oscerd (approved):

orpiske:
Things to note: I have defined the langchain4j versions as properties in the component's pom file, as it may be the case that we will need to live with different versions for different components for a while. 

Please let me know in case you all would like a different approach for keeping these versions elsewhere.

PR #18156, opened 2025-05-22
Update docs to include updated trace property

joshiraezcode:
# Description

Related: https://github.com/hawtio/hawtio/pull/3917

Updates the trace properties in the examples to use current 4.5 property camel.trace.enabled

https://camel.apache.org/manual/camel-4x-upgrade-guide-4_5.html#_camel_spring_boot

# Target

# Tracking

No related Jira issue

No related Jira issue

Doesn't apply (I think?)

### DIFF
--- a/docs/user-manual/modules/ROOT/pages/tracer.adoc
+++ b/docs/user-manual/modules/ROOT/pages/tracer.adoc
@@ -27,7 +27,7 @@ And in Spring Boot
 
 [source,text]
 ----
-camel.trace.enabled = true
+camel.main.tracing = true
 ----
 
 === Setting Tracing in Standby mode
@@ -55,7 +55,7 @@ And in Spring Boot
 
 [source,text]
 ----
-camel.springboot.tracing-standby = true
+camel.main.tracing-standby = true
 ----
 
 If tracer is in standby mode, then tracing is made available, and can be enabled during runtime.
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

docs/user-manual/modules/ROOT/pages/tracer.adoc
```diff
@@ -27,7 +27,7 @@ And in Spring Boot
 
 [source,text]
 ----
-camel.springboot.tracing = true
+camel.trace.enabled = true
```
davsclaus, line 30: This is the old tracer that prints to log, and it should be `camel.main.tracing = true`

davsclaus (changes requested):
oh sorry spotted that the last one should be different

docs/user-manual/modules/ROOT/pages/tracer.adoc
joshiraezcode: Yes, just changed it after chatting with you

joshiraezcode:
I updated the changes to use `camel.main.` migrating from `camel.springboot.` for the tracer properties I have seen in the doc. I think it should be correct now that I understood the difference between backlog and default tracer

davsclaus (approved):

PR #18212, opened 2025-05-29
Feature/camel 16483

vineetsaurabh:
# Description

# Target

# Tracking


components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -90,6 +90,12 @@ public class NatsConfiguration {
     private boolean traceConnection;
     @UriParam(label = "advanced")
     private HeaderFilterStrategy headerFilterStrategy = new DefaultHeaderFilterStrategy();
+    @UriParam(label = "advanced", defaultValue = "true")
+    private boolean jetstreamEnabled = true;
```
oscerd, line 94: I don't think the default value should be true. It's not sure people will use jetstream by default.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -90,6 +90,12 @@ public class NatsConfiguration {
...
+    @UriParam(label = "advanced", defaultValue = "default-stream")
+    private String jetstreamName = "default-stream";
```
oscerd, line 96: Is this the default name of the stream or something you're assuming?

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -454,4 +460,55 @@ public boolean isTraceConnection() {
...
+     * Whether to enable JetStream support for this endpoint.
+     */
+    public boolean isJetstreamEnabled() {
+        return jetstreamEnabled;
+    }
+
+    /**
+     * Sets whether to enable JetStream support for this endpoint.
+     *
+     * @param jetstreamEnabled {@code true} to enable JetStream, {@code false} otherwise.
```
oscerd, line 474: Please remove the param, is not needed. We are going to generate documentation from the javadoc, this would mess the documentation page.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -454,4 +460,55 @@ public boolean isTraceConnection() {
...
+     * will be used if JetStream is enabled.
+     */
+    public String getJetstreamName() {
+        return jetstreamName;
+    }
+
+    /**
+     * Sets the name of the JetStream stream to use.
+     *
+     * @param jetstreamName The name of the JetStream stream.
```
oscerd, line 491: Please remove the param, is not needed. We are going to generate documentation from the javadoc, this would mess the documentation page.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -16,15 +16,15 @@
...
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
-import io.nats.client.Connection;
+import io.nats.client.*;
```
oscerd, line 24: Please avoid * import

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -16,15 +16,15 @@
  */
 package org.apache.camel.component.nats;
 
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
oscerd, line 22: Please avoid * import

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -108,16 +108,53 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
...
-            final NatsMessage.Builder builder = NatsMessage.builder()
-                    .data(body)
-                    .subject(config.getTopic())
-                    .headers(this.buildHeaders(exchange));
-
-            if (ObjectHelper.isNotEmpty(config.getReplySubject())) {
-                final String replySubject = config.getReplySubject();
-                builder.replyTo(replySubject);
+            if (config.isJetstreamEnabled() && this.connection.getServerInfo().isJetStreamAvailable()) {
+                LOG.info("JetStream is available");
```
oscerd, line 112: Use DEBUG.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -108,16 +108,53 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
...
+                    LOG.info("Publish Sequence async: {}", pa.getSeqno());
+                } else {
+                    try {
+                        pa = js.publish(config.getTopic(), body);
+                    } catch (IOException | JetStreamApiException e) {
+                        throw new RuntimeException(e);
+                    }
+                    LOG.info("Publish Sequence synchronously: {}", pa.getSeqno());
+                }
+            } else {
```
oscerd, line 146: I would do the other way around. Jet stream will be not enabled by default, so its part should be in the else.

oscerd (changes requested):
Thanks for the PR.

There are some findings. Also what happens with the consumer?

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: Totally agree!
I shall change it.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: It is not the default name. Rather than leaving it blank, I used a default name.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: OK.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
vineetsaurabh: Ok.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
vineetsaurabh: Yes, it makes sense when changing default.

vineetsaurabh:
Code changes for Consumer is done.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
davsclaus: Break it up into 2 methods, so one is jetstream and the other is non jetstream and as-is today

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java
```diff
@@ -123,26 +130,70 @@ class NatsConsumingTask implements Runnable {
...
-                                Integer.parseInt(NatsConsumer.this.getEndpoint().getConfiguration().getMaxMessages()));
+                    LOG.debug("Setting up standard NATS consumer for topic: {}",
+                            NatsConsumer.this.getEndpoint().getConfiguration().getTopic());
+                    NatsConsumer.this.dispatcher = connection.createDispatcher(new CamelNatsMessageHandler());
+                    if (ObjectHelper.isNotEmpty(queueName)) {
+                        NatsConsumer.this.dispatcher = NatsConsumer.this.dispatcher.subscribe(topic, queueName);
+                    } else {
+                        NatsConsumer.this.dispatcher = NatsConsumer.this.dispatcher.subscribe(topic);
+                    }
+                    if (maxMessages != null) {
```
davsclaus: What is this code doing with maxMessage and unsubscribe ?

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -193,4 +195,65 @@ protected void doStop() throws Exception {
...
+            final String replySubject = config.getReplySubject();
+            builder.replyTo(replySubject);
+        }
+        final Message jetStreamMessage = builder.build();
+
+        PublishAck pa;
+        if (config.isJetstreamAsync()) {
+            CompletableFuture<PublishAck> future = js.publishAsync(jetStreamMessage);
+            try {
+                pa = future.get(1, TimeUnit.SECONDS);
```
davsclaus: Why is the timeout so low at only 1 second

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -193,4 +195,65 @@ protected void doStop() throws Exception {
...
+            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                throw new RuntimeException("Failed to publish message asynchronously with JetStream: " + e.getMessage(), e);
+            }
+            LOG.info("Publish Sequence async: {}", pa.getSeqno());
```
davsclaus: Do not use INFO logging as it spams logs when sending many messages

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -193,4 +195,65 @@ protected void doStop() throws Exception {
...
+        } else {
+            try {
+                pa = js.publish(jetStreamMessage);
+            } catch (IOException | JetStreamApiException e) {
+                throw new RuntimeException("Failed to publish message synchronously with JetStream: " + e.getMessage(), e);
+            }
+            LOG.info("Publish Sequence synchronously: {}", pa.getSeqno());
```
davsclaus: Same here

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java
```diff
@@ -123,26 +130,70 @@ class NatsConsumingTask implements Runnable {
...
-                            NatsConsumer.this.getEndpoint().getConfiguration().getTopic(),
-                            NatsConsumer.this.getEndpoint().getConfiguration().getQueueName());
-                    if (ObjectHelper.isNotEmpty(NatsConsumer.this.getEndpoint().getConfiguration().getMaxMessages())) {
-                        NatsConsumer.this.dispatcher.unsubscribe(
-                                NatsConsumer.this.getEndpoint().getConfiguration().getTopic(),
-                                Integer.parseInt(NatsConsumer.this.getEndpoint().getConfiguration().getMaxMessages()));
-                    }
-                    if (NatsConsumer.this.dispatcher.isActive()) {
-                        NatsConsumer.this.setActive(true);
+                NatsConfiguration config = getEndpoint().getConfiguration();
```
davsclaus, line 134: Can you split up the run method into 2 so we have 1 with jetstream and another with the previous code as-is

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
davsclaus: What is the meaning of the jetstream-name, can you tell more about that

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -90,6 +90,12 @@ public class NatsConfiguration {
     private boolean traceConnection;
     @UriParam(label = "advanced")
     private HeaderFilterStrategy headerFilterStrategy = new DefaultHeaderFilterStrategy();
+    @UriParam(label = "advanced", defaultValue = "false")
+    private boolean jetstreamEnabled = false;
```
davsclaus, line 94: I dont think this option should be advanced as jestream support has been asked for for many years

davsclaus (changes requested):

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: _jetstreamName_ is the name of the stream, identifying the stream and has to be unique within JetStream account.

davsclaus: Is it not better to make the end user define the name instead of using `default-stream` which would otherwise be confusing if they dont know why something fails or happens or what stream is actually used by camel with their account.

vineetsaurabh: The end user can define the name of the stream. In case, the end user does not define the name of the stream, the default name will be taken. 

Example:
```
.to("nats:" + topic +
                "?jetstreamEnabled=" + streamSettings.isNatsJetstreamEnabled() +
                "&jetstreamName=" + streamSettings.getNatsJetstreamName() +
                "&jetstreamAsync=" + streamSettings.isNatsJetstreamAsyncEnabled() +
                "&secure=true")
        .log("(Device) Message sent to NATS server");
```


What is your suggestion?

davsclaus: That they must always define the name to use (no default)

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java
vineetsaurabh: This is an existing code. Calling _unsubscribe_ on a Core NATS client instructs the server to stop sending messages to that specific subscription after _maxMessages_ have been delivered, effectively dropping any subsequent messages for that client.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: Ok. I shall change the code accordingly.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
vineetsaurabh: Changing it to `future.get(config.getRequestTimeout(), TimeUnit.MILLISECONDS);`, which has a default value 20000L

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java
davsclaus: Ah okay its a special feature in camel-nats only - a bit odd as its not something common in the other messaging components

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
davsclaus: It may be good to be able to configure that timeout value, and then leave it as a sensible default such as 20 seconds

vineetsaurabh: Sorry, it is not clear to me - are you suggesting a new timeout value? Or, shall I proceed with the code mentioned above?

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
```diff
@@ -90,6 +90,12 @@ public class NatsConfiguration {
     private boolean traceConnection;
     @UriParam(label = "advanced")
     private HeaderFilterStrategy headerFilterStrategy = new DefaultHeaderFilterStrategy();
+    @UriParam(label = "common", defaultValue = "false")
+    private boolean jetstreamEnabled = false;
+    @UriParam(label = "common")
+    private String jetstreamName;
+    @UriParam(label = "common", defaultValue = "true")
```
davsclaus: This is likely an advanced option, when would you set this to false ?

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
```diff
@@ -104,18 +112,12 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
...
-
-            final NatsMessage.Builder builder = NatsMessage.builder()
-                    .data(body)
-                    .subject(config.getTopic())
-                    .headers(this.buildHeaders(exchange));
-
-            if (ObjectHelper.isNotEmpty(config.getReplySubject())) {
-                final String replySubject = config.getReplySubject();
-                builder.replyTo(replySubject);
+            LOG.debug("Publishing to subject: {}", config.getTopic());
```
davsclaus, line 115: This code in this else block needs to be in a try .. catch (exception) exchange.setExchange(e) callback done(true) idiom that are used with camel routing engine.

davsclaus (commented):
Almost there a few comments

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConfiguration.java
vineetsaurabh: Ok. I will make it advance. 

We should set it false when we need to be sure the message was safely saved before your program does anything else. Also if there is lower throughput, we may not need to process async.

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsProducer.java
vineetsaurabh: Sorry, I do not understand your comment. 

`else` block is the existing code. It was not inside `try .. catch (exception)`.

davsclaus: Something ala
```
            try {
                // do not copy as we already did the copy
                addToQueue(copy, false);
            } catch (SedaConsumerNotAvailableException | IOException e) {
                exchange.setException(e);
                callback.done(true);
                return true;
            }
```

this is from camel-seda but the catch should just be general Exception

davsclaus:
/component-test camel-nats

components/camel-nats/src/main/java/org/apache/camel/component/nats/NatsConsumer.java
```diff
@@ -123,37 +131,88 @@ class NatsConsumingTask implements Runnable {
...
+        private void setupJetStreamConsumer(String topic, String queueName) throws IOException, JetStreamApiException {
+            String streamName = this.configuration.getJetstreamName();
+            String consumerName
+                    = ObjectHelper.isNotEmpty(queueName) ? queueName : "consumer-" + System.currentTimeMillis(); // Generate a default consumer name if queueName is not provided
+            LOG.debug("Setting up JetStream PUSH consumer for stream: '{}', durable: '{}', topic: {} ", streamName,
+                    consumerName, this.configuration.getTopic());
+
+            JetStreamManagement jsm = connection.jetStreamManagement();
+            try {
+                // Try to get the stream, create it if it doesn't exist
```
bvahdat, line 163: Sorry for my late comment on something being already merged, but instead of trying to make use of a (potentially) thrown Exception as the flow control for the logic:

https://ahdak.github.io/blog/effective-java-part-9/

Can't we make use of another API to find it out beforehand if stream available, and then create if required? Maybe using?

[JetStreamManagement#getStreams(String subjectFilter)](https://github.com/nats-io/nats.java/blob/b1b12b4194f479dae3cfed4e130e64a3ebcbee45/src/main/java/io/nats/client/JetStreamManagement.java#L250)

davsclaus: Yeah another PR with this suggestion is welcome

vineetsaurabh: I will create another PR with modified code.

vineetsaurabh: https://github.com/apache/camel/pull/18274

PR #18244, opened 2025-06-02
Camel JBang: Add test for custom Quarkus version

jirisimek98:
Added a new test case to Camel JBang export integration tests. When user specifies `quarkusGroupId` and `quarkusVersion` custom properties, it tests that `export` command produces a Quarkus app with this custom Quarkus build.


davsclaus:
The code is being auto formatted by the build system so you need to commit the formatted code

	modified:   dsl/camel-jbang/camel-jbang-it/src/test/java/org/apache/camel/dsl/jbang/it/ExportITCase.java

PR #18254, opened 2025-06-03
Fix several Camel Jbang functionalities on Windows when classpath url is

apupier:
used

* Checking if the path starts with classpath: before using it with Path.of
* Use Path.of instead of Paths.get as the javadoc is recommending it this way

it allows to get down from 34 failures and 2 errors to 2 errors and 2 failures on Windows

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
@@ -1871,9 +1871,11 @@ public class Run extends CamelCommand {
         }
 
         // skip dirs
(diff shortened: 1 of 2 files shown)


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Run.java
```diff
@@ -1871,9 +1871,11 @@ private boolean skipFile(String name) {
         }
 
         // skip dirs
-        Path path = Paths.get(name);
-        if (Files.exists(path) && Files.isDirectory(path)) {
-            return true;
+        if (!name.startsWith("classpath:")) {
```
apupier, line 1874: it remains the case directory specified through classpath but it wasn't handled before

PR #18269, opened 2025-06-04
CAMEL-22137: camel-infinispan : add Infinispan annotation processor

johnpoth:
# Description

https://issues.apache.org/jira/browse/CAMEL-22137

Seems the infinispan annotation processor needs to be set in the maven compiler plugin ?

# Target

# Tracking

### DIFF
--- a/components/camel-infinispan/camel-infinispan/pom.xml
+++ b/components/camel-infinispan/camel-infinispan/pom.xml
@@ -183,6 +183,24 @@
         </dependencies>
     </dependencyManagement>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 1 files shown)


davsclaus:
What JDK are you using

johnpoth:
Ah yeah I was using 23 while working on another project... Using 17 works like a charm, closing

davsclaus:
If you have a chance then reach out to the infinispan team, maybe they are not aware or something is changed to have it work with never JDKs. 

Adding that explicit maven compiler is likely a good soltuion for us, but I would assume its not really intended that end users need to do that. So likely something they are not aware.

johnpoth:
Yeah looking at their documentation it seems like that's the way to do it  https://infinispan.org/docs/stable/titles/encoding/encoding.html#adding-protostream-processor_marshalling ?

davsclaus:
Ah okay thanks. Then I think your PR is good to merge so we do the same

PR #18343, opened 2025-06-11
CAMEL-22150 - fix(KafkaRecordBatchingProcessor): improve handling of expired records and batch processing

adithyahm24:
# Description

## Cause
- The batching logic in Camel-Kafka has a major issue where records fetched in the current Kafka poll are skipped if the hasExpiredRecords method triggers processing due to the batchingIntervalMs condition. 
- These newly fetched records are not added to the exchangeList, and Kafka's polling behavior advances the offset, causing these records to be lost. 
- This results in data loss as unprocessed messages are neither processed nor retried in subsequent iterations.

### Improved Timeout Handling:
- Added logic to restart the timeoutWatch timer after processing expired batches. This ensures that the timeout is recalculated for subsequent records.
### Enhanced Record Addition Logic:
- Introduced a clear separation between handling expired records and adding new records to the batch. This ensures that new records are always processed after handling any expiration.
### Batch Processing Adjustments:
- Restarted the timeoutWatch timer after processing a batch when the maximum poll records limit is reached. This ensures accurate timing for subsequent batches.

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-22150) filed for the change (usually before you start working on it).

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/batching/KafkaRecordBatchingProcessor.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/batching/KafkaRecordBatchingProcessor.java
@@ -110,6 +110,7 @@ final class KafkaRecordBatchingProcessor extends KafkaRecordProcessor {
             timeoutWatch.takenAndRestart();
         }
 
+        // If timeout has expired, process current batch but continue to handle new records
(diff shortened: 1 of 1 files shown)


davsclaus:
/component-test camel-kafka

PR #18365, opened 2025-06-13
feat(components): camel-micrometer exchanges last timestamp

squakez:
Added a new meter to include the timestamp of the last executed exchange. The value is backed by a thread-safe static variable and it's the way suggested by the framework to mutate the values of any meter.

After the merge, the new metrics will be like:

```
# HELP camel_exchanges_last_timestamp Last exchange processed time in milliseconds since the Unix epoch
# TYPE camel_exchanges_last_timestamp gauge
camel_exchanges_last_timestamp 1.749824808777E12
```

Closes CAMEL-22153

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerConstants.java
+++ b/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerConstants.java
@@ -59,6 +59,7 @@ public final class MicrometerConstants {
     public static final String DEFAULT_CAMEL_ROUTES_RELOADED = "camel.routes.reloaded";
 
     public static final String APP_INFO_METER_NAME = "app.info";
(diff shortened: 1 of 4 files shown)


jpcorreia-promptly:
Hi everyone. Since upgrading `org.apache.camel.springboot:spring-boot` to `v4.13.0` we are often getting the following warning when the deployment gets restarted:
```
This Gauge has been already registered (MeterId{name='camel.exchanges.last.timestamp', tags=[]}), the registration will be ignored. Note that subsequent logs will be logged at debug level.
```

Is this related to these changes?

davsclaus:
as always test with latest release

squakez:
In theory the meter is registered only once. However, as suggested by @davsclaus, please, test against the latest release and if the behavior is still the same, open a Jira issue [1] reporting that. Thanks.

[1] https://issues.apache.org/jira/projects/CAMEL/issues

jpcorreia-promptly:
It's also happening with the latest release. I'll open a Jira issue, thanks.

PR #18387, opened 2025-06-17
CAMEL-22174 Catch TransportExceptions and disconnect share calls so we can recover from network outage

cunningt:
# Description

https://issues.apache.org/jira/browse/CAMEL-22174

This was brought up on the userlist - camel-smb has trouble recovering from TransportExceptions thrown by smbj if there is a network outage that affects the SMB share.   If we catch the exception and disconnect(), camel-smb can recover by reconnecting once the share is again reachable.

I've tested this both with the camel-smb consumer and the producer and it seems to work - once the connection to the SMB share can be reestablished, camel-smb will continue to poll / consume.

# Target

# Tracking


davsclaus:
this is causing streamDownload=true to not function - it appears that the data cannot be read. However its better to have roboust re-connect so we can take a look at how to make streamdownload work again or make it unsupported

PR #18396, opened 2025-06-18
Salesforce component: removed initial replay id check

lorenzobenvenuti:
# Description

See https://issues.apache.org/jira/browse/CAMEL-22173

# Target

For now the MR is targeting 4.10.x because it's easier for me to test the fix in our app

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue]

https://issues.apache.org/jira/browse/CAMEL-22173


davsclaus:
There is a bit of TODO in this code. The old option can be marked as `@Deprecated` at first for older releases. 
The code that stops the route is also debateable as no other components does that. Its better to let the route keep erroring as its not possible to subscribe to that topic. So maybe log a WARN instead.

lorenzobenvenuti:
Thanks for your feedback! As I mentioned in the Jira US, I had a few ideas about how to handle errors without stopping the route. I've pushed some changes with the second option: if the replay id is invalid, when `fallbackToLatestReplayId` is set then the consumer re-subscribes using `ReplayPreset.LATEST`; otherwise, an exception is passed to the consumer `ExceptionHandler` and no other action is taken. This means that the consumer will retry to subscribe with the same id and it will likely receive other errors, but if users want to stop the route they can plug a `BridgeExceptionHandlerToErrorHandler` and handle an `InvalidReplayIdException` in the route exception handler. LMK your thoughts. Meanwhile, I'll try to implement a couple of integration tests.

Thanks,

lorenzo

davsclaus:
Thanks, it would be great if you can get this done today/tomorrow as we cut Camel 4.13 in the weekend and it would be good to get this included in that release.

lorenzobenvenuti:
I've added a couple of integration tests and moved the `fallbackToLatestReplayId` to the endpoint (I thought it was a better option in case someone has multiple subscribers and wants to configure different behaviors); as mentioned earlier, this branch comes from `camel-4.10.x` because that's what we're using and I wanted to be able to test the changes in our application without introducing other variables, but if you think that the code looks good I'll rebase the MR and target `main`. Thanks!

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceEndpoint.java
```diff
@@ -72,6 +72,13 @@ public class SalesforceEndpoint extends DefaultEndpoint {
     @UriParam(label = "consumer", description = "The replayId value to use when subscribing to the Pub/Sub API.")
     private String pubSubReplayId;
 
+    @UriParam(label = "consumer",
+              description = "Whether the pub/sub consumer needs to fallback to the latest replay id when the provided id is not valid. "
+                            + "If set to false, the component will keep retrying; in order to treat this as an exception you can "
+                            + "use BridgeExceptionHandlerToErrorHandler and handle the exception in the route.",
+              defaultValue = "false")
+    private boolean fallbackToLatestReplayId = false;
```
davsclaus: Remove `= false`

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
```diff
@@ -407,10 +350,22 @@ public void onError(Throwable throwable) {
...
-                            initialReplayPreset = ReplayPreset.LATEST;
+                            String currReplayId = null;
+                            if (replayId != null) {
+                                currReplayId = replayId;
+                            } else if (initialReplayPreset == ReplayPreset.CUSTOM) {
+                                currReplayId = initialReplayId;
+                            }
+                            if (fallbackToLatestReplayId) {
+                                initialReplayPreset = ReplayPreset.LATEST;
+                                LOG.error("replay id: " + currReplayId
```
davsclaus: LOG.warn

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/PubSubApiClient.java
```diff
@@ -407,10 +350,22 @@ public void onError(Throwable throwable) {
...
+                                          + " is corrupt. Trying to recover by resubscribing with LATEST replay preset");
+                                replayId = null;
+                            } else {
+                                consumer.getExceptionHandler().handleException(new InvalidReplayIdException(
+                                        "Invalid replay id: " + currReplayId,
```
davsclaus: Maybe use the same word as above, that its corrupt, sine the error case is  ERROR_CORRPUTED

davsclaus (approved):

davsclaus:
Just some small things. Yeah the code looks good. You are welcome to target main branch either as new PR or change this

lorenzobenvenuti:
I've updated the PR, using `main` as the target branch, and addressed the comments. Thanks!

davsclaus:
/component-test camel-salesforce

PR #18404, opened 2025-06-19
INFRA-26918 - Clean often corrupted folder on Jenkins

apupier:
The Jenkins slave is often corrupted failing the build with this error message:
```
Caused by: org.eclipse.aether.RepositoryException: Could not parse
metadata /home/jenkins/.m2/repository/org/apache/camel/camel-buildtools/maven-metadata-local.xml:
in epilog non whitespace content is not allowed but got a (position:
END_TAG seen ...</metadata>\na... @34:2)
```

So when cleaning the workspace, deleting the parent folder. Not deleting the whole .m2 as this would slow down too much the build. Deleting the whole folder to avoid having data not in phase inside the folder.

# Description

# Target

# Tracking

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,6 +97,7 @@ pipeline {
                     stage('Clean workspace') {
                         steps {
                             cleanWs()
+                            sh 'rm -rv /home/jenkins/.m2/repository/org/apache/camel/camel-buildtools'
                             checkout scm
                         }
                     }


davsclaus (approved):

davsclaus:
ah good idea thanks

apupier:
https://issues.apache.org/jira/browse/INFRA-26918?focusedCommentId=17983459&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17983459

even if it won't fix the root issue. This is currently failing too often to be sustainable. i tried to provide a PR to remove the mentioned parent folder https://github.com/apache/camel/pull/18404 but I do not know how to test that without commiting on main branch and waiting for a new build.
Is it possible to try to modify the job on the CI itself?

apupier:
so far I spotted that it was always on the same file so finger crossed that it is this specific file only.
And checked that the Clean workspace is called at the beginning so shoudl be able to try directly (I mean no need to first clean the workspace manually)

PR #18413, opened 2025-06-19
Create template-engines.adoc

skin27:
Documentation: Create an overview of template engines used by Camel.


davsclaus:
I think this page needs some xref links from other pages so you can navigate and find this page

davsclaus:
thanks for the help I have moved the page into correct folder and added a nav

PR #18471, opened 2025-06-25
CAMEL-22201: The clock should seperate elapsed and wall-clock. This a…

davsclaus:
…llows exchange.getCreated to be accurate and has same behaviour as older Camel releases.

# Description

# Target

# Tracking

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/support/MonotonicClockTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/support/MonotonicClockTest.java
@@ -16,29 +16,26 @@
  */
 package org.apache.camel.support;
 
-import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


orpiske:
Now this makes me wonder if the calculation for the elapsed is actually correct ...

davsclaus:
> Now this makes me wonder if the calculation for the elapsed is actually correct ...

In what way? 
If the elpased() method is called then its using the nano clock as before.

oscerd (approved):

orpiske:
> > Now this makes me wonder if the calculation for the elapsed is actually correct ...
> 
> In what way? If the elpased() method is called then its using the nano clock as before.

I was wrong. I got confused about this after reading this on the System.nano documentation:

`The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative)`

But after reviewing the `elapsed` code, I see that we are doing the calculation as suggested there (more specifically, the first part of what they suggest).

PR #18483, opened 2025-06-26
CAMEL-22204: Fix iif expression parsing for Strings

graben:


graben:
Hello @davsclaus, may you have a look on this. IMHO the actual test cases of iif are "wrong". If iff is meant as a comma separated listing of one predicate and two expressions, then quotes should be a legal part of the expressions and not removed. Quoting Strings is only necessary in the predicate I think.

davsclaus:
backport to 4.10.x is not clean and easy. You are welcome to help with this and send another PR for the 4.10.x branch. Make sure to also run tests in camel-simple-joor and all of core tests

PR #18492, opened 2025-06-27
Changed typos in multiple places

rhaetor:
# Description

Documentation: The word "sent" should be "send" in some places


davsclaus:
thanks this has already been done

PR #18535, opened 2025-07-04
CAMEL-22207: Wait for quarkus:run process to complete before performing cleanup tasks

jamesnetherton:


davsclaus (approved):

oscerd (approved):

davsclaus:
We should check CSB if it also needs this kind or that it stops nicely already via camel stop command

jamesnetherton:
> We should check CSB if it also needs this kind or that it stops nicely already via camel stop command

I applied the same logic for CSB as it seems also to have been cleaning up before the app was fully shut down.

davsclaus:
LGTM

PR #18538, opened 2025-07-04
CAMEL-22173 - camel-salesforce: allow to configure fallbackToLatestRe…

lorenzobenvenuti:
…playId for both the endpoint and the component

# Description

Camel-salesforce: move `fallbackToLatestReplayId` to `SalesforceEndpointConfig` in order to allow settings values for both the endpoint and the component

# Target

# Tracking

https://issues.apache.org/jira/browse/CAMEL-22173


lorenzobenvenuti:
I see the pipeline failed because there are some uncommitted files after the build, do I need to anything on my side? I've just noticed that, after running a build locally, I have uncommitted files too (although in my case it's just one: `catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/salesforce.json`)

Thanks!

davsclaus:
Yeah I will run a regen via
```
./etc/scripts/regen.sh
```

It takes some time but I have a fast computer. Its common this happens for contributors as they need to rebuild all of the code when they change something in a component and its just harder to explain how, than do it yourself

lorenzobenvenuti:
TYVM!

PR #18550, opened 2025-07-07
CAMEL-22225: add q parameter to google-drives

davsclaus:
# Description

# Target

# Tracking


davsclaus:
its not possible to add them individually so we only add one a case by case

core/camel-support/src/main/java/org/apache/camel/support/component/AbstractApiProducer.java
```diff
@@ -113,12 +113,14 @@ public void interceptProperties(Map<String, Object> properties) {
...
      * @param  method                API method to invoke.
      * @param  properties            method arguments from endpoint properties and exchange In headers.
      * @return                       API method invocation result.
      * @throws RuntimeCamelException on error. Exceptions thrown by API method are wrapped.
      */
-    protected Object doInvokeMethod(ApiMethod method, Map<String, Object> properties) throws RuntimeCamelException {
+    protected Object doInvokeMethod(Exchange exchange, ApiMethod method, Map<String, Object> properties)
+            throws RuntimeCamelException {
         return ApiMethodHelper.invokeMethod(endpoint.getApiProxy(method, properties), method, properties);
     }
```
apupier: what about keeping a doInvokeMethod(ApiMethod method, Map<String, Object> properties) potentially marked as deprecated to avoid the API breakage?

davsclaus: Yeah we can change this back as we found another way

davsclaus:
f*** I squashed the commits before pushing here, and something does not seems right - need to check what I recently changed that cause a problem.

PR #18552, opened 2025-07-07
Interceptor EIPs should include more information where the intercepti…

Croway:
…on happened

# Description

# Target

# Tracking


core/camel-api/src/main/java/org/apache/camel/ExchangePropertyKey.java
```diff
@@ -57,6 +57,9 @@ public enum ExchangePropertyKey {
     FILE_EXCHANGE_FILE(Exchange.FILE_EXCHANGE_FILE),
     GROUPED_EXCHANGE(Exchange.GROUPED_EXCHANGE),
     INTERCEPTED_ENDPOINT(Exchange.INTERCEPTED_ENDPOINT),
+    INTERCEPTED_ENDPOINT_URI("InterceptedEndpointUri"),
```
davsclaus: These should be new Exchange.XXX like the others and start with CamelXXX so we know its from Camel. Look at one of the others and see as its needed in some code mapping between exchange and this key class

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/InterceptFromReifier.java
```diff
@@ -42,7 +42,11 @@ public Processor createProcessor() throws Exception {
             when = e -> true;
         }
         // set property before processing the child
+
         Processor p = exchange -> {
+            exchange.setProperty(ExchangePropertyKey.INTERCEPTED_ROUTE_ID.getName(), route.getId());
```
davsclaus: Use the key parameter directly so you dont need to do .getName - the former is faster

core/camel-core-reifier/src/main/java/org/apache/camel/reifier/InterceptReifier.java
```diff
@@ -61,11 +62,20 @@ public Processor wrapProcessorInInterceptors(
                 // store the target we are intercepting
                 this.interceptedTarget = target;
 
+                Processor p = exchange -> {
+                    exchange.setProperty(ExchangePropertyKey.INTERCEPTED_ROUTE_ID.getName(),
+                            definition.getParent() != null ? definition.getParent().getId() : route.getRouteId());
```
davsclaus: There is some helper code to get the route id in `ProcessorDefinitionHelper`

davsclaus:
Can you rename the PARENT -> ROUTE

davsclaus:
The documentation is still saying parent
```
The parent's endpoint URI that was intercepted
```

Croway:
> The documentation is still saying parent
> 
> ```
> The parent's endpoint URI that was intercepted
> ```

Nice catch, I am truly sorry.

PR #18537, opened 2025-07-04
CAMEL-22217: Fix potential ClassCastException in BrowseDevConsole

jamesnetherton:

PR #18569, opened 2025-07-09
CAMEL-22217: Fix browse related ClassCastException where CamelMessageTimestamp header is not present

jamesnetherton:
Follow up to https://github.com/apache/camel/pull/18537. There are a few other potential `ClassCastException` I noticed when setting  `dump=false` on `BrowseDevConsole`.


davsclaus:
thanks i have backported to 4.10.x

PR #18590, opened 2025-07-11
Provide first basic test for Camel JBang Update dependency

apupier:
It contains the first test for Camel JBang Dependency Update. Even if not dedicated to the fix, it is indirectly already testing CAMEL-22226. These tests will be the base for more advanced use case.

# Description

# Target

# Tracking


dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/DependencyUpdateTest.java
```diff
@@ -0,0 +1,104 @@
...
+        String camelFilePath = new File(workingDir, "my.camel.yaml").getAbsolutePath();
+        CommandLine.populateCommand(initCommand, camelFilePath);
+        Assertions.assertEquals(0, initCommand.doCall(), initCommandPrinter.getLines().toString());
+        
+        StringPrinter exportCommandPrinter = new StringPrinter();
+        Export exportCommand = new Export(new CamelJBangMain().withPrinter(exportCommandPrinter));
+        CommandLine.populateCommand(exportCommand,
+                "--gav=examples:route:1.0.0",
+                "--dir=" + workingDir,
+                "--camel-version=4.13.0",
```
apupier, line 91: Using a fixed version will allow to have a more stable test but the drawback is that it won't test with the latest version of exported projects. Not sure what's the best

davsclaus: its okay to have a fixed version as that should be working too. Then later something can be added for latest version, but we also do some fixed versions tests in other parts of camel-jbang

apupier: reported https://issues.apache.org/jira/browse/CAMEL-22242 to do it later

PR #18619, opened 2025-07-15
CAMEL-22227 - basic heuristic to determine runtime automatically when

apupier:
doing Camel JBang Dependency update

it allows to have correct dependencies based on current runtime of the Maven project without specifying the --runtime option (which is relatively redundant with the pom.xml)

- ideally this should resolve the Maven pom and look for specific dependencies
- looking to specific word in the file should be enough for most of the cases
- not setting the main type as it is the default fallback anyway

# Description

# Target

# Tracking


oscerd (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/DependencyUpdate.java
```diff
@@ -71,6 +72,16 @@ public Integer doCall() throws Exception {
...
         }
 
+        if (maven && this.runtime == null) {
+            // Basic heuristic to determine if the project is a Quarkus or Spring Boot one.
+            String pomContent = new String(Files.readAllBytes(file));
+            if (pomContent.contains("quarkus")) {
+                runtime = RuntimeType.quarkus;
+            } else if (pomContent.contains("spring-boot")) {
+                runtime = RuntimeType.springBoot;
+            }
```
davsclaus, line 86: if none of these then it should be RuntimeType.main I suppose

apupier: yes. i choose to not setting the main type as it is the default fallback anyway.
The advantages are that it is less code and given that it is a far-from-perfect heuristic, it keeps the default unchanged.

But these are not very clear cut advantages. I can also add to set to main, or only if I detect camel-main as might be a bit too generic

davsclaus: ok or add a code comment that the default null will be main runtime

davsclaus (approved):

PR #18642, opened 2025-07-17
CAMEL-22227 - Fix Camel JBang update dependency for Spring Boot and Quarkus

apupier:
- The major chunk is to take care of the correct directory and not assume the program is launched from the same level than the pom.xml is
- Another point is to not pick the dependencies from dependency management. Thus, still need them for the Camel JBang Runtime Dependency command as it is looking to the bom for the version and runtime type, so introduced a parameter when computing the dependencies.
- avoid setting `camel.main.routes-include-pattern` for Quarkus project. It was added when implementing Quarkus native support but I think the real needed value is the `quarkus.native.resources.includes` parameter. And it was causing trouble to Update command as it was not finding the specified Camel file in classpath

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/DependencyUpdateTest.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/DependencyUpdateTest.java
@@ -17,6 +17,7 @@
 package org.apache.camel.dsl.jbang.core.commands;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -33,6 +34,8 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 import picocli.CommandLine;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 class DependencyUpdateTest extends CamelCommandBaseTest {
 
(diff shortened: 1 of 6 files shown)


dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/DependencyUpdateTest.java
```diff
@@ -59,7 +59,29 @@ void shouldDependencyUpdate(RuntimeType rt) throws Exception {
...
+                "--dir=" + workingDir,
+                new File(workingDir, "pom.xml").getAbsolutePath());
+        int exit = command.doCall();
+        Assertions.assertEquals(0, exit, secondUpdateCommandPrinter.getLines().toString());
+
+        List<String> lines = secondUpdateCommandPrinter.getLines();
+        Assertions.assertEquals(1, lines.size(), secondUpdateCommandPrinter.getLines().toString());
+        Assertions.assertEquals("Updating pom.xml with 1 dependency added", lines.get(0));
+
+        // TODO: check pom.xml content precisely
```
apupier: working on a basic check for this part

jamesnetherton:
> * avoid setting `camel.main.routes-include-pattern` for Quarkus project. It was added when implementing Quarkus native support but I think the real needed value is the `quarkus.native.resources.includes` parameter. And it was causing trouble to Update command as it was not finding the specified Camel file in classpath

FWIW - that should not be necessary since Camel Quarkus 3.22.x. Route XML, YAML etc in the default `src/main/resources/camel` location will automatically get registered as native resources.

davsclaus:
Thanks James that is good to know that this part of CEQ is easier

apupier:
> > * avoid setting `camel.main.routes-include-pattern` for Quarkus project. It was added when implementing Quarkus native support but I think the real needed value is the `quarkus.native.resources.includes` parameter. And it was causing trouble to Update command as it was not finding the specified Camel file in classpath
> 
> FWIW - that should not be necessary since Camel Quarkus 3.22.x. Route XML, YAML etc in the default `src/main/resources/camel` location will automatically get registered as native resources.

good to know. I reported https://issues.apache.org/jira/browse/CAMEL-22254 to improve what is generated with Camel Quarkus 3.22+

apupier:
rebased and finally also tested manually. ready for review

PR #18703, opened 2025-07-23
CAMEL-22266: camel-jbang: use java 21 as default for run/export comma…

davsclaus:
…nds.

# Description

# Target

# Tracking


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Export.java
```diff
@@ -327,7 +327,8 @@ protected void copyDockerFiles(String buildDir) throws Exception {
         Path docker = Path.of(buildDir).resolve("src/main/docker");
         Files.createDirectories(docker);
         String[] ids = gav.split(":");
-        InputStream is = ExportCamelMain.class.getClassLoader().getResourceAsStream("templates/Dockerfile.tmpl");
+        InputStream is
+                = ExportCamelMain.class.getClassLoader().getResourceAsStream("templates/Dockerfile" + javaVersion + ".tmpl");
```
apupier: Shouldn't we fallback to the closest (or at least the latest) Dockerfile in case the user is specifying another java-version than 17 and 21, like 24 or 25?

oscerd (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Export.java
davsclaus: Ah yeah we can drop to 21 but we only support 17 or 21, so if they use 23 then its not our problem

davsclaus: ok we use 21 as default and only 17 if choosen

PR #18732, opened 2025-07-25
Create an OpenAI Mock and Use it in tools test instead of ollama

Croway:
Info about the OpenAI mock can be found into its README.md

### DIFF
--- a/components/camel-ai/camel-langchain4j-tools/pom.xml
+++ b/components/camel-ai/camel-langchain4j-tools/pom.xml
@@ -76,6 +76,12 @@
         <!-- for testing -->
         <dependency>
             <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-infra-openai-mock</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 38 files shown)


orpiske (commented):
This looks very promising. I have some thoughts about things to consider:

1. Our current tests are integration tests and I think we should leave at least 1 or 2 of those. Wdyt? 
2. The tests should be renamed from *IT.java to *Test.java, so that they run with Surefire instead of Failsafe. 
3. You can remove the `@DisabledIfSystemProperty` annotation

orpiske:
Thanks for this, looking forward to having that on on the code 🙏

zbendhiba:
LGTM! That is really helpful to continuous testing when we need to upgrade LangChain4j dependencies.

Croway:
Thanks @orpiske @zbendhiba 

I have reitroduced IT tests with Ollama, I am having issues executing those tests, could you guys try a `mvn verify` on this branch?

If both Tests and ITs are successful, then, it can be merged

zbendhiba:
I can't run the tests with Ollama-test-infra anymore, because the version of Ollama is too slow to run the IT tests. I usually point to a local running version

davsclaus (approved):

PR #18737, opened 2025-07-26
minor: Replaced get() with getProperty() for type safety and clarity

Atharv3221:
Replaced get() with getProperty() for config access.
get() comes from Hashtable and returns a generic Object, while getProperty() is type-safe and clearer.
This improves readability and avoids unnecessary casting.


oscerd (approved):

Croway:
Thanks @Atharv3221 !

PR #18738, opened 2025-07-26
CAMEL-22278: Improve rabbitmq producer performance

Croway:


components/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerIT.java
```diff
@@ -39,6 +47,92 @@
...
+
+        int finalSentCount = messageCounter.get();
+        LOG.info(">>> Total messages sent in {} seconds: {}", totalDurationSeconds, finalSentCount);
+
+        // --- 9. NEW: Receive All Messages from the Queue ---
+        LOG.info("\n--- Draining and Verifying Queue Content ---");
+        AmqpTemplate rabbitConsumerTemplate = new RabbitTemplate(cf);
+
+        int receivedCount = 0;
+        long lastMessageTime = System.currentTimeMillis();
```
orpiske, line 113: Just as a heads up, it might be better to use the `StopWatch` for this, as `currentTimeMillis` is non-monotonic and subject to time drift (i.e.: it can go back in time under some circumstances). This could cause the test to appear flaky, when it's just subject to rare (but still possible) external disturbances to the system clock.

PR #18755, opened 2025-07-29
CAMEL-22282: Removes CamelHttpPath header from http producer on bridge

mcarlett:
In case of spring boot binding the raw http path is appended to the bridged endpoint


davsclaus:
the option skipControlHeaders is likely better than the old bridgeEndpoint option

PR #18756, opened 2025-07-29
Use StopWatch instead of System.currentTimeMillis()

Croway:
@orpiske thanks for the suggestion


components/camel-spring-parent/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerIT.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.camel.RoutesBuilder;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.springrabbit.SpringRabbitMQConstants;
+import org.apache.commons.lang3.time.StopWatch;
```
orpiske: We have our own `StopWatch` in `org.apache.camel.util`. I think it can do all of this for you.

davsclaus (approved):

PR #18778, opened 2025-07-30
CAMEL-22196: Create an AI agent component

zbendhiba:
# Description

# Target

# Tracking


components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -0,0 +1,337 @@
...
+        List<Class<?>> inputGuardrailClasses = parseGuardrailClasses(endpoint.getConfiguration().getInputGuardrails());
+
+        // Output Guardrails
+        List<Class<?>> outputGuardrailClasses = parseGuardrailClasses(endpoint.getConfiguration().getOutputGuardrails());
+
+        // create an AI Service that deals with all of it
+        String response = "";
+
+        // There is a restriction in using the @MemoryId when building an AIService, so we need to have 2 different AIServices : one that deals with memory and one that doesn't
+        if (chatMemoryProvider != null) {
```
orpiske: One of my concerns here is that I think this can grow very complex very fast, given the number of scenarios we could have in the future. 

What do you think about separating this part into an interface and implementation? Something like `Agent` as the interface and then 2 implementations of that interface: one for each branch? That could allow us to grow further in complexity without making the component code very complicated.

orpiske (commented):
Thanks Zineb, at a quick glance this is looking like a great first step for this new component!

I will complete the review later in the morning, but wanted to leave an initial comment about something that came to my mind.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
zbendhiba: @orpiske  Good insights. Let me try this!

Croway:
Looks really good @zbendhiba !!

I was wondering if it would make sense to provide `ChatMemoryStore` default implementations using Camel, for example, we may provide a Camel infinispan/redis ChatMemoryStore implementation, that is automatically added to the Camel registry, this way, the user can configure (via properties for example) the camel-infinispan or redis component, and automagically, the bean is added to the Camel registry

Croway (approved):

orpiske:
> Looks really good @zbendhiba !!
> 
> I was wondering if it would make sense to provide `ChatMemoryStore` default implementations using Camel, for example, we may provide a Camel infinispan/redis ChatMemoryStore implementation, that is automatically added to the Camel registry, this way, the user can configure (via properties for example) the camel-infinispan or redis component, and automagically, the bean is added to the Camel registry

So, this is what we intend to do with the Bean Catalog. I believe we might be able to do that, once we get the changes I  suggested.

zbendhiba:
> tMemoryStore implementation, that is automatically added to the Camel registry, this way, the user can configure (via properties for exa

yes it would be fantastic. And relevant in Camel. I was imagining something to add to the Route like in the case of idempotent EIP  (it could be implemented in many types of components that can persist and clean the memory when it's needed). Of course, I haven't started anything, and any better idea is welcome !

zbendhiba:
@otavioprado  @Croway 
I think my idea is a mix of both your ideas. Currently, there isn't a good implementation of a LangChain4j memory stores. I believe we could reuse our Camel components to achieve the desired behavior, which is why I'm bringing up the Idempotent EIP—it's something to consider.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
zbendhiba: @orpiske  I pushed a new commit. Let me know if that is okay. I'll rebase if you confirm

zbendhiba: I meant I'll squash my commits instead of rebase :)

zbendhiba:
Many thanks for the reviews @orpiske 

that last push, I'm only squashing my commits

oscerd (approved):

oscerd:
This looks really good, thanks a lot.

PR #18921, opened 2025-08-18
CAMEL-22356: Allow Google PubSub consumer pass exception through error handler

seanshi007:
# Description

Downstream currently has no way to know when Google PubSub consumer fails as exception is simply logged and ignored. This change will let Google PubSub consumer to pass exception through bridge error handler (when enabled) to allow downstream to react to the exception. 

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [CAMEL-22356](https://issues.apache.org/jira/browse/CAMEL-22356) filed for the change (usually before you start working on it).


davsclaus:
there is also a `camel-google-pubsub-lite` component - would you mind doing the same fix there also ?

seanshi007:
> there is also a `camel-google-pubsub-lite` component - would you mind doing the same fix there also ?

Sure~ Here is the same fix for PubSub Lite: https://github.com/apache/camel/pull/18922

PR #18996, opened 2025-08-26
CAMEL-22364: camel-jms - Fix rare ConcurrentModificationException in …

davsclaus:
…camel-jms due to destination produced is set as header after sent, which can lead to a race condition when doing request/reply over JMS and reply is already received and continued processing. Thanks to Vladimir Dobos for the analysis.

# Description

# Target

# Tracking


davsclaus:
the test error is due to mq test infra port clash on 1414

PR #19000, opened 2025-08-26
Create group.adoc

skin27:
Camel 4.14.0 enhanced the Core API and MBean to allow more actions by group. This adds the documentation to the user manual. Feel free to enhance/improve.


skin27:
I would suggest that it comes under "Architecture"

davsclaus:
The file should be renamed to `route-group` as the group is for routes, not a generic group terminoligy. Look at route-template etc where its used and how its linked in the nav files.

skin27:
OK, done.

PR #19035, opened 2025-08-29
(chores) camel-ai: replace langchain4j RC versions with their final versions

orpiske:


orpiske:
This supersedes #19030

PR #19071, opened 2025-09-04
feat(CAMEL-22395): camel-langchain4j-tools component improvement for chat memory support

jillyan:
# Description

# Target

# Tracking


Croway:
Hello @jillyan , thanks for the PR, I have two question:

* Why are you targetting camel-4.8.x branch? can't you do this on main? (It will be easier for us to review and test)
* Can you add a test that shows how this feature is supposed to be used?

oscerd (changes requested):
Thanks, but this should go in main. We are not going to add new parameters on old branches.

davsclaus:
Are you going to change this to the main branch, as we only accept new work on the main branch.

If not then we will close this PR

davsclaus:
@orpiske @cunningt is this something that we can try to manually patch for main branch to get this ticket moving forward. Is the feature worth it ?

orpiske:
I don't think it is worth the trouble. The new agents component has memory support in a way that is much easier and simpler to use.

My vote is to close without accepting this.

davsclaus:
okay lets close this.

PR #19097, opened 2025-09-08
Handle Micro versions in camel jbang update list

Croway:
The code didn't handle camel-upgrade-recipes micro versions, ex. 4.14.0.1


ppalaga (commented):
Thanks @Croway, looks good!

PR #19110, opened 2025-09-09
CAMEL-22408: OpenAPI Description showing wrong allowable dates depending on system timezone

ppalaga:
# Description

See https://issues.apache.org/jira/browse/CAMEL-22408

# Target

# Tracking
- [x] https://issues.apache.org/jira/browse/CAMEL-22408

### DIFF
--- a/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiSupport.java
+++ b/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiSupport.java
@@ -25,6 +25,7 @@ import java.text.SimpleDateFormat;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -60,7 +61,12 @@ import static org.apache.camel.openapi.RestDefinitionsResolver.JMX_REST_DEFINITI
  */
 public class RestOpenApiSupport {
 
-    public static DateFormat DEFAULT_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+    private static final DateFormat DEFAULT_DATE_FORMAT;
+    static {
+        final SimpleDateFormat f = new SimpleDateFormat("yyyy-MM-dd");
+        f.setTimeZone(TimeZone.getTimeZone("UTC"));
+        DEFAULT_DATE_FORMAT = f;
+    }
 
     static final String HEADER_X_FORWARDED_PREFIX = "X-Forwarded-Prefix";
     static final String HEADER_X_FORWARDED_HOST = "X-Forwarded-Host";
--- a/components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
+++ b/components/camel-openapi-java/src/test/java/org/apache/camel/openapi/RestOpenApiReaderTest.java
@@ -16,14 +16,6 @@
  */
 package org.apache.camel.openapi;
 
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.TimeZone;
-
-import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.swagger.v3.core.util.Json;
-import io.swagger.v3.core.util.Json31;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.info.Info;
 import org.apache.camel.BindToRegistry;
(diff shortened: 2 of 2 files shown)


ppalaga:
@ffang you may want to review

components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiSupport.java
```diff
@@ -60,7 +61,12 @@
  */
 public class RestOpenApiSupport {
 
-    public static DateFormat DEFAULT_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+    private static final DateFormat DEFAULT_DATE_FORMAT;
+    static {
+        final SimpleDateFormat f = new SimpleDateFormat("yyyy-MM-dd");
+        f.setTimeZone(TimeZone.getTimeZone("UTC"));
```
zhfeng, line 67: In the original test, it used `testDateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));`, not sure if it is the same with `"UTC"`?

ppalaga: Yeah, GMT is basically the old name of UTC https://en.wikipedia.org/wiki/Greenwich_Mean_Time#:~:text=The%20term%20%22GMT%22%20is%20also%20used%20as%20one%20of%20the%20names%20for%20the%20time%20zone%20UTC%2B00%3A00

zhfeng (approved):

oscerd (approved):

bvahdat:
Hi @ppalaga 

As `SimpleDateFormat` is not thread-safe, would we then run into problems where `RestOpenApiSupport#getJsonFromOpenAPIAsString` API is used concurrently through multiple threads? I think safer would be to instantiate it per request / call.

ppalaga:
> Hi @ppalaga
> 
> As `SimpleDateFormat` is not thread-safe, would we then run into problems where `RestOpenApiSupport#getJsonFromOpenAPIAsString` API is used concurrently through multiple threads? I think safer would be to instantiate it per request / call.

I agree, `SimpleDateFormat` is not thread-safe, but I have no idea whether `RestOpenApiSupport#getJsonFromOpenAPIAsString` can be called concurrently. It would be great if some subject matter could check.

bvahdat:
> > Hi @ppalaga
> > As `SimpleDateFormat` is not thread-safe, would we then run into problems where `RestOpenApiSupport#getJsonFromOpenAPIAsString` API is used concurrently through multiple threads? I think safer would be to instantiate it per request / call.
> 
> I agree, `SimpleDateFormat` is not thread-safe, but I have no idea whether `RestOpenApiSupport#getJsonFromOpenAPIAsString` can be called concurrently. It would be great if some subject matter could check.

I meant, what would happen if multiple rest clients simultaneously call the exposed rest endpoint by https://github.com/apache/camel-quarkus/issues/7717 ? Do we maybe run into unexpected results again as originally reported by the issue above?

```
[ERROR]   OpenApiV3Test.arrayTypes:368 1 expectation failed.
JSON path paths.'/api/operation/spec/array/params'.get.parameters[9].schema.enum doesn't match.
Expected: iterable containing ["2023-01-01", "2023-02-02", "2023-03-03"]
  Actual: <[2022-12-31, 2023-02-01, 2023-03-02]>

[ERROR]   OpenApiV3Test.arrayTypes:368 1 expectation failed.
JSON path paths.'/api/operation/spec/array/params'.get.parameters[9].schema.enum doesn't match.
Expected: iterable containing ["2023-01-01", "2023-02-02", "2023-03-03"]
  Actual: <[2022-12-31, 2023-02-01, 2023-03-02]>
``` 

As because the same `SimpleDateFormat` object is now reused across multiple HTTP caller threads through this fix.

ppalaga:
If the code in `RestOpenApiSupport#getJsonFromOpenAPIAsString` can be called concurrently (I am not sure you are confirming that's the case), then no, it would not be the case that issue #7717 is re-introduced again. That would be a different issue, caused by some racing internals of `SimpleDateFormat`. That issue was neither introduced nor worsened by the current PR. 
So if somebody confirms that `RestOpenApiSupport#getJsonFromOpenAPIAsString` can really be called concurrently, than `RestOpenApiSupport.DEFAULT_DATE_FORMAT` has to be transformed to a thread local or inlined into `getJsonFromOpenAPIAsString()`.

bvahdat:
> If the code in `RestOpenApiSupport#getJsonFromOpenAPIAsString` can be called concurrently (I am not sure you are confirming that's the case), then no, it would not be the case that issue #7717 is re-introduced again. That would be a different issue, caused by some racing internals of `SimpleDateFormat`. That issue was neither introduced nor worsened by the current PR. So if somebody confirms that `RestOpenApiSupport#getJsonFromOpenAPIAsString` can really be called concurrently, than `RestOpenApiSupport.DEFAULT_DATE_FORMAT` has to be transformed to a thread local or inlined into `getJsonFromOpenAPIAsString()`.

I'm not much familiar with the codebase but the static signature of `RestOpenApiSupport#getJsonFromOpenAPIAsString` smells after that but I'm not sure.

cc: @zhfeng @oscerd  @davsclaus

davsclaus:
yes it can be called concurrently in rest-dsl etc. So yeah its better to create a local simple data format instance per call

PR #19141, opened 2025-09-10
Defer datasource creation via a factory

Croway:
Add dataSourceFactory property to the sql, jdbc and spring-jdbc endpoints.

This way user can provide a datasource factory that creates DS at runtime.


oscerd (approved):

core/camel-support/src/main/java/org/apache/camel/support/sql/DataSourceFactory.java
```diff
@@ -0,0 +1,12 @@
+package org.apache.camel.support.sql;
+
+import org.apache.camel.CamelContextAware;
+import org.apache.camel.Exchange;
+
+import javax.sql.DataSource;
+
+public interface DataSourceFactory extends CamelContextAware {
```
orpiske, line 37: It would be good to have a bit of javadoc on this one

orpiske (commented):
LGTM, but it would be nice to have a bit of javadoc on the factory.

PR #19142, opened 2025-09-11
CAMEL-22011: camel-xslt-saxon: Allow JSON input body

igarashitm:
# Description

camel-xslt-saxon assumes the input body as XML by default. This introduces an option `useJsonBody` to allow consuming JSON data from Message body, convert it into XML representation of JSON and use it as an XSLT primary input.

# Target

# Tracking
https://issues.apache.org/jira/browse/CAMEL-22011


igarashitm:
For some reason it's failing with
```
Caused by: java.lang.ClassNotFoundException: jakarta.xml.bind.JAXBException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
```
Although `camel-xj` uses the same `xmlunit-assertj3` for the test, it runs OK...

igarashitm:
fixed - `camel-spring-xml` was missing, and then `xmlunit-assertj3` was bringing in too old JAXB dependency.

components/camel-xslt-saxon/src/main/java/org/apache/camel/component/xslt/saxon/SaxonXmlSourceHandlerFactoryImpl.java
```diff
@@ -40,4 +81,51 @@ protected Source getSource(Exchange exchange, Object body) {
         }
         return source;
     }
+
+    private Source convertJsonToXmlSource(String jsonString) {
+        try {
+            net.sf.saxon.s9api.XPathSelector selector = getSaxonJsonToXmlExecutable().load();
```
Croway, line 87: is XPathSelector TrheadSafe? in case can this be initialized only once?

igarashitm: According to the Saxon Javadoc it's up to `XPathExecutable` being thread-safe, and `XPathSelector` is suggested to load every time (which makes sense as XPathSelector will hold compiled XPath expression, i.e. will hold JSON payload)
https://www.saxonica.com/html/documentation12/javadoc/net/sf/saxon/s9api/XPathExecutable.html

PR #19152, opened 2025-09-11
Defer Bean DataSource creation in sql and jdbc components

Croway:


davsclaus:
is there a JIRA for this and it would be good to add to the 4.15 migration guide that this factory API is removed

Croway:
> is there a JIRA for this and it would be good to add to the 4.15 migration guide that this factory API is removed

No need to document it. I introduced it last week, but I am going to remove it. There should be more elegant way to achieve the same.

davsclaus:
OK

davsclaus:
Should this be merged ?

PR #19153, opened 2025-09-12
CAMEL-22412 Add langchain4j-embeddingstore component to support vector stores

cunningt:
# Description

https://issues.apache.org/jira/browse/CAMEL-22412

This PR adds a langchain4j-embeddingstore component  :

It has the following actions
    - ADD - Store embeddings with optional text segments and metadata
    - REMOVE - Delete embeddings by unique id
    - SEARCH - Perform similarity search 

# Target

# Tracking


components/camel-ai/camel-langchain4j-embeddingstore/src/main/java/org/apache/camel/component/langchain4j/embeddingstore/EmbeddingStoreFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ * Factory interface for creating EmbeddingStore instances within the Apache Camel LangChain4j integration.
+ *
+ * <p>
+ * The factory extends {@link CamelContextAware} to ensure proper integration with the Camel context and access to
+ * registry components.
+ * </p>
+ *
+ * @since 4.14.0
+ */
+public interface EmbeddingStoreFactory extends CamelContextAware {
```
orpiske, line 33: I'm wondering if we should split this one in runtime and API, like we did for the agent component. On the other hand, this component is small, so I think this would be overkill.

components/camel-ai/camel-langchain4j-embeddingstore/src/main/java/org/apache/camel/component/langchain4j/embeddingstore/LangChain4jEmbeddingStore.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Message headers used by the embedding store component.
+     *
+     * <p>
+     * These headers control the behavior of embedding store operations and provide additional parameters for search and
+     * filter operations.
+     * </p>
+     */
+    public static class Headers {
```
orpiske, line 49: Recently I had a chat with @davsclaus where we discussed about keeping these headers separately. I ended up reporting CAMEL-22334 for this, so I think it would be good to avoid introducing more of this, so it makes fixing the existing ones a bit easier.

orpiske:
Hey @cunningt, this LGTM, but the only thing that we really need to adjust is the nested headers class. The `-api` thing is just a suggestion and, possibly, overkill (but, please, let me know what you think).

components/camel-ai/camel-langchain4j-embeddingstore/src/main/java/org/apache/camel/component/langchain4j/embeddingstore/LangChain4jEmbeddingStore.java
cunningt: Sounds good to me - split the headers out into `LangChain4jEmbeddingStoreHeaders` and I can start helping out with making those changes to other components.

PR #19165, opened 2025-09-12
MongoDB upgrade - with fixes

orpiske:


orpiske:
This one supersedes #19156.

PR #19173, opened 2025-09-13
CAMEL-22413: Explain the situation to use Kamelet EIP in the catalog

igarashitm:
# Description

Add explanation about when Kamelet EIP should be used instead of Kamelet component, and suggest to use Kamelet component in other cases. 

# Target

# Tracking
https://issues.apache.org/jira/browse/CAMEL-22413

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/model/KameletDefinition.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/model/KameletDefinition.java
@@ -27,7 +27,11 @@ import jakarta.xml.bind.annotation.XmlRootElement;
 import org.apache.camel.spi.Metadata;
 
 /**
- * To call Kamelets in special situations
+ * To call Kamelets in special situations. By default, calling kamelets should be done as endpoints with the kamelet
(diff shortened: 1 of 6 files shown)


core/camel-core-model/src/main/java/org/apache/camel/model/KameletDefinition.java
```diff
@@ -27,7 +27,11 @@
 import org.apache.camel.spi.Metadata;
 
 /**
- * To call Kamelets in special situations
+ * To call Kamelets in special situations. By default, calling kamelets should be done as endpoints with the kamelet
+ * component, such as <code></code>to("kamelet:mykamelet")</code>. When a Kamelet is designed for a special use-case
```
Croway: `<code></code>` ?

davsclaus: yeah the `<code>` should be removed

igarashitm:
OK removed

davsclaus:
This file is updated during the build

	modified:   core/camel-core-model/src/main/java/org/apache/camel/model/KameletDefinition.java

igarashitm:
ugh sorry my bad, rebuilt

PR #19174, opened 2025-09-13
Enhance Build Scan Metadata

tylerbertrand:
# Description

Applies the Common Custom User Data Maven Extension to enhance Develocity Build Scans with additional metadata, like custom values that capture the git commit the build was run at, tags that capture the git branch, links that allow easy navigation between the Build Scan and the CI job that ran the build, and more.

# Target

# Tracking

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -15,4 +15,9 @@
         <artifactId>develocity-maven-extension</artifactId>
         <version>2.1</version>
     </extension>
+    <extension>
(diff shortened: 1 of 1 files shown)


orpiske (changes requested):
Thanks! I need to check this with ASF Infra before we can merge. Marking it as request for changes until we can do that.

orpiske:
Tracker for ASF Infra: https://issues.apache.org/jira/browse/INFRA-27228

PR #19175, opened 2025-09-13
CAMEL-22334 Move nested Headers classes to their own Headers class

cunningt:
# Description

https://issues.apache.org/jira/browse/CAMEL-22334

This commit affects classes which have been previously released in 4.14 and earlier, and would be a breaking change.     Do we need a deprecation strategy here or is it allowable given these are all camel-ai components?

# Target

# Tracking


davsclaus:
Its not a big problem, just add a note in the 4.15 migration guide about this. The header values are the same

davsclaus:
LGTM

PR #19213, opened 2025-09-16
CAMEL-22423: ContextServicePlugin SPI - provides a plugin system for Apache Camel …

Croway:
…that allows automatic discovery and initialization of third-party components during CamelContext startup.

The ContextServicePlugin provides an entry point into the Camel Context for third party plugins. This way, 3d party dependencies can be created to offer independent, additional features to a Camel application.

For example, a set of custom beans can be easily provided by
```
public class MyContextServicePlugin implements ContextServicePlugin {
    @Override
    public void load(CamelContext camelContext) {
        camelContext.getRegistry().bind("myBean", new TestBean());
        camelContext.getRegistry().bind("mySecondBean", new AnotherTestBean());
        ...
    }
}
```
 
 and the service `META-INF/services/org.apache.camel.spi.ContextServicePlugin` with content `com.foo.acme.TestPlugin`

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/ContextServicePlugin.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/ContextServicePlugin.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


Croway:
https://issues.apache.org/jira/browse/CAMEL-22423

jamesnetherton (approved):

PR #19246, opened 2025-09-18
fix(ci): exit on error 

squakez:
# Description

This script is used during github check. When the first maven execution fails, it continues to the second. It is convenient to stop immediately instead.
Additionally with this PR avoid the developer to download the build log file in case of error. He can now have a sneak peak at the workflow log to understand what was wrong.

# Target

# Tracking

### DIFF
--- a/etc/scripts/regen.sh
+++ b/etc/scripts/regen.sh
@@ -26,6 +26,19 @@ git clean -fdx
 rm -Rf **/src/generated/
 
 # Regenerate everything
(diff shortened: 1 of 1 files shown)


squakez:
The checks execute as expected:
```
Run ./etc/scripts/regen.sh
Removing mvnd.zip
Removing mvnd.zip.sha256
✅ mvn -Pregen succeeded.
✅ mvn install for camel-catalog succeeded
```
in case of PR failure, we should get an error with some details.

davsclaus (approved):

PR #19247, opened 2025-09-18
CAMEL-22423: Start the ContextServiceLoaderPluginResolver

Croway:
https://issues.apache.org/jira/browse/CAMEL-22423

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
@@ -70,7 +70,6 @@ import org.apache.camel.ShutdownRunningTask;
 import org.apache.camel.StartupListener;
 import org.apache.camel.StartupStep;
 import org.apache.camel.StartupSummaryLevel;
-import org.apache.camel.StatefulService;
 import org.apache.camel.Suspendable;
 import org.apache.camel.SuspendableService;
 import org.apache.camel.TypeConverter;
@@ -106,6 +105,7 @@ import org.apache.camel.spi.CliConnectorFactory;
 import org.apache.camel.spi.ComponentNameResolver;
 import org.apache.camel.spi.ComponentResolver;
 import org.apache.camel.spi.ConfigurerResolver;
+import org.apache.camel.spi.ContextServiceLoaderPluginResolver;
 import org.apache.camel.spi.DataFormat;
 import org.apache.camel.spi.DataFormatResolver;
 import org.apache.camel.spi.DataType;
@@ -350,7 +350,7 @@ public abstract class AbstractCamelContext extends BaseService
      * Called during object construction to initialize context plugins
      */
     protected void initPlugins() {
-        camelContextExtension.addContextPlugin(StatefulService.class, createContextServiceLoaderPlugin());
+        camelContextExtension.addContextPlugin(ContextServiceLoaderPluginResolver.class, createContextServiceLoaderPlugin());
         camelContextExtension.addContextPlugin(StartupConditionStrategy.class, createStartupConditionStrategy());
         camelContextExtension.addContextPlugin(CamelBeanPostProcessor.class, createBeanPostProcessor());
         camelContextExtension.addContextPlugin(CamelDependencyInjectionAnnotationFactory.class,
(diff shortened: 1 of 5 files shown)


Croway:
> LGTM. However I wonder if this can introduce any breaking compatibility. Above all the fact we're changing the StatefulService class. How it impacts any previous user?
> 
> Also, just out of curiosity, why it affects KameletMain?

Hello @squakez , thanks for the review!

- `How it impacts any previous user?` it does not, I introudced the StatefulService with this [commit](https://github.com/apache/camel/pull/19213/files#diff-fbb1df8a09e10e8b27eea4ee18615247e2647cd28a4692c69540c951d37b8cfcR353), but, with some testing I noticed that it was not enough (or even correct). This is why instead of using StatefulService directly, I am using a new interface.
- `why it affects KameletMain?` good question, for Camel JBang, I want the service to be started when running via Camel JBang, Camel Main, Camel Spring Boot and Camel Quarkus.

squakez (approved):

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
```diff
@@ -2383,6 +2383,23 @@ public void doBuild() throws Exception {
...
+        // Start context service loader plugin to discover and load third-party plugins early in build phase
+        ContextServiceLoaderPluginResolver contextServicePlugin
+                = camelContextExtension.getContextPlugin(ContextServiceLoaderPluginResolver.class);
+        if (contextServicePlugin != null) {
+            try {
+                StartupStep step6 = startupStepRecorder.beginStep(CamelContext.class, null, "Start ContextServiceLoaderPlugin");
+                ServiceHelper.startService(contextServicePlugin);
+                startupStepRecorder.endStep(step6);
+            } catch (Exception e) {
+                LOG.warn("Cannot start context service loader plugin due: {}. Third-party context plugins may not be loaded.",
```
davsclaus, line 2395: Is there a specific reason for catching this exception?

If something failed on startup IMHO its better to fail as otherwise you have a "broken system". The other bootstrap code does not do this.

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/AbstractCamelContext.java
```diff
@@ -350,7 +350,7 @@ protected AbstractCamelContext(boolean build) {
      * Called during object construction to initialize context plugins
      */
     protected void initPlugins() {
-        camelContextExtension.addContextPlugin(StatefulService.class, createContextServiceLoaderPlugin());
+        camelContextExtension.addContextPlugin(ContextServiceLoaderPluginResolver.class, createContextServiceLoaderPlugin());
```
davsclaus, line 353: Yeah it must not be using a generic interface. So having its own interface is a must.

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/KameletMain.java
```diff
@@ -732,6 +734,21 @@ protected CamelContext createCamelContext() {
             throw RuntimeCamelException.wrapRuntimeException(e);
         }
 
+        // Start the context service loader plugin after all dependencies and downloaders are set up
+        // This ensures the classpath is enhanced and ServiceLoader can discover third-party plugins
```
davsclaus, line 738: ok so this force these to be loader earlier than usual (than standard in camel-core) and this allows camel-jbang etc to have custom plugins, for example also download dependencies and whatnot during their startup. Otherwise this kind of code is executed in camel-core already.

????

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/KameletMain.java
```diff
@@ -732,6 +734,21 @@ protected CamelContext createCamelContext() {
...
+        ContextServiceLoaderPluginResolver contextServicePlugin = answer
+                .getCamelContextExtension().getContextPlugin(ContextServiceLoaderPluginResolver.class);
+        if (contextServicePlugin != null) {
+            try {
+                // force start context service loader plugin to discover and load third-party plugins
+                ServiceHelper.startService(contextServicePlugin);
+            } catch (Exception e) {
+                LOG.warn(
```
davsclaus, line 746: Again I think this exception should not be caught but to let it fail fast instead.

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/KameletMain.java
Croway: Exactly, the camel-core code is not executed by Camel JBang, this is why I had to start the service in KameletMain

PR #19297, opened 2025-09-23
feat(components): Camel MDC

squakez:
# Description
Closes CAMEL-21889 which implements proposal in https://github.com/apache/camel/blob/main/proposals/mdc.adoc

With this in place, the user will have just to provide the component in the pom and additionally the `camel.mdc.enabled=true` (plus eventually other configurations as shown in doc). Then he can show the MDC information via regular logging variables (ie, in log4j, `%X{exchangeID}`).

- [x] Core components architecture
- [x] Include headers
- [x] Include user headers
- [x] Unit test
- [x] Doc / Proposal upgrade

I will develop separately the following tasks:

- [ ] CSB starter
- [ ] Quarkus extension
- [ ] Previous MDC deprecation (likely in 4.16)

# Target

# Tracking


components/camel-mdc/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <properties>
+    <firstVersion>4.15.0</firstVersion>
+    <label>logging</label>
+    <supportLevel>Preview</supportLevel>
+  </properties>
+
+  <artifactId>camel-mdc</artifactId>
+  <packaging>jar</packaging>
+  <name>Camel :: MDC</name>
+  <description>Mapped Diagnostic Context component</description>
```
davsclaus: I dont think many knows this is related to logging. So maybe

Logging MDC support

components/camel-mdc/src/main/java/org/apache/camel/mdc/MDCProcessorsInterceptStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public Processor wrapProcessorInInterceptors(
+            CamelContext camelContext,
+            NamedNode processorDefinition, Processor target, Processor nextTarget)
+            throws Exception {
+        return new DelegateAsyncProcessor(new TraceProcessor(target));
+    }
+
+    private class TraceProcessor implements Processor {
```
davsclaus: This would force the downstream processor to be processed synchronously - That works with MDC because it would be the same thread that continue processing .... 

The async is hard to get correct with MDC and we have MDCUnitOfWork doing this, and then some special code in the routing engine when Camel routing engine goes async.

core/camel-main/src/main/docs/main.adoc
```diff
@@ -562,6 +562,15 @@ The camel.telemetryDev supports 4 options, which are listed below.
 |===
 
 
+=== Camel MDC configurations
+The camel.mdc supports 0 options, which are listed below.
```
davsclaus: Oh something seems wrong here saying 0 options

core/camel-main/src/main/java/org/apache/camel/main/MdcConfigurationProperties.java
```diff
@@ -0,0 +1,57 @@
...
+package org.apache.camel.main;
+
+import org.apache.camel.spi.BootstrapCloseable;
+import org.apache.camel.spi.Configurer;
+
+/**
+ * Global configuration for MDC
+ */
+@Configurer(extended = true)
+public class MdcConfigurationProperties implements BootstrapCloseable {
```
davsclaus, line 26: Looks like there should be aded some options here you can configure

davsclaus (commented):
See something like AsyncEndpointRecipientList2Test that has a special test component that process async, which makes Camel continue routing in another thread. This can maybe be used for MDC unit test to see if all this still work when this happens.

squakez:
> See something like AsyncEndpointRecipientList2Test that has a special test component that process async, which makes Camel continue routing in another thread. This can maybe be used for MDC unit test to see if all this still work when this happens.

Sure, I'll give a look at that. However, I think the wiretap (which is used in the async test here) is processing async as well. As the MDC is called "onLog" and later cleared, we are sure we're always consistent as during the onLog we get the values available on the Exchange.

components/camel-mdc/src/main/java/org/apache/camel/mdc/MDCProcessorsInterceptStrategy.java
squakez: That was the point of the proposal in https://github.com/apache/camel/blob/main/proposals/mdc.adoc which this PR is addressing. It inherit the same mechanism we used for tracing, where we use the Exchange as a way to consistently get information throughout all the processing. In this way we don't need any component to take care of context propagation which is hard when we deal with async processing.

core/camel-main/src/main/docs/main.adoc
squakez: I'll have a look.

squakez:
I've fixed the problems, it should be good now. I have also changed the test and included a thread verification name on the async execution to make sure the async component (wiretap) is really executing on a different thread and still the MDC is carried over.

In order to clarify the intent of this PR I've added some more documentation that I hope it clarifies the main aspects of why we should move from the older MDC to this newer implementation. Here an extract:

> The goal of this component is to avoid to work on low level API in Java. In older MDC implementations you had to hack into the code to include MDC such as:

```java
    org.slf4j.MDC.put("myCustomMDCKey", "myCustomKeyValue");
```
> And later you had to make sure to provide MDC context propagation in async components (eg, `wiretap`) in order to make sure to have such context available in the new executing async thread. With this new service, the only thing to do is to add the value as a Camel Exchange header (or a property), for example:
```java
    .setHeader("myCustomMDCKey", simple("myCustomKeyValue"))
```
> include the MDC service and additionally instruct the Camel application to treat that header as a MDC trace (via `camel.mdc.customHeaders=myCustomMDCKey` or `*` to include all headers). You won't need any longer to worry about context propagation as the propagation will be done via Camel Exchange instead.
>
>NOTE: you won't also need any longer to hack the code using Java DSL, as you can put headers in any Camel DSL.
>
>Depending on the logging technology used, you can now include the MDC parameters you want to trace in your logging configuration. For example, in `log4j2` configuration you can include them as shown below:
>
```
... [%X{camel.contextId}, %X{camel.routeId}, %X{camel.exchangeId}, %X{camel.messageId}, %X{customHead}, %X{customProp}]
```
>
>During the execution you can verify the output of the log to see the traces appended to your logger.

Basically we foster a higher level of abstraction to let the user have the feature without the need to do things at a low level. We also make this available to any other DSL, which was not possible before. We simplify also the long term maintenance as we have higher cohesion in a unique component instead of code dealing in multiple places. And we consistently have a context propagation, leveraging the Exchange mechanism instead of any other lower level threading mechanism.

Feel free to ask for any further clarification. Thanks!

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -2741,6 +2781,20 @@ private static CamelMetricsService resolveMicrometerService(CamelContext camelCo
...
+    private static CamelMDCService resolveMDCService(CamelContext camelContext) throws Exception {
+        CamelMDCService answer = camelContext.hasService(CamelMDCService.class);
+        if (answer == null) {
+            answer = camelContext.getRegistry().findSingleByType(CamelMDCService.class);
+        }
+        if (answer == null) {
+            answer = camelContext.getCamelContextExtension().getBootstrapFactoryFinder()
+                    .newInstance("mdc-service", CamelMDCService.class)
+                    .orElseThrow(() -> new IllegalArgumentException(
+                            "Cannot find OpenTelemetryTracer2 on classpath. Add camel-opentelemetry2 to classpath."));
```
davsclaus: Wrong name here

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -1851,6 +1869,28 @@ private void setOtel2Properties(
...
+
+        String loc = mdcProperties.getLocation("enabled");
+        Object obj = mdcProperties.remove("enabled");
+        if (ObjectHelper.isNotEmpty(obj)) {
+            autoConfiguredProperties.put(loc, "camel.mdc.enabled", obj.toString());
+        }
+        boolean enabled = obj != null ? CamelContextHelper.parseBoolean(camelContext, obj.toString()) : true;
+        if (enabled) {
+            CamelMDCService mdc = resolveMDCService(camelContext);
+            setPropertiesOnTarget(camelContext, mdc, mdcProperties, "camel.mdc.", failIfNotSet, true,
```
davsclaus, line 1885: The true value seems like its default enabled

core/camel-main/src/main/java/org/apache/camel/main/MdcConfigurationProperties.java
```diff
@@ -0,0 +1,104 @@
...
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    public String getCustomHeaders() {
+        return customHeaders;
+    }
+
+    /**
+     * Provide the headers you would like to use in the logging. Use `*` value to include all available headers
```
davsclaus: The '*' is not in the doc so maybe try without the quotes

core/camel-main/src/main/java/org/apache/camel/main/MdcConfigurationProperties.java
```diff
@@ -0,0 +1,104 @@
...
+    public void setCustomHeaders(String customHeaders) {
+        this.customHeaders = customHeaders;
+    }
+
+    public String getCustomProperties() {
+        return customProperties;
+    }
+
+    /**
+     * Provide the properties you would like to use in the logging. Use `*` value to include all available properties
```
davsclaus: properties is ambitious, favour using exchange properties

core/camel-main/src/main/docs/main.adoc
```diff
@@ -562,6 +562,18 @@ The camel.telemetryDev supports 4 options, which are listed below.
 |===
 
 
+=== Camel MDC configurations
+The camel.mdc supports 3 options, which are listed below.
+
+[width="100%",cols="2,5,^1,2",options="header"]
+|===
+| Name | Description | Default | Type
+| *camel.mdc.customHeaders* | Provide the headers you would like to use in the logging. Use value to include all available headers |  | String
```
davsclaus: Notice here how * is not in the doc (see lateR)

davsclaus:
Can you add unit test with that async component I pointed previously. Wiretap and Async Component does not execute fully similar.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
squakez: This is the signature of the method:
```
    public static boolean setPropertiesOnTarget(
            CamelContext context, Object target, OrderedLocationProperties properties,
            String optionPrefix, boolean failIfNotSet, boolean ignoreCase,
            OrderedLocationProperties autoConfiguredProperties)
```
the `true` value is for the `ignoreCase`. Or are you maybe referring to something different?

davsclaus: Ah okay, yeah so that was for ignore case

core/camel-main/src/main/java/org/apache/camel/main/MdcConfigurationProperties.java
squakez: It seems the way used in other properties is the `&#42;` character. I've used that one.

core/camel-main/src/main/java/org/apache/camel/main/MdcConfigurationProperties.java
squakez: Good one. Yes, I've changed to exchangeProperties and exchangeHeaders as well.

squakez:
> Can you add unit test with that async component I pointed previously. Wiretap and Async Component does not execute fully similar.

Okey, I've added the test and I had to refactor to manage an AsyncProcessor. Now it seems more consistent and the new thread spanned from the async processors are properly sharing MDC context as well.

PR #19324, opened 2025-09-25
CAMEL-8249 - camel-keycloak - A security component

oscerd:
# Description

In this first implementation I added only an AuthorizationPolicy based on Keycloak. I tried to add also a test-infra module to automate the testing with a container image, but creating realm, users and permissions via SDK, still requires some manual steps on the UI, so it won't work ootb.

Next step:

1. Support admin operation like createRealm, users etc. So the component won't only be a processor with security policy but also a producer only component
2. Support permissions based access to resources

I tested this against a docker instance and the IT test works fine.

# Target

# Tracking


docs/components/modules/others/nav.adoc
```diff
@@ -24,6 +24,8 @@
 ** xref:jasypt.adoc[Jasypt]
 ** xref:jfr.adoc[JFR]
 ** xref:jta.adoc[JTA]
+** xref:keycloak.adoc[Keycloak]
+** xref:keycloak-usage-examples.adoc[Keycloak Security Policy Usage Examples]
```
davsclaus, line 28: I wonder where this adoc file is ?

components/camel-keycloak/src/test/resources/log4j2.properties
```diff
@@ -0,0 +1,50 @@
+## ---------------------------------------------------------------------------
```
davsclaus, line 1: double license header

components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private KeycloakSecurityHelper() {
+        // Utility class
+    }
+
+    public static AccessToken parseAccessToken(String tokenString) throws VerificationException {
+        // Note: In production, you should verify the token signature with the public key
+        // For simplicity, we're just parsing without verification here
+        // You would typically do: TokenVerifier.create(tokenString, AccessToken.class).publicKey(publicKey).verify().getToken()
```
davsclaus, line 42: I guess those certificate and keys should be support as well, maybe in a future PR

oscerd:
For me this is compiling fine locally and the issue seems related to Atlassian only

```
[ERROR] Failed to execute goal on project camel-endpointdsl: Could not resolve dependencies for project org.apache.camel:camel-endpointdsl:jar:4.15.0-SNAPSHOT
[ERROR] dependency: org.apache.camel:camel-keycloak:jar:4.15.0-SNAPSHOT (provided)
[ERROR] 	G:org.apache.camel NOT allowed from atlassian (https://packages.atlassian.com/maven-external/, default, releases)
[ERROR] -> [Help 1]
```

I'll try to merge this and we could check the Jenkins CI on ASF infra.

components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityHelper.java
oscerd: Yes, it's a todo for the next PRs.

docs/components/modules/others/nav.adoc
oscerd: I removed that.

components/camel-keycloak/src/test/resources/log4j2.properties
oscerd: I'll fix that

PR #19373, opened 2025-09-30
Rename dsl.adoc to dsl-summary.adoc

rhaetor:
Rename file `-summary.adoc` for consistency.

# Description

All the other files of this overview type are named `-summary.adoc`. 

(For example: `components/camel-ai/src/main/docs/ai-summary.adoc`) 

It would make sense to have the same naming convention here.

Additionally, the symlink in the docs folder needs to reference the new name.


davsclaus (changes requested):
No its not the same

the summary files are for components only when they have a middle folder

rhaetor:
OK; I don't quite understand this.

Here is the AI page: 
https://camel.apache.org/components/4.14.x/ai-summary.html

For DSL, the page looks to have a similar function: 
https://camel.apache.org/components/4.14.x/others/dsl.html

Also, the navigation on the left hand for miscellaneous components arescrewed up: 
https://camel.apache.org/components/4.14.x/others/index.html

LangChain4j Tokenizer comes before Attachments. (I don't think LangChain4j Tokenizer  should be in the misc category at all.)

and shouldn't DSL item should be a group containing the DSL components? The nav outdents instead of indents the DSL components.

davsclaus:
Well sadly the website doc is a complex build of node stuff that copy and modify files and can be a pain to understand and maintain.

Take a look in docs/gulpfile.js and there are some template files `-nav.adoc.template` and somethow in the others then that tokenizer is computed with 3 levels instead of 2, and some other issues.

Help with this is welcome

davsclaus:
I fixed some of the 3 vs 2 indent in the others-nav file
https://github.com/apache/camel/commit/40a26fd8d67c5b52a3ff8d6d84cd884548c3e547

davsclaus:
Also its not so good to rename pages as this can cause build errors when there are xref that links to the old page from other projects like camel-k.

rhaetor:
OK; still trying to understand how this is supposed to work. 

It looks like you removed group properties on the .adocs and edited the Nav file manually?

Isn't the nav generated automatically from the gulpfile? 

If so, shouldn't the group property still be there. I'm barely able to read the gulpfile, so can't really tell what is going on there.

davsclaus:
no the nav file is generated, but the group should only be there for regular components

PR #19377, opened 2025-09-30
Update index.adoc

antbewes:
Without quotes I was getting error on windows:
PS C:\projects\learn\first-camel-integration> mvn camel:run -Dcamel.main.durationMaxMessages=2 -->
[ERROR] Unknown lifecycle phase ".main.durationMaxMessages=2". ...

# Description

# Target

# Tracking

### DIFF
--- a/docs/main/modules/getting-started/pages/index.adoc
+++ b/docs/main/modules/getting-started/pages/index.adoc
@@ -93,7 +93,7 @@ To run the project you can run the following command:
 
 [source,bash]
 ----
-mvn camel:run -Dcamel.main.durationMaxMessages=2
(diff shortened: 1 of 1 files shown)


davsclaus:
I think its better to add this as a new example for windows users. Also do you really need to quote both the key and the value, can it not just be "aaa.bbb=123"

davsclaus:
okay since this is a intro then lets change the run command to not have that extra command that is just a bit confusing. Instead we tell the user that they can exit via ctrl + c as any other Java application.

davsclaus:
https://github.com/apache/camel/commit/b49d5f259bc462ac857f8d7373cb3cf88928941f

PR #19391, opened 2025-10-01
CAMEL-22475: Camel LangChain4j Agent: custom tools feature

zbendhiba:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent-api/src/main/java/org/apache/camel/component/langchain4j/agent/api/AgentConfiguration.java
+++ b/components/camel-ai/camel-langchain4j-agent-api/src/main/java/org/apache/camel/component/langchain4j/agent/api/AgentConfiguration.java
@@ -57,6 +57,7 @@ public class AgentConfiguration {
     private RetrievalAugmentor retrievalAugmentor;
     private List<Class<?>> inputGuardrailClasses;
     private List<Class<?>> outputGuardrailClasses;
+    private List<Object> customTools; // Custom LangChain4j tools
 
     /**
      * Gets the configured chat model.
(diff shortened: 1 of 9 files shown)


components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConfiguration.java
```diff
@@ -88,4 +93,17 @@ public AgentFactory getAgentFactory() {
...
+     *
+     * @return the custom tools (instances of classes with @Tool methods)
+     */
+    public List<Object> getCustomTools() {
+        return customTools;
+    }
+
+    public void setCustomTools(List<Object> customTools) {
+        this.customTools = customTools;
+    }
```
orpiske: I think this is the only thing that we don't need. As the customTools seem to come from the agent configuration, I believe these are unused. Right?

orpiske (changes requested):

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConfiguration.java
zbendhiba: oh yeah, I started here and then moved to the factory thing. could catch. let me fix this

orpiske (changes requested):
Sometimes IntelliJ is too trigger happy ... I think this is great! I believe the change on `LangChain4jAgentConfiguration`, but everything else looks like a great addition to the agent component.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConfiguration.java
zbendhiba: fixed !!

PR #19395, opened 2025-10-01
CAMEL-22466 Clean up spring-boot / quarkus version references

cunningt:
# Description

`dsl/camel-jbang/camel-jbang-core/src/main/java-templates/org/apache/camel/dsl/jbang/core/common/RuntimeType.java
` has hardcoded versions of quarkus and spring-boot included in it that are hard to maintain - they won't be picked up by dependabot updates.

This is an attempt to manage the spring-boot / quarkus versions for RuntimeType.java consistently with how camel manages other third party versions - add them to parent/pom.xml, and then filter the versions into the source.

Looking for some feedback here - this compiles with no issues for me locally - templating-maven-plugin is executed pre-compile in maven, but I am not sure how IDEs will handle this.   References to RuntimeType might be marked as missing, I'm not sure if IDEs know to look in the `java-templates` directory for sources.      Another strategy might be to filter the versions into a property file and then read that file in RuntimeType for the quarkus/spring-boot versions - I could change to that if that's preferable?

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-core/pom.xml
@@ -215,6 +215,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>templating-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>filter-src</id>
+                        <goals>
+                            <goal>filter-sources</goal>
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

davsclaus:
There is also a SB version in the pom in tooling/maven/camel-repackager-maven-plugin - this can be removed so it uses the version from parent pom now

Croway (approved):

PR #19402, opened 2025-10-02
CAMEL-22485: add a way to clean up resources

orpiske:


core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultContextServiceLoaderPlugin.java
```diff
@@ -58,13 +59,22 @@ public class DefaultContextServiceLoaderPlugin extends ServiceSupport implements
...
                 camelContext.getApplicationContextClassLoader());
         for (ContextServicePlugin plugin : contextServicePlugins) {
             plugin.load(camelContext);
         }
     }
 
+    @Override
+    protected void doStop() throws Exception {
+        if (contextServicePlugins != null) {
+            for (ContextServicePlugin plugin : contextServicePlugins) {
```
davsclaus, line 82: You may want to do a try .. catch and LOG warn about ignoring this exception so all the plugins can do their stop logic as that is more important than if one fails then the others are not unloaded.

core/camel-api/src/main/java/org/apache/camel/spi/ContextServicePlugin.java
```diff
@@ -66,4 +66,11 @@ public interface ContextServicePlugin {
      * @param camelContext the CamelContext being initialized, never {@code null}
      */
     void load(CamelContext camelContext);
+
+    /**
+     * Called during CamelContext stop. Use it to free allocated resources.
+     */
+    default void unload() {
```
davsclaus: Maybe add CamelContext so its same as load

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultContextServiceLoaderPlugin.java
orpiske: Good idea. Let me do that!

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultContextServiceLoaderPlugin.java
```diff
@@ -58,13 +63,27 @@ public class DefaultContextServiceLoaderPlugin extends ServiceSupport implements
      */
     @Override
     protected void doStart() throws Exception {
-        ServiceLoader<ContextServicePlugin> contextServicePlugins = ServiceLoader.load(ContextServicePlugin.class,
+        contextServicePlugins = ServiceLoader.load(ContextServicePlugin.class,
                 camelContext.getApplicationContextClassLoader());
         for (ContextServicePlugin plugin : contextServicePlugins) {
             plugin.load(camelContext);
```
orpiske: I'm thinking we should do the same logic for the unload.

davsclaus (approved):

orpiske:
Failing due to some formatting issues. I will fix after the my Czech class

PR #19403, opened 2025-10-02
Backport CAMEL-22485

orpiske:


oscerd (approved):

davsclaus:
Is this PR ready

orpiske:
> Is this PR ready

Yeah. I think it is fine to merge. We had an issue on CEQ, but we could find a work-around, so it should be OK for now.

PR #19446, opened 2025-10-06
New JMS component option 'replyCorrelationProperty'.

marcinwieckowski:
Adds a new option on the JMS Component configuration: 'replyCorrelationProperty'.

Difference between this and 'correlationProperty'" is that 'correlationProperty' tells which request property holds the correlation id value, and it does not affect the selector for the reply (JMSCorrelationID=<correlation id>), while 'replyCorrelationProperty' tells which reply property will hold the correlation id value, and it does affect the selector for the reply (<replyCorrelationProperty>=<correlation id>).

Use case: reply provider has custom correlation behaviour where it uses custom headers (not JMSMessageID and JMSCorrelationID) for correlation.


marcinwieckowski:
Sorry, I had some problems with generating all the necessary auto-generated files. Now it should be ok.

davsclaus (approved):

davsclaus:
CAMEL-22500

davsclaus:
Thanks for the PR. I have manually merged this. Yeah its a bit difficult due to those source files are regenerated in various places. So closing this PR as the code has been manually merged.

marcinwieckowski:
Thank you.

PR #19468, opened 2025-10-07
CAMEL-22503 - Camel-Docling: Support Docling-serve and being able to invoke the service as API

oscerd:
# Description

# Target

# Tracking

### DIFF
--- a/test-infra/camel-test-infra-docling/src/main/java/org/apache/camel/test/infra/docling/services/DoclingInfraService.java
+++ b/test-infra/camel-test-infra-docling/src/main/java/org/apache/camel/test/infra/docling/services/DoclingInfraService.java
@@ -23,5 +23,5 @@ import org.apache.camel.test.infra.common.services.InfrastructureService;
  */
 public interface DoclingInfraService extends InfrastructureService {
 
-    String getDoclingServerUrl();
+    String doclingServerUrl();
 }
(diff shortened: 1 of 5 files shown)


davsclaus (approved):

test-infra/camel-test-infra-docling/src/main/java/org/apache/camel/test/infra/docling/services/DoclingInfraService.java
```diff
@@ -0,0 +1,27 @@
...
+package org.apache.camel.test.infra.docling.services;
+
+import org.apache.camel.test.infra.common.services.InfrastructureService;
+
+/**
+ * Test infra service for Docling
+ */
+public interface DoclingInfraService extends InfrastructureService {
+
+    String getDoclingServerUrl();
```
Croway: can you rename it to `doclingServeUrl()`? This way tools can map it easily

oscerd: Done.

Croway (approved):

oscerd:
The node is not lucky today. Btw I tested this, I think one problem could be the docling-serve image has something like 5 Gb layer (layer) in its image. So maybe we should disable the IT tests on GH action. Let's see on ASF Jenkins.

PR #21534, opened 2026-02-18
CAMEL-21975: Jbang diagram command to visualize camel routes

vignesh-manel:
JIRA: [CAMEL-21975](https://issues.apache.org/jira/browse/CAMEL-21975)

# Description

Added Playwright-based PNG export for camel diagram via hawtio and Jolokia

# Target

# Tracking


davsclaus:
Thanks for helping with jbang. Can you make this into its own plugin as this should not be in the core.

You can look at one of the existing pluigins such as the new validate
https://github.com/apache/camel/tree/main/dsl/camel-jbang/camel-jbang-plugin-validate

parent/pom.xml
```diff
@@ -450,6 +450,7 @@
         <pgjdbc-ng-driver-version>0.8.9</pgjdbc-ng-driver-version>
         <picocli-version>4.7.7</picocli-version>
         <pinecone-client-version>3.1.0</pinecone-client-version>
+        <playwright-version>1.43.0</playwright-version>
```
Croway: This version is quite old, can you use the latest 1.58.0?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+import org.apache.camel.dsl.jbang.core.commands.process.Hawtio;
+import org.apache.camel.dsl.jbang.core.commands.process.Jolokia;
+import org.apache.camel.dsl.jbang.core.commands.process.StopProcess;
+import org.apache.camel.util.FileUtil;
+import org.apache.camel.util.StopWatch;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "diagram", description = "Visualize Camel routes using Hawtio", sortOptions = false,
+                     showDefaultValues = true)
+public class Diagram extends CamelCommand {
```
Croway: This class is quite big, can you split it in smaller classes?

There is javascript code that is hard to read as it is, please put the js code in the resource folder in .js files.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+                if (code >= 200 && code < 500) {
+                    return;
+                }
+            } catch (Exception e) {
+                // ignore until timeout
+            }
+            if (hawtioProcess != null && !hawtioProcess.process.isAlive()) {
+                throw new IllegalStateException("Hawtio terminated before startup." + formatHawtioOutput(hawtioProcess.output));
+            }
+            Thread.sleep(500);
```
Croway: This is fragile, would it be possible to implement some kind of polling healthcheck instead?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+    }
+
+    private boolean waitForJolokia(String jolokiaUrl, long timeoutMs) {
+        StopWatch watch = new StopWatch();
+        while (watch.taken() < timeoutMs) {
+            if (isJolokiaAvailable(jolokiaUrl)) {
+                return true;
+            }
+            try {
+                Thread.sleep(200);
```
Croway: same

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+        }
+    }
+
+    private void selectRoutesFolder(Page page) {
+        try {
+            waitForSelector(page, "#camel-tree-view .pf-v5-c-tree-view__node", playwrightTimeout / 6);
+            Locator expandAll = page.locator("#camel-tree-view button:has-text(\"Expand all\")");
+            if (expandAll.count() > 0) {
+                expandAll.first().click();
+                page.waitForTimeout(300);
```
Croway: This is fragile as well, there may be alternatives like waitForSelector, or waitForFunction, waitForResponse that can be used instead

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+        prepareDiagramForScreenshot(page);
+    }
+
+    private boolean selectRouteInTree(Page page, String routeId) {
+        try {
+            waitForSelector(page, "#camel-tree-view .pf-v5-c-tree-view__node", playwrightTimeout / 6);
+            Locator expandAll = page.locator("#camel-tree-view button:has-text(\"Expand all\")");
+            if (expandAll.count() > 0) {
+                expandAll.first().click();
+                page.waitForTimeout(300);
```
Croway: same

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+            }
+            Locator search = page.locator("#input-search");
+            if (search.count() > 0) {
+                search.fill(routeId);
+                page.waitForTimeout(300);
+            }
```
Croway: same

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Diagram.java
```diff
@@ -0,0 +1,945 @@
...
+            } catch (com.microsoft.playwright.PlaywrightException e) {
+                printer().printErr("Diagram element changed while rendering, capturing full page instead: " + e.getMessage());
+            }
+        }
+        page.screenshot(new ScreenshotOptions().setPath(outputPath).setFullPage(true));
+    }
+
+    private void normalizeDiagramLayout(Page page) {
+        try {
+            page.evaluate("(() => {"
```
Croway: move to its own js file

orpiske (commented):
Nice idea, I'm but I'm under the impression this would be better done as a plugin. Let's see what others think about it ...

vignesh-manel:
@davsclaus @Croway @orpiske Refactored as plugin, please review

Croway:
Nice @vignesh-manel , were you able to test everything locally? can you provide some commands and outputs?

davsclaus:
there are changes to camel-launcher / camel-jbang run etc - that is of course a bit more worring since this plugin should not break existing behaviour. 

And what web browsers and OS does this work on ? Is it only google chrome ?

vignesh-manel:
> Nice @vignesh-manel , were you able to test everything locally? can you provide some commands and outputs?

I tested with the jar locally using below command from camel-launcher directory, it requires a browser executable which is compatible with chromium

java -jar target/camel-launcher-4.19.0-SNAPSHOT.jar diagram ../camel-jbang-core/src/test/resources/hello.yaml --output=hello.png --playwright-browser-path="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"

if --output is not specified it will open the hawtio console, where the integration can be viewed
if --keep-running option is provided, the integration will keep running in the background
--name=hello can be provided  to capture diagram of integration which is already running in the background
--route-id=xyz can be provided if a there are multiple routes to capture diagram of specific route

vignesh-manel:
> there are changes to camel-launcher / camel-jbang run etc - that is of course a bit more worring since this plugin should not break existing behaviour.
> 
> And what web browsers and OS does this work on ? Is it only google chrome ?

I tried to simplify the changes, right now any chromium based browser is supported, I have tested it only in MacOS

davsclaus:
> > Nice @vignesh-manel , were you able to test everything locally? can you provide some commands and outputs?
> 
> I tested with the jar locally using below command from camel-launcher directory, it requires a browser executable which is compatible with chromium
> 
> java -jar target/camel-launcher-4.19.0-SNAPSHOT.jar diagram ../camel-jbang-core/src/test/resources/hello.yaml --output=hello.png --playwright-browser-path="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
> 
> if --output is not specified it will open the hawtio console, where the integration can be viewed if --keep-running option is provided, the integration will keep running in the background --name=hello can be provided to capture diagram of integration which is already running in the background --route-id=xyz can be provided if a there are multiple routes to capture diagram of specific route

I think this information would be good to add to the documentation, with that little example so others better can understand how to use this feature

docs/user-manual/modules/ROOT/pages/camel-jbang.adoc
```diff
@@ -2484,6 +2484,34 @@ then Camel will attempt to uninstall the Jolokia JVM Agent. However, this may no
 able to do this always, because the JVM is being terminated which can prevent camel-jbang
 from doing JVM process communication to the running Camel integration.
 
+=== Visualizing routes with diagram
+
+The `camel diagram` command uses Hawtio to visualize Camel routes. You can run route files in the background and open Hawtio in a browser:
```
davsclaus: the diagram is a plugin you need to install first via `camel plugin add diagram`

also I think there is some code where we have a known list of plugins we ship so diagram should be added there also

docs/user-manual/modules/ROOT/pages/camel-jbang.adoc
```diff
@@ -2484,6 +2484,34 @@ then Camel will attempt to uninstall the Jolokia JVM Agent. However, this may no
...
+By default Camel JBang stops the background integration when you exit Hawtio; use `--keep-running` to leave it running.
+
+You can export a PNG snapshot using Playwright (requires a Chromium browser binary and executable path configured):
+
+[source,bash]
+----
+camel diagram routes.yaml --output=routes.png --playwright-browser-path=/path/to/chrome
+----
+
+When using `--output`, Hawtio runs headless and the PNG is captured automatically.
```
davsclaus, line 2539: A good idea is to add that actual long command you typed to run on your mac, as mac users need to point to that /Content/XXX or whatever inside those apps - so users would maybe struggle if they dont have a real working example to see

davsclaus (approved):

orpiske (approved):

davsclaus:
btw I read on social media about a new zig rewritten browser that was headless for AI agents only - it was much faster to launch and supported playwright.

docs/user-manual/modules/ROOT/pages/camel-jbang.adoc
vignesh-manel: added

docs/user-manual/modules/ROOT/pages/camel-jbang.adoc
vignesh-manel: added

vignesh-manel:
> btw I read on social media about a new zig rewritten browser that was headless for AI agents only - it was much faster to launch and supported playwright.

Is it this? https://github.com/lightpanda-io/browser

gnodet:
## Code Review

Great feature — visualizing Camel routes directly from `camel` CLI is very useful. Here's detailed feedback on the implementation:

### Dead code

`DiagramPage.java` has several unused/empty methods:
- `selectRoutesFolder()` — private, never called
- `waitForDiagramAssets()` — private, never called
- `prepareDiagramForScreenshot()` — called from two places but has an empty body

These should be removed or connected to the call paths where they're intended.

### `LauncherHelper.normalizeJarPath()` returns `null` — potential NPE

The new `normalizeJarPath()` returns `null` when the path doesn't contain `.jar`. Callers at lines 68 and 162 in `getLauncherJarPath()` previously returned a valid path string but now may return `null` if running from an exploded classpath directory. This is a behavioral regression that could cause NPEs downstream.

### Duplicated Jolokia logic

`JolokiaAttacher` directly uses `com.sun.tools.attach.VirtualMachine` and reimplements PID-finding logic. The existing `Jolokia` command in `camel-jbang-core` already handles this with the proper Jolokia client abstractions and `ProcessBaseCommand.findPids()`. Consider reusing the existing code rather than duplicating it. If the plugin module can't depend on `ProcessBaseCommand`, consider extracting the shared logic into a utility.

### Minimal test coverage

The 5 tests only cover basic parameter binding. There are no tests for:
- `LauncherHelper.normalizeJarPath()` — a utility with several branches now used in critical paths
- `LauncherHelper.findJolokiaAgentJar()` — filesystem scanning
- `JolokiaAttacher` — PID resolution, port finding, agent loading
- `DiagramPngExporter` — the core export workflow
- `PluginHelper.scanLauncherForPlugins()` — nested JAR scanning

At minimum, `normalizeJarPath()` needs unit tests since it's in shared core code affecting all JBang commands.

### Playwright is a heavyweight dependency

Playwright (~50MB+ with browser binaries) is a compile-time dependency of the entire plugin module but only needed for `--output` PNG export. When bundled into the launcher fat JAR, it increases download size for all users. Consider making it a runtime dependency downloaded on demand (similar to how Hawtio downloads its dependencies via `MavenDependencyDownloader`).

### Minor issues

- **`arity = "0..9"`** limits file arguments to 9 — other JBang commands use `"0..*"` for unlimited files. Is this intentional?
- **`new Random().nextLong()`** for log file naming can produce negative numbers and has a NOSONAR suppression. `Files.createTempFile("diagram-run", ".log")` would be cleaner.
- **`--openUrl`** uses camelCase while all other options use kebab-case (`--keep-running`, `--route-id`). Consider adding a kebab-case alias `--open-url`.
- **HTTP 4xx treated as success** in `waitForEndpoint()` — `code >= 200 && code < 500 && code != 404` accepts 401/403/405 as "endpoint available." Should be documented.
- **`URLClassLoader` leak** in `PluginHelper.scanLauncherForPlugins()` — the comment explains the rationale but consider storing loaders for shutdown cleanup.
- **JS code tightly coupled to Hawtio internals** — `diagram-scripts.js` uses PatternFly v5 CSS classes and React Flow internals. Any Hawtio UI update could break this silently. Consider documenting which Hawtio version this was tested against.

### Summary

| Priority | Issue |
|----------|-------|
| High | Dead code (unused/empty methods in DiagramPage) |
| High | `normalizeJarPath` returns null — behavioral regression |
| High | Duplicated Jolokia attach logic |
| High | Minimal test coverage for shared code changes |
| Medium | Playwright heavyweight compile dependency |
| Medium | File arity limited to 9 |
| Low | Random filenames, option naming, HTTP 4xx handling, classloader leak |

vignesh-manel:
> ## Code Review
> Great feature — visualizing Camel routes directly from `camel` CLI is very useful. Here's detailed feedback on the implementation:
> 
> ### Dead code
> `DiagramPage.java` has several unused/empty methods:
> 
> * `selectRoutesFolder()` — private, never called
> * `waitForDiagramAssets()` — private, never called
> * `prepareDiagramForScreenshot()` — called from two places but has an empty body
> 
> These should be removed or connected to the call paths where they're intended.
> 
> ### `LauncherHelper.normalizeJarPath()` returns `null` — potential NPE
> The new `normalizeJarPath()` returns `null` when the path doesn't contain `.jar`. Callers at lines 68 and 162 in `getLauncherJarPath()` previously returned a valid path string but now may return `null` if running from an exploded classpath directory. This is a behavioral regression that could cause NPEs downstream.
> 
> ### Duplicated Jolokia logic
> `JolokiaAttacher` directly uses `com.sun.tools.attach.VirtualMachine` and reimplements PID-finding logic. The existing `Jolokia` command in `camel-jbang-core` already handles this with the proper Jolokia client abstractions and `ProcessBaseCommand.findPids()`. Consider reusing the existing code rather than duplicating it. If the plugin module can't depend on `ProcessBaseCommand`, consider extracting the shared logic into a utility.
> 
> ### Minimal test coverage
> The 5 tests only cover basic parameter binding. There are no tests for:
> 
> * `LauncherHelper.normalizeJarPath()` — a utility with several branches now used in critical paths
> * `LauncherHelper.findJolokiaAgentJar()` — filesystem scanning
> * `JolokiaAttacher` — PID resolution, port finding, agent loading
> * `DiagramPngExporter` — the core export workflow
> * `PluginHelper.scanLauncherForPlugins()` — nested JAR scanning
> 
> At minimum, `normalizeJarPath()` needs unit tests since it's in shared core code affecting all JBang commands.
> 
> ### Playwright is a heavyweight dependency
> Playwright (~50MB+ with browser binaries) is a compile-time dependency of the entire plugin module but only needed for `--output` PNG export. When bundled into the launcher fat JAR, it increases download size for all users. Consider making it a runtime dependency downloaded on demand (similar to how Hawtio downloads its dependencies via `MavenDependencyDownloader`).
> 
> ### Minor issues
> * **`arity = "0..9"`** limits file arguments to 9 — other JBang commands use `"0..*"` for unlimited files. Is this intentional?
> * **`new Random().nextLong()`** for log file naming can produce negative numbers and has a NOSONAR suppression. `Files.createTempFile("diagram-run", ".log")` would be cleaner.
> * **`--openUrl`** uses camelCase while all other options use kebab-case (`--keep-running`, `--route-id`). Consider adding a kebab-case alias `--open-url`.
> * **HTTP 4xx treated as success** in `waitForEndpoint()` — `code >= 200 && code < 500 && code != 404` accepts 401/403/405 as "endpoint available." Should be documented.
> * **`URLClassLoader` leak** in `PluginHelper.scanLauncherForPlugins()` — the comment explains the rationale but consider storing loaders for shutdown cleanup.
> * **JS code tightly coupled to Hawtio internals** — `diagram-scripts.js` uses PatternFly v5 CSS classes and React Flow internals. Any Hawtio UI update could break this silently. Consider documenting which Hawtio version this was tested against.
> 
> ### Summary
> Priority	Issue
> High	Dead code (unused/empty methods in DiagramPage)
> High	`normalizeJarPath` returns null — behavioral regression
> High	Duplicated Jolokia attach logic
> High	Minimal test coverage for shared code changes
> Medium	Playwright heavyweight compile dependency
> Medium	File arity limited to 9
> Low	Random filenames, option naming, HTTP 4xx handling, classloader leak

@gnodet could you please re-review, I have removed the dead code. Changed playwright to runtime dependency which will be downloaded during runtime. Added test for LauncherHelper, addressed the minor issues, however duplicate changes in JolokiaAttacher was not addressed to avoid any impact on existing jolokia functionality

gnodet (commented):

Did a thorough review. Previous reviewer concerns (Croway, davsclaus) have been addressed — good work restructuring as a plugin.

A few remaining items before merge:

1. **Dead code**: `embeddedJolokia` parameter is always `false` — the field, constructor parameter, and `if (embeddedJolokia)` branch in `attachJolokia()` can be removed
2. **Dead code**: `isJolokiaAvailable()` method in `DiagramPngExporter` is defined but never called
3. **Generated docs incomplete**: The `--timeout` option is defined in `DiagramCommand.java` but missing from the generated `camel-jbang-diagram.adoc`. Please regenerate: `mvn install -B -pl dsl/camel-jbang/camel-jbang-core -DskipTests` and commit the updated metadata/docs
4. **Unrelated change**: `DependencyUpdateTest` cleanup (changing temp dir base path) is unrelated to the diagram feature — consider splitting it out or reverting

None of these are blocking but #3 will cause CI failure. Please address and I'll merge.

vignesh-manel:
@gnodet I have removed dead code and updated docs , the unrelated test change was to fix flaky test, after which the build succedded, so I haven't reverted it, but now it looks like there are some other flaky tests which are failing

davsclaus:
there is another approach for a diagram command just merged that does not rely on using web browsers, give that a try and look. 

https://github.com/apache/camel/commit/6b103ea783357819372c960b6c7a3ab680c33bb1

davsclaus:
closing this as we implemented this without using hawtio or the likes. There is a new camel-diagram component that renders and we have a jbang command for this as well. And more to come there.

PR #22158, opened 2026-03-20
CAMEL-23239: Add camel-state-store component with pluggable key-value store

gnodet:
## Summary

Adds a new `camel-state-store` component that provides a unified key-value store API with pluggable backends. This is useful for caching, session state, and scenarios where you need a simple object store (similar to MuleSoft's Object Store).

### Modules

| Module | Description |
|--------|-------------|
| `camel-state-store` | Core component + in-memory backend (default) |
| `camel-state-store-caffeine` | High-performance local cache backend |
| `camel-state-store-redis` | Distributed store via Redisson |
| `camel-state-store-infinispan` | Distributed cache via Hot Rod |

### Operations

`put`, `putIfAbsent`, `get`, `delete`, `contains`, `keys`, `size`, `clear` — with optional per-entry TTL.

### Key Design Points

- **Backend swappability**: develop with in-memory, deploy with Redis/Infinispan — no route changes needed
- **Auto-discovery**: a single `StateStoreBackend` bean in the registry is auto-detected
- **Properties-based configuration**: backends configurable via `camel.beans.*` properties (no Java required)
- `StateStoreBackend` extends `org.apache.camel.Service` for proper lifecycle management

## Changes since last review

All 12 findings from review [#4927079320](https://github.com/apache/camel/pull/22158#pullrequestreview-4927079320) plus the version comment have been addressed:

1. **[Critical] MojoHelper registration** — registered all 4 sub-modules (`camel-state-store`, `camel-state-store-caffeine`, `camel-state-store-redis`, `camel-state-store-infinispan`) instead of just the core module
2. **[Medium] `StateStoreComponent.doStop()` resource leak** — each `backend.stop()` is now wrapped in try-catch, logs warning on failure, and rethrows the first exception after attempting all stops
3. **[Medium] Case-sensitive operation lookup** — added case-insensitive fallback with user-friendly error message listing valid operations
4. **[Medium] Infinispan `Thread.sleep` retry loop** — removed entire retry loop with exponential backoff from `start()`; now fails fast
5. **[Medium] Infinispan IT `RemoteCacheManager` leak** — added `testCacheManager` field with `@AfterEach cleanUp()` in both IT classes
6. **[Medium] Adoc include directives misplaced** — moved `component-configure-options`, `component-endpoint-options`, and `component-endpoint-headers` includes into their correct `START/END` blocks
7. **[Low] Redis/Infinispan managed flag not reset in `stop()`** — both backends now reset `managedRedisson`/`managedCacheManager` to `false` in `stop()`
8. **[Low] Silent backend discard** — explicit backend path now uses `putIfAbsent` and logs WARN when a store name already has a different backend
9. **[Low] No fallback log** — added INFO log when no `StateStoreBackend` found (falling back to in-memory) and DEBUG log for auto-discovery
10. **[Low] InMemoryStateStoreBackend lazy eviction** — `keys()` and `size()` now evict expired entries via `removeIf()` instead of just filtering them
11. **[Low] Redundant caffeine version** — removed explicit `${caffeine-version}` from caffeine backend pom.xml (managed by parent BOM)
12. **[Low] Test gaps** — added: Redis IT delete verification + delete non-existent key, Infinispan IT clear + negative contains + delete non-existent key, core module invalid operation test + case-insensitive operation test + delete non-existent key test
13. **Version update** — rebased on main (4.23.0-SNAPSHOT), updated all version references, `@since` tags, and generated files

## Test plan

- [x] Core unit tests (in-memory backend): put/get, delete, contains, keys, size, clear, putIfAbsent, TTL expiry, operation via header, case-insensitive operations, invalid operation, delete non-existent key, missing key/operation error, auto-discovery, multi-backend fallback
- [x] Caffeine unit tests: put/get, TTL, eviction, properties-based configuration
- [x] Redis integration tests (Testcontainers): put/get, delete with verification, contains (positive + negative), keys/size, putIfAbsent, clear, TTL, delete non-existent key
- [x] Infinispan integration tests (Testcontainers): put/get, delete with verification, contains (positive + negative), keys/size, putIfAbsent, clear, TTL, delete non-existent key, properties-based configuration


orpiske:
Hm,  this issue doesn't seem to exist on ASF Jira.

claudio4j:
The title refers to the wrong issue, CAMEL-23228 is "Add DataWeave to DataSonnet transpiler in camel-jbang".

apupier:
the commit message also needs to be updated with the correct jira issue number

apupier (commented):
Can you elaborate on the difference between these components and the existing Camel caffeine cache component? https://camel.apache.org/components/4.18.x/caffeine-cache-component.html

gnodet:
Thanks for the reviews! I've pushed an update:

- **Fixed the JIRA reference**: commit message now correctly references CAMEL-23239 (was CAMEL-23228, a copy-paste error)
- **Updated the PR description** with a detailed motivation section explaining the rationale and how this relates to existing components

@apupier — regarding the difference with `camel-caffeine-cache`: the state-store component is not a replacement for the technology-specific components. It follows the same pattern as `camel-sql` (generic SQL over any JDBC database) or `camel-jms` (generic messaging over any JMS provider). The key benefit is **ease of use and backend portability** — users choose the capability ("I need a key-value store") and can swap backends without changing route logic. The technology-specific components remain the right choice when users need the full feature set of a particular product (queries, pub/sub, statistics, etc.). See the updated PR description for a detailed comparison table.

gnodet:
### Design note: `StateStoreBackend` interface location

We considered moving the `StateStoreBackend` interface to `camel-api` so it could serve as a shared abstraction alongside the existing SPIs (`IdempotentRepository`, `AggregationRepository`, `StateRepository`). After analysis, we decided against it:

- The existing `camel-api` SPIs are there because **core EIPs consume them** (idempotent consumer, aggregator, etc.). `StateStoreBackend` is only consumed by the `camel-state-store` component itself — nothing in core needs it.
- The existing SPIs have different semantics that don't map cleanly to a generic key-value store: `IdempotentRepository` has two-phase confirm/rollback, `AggregationRepository` stores `Exchange` objects requiring serialization, and technology-specific implementations often leverage native features (JDBC transactions, Cassandra TTL, Kafka offsets).
- Moving it to `camel-api` would add API surface with stricter stability guarantees for no practical benefit today.

If a future core EIP needs a generic key-value store abstraction, it can be promoted then.

components/camel-state-store/camel-state-store/src/main/docs/state-store-component.adoc
```diff
@@ -0,0 +1,331 @@
...
+    <artifactId>camel-state-store-infinispan</artifactId>
+</dependency>
+----
+
+Configure and register:
+
+[source,java]
+----
+@BindToRegistry("infinispanBackend")
+public InfinispanStateStoreBackend infinispan() {
```
Croway, line 345: @gnodet is this kind of configuration doable only via java beans? would it be possible to provide this configuration via properties?

gnodet: Good point! I've pushed an update that addresses this:

**Property-based configuration** — backends can now be fully configured via `application.properties` using Camel's `camel.beans.*` syntax, without writing any Java code:

```properties
# Caffeine example
camel.beans.caffeineBackend = #class:org.apache.camel.component.statestore.caffeine.CaffeineStateStoreBackend
camel.beans.caffeineBackend.maximumSize = 50000
```

```properties
# Redis example
camel.beans.redisBackend = #class:org.apache.camel.component.statestore.redis.RedisStateStoreBackend
camel.beans.redisBackend.redisUrl = redis://myhost:6379
camel.beans.redisBackend.mapName = my-app-state
```

```properties
# Infinispan example
camel.beans.infinispanBackend = #class:org.apache.camel.component.statestore.infinispan.InfinispanStateStoreBackend
camel.beans.infinispanBackend.hosts = myhost:11222
camel.beans.infinispanBackend.cacheName = my-cache
```

**Auto-discovery** — if a single `StateStoreBackend` bean is found in the registry (whether registered via Java or via properties), it is automatically used by all `state-store` endpoints. No `backend=#beanName` reference needed:

```yaml
- route:
    from:
      uri: direct:store
    steps:
      - setHeader:
          name: CamelStateStoreKey
          constant: myKey
      - to:
          uri: state-store:myStore?operation=put
```

Both features are covered by new tests (`CaffeineStateStorePropertiesTest` and `StateStoreAutoDiscoveryTest`).

components/camel-state-store/camel-state-store/src/main/java/org/apache/camel/component/statestore/StateStoreProducer.java
```diff
@@ -0,0 +1,115 @@
...
+                String key = requireKey(message);
+                Object value = message.getBody();
+                Object previous = backend.put(key, value, ttl);
+                message.setBody(previous);
+            }
+            case putIfAbsent -> {
+                String key = requireKey(message);
+                Object value = message.getBody();
+                Object existing = backend.putIfAbsent(key, value, ttl);
+                message.setBody(existing);
```
apupier, line 62: maybe normal but I was surprised when reading it that when performing and operation to store a state, we change the message body with the previous value

apupier: I'm even wondering why modifying the body when setting something in the state store?

gnodet: The return-previous-value behavior follows `java.util.Map.put()` / `Map.putIfAbsent()` semantics:

- **`put`** — returns the previous value (or `null` if key was new), same as `Map.put()`
- **`putIfAbsent`** — returns the existing value if key already existed (meaning nothing was stored), or `null` if the value was stored successfully

This is intentional — it's the only way to communicate the result back to the route, and it enables patterns like idempotent deduplication:

```java
from("kafka:orders")
    .setHeader(StateStoreConstants.KEY, simple("${header.orderId}"))
    .to("state-store:processed?operation=putIfAbsent")
    .choice()
        .when(body().isNotNull())  // non-null means key already existed = duplicate
            .log("Duplicate order, skipping").stop()
    .end()
    .to("direct:process-order");
```

This is consistent with how `camel-caffeine-cache` handles it (the `PUT` action also returns the previous value in the body).

apupier (approved):
Seems fine but I'm really not an expert on this piece of code and what it applies. i think it would be nice to have someone more experimented reviewing it.

components/camel-state-store/camel-state-store-infinispan/src/main/java/org/apache/camel/component/statestore/infinispan/InfinispanStateStoreBackend.java
```diff
@@ -0,0 +1,153 @@
...
+        Exception lastException = null;
+        for (int i = 0; i < 10; i++) {
+            try {
+                cache = cacheManager.administration().getOrCreateCache(cacheName, (String) null);
+                return;
+            } catch (Exception e) {
+                lastException = e;
+                LOG.warn("Failed to access cache '{}' (attempt {}/10): {}", cacheName, i + 1, e.getMessage());
+                try {
+                    Thread.sleep(1000);
```
Croway: what about using resilience4j? would it be overkill?

Croway:
```
# Caffeine example
camel.beans.caffeineBackend = #class:org.apache.camel.component.statestore.caffeine.CaffeineStateStoreBackend
camel.beans.caffeineBackend.maximumSize = 50000

# Redis example
camel.beans.redisBackend = #class:org.apache.camel.component.statestore.redis.RedisStateStoreBackend
camel.beans.redisBackend.redisUrl = redis://myhost:6379
camel.beans.redisBackend.mapName = my-app-state

 # Infinispan example
camel.beans.infinispanBackend = #class:org.apache.camel.component.statestore.infinispan.InfinispanStateStoreBackend
camel.beans.infinispanBackend.hosts = myhost:11222
camel.beans.infinispanBackend.cacheName = my-cache
```

This sounds fine, but from a usability perspective, it would be nice to be able to reuse the existing Camel Infinispan or Redis components. This is not the first time we are "reinventing the wheel" for this kind of use case.

Given that we already have Infinispan and Redis components with their own configurations — which are supposed to be more feature-rich than the current state store implementation — I was wondering whether we could find a way to reuse those components (or at least part of their logic) for these use cases.

fyi @gnodet @orpiske @oscerd @apupier

oscerd:
> ```
> # Caffeine example
> camel.beans.caffeineBackend = #class:org.apache.camel.component.statestore.caffeine.CaffeineStateStoreBackend
> camel.beans.caffeineBackend.maximumSize = 50000
> 
> # Redis example
> camel.beans.redisBackend = #class:org.apache.camel.component.statestore.redis.RedisStateStoreBackend
> camel.beans.redisBackend.redisUrl = redis://myhost:6379
> camel.beans.redisBackend.mapName = my-app-state
> 
>  # Infinispan example
> camel.beans.infinispanBackend = #class:org.apache.camel.component.statestore.infinispan.InfinispanStateStoreBackend
> camel.beans.infinispanBackend.hosts = myhost:11222
> camel.beans.infinispanBackend.cacheName = my-cache
> ```
> 
> This sounds fine, but from a usability perspective, it would be nice to be able to reuse the existing Camel Infinispan or Redis components. This is not the first time we are "reinventing the wheel" for this kind of use case.
> 
> Given that we already have Infinispan and Redis components with their own configurations — which are supposed to be more feature-rich than the current state store implementation — I was wondering whether we could find a way to reuse those components (or at least part of their logic) for these use cases.
> 
> fyi @gnodet @orpiske @oscerd @apupier

I'm failing to see the reason why we should have another component of this kind, it will cause another dose of entropy.

orpiske:
> ```
> # Caffeine example
> camel.beans.caffeineBackend = #class:org.apache.camel.component.statestore.caffeine.CaffeineStateStoreBackend
> camel.beans.caffeineBackend.maximumSize = 50000
> 
> # Redis example
> camel.beans.redisBackend = #class:org.apache.camel.component.statestore.redis.RedisStateStoreBackend
> camel.beans.redisBackend.redisUrl = redis://myhost:6379
> camel.beans.redisBackend.mapName = my-app-state
> 
>  # Infinispan example
> camel.beans.infinispanBackend = #class:org.apache.camel.component.statestore.infinispan.InfinispanStateStoreBackend
> camel.beans.infinispanBackend.hosts = myhost:11222
> camel.beans.infinispanBackend.cacheName = my-cache
> ```
> 
> This sounds fine, but from a usability perspective, it would be nice to be able to reuse the existing Camel Infinispan or Redis components. This is not the first time we are "reinventing the wheel" for this kind of use case.
> 
> Given that we already have Infinispan and Redis components with their own configurations — which are supposed to be more feature-rich than the current state store implementation — I was wondering whether we could find a way to reuse those components (or at least part of their logic) for these use cases.
> 
> fyi @gnodet @orpiske @oscerd @apupier

**Pros:** 
- I like the consistency 
- I like that it has a swappable backend 
- I like the simplicity of the API 

**Cons:** 
- We already have code that does this, even though it's inconsistent, it is well known and documented
- The existing code already fits into Camel idioms 

**Conclusion**:

So, I am not opposed to have this one on the code base and I think there is potential for making stateful operations/caching a bit more elegant. The consistency of the operations make the behavior a bit more predictable regardless of the backend in use. 

However, I believe that we should have a greater discussion about what our vision and/or medium term goals for the problem is. Even if the API is elegant, it may not be enough if we don't collectively share the same vision about how we would like this to be ...

As such, I'm currently leaning towards 0 on this one and would like to hear more feedback from the community before having a decision on -1 or +1 from my side.

oscerd (changes requested):
Thanks for this — `camel-state-store` is a well-structured component with a clean pluggable SPI, thorough docs, and the generated files / DSL wiring are all in place. A couple of blockers before merge:

**Blocking**
- CI is red: the module's parent is still `org.apache.camel:components:4.19.0-SNAPSHOT` while `main` is now `4.21.0-SNAPSHOT`, so the `-Pregen` build can't resolve the parent POM. A rebase on `main` should fix this and also refresh the stale `4.19.0` markers (the `@UriEndpoint(firstVersion=...)`, the four module poms, and `:since: 4.19` in the docs), and re-run the generated files.
- The new tests use `Thread.sleep()` for the TTL-expiry waits (e.g. in `StateStoreTtlTest`, `CaffeineStateStoreBackendTest`, and the Infinispan/Redis ITs). Per the project's testing guideline these should move to Awaitility (`await().atMost(...).untilAsserted(...)`), which is also less flaky on slow CI agents.

**Minor / question**
- A couple of files use the FQCN form `@org.apache.camel.spi.annotations.Component(...)` and `java.util.Map.of(...)` rather than imports — OpenRewrite normally normalizes this, but the regen step fails before that phase.
- `MojoHelper.getComponentPath()` lists `camel-state-store` but not the three backend modules (caffeine/redis/infinispan). The generated files are committed so it evidently worked, but it may be worth registering them the way `camel-infinispan` does — could you confirm?

Solid contribution overall — the asks here are mostly mechanical.

---

gnodet (commented):

## Self-review: CAMEL-23239 — Add camel-state-store component (re-review after 6 new commits)

Well-designed component with clean SPI, smart auto-discovery, comprehensive docs, and good test coverage. The PR description is exemplary. However, several mechanical issues need to be fixed before this can merge.

### Blocking Issues

**1. Stale version `4.19.0-SNAPSHOT`** (already flagged by @oscerd)

All POM files, `@UriEndpoint(firstVersion)`, doc `:since:` tags, and generated JSON files reference `4.19.0-SNAPSHOT` while `main` is at `4.22.0-SNAPSHOT`. Needs a rebase and version update.

**2. `Thread.sleep()` in tests and production code** (already flagged by @oscerd)

8 occurrences across test files, plus one in production code:
- `InfinispanStateStoreBackend.start()` line ~111: `Thread.sleep(1000)` in a 10-retry loop — this blocks the thread for up to 10s during startup. Consider exponential backoff or Camel's retry mechanism.
- All TTL expiry tests use `Thread.sleep(1000)` or `Thread.sleep(2500)` — should use Awaitility per project conventions.

**3. FQCN usage in non-generated code**

- `InMemoryStateStoreBackend.java`: `.map(java.util.Map.Entry::getKey)` → import `Map.Entry`
- `StateStoreComponent.java`: `@org.apache.camel.spi.annotations.Component("state-store")` → import and use `@Component`

### Important (non-blocking)

**4. `StateStoreBackend` doesn't implement Camel's `Service` interface**

The SPI defines its own `start()`/`stop()` instead of extending `org.apache.camel.Service`. Extending `Service` would let `ServiceHelper` manage backends and enable Camel lifecycle state tracking. This is an intentional decision (SPI not in camel-api), but worth documenting in the interface Javadoc.

**5. Backend lifecycle race in `getOrCreateBackend()`**

When two endpoints with the same `storeName` but different explicit backends race, only the first wins silently. No warning is logged. Also, registry auto-discovery inside `computeIfAbsent` could theoretically deadlock with lazy bean creation.

**6. Redis `redisUrl` may contain credentials — not marked secret**

`redisUrl` can embed auth credentials (`redis://user:password@host:6379`). While backends are configured via `camel.beans.*` syntax (not `@UriParam`), recommend documenting that credentials can be embedded and suggesting vault integration for production.

**7. Missing `@since` tags on public classes**

`StateStoreBackend`, `StateStoreConstants`, `StateStoreOperations`, etc. should have `@since 4.22` (or whatever version after rebase) for consistency.

### Minor

- `InMemoryStateStoreBackend.putIfAbsent()` uses `Object[]` capture — `AtomicReference` would be cleaner
- `size()` and `keys()` are O(n) with TTL filtering — documented design choice but worth noting
- `InfinispanStateStoreBackend` retry loop: swallows intermediate exceptions, only throws the last one
- MojoHelper registers only `camel-state-store`, not backend modules — this appears correct (backends are "other", not components) but confirm per @oscerd's question

### Positives

- ✅ Clean, minimal SPI (9 methods) — appropriate for key-value abstraction
- ✅ Smart auto-discovery: single backend in registry used automatically
- ✅ Per-entry TTL with endpoint default + per-message header override
- ✅ Consistent `put` returns-previous-value semantics across all backends
- ✅ Backend sharing via `storeName`-keyed map in component
- ✅ Comprehensive documentation (430+ lines with Java/YAML DSL tabs)
- ✅ 50+ tests across all modules

**Recommendation**: Address the 3 blocking issues (rebase, Thread.sleep, FQCNs) — all mechanical fixes. The component design and functionality are solid and ready for merge after that.

components/camel-state-store/camel-state-store-infinispan/src/main/java/org/apache/camel/component/statestore/infinispan/InfinispanStateStoreBackend.java
gnodet: Good suggestion! I considered it but decided against it to keep the backend module dependency-minimal. resilience4j would add a transitive dependency just for a startup retry loop that only runs once during backend initialization. The exponential backoff loop (now updated from fixed 1s to 200ms–5s backoff) is straightforward and self-contained — it handles the narrow case of Infinispan server startup race conditions.

If the retry logic were more complex (circuit-breaking on operations, bulkhead patterns, etc.), resilience4j would be the right tool, but for a one-time startup sequence I think plain retry-with-backoff is the pragmatic choice.

gnodet:
Thanks for the thoughtful feedback from @Croway, @oscerd, and @orpiske on the overlap question. I've updated the PR description with a detailed section addressing this, but let me summarize the key points here:

### Why a new component instead of reusing existing ones

The existing Camel components (caffeine-cache, infinispan, spring-redis) are **technology-specific** — they expose the full API of each product (queries, pub/sub, statistics, events, etc.). That's their strength, but it's also their overhead when you just need `store(key, value)` / `retrieve(key)`.

**The primary motivation is MuleSoft migration.** MuleSoft's Object Store is a core primitive that MuleSoft users rely on heavily. When migrating MuleSoft flows to Camel, there's no direct equivalent — users are forced to pick a technology (Caffeine? Redis? Infinispan?) and learn its full component API, when all they need is a simple key-value store.

### What the backend modules do (and don't do)

The backend modules (`camel-state-store-caffeine`, etc.) are **thin adapters** — they implement the `StateStoreBackend` interface using each technology's native client library directly. They don't duplicate or wrap the existing Camel components. Think of it like how `camel-jms` has its own `ConnectionFactory` setup rather than wrapping `camel-activemq`.

### This follows established Camel patterns

- `camel-sql` / `camel-jdbc` vs vendor-specific SQL components
- `camel-jms` vs `camel-activemq`, `camel-amqp`
- `camel-jcache` (JSR-107) vs `camel-caffeine-cache`, `camel-ehcache`

### Regarding @orpiske's point about medium-term vision

I agree that a broader discussion about stateful operations in Camel would be valuable (and CAMEL-11114 "Create cache DSL" is related). This component is designed to be compatible with a future direction: the `StateStoreBackend` interface is currently scoped to the component module, but could be promoted to `camel-api` if a core cache DSL/EIP emerges.

I've pushed an update addressing all the code-level review feedback (rebased, squashed, versions fixed, Thread.sleep → Awaitility, AssertJ, @since tags, FQCN fixes). Happy to discuss the architectural direction further.

davsclaus:
@gnodet  this needs to be changed tro 4.23.0-SNAPSHOT

davsclaus (commented):
## Review of camel-state-store

Thanks for this work, @gnodet — the component is well-structured, follows Camel conventions well, and has solid test coverage across all four backends. Here are the findings from the review.

### Critical

1. **MojoHelper missing backend module registrations** — `MojoHelper.getComponentPath()` only registers `camel-state-store` but not `camel-state-store-caffeine`, `camel-state-store-redis`, `camel-state-store-infinispan`. Without this, catalog/docs generation will not discover the backend modules' metadata. The fix should use `Arrays.asList(...)` to include all four sub-modules, similar to how `camel-test` registers its sub-modules.

### Medium

2. **`StateStoreComponent.doStop()` resource leak** — If one backend's `stop()` throws, subsequent backends are never stopped. Each `stop()` call should be wrapped in try-catch so all backends get a chance to clean up.

3. **Case-sensitive operation lookup** — `StateStoreProducer.determineOperation()` uses `StateStoreOperations.valueOf(headerOp.toString())` which throws a cryptic `IllegalArgumentException` for `"PUT"` vs `"put"`. Consider case-insensitive lookup or a try-catch with a user-friendly error message listing valid operations.

4. **Infinispan `Thread.sleep()` in `start()`** — The retry loop with exponential backoff (up to ~30s) blocks `CamelContext.start()`. A component's `start()` method ideally should not compensate for infrastructure not being ready — that is the deployer's responsibility. Consider failing fast, or lazy cache acquisition on first operation.

5. **Infinispan IT `RemoteCacheManager` leak** — The test creates a `RemoteCacheManager` externally via `setCacheManager()` but never closes it (no `@AfterAll`). Since `managedCacheManager=false`, `stop()` won't close it either.

6. **Adoc include directives misplaced** — `component-configure-options.adoc` include is inside the `component options: START/END` block instead of the `component-configure options: START/END` block. This will cause the generated options tables to render in unexpected locations.

### Low

7. **Redis/Infinispan `managed*` flag not reset in `stop()`** — If ownership changes between restart cycles (e.g., `stop()`, then `setRedisson(externalClient)`, then `start()`), the externally provided client gets erroneously shut down on the next `stop()`.

8. **Silent backend discard** — When a `storeName` already has a backend, an explicitly configured `backend=#myBackend` on a new endpoint is silently ignored. Should at least log at WARN level.

9. **No fallback log** — When zero `StateStoreBackend` beans are found in the registry, no log indicates the fallback to `InMemoryStateStoreBackend`. A DEBUG/INFO message would help troubleshooting.

10. **`InMemoryStateStoreBackend.keys()`** — Scans expired entries but doesn't evict them, leading to memory accumulation for high-churn short-TTL workloads.

11. **Redundant caffeine version** — `camel-state-store-caffeine/pom.xml` specifies `${caffeine-version}` explicitly; it's already managed in parent dependency management.

12. **Test gaps:**
  - Redis IT `testDelete()` doesn't verify the key is gone after deletion (compare to Caffeine/InMemory tests which do a follow-up `get`)
  - Infinispan IT missing explicit `testClear()` and negative `contains` case
  - No test for invalid operation name in header (e.g., `"BOGUS"`)
  - No test for delete of non-existent key (behavior may differ between backends)

gnodet:
Thanks for the thorough review @davsclaus! All 12 findings have been addressed in the latest push. Here's a summary:

### Critical
1. **MojoHelper backend module registrations** — ✅ Fixed. Added `case "camel-state-store":` in `MojoHelper.getComponentPath()` registering all 4 sub-modules (camel-state-store, camel-state-store-caffeine, camel-state-store-redis, camel-state-store-infinispan).

### Medium
2. **`StateStoreComponent.doStop()` resource leak** — ✅ Fixed. Each `backend.stop()` is now wrapped in try-catch. Exceptions are logged at WARN, the first is captured and re-thrown after all backends have been stopped and the map cleared.

3. **Case-sensitive operation lookup** — ✅ Fixed. `determineOperation()` first tries exact `valueOf()`, then falls back to case-insensitive matching. On failure, throws a clear error listing all valid operations.

4. **Infinispan `Thread.sleep()` in `start()`** — ✅ Fixed. Removed the retry loop entirely. `start()` now creates the `RemoteCacheManager` and calls `getOrCreateCache()` directly — fails fast if the server is unavailable.

5. **Infinispan IT `RemoteCacheManager` leak** — ✅ Fixed. Added `@AfterEach cleanUp()` that stops and nulls the `testCacheManager`.

6. **Adoc include directives misplaced** — ✅ Fixed. The includes now match the standard Camel pattern. CI generated-files check passes.

### Low
7. **Redis/Infinispan `managed*` flag not reset in `stop()`** — ✅ Fixed. Both `RedisStateStoreBackend.stop()` and `InfinispanStateStoreBackend.stop()` now reset `managedRedisson`/`managedCacheManager` to `false`.

8. **Silent backend discard** — ✅ Fixed. A WARN log is now emitted when a `storeName` already has a backend and an explicit `backend=#ref` is ignored.

9. **No fallback log** — ✅ Fixed. Three cases covered: INFO when falling back to `InMemoryStateStoreBackend`, DEBUG when auto-discovering a single backend, WARN when multiple backends are found and cannot auto-select.

10. **`InMemoryStateStoreBackend.keys()` eviction** — ✅ Fixed. Both `keys()` and `size()` now call `evictExpired()` before returning results, using `removeIf(e -> e.getValue().isExpired())`.

11. **Redundant caffeine version** — ✅ Fixed. Removed the explicit `${caffeine-version}` from `camel-state-store-caffeine/pom.xml`; it inherits from parent BOM.

12. **Test gaps** — ✅ Fixed. Added:
  - Redis IT `testDelete()` now verifies key is gone after deletion
  - Infinispan IT `testClear()` and negative `contains("missing")` case
  - `testInvalidOperationThrows()` for bogus operation name
  - `testDeleteNonExistentKey()` across in-memory, Redis, and Infinispan
  - `testCaseInsensitiveOperation()` for mixed case operations

gnodet:
@oscerd — the blocking items from your review have also been addressed:

- **CI red / version mismatch** — ✅ Rebased on main, all version references updated to 4.23.0-SNAPSHOT
- **Thread.sleep() in tests** — ✅ Migrated all TTL waits to Awaitility (`await().atMost(...).untilAsserted(...)`)
- **FQCN usage** — ✅ Replaced all fully qualified class names with imports
- **MojoHelper registration** — ✅ All 4 sub-modules registered (see finding #1 above)

oscerd (approved):

gnodet:
**CI note:** The Java 25 build failure is unrelated to this PR — it's caused by the known `camel-pqc` KeyStore test regression on Java 25 (JDK-native ML-DSA keys not recognized by Bouncy Castle). This is fixed in #25510. Once that merges, a rebase will make CI green here.

gnodet:

Note: the `StateStoreBackend` SPI in this PR could eventually be reworked on top of the unified `KeyValueRepository` SPI being introduced in #25631 (CAMEL-24463). That effort aims to provide a single key-value abstraction that can back `IdempotentRepository`, `AggregationRepository`, cache, and state store — so adding one backend implementation gives all EIP integrations for free.

For now, this PR keeps `StateStoreBackend` self-contained as a Preview component, and unification would happen as a follow-up once the `KeyValueRepository` SPI is finalized.

PR #22197, opened 2026-03-23
CAMEL-23226: Add TUI dashboards for Camel JBang

gnodet:
JIRA: [CAMEL-23226](https://issues.apache.org/jira/browse/CAMEL-23226)

## Summary

Add full-screen TUI (Terminal User Interface) dashboards to Camel JBang using TamboUI, providing rich interactive monitoring and browsing capabilities directly in the terminal. The TUI is implemented as a separate plugin module (`camel-jbang-plugin-tui`), loaded via the `@CamelJBangPlugin` SPI mechanism.

### New Commands

| Command | Description |
|---------|-------------|
| `camel monitor` | Multi-tab dashboard with overview, routes, processors, health, endpoints, log, and trace tabs |
| `camel catalog-tui` | Interactive component catalog browser with search and detail view |

### Monitor Tabs

1. **Overview** — Live integration list with status, uptime, throughput sparklines, memory gauge
2. **Routes** — Route list with state, message counts, timing stats, and processor detail
3. **Processors** — Per-route processor breakdown with message counts and timing
4. **Health** — Health check dashboard with UP/DOWN indicators
5. **Endpoints** — Endpoint list with direction, component, and route association
6. **Log** — Scrollable log viewer with level filtering (TRACE/DEBUG/INFO/WARN/ERROR), follow mode, and detail panel
7. **Trace** — Exchange trace viewer with detail panel for headers, body, and exchange properties

### Key Features

- Live data refresh with configurable intervals
- Keyboard navigation (arrow keys, tab switching via number keys)
- Color-coded status indicators (green/yellow/red)
- Sparkline throughput charts in overview
- Log level filtering and follow mode
- Trace detail with headers/body toggle
- Memory gauge with heap usage
- Vanishing effect for stopped integrations
- Fuzzy search in catalog browser

### Dependencies

- Add TamboUI 0.2.0-SNAPSHOT dependencies (`tamboui-tui`, `tamboui-widgets`, `tamboui-jline3-backend`)

### Module Structure

- `dsl/camel-jbang/camel-jbang-plugin-tui/` — new plugin module
  - `CamelMonitor.java` — multi-tab monitor dashboard
  - `CamelCatalogTui.java` — interactive catalog browser
  - `TuiHelper.java` — shared TUI utilities
  - `TuiPlugin.java` — plugin registration via SPI

## Test plan

- [x] `camel monitor` shows running integrations with live stats across all tabs
- [x] Log tab correctly parses ANSI-stripped log lines with regex-based parser
- [x] Trace tab correctly parses backlog trace JSON format
- [x] All null fields handled safely (no NPEs in rendering)
- [x] `camel catalog-tui` browses components with search and detail view
- [x] All TUI views handle Ctrl+C/q cleanly


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelCatalogTui.java
```diff
@@ -0,0 +1,697 @@
...
+    @Override
+    public Integer doCall() throws Exception {
+        try {
+            Class.forName("dev.tamboui.tui.event.KeyModifiers");
+            Class.forName("dev.tamboui.tui.event.KeyEvent");
+            Class.forName("dev.tamboui.tui.event.KeyCode");
+            Class.forName("picocli.CommandLine$IExitCodeGenerator");
+        } catch (ClassNotFoundException e) {
+            // ignore
+        }
```
apupier: What's the interest of callingClass;forName to then ignore exception and do nothing about it?

gnodet: _Claude Code on behalf of Guillaume Nodet_

Good question! This eagerly pre-loads classes that the TUI input reader daemon thread needs. Without this, during JVM shutdown the classloader may already be closing (especially with JBang's custom classloader) while the input reader thread is still trying to load `KeyModifiers`, `KeyEvent`, etc. lazily — causing `ClassNotFoundException` stack traces on exit.

The comment was missing on this file (the others already had it). Added it now, and will also extract this into a shared utility to avoid the duplication.

Croway:
Why is the TUI part of Camel JBang core? what about having it as a new plugin?

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelCatalogTui.java
gansheer: If one fails the following classes are not pre-loeaded and there is no error handling that could help understand what happened.

gnodet: Sure, though, if one fails, the _real_ code (main thread) will certainly fail too with an error that will be logged somewhere.  This is really only to avoid the daemon thread to fail with a non actionable exception.  I guess another possibility would be to swallow exceptions when closing, but that may hide other problems.

gnodet:

@Croway The TUI is already implemented as a separate plugin module (`camel-jbang-plugin-tui`), not in camel-jbang-core. The initial commit had it in core but it was moved to its own plugin in a follow-up commit (549d945). The plugin is loaded via the `@CamelJBangPlugin` annotation and SPI mechanism.

davsclaus:
Do you have some screen shots ;)

davsclaus:
@gnodet would you mind rebasing this so its on top of latest code, thanks

davsclaus:
Screenshot 2026-04-18 at 10 40 26

davsclaus:
okay found a solution to make tamboui works as a custom plugin in terms of classloading. 

However the core commands can benefit from tamboui as well, so if we do a big v2 of camel-jbang then tamboui should ideally be used in core and be the UI element.

davsclaus (approved):
This is a great start for this experiment and to build up a TUI dashboard.

davsclaus:
@gnodet with the TUI in your maven pilot, then you can use that learning to improve this TUI to be better.

However in the mean time lets get something merged so more users can see, try, and provide feedback.

PR #22544, opened 2026-04-10
camel-telemetry: fix trace context propagation for messaging components

Croway:
## Summary
- `AbstractMessagingSpanDecorator` used the default `CamelHeadersSpanContextPropagationExtractor` which only handles `String`-valued headers. Messaging transports like Kafka deliver headers as `byte[]`, so trace context headers (e.g. `traceparent`) were silently dropped, breaking distributed trace propagation across services.
- Add `CamelMessagingHeadersSpanContextPropagationExtractor` that handles both `String` and `byte[]` headers, and override `getExtractor()` in `AbstractMessagingSpanDecorator` so all messaging components (Kafka, AMQP, SJMS, STOMP, Spring RabbitMQ, Azure Service Bus, etc.) benefit from the fix.
- The issue was discovered while upgrading the camel-spring-boot opentelemetry example from `camel-opentelemetry` to `camel-opentelemetry2`.

## Test plan
- [x] 91 camel-telemetry tests pass (84 existing + 7 new for the messaging extractor)
- [x] Verified with the opentelemetry Spring Boot example: async Kafka traces now show 20 spans across all 4 services in a single trace (was broken into 7 separate traces before the fix)

### DIFF
--- a/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/propagation/CamelHeadersSpanContextPropagationExtractor.java
+++ b/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/propagation/CamelHeadersSpanContextPropagationExtractor.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.telemetry.propagation;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


squakez (commented):
I think that the decode of the "dash" is something exclusive of JMS protocol, I am not aware of other protocols, for example Kafka, having the same limitation. I you notice, the encoding part belongs to `CamelJMSHeadersSpanContextPropagationInjector` class, so, I guess that if anything has to be done with this class, it makes sense to have the specular encoding part. Same for testing. It makes sense to have an encode/decode test using for consistency reason

squakez:
I think you can instead change the AbstractMessaging and deal with the string/byte conversion there leaving the "dashing" problem to the specific JMS subclass and separating clearly the concerns.

squakez (approved):
This version is fine. We could work in the future to make the JMS one to extend from this, but it is just a minor optimization we can think about.

PR #22547, opened 2026-04-10
CAMEL-23267: Use simple LinkedHashMap in MemoryIdempotentRepository which is used as in progress repo for file based components

davsclaus:
# Description

# Target

# Tracking


core/camel-support/src/main/java/org/apache/camel/support/processor/idempotent/MemoryIdempotentRepository.java
```diff
@@ -164,7 +168,12 @@ public void setCacheSize(int cacheSize) {
     @Override
     protected void doStart() throws Exception {
         if (cache == null) {
-            cache = LRUCacheFactory.newLRUCache(cacheSize <= 0 ? MAX_CACHE_SIZE : cacheSize);
+            cache = new LinkedHashMap<>() {
+                @Override
+                protected boolean removeEldestEntry(Entry<String, Object> eldest) {
+                    return size() > cacheSize;
+                }
+            };
```
gnodet, line 176: The old code had a guard for non-positive cache sizes: `cacheSize <= 0 ? MAX_CACHE_SIZE : cacheSize`. The new code relies on the field default `cacheSize = MAX_CACHE_SIZE` but removes the runtime guard.

If `setCacheSize(0)` or `setCacheSize(-1)` is called before `doStart()`, the `LinkedHashMap` would evict on every insertion (`size() > 0`), making the repository unable to hold any entries. The old code treated this as "use default 1000."

Consider keeping the guard:
```java
int effectiveSize = cacheSize <= 0 ? MAX_CACHE_SIZE : cacheSize;
cache = new LinkedHashMap<>() {
    @Override
    protected boolean removeEldestEntry(Entry<String, Object> eldest) {
        return size() > effectiveSize;
    }
};
```

core/camel-support/src/main/java/org/apache/camel/support/processor/idempotent/MemoryIdempotentRepository.java
```diff
@@ -59,14 +59,18 @@ public MemoryIdempotentRepository(Map<String, Object> set) {
     }
 
     /**
-     * Creates a new memory based repository using a {@link LRUCache} with a default of 1000 entries in the cache.
+     * Creates a new memory based repository using a {@link java.util.LinkedHashMap} as its store, with 1000 maximum
+     * capacity. When a new entry is added and the store has reached its maximum capacity, the oldest entry is removed.
+     *
      */
     public static IdempotentRepository memoryIdempotentRepository() {
```
gnodet, line 66: Nit: trailing blank line in the Javadoc (line before `*/`).

gnodet (commented):

## Review of CAMEL-23267: Replace LRUCache with LinkedHashMap

Nice simplification compared to #22311 — changing the base `doStart()` method fixes all callers (file, AWS S3, IBM COS, FTP, SMB) without adding new API surface.

### What's good

- Upgrade guide entry included
- Test for eviction behavior included
- `cacheSize` field now has a sensible default (`MAX_CACHE_SIZE`), fixing JMX `getMaxCacheSize()` returning 0
- `LRUCache`/`LRUCacheFactory` imports cleanly removed
- Semantic equivalence is correct: `SimpleLRUCache.get()` did not record access in `lastChanges`, so it was already tracking insertion/update order rather than true LRU. The switch to `LinkedHashMap` is behaviorally equivalent.

### Findings

See inline comments for specific items.

### Pre-existing note (non-blocking)

`getCacheSize()` (line 153) calls `cache.size()` and `doStop()` calls `cache.clear()` — both without acquiring `cacheAndStoreLock`. With `SimpleLRUCache` (backed by `ConcurrentHashMap`) these were benign. With a plain `LinkedHashMap` they become data races. Not introduced by this PR, but the switch from a concurrent backing structure to a non-concurrent one makes it slightly more relevant.

### PR housekeeping

- The PR description is just the template boilerplate — it would help to fill in context about CAMEL-23267 and the approach taken.
- If #22547 is merged, #22311 should be closed. It would be nice to acknowledge tga's contribution since the original investigation and test were theirs.

PR #22577, opened 2026-04-13
Add SpanKind parameter to telemetry API to support proper span type

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-23312
    
  Adds SpanKind parameter to the Camel telemetry API to enable proper span type classification
  (CLIENT, SERVER, PRODUCER, CONSUMER, INTERNAL) across different observability backends.            
   
  Changes:                                                                                           
  - Introduces org.apache.camel.telemetry.SpanKind enum
  - Updates SpanLifecycleManager.create() signature to accept span kind                              
  - Implements span kind mapping in tracer implementations:            
    - OpenTelemetry2: Maps to native io.opentelemetry.api.trace.SpanKind                             
    - Micrometer: Accepts parameter for API compatibility (kind determined by Observation context)  without using it
    - TelemetryDev: Stores as span tag                                                               
  - Extends SpanDecorator and decorator implementations to provide span kind information             
  - Migrates test code from Thread.sleep() to Awaitility for more reliable async testing             
                                                                                                     
  Impact: Enables proper span classification in distributed traces, improving observability for      
  messaging and HTTP components.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/SpanKind.java
+++ b/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/SpanKind.java
@@ -17,18 +17,23 @@
 package org.apache.camel.telemetry;
 
 /**
- * This interface is used to manage the lifecycle of a Span.
+ * Span kind constants for telemetry tracing.
+ * <p>
+ * These values describe the relationship between the span and its parent:
+ * <ul>
+ * <li>CLIENT - The span covers a client-side call to a remote service</li>
+ * <li>SERVER - The span covers server-side handling of a remote request</li>
+ * <li>PRODUCER - The span covers the production of a message to a remote system (e.g., message broker, queue, HTTP
+ * endpoint)</li>
+ * <li>CONSUMER - The span covers the consumption of a message from a remote system (e.g., message broker, queue, HTTP
+ * endpoint)</li>
+ * <li>INTERNAL - The span represents internal operations with no remote interaction</li>
+ * </ul>
  */
-public interface SpanLifecycleManager {
-
-    Span create(String spanName, SpanKind kind, Span parent, SpanContextPropagationExtractor extractor);
-
-    void activate(Span span);
-
-    void deactivate(Span span);
-
-    void close(Span span);
-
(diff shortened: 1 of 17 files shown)


components/camel-telemetry/src/main/java/org/apache/camel/telemetry/SpanKind.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.camel.telemetry;
+
+/**
+ * Span kind constants for telemetry tracing.
+ * <p>
+ * These values describe the relationship between the span and its parent:
+ * <ul>
+ * <li>CLIENT - The span covers a client-side call to a remote service</li>
+ * <li>SERVER - The span covers server-side handling of a remote request</li>
+ * <li>PRODUCER - The span covers the production of a message to a remote broker/queue</li>
```
davsclaus: I dont think this is limited to a remote broker/queue, it can be any kind of remote networking, even a HTTP REST call and so on.

davsclaus (approved):

davsclaus:
There is some API changes it would be nice to add this in the 4.20 upgrade guide doc.

squakez (commented):
I will have a thoroughly look asap. However, in the while just consider that the "spanKind" concept is an opentelemetry concept only and not necessarily supported by any other telemetry components. Ideally this feature should be developed at subcomponent level only and we need to evaluate if any generic "extension point" is really required on the abstract telemetry component.

I will give you more details shortly.

squakez (changes requested):
Thanks for the work but I think what you're trying to achieve is already covered by the presence of the SpanDecorator, whose goal is exactly to include any additional information to the Span before or after the trace is generated (via methods `public void beforeTracingEvent(Span span, Exchange exchange, Endpoint endpoint)` and `public void afterTracingEvent(Span span, Exchange exchange)`).

I think that the kind can be generated and set on the span directly in each decorator, which would be also cleaner from an API point of view.

JiriOndrusek:
Superseded by https://github.com/apache/camel/pull/22610

PR #22587, opened 2026-04-13
Use JDK21 when deploying SNAPSHOT

Croway:
with jdk 17 the multi jar deployment for virtual threads does not work, use jdk21 instead.

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 def AGENT_LABEL = env.AGENT_LABEL ?: 'ubuntu'
-def JDK_NAME = env.JDK_NAME ?: 'jdk_17_latest'
+def JDK_NAME = env.JDK_NAME ?: 'jdk_21_latest'
 
 def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly -Dmaven.compiler.fork=true "
 


davsclaus (approved):

apupier:
> with jdk 17 the multi jar deployment for virtual threads does not work, use jdk21 instead.

Do you mean that we must use a JDK 21+ to compile Camel source code to have virtual threads potentially used when running with JDk 21+?
If yes, it is a bit weird and even more importantly it means that we also need to ensure that the "release" bits are compiled with JDK21+ to have it usable. Which I'm not sure how to enforce as it is built on the local machine of the releaser.

Croway:
> > with jdk 17 the multi jar deployment for virtual threads does not work, use jdk21 instead.
> 
> Do you mean that we must use a JDK 21+ to compile Camel source code to have virtual threads potentially used when running with JDk 21+? If yes, it is a bit weird and even more importantly it means that we also need to ensure that the "release" bits are compiled with JDK21+ to have it usable. Which I'm not sure how to enforce as it is built on the local machine of the releaser.

@apupier at the moment we are not enforcing it, but you have a good point, we should, it should be doable via the maven-enforcer-plugin.

Gregor, the release manager, always released Camel with JDK 21 (or JDK 25, since there is a profile for it) and this is the correct approach, given that the compatibility is [JDK17](https://github.com/apache/camel/blob/main/pom.xml#L113) anyway (@davsclaus @oscerd may you confirm I didn't write any blasphemy please?).

Anyway, the issue with the SNAPSHOT is that the JDK21 (virtual thread) snapshot is not released, and one spring boot test is failing when running the test suite with jdk21 since the jar is not released as expected.

I'd merge it, and we can discuss a `maven-enforcer-plugin` configuration, wdyt?

davsclaus:
If you look inside the MANIFEST.MF of the 4.19 release you can see
```
Manifest-Version: 1.0
Created-By: Maven JAR Plugin 3.5.0
Java-Version: 17
Build-Jdk-Spec: 21
Implementation-Title: Camel :: Core Engine
Implementation-Version: 4.19.0
Implementation-Vendor: The Apache Software Foundation
```

Does that not indicate that JDK21 was used during the release

Croway:
Exactly! and it is wrong-ish, the JDK25 profile was not activated in this case. Gregor used JDK21 to build Camel most probably.

Most of the byte code target is Java 17 `Java-Version: 17`, but was built with `Build-Jdk-Spec: 21`, therefore, the Virtual Threads are enabled, but the `ScopeValue` (JDK25 profile) are not present in this release.

We should enforce the JDK used to build Camel

PR #22610, opened 2026-04-14
CAMEL-23312: Add SpanKind support to OpenTelemetry2 component 

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-23312
superseeds https://github.com/apache/camel/pull/22577

  Adds OpenTelemetry SpanKind support to camel-opentelemetry2 component:                                                                                                                      
                                                                                                                                                                                              
  - Introduced `OpenTelemetryHttpSpanDecorator` and `OpenTelemetryMessagingSpanDecorator` interfaces defining SpanKind for EVENT_SENT/EVENT_RECEIVED operations                               
  - Modified `OpenTelemetryTracer` to pass decorator context via ThreadLocal without changing SpanLifecycleManager API
  - All concrete decorators extend camel-telemetry decorators (inherit component-specific behavior like Kafka tags) and implement OpenTelemetry interfaces (add SpanKind)                     
  - HTTP components use CLIENT/SERVER, messaging components use PRODUCER/CONSUMER, others default to INTERNAL 

# Description

<!--
- Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
-->

# Target

- [ ] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking
- [ ] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

<!--
# *Note*: trivial changes like, typos, minor documentation fixes and other small items do not require a JIRA issue. In this case your pull request should address just this issue, without pulling in other changes.
-->

# Apache Camel coding standards and style

- [ ] I checked that each commit in the pull request has a meaningful subject line and body.

<!--
If you're unsure, you can format the pull request title like `[CAMEL-XXX] Fixes bug in camel-file component`, where you replace `CAMEL-XXX` with the appropriate JIRA issue.
-->

- [ ] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

<!--
You can run the aforementioned command in your module so that the build auto-formats your code. This will also be verified as part of the checks and your PR may be rejected if if there are uncommited changes after running `mvn clean install -DskipTests`.

You can learn more about the contribution guidelines at https://github.com/apache/camel/blob/main/CONTRIBUTING.md
-->

PR #22612, opened 2026-04-14
CAMEL-23312: Add SpanKind support to OpenTelemetry2 component

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-23312
superseeds https://github.com/apache/camel/pull/22577

  - Adds SpanKind support to the OpenTelemetry2 component to enable proper distributed tracing
  semantics                                                                                          
  - Implements SpanKind determination through the SpanDecorator API with HTTP components using
  CLIENT/SERVER and messaging components using PRODUCER/CONSUMER                                     
  - Includes comprehensive test coverage with mock HTTP and Kafka components to verify SpanKind
  behavior and decorator property inheritance 
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetrySpanAdapter.java
+++ b/components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetrySpanAdapter.java
@@ -16,14 +16,20 @@
  */
 package org.apache.camel.opentelemetry2;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.context.Scope;
+import org.apache.camel.telemetry.Op;
 import org.apache.camel.telemetry.TagConstants;
 
 public class OpenTelemetrySpanAdapter implements org.apache.camel.telemetry.Span {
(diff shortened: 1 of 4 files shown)


JiriOndrusek:
@squakez  here is the refactored PR (much simpler and cleaner), I force pushed this change to my branch, unfortunately this blocks reopening of https://github.com/apache/camel/pull/22610 (I didn't know about such limitation). Therefore this is the new  PR.

squakez (changes requested):
Hello, thanks again for the work but it appears to be quite far from what I was proposing.

The change I'm advocating for are quite simple and compliant with the actual design. What I think you need to do is to change the implementation of the dedicated `beforeTracingEvent(Span span, Exchange exchange, Endpoint endpoint)` for those decorators (messaging and http based) which you want to add a `kind`. For example:

https://github.com/apache/camel/blob/061ae7d6dc0f0e5a05af83bc4bb3484c2e128108/components/camel-telemetry/src/main/java/org/apache/camel/telemetry/decorators/AbstractHttpSpanDecorator.java#L67-L75

Here you need to add a tag to the existing `Span` object, i.e. `span.setTag("kind", "my-kind")`. The method gives you everything you need to perform the action. If you need the `Op` type, you can extend both the before and after tracing with such an object. In this way it will be available for any type evaluation inside that method.

Hope it helps.

squakez:
Alternatively we could even think to expose the "kind" in the Span (ie, `span.kind()`), just for the sake to avoid adding the Op in the decorator methods. In this way all you need is self contained and indeed you can have as much as Kind types as you need in the SpanAdapter which is implemented at a concrete component level. Yeah, I think this is the best way to handle to minimize the impact and be compliant with the design.

JiriOndrusek:
Hi  @squakez,

I've tried to follow your hints. 
1 - I agree with avoiding Op in decorators
2 - I added `span.kind()` into Span and started from it.
Unfortunately the `io.opentelemetry.api.trace.Span` needs to know its spanKind before starting. I had to deffer the start of the span into the `makeCurrent` method to be able to detect the spanKind. I put the spanKind detection into `OpenTelemetrySpanAdapter` and kept it completely self contained. (I used other tags for it) So I actually did not need the `span.kind()` anymore. What do you think? from my PoV the change is quite small and isolated to opentelemetry2. No API is changed. Test suite is the same as in previous attempts.

components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetrySpanAdapter.java
```diff
@@ -16,38 +16,82 @@
...
     private Scope baggageScope;
 
-    protected OpenTelemetrySpanAdapter(Span otelSpan, Baggage baggage) {
-        this.otelSpan = otelSpan;
-        // We store an important flag in the baggage in order to verify if the
-        // root span was generated internally or from a third party dependency.
-        this.baggage = baggage.toBuilder().put(BAGGAGE_CAMEL_FLAG, "true").build();
+    // For deferred span creation
+    private SpanBuilder spanBuilder;
+    private final Map<String, String> pendingTags = new HashMap<>();
```
davsclaus, line 47: should these map/list be concurrent variants in case multi threads calls at the same time

JiriOndrusek: I discussed the changes with @squakez  and I'll open the new PR which does not need all this code for defferring span creation

JiriOndrusek:
superseded by https://github.com/apache/camel/pull/22653

PR #22651, opened 2026-04-16
CAMEL-23334: Add preferLocal mode to avoid remote checks for cached plugin artifacts

Croway:
I have been facing usability issues with the Forage plugin, in particular, when using the SNAPSHOT, but the same applies for the released version, every time a camel command is executed, the plugin's remote repository was checked, even if the plugin was already present in the local repo. With slow repositories (even sonatype), the camel CLI was not usable (20-30 sec for each command).

## Summary

- Adds a `preferLocal` flag to `MavenDownloader` / `MavenDownloaderImpl` that, when enabled, checks if requested artifacts exist in the local Maven repository and attempts offline resolution first, skipping remote repository checks (especially SNAPSHOT metadata). Falls back to normal online resolution if offline fails.
- `PluginHelper.downloadPlugin()` opts in to this mode, reducing JBang plugin resolution from ~30-90s to ~2-3s on every CLI invocation when the plugin is already cached locally.
- Default behavior (`preferLocal=false`) is unchanged for all existing callers.

## Test plan

- [x] New tests `testPreferLocalResolvesFromCache` and `testPreferLocalFallsBackToOnline` in `MavenDownloaderImplTest`
- [x] Existing `camel-tooling-maven` and `camel-jbang-core` tests pass
- [x] Manual testing: `camel version` with a 3rd-party SNAPSHOT plugin installed completes in ~2s (vs ~90s before)

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -219,6 +219,8 @@ public final class PluginHelper {
         if (repos != null && !repos.isBlank()) {
             downloader.setRepositories(repos);
         }
+        // prefer resolving from local Maven repository to avoid expensive remote SNAPSHOT metadata checks
+        downloader.setPreferLocal(true);
         downloader.start();
         // downloads and adds to the classpath
         downloader.downloadDependencyWithParent("org.apache.camel:camel-jbang-parent:pom:" + camelVersion, group,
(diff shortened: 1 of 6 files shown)


davsclaus (approved):

davsclaus:
its a shame that we have this plugin loader that ends up being slow, even locally if that is 2-3s per command. Especially when you run a command that dont use any plugins, eg `camel get` - the plugin loader could maybe be made smarter to only load if the command to be executed is not a built-in command.

Croway:
yes, I was thinking about that, I'll open an issue.

Moreover there are commands like `camel cmd` for example, that behind the hood invoke `run()`, and the plugin is dowloaded twice! these use cases needed more than a minute

oscerd (approved):

Croway:
@davsclaus https://issues.apache.org/jira/browse/CAMEL-23335

Croway:
The CI failures are unrelated

davsclaus:
LGTM

PR #22638, opened 2026-04-15
CAMEL-23331: Add support for blob snapshot creation in camel-azure-storage-blob

oscerd:
## Summary

- Add `createBlobSnapshot` producer operation that creates a read-only snapshot of a blob via the Azure SDK's `BlobClientBase.createSnapshotWithResponse()`
- The snapshot ID is returned in the exchange body and via the new `CamelAzureStorageBlobSnapshotId` header
- Includes unit test, documentation update, and all regenerated files (catalog, endpoint DSL)

### Review follow-up (commit 2)

- `BlobConfiguration`: fix pre-existing metadata gap — `getChangeFeed` is a valid producer operation (declared in `BlobOperationsDefinition`, wired in `BlobProducer`) but was missing from the `@UriParam` `enums` list. Added alongside `copyBlob` and `createBlobSnapshot`.
- `BlobOperationsDefinition`: fix pre-existing copy-paste error in the `copyBlob` Javadoc (was a duplicate of `getChangeFeed`'s description, dating back to CAMEL-17161 in Nov 2021).

## Affected files

- `BlobOperationsDefinition` — new `createBlobSnapshot` enum value + corrected `copyBlob` Javadoc
- `BlobConstants` — new `BLOB_SNAPSHOT_ID` header constant
- `BlobConfiguration` — updated operation enums list (`createBlobSnapshot`, `copyBlob`, `getChangeFeed`)
- `BlobExchangeHeaders` — new `snapshotId()` fluent setter
- `BlobClientWrapper` — new `createSnapshot()` wrapper delegating to Azure SDK
- `BlobOperations` — new `createBlobSnapshot()` operation implementation
- `BlobProducer` — wired `createBlobSnapshot` in the switch statement
- Documentation and unit test added

## Test plan

- [x] Unit test `BlobOperationsTest.testCreateBlobSnapshot()` passes
- [x] All existing tests pass (32 tests, 0 failures)
- [x] Full `mvn clean install -DskipTests` from root succeeds — all generated files committed
- [ ] Integration test against Azure/Azurite (requires credentials)

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Andrea Cosentino

PR #22656, opened 2026-04-16
CAMEL-23331: Add support for blob snapshot retrieval in camel-azure-storage-blob

oscerd:
## Summary

Follow-up to #22638. Completes the second half of CAMEL-23331 by adding the `snapshotId` read-path.
After this PR, `getBlob`, `downloadBlobToFile` and `downloadLink` can target a specific blob snapshot version
instead of the live blob.

## Changes

- `BlobConfiguration` — new `snapshotId` `@UriParam(common)`
- `BlobConstants` — `BLOB_SNAPSHOT_ID` description updated to cover both roles (input for read ops / output of `createBlobSnapshot`)
- `BlobExchangeHeaders` — new `getSnapshotIdFromHeaders()` reader (setter `snapshotId()` already existed from #22638)
- `BlobConfigurationOptionsProxy` — resolve `snapshotId` from header, fall back to config
- `BlobClientWrapper` — new `withSnapshot(String)` returning a wrapper scoped to the snapshot via `BlobClient.getSnapshotClient`, or `this` when the id is empty
- `BlobOperations` — thread `withSnapshot` through the three read operations; fix the `downloadLink` URL composition to handle the snapshot-scoped URL's existing `?snapshot=<id>` query string before appending the SAS token
- Documentation — new section "Reading a specific blob snapshot" with a route example
- Regenerated catalog JSON, endpoint DSL, component DSL, configurers

### Unit tests (34 total, 0 failures)
- New `BlobOperationsTest.testGetBlobWithSnapshotIdReadsFromScopedClient()` — verifies the snapshot-scoped client is used when `snapshotId` is configured
- New `BlobOperationsTest.testGetBlobWithSnapshotIdHeaderOverridesConfig()` — verifies header precedence over URI param

### Integration tests against Azurite (new `BlobSnapshotOperationsIT`, 5 tests)
- `testCreateSnapshotReturnsId` — snapshot creation returns a non-null id and matching exchange header
- `testReadBlobViaSnapshotIdOnConfig` — `getBlob` honors `snapshotId` set on configuration
- `testReadBlobViaSnapshotIdOnHeader` — `getBlob` honors header override
- `testReadWithNonExistentSnapshotIdFails` — reading with a bogus snapshot id fails at the SDK level (proves the id is actually threaded into the request, not silently dropped)
- `testDownloadLinkContainsSnapshotQueryParam` — `downloadLink` produces a distinct URL containing the `?snapshot=` query parameter when targeting a snapshot

Note: the IT does not attempt to overwrite the live blob after taking a snapshot because Azurite rejects that with `SnapshotsPresent/409` (stricter than real Azure). The assertions still prove correctness at the URL/SDK level.

## JIRA

https://issues.apache.org/jira/browse/CAMEL-23331 (the ticket's original scope included both creation and retrieval; #22638 delivered the first half, this PR delivers the second)

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Andrea Cosentino


Croway (approved):

davsclaus (approved):

apupier (dismissed):
requires regen

apupier (commented):
Shouldn't we have an integration test using Azurite test container?

oscerd:
Yes, I'm still trying to see how to make it work with Azurite

apupier (approved):

PR #22666, opened 2026-04-17
Skip JDk version check on Jenkins CI for JDK 17

apupier:
to allow testing with JDK 17.
A more precise way would be to build with JDk 21 and launch only the test with JDK 17 but it requires far much work.

# Description

# Target

# Tracking

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,6 +17,7 @@
 def MAVEN_PARAMS = "-B -e -fae -V -Dnoassembly -Dmaven.compiler.fork=true -Dsurefire.rerunFailingTestsCount=2 -Dfailsafe.rerunFailingTestsCount=1"
 def MAVEN_TEST_PARAMS = env.MAVEN_TEST_PARAMS ?: "-Dci.env.name=apache.org"
 def MAVEN_TEST_PARAMS_UBUNTU = env.MAVEN_TEST_PARAMS ?: "-Dci.env.name=apache.org"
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

apupier:
changed the pipeline script directly to check the changes https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/ it will be number 1793

Croway (approved):

apupier:
> changed the pipeline script directly to check the changes https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/ it will be number 1793

tried a bunch of things for the pipeline, updqted the PR, this one is reflecting the nez set of change: https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/1800/

compared to initial attempt:

also launch test with jdk 21 on ppc64le and s390x qs jdk 21 is our main
target (it also avoids having specific case for them)

currently all enforcers rules are disabled with jdk17, to improve to
skip only the java check

note that I tried to use global environment conditional directly but not
managed to make it working, thus the condition at the mvnw level

Croway (approved):

Croway:
@apupier `A more precise way would be to build with JDk 21 and launch only the test with JDK 17` exactly, but there are Jenkins jobs that test with jdk21 and 25, therefore, I think we are good as is.

PR #22675, opened 2026-04-17
Using OpenTelemetry instrumentation via Spring Boot starter and OpenTelemetry2

beskow:
# Description

Document how to configure and use opentelemetry-spring-boot-starter as an alternative to the OpenTelemetry agent (see https://issues.apache.org/jira/browse/CAMEL-23343).

# Target

# Tracking

### DIFF
--- a/components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
+++ b/components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
@@ -116,8 +116,7 @@ As an alternative, you can use the agent's built-in MDC integration.
 An alternative to the OpenTelemetry agent is to use https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/[OpenTelemetry Spring Boot starter].
 
 To ensure version alignment across all OpenTelemetry dependencies, you should
-import the opentelemetry-instrumentation-bom BOM (choose a version that aligns with
-the Opentelemetry dependencies that are used within the given Camel version used).
+import the opentelemetry-instrumentation-bom BOM.
 
 [source,xml]
 ----
--- a/components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
+++ b/components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
@@ -116,7 +116,8 @@ As an alternative, you can use the agent's built-in MDC integration.
 An alternative to the OpenTelemetry agent is to use https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/[OpenTelemetry Spring Boot starter].
 
 To ensure version alignment across all OpenTelemetry dependencies, you should
-import the opentelemetry-instrumentation-bom BOM:
+import the opentelemetry-instrumentation-bom BOM (choose a version that aligns with
+the Opentelemetry dependencies that are used within the given Camel version used).
 
 [source,xml]
 ----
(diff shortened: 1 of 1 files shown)


components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
```diff
@@ -110,6 +110,78 @@ As an alternative, you can use the agent's built-in MDC integration.
...
+[source,xml]
+----
+<dependency>
+    <groupId>io.opentelemetry.instrumentation</groupId>
+    <artifactId>opentelemetry-spring-boot-starter</artifactId>
+</dependency>
+----
+
+The starter auto-configures most aspects of the OpenTelemetry instrumentation, with two exceptions:
+No `Tracer` or ContextPropagators` beans are autoconfigured, they need to be configured manually:
```
davsclaus: Missing a quote here

beskow: Thanks, fixed!

davsclaus (approved):

oscerd (approved):

components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
```diff
@@ -110,6 +110,78 @@ As an alternative, you can use the agent's built-in MDC integration.
...
+import the opentelemetry-instrumentation-bom BOM:
+
+[source,xml]
+----
+<dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>io.opentelemetry.instrumentation</groupId>
+            <artifactId>opentelemetry-instrumentation-bom</artifactId>
+            <version>2.26.1</version>
```
squakez, line 128: Probably you want to set here a version compatible with the Opentelemetry dependencies that are used within a given Camel version used. We should add a note.

squakez (commented):
LGTM, just a minor note would be required for compatibility reasons.

components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
beskow: Good point, I'll have added that!

squakez: Thanks, but I think that note is not enough. I'm adding a comment inline to show how I think it should be. Feel free to accept the inline change if you want.

components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
```diff
@@ -110,6 +110,82 @@ As an alternative, you can use the agent's built-in MDC integration.
...
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>
+    </dependencies>
+</dependencyManagement>
+----
+
+Note: Import the OpenTelemetry BOMs before any other BOMs in your project. For example, if you
+import the spring-boot-dependencies BOM, you have to declare it after the OpenTelemetry BOMs.
```
squakez:
```suggestion
import the spring-boot-dependencies BOM, you have to declare it after the OpenTelemetry BOMs. Also make sure that the version you choose is compatible with the Opentelemetry version used by the Camel version you're using.
```

components/camel-opentelemetry2/src/main/docs/opentelemetry2.adoc
beskow: Sure! I had added the same info in the paragraph above, but it might be more clear the way you suggested. Now applied, thanks!

squakez (approved):

PR #22691, opened 2026-04-20
Update Jenkinsfile to use JDK 21 for JBang tests

apupier:
# Description

given that there is now an enforcer to compile with JDK 21 by default. I propose this first iteration to jhave green build back.

ideally, we should compile with JDK 21 and tests with JDK 17, 21 and 25.
A good intermediate step which shoudl be enough would be to create a matrix jobs whcih is compiling and testing with JDK 17, 21 and 25 with the JDk 17 one skipping the enforcer java rule.

Note that I also removed the branches from the triggeres as there are specific jobs for that

# Target

# Tracking

### DIFF
--- a/Jenkinsfile.jbangtest
+++ b/Jenkinsfile.jbangtest
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 def AGENT_LABEL = env.AGENT_LABEL ?: 'ubuntu'
-def JDK_NAME = env.JDK_NAME ?: 'jdk_17_latest'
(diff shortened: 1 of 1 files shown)


apupier:
I modified the script on Jenkins on this build https://ci-builds.apache.org/job/Camel/job/Camel%20Jbang%20IT%20Tests/job/main/435/cloudbees-pipeline-explorer/

we can notice that it goes further (even if failing)

Croway (approved):

PR #22769, opened 2026-04-24
CAMEL-23352: Add syncOptimisticRetry option to Aggregate EIP

code-massel:
When optimisticLocking is enabled, retries are normally scheduled asynchronously on a background ScheduledExecutorService, which switches threads and breaks transactional use cases where the aggregation repository and inbound message store must commit/rollback within a single transaction boundary.

The new syncOptimisticRetry flag (default false) causes the retry delay to execute via Thread.sleep on the calling thread instead, preserving the transaction context. This restores the Camel 2.x behavior for users who need it, without changing the default async behavior.

# Description

When `optimisticLocking` is enabled on the Aggregate EIP, retry failures are scheduled asynchronously on a background `ScheduledExecutorService`. This thread switch breaks transactional use cases where the aggregation repository and inbound message store must commit/rollback within a single transaction boundary (transaction context is bound to the originating thread).

This PR adds a new `syncOptimisticRetry` option (default `false`) to the `aggregate` EIP. When enabled, the retry delay runs via `Thread.sleep` on the calling thread instead of scheduling on a background executor, preserving the transaction context. Default is `false` so existing routes are unaffected.

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a https://issues.apache.org/jira/browse/CAMEL-23352 filed for the change.


davsclaus:
when you add new options to EIPs you generally need to rebuild the entire code base
```
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/aggregate.json
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd
	modified:   core/camel-core-model/src/generated/resources/META-INF/org/apache/camel/model/aggregate.json
	modified:   core/camel-xml-io/src/generated/java/org/apache/camel/xml/in/ModelParser.java
	modified:   core/camel-xml-io/src/generated/java/org/apache/camel/xml/out/ModelWriter.java
	modified:   core/camel-yaml-io/src/generated/java/org/apache/camel/yaml/out/ModelWriter.java
	modified:   dsl/camel-yaml-dsl/camel-yaml-dsl-deserializers/src/generated/java/org/apache/camel/dsl/yaml/deserializers/ModelDeserializers.java
	modified:   dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl.json
```

davsclaus:
for example run from root

    mvn clean install -P fastinstall

core/camel-core-model/src/main/java/org/apache/camel/model/AggregateDefinition.java
```diff
@@ -90,6 +90,9 @@ public class AggregateDefinition extends OutputDefinition<AggregateDefinition>
     @Metadata(javaType = "java.lang.Boolean")
     private String optimisticLocking;
     @XmlAttribute
+    @Metadata(label = "advanced", javaType = "java.lang.Boolean", defaultValue = "false")
+    private String syncOptimisticRetry;
```
gnodet: Nit / naming: consider `optimisticLockingSyncRetry` (or `optimisticLockRetrySync`) to group it visually with the existing `optimisticLocking` field in docs and configuration. `syncOptimisticRetry` reads a bit ambiguously ("synchronize" vs "synchronous"). Not a blocker — just a readability suggestion.

core/camel-core-model/src/main/java/org/apache/camel/model/AggregateDefinition.java
```diff
@@ -1004,6 +1016,16 @@ public AggregateDefinition optimisticLocking() {
         return this;
     }
 
+    /**
+     * When optimistic locking is enabled, retries happen synchronously in the same thread instead of being scheduled on
+     * a background thread. This preserves transaction context for repositories that require single-thread transactional
+     * guarantees.
+     */
```
gnodet, line 1023: Nice Javadoc. Consider adding a note that this option only takes effect when `optimisticLocking` is also enabled:

```suggestion
    /**
     * When optimistic locking is enabled, retries happen synchronously in the same thread instead of being scheduled on
     * a background thread. This preserves transaction context for repositories that require single-thread transactional
     * guarantees. Only takes effect when {@link #optimisticLocking()} is also enabled.
     */
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,117 @@
...
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:start")
+                        .aggregate(header("id"), new BodyInAggregatingStrategy())
+                        .aggregationRepository(repository)
+                        .optimisticLocking()
+                        .syncOptimisticRetry()
+                        .optimisticLockRetryPolicy(
+                                new OptimisticLockRetryPolicy().maximumRetries(10).retryDelay(0))
```
gnodet, line 204: Good test coverage overall. One observation: `retryDelay(0)` means `doDelay()` returns immediately without sleeping, so the `Thread.sleep` + `InterruptedException` code path isn't exercised.

Consider adding a small test with a nonzero delay (e.g., `retryDelay(10)`) to verify the synchronous sleep actually happens on the calling thread under realistic conditions. A test that interrupts the thread during retry would further strengthen coverage of the `InterruptedException` branch. Not a blocker, but would increase confidence.

gnodet (changes requested):
## Review: CAMEL-23352 — Add `syncOptimisticRetry` option to Aggregate EIP

Thank you for this contribution! The use case is well-motivated — async optimistic locking retries breaking transaction context is a real problem, and adding an opt-in synchronous retry is a clean approach. The core implementation logic is correct.

A few items need attention before this can merge.

### Blocker: Missing generated files

The PR modifies 4 hand-written files but does not include any regenerated downstream artifacts. CI will fail at the "Fail if there are uncommitted changes" step. Please run:

```bash
mvn install -B -pl core/camel-core-model -DskipTests
```

Then check `git status` and commit all generated file changes. This will update at minimum:
- `core/camel-core-model/src/generated/` — JSON model metadata, configurers
- `dsl/camel-yaml-dsl/` — YAML DSL schema and deserializers

The auto-generated options table in the aggregate EIP docs (`aggregate-eip.adoc`) will also pick up the new option from the Javadoc after regeneration, so a separate prose change is likely not needed.

### Implementation

The core retry logic is sound:
- The `syncOptimisticRetry` branch correctly calls `doDelay(attempt)` then `continue`s the retry loop on the calling thread.
- `InterruptedException` handling correctly restores the interrupt flag and propagates via `exchange.setException()`.
- Default (`false`) preserves existing async behavior — no regression risk.
- The reifier and copy constructor correctly wire the new option.

See inline comments for minor suggestions.

core/camel-core-model/src/main/java/org/apache/camel/model/AggregateDefinition.java
davsclaus: I like the suggested rename of the option

core/camel-core-model/src/main/java/org/apache/camel/model/AggregateDefinition.java
code-massel: good point. done

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
code-massel: Added two new tests: one verifying synchronous retry with a non-zero delay stays on the calling thread, and one exercising the InterruptedException path when the thread is interrupted mid-retry

code-massel:
Thanks for the review! I've addressed your feedback in the latest push:

- Renamed syncOptimisticRetry → optimisticLockingSyncRetry across all files for consistency with the existing optimisticLocking option naming
- Updated Javadoc to clarify the flag only takes effect when optimisticLocking() is also enabled
- Added two new tests: one verifying synchronous retry with a non-zero delay stays on the calling thread, and one exercising the InterruptedException path when the thread is interrupted mid-retry
- Regenerated all downstream artifacts (catalog, YAML DSL, XML IO, XSD schemas)

Let me know if there's anything else to adjust

core/camel-core-model/src/main/java/org/apache/camel/model/AggregateDefinition.java
code-massel: renamed to optimisticLockingSyncRetry

davsclaus (approved):

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+                    e -> {
+                        e.getMessage().setHeader("id", "interrupt-group");
+                        e.getMessage().setBody("X");
+                    });
+            resultExchange.set(exchange);
+        });
+        sender.start();
+
+        // Let at least one retry attempt happen before interrupting
+        Thread.sleep(100);
```
Croway: Can you use Awaitility instead of this Thread.sleep?

Croway:
There are some uncommitted changes:
```
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd
	modified:   dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl-canonical.json
	```

**Comment by davsclaus** [comment]
@code-massel can you do a full rebuild and add the modified files to this PR

mvn clean install -P fastinstall

from the root folder

**Review by gnodet — DISMISSED** [review]
## Second Review: CAMEL-23352 — `optimisticLockingSyncRetry` option for Aggregate EIP

Thank you for addressing the first round of feedback — the rename to `optimisticLockingSyncRetry`, the improved Javadoc, and the two additional tests (nonzero delay + interrupt path) are all solid improvements. The core implementation in `AggregateProcessor.doInOptimisticLock()` looks correct: the synchronous `doDelay()` + `continue` path cleanly preserves the calling thread, and the `InterruptedException` handling is done properly with `Thread.currentThread().interrupt()`.

This second round focuses on the **test class**, specifically around style conventions and test robustness.

### Summary of Findings

| # | Severity | Finding |
|---|----------|---------||
| 1 | **Must fix** | `Thread.sleep(100)` in `testSyncRetryInterrupted` — use Awaitility (project rule, also flagged by @Croway) |
| 2 | **Must fix** | Conditional assertions silently pass — `if (aggregateThreadName != null)` and `if (result != null && ...)` mean the test can succeed without ever verifying the core behavior |
| 3 | **Should fix** | Test class and methods are `public` — newer tests in this package use package-private (see `OptimisticLockRetryPolicyTest`, `AggregateCompletionOnlyTwoTest`) |
| 4 | **Should fix** | Unused variable `callerThread` in `testSyncRetryHappensInSameThread` |

See inline comments for details and suggested fixes.

---
*This review covers project conventions and test quality. It does not replace specialized tools (CodeRabbit, SonarCloud, etc.).*

**Review by gnodet — DISMISSED** [review]
## Second Review: CAMEL-23352 — `optimisticLockingSyncRetry` option for Aggregate EIP

Thank you for addressing the first round of feedback — the rename to `optimisticLockingSyncRetry`, the improved Javadoc, and the two additional tests (nonzero delay + interrupt path) are all solid improvements. The core implementation in `AggregateProcessor.doInOptimisticLock()` looks correct: the synchronous `doDelay()` + `continue` path cleanly preserves the calling thread, and the `InterruptedException` handling is done properly with `Thread.currentThread().interrupt()`.

This second round focuses on the **test class**, specifically around style conventions and test robustness.

### Summary of Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | **Must fix** | `Thread.sleep(100)` in `testSyncRetryInterrupted` — use Awaitility (project rule, also flagged by @Croway) |
| 2 | **Must fix** | Conditional assertions silently pass — `if (aggregateThreadName != null)` and `if (result != null && ...)` mean the test can succeed without ever verifying the core behavior |
| 3 | **Should fix** | Test class and methods are `public` — newer tests in this package use package-private (see `OptimisticLockRetryPolicyTest`, `AggregateCompletionOnlyTwoTest`) |
| 4 | **Should fix** | Unused variable `callerThread` in `testSyncRetryHappensInSameThread` |

See inline comments for details and suggested fixes.

---
*This review covers project conventions and test quality. It does not replace specialized tools (CodeRabbit, SonarCloud, etc.).*

File: core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java [hunk 5]
```diff
@@ -0,0 +1,215 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests that optimistic locking retries happen synchronously in the same thread when optimisticLockingSyncRetry is
+ * enabled.
+ */
+public class AggregateOptimisticLockSyncRetryTest extends ContextTestSupport {
```
gnodet: **Style**: In this package, newer tests use package-private visibility (see `OptimisticLockRetryPolicyTest` and `AggregateCompletionOnlyTwoTest`). JUnit 5 does not require `public` on classes or test methods.

```suggestion
class AggregateOptimisticLockSyncRetryTest extends ContextTestSupport {
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+            aggregateThreadName = Thread.currentThread().getName();
+            if (count <= FAIL_FIRST_N_ATTEMPTS) {
+                throw new OptimisticLockingException();
+            }
+            return super.add(camelContext, key, oldExchange, newExchange);
+        }
+    };
+
+    @Test
+    public void testSyncRetryHappensInSameThread() throws Exception {
```
gnodet: Same visibility nit — `public` is not needed on test methods with JUnit 5.

```suggestion
    void testSyncRetryHappensInSameThread() throws Exception {
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        MockEndpoint mock = getMockEndpoint("mock:result");
+        mock.expectedMessageCount(1);
+
+        String callerThread = Thread.currentThread().getName();
```
gnodet: **Dead code**: `callerThread` is assigned but never referenced in any assertion. Remove it, or better yet, use it to strengthen the test by asserting the retry thread matches the caller thread.

```suggestion
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        // (the route's thread), NOT in the AggregateOptimisticLockingExecutor thread pool.
+        // The key assertion is that the thread name does NOT contain the async executor name.
+        if (aggregateThreadName != null) {
+            assertFalse(aggregateThreadName.contains("AggregateOptimisticLockingExecutor"),
+                    "Expected synchronous retry but found async executor thread: " + aggregateThreadName);
+        }
+    }
+
+    @Test
+    public void testSyncRetryCompletes() throws Exception {
```
gnodet:
```suggestion
    void testSyncRetryCompletes() throws Exception {
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+        // Reset counter so next message triggers failures and retries
+        addCounter.set(0);
+        template.sendBodyAndHeader("direct:start", "B", "id", 1);
+
+        mock.assertIsSatisfied();
+    }
+
+    @Test
+    public void testSyncRetryWithNonZeroDelayStaysOnCallingThread() throws Exception {
```
gnodet:
```suggestion
    void testSyncRetryWithNonZeroDelayStaysOnCallingThread() throws Exception {
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+        mock.assertIsSatisfied();
+
+        assertTrue(delayedCounter.get() > 2, "Expected retries to occur");
+        assertFalse(retryThreadName.get().contains("AggregateOptimisticLockingExecutor"),
+                "Expected synchronous retry thread but got: " + retryThreadName.get());
+    }
+
+    @Test
+    public void testSyncRetryInterrupted() throws Exception {
```
gnodet:
```suggestion
    void testSyncRetryInterrupted() throws Exception {
```

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
gnodet: **Project rule violation** (also flagged by @Croway): `Thread.sleep()` is not allowed in test code — use [Awaitility](https://github.com/awaitility/awaitility) instead (see CLAUDE.md "Asynchronous Testing" section).

The intent here is to wait until at least one retry attempt has occurred before interrupting. This can be done deterministically:

```suggestion
        await().atMost(5, TimeUnit.SECONDS)
               .untilAsserted(() -> assertTrue(interruptCounter.get() > 0,
                       "Expected at least one retry attempt before interrupting"));
```

Requires adding:
```java
import static org.awaitility.Awaitility.await;
import java.util.concurrent.TimeUnit;
```

gnodet (changes requested):
## Second Review: CAMEL-23352 — `optimisticLockingSyncRetry` option for Aggregate EIP

Thank you for addressing the first round of feedback — the rename to `optimisticLockingSyncRetry`, the improved Javadoc, and the two additional tests (nonzero delay + interrupt path) are all solid improvements. The core implementation in `AggregateProcessor.doInOptimisticLock()` looks correct: the synchronous `doDelay()` + `continue` path cleanly preserves the calling thread, and the `InterruptedException` handling is done properly with `Thread.currentThread().interrupt()`.

This second round focuses on the **test class**, specifically around style conventions and test robustness.

### Summary of Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | **Must fix** | `Thread.sleep(100)` in `testSyncRetryInterrupted` — use Awaitility (project rule, also flagged by @Croway) |
| 2 | **Must fix** | Conditional assertions silently pass — `if (aggregateThreadName != null)` and `if (result != null && ...)` mean the test can succeed without ever verifying the core behavior |
| 3 | **Should fix** | Test class and methods are `public` — newer tests in this package use package-private (see `OptimisticLockRetryPolicyTest`, `AggregateCompletionOnlyTwoTest`) |
| 4 | **Should fix** | Unused variable `callerThread` in `testSyncRetryHappensInSameThread` |

See inline comments for details and suggested fixes.

---
*This review covers project conventions and test quality. It does not replace specialized tools (CodeRabbit, SonarCloud, etc.).*

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        assertTrue(addCounter.get() > FAIL_FIRST_N_ATTEMPTS,
+                "Expected more than " + FAIL_FIRST_N_ATTEMPTS + " attempts, got " + addCounter.get());
+
+        // Since optimisticLockingSyncRetry is enabled, the retry should happen in a Camel thread
+        // (the route's thread), NOT in the AggregateOptimisticLockingExecutor thread pool.
+        // The key assertion is that the thread name does NOT contain the async executor name.
+        if (aggregateThreadName != null) {
+            assertFalse(aggregateThreadName.contains("AggregateOptimisticLockingExecutor"),
+                    "Expected synchronous retry but found async executor thread: " + aggregateThreadName);
+        }
```
gnodet: **Silent pass**: If `aggregateThreadName` is `null` (e.g., the retry path was never hit), this assertion is skipped and the test passes without verifying anything. Make it unconditional:

```suggestion
        assertNotNull(aggregateThreadName, "Expected aggregateThreadName to be captured during retry");
        assertFalse(aggregateThreadName.contains("AggregateOptimisticLockingExecutor"),
                "Expected synchronous retry but found async executor thread: " + aggregateThreadName);
```

(also add `import static org.junit.jupiter.api.Assertions.assertNotNull;`)

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        Thread.sleep(100);
+        sender.interrupt();
+        sender.join(5000);
+
+        // The exchange should have been completed with an InterruptedException
+        Exchange result = resultExchange.get();
+        if (result != null && result.getException() != null) {
+            assertTrue(result.getException() instanceof InterruptedException,
+                    "Expected InterruptedException but got: " + result.getException());
+        }
```
gnodet: **Silent pass**: If `result` is `null` or has no exception, this test passes without verifying the interrupt behavior at all. Make the assertions unconditional:

```suggestion
        assertNotNull(result, "Expected exchange result from interrupted sender");
        assertNotNull(result.getException(), "Expected an exception on the exchange after interrupt");
        assertInstanceOf(InterruptedException.class, result.getException(),
                "Expected InterruptedException but got: " + result.getException());
```

(`assertInstanceOf` is available in JUnit 5.8+ and reads more cleanly than `assertTrue(... instanceof ...)`)

code-massel:
Thanks for the second round of feedback — all items addressed and pushed (10bd777b303).

**Test class fixes:**

1. Replaced `Thread.sleep(100)` in `testSyncRetryInterrupted` with `Awaitility.await().atMost(5s).until(() -> interruptCounter.get() > 0)` — no more timing-dependent sleep before interrupt.
2. Replaced conditional `if (aggregateThreadName != null)` and `if (result != null && ...)` assertions with `assertNotNull(...)` + unconditional assertions — tests now actually fail if the core behavior is not exercised.
3. Removed `public` from class and all test methods — now package-private, consistent with `AggregateCompletionOnlyTwoTest` and others in the same package.
4. Removed unused `callerThread` variable in `testSyncRetryHappensInSameThread`.

**Missing generated file:**

`camelYamlDsl-canonical.json` was not present on the branch because the file was added to `main` after the branch was created. Rebased the branch onto current `main` and regenerated — the file now includes the `optimisticLockingSyncRetry` option (6 insertions).

All 4 tests pass locally. Let me know if there is anything else to address.

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
code-massel: done

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
code-massel: done

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
code-massel: done

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateOptimisticLockSyncRetryTest.java
code-massel: done

davsclaus:
Sorry but its safer to do a full rebuild when you change EIPs or whatnot as it affect other DSLs, catalog etc

There are uncommitted changes
HEAD detached at pull/22769/merge
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd

davsclaus (approved):

PR #22788, opened 2026-04-27
CAMEL-23370 fix `camel.main.routes-include-pattern` usage with multi-level wildcard matchers in springboot jars

vilmosnagy:
Using `classpath:**/*.xml` (like any xml in any dir) was broken since _probably_ CAMEL-21944 & #17716

# Description

See [CAMEL-23370](https://issues.apache.org/jira/browse/CAMEL-23370); but it seems to me that multi-level include-route matchers are broken since #17716.

This worked with Camel 4.11:

```yaml
camel:
  main:
    routes-include-pattern: classpath:camel/**/*.xml
```

and does not work with latest snapshot.

I'm not sure how to write tests for my use-case though; and not sure if it ain't broke anything, 'cause there seem to bit not much tests around this part of the codebase.

# Target

# Tracking

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/scan/DefaultPackageScanResourceResolver.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/scan/DefaultPackageScanResourceResolver.java
@@ -216,7 +216,7 @@ public class DefaultPackageScanResourceResolver extends BasePackageScanResolver
             boolean match = PATH_MATCHER.match(subPattern, shortName);
             LOG.debug("Found resource: {} matching pattern: {} -> {}", shortName, subPattern, match);
             if (match) {
-                Resource resource = new PackageScanJarResource("jar", url, name);
+                Resource resource = new PackageScanJarResource("jar", url, shortName);
                 resources.add(resource);
             }
         }
--- a/core/camel-support/src/main/java/org/apache/camel/support/scan/PackageScanJarResource.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/scan/PackageScanJarResource.java
@@ -25,7 +25,6 @@ import java.net.URL;
 import java.net.URLClassLoader;
 
 import org.apache.camel.support.ResourceSupport;
-import org.apache.camel.util.FileUtil;
 import org.apache.camel.util.StringHelper;
 
 public class PackageScanJarResource extends ResourceSupport {
@@ -33,8 +32,8 @@ public class PackageScanJarResource extends ResourceSupport {
     private final URL url;
     private final URLClassLoader uc;
 
-    public PackageScanJarResource(String scheme, URL url, String location) {
-        super(scheme, url.getFile() + FileUtil.stripPath(location));
+    public PackageScanJarResource(String scheme, URL url, String shortName) {
+        super(scheme, url.getFile() + shortName);
         this.url = url;
         this.uc = new URLClassLoader(new URL[] { url });
     }


davsclaus (approved):

Croway:
Hi @vilmosnagy thanks for the fix. You can add a test in the camel-spring-boot repository, for example, similar to the following one https://github.com/apache/camel-spring-boot/blob/4bdea8a87e75c67d0a242752264f42646ea8551b/core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelXmlRoutesTest.java#L35

vilmosnagy:
Hi @Croway, thanks for the direction!

I've added a test in my fork:  
https://github.com/apache/camel-spring-boot/compare/main...vilmosnagy:camel-spring-boot:vilmos/testing-wildcard-globs

My understanding is that this bug is only triggered when Camel is run via `java -jar`. All tests in my production project pass when run with the Maven Surefire Plugin using glob matchers, and the issue only appears when starting the application on Kubernetes.

From debugging, it seems that when running with Surefire, `file.isDirectory()` returns `true` here:  
https://github.com/apache/camel/blob/main/core/camel-support/src/main/java/org/apache/camel/support/scan/DefaultPackageScanResourceResolver.java#L165-L168  
which causes a different code path to be executed.

The only way I was able to reproduce the faulty behavior was:
- adding a new module called `test-resources-in-jar` and placing the route there  
- adding this module as a test-scoped dependency to the `camel-spring-boot` module  
- running the tests via the Maven CLI (in this case, they fail)

Interestingly, my JetBrains IDE seems to use a different classloading approach and still loads these resources directly from the filesystem rather than from the JAR, so the issue does not appear there.

If you have a better idea for reproducing this, please let me know. Also, `camel-observation-starter` has failing tests on my machine on the `main` branch, so I’m not entirely sure all tests ran successfully—but as far as I can tell, my change resolves this issue.

davsclaus:
yeah its not easy to test and the idea of building a -jar as another maven module is correct. I wonder if the test can be moved into the `tests` folder instead so its not in the core. And this would then require to add a new module specially for this test `camel-itest-spring-boot-farjar` or a name like that. Then we can add other farjar tests in the future

vilmosnagy:
@davsclaus I almost have the test done: do you have any example on how can I set a custom `camel.main.routes-include-pattern` value in the `camel-itest-spring-boot` package, or should I create a new way for that? I have ideas, but in a hurry right now. If you don't have a ready-to-copy solution, I'll describe my idea over the weekend.

davsclaus:
I am not sure how you run the test, though the regular unit tests for SB allow to declare custom values for its `application.properties` in those annotations. 

But feel free to provide what you have and we can take a look to help last bit. But we dont have a release in a hurry so there is plenty of time to get this done over the following weeks

vilmosnagy:
I hope this is the test you were thinking of: https://github.com/apache/camel-spring-boot/pull/1773

At first, I tried to write a test similar to the existing ones in the `camel-itest-spring-boot` module. As far as I understand, they don’t use the `@SpringBootTest` annotation (or other Spring Boot testing features). Instead, they generate a temporary Maven project and bootstrap a Spring Boot application from the built artifacts of that project.

I wasn’t able to write a similar test for loading XML routes from another module, because the `test-resources-in-jar` JAR would not be accessible from the generated Maven project in that setup.

In the end, I moved my previous attempt to the `tests` module and added test-scoped dependencies to the `camel-itest-spring-boot` module. This allows running tests annotated with `@SpringBootTest`.

I’m not fully happy with this solution, but I don’t see a better alternative at the moment. This still applies from my previous comment:

> Interestingly, my JetBrains IDE seems to use a different classloading approach and still loads these resources directly from the filesystem rather than from the JAR, so the issue does not appear there.

Croway:
I reviewed https://github.com/apache/camel-spring-boot/pull/1773 . I think that we can merge it, and have the tests merged later. Thanks for your contributions @vilmosnagy

davsclaus:
ok

PR #22793, opened 2026-04-28
feat(components): log on shutdown prometheus format

squakez:
Possibility to select json (default) or prometheus format when logging metric traces

With this PR we give the possibility to select wich format to use. By default `camel-observability-services` will use `prometheus` in order to simplify the metrics extraction from external tooling (eg, camel-monitor-operator):

```
2026-04-28 11:08:27.574  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : Micrometer service is stopping, here a list of metrics collected so far.
2026-04-28 11:08:27.608  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : #METRIC-START## HELP app_info  #METRIC-END#
2026-04-28 11:08:27.608  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : #METRIC-START## TYPE app_info gauge#METRIC-END#
2026-04-28 11:08:27.608  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : #METRIC-START#app_info{camel_context="camel-1",camel_runtime_provider="Main",camel_runtime_version="4.21.0-SNAPSHOT",camel_version="4.21.0-SNAPSHOT"} 1#METRIC-END#
2026-04-28 11:08:27.611  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : #METRIC-START## HELP jvm_memory_max_bytes The maximum amount of memory in bytes that can be used for memory management#METRIC-END#
2026-04-28 11:08:27.611  WARN 247539 --- [           main] crometer.prometheus.MicrometerPrometheus : #METRIC-START## TYPE jvm_memory_max_by
```

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-micrometer-prometheus/src/main/java/org/apache/camel/component/micrometer/prometheus/MicrometerPrometheus.java
+++ b/components/camel-micrometer-prometheus/src/main/java/org/apache/camel/component/micrometer/prometheus/MicrometerPrometheus.java
@@ -106,6 +106,8 @@ public class MicrometerPrometheus extends ServiceSupport implements CamelMetrics
     private boolean skipCamelInfo = false;
     @Metadata(defaultValue = "false")
     private boolean logMetricsOnShutdown = false;
+    @Metadata(defaultValue = "json", enums = "json,prometheus")
+    private String logMetricsOnShutdownFormat = "json";
     @Metadata(defaultValue = "0.0.4", enums = "0.0.4,1.0.0")
     private String textFormatVersion = "0.0.4";
     @Metadata
@@ -307,6 +309,18 @@ public class MicrometerPrometheus extends ServiceSupport implements CamelMetrics
         this.logMetricsOnShutdownFilters = logMetricsOnShutdownFilters;
     }
 
+    public String getLogMetricsOnShutdownFormat() {
+        return logMetricsOnShutdownFormat;
+    }
+
+    /**
+     * Format of metrics to log when application is shutting down. Either `json` (default) or `prometheus` format
+     * accepted.
+     */
+    public void setLogMetricsOnShutdownFormat(String logMetricsOnShutdownFormat) {
+        this.logMetricsOnShutdownFormat = logMetricsOnShutdownFormat;
+    }
+
     @Override
     protected void doInit() throws Exception {
         super.doInit();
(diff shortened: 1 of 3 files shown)


components/camel-micrometer-prometheus/src/main/java/org/apache/camel/component/micrometer/prometheus/MicrometerPrometheus.java
```diff
@@ -482,6 +500,45 @@ private void initBinders() throws IOException {
...
+    }
+
+    void logMetricsOnShutdown(String... filters) {
+        String[] scrapes = meterRegistry.scrape().split("\n");
+        for (String s : scrapes) {
+            if (matchesFilter(s, filters)) {
+                // We put on warn level to make sure it is printed even if the log is
+                // at higher levels. Important: we also include a start and end tag to make sure the
+                // scraper can more easily identify the metric content.
+                LOG.warn("#METRIC-START#" + s + "#METRIC-END#");
```
davsclaus, line 521: I wonder if anyone really turn of INFO logging in production?

Just wonder if WARN logging always is causing some concern for end users when they see that.

davsclaus (approved):

Croway:
@squakez does `logmetricsonshutdownformat` need to be mirrored in https://github.com/apache/camel/blob/46d917320b4fb8c0c3d417f9d3b734bb1617c072/core/camel-main/src/main/java/org/apache/camel/main/MetricsConfigurationProperties.java#L27 ?

components/camel-micrometer-prometheus/src/main/java/org/apache/camel/component/micrometer/prometheus/MicrometerPrometheus.java
squakez: I worked in companies that did that for performance reasons. It should not be a concern in this case as the log only trace once before shutting down.

squakez:
> @squakez does `logmetricsonshutdownformat` need to be mirrored in
> 
> https://github.com/apache/camel/blob/46d917320b4fb8c0c3d417f9d3b734bb1617c072/core/camel-main/src/main/java/org/apache/camel/main/MetricsConfigurationProperties.java#L27
> ?

It depends. Right now the main user of this property is the observability services. If we want to expose that in main, then we should include it there. However, I'd avoid that and just stick to the usage of this feature via observability services (unless some user is requiring that on purpose).

components/camel-micrometer-prometheus/src/main/java/org/apache/camel/component/micrometer/prometheus/MicrometerPrometheus.java
davsclaus: ok

Croway (approved):

PR #22811, opened 2026-04-28
Fix camel-joor MultiCompile classpath for Spring Boot fat-jar nested JARs

smongiar:
## Summary

- Fix `MultiCompile.java` to handle non-`file://` URLs (e.g. `jar:nested:`) in the compiler classpath
- When running inside a Spring Boot fat-jar, `URLClassLoader.getURLs()` returns `jar:nested:` protocol URLs for JARs inside `BOOT-INF/lib/`
- The existing code at line 131 silently drops these URLs (`if ("file".equals(url.getProtocol()))`), causing `javax.tools.JavaCompiler` to fail with missing classes
- Fix extracts non-file URLs to temporary JAR files and adds them to the compiler classpath, with cleanup in a finally block

## Root cause

The `camel-joor` runtime Java compiler builds its `-classpath` from `URLClassLoader.getURLs()`. In a standard flat JAR, these are `file://` URLs. In a Spring Boot repackaged JAR (nested format), they are `jar:nested:/path/to/app.jar/!BOOT-INF/lib/some-dep.jar` — which get silently skipped.

This manifests as compilation failures for auto-configure scripts like `spring.datasource.url.java` with errors like:
```
error: package org.apache.camel.main does not exist
```

### DIFF
--- a/components/camel-joor/src/main/java/org/apache/camel/language/joor/MultiCompile.java
+++ b/components/camel-joor/src/main/java/org/apache/camel/language/joor/MultiCompile.java
@@ -18,6 +18,8 @@ package org.apache.camel.language.joor;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.StringWriter;
 import java.lang.invoke.MethodHandles;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus:
If this makes it possible on SB then we can remove the limitions at for CSB at
https://camel.apache.org/components/4.18.x/languages/java-language.html#_limitations

davsclaus (approved):

PR #22835, opened 2026-04-29
Fix camel hawtio to explicitly set host address for Jetty binding

smongiar:
## Summary

- Add explicit `setHost("127.0.0.1")` call when launching hawtio-embedded, instead of relying on its default `"localhost"`
- Add `--host` CLI option to `camel hawtio` command for user-configurable binding
- Update browser URL to use the configured host instead of hardcoded `localhost`

## Problem

On s390x and ppc64le architectures, `camel hawtio` fails with a `ConnectException` caused by `ClosedChannelException`. The root cause is that hawtio-embedded's default host `"localhost"` is resolved via `InetAddress.getByName("localhost")`, which may return `::1` (IPv6 loopback) on these systems. If IPv6 is disabled or NIO sockets behave differently, Jetty binds to an unreachable address.

The `camel jolokia` command (JVM agent, no Jetty) works fine on the same systems.

## Error trace (s390x)

```
java.net.ConnectException
Caused by: java.nio.channels.ClosedChannelException
  at sun.nio.ch.SocketChannelImpl.ensureOpen(...)
  at sun.nio.ch.SocketChannelImpl.beginConnect(...)
```

## Fix

Use explicit `127.0.0.1` (IPv4 loopback) as the default host to avoid ambiguous `localhost` resolution across architectures.


davsclaus (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/Hawtio.java
```diff
@@ -116,14 +120,16 @@ protected Integer callHawtio() throws Exception {
...
+                ObjectHelper.invokeMethod(m, hawt, host);
                 m = clazz.getMethod("setPort", Integer.class);
                 ObjectHelper.invokeMethod(m, hawt, port);
                 m = clazz.getMethod("run");
                 ObjectHelper.invokeMethod(m, hawt);
 
                 if (openUrl) {
                     // open web browser
-                    String url = "http://localhost:" + port + "/hawtio";
+                    String url = "http://" + host + ":" + port + "/hawtio";
```
Croway, line 132: is it ok to change from localhost, to the default 127.0.0.1?

smongiar: It should be the default value, yet. It's at line [46](https://github.com/apache/camel/pull/22835/changes#diff-58a14a72becb5f8554757084eb407003122e4d8539705f493f8739c07f6cec39R46). Am I wrong?  `127.0.0.1` is intentional to avoid IPv6 resolution issues on s390x/ppc64le

Croway: yes, but it was "http://localhost:" previously, now the default will be "http://127.0.0.1:"

smongiar: That's ok, it's related to IPv6 resolution issue

Croway (approved):

davsclaus:
modified:   dsl/camel-jbang/camel-jbang-core/src/generated/resources/META-INF/camel-jbang-commands-metadata.json

davsclaus:
more modified files

	modified:   docs/user-manual/modules/ROOT/pages/jbang-commands/camel-jbang-hawtio.adoc
	modified:   dsl/camel-jbang/camel-jbang-core/src/generated/resources/META-INF/camel-jbang-commands-metadata.json

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/Hawtio.java
davsclaus: yeah 127.0.0.1 is better to use than localhost

davsclaus:
@smongiar there are still a few modified files, its often just easier to do a full rebuild from root

mvn clean install -P fastinstall

davsclaus (approved):

PR #22840, opened 2026-04-29
camel-salesforce: Improve error resilience in subscription failure handling

shaipan:
Improve error handling in SubscriptionHelper to recover from more transient Salesforce
  Streaming API failures instead of aborting the subscription.

  - Treat IOException, TransportException, and TimeoutException as retryable temporary errors
  - Retry with backoff on authentication (401), authorization (403), security policy denial,
    org daily event limit, and missing error message failures
  - Handle invalid channel errors (400) by removing the channel from the subscription list
    instead of retrying indefinitely

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
@@ -29,6 +29,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
@@ -50,6 +51,7 @@ import org.cometd.client.BayeuxClient;
 import org.cometd.client.BayeuxClient.State;
 import org.cometd.client.http.jetty.JettyHttpClientTransport;
 import org.cometd.client.transport.ClientTransport;
+import org.cometd.common.TransportException;
 import org.eclipse.jetty.client.Request;
 import org.eclipse.jetty.http.HttpCookie;
 import org.eclipse.jetty.http.HttpCookieStore;
@@ -82,6 +84,10 @@ public class SubscriptionHelper extends ServiceSupport {
     private static final String SERVER_TOO_BUSY_ERROR = "503::";
     private static final String AUTHENTICATION_INVALID = "401::Authentication invalid";
     private static final String INVALID_REPLAY_ID_PATTERN = "400::The replayId \\{.*} you provided was invalid.*";
+    private static final String CHANNEL_INVALID_PATTERN = "400::The channel.*";
+    private static final String DENIED_BY_SEC_POLICY = "403:denied_by_security_policy";
+    private static final String AUTHORIZATION_ERROR = "403::";
+    private static final String ORG_LIMIT_ERROR = "403::Organization total events daily limit exceeded";
 
     BayeuxClient client;
 
@@ -219,7 +225,9 @@ public class SubscriptionHelper extends ServiceSupport {
         boolean abort = true;
 
         LOG.warn(msg);
-        if (isTemporaryError(message)) {
+        if (isTemporaryError(message) || error.equals(AUTHENTICATION_INVALID) || error.startsWith(DENIED_BY_SEC_POLICY)
+                || error.startsWith(AUTHORIZATION_ERROR) || error.equals(ORG_LIMIT_ERROR)
+                || error.equals("Missing error message")) {
 
             // retry after delay
             final long backoff = handshakeBackoff.getAndAdd(backoffIncrement);
@@ -229,8 +237,6 @@ public class SubscriptionHelper extends ServiceSupport {
                 abort = false;
 
                 LOG.debug("Pausing for {} msecs before subscribe attempt", backoff);
-                // Use Camel's task API for backoff delay instead of Thread.sleep()
-                // We use initialDelay for the actual delay, and maxIterations(1) to run once
                 Tasks.foregroundTask()
                         .withBudget(Budgets.iterationBudget()
                                 .withMaxIterations(1)
@@ -255,6 +261,9 @@ public class SubscriptionHelper extends ServiceSupport {
             for (var consumer : consumers) {
                 subscribe(consumer);
             }
+        } else if (error.matches(CHANNEL_INVALID_PATTERN)) {
+            LOG.warn("Channel invalid for channel {}, removing from subscription list", channelName);
+            channelsToSubscribe.remove(channelName);
         }
 
         if (abort && client != null) {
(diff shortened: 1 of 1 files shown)


davsclaus:
@jeremyross just in case you have a bit time to quickly take a look

davsclaus (approved):

oscerd (approved):

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
```diff
@@ -219,7 +225,9 @@ private void subscriptionFailed(StreamingApiConsumer firstConsumer, Message mess
         boolean abort = true;
 
         LOG.warn(msg);
-        if (isTemporaryError(message)) {
+        if (isTemporaryError(message) || error.equals(AUTHENTICATION_INVALID) || error.startsWith(DENIED_BY_SEC_POLICY)
+                || error.startsWith(AUTHORIZATION_ERROR) || error.equals(ORG_LIMIT_ERROR)
```
gnodet: Nit: `AUTHORIZATION_ERROR` is `"403::"` and `ORG_LIMIT_ERROR` is `"403::Organization total events daily limit exceeded"`. Since `ORG_LIMIT_ERROR` starts with `"403::"`, the `startsWith(AUTHORIZATION_ERROR)` check already covers it — the `equals(ORG_LIMIT_ERROR)` is redundant.

```suggestion
                || error.startsWith(AUTHORIZATION_ERROR)
```

(Note: `DENIED_BY_SEC_POLICY` uses a single colon `"403:denied_by_security_policy"` so its separate `startsWith` check is correctly needed.)

gnodet (commented):
Thanks for improving the error resilience in the Salesforce Streaming subscription handling, @shaipan! This addresses a real operational pain point. A few observations:

**Missing tests**: Per [project standards](https://github.com/apache/camel/blob/main/CLAUDE.md), every PR should include tests for new behavior. The new error handling paths in `subscriptionFailed()` and the extended `isTemporaryError()` method are not covered by the existing `SubscriptionHelperTest`. Unit tests verifying that `IOException`, `TransportException`, and `TimeoutException` are treated as temporary, and that the various 401/403 error strings trigger retry, would strengthen confidence in these changes.

**401/403 retry without token refresh**: Treating `AUTHENTICATION_INVALID` (401) and authorization errors (403) as retryable is reasonable for transient failures. However, the retry path re-calls `subscribe()` without explicitly triggering a session token refresh. If the token is genuinely expired, all retries may fail identically, consuming the full backoff budget before aborting. Worth confirming that the CometD handshake/subscribe path handles token refresh internally.

**Minor**: Two comments explaining the non-obvious `Tasks.foregroundTask()` usage (using `initialDelay` as a sleep and `maxIterations(1)` to run once) were removed. The pattern isn't self-documenting — consider keeping them.

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
```diff
@@ -219,7 +225,9 @@ private void subscriptionFailed(StreamingApiConsumer firstConsumer, Message mess
...
+                || error.equals("Missing error message")) {
```
gnodet: This matches the synthetic default set earlier when `error == null`:
```java
if (error == null) {
    error = "Missing error message";
}
```
Previously, a message with **no error field at all** would abort the subscription. Now it silently retries with backoff. This is a significant behavioral change — unknown/unclassified failures become retryable instead of surfacing to the consumer's exception handler. If there's genuinely no error information, retrying blindly (up to `maxBackoff`) may mask real problems.

Was this intentional? If so, it might be worth a comment explaining the rationale.

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
```diff
@@ -255,6 +261,9 @@ private void subscriptionFailed(StreamingApiConsumer firstConsumer, Message mess
             for (var consumer : consumers) {
                 subscribe(consumer);
             }
+        } else if (error.matches(CHANNEL_INVALID_PATTERN)) {
+            LOG.warn("Channel invalid for channel {}, removing from subscription list", channelName);
+            channelsToSubscribe.remove(channelName);
```
gnodet, line 266: In this branch, `abort` stays `true`, so the code falls through to:
```java
if (abort && client != null) {
    consumer.getExceptionHandler().handleException(...);
}
```
Meaning the channel is removed from `channelsToSubscribe` **and** all consumers for that channel receive an exception. Is that the intended behavior? If so, it makes sense (clean up + notify), but setting `abort = false` after the removal would be the alternative if the intent is to silently drop the bad channel.

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
shaipan: > Nit: `AUTHORIZATION_ERROR` is `"403::"` and `ORG_LIMIT_ERROR` is `"403::Organization total events daily limit exceeded"`. Since `ORG_LIMIT_ERROR` starts with `"403::"`, the `startsWith(AUTHORIZATION_ERROR)` check already covers it — the `equals(ORG_LIMIT_ERROR)` is redundant.
> 
> (Note: `DENIED_BY_SEC_POLICY` uses a single colon `"403:denied_by_security_policy"` so its separate `startsWith` check is correctly needed.)

@gnodet  Fixed this

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
shaipan: removed

components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/streaming/SubscriptionHelper.java
shaipan: Keeping abort = true here intentionally. An invalid channel is a permanent error. consumers should be notified so the issue surfaces rather than being silently ignored.

shaipan:
> Thanks for improving the error resilience in the Salesforce Streaming subscription handling, @shaipan! This addresses a real operational pain point. A few observations:
> 
> **Missing tests**: Per [project standards](https://github.com/apache/camel/blob/main/CLAUDE.md), every PR should include tests for new behavior. The new error handling paths in `subscriptionFailed()` and the extended `isTemporaryError()` method are not covered by the existing `SubscriptionHelperTest`. Unit tests verifying that `IOException`, `TransportException`, and `TimeoutException` are treated as temporary, and that the various 401/403 error strings trigger retry, would strengthen confidence in these changes.
> 
> **401/403 retry without token refresh**: Treating `AUTHENTICATION_INVALID` (401) and authorization errors (403) as retryable is reasonable for transient failures. However, the retry path re-calls `subscribe()` without explicitly triggering a session token refresh. If the token is genuinely expired, all retries may fail identically, consuming the full backoff budget before aborting. Worth confirming that the CometD handshake/subscribe path handles token refresh internally.
> 
> **Minor**: Two comments explaining the non-obvious `Tasks.foregroundTask()` usage (using `initialDelay` as a sleep and `maxIterations(1)` to run once) were removed. The pattern isn't self-documenting — consider keeping them.
> 

@gnodet  The methods are private and the existing test pattern in the file doesn't cover listener callbacks. 
Reverted the comments that were removed in earlier PR.

davsclaus:
Thanks @shaipan

shaipan:
Hi @gnodet @davsclaus are we good to commit this? Or any more chanegs needed from my side. I have 2 3 more bugs in this area to commit.

davsclaus (approved):

davsclaus:
I created a ticket to capture your PRs in CAMEL-23391 as we need this for the release notes to track notice changes in this component

PR #22887, opened 2026-05-03
CAMEL-23402: Route structure dump to walk model for structure to be accurate

davsclaus:
# Description

# Target

# Tracking


gnodet (commented):
## Review: CAMEL-23402 — Route structure dump to walk model

The core change (model-based walking instead of JMX MBeans) is well-motivated and aligned with the JIRA description. Removes the JMX dependency entirely, which means the structure dump now works even when JMX is disabled. CI is green across JDK 17, 21, and 25.

### Findings

**1. Source-breaking API change — `ModelDumpLine` record (High)**
`ModelDumpLine` has been a released public API since 4.16.0 (5 releases). Adding the `description` field changes the positional constructor from 5 to 6 parameters, which will break any external code constructing `ModelDumpLine` directly at compile time.

**2. Missing test for the core JIRA scenario (Medium)**
CAMEL-23402 specifically says the dumper does not include `doCatch`. The new test covers `choice`/`when`/`otherwise`, but there is no test for `doTry`/`doCatch`/`doFinally` — the exact scenario the JIRA describes.

**3. `from` line inherits route description — intentional? (Low)**
In `DefaultModelToStructureDumper`, both the `route` line and the `from` line receive `def.getDescription()`. The `from` endpoint is not the route itself — should this be `null` or `def.getInput().getDescription()` instead?

**4. `getLevel()` generalization (Question)**
The level calculation has special-casing for `choice`/`when`/`otherwise`. Other container EIPs (`doTry`, `split`, `multicast`, `loop`, `aggregate`) may need similar treatment. Without broader tests it's hard to tell if the current logic generalizes correctly.

### Positive observations
- No more JMX dependency — significant improvement
- `ProcessorDefinitionHelper.doFindType` correctly includes `OtherwiseDefinition` itself now
- `CatchDefinition` label fix is cleaner
- Good test structure with deterministic node IDs
- Dump moved before route startup helps troubleshooting startup failures

oscerd (approved):

PR #22926, opened 2026-05-04
chore: Replace commons-dbcp with HikariCP in camel-jooq

tmielke:
Replaced the unmaintained commons-dbcp:1.4 library with HikariCP 7.0.2 for JDBC connection pooling in camel-jooq test infrastructure.

Changes:
- Updated parent POM to include hikaricp-version property (7.0.2)
- Replaced commons-dbcp dependency with HikariCP (test scope)
- Updated Spring XML configuration to use HikariDataSource
- Updated connection pool properties (maximumPoolSize, minimumIdle)
- Updated documentation example to reflect HikariCP usage
- All tests pass successfully

Made with help from AI tools.

### DIFF
--- a/components/camel-jooq/pom.xml
+++ b/components/camel-jooq/pom.xml
@@ -54,9 +54,10 @@
             <version>${jakarta-annotation-api-version}</version>
         </dependency>
         <dependency>
-            <groupId>commons-dbcp</groupId>
-            <artifactId>commons-dbcp</artifactId>
-            <version>${commons-dbcp-version}</version>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>${hikaricp-version}</version>
+            <scope>test</scope>
         </dependency>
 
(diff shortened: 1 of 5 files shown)


apupier:
Given that we are using commons-dbcp2 for other components, I'm wondering if it should not try to use dbcp2 instead of a completely different library.

dbcp2 is not very active but still has regular releases

tmielke:
True, dbcp2 would have been an option as well. I evaluated between HikariCP and Agroal but it seems HikariCP is even more popular and wide spread these days. Given the PR also updates the component documentation, it might be worth recommending today's modern connection pooling libraries?

apupier:
Quarkus is using Agroal as far as I know. it is mentioning it here https://quarkus.io/guides/datasource so if changing from dbcp, i think that moving to agroal seems a better fit than HikariCP. (would worth to validate with Quarkus team)

tmielke:
After further discussion I will rewrite a new proposal and use Agroal instead of HikariCP.

PR #22935, opened 2026-05-05
Enforce import style: auto-shorten FQCNs via OpenRewrite in CI

gnodet:
## Summary

- Add `rewrite-maven-plugin` with `ShortenFullyQualifiedTypeReferences` recipe to automatically replace unnecessary fully qualified class names with simple names + imports
- Plugin is gated behind a `-Prewrite` Maven profile — only runs in CI, not during local builds
- Uses `runNoFork` goal with `runPerSubmodule=true` to process each module immediately during the build
- Generated code (`src/generated/`) is excluded from processing
- Both PR and main build workflows now pass `-Prewrite` to `regen.sh`, and the existing "fail if uncommitted changes" step catches any unfixed FQCNs
- Update `CLAUDE.md` with import style conventions for AI agents
- Fix all 598 existing unnecessary FQCNs across 187 modules

## How it works

1. CI runs `regen.sh` with `-Prewrite` → OpenRewrite shortens unnecessary FQCNs and adds imports
2. `impsort` sorts the new imports (already runs in the same build)
3. `formatter` formats the code (already runs in the same build)
4. Existing `git status --porcelain` step fails if any files were modified

Local developers can opt-in with `mvn install -Prewrite` to auto-fix FQCNs before pushing.

## Test plan

- [x] Normal build (`mvn install`) does not run OpenRewrite
- [x] Build with `-Prewrite` runs OpenRewrite and shortens FQCNs per module
- [x] Generated files (`src/generated/`) are excluded
- [x] Modified files are detected by `git status --porcelain`
- [x] Full repo run fixes 598 files across 187 modules
- [ ] CI green


davsclaus (approved):

apupier (commented):
Why is it not applied during normal build like for other formatting?

oscerd (approved):
LGTM

gnodet:
> Why is it not applied during normal build like for other formatting?

It takes a bit of time.  The PR workflow is fine because it usually does not rebuild everything.  But we can apply it globally if we want.

components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramRenderer.java
apupier, line 37: missing import for jline Colors

apupier (changes requested):

davsclaus:
LGTM

PR #22942, opened 2026-05-05
fix(components): remove thread scopes

squakez:
Aligns the changed design as described in https://github.com/apache/camel/blob/main/proposals/tracing.adoc

Ref CAMEL-23380

# Description

# Target

# Tracking


oscerd (approved):

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/TraceProcessorsMicrometerObsInterceptStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+ * for any custom Micrometer Observability activity in the processor to be available.
+ */
+public class TraceProcessorsMicrometerObsInterceptStrategy implements InterceptStrategy {
+
+    // NOTE: this is an implementation detail that the interceptor should not know.
+    // We are temporarily using this to evaluate as a patch for a context leak problem we're suffering.
+    // Once we are clear this is correctly fixed and no more corner cases, then, we should change the TraceProcessorsInterceptStrategy
+    // class in camel-telemetry component in order to be able to retrieve the span before executing the processor and perform
+    // a similar logic of what we're doing here.
+    private SpanStorageManagerExchange spanStorage = new SpanStorageManagerExchange();
```
gnodet, line 44: The inline comment acknowledges this is temporary. It would be good to track this as a follow-up JIRA (or add a note in CAMEL-23380) so the intent to push this pattern into the base `TraceProcessorsInterceptStrategy` in `camel-telemetry` doesn't get lost.

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/TraceProcessorsMicrometerObsInterceptStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+        @Override
+        public boolean process(Exchange exchange, AsyncCallback callback) {
+            Span activeSpan = spanStorage.peek(exchange);
+            if (activeSpan != null) {
+                MicrometerObservabilitySpanAdapter otelSpan = (MicrometerObservabilitySpanAdapter) activeSpan;
+                try (Tracer.SpanInScope scope = tracer.withSpan(otelSpan.getSpan())) {
+                    return processor.process(exchange, doneSync -> {
+                        callback.done(doneSync);
+                    });
+                }
```
gnodet, line 88: In the async path, the `try-with-resources` closes the `SpanInScope` as soon as `processor.process()` returns, which means the scope is closed _before_ the async callback fires (when `doneSync=false`). This is likely intentional (scope is only needed for thread-local context on the calling thread), but it differs from the base `TraceProcessorsInterceptStrategy` which defers `endProcessorSpan()` into the callback's `finally` block.

Might be worth a brief comment confirming this is by design, since a future reader may view it as a bug.

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/MicrometerObservabilityTracer.java
```diff
@@ -106,6 +106,9 @@ protected void initTracer() {
                      "This can be useful for development only, avoid this in a production environment.");
         }
 
+        InterceptStrategy interceptStrategy = new TraceProcessorsMicrometerObsInterceptStrategy(tracer);
+        getCamelContext().getCamelContextExtension().addInterceptStrategy(interceptStrategy);
```
gnodet, line 110: This interceptor is registered unconditionally, while the base `TraceProcessorsInterceptStrategy` (registered in `Tracer.doInit()`) is also always added. That means every processor is now wrapped by **two** intercept strategies — the base one (which gates span creation internally on `isTraceProcessors()`) and this new one (which always does `spanStorage.peek()` + `tracer.withSpan()`).

For users with `traceProcessors=false` (like the JIRA reporter), this adds per-processor overhead that wasn't there before. Could this be gated behind a similar flag, or is the unconditional scope management intentional?

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/MicrometerObservabilitySpanAdapter.java
```diff
@@ -79,12 +71,8 @@ protected void close() {
...
 
     protected void deactivate() {
-        if (baggage != null) {
-            baggage.close();
-        }
-        if (scope != null) {
-            scope.close();
-        }
+        // NOTE: we had changed the implementation not to depend
+        // any longer by thread scopes
```
gnodet: Nit: grammar — "depend ... by" should be "depend ... on".

```suggestion
        // NOTE: the implementation no longer depends on thread scopes
```

gnodet (commented):
Thanks for tackling this tricky scope leak issue, @squakez — the new approach (short-lived scopes per processor call instead of long-lived scopes on the span adapter) is fundamentally sounder and should fix the OOM.

I have a few inline comments plus some general items:

### General findings

**Orphaned `MicrometerObservabilityTrace` class**
`MicrometerObservabilityTracerTestSupport` was the only consumer of `MicrometerObservabilityTrace.java`. Since the support class is deleted in this PR, `MicrometerObservabilityTrace` has no remaining references and should be deleted as well to avoid dead code.

**Commit message convention**
Per project guidelines, the commit format should be `CAMEL-23380: <description>` rather than conventional-commit style (`fix(components): remove thread scopes`).

**PR description**
The PR body currently contains only the unchecked template. Given the complexity of this scope management redesign, a description explaining *what* changed, *why*, and the tradeoffs would help reviewers and future maintainers.

---

> _Note: This review covers project conventions and code-level findings. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud)._

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/TraceProcessorsMicrometerObsInterceptStrategy.java
squakez: It was already there: https://issues.apache.org/jira/browse/CAMEL-23413

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/TraceProcessorsMicrometerObsInterceptStrategy.java
squakez: yes, scope is only needed on that thread.

components/camel-micrometer-observability/src/main/java/org/apache/camel/micrometer/observability/MicrometerObservabilityTracer.java
squakez: Yes, this is temporary. As soon as this is confirmed to work and no other side effects, we'll kick this off as part of https://issues.apache.org/jira/browse/CAMEL-23413 work

PR #22944, opened 2026-05-05
Align Camel JBang to the latest 4.18.2 release

Croway:


oscerd (approved):

oscerd:
I think we should do this on 4.18.x and release a camel-jbang docker image with 4.18.2, in main we should point to 4.20.0 and release the image related to 4.20

Croway:
> I think we should do this on 4.18.x and release a camel-jbang docker image with 4.18.2, in main we should point to 4.20.0 and release the image related to 4.20

I can backport this PR to 4.18.x, but we always provided the latest LTS on main, so that users when installing camel via `jbang app install camel@apache/camel` get the latest LTS

PR #22945, opened 2026-05-05
CAMEL-23409: camel-sjms - Disable ObjectMessage by default

oscerd:
## Summary

Brings `camel-sjms` and `camel-sjms2` in line with [CAMEL-23373](https://issues.apache.org/jira/browse/CAMEL-23373), which disabled JMS `ObjectMessage` support in `camel-jms` by default in 4.21.

A new `objectMessageEnabled` option (default `false`) has been added on `SjmsComponent` and `SjmsEndpoint`. `camel-sjms2` inherits the change via `Sjms2Component extends SjmsComponent` and `Sjms2Endpoint extends SjmsEndpoint`.

## Behavior change (when `objectMessageEnabled=false`, the new default)

`JmsBinding` now refuses to:

- create an `ObjectMessage` from a `Serializable` body (auto-detected `jmsMessageType=Object` or explicitly set)
- create an `ObjectMessage` for `transferException` replies
- extract the body of a received `jakarta.jms.ObjectMessage`

The error message points users at the new option:

> JMS ObjectMessage is disabled by default for security reasons (...). Set `objectMessageEnabled=true` on the SJMS endpoint or component to enable it.

To restore the previous behavior, enable the option at the component or endpoint level:

```properties
camel.component.sjms.objectMessageEnabled=true
camel.component.sjms2.objectMessageEnabled=true
```

```text
sjms:queue:foo?objectMessageEnabled=true
sjms2:queue:foo?objectMessageEnabled=true
```

## Files changed

- `SjmsEndpoint`, `SjmsComponent`: new `objectMessageEnabled` option, both annotations carry `security = "insecure:serialization"` so the Camel security policy sees a uniform signal at component and endpoint level.
- `JmsBinding`: gates ObjectMessage creation/extraction; new constructor parameter (older constructors preserved as compatibility wrappers, defaulting to `false`).
- `SjmsObjectMessageEnabledTest`: integration test covering producer/consumer default-disabled behavior, endpoint-level override, and component-level override (broker-based).
- `JmsBindingObjectMessageEnabledTest`: focused unit test for the `transferException` reply path guard, verified directly via `JmsBinding.makeJmsMessage(..., cause)` — broker round-trip would surface as a producer-side timeout and obscure the actual guard message.
- `camel-4x-upgrade-guide-4_21.adoc`: extended the existing CAMEL-23373 entry with a new `camel-sjms` / `camel-sjms2` section.
- Generated artifacts regenerated and committed: catalog (`sjms.json`, `sjms2.json`), configurer / URI factory classes, and endpoint DSL builder factories (`SjmsEndpointBuilderFactory`, `Sjms2EndpointBuilderFactory`).

## Review feedback addressed (gnodet)

- Added `security = "insecure:serialization"` to the component-level `@Metadata` so the catalog is consistent with the endpoint-level `@UriParam` and the camel-jms precedent.
- Added test coverage for the `transferException` reply path, the only previously-untested `objectMessageEnabled` guard.

## Test plan

- [x] `mvn test` in `components/camel-sjms` (118 tests, 0 failures — adds `JmsBindingObjectMessageEnabledTest`)
- [x] `mvn test` in `components/camel-sjms2` (23 tests, 0 failures)
- [x] Full root build: `mvn clean install -DskipTests -Dquickly`
- [x] Rebased onto current `main` (resolves prior conflict)
- [x] Coverage:
  - Producer refuses `Serializable` body when `objectMessageEnabled=false`
  - Producer accepts `Serializable` body when `objectMessageEnabled=true` (verified ObjectMessage is sent)
  - Consumer refuses externally-sent `ObjectMessage` when `objectMessageEnabled=false`
  - Component-level `objectMessageEnabled=true` propagates to endpoints
  - `transferException` reply path refused when `objectMessageEnabled=false` (unit test)

---


davsclaus (approved):

gnodet (commented):
Nice work bringing `camel-sjms` / `camel-sjms2` in line with the `camel-jms` CAMEL-23373 change. The approach is clean, the guard placement covers all three ObjectMessage paths, and the backward-compatible `JmsBinding` constructors are a good touch. Upgrade guide and generated artifacts look good.

Two findings below — one concrete fix and one optional test coverage note.

_This is a convention/rules review and does not replace specialized tools (CodeRabbit, SonarCloud, etc.)._

components/camel-sjms/src/main/java/org/apache/camel/component/sjms/SjmsComponent.java
```diff
@@ -83,6 +83,14 @@ public class SjmsComponent extends HeaderFilterStrategyComponent {
                             + " a conservative default filter denying java.net.** and otherwise allowing java.**, javax.**"
                             + " and org.apache.camel.** is applied.")
     private String deserializationFilter;
+    @Metadata(label = "advanced",
```
gnodet: The `@Metadata` annotation here is missing `security = "insecure:serialization"`. The `@Metadata` annotation does support the `security` attribute (defined in `Metadata.java:103`).

For comparison, the endpoint-level `@UriParam` in `SjmsEndpoint.java` correctly includes `security = "insecure:serialization"`, and in `camel-jms` both the component-level and endpoint-level catalog entries have `"security": "insecure:serialization"` (verified in `jms.json`).

Without this, the generated `sjms.json` / `sjms2.json` catalog files are inconsistent: the component-level property lacks the security tag while the endpoint-level parameter has it. This matters for the Camel security policy (`camel.security.*`) which uses these tags to warn or block insecure options at startup.

```suggestion
    @Metadata(label = "advanced", security = "insecure:serialization",
```

components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/SjmsObjectMessageEnabledTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Verifies that JMS ObjectMessage support is disabled by default in camel-sjms and can be enabled via the
+ * {@code objectMessageEnabled} option.
+ */
+public class SjmsObjectMessageEnabledTest extends JmsTestSupport {
```
gnodet, line 39: Nit / optional: the test class covers four scenarios nicely (producer disabled, producer enabled, consumer disabled, component-level propagation). One gap: the `transferException` reply path (`JmsBinding.createJmsMessage(Exception, Session)`) also has a new guard but isn't tested.

A test using an InOut exchange with `transferException=true` and `objectMessageEnabled=false` (the default) would verify that the guard blocks exception serialization. This is a distinct code path from the other three guards, and `transferException` is explicitly called out in the option description as requiring `objectMessageEnabled=true`.

Not a blocker — just a note for completeness.

components/camel-sjms/src/main/java/org/apache/camel/component/sjms/SjmsComponent.java
oscerd: Fixed in 1787b6cb143 — added `security = "insecure:serialization"` to the `@Metadata` on `SjmsComponent.objectMessageEnabled`. The regenerated `sjms.json` / `sjms2.json` and catalog entries now consistently carry the tag at both component (kind=property) and endpoint (kind=parameter) levels, so the Camel security policy sees a uniform signal. Thanks for catching this.

components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/SjmsObjectMessageEnabledTest.java
oscerd: Addressed in 1787b6cb143 — added `JmsBindingObjectMessageEnabledTest` covering the `transferException` reply path. I went unit-style (direct `JmsBinding.makeJmsMessage(...)` invocation with a non-null `cause`) rather than full integration: when the reply ObjectMessage can't be created on the consumer, the exception is caught inside `EndpointMessageListener` and the producer just sees a request timeout, which obscures the actual guard message. The unit test asserts the `IllegalStateException` carries both the `transferException reply` operation tag and the `objectMessageEnabled=true` guidance, which is what we actually want to verify. Happy to add an integration-style test as well if you'd prefer the broker round-trip coverage on top.

davsclaus (approved):

PR #22953, opened 2026-05-05
chore: migrate camel-jooq from commons-dbcp to commons-dbcp2

tmielke:
Replace unmaintained commons-dbcp 1.4 with modern commons-dbcp2 2.14.0 for test configuration. Changes include:

- Update Maven dependency to org.apache.commons:commons-dbcp2:2.14.0 with test scope
- Change package name from org.apache.commons.dbcp to org.apache.commons.dbcp2 in Spring configuration
- Rename maxActive property to maxTotal
- Update documentation example to reflect new package name
- Remove unused commons-dbcp-version property from parent POM

This aligns camel-jooq with other components (camel-jpa, camel-quarkus, camel-spring-batch) that already use commons-dbcp2 and removes the unmaintained commons-dbcp library.

Made with help from AI tools.

### DIFF
--- a/components/camel-jooq/pom.xml
+++ b/components/camel-jooq/pom.xml
@@ -54,9 +54,10 @@
             <version>${jakarta-annotation-api-version}</version>
         </dependency>
         <dependency>
-            <groupId>commons-dbcp</groupId>
-            <artifactId>commons-dbcp</artifactId>
-            <version>${commons-dbcp-version}</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-dbcp2</artifactId>
+            <version>${commons-dbcp2-version}</version>
+            <scope>test</scope>
         </dependency>
 
         <!-- test dependencies -->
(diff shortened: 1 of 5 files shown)


oscerd (approved):

Croway (approved):

components/camel-jooq/pom.xml
```diff
@@ -54,9 +54,10 @@
             <version>${jakarta-annotation-api-version}</version>
         </dependency>
         <dependency>
-            <groupId>commons-dbcp</groupId>
-            <artifactId>commons-dbcp</artifactId>
-            <version>${commons-dbcp-version}</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-dbcp2</artifactId>
+            <version>${commons-dbcp2-version}</version>
+            <scope>test</scope>
```
apupier, line 60: Is it expected that we add the scope test?

If yes, I think that in the documentation we must add that user needs to add the Maven dependency to have it working.
If not, need to remove the scope test

apupier (commented):
note: there is a camel-example that will need to be updated too when the PR is accepted and merged

components/camel-jooq/pom.xml
tmielke: Yes, the artifact is only used in test scope. We could add a note to the docs that the example requires commons-dbcp2 on the classpath but the we also do no instruct to add the actual DB driver jar.

davsclaus (approved):

apupier (approved):

PR #22954, opened 2026-05-05
Revert "chore(camel-test-infra-infinispan): upgrade infinispan.container to 16.2"

apupier:
Reverts apache/camel#22316

Infinispan upgrade to 16.2 is causing the tests to be very very flaky. Like 4 out 5 times failing.

see https://issues.apache.org/jira/browse/CAMEL-23410

I reproduced locally the failure several times with 16.2, and it always passed with 16.1

note that it is currently causing the CI to go on timeout on all builds, there si always at least one combination on which there is the problem and for which it goes on timEout.


oscerd (approved):

apupier:
I created https://issues.apache.org/jira/browse/CAMEL-23421 to take care of the upgrade later on

PR #22962, opened 2026-05-05
CAMEL-23424: Add --node-label option to route-diagram for description display

gnodet:
[CAMEL-23424](https://issues.apache.org/jira/browse/CAMEL-23424)

## Summary

- Add `--node-label` option to `camel cmd route-diagram` with three modes:
  - `code` (default): show URI/code in diagram nodes (current behavior)
  - `description`: prefer description over code if available, fall back to code
  - `both`: show description and code together (separate lines in PNG, pipe-separated in text)
- Parse `description` field from route structure JSON into `NodeInfo`
- Fix `DefaultDumpRoutesStrategy.dumpAsJSon()` to include `description` in structure JSON (was already present in `RouteStructureDevConsole`)
- Expose `nodeLabel` parameter in MCP tool
- 10 new tests covering all three modes (layout, text diagram, PNG rendering, parsing)

## Examples

| `--node-label code` | `--node-label description` | `--node-label both` (default) |
|---|---|---|
| diagram-code | diagram-description | diagram-both |

## Test plan

- [x] 47 diagram tests pass (8 new for node-label modes)
- [x] 4 jbang-core action tests pass (2 new for description parsing and mode parsing)
- [x] 3 MCP tools tests pass (updated for new parameter)
- [x] Visual verification with all three modes generating correct PNG output


gnodet:

@davsclaus I changed the default for `--node-label` to `both` so that descriptions are shown alongside the code by default. This way users who set descriptions on their EIP nodes get the benefit immediately without needing to discover the option, while the URI/code is still visible for technical context.

If you think `code` (current behavior) or `description` (description-only when available) would be a better default, happy to change it. What do you think?

davsclaus:
> 
> @davsclaus I changed the default for `--node-label` to `both` so that descriptions are shown alongside the code by default. This way users who set descriptions on their EIP nodes get the benefit immediately without needing to discover the option, while the URI/code is still visible for technical context.
> 
> If you think `code` (current behavior) or `description` (description-only when available) would be a better default, happy to change it. What do you think?

Yes both of them is good as default, then the diagram has it from the start, so the user can see its possible, and they can change it if they only want one of them.

components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramLayoutEngine.java
```diff
@@ -226,6 +243,30 @@ int computeNodeHeight(String label) {
         return baseNodeHeight + (lineCount - 1) * lineSpacing;
     }
 
+    String resolveLabel(NodeInfo info) {
+        String code = cleanLabel(info.code);
+        switch (nodeLabelMode) {
+            case DESCRIPTION:
+                return info.description != null && !info.description.isBlank() ? info.description : code;
+            case BOTH:
+                return code;
```
davsclaus: This seems wrong that BOTH returns code only. Should it not be description + code, like its done in the RouteDiagramRenderer class in this PR

davsclaus (changes requested):

gnodet:

@davsclaus Good catch — fixed. `resolveLabel()` now returns `description + code` for BOTH mode when a description is present, consistent with `resolveWrappedLines()` and the text renderer.

davsclaus (approved):

gnodet:

@davsclaus Fixed. `resolveLabel()` now returns `List<String>` — no encoding tricks. For BOTH mode it returns `[description, code]` as two separate elements. The layout engine flatMaps each through `wrapLabel()` for the PNG (separate lines), and the text renderer joins with ` | `.

The previous `label` field on `LayoutNode` was dead code (PNG rendering always used `wrappedLines`, never `label`), so it was removed.

PR #22971, opened 2026-05-05
chore(camel-diagram): Code cleanup and Unicode-safe text wrapping

gnodet:
**Depends on:** #22962

## Summary

Code cleanup and quality improvements for the `camel-diagram` module:

- Fix `Graphics2D` resource leak in constructor (try-finally)
- Replace `int[]` bounding box pattern with typed `Bounds` class
- Reduce visibility of internal-only APIs (`H_GAP`, `LABEL_OFFSET`, `findLastLayoutNode`, `findMaxY`, etc.)
- Remove redundant instance `resolveLabel` wrapper
- Extract `getTopY()` helper to deduplicate scope-check ternary in renderer
- Share `NODE_TEXT_PADDING` via getter instead of duplicating magic constant
- Named constants for magic numbers (`DEFAULT_FONT_SIZE`, `DEFAULT_BOX_WIDTH`, `DEFAULT_NODE_HEIGHT`)
- Fix `wrapLabel` to use `BreakIterator` for grapheme-safe Unicode handling (emojis, combining marks, ZWJ sequences like 👨‍👩‍👧‍👦)
- Add null-guard in JSON route parsing

## Test plan

- [x] 50 diagram tests pass (3 new for Unicode wrapping: emoji, combining marks, ZWJ sequences)
- [x] 4 jbang-core action tests pass
- [x] External consumer (`CamelRouteDiagramAction`) compiles and tests pass

### DIFF
--- a/components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramHelper.java
+++ b/components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramHelper.java
@@ -48,7 +48,11 @@ public final class RouteDiagramHelper {
         }
 
         for (int i = 0; i < arr.size(); i++) {
-            JsonObject o = (JsonObject) arr.get(i);
+            Object item = arr.get(i);
+            if (!(item instanceof JsonObject)) {
+                continue;
+            }
+            JsonObject o = (JsonObject) item;
             RouteInfo route = new RouteInfo();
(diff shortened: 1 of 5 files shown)


davsclaus:
I think the omc file should not be included in this PR.

Maybe we should add .omc to git ignore ?

davsclaus (approved):

PR #22979, opened 2026-05-05
CAMEL-23430: Fix flaky tests in camel-kafka

gnodet:
[CAMEL-23430](https://issues.apache.org/jira/browse/CAMEL-23430)

Fix two flaky tests in camel-kafka:

**KafkaConsumerAutoInstResumeRouteStrategyIT.testOffsetIsBeingChecked:**
- Producer was never flushed or closed after sending messages — messages may not have been delivered to Kafka when the test starts consuming
- `mock.assertIsSatisfied()` was called without any wait timeout — the assertion checked immediately before the route had time to consume and process messages
- Fix: use try-with-resources for the producer with an explicit flush, and use Awaitility to wait for the mock assertion

**KafkaIdempotentRepositoryEagerIT.createRepositoryTopic** (via KafkaTestUtil.createTopic):
- The `describeTopics()` future had only a 5-second timeout, which is too tight when the Kafka broker is still warming up
- The AdminClient was never closed, leaking resources
- Fix: increase timeout to 30 seconds and use try-with-resources for the AdminClient

### DIFF
--- a/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
+++ b/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
@@ -70,7 +70,7 @@ public class KafkaConsumerAutoInstResumeRouteStrategyIT extends BaseKafkaTestSup
         MockEndpoint mock = contextExtension.getMockEndpoint(KafkaTestUtil.MOCK_RESULT);
 
         mock.expectedMessageCount(10);
-        Awaitility.await().atMost(30, TimeUnit.SECONDS).untilAsserted(() -> mock.assertIsSatisfied());
+        Awaitility.await().atMost(25, TimeUnit.SECONDS).untilAsserted(() -> mock.assertIsSatisfied());
     }
 
     @AfterEach
--- a/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
+++ b/components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
@@ -19,6 +19,7 @@ package org.apache.camel.component.kafka.integration;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
@@ -29,6 +30,7 @@ import org.apache.camel.processor.resume.TransientResumeStrategy;
 import org.apache.camel.processor.resume.kafka.KafkaResumeStrategyConfigurationBuilder;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.awaitility.Awaitility;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


oscerd (approved):

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/common/KafkaTestUtil.java
```diff
@@ -91,14 +91,13 @@ public static void configureKafkaComponent(CamelContext context, String bootstra
...
-
-        try {
-            TopicDescription td = tdFuture.get(5L, TimeUnit.SECONDS);
+        try (AdminClient kafkaAdminClient = createAdminClient(service)) {
+            NewTopic testTopic = new NewTopic(topic, numPartitions, CreateTopicsRequest.NO_REPLICATION_FACTOR);
+            kafkaAdminClient.createTopics(Collections.singleton(testTopic));
+            KafkaFuture<TopicDescription> tdFuture
+                    = kafkaAdminClient.describeTopics(Collections.singletonList(topic)).topicNameValues().get(topic);
+
+            TopicDescription td = tdFuture.get(30L, TimeUnit.SECONDS);
```
apupier: I do not think that chaging the tiemout will change anything. Based on the stack trace, I understand that the answer is returned as soon as it is possible and i tis throwing:
```
.UnknownTopicOrPartitionException: This server does not host this topic-partition.
```

In case, it is a matter of time that the topic are not created, we need to use an await() and check several times if there are some topics. (which we do not know if it is a matter of time or if there is a differnt problem that can prevent the creation of the topics from time to time)

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
```diff
@@ -54,20 +56,21 @@ public static KafkaResumeStrategyConfigurationBuilder getDefaultKafkaResumeStrat
...
 
     @Test
     @Timeout(value = 30)
-    public void testOffsetIsBeingChecked() throws InterruptedException {
+    public void testOffsetIsBeingChecked() {
         MockEndpoint mock = contextExtension.getMockEndpoint(KafkaTestUtil.MOCK_RESULT);
 
         mock.expectedMessageCount(10);
-        mock.assertIsSatisfied();
+        Awaitility.await().atMost(30, TimeUnit.SECONDS).untilAsserted(() -> mock.assertIsSatisfied());
```
apupier: I think we need to be slightly lower than the timeout of the test to benefit from this assertion to fail with precise message instead of the genric timeout for the test.

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/common/KafkaTestUtil.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good point, you're right that the exception is `UnknownTopicOrPartitionException` which returns immediately rather than timing out. I've replaced the simple timeout increase with Awaitility polling: it now retries `describeTopics` every second for up to 30 seconds, using `ignoreExceptions()` to treat the `UnknownTopicOrPartitionException` as a "not yet" signal.

components/camel-kafka/src/test/java/org/apache/camel/component/kafka/integration/KafkaConsumerAutoInstResumeRouteStrategyIT.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch — lowered the Awaitility timeout to 25 seconds (vs the 30s `@Timeout`) so the assertion failure message is precise rather than a generic JUnit timeout.

apupier (approved):

PR #22980, opened 2026-05-06
CAMEL-23428: Fix flaky RabbitMQProducerInvalidExchangeIT test

gnodet:
[CAMEL-23428](https://issues.apache.org/jira/browse/CAMEL-23428)

The `RabbitMQProducerInvalidExchangeIT.testProducer` test is flaky because it asserts the direct cause of `CamelExecutionException` is `ShutdownSignalException`. However, Spring AMQP may wrap the underlying `ShutdownSignalException` in an `AmqpException` depending on timing, causing intermittent failures.

**Fix:** Use AssertJ's `hasRootCauseInstanceOf` to check the exception cause chain. This handles both cases and provides better diagnostics (full stack trace) on failure:
- Direct cause: `CamelExecutionException` → `ShutdownSignalException`
- Wrapped cause: `CamelExecutionException` → `AmqpException` → `ShutdownSignalException`

### DIFF
--- a/components/camel-spring-parent/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerInvalidExchangeIT.java
+++ b/components/camel-spring-parent/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerInvalidExchangeIT.java
@@ -50,7 +50,13 @@ public class RabbitMQProducerInvalidExchangeIT extends RabbitMQITSupport {
 
         final CamelExecutionException exception = Assertions.assertThrows(CamelExecutionException.class,
                 () -> template.sendBody("direct:start", "Hello World"));
-        Assertions.assertInstanceOf(ShutdownSignalException.class, exception.getCause());
+        // Spring AMQP may wrap ShutdownSignalException in AmqpException depending on timing
+        Throwable cause = exception.getCause();
+        while (cause != null && !(cause instanceof ShutdownSignalException)) {
+            cause = cause.getCause();
+        }
+        Assertions.assertInstanceOf(ShutdownSignalException.class, cause,
+                "Expected ShutdownSignalException in cause chain, but got: " + exception.getCause());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

oscerd (approved):

oscerd (approved):

components/camel-spring-parent/camel-spring-rabbitmq/src/test/java/org/apache/camel/component/springrabbit/integration/RabbitMQProducerInvalidExchangeIT.java
```diff
@@ -50,7 +50,13 @@ public void testProducer() {
...
         final CamelExecutionException exception = Assertions.assertThrows(CamelExecutionException.class,
                 () -> template.sendBody("direct:start", "Hello World"));
-        Assertions.assertInstanceOf(ShutdownSignalException.class, exception.getCause());
+        // Spring AMQP may wrap ShutdownSignalException in AmqpException depending on timing
+        Throwable cause = exception.getCause();
+        while (cause != null && !(cause instanceof ShutdownSignalException)) {
+            cause = cause.getCause();
+        }
+        Assertions.assertInstanceOf(ShutdownSignalException.class, cause,
+                "Expected ShutdownSignalException in cause chain, but got: " + exception.getCause());
```
apupier: I think we could use https://javadoc.io/doc/org.assertj/assertj-core/3.27.7/org/assertj/core/api/AbstractThrowableAssert.html#hasCauseInstanceOf(java.lang.Class)

it will allow to have shorter code and also a better error message with the full stack trace.
Especially given that it sounds like a stab in the dark that the AMQPEXception is wrapping the shutdownexception. At least, if i tis not the case, i twill help to investigate more.

gnodet: Good point, thanks! Switched to `assertThat(exception).hasRootCauseInstanceOf(ShutdownSignalException.class)` — uses `hasRootCauseInstanceOf` rather than `hasCauseInstanceOf` since the `ShutdownSignalException` is the deepest cause in both the direct and wrapped scenarios. This also gives the full stack trace on failure as you suggested.

PR #22981, opened 2026-05-06
CAMEL-22525: Fix flaky AS2 tests by eliminating port allocation race condition

gnodet:
[CAMEL-22525](https://issues.apache.org/jira/browse/CAMEL-22525)

## Summary

- Add `getLocalPort()` to `AS2ServerConnection` and `AS2AsyncMDNServerConnection` to expose the actual bound port
- Eliminate the TOCTOU race condition in AS2 tests by binding to port 0 (OS assigns a free port atomically) instead of pre-allocating a port with `AvailablePortFinder` which probes then closes a socket, leaving a window for another process to steal the port
- Rename non-final static fields to camelCase (`TARGET_PORT` → `targetPort`, `RECIPIENT_DELIVERY_ADDRESS` → `recipientDeliveryAddress`)

## Root Cause

`AvailablePortFinder.find()` opens a `ServerSocket` on port 0, records the assigned port, then **closes** the socket. When `AS2ServerConnection` later tries to bind to that port, another process may have already claimed it — a classic TOCTOU (time-of-check-to-time-of-use) race condition. This was particularly flaky on s390x CI where multiple processes compete for ports.

The prior fix (SO_REUSEADDR in #22095) helped with TIME_WAIT state but didn't eliminate the fundamental race.

## Fix

Instead of pre-allocating a port number, tests now:
1. Create `AS2ServerConnection` with port **0** — the OS assigns a free port atomically during `bind()`
2. Read the actual port back via the new `getLocalPort()` method
3. Use that port to configure the client

This completely eliminates the race condition since port assignment and binding happen in a single atomic operation.

## Test plan

- [x] All 94 camel-as2-api tests pass
- [x] All 12 camel-as2-component tests pass (1 skipped: manual Mendelson test)

### DIFF
--- a/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2ServerConnection.java
+++ b/components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2ServerConnection.java
@@ -639,6 +639,10 @@ public class AS2ServerConnection {
         consumerConfigurations.put(path, config);
     }
 
+    public int getLocalPort() {
+        return serversocket != null ? serversocket.getLocalPort() : -1;
+    }
+
     public void close() {
         if (acceptorThread != null) {
             lock.lock();
--- a/components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
+++ b/components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
@@ -84,13 +84,13 @@ public class AS2MessageTestBase {
 
     protected static final String METHOD = "POST";
     protected static final String TARGET_HOST = "localhost";
-    protected static int TARGET_PORT;
+    protected static int targetPort;
     protected static final Duration HTTP_SOCKET_TIMEOUT = Duration.ofSeconds(5);
     protected static final Duration HTTP_CONNECTION_TIMEOUT = Duration.ofSeconds(5);
     protected static final Integer HTTP_CONNECTION_POOL_SIZE = 5;
     protected static final Duration HTTP_CONNECTION_POOL_TTL = Duration.ofMinutes(15);
     protected static final Certificate[] VALIDATE_SIGNING_CERTIFICATE_CHAIN = null;
-    protected static String RECIPIENT_DELIVERY_ADDRESS;
+    protected static String recipientDeliveryAddress;
     protected static final String AS2_VERSION = "1.1";
     protected static final String USER_AGENT = "Camel AS2 Endpoint";
     protected static final String REQUEST_URI = "/";
@@ -171,7 +171,7 @@ public class AS2MessageTestBase {
         aSettings.setSenderData(AS2_NAME, FROM, "openas2a_alias");
 
         // Fixed receiver
-        aSettings.setReceiverData(AS2_NAME, "openas2b_alias", "http://" + TARGET_HOST + ":" + TARGET_PORT + "/");
+        aSettings.setReceiverData(AS2_NAME, "openas2b_alias", "http://" + TARGET_HOST + ":" + targetPort + "/");
         aSettings.setReceiverCertificate(issueCert);
 
         // AS2 stuff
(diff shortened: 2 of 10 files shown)


davsclaus (approved):

oscerd (approved):

components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
```diff
@@ -86,14 +84,13 @@ public class AS2MessageTestBase {
...
-    @RegisterExtension
-    protected static AvailablePortFinder.Port TARGET_PORT = AvailablePortFinder.find();
+    protected static int TARGET_PORT;
     protected static final Duration HTTP_SOCKET_TIMEOUT = Duration.ofSeconds(5);
     protected static final Duration HTTP_CONNECTION_TIMEOUT = Duration.ofSeconds(5);
     protected static final Integer HTTP_CONNECTION_POOL_SIZE = 5;
     protected static final Duration HTTP_CONNECTION_POOL_TTL = Duration.ofMinutes(15);
     protected static final Certificate[] VALIDATE_SIGNING_CERTIFICATE_CHAIN = null;
-    protected static final String RECIPIENT_DELIVERY_ADDRESS = "http://localhost:" + TARGET_PORT.getPort() + "/handle-receipts";
+    protected static String RECIPIENT_DELIVERY_ADDRESS;
```
apupier: given that it is no more a final attribute:

"Rename this field "RECIPIENT_DELIVERY_ADDRESS" to match the regular expression '^[a-z][a-zA-Z0-9]*$'."

components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
```diff
@@ -86,14 +84,13 @@ public class AS2MessageTestBase {
 
     protected static final String METHOD = "POST";
     protected static final String TARGET_HOST = "localhost";
-    @RegisterExtension
-    protected static AvailablePortFinder.Port TARGET_PORT = AvailablePortFinder.find();
+    protected static int TARGET_PORT;
```
apupier: given that it is no more a final attribute:

"Rename this field "RECIPIENT_DELIVERY_ADDRESS" to match the regular expression '^[a-z][a-zA-Z0-9]*$'."

components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2ServerConnection.java
```diff
@@ -639,6 +639,10 @@ public void registerConsumerConfiguration(String path, AS2ConsumerConfiguration
         consumerConfigurations.put(path, config);
     }
 
+    public int getLocalPort() {
+        return serversocket != null ? serversocket.getLocalPort() : -1;
```
apupier, line 643: shouldn't we return 0 to let the system pick an ephemeral port?

components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch — renamed `TARGET_PORT` to `targetPort` since it's no longer a constant. Fixed in 7413a4a.

components/camel-as2/camel-as2-api/src/test/java/org/apache/camel/component/as2/api/AS2MessageTestBase.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch — renamed `RECIPIENT_DELIVERY_ADDRESS` to `recipientDeliveryAddress` since it's no longer a constant. Fixed in 7413a4a.

components/camel-as2/camel-as2-api/src/main/java/org/apache/camel/component/as2/api/AS2ServerConnection.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Returning `-1` is intentional — it follows the `ServerSocket.getLocalPort()` convention where `-1` means "not bound / closed". Returning `0` would be misleading since `0` means "pick an ephemeral port" in socket APIs, which could cause silent bugs if the value is accidentally passed to another `new ServerSocket(0)` or `bind()` call. The `-1` sentinel makes the error explicit.

PR #22983, opened 2026-05-06
CAMEL-23216: Fix flaky mina sftp tests

gnodet:
[CAMEL-23216](https://issues.apache.org/jira/browse/CAMEL-23216)

## Summary

Several mina-sftp integration tests were flaky due to the embedded SFTP server not being fully ready to accept connections when tests started. The server restarts per test (`beforeEach`/`afterEach`), and without a readiness check, tests could attempt SSH connections before the server was listening.

## Changes

- **Add server readiness probe** in `SftpEmbeddedInfraService.setUpServer()` — after starting the embedded SSHD server, verify it accepts TCP connections before returning. Throws `IOException` if the server is not ready within 5 seconds, failing fast with a clear error message instead of proceeding to confusing timeout errors.

## Test plan

- [x] All 26 previously-flaky tests pass (MinaSftpAdvancedFileOperationsIT + MinaSftpAuthenticationIT)
- [x] Full mina-sftp integration test suite passes (249 tests, 0 failures)

### DIFF
--- a/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
+++ b/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
@@ -20,6 +20,7 @@ package org.apache.camel.test.infra.ftp.services.embedded;
 import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.Socket;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -136,6 +137,23 @@ public class SftpEmbeddedInfraService extends AbstractService implements FtpInfr
         sshd.start();
 
         port = ((InetSocketAddress) sshd.getBoundAddresses().iterator().next()).getPort();
+
+        waitForServerReady();
+    }
+
+    private void waitForServerReady() throws IOException {
+        long deadline = System.nanoTime() + 30_000_000_000L; // 30 seconds
+        IOException lastException = null;
+        while (System.nanoTime() < deadline) {
+            try (Socket socket = new Socket()) {
+                socket.connect(new InetSocketAddress("localhost", port), 1000);
+                return;
+            } catch (IOException e) {
+                lastException = e;
+                Thread.onSpinWait();
+            }
+        }
+        throw new IOException("SFTP server not ready after 30 seconds on port " + port, lastException);
     }
 
     protected PublickeyAuthenticator getPublickeyAuthenticator() {
--- a/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
+++ b/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
@@ -28,6 +28,7 @@ import java.security.KeyPair;
 import java.security.PublicKey;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
 import org.apache.camel.spi.annotations.InfraService;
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

oscerd (approved):

test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
```diff
@@ -136,6 +137,21 @@ public void setUpServer() throws Exception {
...
+        long deadline = System.nanoTime() + 5_000_000_000L; // 5 seconds
+        while (System.nanoTime() < deadline) {
+            try (Socket socket = new Socket()) {
+                socket.connect(new InetSocketAddress("localhost", port), 1000);
+                return;
+            } catch (IOException e) {
+                Thread.onSpinWait();
+            }
+        }
+        LOG.warn("SFTP server may not be ready after timeout on port {}", port);
```
apupier: Why is it a log?
Shouldn't we throw an exception? or use an assertion to prevent the test from going further?

apupier (commented):
the increase of timeout doesn't sound very useful. I looked to the time spent for several of the tests and the max amount of time on Jenkins Ci for one test is 12 seconds. With most of them around 2-3 seconds. When the tests are failing it was all of them. If it was due to the timeout, we should have only the ones with teime spent close to timeout impacted.
In case, the sftpservice is not fixing the issue, it will slow down the whole build.

test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
gnodet: Good point, fixed. The method now throws `IOException` if the server is not ready within 5 seconds, so tests fail fast with a clear error message instead of proceeding to confusing timeout errors.

gnodet (commented):
Thanks for the review @apupier. You're right — the timeout increase wasn't addressing the root cause. When the tests fail, they all fail simultaneously because the embedded server isn't ready, not because individual tests are too slow.

I've dropped all the `@Timeout` changes and kept only the server readiness probe in `SftpEmbeddedInfraService`. It now throws `IOException` (instead of logging a warning) if the server isn't ready within 5 seconds, so failures surface immediately with a clear message.

test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
davsclaus: -1

This is an infra service (not an test). This startup a FTP server in a docker container and if that takes longer than 5 seconds then it should be given more time to do so. Maybe if it takes longer than 30 seconds then something is wrong.

davsclaus (changes requested):
-1

This is an infra service (not an test). This startup a FTP server in a docker container and if that takes longer than 5 seconds then it should be given more time to do so. Maybe if it takes longer than 30 seconds then something is wrong.

test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
gnodet: Good point — increased the timeout to 30 seconds and it throws `IOException` on failure.

gnodet: Fixed — timeout increased to 30 seconds using `TimeUnit.SECONDS.toNanos(30)` for readability, and it throws `IOException` on failure.

gnodet (commented):
@davsclaus Thanks for the feedback. Fixed in the latest commit — timeout increased to 30 seconds using `TimeUnit.SECONDS.toNanos(30)`, and it throws `IOException` on failure. Ready for re-review.

davsclaus (approved):

apupier (approved):

PR #22989, opened 2026-05-06
CAMEL-23437: Fix and re-enable flaky DisruptorReconfigureWithBlockingProducerTest

gnodet:
[CAMEL-23437](https://issues.apache.org/jira/browse/CAMEL-23437)

## Summary

- Fix and re-enable `DisruptorReconfigureWithBlockingProducerTest`, which has been `@Disabled` since 2019

## Root causes of flakiness

1. **Fragile wall-clock timing assertion** (`watch.taken() < 2000`): only 400ms margin over the theoretical 1600ms drain time, easily exceeded on slow CI machines
2. **Race between producer and reconfiguration**: the `ConsumerEventHandler` processes events asynchronously (calls `processor.process()` with a no-op callback and returns immediately), so the ring buffer drains in microseconds. The producer could send all 12 messages to the OLD disruptor before `addRoutes()` nullified the reference, causing `mock:b` to receive far fewer messages than expected
3. **Non-volatile `exception` field** across threads and tight 5-second timeouts

## Changes

- Remove `@Disabled` annotation to re-enable the test
- Use explicit `CountDownLatch` synchronization: producer sends a first batch of 8 messages, signals, then waits for reconfiguration to complete before sending the second batch of 12 messages. This ensures the second batch deterministically goes through both consumers
- Use `blockWhenFull=true` on the producer URI (matching the test's name)
- Remove the fragile `StopWatch` timing assertion
- Make `exception` field `volatile` for proper cross-thread visibility
- Increase timeouts to generous values (10s/30s) and add explicit timeout to `MockEndpoint.assertIsSatisfied`
- Add proper error handling with `finally` for `resultLatch`

Verified stable with 30 consecutive passes locally and full disruptor test suite (102 tests).

### DIFF
--- a/components/camel-disruptor/src/test/java/org/apache/camel/component/disruptor/DisruptorReconfigureWithBlockingProducerTest.java
+++ b/components/camel-disruptor/src/test/java/org/apache/camel/component/disruptor/DisruptorReconfigureWithBlockingProducerTest.java
@@ -22,26 +22,28 @@ import java.util.concurrent.TimeUnit;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.test.junit6.CamelTestSupport;
-import org.apache.camel.util.StopWatch;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@Disabled("CAMEL-13629: Flaky test")
+/**
+ * Tests that reconfiguring a Disruptor (adding a consumer) works correctly while a producer is actively sending
+ * messages, and that the new consumer receives messages sent after reconfiguration.
+ */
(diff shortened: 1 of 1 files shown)


apupier (approved):
How many times have you launched it locally?
Were you able to reproduce the flakiness locally without the change?

gnodet (commented):
Thanks for the review @apupier!

> How many times have you launched it locally?

30 consecutive runs (10 + 20), all passed. The test completes consistently in under 1 second.

> Were you able to reproduce the flakiness locally without the change?

Yes. With just removing `@Disabled` (keeping the original test logic), it failed 3 out of 5 runs — `mock:b` would receive only 6 messages instead of the expected minimum 10, timing out at 30 seconds. The root cause is that the `ConsumerEventHandler` processes events asynchronously (calls `processor.process()` with a no-op callback and returns immediately), so the ring buffer drains in microseconds. The producer could send all 12 messages to the old disruptor before `addRoutes()` completed the reconfiguration.

davsclaus (approved):

PR #22991, opened 2026-05-06
camel-diagram: Fix route source name stripping line numbers

gnodet:
Fix route source display in diagram that was stripping line numbers from source names like `cheese.java:9`, showing just `9` instead.

## Problem

`extractSourceName()` used `lastIndexOf(':')` to strip URI scheme prefixes (e.g. `file:`), but this also stripped line number suffixes from sources like `cheese.java:9`, leaving just `9`.

## Fix

Changed to use `indexOf(':')` (first colon) with a check that the part before the colon consists only of letters (a valid URI scheme like `file`, `classpath`, `ref`). This correctly handles:
- `file:/path/to/my-route.yaml` → `my-route.yaml`
- `classpath:my-route.yaml` → `my-route.yaml`
- `cheese.java:9` → `cheese.java:9` (preserved)
- `file:/path/to/cheese.java:9` → `cheese.java:9` (preserved)

## Test plan

- [x] Added unit tests for all source name extraction scenarios
- [x] All 58 tests pass in camel-diagram module

### DIFF
--- a/components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramHelper.java
+++ b/components/camel-diagram/src/main/java/org/apache/camel/diagram/RouteDiagramHelper.java
@@ -86,10 +86,15 @@ public final class RouteDiagramHelper {
         if (source == null || source.isBlank()) {
             return null;
         }
-        // strip scheme prefix (e.g. "file:")
-        int colon = source.lastIndexOf(':');
-        if (colon >= 0 && colon < source.length() - 1) {
-            source = source.substring(colon + 1);
+        // strip URI scheme prefix (e.g. "file:", "classpath:") — only if the part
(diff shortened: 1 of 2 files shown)


davsclaus:
this is no longer needed as the json source from route structure dump has removed the line number and include this in a seperate "line" attribute

gnodet:

True, but the code is still wrong regardless — it uses `lastIndexOf(':')` to strip the scheme prefix, which means it keeps only the part after the **last** colon rather than stripping the **first** part (the scheme). For example, `classpath:my-route.yaml:10` would produce `10` instead of `my-route.yaml:10`.

The fix changes to `indexOf(':')` with a check that the part before the colon is all letters (a valid URI scheme), so it correctly strips only the scheme prefix.

davsclaus (approved):

PR #23003, opened 2026-05-06
CAMEL-23334: Fix lazy plugin loading to skip unnecessary Maven resolution

Croway:
## Summary

- Fix `PluginHelper.addPlugins()` which was downloading ALL configured plugins before filtering by the target command, causing expensive Maven resolution on every CLI invocation — even for commands like `camel version` that don't use any plugin.
- Move the command-name filter into `getActivePlugins()` so it runs **before** `getPlugin()`/`downloadPlugin()`, skipping resolution entirely for plugins that don't match the current command.
- Add tests for the filtered and unfiltered `getActivePlugins` overloads.

**Before**: `camel version` with a third-party plugin (453 transitive deps) → ~3.7s wasted on Maven resolution
**After**: `camel version` → zero Maven resolution, instant

## Test plan

- [x] Unit tests pass (`mvn verify` in `camel-jbang-core` — 379 tests, 0 failures)
- [x] E2E: `camel version` with forage plugin installed produces no resolution messages
- [x] E2E: `camel forage` still resolves the forage plugin correctly
- [x] E2E: `camel plugin` lists all configured plugins

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -117,14 +117,11 @@ public final class PluginHelper {
             }
         }
 
-        // Fall back to JSON configuration for additional or missing plugins
-        Map<String, Plugin> plugins = getActivePlugins(main, repos);
+        // Fall back to JSON configuration for additional or missing plugins.
+        // Pass target so only matching plugins are downloaded — avoids expensive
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus (approved):

davsclaus:
and jbang itself may eventually also become faster, and because camel jbang has a lot of commands and options then this will be good for us as well: https://github.com/jbangdev/jbang/pull/2453

Croway:
> and jbang itself may eventually also become faster, and because camel jbang has a lot of commands and options then this will be good for us as well: [jbangdev/jbang#2453](https://github.com/jbangdev/jbang/pull/2453)

nice!

PR #23005, opened 2026-05-06
CAMEL-23436: camel-netty - Add type convert from InputStream to ByteBuf

davsclaus:
# Description

# Target

# Tracking

### DIFF
--- a/components/camel-netty/src/main/java/org/apache/camel/component/netty/NettyConverter.java
+++ b/components/camel-netty/src/main/java/org/apache/camel/component/netty/NettyConverter.java
@@ -122,6 +122,13 @@ public final class NettyConverter {
     }
 
     @Converter
+    public static ByteBuf toByteBuffer(InputStream is, Exchange exchange) {
+        // convert to byte array first
+        byte[] arr = exchange.getContext().getTypeConverter().convertTo(byte[].class, exchange, is);
+        return toByteBuffer(arr);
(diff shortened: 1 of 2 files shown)


oscerd (approved):

components/camel-netty/src/main/java/org/apache/camel/component/netty/NettyConverter.java
```diff
@@ -121,6 +121,13 @@ public static ByteBuf toByteBuffer(String s, Exchange exchange) {
         return toByteBuffer(bytes);
     }
 
+    @Converter
+    public static ByteBuf toByteBuffer(InputStream is, Exchange exchange) {
+        // convert to byte array first
+        byte[] arr = exchange.getContext().getTypeConverter().convertTo(byte[].class, exchange, is);
```
Croway, line 127: is this the only approach? is it possible to create a bytebuf from an InputStream without creating an intermediate byte[]?

davsclaus: netty ByteBuf.readInputStream will create a byte array internally anyway so you end up with same situation, and you have the trouble to need to provide the total length of the input stream - so there is no free lunch to read a big stream without it all end up as bytes inside the JVM memory

Croway (approved):

PR #23010, opened 2026-05-06
CAMEL-23418: camel-ocsf - Upgrade OCSF schema to 1.8.0

oscerd:
# Description

Upgrade the bundled OCSF (Open Cybersecurity Schema Framework) schemas in `camel-ocsf` from 1.7.0 to 1.8.0, regenerating the JSON schemas and the POJOs derived from them via `jsonschema2pojo`.

JIRA: [CAMEL-23418](https://issues.apache.org/jira/browse/CAMEL-23418)

# Changes

- **Schemas regenerated** from OCSF 1.8.0 — every file in `components/camel-ocsf/src/main/resources/schema/` now references `https://schema.ocsf.io/1.8.0/...` and reflects 1.8.0 attributes.
- **`Packet.json` and `Token.json` added** — OCSF 1.8.0 introduces a `packet` reusable object that is referenced from every network activity class (`network_activity.packet_list`, `dns_activity.packet_list`, etc.), and a `token` object referenced from `api.token`. Adding them to `OBJECTS_TO_GENERATE` and `PROPERTY_TO_OBJECT_MAP` produces proper top-level POJOs and `$ref` links instead of inline anonymous types.
- **`generate-ocsf-schemas.py`**:
  - Default `--version` bumped from `1.7.0` to `1.8.0`.
  - New `load_ocsf_schema()` falls back to downloading the schema export from `https://schema.ocsf.io/{version}/export/schema?profiles=&extensions=win,linux,macos` when the requested version is not bundled inside the `ocsf-json-schema` Python package. This is required today because PyPI `ocsf-json-schema 1.1.0` only bundles up to 1.7.0.
- **`OcsfConstants.SCHEMA_VERSION`** updated from `1.7.0` to `1.8.0`. All existing OCSF event class UIDs are unchanged between 1.7.0 and 1.8.0, so no other constants needed updating.
- **Docs** (`ocsf-dataformat.adoc`): version references and object count (68 → 70) refreshed; `Packet` and `Token` listed in the object summary; a NOTE was added about the new download fallback.
- **Tests** (`OcsfEventExampleTest`, `ocsf-detection-finding-example.json`): version strings in test fixtures and explanatory comments updated from `1.7.0` to `1.8.0`. Behavior is unchanged (attacks remain on `FindingInfo` in 1.8.0).

# Verification

- `mvn clean test` in `components/camel-ocsf` — all 8 tests pass.
- `mvn clean install -DskipTests -pl :camel-ocsf -am` from the repo root — BUILD SUCCESS.
- `git grep -n 1\\.7\\.0 components/camel-ocsf/` returns no matches (confirming nothing was missed).
- Class UIDs were diffed between 1.7.0 (bundled) and 1.8.0 (downloaded) — no UID changes.

# Note on the schema download fallback

The fallback uses the public `schema.ocsf.io` export endpoint and runs only at generator time (a developer step), not at component build time. The Maven build still consumes the committed JSON schemas, so it remains fully offline. As soon as `ocsf-json-schema` ships a release that bundles 1.8.0, the fallback will simply not fire.

---


davsclaus (approved):

apupier (commented):
Would it imply end-users to potentially have changes to do in their Camel routes?

oscerd:
@apupier no — this should be a transparent upgrade for end-users.

Looking at what actually lands in the generated `org.apache.camel.dataformat.ocsf.model.*` classes (the script generates with `profiles=[]`, so profile-bound attributes are filtered out in both versions):

- **No fields are removed** from any of the 34 event classes or 68 reusable objects already generated for 1.7.0.
- The change is purely additive: ~30 new fields across event classes (e.g. `app_protocol_name`, `network_observation_point`, `packet_list` on the network classes), ~15 new fields across reusable objects (e.g. `account.is_disabled`, `device.mac_vendor`, `api.token`), and 2 new top-level POJOs (`Packet`, `Token`).
- Every existing OCSF event class UID is stable across 1.7.0 and 1.8.0.

So routes that marshal/unmarshal OCSF events, populate fields via setters, or read existing getters keep working as-is. Existing 1.7.0 JSON payloads continue to deserialize cleanly. The only behavioural change is that fields newly typed in 1.8.0 (such as `packet_list` on the network activity classes) now land on a typed getter instead of `additionalProperties` — that's a non-breaking improvement.

The one tiny caveat is the `OcsfConstants.SCHEMA_VERSION` constant value moves from `"1.7.0"` to `"1.8.0"`. The constant itself still exists with the same signature; only the string value changes, so code that hardcodes `equals("1.7.0")` against it would need to update.

PR #23021, opened 2026-05-06
CAMEL-22640: Add NullAway validation profile and fix null annotations in camel-api

gnodet:
[CAMEL-22640](https://issues.apache.org/jira/browse/CAMEL-22640)

## Summary

Add comprehensive [JSpecify](https://jspecify.dev/) null safety annotations to `camel-api`, and a NullAway compile-time validation profile to enforce them.

### JSpecify null annotations on `camel-api` (283 files)
- Add `org.jspecify:jspecify` dependency to `camel-api`
- Mark all 16 packages in `camel-api` as `@NullMarked` via `package-info.java` files (`org.apache.camel`, `o.a.c.spi`, `o.a.c.vault`, `o.a.c.support.jsse`, `o.a.c.health`, `o.a.c.resume`, etc.)
- Add ~1200 `@Nullable` annotations across the public API surface (interfaces, abstract classes, exception constructors, return types, method parameters) to accurately document nullability contracts
- Key areas annotated:
  - `CamelContext` — nullable returns for `getComponent()`, `hasComponent()`, `getRoute()`, `getProcessor()`, `resolveDataFormat()`, `resolveTransformer()`, `resolveValidator()`, `getVariable()`, `getGlobalOption()`, etc.
  - `Exchange` / `Message` — nullable bodies, headers, properties
  - `AggregationStrategy` — nullable old/new exchange parameters and return
  - Exception classes — nullable causes and detail fields
  - JSSE/SSL classes — nullable context parameters, key/trust managers
  - `spi.*` — nullable returns on resolvers, factories, registries, repositories
  - `vault.*` — nullable optional configuration fields
  - `resume.*` — nullable serialization results
- Add null guards where `@Nullable` values are dereferenced (e.g., `HealthCheckHelper`, `RoutesLoader`, `GroovyScriptCompiler`, JSSE parameter classes)
- Update Velocity template for generated `ExchangeConstantProvider` to include `@Nullable`

### NullAway validation profile (`parent/pom.xml`)
- Add opt-in `nullcheck` Maven profile using [NullAway](https://github.com/uber/NullAway) via Error Prone
- Error Prone 2.49.0 + NullAway 0.13.4
- JSpecify mode enabled, only checks `@NullMarked` packages
- All other Error Prone checks disabled — only NullAway runs
- Requires JDK 21+ (forked compiler with `--add-exports` flags)
- **Opt-in only**: `mvn install -Pnullcheck -DskipTests`

### Test fix
- Fix import formatting in `AggregateProcessorTest` to match updated `AggregationStrategy` signature

## Verification

```bash
# Run NullAway on camel-api (requires JDK 21+)
JAVA_HOME=$HOME/.sdkman/candidates/java/21 mvn clean install -B -pl core/camel-api -Pnullcheck -DskipTests -am
# Result: BUILD SUCCESS, 0 NullAway warnings

# Also verified clean across full camel-core module tree
JAVA_HOME=$HOME/.sdkman/candidates/java/21 mvn clean install -B -pl core/camel-core -Pnullcheck -DskipTests -am
# Result: BUILD SUCCESS, 0 NullAway warnings
```


davsclaus:
I assume you can close your old PR

davsclaus:
There are uncommitted changes
HEAD detached at pull/23021/merge
	modified:   core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateProcessorTest.java

davsclaus (approved):

davsclaus:
It may be good to add a note in the 4.21 upgrade guide that camel-api has dependency on jspecify JAR

apupier (commented):
last release of jspecify was 2 years ago.

Is the benefit really that huge to introduce a dependency on a project which doesn't seem very active?

gnodet:
@davsclaus Good point — I've added a note to the 4.21 upgrade guide and also made the jspecify dependency `<optional>true</optional>` so it's not pulled transitively into consumer projects.

gnodet:
@apupier JSpecify 1.0.0 is a stable release — the project only contains 4 annotation types (\`@Nullable\`, \`@NonNull\`, \`@NullMarked\`, \`@NullUnmarked\`), so there's very little reason for frequent releases. It's backed by Google, JetBrains, and others, and is becoming the de facto standard for null annotations in the Java ecosystem (supported by NullAway, Error Prone, Kotlin, IntelliJ, etc.).

That said, I've made the dependency \`<optional>true</optional>\` so it's not pulled transitively into consumer projects. Annotations gracefully degrade if the JAR isn't on the classpath — they simply become invisible to reflection.

core/camel-api/src/main/java/org/apache/camel/spi/OAuthClientConfig.java
```diff
@@ -24,46 +28,46 @@
...
 
-    public String getClientSecret() {
+    public @Nullable String getClientSecret() {
         return clientSecret;
     }
 
     public OAuthClientConfig setClientSecret(String clientSecret) {
-        this.clientSecret = clientSecret;
+        this.clientSecret = Objects.requireNonNull(clientSecret, "clientSecret");
         return this;
```
apupier, line 54: the clientSecret is nullable, why do we requireNonNull?

core/camel-api/src/main/java/org/apache/camel/spi/Transformer.java
```diff
@@ -55,7 +58,7 @@ protected Transformer() {
     }
 
     protected Transformer(String name) {
-        this.name = name;
+        this.name = Objects.requireNonNull(name, "name");
```
apupier, line 61: the name is marked as Nullable but here there is a requireNonNull

apupier (changes requested):
most of (all?) requireNonNullMessage are not giving any information. Either remove them or provide a useful message.

To be honest, I feel like this is the kind of changes that shouldn't be done in a bulk change and reviewed carefully at each place. I checked only few files and found inconsistencies

core/camel-api/src/main/java/org/apache/camel/spi/Transformer.java
gnodet: For this constructor, yes.  But the no-arg constructor does not set the `name` field, which ends up being `null`, hence the `@Nullable`.

core/camel-api/src/main/java/org/apache/camel/spi/OAuthClientConfig.java
gnodet: For this constructor, yes. But the no-arg constructor does not set the `clientSecret` field, which ends up being `null`, hence the @Nullable.

gnodet:
> most of (all?) requireNonNullMessage are not giving any information. Either remove them or provide a useful message.
> 
> To be honest, I feel like this is the kind of changes that shouldn't be done in a bulk change and reviewed carefully at each place. I checked only few files and found inconsistencies

This PR does not modify current semantics.  So most fields are annotated with `@Nullable` because there's a no-arg constructor, even though it's not _supposed_ to be `null` at runtime.  This is a design flaw in the API, but cannot be fixed by annotations.  This PR also adds a profile which uses compile-time source analysis via Error Prone to make sure the whole code actually follows the annotations.

I could improve the `requireNonNull` messages, but beyond the name of the field, anything else is just noise imho, but I can do some "xxx must be non null` instead of just "xxx".   Though the JDK itself is using just the field name, and so does Guava.  Spring is different and uses `Assert.notNull(obj, "'obj' must not be null")`.

core/camel-api/src/main/java/org/apache/camel/resume/ResumeStrategyConfiguration.java
```diff
@@ -17,21 +17,24 @@
...
 import org.apache.camel.resume.cache.ResumeCache;
+import org.jspecify.annotations.Nullable;
 
 /**
  * Basic configuration holder for resume strategies
  */
 public abstract class ResumeStrategyConfiguration {
-    private Cacheable.FillPolicy cacheFillPolicy;
-    private ResumeCache<?> resumeCache;
+    private Cacheable.@Nullable FillPolicy cacheFillPolicy;
```
davsclaus, line 29: this looks a bit funny

core/camel-api/src/main/java/org/apache/camel/resume/ResumeStrategyConfiguration.java
```diff
@@ -17,21 +17,24 @@
...
+    private @Nullable ResumeCache<?> resumeCache;
 
     /**
      * Gets the {@link org.apache.camel.resume.Cacheable.FillPolicy} for the cache used in the strategy
      *
      * @return the fill policy to use
      */
-    public Cacheable.FillPolicy getCacheFillPolicy() {
+    public Cacheable.@Nullable FillPolicy getCacheFillPolicy() {
```
davsclaus, line 37: here as well

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -119,34 +124,47 @@ default boolean allowNull() {
...
+            throws TypeConversionException {
+        if (value == null) {
+            return null;
+        }
         return convertTo(value.getClass(), type, exchange, value);
     }
 
     @Override
-    default <T> T mandatoryConvertTo(Class<T> type, Object value)
+    default <T> T mandatoryConvertTo(Class<T> type, @Nullable Object value)
```
davsclaus, line 144: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -119,34 +124,47 @@ default boolean allowNull() {
...
+    default <T> @Nullable T convertTo(Class<T> type, @Nullable Object value) throws TypeConversionException {
+        if (value == null) {
+            return null;
+        }
         return convertTo(value.getClass(), type, null, value);
     }
 
     @Override
-    default <T> T convertTo(Class<T> type, Exchange exchange, Object value) throws TypeConversionException {
+    default <T> @Nullable T convertTo(Class<T> type, @Nullable Exchange exchange, @Nullable Object value)
```
davsclaus, line 135: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -119,34 +124,47 @@ default boolean allowNull() {
     }
 
     @Override
-    default <T> T convertTo(Class<T> type, Object value) throws TypeConversionException {
+    default <T> @Nullable T convertTo(Class<T> type, @Nullable Object value) throws TypeConversionException {
```
davsclaus, line 127: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -91,7 +95,8 @@ default <T> T tryConvertTo(Class<?> from, Class<T> to, Exchange exchange, Object
      * @throws TypeConversionException            is thrown if error during type conversion
      * @throws NoTypeConversionAvailableException if no type converters exists to convert to the given type
      */
-    default <T> T mandatoryConvertTo(Class<?> from, Class<T> to, Exchange exchange, Object value)
+    @SuppressWarnings("NullAway")
+    default <T> T mandatoryConvertTo(Class<?> from, Class<T> to, @Nullable Exchange exchange, @Nullable Object value)
             throws TypeConversionException, NoTypeConversionAvailableException {
         Object t = convertTo(from, to, exchange, value);
```
davsclaus, line 101: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -119,34 +124,47 @@ default boolean allowNull() {
...
-    default <T> T mandatoryConvertTo(Class<T> type, Exchange exchange, Object value)
+    default <T> T mandatoryConvertTo(Class<T> type, @Nullable Exchange exchange, @Nullable Object value)
             throws TypeConversionException, NoTypeConversionAvailableException {
-        return mandatoryConvertTo(value.getClass(), type, exchange, value);
+        return mandatoryConvertTo(value != null ? value.getClass() : type, type, exchange, value);
     }
 
     @Override
-    default <T> T tryConvertTo(Class<T> type, Object value) {
+    default <T> @Nullable T tryConvertTo(Class<T> type, @Nullable Object value) {
```
davsclaus, line 156: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -119,34 +124,47 @@ default boolean allowNull() {
...
+        if (value == null) {
+            return null;
+        }
         return tryConvertTo(value.getClass(), type, null, value);
     }
 
     @Override
-    default <T> T tryConvertTo(Class<T> type, Exchange exchange, Object value) {
+    default <T> @Nullable T tryConvertTo(Class<T> type, @Nullable Exchange exchange, @Nullable Object value) {
```
davsclaus, line 164: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -65,7 +68,8 @@ public interface BulkTypeConverters extends Ordered, TypeConverter {
      * @param  value the value to be converted
      * @return       the converted value, or <tt>null</tt> if not possible to convert
      */
-    default <T> T tryConvertTo(Class<?> from, Class<T> to, Exchange exchange, Object value) throws TypeConversionException {
+    default <T> @Nullable T tryConvertTo(Class<?> from, Class<T> to, @Nullable Exchange exchange, @Nullable Object value)
```
davsclaus, line 71: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/BulkTypeConverters.java
```diff
@@ -53,7 +55,8 @@ public interface BulkTypeConverters extends Ordered, TypeConverter {
      *                                 return null.
      * @throws TypeConversionException is thrown if error during type conversion
      */
-    <T> T convertTo(Class<?> from, Class<T> to, Exchange exchange, Object value) throws TypeConversionException;
+    <T> @Nullable T convertTo(Class<?> from, Class<T> to, @Nullable Exchange exchange, @Nullable Object value)
```
davsclaus, line 58: the value is not null - you cannot converter a null value so this should be changed

core/camel-api/src/main/java/org/apache/camel/spi/ComponentCustomizer.java
```diff
@@ -139,7 +140,7 @@ public boolean test(String s, Component target) {
      */
     class Builder<T extends Component> {
         private final Class<T> type;
-        private BiPredicate<String, Component> condition;
+        private @Nullable BiPredicate<String, Component> condition;
```
davsclaus, line 143: why is this nullable. The point of the customizer is to call the condition

core/camel-api/src/main/java/org/apache/camel/spi/DataFormatCustomizer.java
```diff
@@ -139,7 +140,7 @@ public boolean test(String s, DataFormat target) {
      */
     class Builder<T extends DataFormat> {
         private final Class<T> type;
-        private BiPredicate<String, DataFormat> condition;
+        private @Nullable BiPredicate<String, DataFormat> condition;
```
davsclaus, line 143: same here as last comment

core/camel-api/src/main/java/org/apache/camel/spi/LifecycleStrategy.java
```diff
@@ -199,8 +200,8 @@ default void onServiceRemove(CamelContext context, Service service, Route route,
      * @param threadPoolProfileId id of the thread pool profile, if used for creating this thread pool (can be null)
      */
     void onThreadPoolAdd(
-            CamelContext camelContext, ThreadPoolExecutor threadPool, String id,
-            String sourceId, String routeId, String threadPoolProfileId);
+            CamelContext camelContext, ThreadPoolExecutor threadPool, @Nullable String id,
```
davsclaus, line 203: we should likely enforce that some of these are not null as creating a thread pool we should have id / source id to identify it.

core/camel-api/src/main/java/org/apache/camel/spi/LifecycleStrategy.java
```diff
@@ -222,8 +223,8 @@ void onThreadPoolAdd(
      * @param threadPoolProfileId id of the thread pool profile, if used for creating this thread pool (can be null)
      */
     default void onThreadPoolAdd(
-            CamelContext camelContext, ExecutorService executorService, String id,
-            String sourceId, String routeId, String threadPoolProfileId) {
+            CamelContext camelContext, ExecutorService executorService, @Nullable String id,
```
davsclaus, line 226: we should likely enforce that some of these are not null as creating a thread pool we should have id / source id to identify it.

gnodet:
Thanks for the thorough review. The intent of this PR is to document **current reality** (what can actually be null today), not **intentionality** (what should ideally be non-null). A few responses:

- **BulkTypeConverters `value`**: `CoreTypeConverterRegistry.doConvertTo()` and `fastConvertTo()` both explicitly handle `null` values (returning `null`, or converting to primitive defaults). There's also `ConverterTest.testNullConverter()` confirming this. The `@Nullable` reflects reality.
- **ComponentCustomizer/DataFormatCustomizer `condition`**: The field is an optional builder parameter — `condition()` handles `null` by providing a default predicate. The `@Nullable` matches existing behavior.
- **LifecycleStrategy thread pool params**: The existing Javadoc already says "(can be null in special cases)" for `sourceId`, `routeId`, `threadPoolProfileId`. The `@Nullable` just formalizes what the docs state.
- **ResumeStrategyConfiguration**: The `Cacheable.@Nullable FillPolicy` syntax looks unusual but is the correct JSpecify placement for type-use annotations on nested types — the `@Nullable` must go directly before the simple class name, not before the outer type.

Where we agree some parameters *should* be non-null (like thread pool `id`), that's a separate behavioral change — tightening the contract and adding validation. Happy to do that in follow-up PRs where each case can be reviewed individually. This PR intentionally avoids changing semantics.

davsclaus:
> Where we agree some parameters _should_ be non-null (like thread pool `id`), that's a separate behavioral change — tightening the contract and adding validation. Happy to do that in follow-up PRs where each case can be reviewed individually. This PR intentionally avoids changing semantics.

Yes I agree this is better. Follow up PRs to improve the null vs non-null in specific areas afterwards where we can correct/improve the current API. And leave this PR as-is.

davsclaus:
LGTM

apupier:
It broke 3 tests:

https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/1846/testReport/org.apache.camel.component.jsonvalidator/JsonValidationExceptionTest/testErrorsEmpty/

```
java.lang.NullPointerException: message
	at java.base/java.util.Objects.requireNonNull(Objects.java:259)
	at org.apache.camel.ValidationException.<init>(ValidationException.java:35)
	at org.apache.camel.component.jsonvalidator.JsonValidationException.<init>(JsonValidationException.java:44)
	at org.apache.camel.component.jsonvalidator.JsonValidationExceptionTest.testErrorsEmpty(JsonValidationExceptionTest.java:45)
```

https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/1846/testReport/org.apache.camel.component.jsonvalidator/ValidatorRouteTest/testNullHeader/

```
org.opentest4j.AssertionFailedError: Exception should be correct type ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:158)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:139)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:69)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:41)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:228)
	at org.apache.camel.component.jsonvalidator.ValidatorRouteTest.testNullHeader(ValidatorRouteTest.java:113)
```

https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/1846/testReport/org.apache.camel.component.rest.openapi.validator.client/OpenApiRestClientRequestValidatorTest/testValidateBody/

```
java.lang.NullPointerException: validationLevels
	at java.base/java.util.Objects.requireNonNull(Objects.java:259)
	at org.apache.camel.spi.RestConfiguration.setValidationLevels(RestConfiguration.java:660)
	at org.apache.camel.component.rest.openapi.validator.client.OpenApiRestClientRequestValidatorTest.testValidateBody(OpenApiRestClientRequestValidatorTest.java:79)
```

apupier:
reported https://issues.apache.org/jira/browse/CAMEL-23542

apupier:
another regression due to this PR:

```
org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a Lambda expression in org.apache.camel.component.aws2.s3.integration.S3ConsumerIgnoreBodyIT null within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1160)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
	at org.apache.camel.component.aws2.s3.integration.S3ConsumerIgnoreBodyIT.sendIn(S3ConsumerIgnoreBodyIT.java:54)
Caused by: java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:204)
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:101)
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:81)
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:103)
	... 5 more
```

gnodet:
The S3ConsumerIgnoreBodyIT failure is not related to this PR (no S3 files were touched here). The root cause is likely #23086 (CAMEL-23132: ListObjectsV2 migration) which made `contentType` return null in some cases, causing `Optional.of(contentType)` to throw NPE in `includeS3Object()`. This silently kills the consumer poll and causes the Awaitility timeout.

This was already fixed in #23300 by changing `Optional.of` to `Optional.ofNullable`.

gnodet:
Regarding the ValidationException / CamelExchangeException / RestConfiguration test failures reported [above](https://github.com/apache/camel/pull/23021#issuecomment-4441975797): the revert in #23289 correctly removed the `Objects.requireNonNull()` runtime checks, but also removed the `@Nullable` JSpecify annotations. Follow-up PR #23304 restores the annotations without the runtime enforcement.

PR #23035, opened 2026-05-07
chore: fix FQCN in AggregateProcessorTest

gnodet:
## Summary

- Fix fully-qualified `java.util.concurrent.TimeUnit.SECONDS` usage in `AggregateProcessorTest` — add proper import and use simple class name instead.
- This was introduced in commit 9e5c829e7a25 (CAMEL-22539) and causes CI failure in the "Fail if there are uncommitted changes" step because the OpenRewrite FQCN-shortening plugin rewrites it during the build.

## Test plan

- [x] `mvn compile test-compile -pl core/camel-core` passes
- [ ] CI "uncommitted changes" check passes

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateProcessorTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/aggregator/AggregateProcessorTest.java
@@ -18,6 +18,7 @@ package org.apache.camel.processor.aggregator;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


gnodet:
The CAMEL-22539 PR (#23024) was created and CI-tested before the OpenRewrite FQCN enforcement (#22935) was merged. Its CI passed because the rewrite rule wasn't active yet. When it was subsequently merged into `main` — which now includes the OpenRewrite plugin — the build rewrites the inline `java.util.concurrent.TimeUnit.SECONDS` to use a proper import, producing an uncommitted diff that fails the "Fail if there are uncommitted changes" CI check.

davsclaus (approved):

gnodet:
Duplicated by #23036

PR #23039, opened 2026-05-07
Revert "Enforce import style: auto-shorten FQCNs via OpenRewrite in CI"

apupier:
Reverts apache/camel#22935

proposal to revert the change as it seems to quadruple the time of each PR check build for low value. (just teach the AI to do basic junior level work)


davsclaus:
we could maybe revert only that rule that reformat the code, or have it as a manual step only we can occationally run and manually commit. The code changes in the java source seems okay.

apupier:
> we could maybe revert only that rule that reformat the code, or have it as a manual step only we can occationally run and manually commit. The code changes in the java source seems okay.

you're right, I forgot that it was doen a single commit/PR

apupier:
created https://github.com/apache/camel/pull/23041

PR #23042, opened 2026-05-07
chore: upgrade okhttp to version 5.3.2

tmielke:
Upgrading dependency `com.squareup.okhttp3:okhttp` to latest `5.3.2` version (already defined in `parent/pom.xml`) 
as currently used version `3.14.9` is 6 years old!

This updates `camel-jetty` and `camel-olingo2` to use `com.squareup.okhttp3:okhttp-jvm:5.3.2`.
Component `camel-huaweicloud-obs` references okhttp3 as compile time dependency but it is not used there at all. So okhttp3 got removed as a dependency there. 

All unit tests of these components pass, with the exception of `JettySessionSupportTest` in `camel-jetty`. However this test fails for me on main branch as well and does not use okhttp classes in any way.

The only component left to use `com.squareup.okhttp3:okhttp:3.14.9` is `camel-salesforce`. A quick attempt to upgrade  to version `5.3.2` here as well failed and breaks many unit tests. So this component will require a separate investigation.


davsclaus (approved):

Croway:
I'll merge it, but tomorrow let's check the CI https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/

PR #23043, opened 2026-05-07
camel-jbang: Fix Jolokia command when running from camel-launcher nested JAR

smongiar:
When `camel-launcher` is packaged as a Spring Boot executable JAR using `camel-repackager-maven-plugin`, the Jolokia agent JAR is nested inside `BOOT-INF/lib/` and cannot be resolved as a real filesystem path.

`OptionsAndArgs.lookupJarFile()` uses `getProtectionDomain().getCodeSource().getLocation()` which returns the fat JAR instead of the nested Jolokia agent JAR. When `VirtualMachine.loadAgent()` tries to load the fat JAR as an agent, it fails with:

```
ERROR: Cannot execute jolokia command due: Error while loading Jolokia agent to a JVM process
```

This fix detects when the resolved JAR is not a valid Jolokia agent (by checking the manifest for `Agent-Class`) and extracts the actual agent JAR from inside the fat JAR to a temp file, allowing `loadAgent()` to use a real filesystem path.

The temp file is cleaned up in a `finally` block after use.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/Jolokia.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/Jolokia.java
@@ -16,7 +16,15 @@
  */
 package org.apache.camel.dsl.jbang.core.commands.process;
 
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
 
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

Croway:
@davsclaus https://github.com/apache/camel/pull/23043/changes#diff-bd6c1fc0829371210347b5502274eb69e57fef7b2a526b5bc5c964117c020ba0R240 will be part of camel-launcher, is it expected?

davsclaus:
that would be expected if this makes the jolokia goal work (and that we want to have such feature in the launcher) - it would be needed for hawtio goal as well

smongiar:
> @davsclaus https://github.com/apache/camel/pull/23043/changes#diff-bd6c1fc0829371210347b5502274eb69e57fef7b2a526b5bc5c964117c020ba0R240 will be part of camel-launcher, is it expected?

 'camel-jbang` is misleading in description. It's involved, but it's specifically a camel-launcher bug, we need this, for hawtio integration

davsclaus:
LGTM what do you say @Croway

PR #23090, opened 2026-05-08
CAMEL-23469: Fix infinite redelivery loop when child route removes headers

Croway:
## Summary

- Fix `RedeliveryErrorHandler.incrementRedeliveryCounter()` to use the internal `redeliveryCounter` field as the authoritative counter instead of reading from the `CamelRedeliveryCounter` exchange header, which user routes can remove (e.g., via `removeHeaders("*")`)
- Fix `prepareExchangeForRedelivery()` to restore redelivery headers from internal state rather than saving/restoring from exchange headers — resolves a [longstanding TODO](https://github.com/apache/camel/blob/557b15f88e63cb9dfc8804eac7fbb8e5302b5e07/core/camel-core-processor/src/main/java/org/apache/camel/processor/errorhandler/RedeliveryErrorHandler.java#L1280-L1281) present since Camel 2.8
- Add regression test (`RedeliverToSubRouteRemoveHeadersTest`) with two test methods: one verifying the infinite loop is fixed, one verifying `onRedelivery` processors see the correct counter

**Scenario:** parent route with `onException`/`maximumRedeliveries(3)` calls a child route (via `direct:`) that uses `NoErrorHandler` and `removeHeaders("*")`. The `removeHeaders` wipes `CamelRedeliveryCounter` during each attempt. Before this fix, the counter reset to 1 every time, causing an infinite redelivery loop.

The defensive-copy mechanism (`copyFrom(original)`) was introduced in Camel 2.8.0, but `incrementRedeliveryCounter()` was never updated to use the internal field — it kept reading from the header as it had since before the defensive copy existed. In Camel 2.7, `prepareExchangeForRedelivery()` did not copy from the original, so headers survived intact and this bug did not manifest.

---

Fix three instances of internal state being read from mutable exchange message headers instead of authoritative internal fields, making them vulnerable to removeHeaders("*") in sub-routes.

- *decrementRedeliveryCounter()*: Moved from the outer RedeliveryErrorHandler class into RedeliveryTask where the internal redeliveryCounter field is accessible. Uses the same Math.max(header, field) pattern as incrementRedeliveryCounter() to respect nested error handler counters while falling back to the internal field when headers are wiped.
- *PollEnricher save/restore*: Added redeliveryCounter and redeliveryMaxCounter fields to ExchangeExtension so redelivery state survives removeHeaders("*"). RedeliveryErrorHandler populates these fields alongside headers. PollEnricher now reads from ExchangeExtension instead of headers for its save/restore around aggregation with bridgeErrorHandler.                                                                                                                       
- *SagaProcessor coordinator ID*: The Long-Running-Action header is the sole storage for the saga coordinator ID. If removed, saga coordination fails silently (REQUIRED), throws (MANDATORY), or breaks compensation (REQUIRES_NEW). Now dual-stored in ExchangeExtension alongside the header. getCurrentSagaCoordinator() reads from extension first, falls back to the header for LRA protocol interoperability.

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/ExchangeExtension.java
+++ b/core/camel-api/src/main/java/org/apache/camel/ExchangeExtension.java
@@ -129,32 +129,35 @@ public interface ExchangeExtension {
     void setRedeliveryExhausted(boolean redeliveryExhausted);
 
     /**
-     * Gets the internal redelivery counter, which is resilient to header removal. Returns -1 if not set.
+     * Gets the redelivery counter. Returns -1 if not set. The corresponding header is set for backward compatibility.
      */
     int getRedeliveryCounter();
 
     /**
-     * Sets the internal redelivery counter, which is resilient to header removal. Set to -1 to clear.
+     * Sets the redelivery counter. Set to -1 to clear. The corresponding header is set for backward compatibility.
      */
     void setRedeliveryCounter(int redeliveryCounter);
 
     /**
-     * Gets the internal maximum redelivery counter, which is resilient to header removal. Returns -1 if not set.
+     * Gets the maximum redelivery counter. Returns -1 if not set. The corresponding header is set for backward
+     * compatibility.
      */
     int getRedeliveryMaxCounter();
 
     /**
-     * Sets the internal maximum redelivery counter, which is resilient to header removal. Set to -1 to clear.
+     * Sets the maximum redelivery counter. Set to -1 to clear. The corresponding header is set for backward
+     * compatibility.
      */
     void setRedeliveryMaxCounter(int redeliveryMaxCounter);
 
     /**
-     * Gets the internal saga long running action ID, which is resilient to header removal. Returns null if not set.
+     * Gets the saga long running action ID. Returns null if not set. The corresponding header is set for backward
+     * compatibility.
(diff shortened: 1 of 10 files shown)


oscerd (approved):

davsclaus:
yeah so Camel has historically stored such metadata as headers and/or exchange properties. EIPs usually use properties but the error handler was headers and as such it kinda got stuck there - end user can check those headers to track how many time attempt of redelivery and so on.

Now that the exchange has this rather new state stuff we can put it there, and copy over the info to the headers so they are read-only.

core/camel-api/src/main/java/org/apache/camel/ExchangeExtension.java
```diff
@@ -128,6 +128,36 @@ public interface ExchangeExtension {
      */
     void setRedeliveryExhausted(boolean redeliveryExhausted);
 
+    /**
+     * Gets the internal redelivery counter, which is resilient to header removal. Returns -1 if not set.
```
davsclaus: do not call this internal, it would become the official API for this information. And then we copy it over to the headers as well for backwards compatibility (as read only headers)

davsclaus (approved):

PR #23120, opened 2026-05-11
chore(components): specify custom span nesting in opentelemetry

squakez:
As we have moved the scope only for Processors, any call to the endpoint, which generates an event is not providing the scope. This is needed by custom span to nest the execution under the specific processor. Added some test and documentation to clarify this architecture constraint.

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-opentelemetry2/src/test/java/org/apache/camel/opentelemetry2/SpanToBeanTest.java
+++ b/components/camel-opentelemetry2/src/test/java/org/apache/camel/opentelemetry2/SpanToBeanTest.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.opentelemetry2;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import org.apache.camel.CamelContext;
+import org.apache.camel.CamelContextAware;
+import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
+import org.apache.camel.RoutesBuilder;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.opentelemetry2.CamelOpenTelemetryExtension.OtelTrace;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SpanToBeanTest extends OpenTelemetryTracerTestSupport {
+
+    Tracer tracer = otelExtension.getOpenTelemetry().getTracer("spanInjection");
+
+    @Override
+    protected CamelContext createCamelContext() throws Exception {
+        OpenTelemetryTracer tst = new OpenTelemetryTracer();
+        tst.setTracer(tracer);
+        tst.setContextPropagators(otelExtension.getOpenTelemetry().getPropagators());
+        CamelContext context = super.createCamelContext();
+        CamelContextAware.trySetCamelContext(tst, context);
+        tst.init(context);
+        return context;
+    }
+
+    @Test
+    void testRouteSingleRequest() throws IOException {
+        template.sendBody("direct:start", "my-body");
+        Map<String, OtelTrace> traces = otelExtension.getTraces();
+        assertEquals(1, traces.size());
+        checkTrace(traces.values().iterator().next());
+    }
+
+    private void checkTrace(OtelTrace trace) {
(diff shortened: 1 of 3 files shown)


squakez:
@davsclaus @oscerd I need you to please have a look at this one. Please, advise if the documentation explaining the design constraint is wrong or should be reworded.

Also, I wonder if this analysis is overlooking something. I understand that when we configure, for example, a bean to be called as an endpoint, then, the execution "converts" it as an event. Is there any possibility to have it as a regular processor and make the otel component to intercept it in the "InterceptStrategy" instead?

Thanks!

davsclaus (approved):

oscerd (approved):
It looks ok to me and it doesn't need to be reworded.

components/camel-opentelemetry2/src/test/java/org/apache/camel/opentelemetry2/SpanToBeanTest.java
```diff
@@ -0,0 +1,131 @@
...
+        };
+    }
+
+    class MyBean {
+        // NOTE: the commented annotation below would work only when an agent or a runtime framework (quarkus or spring)
+        // is available. We simulate it creating the Span by hand instead.
+        //@WithSpan
+        public void helloWorld() {
+            Span mySpan = tracer.spanBuilder("helloworld").startSpan();
+            System.out.println("*** " + mySpan);
```
gnodet: Looks like leftover debug output — please remove.

```suggestion
```

components/camel-opentelemetry2/src/test/java/org/apache/camel/opentelemetry2/SpanToBeanTest.java
```diff
@@ -0,0 +1,131 @@
...
+        SpanData beanProcessor = spans.get(4);
+        SpanData beanMethod = spans.get(5);
+        SpanData log = spans.get(6);
+        SpanData innerToLog = spans.get(7);
+
+        // Validate span completion
+        assertTrue(testProducer.hasEnded());
+        assertTrue(direct.hasEnded());
+        assertTrue(innerLog.hasEnded());
+        assertTrue(beanProcessor.hasEnded());
```
gnodet, line 79: `beanTo` is declared on line 66 but its completion is not validated here — all other 7 spans are checked. Looks like an oversight.

```suggestion
        assertTrue(beanTo.hasEnded());
        assertTrue(beanProcessor.hasEnded());
```

components/camel-telemetry/src/test/java/org/apache/camel/telemetry/SpanToBeanTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertEquals(Op.EVENT_RECEIVED.toString(), direct.getTag("op"));
+
+        // Validate hierarchy
+        assertNull(testProducer.getTag("parentSpan"));
+        assertEquals(testProducer.getTag("spanid"), direct.getTag("parentSpan"));
+        assertEquals(direct.getTag("spanid"), innerLog.getTag("parentSpan"));
+        assertEquals(direct.getTag("spanid"), log.getTag("parentSpan"));
+        assertEquals(direct.getTag("spanid"), toBean.getTag("parentSpan"));
+        assertEquals(toBean.getTag("spanid"), bean.getTag("parentSpan"));
+        assertEquals(bean.getTag("spanid"), beanMethod.getTag("parentSpan"));
```
gnodet, line 101: Nit: this asserts `beanMethod` is parented to the **bean processor** span, while the equivalent assertion in `camel-opentelemetry2` `SpanToBeanTest` (line 98) asserts the parent is the **endpoint** span (`beanTo`).

The difference is because the mock here explicitly reads the parent from exchange storage (`SpanStorageManagerExchange().peek(exchange)` → gets the processor), whereas the otel2 test uses implicit OTel context propagation (→ gets the endpoint scope). This is actually the core behavioral difference the documentation describes.

A short comment here noting *why* the hierarchy differs from the otel2 test would help future readers who compare the two.

components/camel-telemetry/src/test/java/org/apache/camel/telemetry/SpanToBeanTest.java
```diff
@@ -0,0 +1,151 @@
...
+                from("direct:start")
+                        .routeId("start")
+                        .log("A message")
+                        .to("bean:myBean")
+                        .process(new Processor() {
+                            @Override
+                            public void process(Exchange exchange) throws Exception {
+                                exchange.getIn().setHeader("operation", "fake");
+                            }
+                        })
```
gnodet, line 132: Nit: this `.process(...)` step is present in the telemetry test but absent from the otel2 equivalent. It doesn't produce a span and only sets a header that isn't asserted anywhere. Removing it would keep the routes identical across modules and make the behavioral difference (span nesting) easier to compare side-by-side.

gnodet (changes requested):
Thanks for documenting this architectural constraint, @squakez — this is a genuinely useful addition that will save users from confusing debugging sessions when custom spans don't nest where expected.

A couple of items to address:

**Required:**
1. Remove `System.out.println` debug output in the otel2 test
2. Add missing `beanTo.hasEnded()` assertion in the otel2 test

**Suggestions (non-blocking):**
3. The telemetry test route has an extra `.process(...)` step not present in the otel2 test — consider aligning for easier cross-module comparison
4. The two tests assert different parent hierarchies for `beanMethod` — a clarifying comment in the telemetry test would help future readers

> _Note: this is a project-conventions review and does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud._

components/camel-telemetry/src/test/java/org/apache/camel/telemetry/SpanToBeanTest.java
squakez: I don't think it's needed. It is explained in the otel2 test, where, as we miss the possibility to get properly the scope, the span belongs to the endpoint directly.

components/camel-telemetry/src/test/java/org/apache/camel/telemetry/SpanToBeanTest.java
squakez: It's on purpose. We are excluding by default the custom processors, reason why a span is not generated. I will include the process in the otel2 test as well to make it even.

gnodet (approved):
All feedback addressed — thanks for the quick turnaround, @squakez!

- `System.out.println` removed ✓
- Missing `beanTo.hasEnded()` assertion added ✓
- Routes aligned across modules ✓

LGTM.

PR #23124, opened 2026-05-11
Improve mdc.adoc

abedurftig:
I was trying to get this running in my Spring Boot application. I noticed that the documentation could be improved.

- update the usage; auto-configuration only works of the camel-mdc-starter is added
- otherwise service must be created and added
- camel.mdc properties names didn't match implementation

### DIFF
--- a/components/camel-mdc/src/main/docs/mdc.adoc
+++ b/components/camel-mdc/src/main/docs/mdc.adoc
@@ -70,6 +70,6 @@ The configuration properties for the MDC component are:
 |=======================================================================
 |Option |Default |Description
 |`camel.mdc.enabled`| false | Enable the MDC logging.
-|`camel.mdc.customHeaders` |  | Provide the exchange headers you would like to trace in MDC. Use `*` value as a wildcard to include more exchange headers at once, for example `CAMEL_HTTP_*` or only the `*` to include all available exchange headers.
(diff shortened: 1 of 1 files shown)


davsclaus:
For Spring Boot you should always use the -starter - that is for all of the camel components.

davsclaus (approved):

squakez (commented):
We don't generally explain in the components how to deal when adopting springboot or main (or quarkus). It is kind of implicity that, in order to use a component, then you have to add the component dependency or starter (or quarkus extension).

As for the property names, yes, they had to be fixed, thanks. However, I'd keep the example explanation in order to show the user how to use wildcards when they want to include any pattern for a property or header.

IMO, the only change we should add are the property names changes.

abedurftig:
Thanks for your feedback. I was not able to get it working right away. The fact that there might be a starter occurred too late, that's why I updated the docs. But if this is a convention, then maybe redundant.

Are these conventions documented somewhere?

Should I revert everything but the property names?

davsclaus:
> Thanks for your feedback. I was not able to get it working right away. The fact that there might be a starter occurred too late, that's why I updated the docs. But if this is a convention, then maybe redundant.
> 
> Are these conventions documented somewhere?

in the camel spring boot docs
https://camel.apache.org/camel-spring-boot/4.18.x/index.html


> 
> Should I revert everything but the property names?

yes thanks as suggested by @squakez

abedurftig:
Alright. There were some changes on main on the exact lines of code, so I rebased and cleaned up. So, the changes are really only the property names.

> in the camel spring boot docs

Thanks for sharing.

PR #23125, opened 2026-05-11
CAMEL-23459: Add TTL cleanup for pending async tasks to prevent memory leak

oscerd:
## Summary

Fixes memory leak in camel-docling component where pending async conversion tasks accumulate indefinitely in the shared `pendingAsyncTasks` map.

## Changes

- **AsyncTaskEntry**: New wrapper class that tracks `CompletableFuture` with creation timestamp
- **asyncTaskTtl**: New configuration option (default 86400000ms / 24 hours)
- **Scheduled cleanup**: Background task runs every 10% of TTL (minimum 1 second) to evict expired entries
- **Logging**: DEBUG-level logging for lifecycle, evictions, and cleanup summary
- **Documentation**: Updated `docling-component.adoc` with new configuration option
- **Tests**: Added `DoclingAsyncTaskTtlTest` with three test scenarios (Awaitility-based, no `Thread.sleep`)

## Risk Assessment

**Low risk** - The change is additive and backward compatible:
- Existing behavior unchanged (tasks still tracked and consumed normally)
- New cleanup only removes entries that would never be consumed
- Default TTL (24 hours) is conservative for typical use cases
- Cleanup runs in background thread, no impact on processing

## Testing

- Unit tests verify TTL eviction after configured timeout
- Unit tests verify tasks not evicted before TTL expires
- Unit tests verify multiple tasks evicted correctly
- camel-docling module: 49 tests, 0 failures, 0 errors
- Full reactor build (`mvn clean install -DskipTests`): SUCCESS, no uncommitted regen artifacts

## Review feedback addressed (7c1479fa90e)

- Replaced `Thread.sleep` in tests with Awaitility `during/atMost` pattern (@gnodet)
- Lowered cleanup-interval floor from 60s → 1s so short TTLs actually fire (@gnodet)
- Corrected misleading test comment and added scheduler-jitter headroom (@gnodet)
- Lowered routine lifecycle logs from `INFO` → `DEBUG` (@gnodet)
- Regenerated stale top-level catalog/DSL artifacts (`docling.json`, `DoclingComponentBuilderFactory`, `DoclingEndpointBuilderFactory`) (@davsclaus, @gnodet)
- Updated pre-existing `DoclingAsyncConversionTest` to use `AsyncTaskEntry` (drive-by — the old test was still inserting raw `CompletableFuture` and throwing `ClassCastException`)

## Related

- JIRA: https://issues.apache.org/jira/browse/CAMEL-23459
- Component: camel-docling (async conversion feature)

---
_Bob on behalf of Andrea Cosentino_


Croway (approved):

davsclaus (approved):

davsclaus:
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/docling.json
	modified:   dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/DoclingComponentBuilderFactory.java
	modified:   dsl/camel-endpointdsl/src/generated/java/org/apache/camel/builder/endpoint/dsl/DoclingEndpointBuilderFactory.java

components/camel-ai/camel-docling/src/test/java/org/apache/camel/component/docling/DoclingAsyncTaskTtlTest.java
```diff
@@ -0,0 +1,131 @@
...
+        String taskId = "test-task-2";
+        AsyncTaskEntry entry = new AsyncTaskEntry(taskId, new java.util.concurrent.CompletableFuture<>());
+        component.getPendingAsyncTasks().put(taskId, entry);
+
+        // Verify task is present
+        assertEquals(1, component.getPendingAsyncTasks().size());
+
+        // Wait a short time (less than TTL)
+        Thread.sleep(1000);
+
```
gnodet: `Thread.sleep()` is not allowed in test code per project conventions — please use Awaitility instead. For example:

```java
// Verify task remains present during the wait period (well below TTL)
await().during(1, TimeUnit.SECONDS)
       .atMost(2, TimeUnit.SECONDS)
       .untilAsserted(() -> assertEquals(1, component.getPendingAsyncTasks().size()));
```

This is more robust and avoids flaky timing issues.

components/camel-ai/camel-docling/src/test/java/org/apache/camel/component/docling/DoclingAsyncTaskTtlTest.java
```diff
@@ -0,0 +1,131 @@
...
+        String taskId = "test-task-1";
+        AsyncTaskEntry entry = new AsyncTaskEntry(taskId, new java.util.concurrent.CompletableFuture<>());
+        component.getPendingAsyncTasks().put(taskId, entry);
+
+        // Verify task is present
+        assertEquals(1, component.getPendingAsyncTasks().size());
+
+        // Wait for TTL to expire and cleanup to run
+        // Cleanup runs every 10% of TTL (minimum 1 minute), but for 2s TTL that's 200ms
+        // Add buffer for cleanup execution
```
gnodet: Nit: this comment says "for 2s TTL that's 200ms" but the production code does `Math.max(ttl / 10, 60000)`, which clamps to a **1-minute minimum**, not 200ms.

If the cleanup interval is really 60s, the `await().atMost(5, SECONDS)` on line 60 will time out before the first cleanup fires. Worth verifying this test passes reliably — the comment at minimum should be corrected.

components/camel-ai/camel-docling/src/main/java/org/apache/camel/component/docling/DoclingComponent.java
```diff
@@ -51,6 +58,26 @@ public DoclingComponent(CamelContext context) {
...
+        cleanupExecutor = getCamelContext().getExecutorServiceManager()
+                .newScheduledThreadPool(this, "DoclingAsyncTaskCleanup", 1);
+
+        cleanupExecutor.scheduleWithFixedDelay(
+                this::cleanupExpiredTasks,
+                cleanupInterval,
+                cleanupInterval,
+                TimeUnit.MILLISECONDS);
+
+        LOG.info("Started async task cleanup with TTL={}ms, cleanup interval={}ms", ttl, cleanupInterval);
```
gnodet: Suggestion: consider using `LOG.debug(...)` here and at the other `LOG.info` calls in `cleanupExpiredTasks()` (line 142) and `doStop()` (line 153). For a component library, routine lifecycle events and periodic housekeeping are typically logged at DEBUG to avoid noisy output for end users.

gnodet (changes requested):
Nice work addressing the memory leak, Andrea — the TTL-based eviction approach is clean and backward-compatible.

A few items to address before this can merge:

### Blocking

1. **Missing regenerated files** — CI is failing on the "uncommitted changes" check. As Claus already noted, `catalog/camel-catalog/.../docling.json`, `dsl/camel-componentdsl/.../DoclingComponentBuilderFactory.java`, and `dsl/camel-endpointdsl/.../DoclingEndpointBuilderFactory.java` need to be regenerated and committed.

2. **`Thread.sleep()` in test code** — `testAsyncTaskNotEvictedBeforeTtl()` uses `Thread.sleep(1000)`. The project convention requires Awaitility instead (see CLAUDE.md "Asynchronous Testing" section). See inline comment for a suggested replacement.

### Non-blocking suggestions

3. **Test comment vs actual behavior** — The comment in `testAsyncTaskTtlEviction` says "for 2s TTL that's 200ms" but the code enforces a 1-minute minimum (`Math.max(ttl / 10, 60000)`). If the cleanup interval is actually 60s, the `await().atMost(5, SECONDS)` assertion may be fragile. Worth double-checking this passes reliably.

4. **Logging level** — `doStart()`, `doStop()`, and periodic cleanup all log at `INFO`. For a component library, routine lifecycle and housekeeping are typically `DEBUG` to avoid noisy output for end users. See inline comment.

components/camel-ai/camel-docling/src/test/java/org/apache/camel/component/docling/DoclingAsyncTaskTtlTest.java
oscerd: Fixed in 7c1479fa90e — replaced `Thread.sleep(1000)` with the suggested Awaitility `during/atMost` pattern.

components/camel-ai/camel-docling/src/test/java/org/apache/camel/component/docling/DoclingAsyncTaskTtlTest.java
oscerd: Good catch — you're right that the comment was wrong and, more importantly, with the 60s minimum the cleanup interval was *60s* for a 2s TTL test, so the `await().atMost(5, SECONDS)` was unreachable.

Fixed in 7c1479fa90e by:
1. Lowering the cleanup-interval floor from 60s to **1s**. The cleanup is cheap when no tasks are expired (early-returns on empty map; otherwise just a hashmap iteration), so the lower minimum is safe and makes the configurable TTL actually meaningful for short values. For the default 24h TTL this changes nothing (ttl/10 = 2.4h > 1s either way).
2. Correcting the test comment to match: "Cleanup runs every 10% of TTL with a 1 second minimum, so for a 2s TTL the interval is 1s."
3. Bumping `atMost` from 5s → 8s to give headroom for TTL (2s) + first cleanup tick (~1s) + scheduler jitter.
4. Also overrode `isUseAdviceWith()` to return `true` — otherwise CamelTestSupport auto-starts the context (and the component's `doStart`) *before* the test code sets the TTL, so the scheduler was being built with the default 24h TTL even when the test set 2s.

All 3 tests in `DoclingAsyncTaskTtlTest` now pass deterministically (~7s for the whole class).

components/camel-ai/camel-docling/src/main/java/org/apache/camel/component/docling/DoclingComponent.java
oscerd: Fixed in 7c1479fa90e — lowered all three call sites (`doStart`, `cleanupExpiredTasks` summary line, `doStop`) from `LOG.info` to `LOG.debug`. The per-task eviction log was already at DEBUG.

oscerd:
Pushed 7c1479fa90e addressing all of @gnodet's and @davsclaus's review feedback:

**Blocking items**
- **Stale generated files** (@davsclaus / @gnodet "Blocking 1"): regenerated and committed `catalog/.../docling.json`, `DoclingComponentBuilderFactory.java`, and `DoclingEndpointBuilderFactory.java`. Verified that a clean `mvn clean install -DskipTests` from the repo root now produces no further uncommitted regen artifacts.
- **`Thread.sleep` in tests** (@gnodet "Blocking 2"): replaced with the suggested Awaitility `during/atMost` pattern.

**Non-blocking suggestions (both applied)**
- **Test comment vs actual behavior** (@gnodet #3): this turned out to be more than a comment error — with the 60s minimum, the cleanup-interval was actually 60s for the 2s test TTL, so the existing `atMost(5, SECONDS)` could never fire. Lowered the cleanup-interval floor from 60s → 1s (cleanup is cheap when the map is empty, so the lower minimum is safe and makes short TTLs actually meaningful), corrected the test comment, bumped `atMost` to 8s for headroom, and added `isUseAdviceWith() = true` so the test-configured TTL is in effect by the time `doStart` runs.
- **Logging level** (@gnodet #4): lowered `doStart`/`doStop`/cleanup-summary lines from `LOG.info` to `LOG.debug`.

**Drive-by**
While running the test suite I also found that the pre-existing `DoclingAsyncConversionTest` wasn't updated when the PR changed the map's value type to `AsyncTaskEntry` — it was still inserting raw `CompletableFuture` and producing `ClassCastException` on 3 tests. Updated the test to wrap in `AsyncTaskEntry`.

**Verification**
- camel-docling: 49 tests, 0 failures, 0 errors (was 3 ClassCastException errors on `DoclingAsyncConversionTest` + flaky `DoclingAsyncTaskTtlTest`).
- Full reactor `mvn clean install -DskipTests` from root: SUCCESS in 8:42.
- `git status` clean after the full reactor build — no further regen artifacts.

cc @Croway @davsclaus @gnodet for re-review when convenient.

davsclaus (approved):

Croway (approved):

gnodet (approved):
All feedback addressed — Thread.sleep replaced with Awaitility, misleading comment fixed, LOG.info demoted to DEBUG, and generated files regenerated. CI is green. LGTM.

PR #23127, opened 2026-05-11
CAMEL-23302: camel-xslt - Auto-disable content cache in routes-reload (dev) mode

ammachado:
# Description

[CAMEL-23302: camel-xslt - Auto-disable content cache in routes-reload (dev) mode](https://issues.apache.org/jira/browse/CAMEL-23302)

When `camel.main.routesReloadEnabled=true` (set automatically by `camel run --dev`), Camel now sets `contentCache` to `false` on resource-based components (components exposing `contentCache` option, such as xslt, freemarker, velocity, mustache, jte, jolt, jslt, jsonata, json-validator, json-patch, mvel, chunk, stringtemplate, language) so that edits to the resource file are picked up on the next message without restarting the route.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

# Target

# Tracking


components/camel-xslt/src/main/java/org/apache/camel/component/xslt/XsltComponent.java
```diff
@@ -102,7 +108,8 @@ public boolean isContentCache() {
      * stylesheet file on each message processing. This is good for development. A cached stylesheet can be forced to
      * reload at runtime via JMX using the clearCachedStylesheet operation.
      */
-    public void setContentCache(boolean contentCache) {
+    @Override
+    public void setContentCache(Boolean contentCache) {
         this.contentCache = contentCache;
     }
```
gnodet: The `setContentCache(boolean)` → `setContentCache(Boolean)` signature change breaks binary backward compatibility (JVM method descriptor changes from `(Z)V` to `(Ljava/lang/Boolean;)V`). Existing compiled code calling `component.setContentCache(true)` would get a `NoSuchMethodError` at runtime.

Consider keeping the old `boolean` overload as a deprecated bridge:

```suggestion
    @Override
    public void setContentCache(Boolean contentCache) {
        this.contentCache = contentCache;
    }

    /**
     * @deprecated use {@link #setContentCache(Boolean)}
     */
    @Deprecated
    public void setContentCache(boolean contentCache) {
        this.contentCache = contentCache;
    }
```

core/camel-main/src/main/java/org/apache/camel/main/DevModeContentCacheStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Lifecycle strategy that disables content caching on resource-based components when routes-reload is enabled. Lets
+ * users edit a resource file (e.g. an XSLT stylesheet) and see the change applied live, without having to set
+ * {@code contentCache=false} on every endpoint.
+ *
+ * Only flips components that have not been explicitly configured by the user (i.e.
+ * {@link ContentCacheAware#getContentCache()} returns {@code null}); explicit user settings are preserved.
+ */
+public class DevModeContentCacheStrategy extends LifecycleStrategySupport {
```
gnodet: Nit: this class is only instantiated from `DefaultConfigurationConfigurer` in the same package. Consider reducing visibility to package-private:

```suggestion
class DevModeContentCacheStrategy extends LifecycleStrategySupport {
```

gnodet (commented):
Nice work on this PR! The tri-state `Boolean` approach for the `ContentCacheAware` SPI is a clean design, and the lifecycle strategy is the right integration point for dev-mode behavior.

A few items — see inline comments for specific suggestions:

1. **Binary backward compatibility** (setter signature change): see inline suggestion on `XsltComponent.java`
2. **Class visibility**: `DevModeContentCacheStrategy` could be package-private (see inline)
3. **JIRA link**: Consider adding `[CAMEL-23302](https://issues.apache.org/jira/browse/CAMEL-23302)` to the PR description body
4. **Upgrade guide**: Since `ContentCacheAware` is a new public SPI in `camel-api`, consider mentioning it in the 4.21 upgrade guide so other component authors know they can opt in

components/camel-xslt/src/main/java/org/apache/camel/component/xslt/XsltComponent.java
```diff
@@ -45,7 +46,7 @@ public class XsltComponent extends DefaultComponent {
     @Metadata
     private boolean allowTemplateFromHeader;
     @Metadata(defaultValue = "true")
-    private boolean contentCache = true;
+    private Boolean contentCache;
```
Croway: isn't the default false?

Croway:
@ammachado there are uncommitted changes
```
There are uncommitted changes
HEAD detached at pull/23127/merge
	modified:   dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/XjComponentBuilderFactory.java
	modified:   dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/XsltComponentBuilderFactory.java
	modified:   dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/XsltSaxonComponentBuilderFactory.java
```

You can `cd dsl && man clean install -DskipTests` once camel-xslt is built locally to quickly generate those files

gnodet:
**Alternative approach worth considering: use the existing `PropertyConfigurer` API instead of a new SPI**

Rather than introducing `ContentCacheAware` in `camel-api` and migrating components one by one, the `DevModeContentCacheStrategy` could use the existing generated `PropertyConfigurer` to detect and set the `contentCache` property on any component that has it:

```java
@Override
public void onComponentAdd(String name, Component component) {
    PropertyConfigurer configurer = PluginHelper.getConfigurerResolver(component.getCamelContext())
        .resolvePropertyConfigurer(name + "-component", component.getCamelContext());
    if (configurer instanceof PropertyConfigurerGetter getter) {
        Object val = getter.getOptionValue(component, "contentCache", true);
        if (Boolean.TRUE.equals(val)) {
            configurer.configure(component.getCamelContext(), component, "contentCache", false, true);
            LOG.info("Routes-reload is enabled: disabling contentCache on component '{}' for live resource reload", name);
        }
    }
}
```

**Advantages:**
- No new SPI interface in `camel-api` (smaller API surface)
- Works immediately for **all ~13 components** with `contentCache` (freemarker, velocity, mustache, jte, jolt, jslt, jsonata, json-validator, json-patch, mvel, chunk, stringtemplate, language) — no per-component migration needed
- No `boolean` → `Boolean` field change, so no binary compatibility concern
- The timing works for the common case: user properties via `camel.component.<name>.contentCache=true` are bound *after* `onComponentAdd`, so they naturally override the strategy's change

**Downside:**
- Cannot distinguish "default `true`" from "user explicitly set `true` programmatically before adding the component to the context" (e.g., `component.setContentCache(true); context.addComponent(...)`). The strategy would wrongly override it in that scenario. The tri-state `Boolean` (null = unset) in the current PR handles this edge case correctly.
- That said, programmatic pre-registration configuration is uncommon with Camel Main, which is the primary consumer of this feature.

Not a blocker — just flagging the trade-off for consideration.

components/camel-xslt/src/main/java/org/apache/camel/component/xslt/XsltComponent.java
ammachado: Using `PropertyConfigurer` implementation instead.

davsclaus (approved):

PR #23178, opened 2026-05-13
CAMEL-23495: Replace Thread.sleep() with Awaitility and fix ~30 flaky tests

gnodet:
[CAMEL-23495](https://issues.apache.org/jira/browse/CAMEL-23495)

## Summary

Fix ~30 flaky tests across `camel-core` and `camel-management` by replacing `Thread.sleep()` with Awaitility, increasing tight timeouts, fixing race conditions, and fixing bugs. Changes span 32 test files.

## Root causes addressed

| Category | Tests | Fix |
|---|---|---|
| **Thread.sleep() for synchronization** | Suspend/resume tests, management tests, executor tests, redelivery tests (~15 files) | Replace with `await().atMost().until()` or `untilAsserted()` |
| **Tight timeouts under CI load** | Seda, multicast, file consumer, aggregator tests (~10 files) | Increase `resultWaitTime`, assertion timeouts, poll delays |
| **Race conditions** | `AggregateForceCompletionOnStopParallelTest` | Awaitility after async `context.stop()` / `stopRoute()` |
| **File consumer partial reads** | `MarkerFileExclusiveReadLockStrategy*` tests | Increase `initialDelay` from 0 to 1000ms to prevent reading during writes |
| **Bugs** | `AsyncEndpointRedeliveryErrorHandlerNonBlockedDelay2Test` | Reset static counter; fix `mock:result` → `mock:before` copy-paste bug |
| **Deterministic ordering** | `MulticastParallelStreamingTest`, `RecipientListParallelStreamingTest` | Use `syncDelayed()` instead of `asyncDelayed()` |

## Verification

All 30 modified test classes were run in a **100-iteration loop** to verify stability:

- **camel-core (29 tests):** **0 failures out of 100 iterations**
- **camel-management (1 test):** **0 failures out of 100 iterations**

Note: `AggregateCompleteAllOnStopWithIntervalTest` was reverted to its original state — it has a real bug where `completeAllOnStop()` with `completionInterval()` never completes aggregations during shutdown. Tracked as [CAMEL-23513](https://issues.apache.org/jira/browse/CAMEL-23513).

_Claude Code on behalf of Guillaume Nodet_

PR #23180, opened 2026-05-13
CAMEL-23495: Replace Thread.sleep() with Awaitility in camel-core and camel-management tests

gnodet:
[CAMEL-23495](https://issues.apache.org/jira/browse/CAMEL-23495)

Replace `Thread.sleep()` synchronization calls with Awaitility in 15 test files across camel-core and camel-management to reduce flakiness under CI load.

### Suspend/resume tests (4 files)
- `RouteSedaSuspendResumeTest`, `DefaultCamelContextSuspendResumeRouteTest`, `TwoRouteSuspendResumeTest`, `SedaConsumerSuspendResumeTest`
- Replaced `Thread.sleep(1000L)` + existing `await()` with a single `await().pollDelay(1s).atMost(5s)` to give SEDA consumer thread time to idle after route suspension

### Management tests (2 files)
- `ManagedInflightStatisticsTest` — replaced `Thread.sleep(250)` and `Thread.sleep(200)` with `await().until()` checking JMX inflight exchange count
- `DefaultExecutorServiceManagerTest` — replaced `Thread.sleep(3000)` with a `started` CountDownLatch + `await()` to detect when the pool task begins

### Async/scheduler tests (2 files)
- `SchedulerMulticastParallelGreedyTest` — replaced `Thread.sleep(50)` with `MockEndpoint.setAssertPeriod(200)` to verify no extra messages arrive
- `ThreadsRejectedExecutionWithDeadLetterTest` — replaced `Thread.sleep(100)` with `await().until()` checking the rejected message arrived at dead letter

### Aggregation tests (3 files)
- `AggregateGroupedExchangeBatchSizeTest` — replaced `Thread.sleep(1000)` with `await()` for the second batch via completion timeout
- `AggregateDiscardOnTimeoutTest` — replaced `Thread.sleep(250)` with `MockEndpoint.setAssertPeriod(500)`
- `AggregateClosedCorrelationKeyTest` — replaced `Thread.sleep(200)` with `await().until()` checking all aggregated results arrived

### Redelivery/shutdown tests (3 files)
- `NotAllowRedeliveryWhileStoppingTest`, `NotAllowRedeliveryWhileStoppingDeadLetterChannelTest` — replaced `Thread.sleep(500)` with `await()` checking inflight repository
- `RedeliveryDeadLetterErrorHandlerNoRedeliveryOnShutdownTest` — replaced `Thread.sleep(500)` with `await().until(() -> counter.get() >= 20)`

### Resequencer test (1 file)
- `ResequenceStreamRejectOldExchangesTest` — replaced `Thread.sleep(100)` with `await().until()` checking first message delivered before sending the rest

## Out of scope
The following `Thread.sleep()` uses were intentionally left unchanged:
- Sleeps inside `Processor.process()` or `Runnable.run()` simulating slow I/O
- Timing-specific tests (completion intervals, rate calculations, timeout ordering)
- Pacing sleeps between message sends where the delay is the test's intent

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
@@ -59,11 +59,14 @@ public class SedaConsumerSuspendResumeTest extends ContextTestSupport {
         // mode where
         // it would poll and route (there is a little slack (up till 1 sec)
         // before suspension is empowered)
-        // wait for queues to empty and consumer to complete its poll cycle
-        await().pollDelay(1, TimeUnit.SECONDS)
-                .atMost(5, TimeUnit.SECONDS)
+        // wait for queues to empty
+        await().atMost(5, TimeUnit.SECONDS)
                 .until(() -> context.getEndpoint("seda:foo", SedaEndpoint.class).getQueue().isEmpty()
                         && context.getEndpoint("seda:bar", SedaEndpoint.class).getQueue().isEmpty());
+        // give consumer thread time to idle after queues empty
+        await().pollDelay(1, TimeUnit.SECONDS)
+                .atMost(5, TimeUnit.SECONDS)
+                .untilAsserted(() -> assertEquals("Suspended", consumer.getStatus().name()));
 
         template.sendBody("seda:foo", "B");
         // wait a little to ensure seda consumer thread would have tried to poll
--- a/core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
@@ -59,15 +59,11 @@ public class SedaConsumerSuspendResumeTest extends ContextTestSupport {
         // mode where
         // it would poll and route (there is a little slack (up till 1 sec)
         // before suspension is empowered)
-        await().atMost(1, TimeUnit.SECONDS)
+        // wait for queues to empty and consumer to complete its poll cycle
+        await().pollDelay(1, TimeUnit.SECONDS)
+                .atMost(5, TimeUnit.SECONDS)
                 .until(() -> context.getEndpoint("seda:foo", SedaEndpoint.class).getQueue().isEmpty()
                         && context.getEndpoint("seda:bar", SedaEndpoint.class).getQueue().isEmpty());
-
-        // even though we wait for the queues to empty, there is a race condition where the consumer
-        // may still process messages while it's being suspended due to asynchronous message handling.
-        // as a result, we need to wait a bit longer to ensure that the seda consumer is suspended before
-        // sending the next message.
-        Thread.sleep(1000L);
 
         template.sendBody("seda:foo", "B");
         // wait a little to ensure seda consumer thread would have tried to poll
--- a/core/camel-core/src/test/java/org/apache/camel/impl/DefaultCamelContextSuspendResumeRouteTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/DefaultCamelContextSuspendResumeRouteTest.java
@@ -49,9 +49,12 @@ public class DefaultCamelContextSuspendResumeRouteTest extends ContextTestSuppor
 
         context.suspend();
 
-        // wait for the seda consumer to complete its current poll cycle and become idle
+        // wait for the context to be fully suspended
         Awaitility.await().atMost(5, TimeUnit.SECONDS)
-                .pollDelay(1, TimeUnit.SECONDS)
+                .until(() -> context.isSuspended());
+        // give seda consumer thread time to complete its current poll cycle
+        Awaitility.await().pollDelay(1, TimeUnit.SECONDS)
+                .atMost(5, TimeUnit.SECONDS)
                 .untilAsserted(() -> Assertions.assertDoesNotThrow(() -> template.sendBody("seda:foo", "B")));
 
         mock.assertIsSatisfied(1000);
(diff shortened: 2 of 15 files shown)


davsclaus (approved):

core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
```diff
@@ -59,16 +59,12 @@ public void testSuspendResume() throws Exception {
...
+                .atMost(5, TimeUnit.SECONDS)
                 .until(() -> context.getEndpoint("seda:foo", SedaEndpoint.class).getQueue().isEmpty()
                         && context.getEndpoint("seda:bar", SedaEndpoint.class).getQueue().isEmpty());
 
-        // even though we wait for the queues to empty, there is a race condition where the consumer
-        // may still process messages while it's being suspended due to asynchronous message handling.
-        // as a result, we need to wait a bit longer to ensure that the seda consumer is suspended before
-        // sending the next message.
-        Thread.sleep(1000L);
-
```
apupier, line 71: there is no await replacement then this thread;sleep. This could only be worse than it currently is

core/camel-core/src/test/java/org/apache/camel/impl/DefaultCamelContextSuspendResumeRouteTest.java
```diff
@@ -49,15 +49,9 @@ public void testSuspendResume() throws Exception {
 
         context.suspend();
 
-        // even though we wait for the route to suspend, there is a race condition where the consumer
-        // may still process messages while it's being suspended due to asynchronous message handling.
-        // as a result, we need to wait a bit longer to ensure that the seda consumer is suspended before
-        // sending the next message.
-        Thread.sleep(1000L);
```
apupier, line 56: there isn't a wait condition anymore ensuring that the message is suspended

core/camel-core/src/test/java/org/apache/camel/impl/TwoRouteSuspendResumeTest.java
```diff
@@ -49,16 +49,10 @@ public void testSuspendResume() throws Exception {
...
-        // need to give seda consumer thread time to idle
-        await().atMost(1, TimeUnit.SECONDS).until(() -> {
-            return context.getEndpoint("seda:foo", SedaEndpoint.class).getQueue().isEmpty();
-        });
-
-        // even though we wait for the queues to empty, there is a race condition where the consumer
-        // may still process messages while it's being suspended due to asynchronous message handling.
-        // as a result, we need to wait a bit longer to ensure that the seda consumer is suspended before
-        // sending the next message.
-        Thread.sleep(1000L);
```
apupier, line 61: there is no await replacement then this thread;sleep. This could only be worse than it currently is

apupier (changes requested):

core/camel-core/src/test/java/org/apache/camel/impl/TwoRouteSuspendResumeTest.java
gnodet: What about the `pollDelay(1, TimeUnit.SECONDS)` ?

core/camel-core/src/test/java/org/apache/camel/impl/DefaultCamelContextSuspendResumeRouteTest.java
gnodet: What about the `pollDelay(1, TimeUnit.SECONDS)` ?

core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
gnodet: What about the `pollDelay(1, TimeUnit.SECONDS)` ?

core/camel-core/src/test/java/org/apache/camel/impl/TwoRouteSuspendResumeTest.java
apupier: The poll Delay is before checking for for the suspension and the big comment is explaining that we need to wait *after* this point due to async behavior

gnodet:
Thanks for the review! You're right — the `pollDelay` was placed before the condition check, but the original `Thread.sleep` was after the queue-empty check. Fixed in ef247030d5b8: now using two-step await:

1. First `await().until(queue.isEmpty())` — checks the condition
2. Second `await().pollDelay(1s).untilAsserted(...)` — gives the consumer thread time to idle after the queue empties

core/camel-core/src/test/java/org/apache/camel/impl/TwoRouteSuspendResumeTest.java
gnodet: You're right, the `pollDelay` was before the condition check, but the original `Thread.sleep` was after the queue-empty await. Fixed in ef247030 — now split into two steps:

1. `await().until(queue.isEmpty())` — checks the condition first
2. `await().pollDelay(1s).untilAsserted(...)` — then gives the consumer thread time to idle

This preserves the original semantics: wait for the queue to empty, *then* wait for the consumer to stop its poll loop.

core/camel-core/src/test/java/org/apache/camel/component/seda/SedaConsumerSuspendResumeTest.java
gnodet: Fixed in ef247030 — now split into two `await()` calls: first checks queues are empty, then uses `pollDelay(1s)` to give the consumer thread time to idle after the queues empty. The second await also asserts the consumer status is Suspended.

core/camel-core/src/test/java/org/apache/camel/impl/DefaultCamelContextSuspendResumeRouteTest.java
gnodet: Fixed in ef247030 — now split into two `await()` calls: first checks `context.isSuspended()`, then uses `pollDelay(1s)` to give the seda consumer thread time to complete its poll cycle before sending the test message.

gnodet:
Commits cherry-picked into #23178 which consolidates all flaky test fixes.

PR #23183, opened 2026-05-13
CAMEL-23497: Fix NPE in ManagedTransformerRegistry.listTransformers()

gnodet:
[CAMEL-23497](https://issues.apache.org/jira/browse/CAMEL-23497)

Handle null `from`/`to` DataType in `ManagedTransformerRegistry.listTransformers()`. The `Transformer.getFrom()` and `getTo()` methods are `@Nullable` — when a transformer is registered by name only (without explicit from/to types), these can be null, causing an NPE when building the JMX CompositeData.

- Add null checks for `from`/`to` in `listTransformers()`, using empty string when null
- Fix incorrect test assertions that expected `"camel:any"` for a named transformer with null from/to
- Add dedicated test `testListTransformersWithNullFromTo` that registers a transformer with null from/to directly in the registry and verifies `listTransformers()` handles it correctly

### DIFF
--- a/core/camel-management/src/main/java/org/apache/camel/management/mbean/ManagedTransformerRegistry.java
+++ b/core/camel-management/src/main/java/org/apache/camel/management/mbean/ManagedTransformerRegistry.java
@@ -94,7 +94,9 @@ public class ManagedTransformerRegistry extends ManagedService implements Manage
 
                 CompositeData data = new CompositeDataSupport(
                         ct, new String[] { "name", "from", "to", "static", "dynamic", "description" },
-                        new Object[] { name, from.toString(), to.toString(), fromStatic, fromDynamic, desc });
+                        new Object[] {
+                                name, from != null ? from.toString() : "", to != null ? to.toString() : "",
+                                fromStatic, fromDynamic, desc });
                 answer.put(data);
             }
             return answer;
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

apupier (approved):
would be nice to add a test covering this case

EDIT: in fact, this was caught by a test  based on https://issues.apache.org/jira/browse/CAMEL-23497

gnodet (commented):

Regarding the `"camel:any"` → `""` change in the existing test assertions:

`MyTransformer` extends `Transformer` with no `@DataTypeTransformer` annotation and is registered via `transformer().name("custom").withJava(MyTransformer.class)`. The `Transformer` constructor only sets `from`/`to` if the annotation is present — so they stay `null`.

The CAMEL-18698 commit changed the test expectation from `assertEquals(null, from)` to `assertEquals("camel:any", from)`, but that was incorrect — `from` was never actually `"camel:any"`, it was `null`. The test never passed after that change because the NPE on line 97 (`from.toString()`) would throw *before* the assertion was even reached.

With the null-safety fix, `listTransformers()` now returns `""` for null DataTypes instead of throwing. So the assertion needs to match what actually comes back: `""`.

PR #23185, opened 2026-05-13
CAMEL-22640: Guard JSSE factory beans against null CamelContext

Croway:
Fixes the test failure `org.apache.camel.util.spring.SecureRandomParametersFactoryBeanTest.testKeyStoreParameters`

## Summary

- The JSpecify commit (CAMEL-22640) added `Objects.requireNonNull` to `JsseParameters.setCamelContext()`, which is correct per the `CamelContextAware` contract.
- The JSSE factory beans in `camel-core-xml` unconditionally called `newInstance.setCamelContext(getCamelContext())`, passing `null` when no `CamelContext` was injected (e.g., Spring XML tests without a `<camelContext>` element). This now throws `NullPointerException`.
- Added a `setCamelContextIfPresent()` helper in `AbstractJsseUtilFactoryBean` that only propagates the context when non-null, and replaced all 16 call sites across 8 factory beans.

## Test plan

- [x] `SecureRandomParametersFactoryBeanTest` passes (was previously failing with NPE)
- [x] `SSLContextParametersFactoryBeanTest` passes
- [x] `KeyStoreParametersFactoryBeanTest` passes

### DIFF
--- a/core/camel-core-xml/src/main/java/org/apache/camel/core/xml/util/jsse/AbstractBaseSSLContextParametersFactoryBean.java
+++ b/core/camel-core-xml/src/main/java/org/apache/camel/core/xml/util/jsse/AbstractBaseSSLContextParametersFactoryBean.java
@@ -19,6 +19,7 @@ package org.apache.camel.core.xml.util.jsse;
 import jakarta.xml.bind.annotation.XmlAttribute;
 import jakarta.xml.bind.annotation.XmlTransient;
 
+import org.apache.camel.CamelContextAware;
 import org.apache.camel.spi.Metadata;
 import org.apache.camel.support.jsse.BaseSSLContextParameters;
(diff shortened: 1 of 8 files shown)


davsclaus:
you can use `trySetCamelContext` on `CamelContextAware` which is intended for this

oscerd (approved):

gnodet (approved):
Looks good — the fix correctly guards all 16 JSSE factory bean call sites against null `CamelContext` after the JSpecify `Objects.requireNonNull` addition.

One suggestion (already noted by @davsclaus): consider using the existing `CamelContextAware.trySetCamelContext(object, getCamelContext())` static method instead of the new `setCamelContextIfPresent()` helper — it does exactly the same null check and is the established pattern in the codebase.

davsclaus (approved):

PR #23193, opened 2026-05-13
Improve Javadoc across camel-api: URLs, Exchange, SPI contracts, and exception constructors

gnodet:
## Summary

Third batch of Javadoc improvements for the camel-api module:

- **HTTPS URLs**: Update all 67 `http://camel.apache.org` URLs to `https://` across 39 Javadoc comments (excluding XML namespace URIs)
- **Broken link fix**: Replace dead Spring 2.5.x API link in `Synchronization.java` with `{@code}` FQCN reference
- **Exchange rewrite**: Rewrite class-level Javadoc to lead with `getMessage()` instead of deprecated `getIn()/getOut()` pattern, with `ExchangePattern` mention preserved
- **SPI contracts**: Add contract documentation to `Language`, `AggregationRepository`, and `DataFormatFactory` explaining lifecycle, thread-safety, and usage
- **Exception constructors**: Add `@param` tags to constructors of 20 most important exception classes
- **StreamCacheException**: Add missing class-level Javadoc

All changes are Javadoc-only — no functional code modified.

## Test plan

- [x] `mvn install -pl core/camel-api -DskipTests` passes
- [x] Formatter applied (`mvn formatter:format impsort:sort`)
- [x] No generated file changes needed
- [x] `@XmlSchema` namespace URI preserved as `http://` (XML identifier, not web URL)

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/Exchange.java
+++ b/core/camel-api/src/main/java/org/apache/camel/Exchange.java
@@ -46,7 +46,7 @@ import org.jspecify.annotations.Nullable;
  * <p/>
  * To access or modify the current message, use the {@link #getMessage()} method. Camel uses {@link #getMessage()} to
  * obtain the current message during routing. If a {@link Processor} modifies the message, those changes are visible to
- * subsequent processors.
+ * subsequent processors. The {@link ExchangePattern} determines whether a reply is expected (InOut) or not (InOnly).
  */
 @ConstantProvider("org.apache.camel.ExchangeConstantProvider")
 public interface Exchange extends VariableAware {
--- a/core/camel-api/src/main/java/org/apache/camel/Exchange.java
+++ b/core/camel-api/src/main/java/org/apache/camel/Exchange.java
@@ -28,9 +28,9 @@ import org.jspecify.annotations.Nullable;
  * An Exchange is the message container holding the information during the entire routing of a {@link Message} received
  * by a {@link Consumer}.
  * <p/>
- * During processing down the {@link Processor} chain, the {@link Exchange} provides access to the current (not the
- * original) request and response {@link Message} messages. The {@link Exchange} also holds meta-data during its entire
- * lifetime stored as properties accessible using the various {@link #getProperty(String)} methods. The
+ * During processing down the {@link Processor} chain, the {@link Exchange} provides access to the current
+ * {@link Message} via {@link #getMessage()}. The {@link Exchange} also holds meta-data during its entire lifetime
+ * stored as properties accessible using the various {@link #getProperty(String)} methods. The
  * {@link #setProperty(String, Object)} is used to store a property. For example, you can use this to store security,
  * SLA related data or any other information deemed useful throughout processing. If an {@link Exchange} failed during
  * routing the {@link Exception} that caused the failure is stored and accessible via the {@link #getException()}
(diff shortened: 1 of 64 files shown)


core/camel-api/src/main/java/org/apache/camel/spi/Synchronization.java
```diff
@@ -22,8 +22,7 @@
 /**
  * Provides a hook for custom {@link org.apache.camel.Processor} or {@link org.apache.camel.Component} instances to
  * respond to completed or failed processing of an {@link Exchange} rather like Spring's
- * <a href="http://static.springframework.org/spring/docs/2.5.x/api/org/springframework/transaction/
- * support/TransactionSynchronization.html">TransactionSynchronization</a>
+ * {@code TransactionSynchronization}
```
apupier: it was pointing to Spring class, I guess that when using just @code without fqcn it expects to point ot an internal class in same package

core/camel-api/src/main/java/org/apache/camel/InvalidPayloadRuntimeException.java
```diff
@@ -27,21 +27,41 @@ public class InvalidPayloadRuntimeException extends RuntimeExchangeException {
...
+     * @param cause    the cause of the failure
+     */
     public InvalidPayloadRuntimeException(Exchange exchange, Class<?> type, Throwable cause) {
         this(exchange, type, exchange.getIn(), cause);
     }
 
+    /**
+     * @param exchange the exchange that caused the error
+     * @param type     the expected body type
+     * @param message  the message whose body could not be converted
```
apupier: is it really only when the body cannot be converted? I understand from the class name that the payload is invalid but ther ei no assumption that it will be converted

core/camel-api/src/main/java/org/apache/camel/package-info.java
```diff
@@ -18,7 +18,7 @@
  * The core Camel API.
  */
 @NullMarked
-@XmlSchema(namespace = "http://camel.apache.org/schema/spring",
+@XmlSchema(namespace = "https://camel.apache.org/schema/spring",
```
apupier: I do not think that namespace can be changed

core/camel-api/src/main/java/org/apache/camel/Exchange.java
```diff
@@ -39,28 +39,14 @@
...
- * will lazily instantiate a {@link org.apache.camel.support.DefaultMessage} which you could populate. As an alternative
- * you could also instantiate your specialized {@link Message} and set it on the exchange using the
- * {@link #setOut(org.apache.camel.Message)} method. Please note that a {@link Message} contains not only the body but
- * also headers and attachments. If you are creating a new {@link Message} the headers and attachments of the in
- * {@link Message} are not automatically copied to the out by Camel, and you'll have to set the headers and attachments
- * you need yourself. If your {@link Processor} is not producing a different {@link Message} but only needs to slightly
- * modify the in, you can simply update the in {@link Message} returned by {@link #getIn()}.
+ * To access or modify the current message, use the {@link #getMessage()} method. Camel uses {@link #getMessage()} to
+ * obtain the current message during routing. If a {@link Processor} modifies the message, those changes are visible to
+ * subsequent processors.
```
apupier: what is the reason to remove several details?

apupier (changes requested):

core/camel-api/src/main/java/org/apache/camel/spi/Synchronization.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good point — updated to use the FQCN: `{@code org.springframework.transaction.support.TransactionSynchronization}` so it's clear this refers to a Spring class.

core/camel-api/src/main/java/org/apache/camel/InvalidPayloadRuntimeException.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

You're right — the description was too narrow. Updated to: `@param message the message with the invalid or missing payload`. Same fix applied to `InvalidPayloadException`.

core/camel-api/src/main/java/org/apache/camel/package-info.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch — the `@XmlSchema` namespace is an XML identifier, not a web URL. Changing `http://` to `https://` would break namespace matching. Reverted.

core/camel-api/src/main/java/org/apache/camel/Exchange.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

The removed paragraphs described the in/out message model (`getIn()`/`getOut()`/`setOut()`) which was deprecated in Camel 3.0. The old Javadoc was teaching users a deprecated pattern. The replacement directs users to `getMessage()` which is the current API. I've also added back a mention of `ExchangePattern` since MEP is still an active concept.

davsclaus (approved):

apupier (approved):

PR #23194, opened 2026-05-13
CAMEL-22640: Tighten @Nullable annotations where non-null is enforced

gnodet:
[CAMEL-22640](https://issues.apache.org/jira/browse/CAMEL-22640)

## Summary

Follow-up to #23021 — removes `@Nullable` from parameters that are already enforced as non-null or are never null in practice. The previous PR documented *current reality*; this PR tightens the contract where *intentionality* is clear.

## Changes

- **`ResolveEndpointFailedException.uri`**: Remove `@Nullable` from field, constructors, and getter — zero callers across the entire codebase pass null. Add `requireNonNull` enforcement.
- **`LifecycleStrategy.onThreadPoolAdd()` `id` param**: Remove `@Nullable` — `BaseExecutorServiceManager.onThreadPoolCreated()` always computes a non-null id (with `StringHelper.notEmpty()` enforcement). The original Javadoc "(can be null in special cases)" was accurate when the method was introduced in CAMEL-3437 (the `else` branch left `id` as `null` when `source` was not a `ProcessorDefinition` or `String`), but the caller was later fixed to always generate an id using a `source.getClass().getSimpleName() + hashcode` fallback.

## Not changed

Exception `transient` fields (`InvalidPayloadException.type`, `ExpectedBodyTypeException.exchange/expectedBodyType`) are kept `@Nullable` — although constructors enforce non-null via `requireNonNull`, the `transient` keyword means these fields will be null after Java serialization/deserialization.

## Verification

- `mvn install -B -pl core/camel-api -DskipTests -am` — BUILD SUCCESS
- `mvn install -B -pl core/camel-core -DskipTests` — BUILD SUCCESS (downstream compilation)
- `mvn formatter:format impsort:sort -B -pl core/camel-api` — no formatting changes


davsclaus (approved):

core/camel-api/src/main/java/org/apache/camel/ResolveEndpointFailedException.java
```diff
@@ -18,35 +18,33 @@
...
  */
 public class ResolveEndpointFailedException extends RuntimeCamelException {
 
-    private final @Nullable String uri;
+    private final String uri;
 
-    public ResolveEndpointFailedException(@Nullable String uri, Throwable cause) {
-        super("Failed to resolve endpoint: " + sanitizeUri(uri) + " due to: "
+    public ResolveEndpointFailedException(String uri, Throwable cause) {
+        super("Failed to resolve endpoint: " + sanitizeUri(Objects.requireNonNull(uri, "uri")) + " due to: "
```
apupier: the Objects.requireNonNull shoudl provide a meaningful message or it is uselles and just blur the codebase

core/camel-api/src/main/java/org/apache/camel/ResolveEndpointFailedException.java
```diff
@@ -18,35 +18,33 @@
...
               + Objects.requireNonNull(cause, "cause").getMessage(), cause);
         this.uri = sanitizeUri(uri);
     }
 
-    public ResolveEndpointFailedException(@Nullable String uri, String message) {
-        super("Failed to resolve endpoint: " + sanitizeUri(uri) + " due to: "
+    public ResolveEndpointFailedException(String uri, String message) {
+        super("Failed to resolve endpoint: " + sanitizeUri(Objects.requireNonNull(uri, "uri")) + " due to: "
```
apupier: the Objects.requireNonNull shoudl provide a meaningful message or it is uselles and just blur the codebase

core/camel-api/src/main/java/org/apache/camel/ResolveEndpointFailedException.java
gnodet: That's how the JDK handles it, and it's consistent with the rest of the codebase (not completely, but mostly). `NullPointerException("uri must not be null")` doesn't add information over `NullPointerException("uri")` — the "must not be null" part is already conveyed by the exception type itself. The parameter name is the useful signal; the rest is noise.

gnodet: If we could come up with a real value added message, that would be different, but here, we're just checking an argument, we don't have any information to provide.

davsclaus: yeah this is fine and how other frameworks also would do it - also those null values are rarely happening and if so its during development.

davsclaus:
LGTM

PR #23195, opened 2026-05-13
Add @since Javadoc tags to camel-api public types and methods

gnodet:
## Summary

- Add 317 `@since` Javadoc tags to `core/camel-api` tracking when public API elements were introduced (Camel 3.0 through 4.21)
- 266 class-level tags on types (interfaces, classes, enums, annotations) introduced in 3.0+, covering all subpackages (`health/`, `cluster/`, `console/`, `vault/`, `resume/`, `clock/`, `spi/`, etc.)
- 51 method-level tags on methods added to pre-existing interfaces (`Exchange`, `CamelContext`, `Message`)
- Update `AGENTS.md` with convention requiring `@since` tags on all new public API additions going forward

Version provenance was determined using `git log --diff-filter=A` and `git tag --contains` to map each file/method to the release that first included it. Pre-3.0 files (moved during the CAMEL-13040 module reorganization) were excluded.

## Test plan

- [x] `mvn install -B -pl core/camel-api -DskipTests` passes
- [x] `mvn formatter:format impsort:sort -B -pl core/camel-api` produces no changes
- [x] Spot-checked tags against git history for key files (Exchange, CamelContext, VariableAware, HealthCheck, vault configs)


davsclaus:
there are merge conflicts

davsclaus (approved):

PR #23263, opened 2026-05-17
CAMEL-22894: Refactor `SimpleFunctionExpression`: introduce function-factory dispatcher (part 1)

ammachado:
# Description

First incremental step toward [CAMEL-22894](https://issues.apache.org/jira/browse/CAMEL-22894): break up the 4,500-line `SimpleFunctionExpression` that today carries both runtime Simple language and compile-time CSimple code-generation responsibilities, with no behavior change.

This PR establishes the pattern and ports the first batch of small built-in families. Larger families (`header`, `exchangeProperty`, the rest) will follow in subsequent PRs once the dispatcher pattern lands.

**Three commits, each independently reviewable:**

1. **`fc43e0d` — Deprecate `SimpleLanguageFunctionFactory#createCode`.** The `createCode(...)` method on the public `SimpleLanguageFunctionFactory` SPI (`@since 4.10`) exists solely to support csimple, which is already `@Deprecated(since = \"4.19\")` and slated for removal in 5.0. Marked the method `@Deprecated(since = \"4.21\")` and made it a `default` returning `null`, so implementations that don't care about csimple no longer have to override it. The three in-tree implementations (`camel-attachments`, `camel-base64`, `camel-jsoup`) keep their existing overrides and pick up the deprecation marker via inheritance.

2. **`4beafc9` — Centralize external-component function dispatch in `SimpleFunctionDispatcher`.** `SimpleFunctionExpression` previously had six near-identical methods plus six gating predicate blocks dedicated to dispatching to the three external `SimpleLanguageFunctionFactory` implementations (`camel-attachments`, `camel-base64`, `camel-jsoup`). Moved the dispatch into a new internal `SimpleFunctionDispatcher` that owns the list of known component factories and their gating predicates, exposing `tryCreate(...)` / `tryCreateCode(...)` entry points. Behaviour preserved exactly: same factories, same order, same gating (so a missing component still surfaces the helpful \"add this JAR\" error via `ResolverHelper.resolveMandatoryBootstrapService`), and `camel-jsoup` remains excluded from the csimple `createCode` path because its `createCode` throws `UnsupportedOperationException`.

3. **`63d326e` — Extract `random`, `skip`, `collate`, `join` into `SimpleLanguageFunctionFactory` classes.** Moved four self-contained Simple functions out of `SimpleFunctionExpression` into their own factory implementations under `org.apache.camel.language.simple.functions`, registered as built-ins on `SimpleFunctionDispatcher`. The dispatcher consults built-ins before external component factories, preserving today's priority. Both `createFunction` (Simple) and `createCode` (csimple) live side by side in each class, instead of being ~2,000 lines apart in `SimpleFunctionExpression`.

**Net structural impact so far:**

- `SimpleFunctionExpression.java`: 4,526 → ~4,255 lines (−270 LOC), with the new dispatcher (~150 LOC) and four small factory classes (~300 LOC combined).
- Adding a new external-component factory is now a one-line edit in `SimpleFunctionDispatcher` instead of a six-block change spread across `SimpleFunctionExpression`.
- Built-in extraction template established for follow-up PRs.

**Out of scope for this PR** (and the broader refactor, see plan locked in JIRA discussion):

- No public-API additions (no new SPI, no `addFunctionFactory(...)` on `SimpleFunctionRegistry`).
- No CSimple-side restructure: no visitor pattern, no type-safe code-generator, no parameterized cross-tests. The current generated csimple Java source is identical before/after.
- No classloader auto-discovery of factories — would not work in Quarkus native; current explicit allowlist is the right shape.

**Verification:**

- `mvnd clean install -Dquickly -Dci.env.name=local` from repo root: ✅ BUILD SUCCESS.
- `SimpleTest` (227), `CSimpleTest` (1), `SimpleCustomFunctionTest` (5), `CSimplePredicateParserTest` (4), `CSimpleExpressionParserTest` (13), `SimpleExpressionParserNodesTest` (2), `SimplePredicateParserNodesTest` (3): all green.
- `camel-attachments` `SimpleAttachmentTest` (10), `camel-base64` `SimpleBase64Test` (2), full `camel-jsoup` test suite (3): all green.
- `mvn formatter:format impsort:sort` clean.

**Review follow-up after the initial proof-of-concept commits:**

The first three commits were intentionally scoped to prove the extraction/dispatcher pattern and get reviewer feedback on the shape before applying it more broadly. Based on that review, this branch now also:

- restores the original built-in evaluation order
- extracts `codeSplitSafe(...)` into `SimpleFunctionHelper`
- fixes the pre-existing `skip(...)` runtime/code inconsistency
- moves the touched function coverage into dedicated factory unit tests

# Target

# Tracking

- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-22894) filed for the change (usually before you start working on it).

---


davsclaus:
We are not keeping csimple until camel 5 in so many years. It will be removed sooner. Its not really in use by end users. And users who use it can easily use regular simple

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+        for (Entry entry : EXPRESSION_ENTRIES) {
+            if (!entry.claims.test(function)) {
+                continue;
+            }
+            SimpleLanguageFunctionFactory factory = resolve(camelContext, entry.jarName);
+            Expression answer = factory.createFunction(camelContext, function, index);
+            if (answer != null) {
+                return answer;
```
davsclaus, line 98: Minor observation: built-in factories (`random`, `skip`, `collate`, `join`) were previously evaluated inside `createSimpleExpressionMisc()` which runs *before* `createSimpleExpressionMath()`. Now they are evaluated here, *after* math. Since these function prefixes don't overlap with math patterns, this is safe — but the PR description's claim of "Behaviour preserved exactly" should note this ordering shift.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/functions/JoinFunctionFactory.java
```diff
@@ -0,0 +1,107 @@
...
+                exp = tokens[2];
+            } else if (tokens.length == 2) {
+                separator = tokens[0];
+                prefix = tokens[1];
+            } else {
+                separator = tokens[0];
+            }
+        }
+        return SimpleExpressionBuilder.joinExpression(exp, separator, prefix);
+    }
```
davsclaus, line 65: Nit for follow-up: this creates a coupling back to the class being decomposed. As more factories are extracted, `codeSplitSafe` is a good candidate to move into a shared utility class.

davsclaus (commented):
## Review — CAMEL-22894: Refactor `SimpleFunctionExpression`: introduce function-factory dispatcher (part 1)

Well-structured refactoring PR with clear scope boundaries. The three-commit structure is easy to follow, and the PR description is exemplary. CI is green and no blocking issues found.

### Observations

**1. (Low) Evaluation order change for `random`, `skip`, `collate`, `join`**

In the original code these four functions are evaluated inside `createSimpleExpressionMisc()` which runs *before* `createSimpleExpressionMath()`. After this PR they are evaluated via the dispatcher *after* math. Since these function names (`random(`, `skip(`, `collate(`, `join(`) have no overlap with math function patterns, this is harmless in practice. But the PR description states "Behaviour preserved exactly" — this is a safe ordering change rather than exact preservation, and worth acknowledging.

**2. (Low) `JoinFunctionFactory` couples back to `SimpleFunctionExpression.codeSplitSafe()`**

For the follow-up PRs, `codeSplitSafe` should move to a shared utility (e.g. `SimpleParserHelper`) as more factories are extracted. Not blocking for this PR.

**3. (Low) No dedicated unit tests for individual factory classes**

The existing `SimpleTest` (227 tests), `CSimpleTest`, etc. cover behavioral correctness. As more factories are extracted in follow-up PRs, per-factory unit tests would help catch regressions faster.

### Verified Clean

- `SkipFunctionFactory` correctly uses `StringHelper.before()` in `createFunction` and `StringHelper.beforeLast()` in `createCode`, matching the original.
- The `@Deprecated` default `createCode` returning `null` is safe — existing overrides in camel-attachments/base64/jsoup continue working.
- `@Deprecated(since = "4.21")` without `forRemoval` is consistent with existing csimple deprecation convention.
- CI green, formatting clean.

davsclaus (approved):

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/functions/SkipFunctionFactory.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public final class SkipFunctionFactory implements SimpleLanguageFunctionFactory {
+
+    @Override
+    public Expression createFunction(CamelContext camelContext, String function, int index) {
+        String remainder = ifStartsWithReturnRemainder("skip(", function);
+        if (remainder == null) {
+            return null;
+        }
+        String values = StringHelper.before(remainder, ")");
```
gnodet: **Bug (pre-existing, but should fix now):** This uses `StringHelper.before(remainder, ")")` while `createCode` below (line 53) uses `StringHelper.beforeLast(remainder, ")")`. The difference matters if the argument ever contains a literal `)` — `before` stops at the first one, `beforeLast` stops at the last.

This inconsistency was faithfully ported from the original code (runtime line 1196 vs code-gen line 3266), but since we're extracting this into its own class, this is the right time to fix it. All other factories consistently use `beforeLast`.

```suggestion
        String values = StringHelper.beforeLast(remainder, ")");
```

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
```diff
@@ -401,33 +400,10 @@ private Expression doCreateSimpleExpression(CamelContext camelContext, String fu
...
-        // html
-        if ("htmlClean".equals(function) || "htmlParse".equals(function) || "htmlDecode".equals(function)
-                || ifStartsWithReturnRemainder("htmlClean", function) != null
-                || ifStartsWithReturnRemainder("htmlParse", function) != null
-                || ifStartsWithReturnRemainder("htmlDecode", function) != null) {
-            Expression exp = createSimpleHtml(camelContext, function);
-            if (exp != null) {
-                return exp;
-            }
+        // functions from external components (attachments, base64, html, ...)
```
gnodet, line 407: **Ordering change:** In the original code, `random`, `skip`, `collate`, `join` were evaluated inside `createSimpleExpressionMisc()` (called ~10 lines above, before `createSimpleExpressionMath`). Now they're evaluated here via the dispatcher, which runs *after* math.

In practice this is safe because `random(`, `skip(`, `collate(`, `join(` don't collide with math expression patterns. But the PR description says "behaviour preserved exactly" — this subtle shift in evaluation order should at least be acknowledged.

Alternatively, the dispatcher call for built-ins could be placed before the `createSimpleExpressionMisc` call to truly preserve the original ordering. The same applies to the `doCreateCode` path below.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
```diff
@@ -401,33 +400,10 @@ private Expression doCreateSimpleExpression(CamelContext camelContext, String fu
             return math;
         }
```
gnodet, line 406: Nit: The comment says "external components" but the dispatcher now also handles built-in functions (`random`, `skip`, `collate`, `join`). Consider:

```suggestion
        // functions from built-in factories and external components (attachments, base64, html, ...)
```

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/functions/JoinFunctionFactory.java
```diff
@@ -0,0 +1,107 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.language.simple.functions;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.Expression;
+import org.apache.camel.language.simple.SimpleExpressionBuilder;
+import org.apache.camel.language.simple.ast.SimpleFunctionExpression;
```
gnodet: **Circular dependency:** This import creates a dependency from the extracted factory back to `SimpleFunctionExpression` — the very class we're decomposing. `codeSplitSafe` is a utility method that doesn't belong on `SimpleFunctionExpression`.

Consider moving `codeSplitSafe` to a utility class (e.g., `SimpleParserHelper` or `StringQuoteHelper`) in a preparatory commit. That would make the extracted factories truly standalone and avoid the circular reference:
`SimpleFunctionExpression` → `SimpleFunctionDispatcher` → `JoinFunctionFactory` → `SimpleFunctionExpression`.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.camel.language.simple.functions.JoinFunctionFactory;
+import org.apache.camel.language.simple.functions.RandomFunctionFactory;
+import org.apache.camel.language.simple.functions.SkipFunctionFactory;
+import org.apache.camel.spi.SimpleLanguageFunctionFactory;
+import org.apache.camel.support.ResolverHelper;
+
+import static org.apache.camel.language.simple.ast.SimpleFunctionExpression.ifStartsWithReturnRemainder;
+
+/**
+ * Resolves {@link SimpleLanguageFunctionFactory} implementations shipped by external Camel components (currently
```
gnodet: The Javadoc says this class "resolves `SimpleLanguageFunctionFactory` implementations shipped by external Camel components" — but it now also manages the built-in factories (`BUILT_INS`). Consider updating to reflect both roles, e.g.:

> Dispatches Simple/CSimple function lookup to built-in function factories and to `SimpleLanguageFunctionFactory` implementations shipped by external Camel components (currently camel-attachments, camel-base64, camel-jsoup).

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
```diff
@@ -0,0 +1,148 @@
...
+                || "htmlParse".equals(function)
+                || "htmlDecode".equals(function)
+                || ifStartsWithReturnRemainder("htmlClean", function) != null
+                || ifStartsWithReturnRemainder("htmlParse", function) != null
+                || ifStartsWithReturnRemainder("htmlDecode", function) != null;
+    }
+
+    private record Entry(String jarName, Predicate<String> claims) {
+    }
+}
```
gnodet, line 179: Nit: Records are used sparingly in Camel core. The project guidelines say "do NOT use Records unless already present in the file." Since this is a new file, it's a judgment call — a simple `private static final class Entry` with two fields would be more conventional for this codebase. Not blocking, but other committers may flag it.

gnodet (changes requested):

Good first step toward decomposing the 4,500-line `SimpleFunctionExpression`. The commit structure is clean — deprecation, dispatcher, extraction are each independently reviewable. Two issues to address before merge (see inline comments), plus a few suggestions.

**Summary of findings:**

1. **(blocking) `SkipFunctionFactory`: `before` vs `beforeLast` inconsistency** — `createFunction` uses `StringHelper.before()` while `createCode` uses `StringHelper.beforeLast()`. This is a pre-existing bug faithfully ported from the original code, but since we're touching this code now, we should fix it.

2. **(blocking) Ordering change for extracted built-ins** — `random`, `skip`, `collate`, `join` previously evaluated in `createSimpleExpressionMisc()` (before `createSimpleExpressionMath()`). Now they evaluate via the dispatcher (after math). Safe in practice since these prefixes don't collide with math patterns, but the PR description claims "behaviour preserved exactly" — this should be acknowledged or addressed.

3. **(major) Circular dependency** — `JoinFunctionFactory` calls `SimpleFunctionExpression.codeSplitSafe(...)`, creating a dependency from the extracted factory back into the monolith being decomposed.

4. **(minor) Javadoc/comment mismatches** — The dispatcher Javadoc and inline comments don't reflect that it now handles both built-in and external functions.

ammachado:
The first three commits were intentionally scoped to prove the extraction/dispatcher pattern and get reviewer feedback on the shape before applying it more broadly. Based on that review, this branch now also:

- restores the original built-in evaluation order
- extracts `codeSplitSafe(...)` into `SimpleFunctionHelper`
- fixes the pre-existing `skip(...)` runtime/code inconsistency
- moves the touched function coverage into dedicated factory unit tests

gnodet (approved):
All review feedback has been addressed:

- `SkipFunctionFactory` `before`/`beforeLast` inconsistency fixed
- Evaluation ordering restored (built-ins dispatched before misc/math)
- Circular dependency removed (`codeSplitSafe` moved to `SimpleFunctionHelper`)
- Javadoc and comments updated to reflect both built-in and external dispatch
- `record Entry` replaced with conventional `static final class`
- Dedicated per-factory unit tests added, routed through the full language pipeline

davsclaus (approved):

davsclaus:
thanks part 2 is welcome

PR #23283, opened 2026-05-18
CAMEL-23516: camel-xmpp - Use dedicated HeaderFilterStrategy aligned with sibling components

oscerd:
## Summary

`camel-xmpp` defaulted its `headerFilterStrategy` to a bare
`new DefaultHeaderFilterStrategy()` with no configuration — at
`XmppEndpoint.java:106` (the production path, passed into `XmppBinding`)
and in the `XmppBinding()` no-arg constructor. The base strategy with no
`inFilter` / `inFilterPattern` / `inFilterStartsWith` returns `false` for
every header, so inbound XMPP message properties (`XmppBinding.extractHeadersFrom`,
`applyFilterToExternalHeaders`) are copied onto the Camel `Exchange`
as-is — including `Camel*` / `camel*` / `org.apache.camel.*` keys that
drive component behaviour downstream.

This aligns `camel-xmpp` with the rest of the component catalog
(`camel-kafka`, `camel-mail`, `camel-coap`, `camel-google-pubsub`,
`camel-jms`, `camel-sjms`, and `camel-nats` via CAMEL-23515) which all
ship a dedicated `HeaderFilterStrategy` subclass.

## Changes

- New `XmppHeaderFilterStrategy` following the
  `KafkaHeaderFilterStrategy` / `MailHeaderFilterStrategy` shape
  (`setLowerCase(true)`, in/out `setInFilterStartsWith(CAMEL_FILTER_STARTS_WITH)`).
- `XmppEndpoint` and the `XmppBinding` no-arg constructor now default to
  `new XmppHeaderFilterStrategy()`; the now-unused
  `DefaultHeaderFilterStrategy` import is removed from both
  (`XmppHeaderFilterStrategy` is in the same package).
- Unit test covering inbound/outbound `Camel*` filtering, lowercase
  variants, and pass-through of non-Camel headers.
- 4.21 upgrade-guide entry (`=== camel-xmpp`).

## Test plan

- [x] Module build: `mvn -pl components/camel-xmpp -DskipTests install`
- [x] New unit test: `XmppHeaderFilterStrategyTest` (4 tests, green)
- [ ] CI: full reactor build + existing camel-xmpp suite

## JIRA

https://issues.apache.org/jira/browse/CAMEL-23516

Sibling of CAMEL-23515 (camel-nats, merged to main/4.18.x/4.14.x).
Backports to 4.18.x and 4.14.x planned per the `fixVersions`.

### DIFF
--- a/components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppHeaderFilterStrategy.java
+++ b/components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppHeaderFilterStrategy.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.xmpp;
+
+import org.apache.camel.support.DefaultHeaderFilterStrategy;
+
+public class XmppHeaderFilterStrategy extends DefaultHeaderFilterStrategy {
+
+    public XmppHeaderFilterStrategy() {
+        initialize();
+    }
+
+    protected void initialize() {
+        setLowerCase(true);
+        // filter headers begin with "Camel" or "org.apache.camel"
+        setOutFilterStartsWith(CAMEL_FILTER_STARTS_WITH);
+        setInFilterStartsWith(CAMEL_FILTER_STARTS_WITH);
+    }
+
+}
--- a/components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppBinding.java
+++ b/components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppBinding.java
@@ -22,7 +22,6 @@ import java.util.Set;
 
 import org.apache.camel.Exchange;
 import org.apache.camel.spi.HeaderFilterStrategy;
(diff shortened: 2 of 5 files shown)


components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppHeaderFilterStrategy.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.camel.support.DefaultHeaderFilterStrategy;
+
+public class XmppHeaderFilterStrategy extends DefaultHeaderFilterStrategy {
+
+    public XmppHeaderFilterStrategy() {
+        initialize();
+    }
+
+    protected void initialize() {
```
davsclaus, line 27: why does the default header filter not just do this? Otherwise we end up creating a gazillion custom strategies to filter out camel stuff.

A custom strategy should not be needed in most cases. We have a few related to HTTP/REST.

I guess this is maybe a bit late when we have fixed so many components like this style.

Maybe I am missing something. (not using AI)

oscerd: This could be done for sure, I don't know what would be the impact on tests in general. How do we want to go with this?

davsclaus: Lets create a JIRA to investigate later. We can do this for now since we have done it so many before. Then after all of this. We can check and if we can just update the default, then we can change on main on for future releases.

davsclaus (approved):

components/camel-xmpp/src/main/java/org/apache/camel/component/xmpp/XmppHeaderFilterStrategy.java
oscerd: Logged the follow-up as CAMEL-23543 — "camel-core - Investigate centralizing Camel* inbound header filtering in DefaultHeaderFilterStrategy": https://issues.apache.org/jira/browse/CAMEL-23543

It captures the proposal to filter `Camel`/`camel`/`org.apache.camel.` by default in `DefaultHeaderFilterStrategy`, the relationship to CAMEL-23379 (which set `lowerCase=true` by default but intentionally did not add inbound `Camel*` filtering), the list of components currently shipping their own subclass, and the backward-compat / upgrade-guide / PMC-sign-off constraints for changing the base default on `main` for a future release.

Proceeding with the per-component strategy here for now, consistent with the prior components.

PR #23300, opened 2026-05-18
chore: fix latent NPE in S3 consumer includeS3Object with null contentType

Croway:
## Summary

Fix Optional.of → Optional.ofNullable in AWS2S3Consumer.includeS3Object() to handle null contentType when includeFolders is false

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisResumeStrategyConfiguration.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisResumeStrategyConfiguration.java
@@ -60,7 +60,9 @@ public class KinesisResumeStrategyConfiguration extends ResumeStrategyConfigurat
         @Override
         public KinesisResumeStrategyConfiguration build() {
             KinesisResumeStrategyConfiguration result = new KinesisResumeStrategyConfiguration();
-            result.setResumeCache(resumeCache);
+            if (resumeCache != null) {
+                result.setResumeCache(resumeCache);
+            }
             result.setCacheFillPolicy(fillPolicy);
             return result;
         }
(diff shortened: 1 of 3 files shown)


core/camel-core-processor/src/main/java/org/apache/camel/processor/LogProcessor.java
```diff
@@ -77,7 +77,9 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
                 if (listeners != null && !listeners.isEmpty()) {
                     msg = fireListeners(exchange, msg);
                 }
-                logger.doLog(msg);
+                if (msg != null) {
+                    logger.doLog(msg);
```
apupier: this is changing the initial behaviour. Instead of loggign one line with no content apart from the time/loggername information, there won't be any more log which will be silently ignored.

I proposed https://github.com/apache/camel/pull/23299

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisResumeStrategyConfiguration.java
```diff
@@ -60,7 +60,9 @@ public KinesisResumeStrategyConfigurationBuilder withResumeCache(ResumeCache<?>
         @Override
         public KinesisResumeStrategyConfiguration build() {
             KinesisResumeStrategyConfiguration result = new KinesisResumeStrategyConfiguration();
-            result.setResumeCache(resumeCache);
+            if (resumeCache != null) {
+                result.setResumeCache(resumeCache);
+            }
```
apupier: I think it would be better that the ResumeStartegyConfiguration.setResumeCache accepts null, as it was before.
Currently, we removed the possibility to set back resumecache to a null value

apupier:
relates to regression introduced by https://github.com/apache/camel/pull/23021

components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/consumer/KinesisResumeStrategyConfiguration.java
apupier: created https://github.com/apache/camel/pull/23302

apupier: and when reading code, it is even making less sense to forbid to provide a null value as the attribute in the class is marked as @Nullable

Croway:
good point @apupier , it is not the first time that I need to `if(.. != null) { ..set... }`, and I am not proud of, if it can be handled by reverting some changes, that's better

oscerd (approved):

davsclaus:
I guess this PR is outdated and not needed as the 2 PRs from @apupier resolves this

Croway:
rebased

davsclaus (approved):

PR #23304, opened 2026-05-18
CAMEL-23542: Restore @Nullable annotations removed by the requireNonNull revert

gnodet:
[CAMEL-23542](https://issues.apache.org/jira/browse/CAMEL-23542)

Follow-up to #23289.

## Summary

The revert in #23289 correctly removed the `Objects.requireNonNull()` runtime checks that broke 3 tests, but also removed the `@Nullable` JSpecify annotations. This PR restores them to accurately document the nullability contracts:

- **`ValidationException`**: `exchange`, `message`, and `cause` parameters are nullable — callers pass null exchange in tests, `e.getMessage()` returns null, and callers may hold a cause in a variable that could be null
- **`CamelExchangeException`**: `message` and `cause` parameters are nullable — consistent with `createExceptionMessage()` which already accepts all-`@Nullable` parameters, and `Exception(String, Throwable)` handles null cause
- **`RestConfiguration.setValidationLevels`**: parameter is nullable — consistent with the field and getter which are already `@Nullable`

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/CamelExchangeException.java
+++ b/core/camel-api/src/main/java/org/apache/camel/CamelExchangeException.java
@@ -32,7 +32,7 @@ public class CamelExchangeException extends CamelException {
      * @param message  the detail message
      * @param exchange the exchange that caused the error
      */
-    public CamelExchangeException(String message, @Nullable Exchange exchange) {
+    public CamelExchangeException(@Nullable String message, @Nullable Exchange exchange) {
         super(CamelExchangeException.createExceptionMessage(message, exchange, null));
         this.exchange = exchange;
     }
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (approved):

core/camel-api/src/main/java/org/apache/camel/CamelExchangeException.java
```diff
@@ -42,7 +42,7 @@ public CamelExchangeException(String message, @Nullable Exchange exchange) {
      * @param exchange the exchange that caused the error
      * @param cause    the cause of the failure
      */
-    public CamelExchangeException(String message, @Nullable Exchange exchange, Throwable cause) {
+    public CamelExchangeException(@Nullable String message, @Nullable Exchange exchange, Throwable cause) {
```
apupier: I think that the cause is Nullable too

gnodet: Good point — added `@Nullable` on `Throwable cause` too (both `CamelExchangeException` and `ValidationException`). Java's `Exception(String, Throwable)` and `createExceptionMessage()` both handle null cause correctly, and it avoids forcing callers to branch on whether cause is null to pick the right constructor.

davsclaus:
https://issues.apache.org/jira/browse/CAMEL-23552

davsclaus:
LGTM

PR #23305, opened 2026-05-18
CAMEL-23521: camel-core - XML route dumping is missing target element closing tags

davsclaus:
## Summary

Fixes CAMEL-23521: When XML route dumping is performed via `DefaultDumpRoutesStrategy`, the output contains the opening `<routes>`, `<rests>`, or `<routeTemplates>` tag with its namespace attribute but is missing the corresponding closing tag, producing malformed XML.

**Root cause:** `doDumpXml()` strips the outer container element by matching its opening and closing tags via simple string replacement. The old JAXB/Spring dumper emitted `xmlns="http://camel.apache.org/schema/spring"`, which was explicitly stripped first to produce a bare `<routes>`. The lightweight xml-io dumper (`LwModelToXMLDumper`, the current default) uses `xmlns="http://camel.apache.org/schema/xml-io"` instead. Because only the Spring namespace was handled, the regex `<routes>` never matched the opening tag (which still had the xml-io namespace attribute), but `</routes>` (no attributes) was correctly matched and removed — leaving an orphaned opening tag.

**Fix:** Replace the exact `<routes>` string match with a regex `<routes(?:\s[^>]*)?>` that matches the opening tag with or without any namespace/xmlns attributes. This handles both dumpers (xml-io and Spring/JAXB) and any future dumpers regardless of their namespace conventions.

## Changes

- `core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultDumpRoutesStrategy.java` — fix in `doDumpXml()`: use regex to match outer container opening tag with optional attributes
- `core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java` — new test verifying that dumped XML has correct structure (outer containers stripped, individual elements properly closed, `<camel>` wrapper present)

## Test plan

- [x] New test `DefaultDumpRoutesStrategyXmlTest.testDumpXmlHasCorrectStructure` verifies that routes and routeTemplates dump files are well-structured (outer container stripped, individual elements present and closed, `<camel>` wrapper intact)
- [x] All existing `DumpModel*` tests pass
- [x] Full root build (`mvn clean install -DskipTests`) passes

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java
@@ -25,9 +25,7 @@ import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.util.IOHelper;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class DefaultDumpRoutesStrategyXmlTest extends ContextTestSupport {
 
@@ -50,7 +48,7 @@ public class DefaultDumpRoutesStrategyXmlTest extends ContextTestSupport {
     public void testDumpXmlHasCorrectStructure() throws Exception {
         File dir = testDirectory().toFile();
         File[] files = dir.listFiles();
-        assertNotNull(files, "Expected dump files in " + dir);
+        assertThat(files).as("Expected dump files in %s", dir).isNotNull().isNotEmpty();
 
         StringBuilder all = new StringBuilder();
         for (File f : files) {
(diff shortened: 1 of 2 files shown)


oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // outer container tags must be stripped entirely (both opening and closing) - CAMEL-23521
+        assertFalse(xml.contains("<routes"), "Outer <routes> wrapper must be stripped");
+        assertFalse(xml.contains("<routeTemplates"), "Outer <routeTemplates> wrapper must be stripped");
+
+        // individual route elements must be present and properly closed
+        assertTrue(xml.contains("<route "), "Expected individual <route> elements");
+        assertTrue(xml.contains("</route>"), "Expected </route> closing tag");
+        assertTrue(xml.contains("<routeTemplate "), "Expected individual <routeTemplate> elements");
+        assertTrue(xml.contains("</routeTemplate>"), "Expected </routeTemplate> closing tag");
```
apupier: using AssertJ/XMLUnit,  it would help to have better readability when the test is failing. For instance to show what is current content of `xml`

davsclaus: Done — switched all assertions to AssertJ `assertThat(...).as("...\nActual xml:\n%s", xml).contains(...)` / `doesNotContain(...)`. When a check fails, the full dumped XML is now included in the failure message, which makes debugging much easier. Committed in d6c7492b4a42.

core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java
```diff
@@ -0,0 +1,93 @@
...
+                .doesNotContain("<routes");
+        assertThat(xml).as("Outer <routeTemplates> wrapper must be stripped\nActual xml:\n%s", xml)
+                .doesNotContain("<routeTemplates");
+
+        // individual route elements must be present and properly closed
+        assertThat(xml).as("Expected individual <route> elements\nActual xml:\n%s", xml)
+                .contains("<route ", "</route>");
+        assertThat(xml).as("Expected individual <routeTemplate> elements\nActual xml:\n%s", xml)
+                .contains("<routeTemplate ", "</routeTemplate>");
+    }
```
apupier: the assertion could be chained so that it gives an overview of all the part which are failing in a single time

apupier (approved):

core/camel-core/src/test/java/org/apache/camel/impl/DefaultDumpRoutesStrategyXmlTest.java
davsclaus: Done — chained all assertions into a single `assertThat(xml).as("...", xml).contains(...).doesNotContain(...).contains(...)` call. The full XML context is now shown once in the `as()` description and all conditions are visible together in one invocation. Committed in 37b7454.

PR #23307, opened 2026-05-18
CAMEL-23479: Fix camel.main.virtualThreadsEnabled not working

davsclaus:
## Summary

Fixes two distinct bugs where `camel.main.virtualThreadsEnabled=true` failed to enable virtual threads:

- **Race condition (properties file path)**: `autoconfigure()` set the system property `camel.threads.virtual.enabled` *after* `camelContext.build()` had already triggered `ThreadType.current()`, which uses a permanent double-checked-locking cache. Once cached as `PLATFORM`, the value was never re-evaluated.
- **Programmatic path always broken**: `main.configure().withVirtualThreadsEnabled(true)` set `mainConfigurationProperties.virtualThreadsEnabled=true` but `autoConfigurationSingleOption` only reads external sources — the system property was never set, so `ThreadType.current()` always returned `PLATFORM`.

**Fix:**

1. Add `configureVirtualThreadsEarly()` to `BaseMainSupport`, called *before* `camelContext.build()`. It checks all config sources in priority order: programmatic → `initialProperties` → `PropertiesComponent` (application.properties) → ENV vars → JVM system props.
2. Add `ThreadType.enable()` in `camel-util` to directly set the cached value to `VIRTUAL`, overriding any value that may have been cached earlier in the bootstrap sequence.
3. New test class `MainVirtualThreadsTest` covering both previously-broken paths.

## Test plan

- [ ] `MainVirtualThreadsTest.testProgrammaticVirtualThreadsEnabled` — verifies `main.configure().withVirtualThreadsEnabled(true)` correctly sets `ThreadType.VIRTUAL`
- [ ] `MainVirtualThreadsTest.testInitialPropertyVirtualThreadsEnabled` — verifies `main.addInitialProperty("camel.main.virtualThreadsEnabled", "true")` correctly sets `ThreadType.VIRTUAL`
- [ ] Both tests pass in isolation and as part of the full `camel-main` test suite

Relates to: https://issues.apache.org/jira/browse/CAMEL-23479

### DIFF
--- a/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
+++ b/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
@@ -119,6 +119,7 @@ import org.apache.camel.util.OrderedProperties;
 import org.apache.camel.util.SecurityUtils;
 import org.apache.camel.util.SecurityViolation;
 import org.apache.camel.util.StringHelper;
+import org.apache.camel.util.concurrent.ThreadType;
 import org.apache.camel.vault.VaultConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -958,6 +1011,8 @@ protected void postProcessCamelContext(CamelContext camelContext) throws Excepti
         configureRoutesLoader(camelContext);
         // configure custom main listeners
         configureMainListener(camelContext);
+        // configure virtual threads early before build() to avoid ThreadType DCL race
+        configureVirtualThreadsEarly(camelContext);
```
Croway, line 1006: is https://github.com/davsclaus/camel/blob/e93773b5b3e3cbd992c5170b905614b738160a96/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java#L593 still needed? or can be removed in favour of `configureVirtualThreadsEarly` ?

davsclaus: Good catch! Yes, the autoConfigurationSingleOption call for virtualThreadsEnabled is now redundant — configureVirtualThreadsEarly() covers all the same configuration sources and runs at the right moment (before camelContext.build()). Removed it in the follow-up commit.

Croway (approved):

PR #23327, opened 2026-05-19
CAMEL-23499 - Disable again several FileToFtps* tests

apupier:
Several FTP tests were re-enabled recently with
https://github.com/apache/camel/pull/23078 but unfortunately several are broken. They always fail on CI or when launching from Maven locally (at least for me)
Also note that some of them are disabled in case the JVM does not support specific algorithm, which means that they are not raising an error in this case.

These tests are passing if I launch them from my IDE.

# Description

# Target

# Tracking


davsclaus (approved):

apupier:
note that it will be nice to unskip all the skipped tests one day but better to do it without impacting the main branch

PR #23331, opened 2026-05-19
chore: add --enable-native-access=ALL-UNNAMED to CamelJBang launcher

gnodet:
## Summary

- Add `--enable-native-access=ALL-UNNAMED` JBang directive to both CamelJBang launcher copies to suppress JDK 24+ restricted method warnings (JEP 472).
- JLine's FFM and JNI terminal providers call restricted methods like `MemorySegment.reinterpret` and `Linker.downcallHandle`. Without this flag, JDK 24+ prints warnings and JDK 26+ will block these calls with `IllegalCallerException`.
- Since JBang runs everything on the classpath (unnamed module), `ALL-UNNAMED` is the correct value.

## References

- [JEP 472](https://openjdk.org/jeps/472)
- [JLine troubleshooting: JDK 24 restricted method warning](https://jline.org/docs/troubleshooting/#jdk-24-restricted-method-warning)

## Test plan

- [ ] Verify CamelJBang launches without restricted method warnings on JDK 24+
- [ ] Verify no behavioral change on JDK 21-23

### DIFF
--- a/dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java
+++ b/dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java
@@ -20,6 +20,7 @@
 //JAVA 21+
 //REPOS central=https://repo1.maven.org/maven2,apache-snapshot=https://repository.apache.org/content/groups/snapshots/
 //JAVA_OPTIONS -Dcamel.jbang.quarkusVersion=3.33.1.1
+//JAVA_OPTIONS --enable-native-access=ALL-UNNAMED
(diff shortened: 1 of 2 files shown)


davsclaus:
oh for some reason we have a copy of the same file in
https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-main/src/main/jbang/main/CamelJBang.java

So both needs updates.

davsclaus (approved):

oscerd (approved):

PR #23334, opened 2026-05-19
CAMEL-23548: camel-aws2-s3 - allow Simple expressions in destinationB…

GuntherSchrijvers:
# Description

The `camel-aws2-s3` consumer supports a `moveAfterRead` option that copies consumed objects to a destination bucket. The `destinationBucketPrefix` and `destinationBucketSuffix` options allow customising the destination key, but previously only accepted literal strings.

This PR adds Simple expression language support to both options, following the same pattern already used by `AWS2S3Utils.determineKey()` and `AWS2S3Utils.determineBucketName()`: `LanguageSupport.hasSimpleFunction()` detects whether the value contains a `${...}` expression; if so, the `"simple"` language is resolved from the Camel context and evaluated against the current exchange.

Literal strings and `null`/empty values continue to behave exactly as before, so existing routes using `RAW(somePrefix)` are unaffected.

A practical use case is organising moved objects by date:
```
aws2-s3://mybucket?moveAfterRead=true&destinationBucket=archive
  &destinationBucketPrefix=RAW(${date:now:yyyy/MM/dd}/)
```

Changes:
- `AWS2S3Utils`: two new static helpers `evaluateDestinationBucketPrefix` / `evaluateDestinationBucketSuffix`
- `AWS2S3Consumer.processCommit()`: uses the new helpers instead of plain string concatenation
- `AWS2S3UtilsTest`: 6 new unit tests (literal pass-through, Simple expression evaluation, null/empty for prefix and suffix)
- `aws2-s3-component.adoc`: documents expression support with a static and a dynamic date example

# Target

# Tracking

- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-23548) filed for the change.

### DIFF
--- a/components/camel-aws/camel-aws2-s3/src/main/docs/aws2-s3-component.adoc
+++ b/components/camel-aws/camel-aws2-s3/src/main/docs/aws2-s3-component.adoc
@@ -1128,19 +1128,35 @@ This will require specifying the destinationBucket option. As example:
 
 In this case, the objects consumed will be moved to _myothercamelbucket_ bucket and deleted from the original one (because of deleteAfterRead set to true as default).
 
-You have also the possibility of using a key prefix/suffix while moving the file to a different bucket. The options are destinationBucketPrefix and destinationBucketSuffix.
+You have also the possibility of using a key prefix/suffix while moving the file to a different bucket.
+The options are `destinationBucketPrefix` and `destinationBucketSuffix`.
 
-Taking the above example, you could do something like:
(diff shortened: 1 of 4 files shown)


oscerd (approved):

GuntherSchrijvers:
Sorry for the force pushes: I accidentally used the wrong account to commit. It is solved now and I will not be doing any more.

davsclaus (approved):

PR #22192, opened 2026-03-23
CAMEL-21960: camel-jbang - Auto-detect application.properties in run and export

gnodet:
JIRA: [CAMEL-21960](https://issues.apache.org/jira/browse/CAMEL-21960)

## Summary

- Auto-detect `application.properties` in the base directory and include it in the files list for both `camel run` and `camel export` commands
- Previously, running `camel run foo.camel.yaml` would load `application.properties` for configuration but not include it as a file — users had to explicitly add it or use `camel run *`
- The profile-specific `application-{profile}.properties` was already auto-detected for `run`; this adds the same for the base `application.properties`, and adds both to `export`
- Fix pre-existing `DependencyUpdateTest` failure for `main` runtime: Export and DependencyUpdate use different dependency discovery paths (runtime vs catalog), causing false mismatches — exclude `main` from export-based tests
- Document the new auto-detection behavior in the 4.21 upgrade guide and camel-jbang docs

## Test plan

- [x] Existing tests pass (311 tests, 0 failures)
- [x] `DependencyUpdateTest` fixed for incremental CI builds
- [ ] Manual: `camel run foo.camel.yaml` with `application.properties` in the same directory should auto-include it
- [ ] Manual: `camel export` with explicit files should auto-include `application.properties`
- [ ] Manual: `camel run foo.camel.yaml application.properties` should not duplicate the file
- [ ] Manual: `camel run --empty` should not auto-include `application.properties`

PR #23348, opened 2026-05-20
chore: clarify documentation requirements in AGENTS.md

gnodet:
## Summary

- Expand the documentation bullet in the Code Quality section to explicitly call out the upgrade guide and component/command docs
- Ensures behavioral changes (new defaults, auto-detection, removed options, changed headers) are not shipped without user-facing documentation

This was prompted by a review on #22192 where upgrade guide and jbang doc updates were requested after the initial implementation.

## Test plan

- [x] Documentation-only change, no code impact

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,6 +78,8 @@ When creating a PR, **always identify and request reviews** from the most releva
   command or component documentation page. This includes: changed defaults, new auto-detection
   behavior, removed or renamed options, changed header names or values, API/SPI signature changes,
   removed or deprecated components, migrated libraries, and renamed documentation pages.
+  For backported changes, the upgrade guide entry must be added on the `main` branch (not on the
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

oscerd (commented):
I guess we should also state that for backport operation the migration guide to be upgraded is always on main branch.

davsclaus:
> I guess we should also state that for backport operation the migration guide to be upgraded is always on main branch.

yes as the website is using the main branch only for the migration guides that are published. So any changes in LTS branches is not visible and not really in use

gnodet:

@oscerd Good point — added a note that for backported changes, the upgrade guide entry must be added on `main` (not on the maintenance branch). See d0b95fa7.

oscerd (approved):

PR #23363, opened 2026-05-20
Use newest Maven and Java 21 versions in .sdkmanrc and Maven Wrapper

ppalaga:


Croway:
@ppalaga there is a module that requires java 25, would it be possible to update mvnw with java 25? anyway, camel is a multjar project with jdk17 baseline

davsclaus:
I think you mean sdkman?

Croway:
uh, yeah, sdkman https://github.com/apache/camel/pull/23363/changes#diff-909bca673d9d35e9c0672dd22a0dd6bbe1d6dcbe94ccc9221d26021badf1171dR3

davsclaus:
yeah it can be 25 imho. I use that too sometimes, but tend to be on 21 for my local development

davsclaus (approved):

davsclaus:
When releasing we need to use 25

apupier (approved):
agree that we could provide Java 25 for sdkman as it is (will be?) the version required to do the release

ppalaga:
1f35cdb:

* Set java version to 25-tem in .sdkmanrc

davsclaus (approved):

PR #23378, opened 2026-05-20
CAMEL-23571: Ensure GooglePubsubComponent does not leave orphan channels

driseley:
# Description

When using the Google PubSub emulator, orphaned channels are left open by the GooglePubsubComponent that are cleaned up via garbage collection.  This PR resolves this by using an InstantiatingGrpcChannelProvider instead of a FixedTransportChannelProvider and adds assertions to the tests to show that the ManagedChannelOrphanWrapper is not performing cleanup by checking the log output.  See CAMEL-23571

# Target

# Tracking

### DIFF
--- a/components/camel-google/camel-google-pubsub/src/test/java/org/apache/camel/component/google/pubsub/PubsubTestSupport.java
+++ b/components/camel-google/camel-google-pubsub/src/test/java/org/apache/camel/component/google/pubsub/PubsubTestSupport.java
@@ -18,13 +18,14 @@ package org.apache.camel.component.google.pubsub;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.List;
 import java.util.Properties;
 import java.util.logging.LogManager;
 
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.NoCredentialsProvider;
-import com.google.api.gax.grpc.GrpcTransportChannel;
-import com.google.api.gax.rpc.FixedTransportChannelProvider;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
@@ -33,7 +34,6 @@ import com.google.pubsub.v1.ProjectSubscriptionName;
 import com.google.pubsub.v1.Subscription;
 import com.google.pubsub.v1.Topic;
 import com.google.pubsub.v1.TopicName;
-import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import org.apache.camel.BindToRegistry;
 import org.apache.camel.CamelContext;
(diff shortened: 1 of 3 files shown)


components/camel-google/camel-google-pubsub/src/test/java/org/apache/camel/component/google/pubsub/PubsubTestSupport.java
```diff
@@ -86,6 +94,24 @@ public Properties loadRegProperties() {
...
 
+    @Override
+    protected void setupResources() throws Exception {
+        ORPHAN_APPENDER.reset();
+        super.setupResources();
+    }
+
+    @Override
+    protected void cleanupResources() throws Exception {
+        super.cleanupResources();
```
davsclaus, line 105: `System.runFinalization()` was deprecated for removal in Java 18 (JEP 421) and is a no-op in Java 21+. Since Camel targets Java 17+, this will produce deprecation warnings on newer JDKs.

Consider removing the `System.runFinalization()` call — the `System.gc()` hint alone is sufficient to trigger the `ManagedChannelOrphanWrapper` cleanup (which uses phantom references in newer gRPC versions, not finalizers).

```suggestion
        System.gc();
```

davsclaus (commented):
Thanks for this contribution, David — the fix is clean and well-motivated. Replacing `FixedTransportChannelProvider` with `InstantiatingGrpcChannelProvider` is the right approach to ensure channels are properly lifecycle-managed rather than left for GC finalization.

### Confirmed Issues

**1. (Medium) `System.runFinalization()` is deprecated for removal since Java 18**
See inline comment on `PubsubTestSupport.java`.

**2. (Low) Orphan detection test is inherently non-deterministic**
`System.gc()` is only a hint — the JVM may choose not to collect orphaned channels. If GC doesn't finalize the objects during the test, the assertion passes vacuously (no orphan messages = test passes), so it won't catch regressions. This is a known limitation of the approach rather than a bug — the test can prove orphans exist but cannot prove they don't.

### Question

The `endpoint` field is documented as "Endpoint to use with local Pub/Sub emulator." The old code used `ManagedChannelBuilder.forTarget()`, which accepts various gRPC target formats (e.g., `dns:///host:port`). `InstantiatingGrpcChannelProvider.setEndpoint()` expects strictly `host:port`. Is this a potential compatibility concern for users who may have been using non-standard target formats?

### Positive Observations

- The core fix is clean — replacing 4 identical blocks with the proper lifecycle-managed provider
- Good test hygiene: adds a custom Log4j appender to detect orphans rather than just trusting the fix
- Test support class is also fixed, not just production code
- Single focused commit with proper JIRA reference
- CI passes

**Note:** This review covers project rules and conventions. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

components/camel-google/camel-google-pubsub/src/test/java/org/apache/camel/component/google/pubsub/PubsubTestSupport.java
driseley: Thanks Claus - I have removed the `System.runFinalization()` call and force pushed

oscerd (approved):

Croway (approved):

davsclaus (approved):

PR #23420, opened 2026-05-21
CAMEL-23564: Document how to access Baggage in custom processor via traceProcessors

Croway:


davsclaus (approved):

squakez (commented):
Any reason why to add this info? That's how it normally works, so, either it is added via header or via any other mean, it's the way the baggage can be accessed, but it is not specific to the header setting.

squakez (approved):

PR #23443, opened 2026-05-21
CAMEL-23491: Add known 3rd party plugin catalog to camel-jbang

davsclaus:
## Summary
- Add a human-curated catalog of known 3rd party plugins (`known-plugins.json`) so users can install them by name without specifying GAV coordinates (e.g., `camel plugin add forage`)
- Add `vendor` field to `PluginType` enum to distinguish ASF vs Community plugins
- `camel plugin get --all` now shows three sections: active, supported (ASF), and known 3rd party (Community)
- Users can pin a specific version with `--version` flag; defaults to `LATEST` for 3rd party plugins
- Initial catalog includes: forage, camel-kit

## Test plan
- [x] All 21 existing plugin tests pass (PluginGetTest, PluginAddTest, PluginDeleteTest, PluginHelperTest)
- [x] Updated PluginGetTest assertions for new VENDOR column
- [ ] Manual: `camel plugin get --all` shows vendor column and 3rd party section
- [ ] Manual: `camel plugin add forage` auto-fills GAV from catalog
- [ ] Manual: `camel plugin add forage --version=1.2.3` pins specific version

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -585,7 +585,7 @@ public final class PluginHelper {
                 String text = new String(is.readAllBytes(), java.nio.charset.StandardCharsets.UTF_8);
                 JsonArray arr = (JsonArray) Jsoner.deserialize(text);
                 List<JsonObject> result = new ArrayList<>(arr.size());
-                for (Object o : arr) {
+                String text = new String(is.readAllBytes(), StandardCharsets.UTF_8);
                     if (o instanceof JsonObject jo) {
                         result.add(jo);
                     }
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -50,6 +50,7 @@ import org.apache.camel.spi.FactoryFinder;
 import org.apache.camel.support.ObjectHelper;
 import org.apache.camel.tooling.maven.MavenGav;
 import org.apache.camel.util.IOHelper;
+import org.apache.camel.util.json.JsonArray;
 import org.apache.camel.util.json.JsonObject;
 import org.apache.camel.util.json.Jsoner;
 import picocli.CommandLine;
@@ -573,6 +574,37 @@ public final class PluginHelper {
         } catch (IOException e) {
             throw new RuntimeCamelException("Failed to save plugin configuration", e);
         }
+    }
+
+    /**
+     * Loads the list of known 3rd party plugins from the classpath resource known-plugins.json.
+     */
+    public static List<JsonObject> loadKnownPlugins() {
+        try (InputStream is = PluginHelper.class.getClassLoader().getResourceAsStream("known-plugins.json")) {
+            if (is != null) {
+                String text = new String(is.readAllBytes(), java.nio.charset.StandardCharsets.UTF_8);
+                JsonArray arr = (JsonArray) Jsoner.deserialize(text);
+                List<JsonObject> result = new ArrayList<>(arr.size());
+                for (Object o : arr) {
+                    if (o instanceof JsonObject jo) {
+                        result.add(jo);
+                    }
+                }
+                return result;
+            }
+        } catch (Exception e) {
+            // ignore
+        }
+        return List.of();
+    }
+
+    /**
+     * Finds a known 3rd party plugin by name (case-insensitive).
+     */
+    public static Optional<JsonObject> findKnownPlugin(String name) {
+        return loadKnownPlugins().stream()
+                .filter(p -> name.equalsIgnoreCase(p.getString("name")))
+                .findFirst();
     }
 
     public static void enable(PluginType pluginType) {
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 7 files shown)


davsclaus:
@Croway do you have a better forage jbang plugin description we can use
@luigidemasi the same for you for camel kit

Croway:
for Forage, something like `Configure components via opinionated bean factories`

dsl/camel-jbang/camel-jbang-core/src/main/resources/known-plugins.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  {
+    "name": "camel-kit",
+    "command": "kit",
+    "description": "Design Apache Camel Integrations with AI",
+    "firstVersion": "4.10.0",
+    "groupId": "io.github.luigidemasi",
+    "artifactId": "camel-kit-jbang-plugin",
+    "vendor": "Community"
+  }
```
luigidemasi:
```json
  {
      "name": "camel-kit",
      "command": "kit",
      "description": "Turn AI coding agents (Claude, IBM Bob, Gemini and more) into Apache Camel experts — design, migrate, validate Camel integrations",
      "firstVersion": "4.18.0",
      "groupId": "io.github.luigidemasi",
      "artifactId": "camel-jbang-plugin-kit",
      "vendor": "Community"
  }
```

gnodet (commented):
Good feature -- the known-plugin catalog is a nice UX improvement. A couple of items to address:

1. **FQCN in `PluginHelper.loadKnownPlugins()`**: `java.nio.charset.StandardCharsets.UTF_8` is used inline instead of importing `StandardCharsets`. The project convention (and the OpenRewrite CI check) requires using simple class names with an import statement.

2. **Forage description**: @Croway suggested a better description in the comments: _"Configure components via opinionated bean factories"_. The `known-plugins.json` should be updated accordingly.

Otherwise the implementation looks clean, the tests are properly updated, and the documentation is good.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
```diff
@@ -575,6 +576,37 @@ public static void savePluginConfig(JsonObject plugins) {
...
+    /**
+     * Loads the list of known 3rd party plugins from the classpath resource known-plugins.json.
+     */
+    public static List<JsonObject> loadKnownPlugins() {
+        try (InputStream is = PluginHelper.class.getClassLoader().getResourceAsStream("known-plugins.json")) {
+            if (is != null) {
+                String text = new String(is.readAllBytes(), java.nio.charset.StandardCharsets.UTF_8);
+                JsonArray arr = (JsonArray) Jsoner.deserialize(text);
+                List<JsonObject> result = new ArrayList<>(arr.size());
+                for (Object o : arr) {
```
gnodet, line 589: FQCN usage: `java.nio.charset.StandardCharsets.UTF_8` should be replaced with `StandardCharsets.UTF_8` and `import java.nio.charset.StandardCharsets;` added to the import block. The project convention (enforced by OpenRewrite in CI) forbids inline FQCNs.

```suggestion
                String text = new String(is.readAllBytes(), StandardCharsets.UTF_8);
```

dsl/camel-jbang/camel-jbang-core/src/main/resources/known-plugins.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "forage",
+    "command": "forage",
+    "description": "Utilities for working with Forage components",
```
gnodet: @Croway suggested a better description for Forage: _"Configure components via opinionated bean factories"_.

```suggestion
    "description": "Configure components via opinionated bean factories",
```

davsclaus:
> for Forage, something like `Configure components via opinionated bean factories`

@Croway  can we come up with a better description. As an end user I would not really what that means and why I should care. 

Its something about connectors and adapters that adjust to chosen runtimes, and their best practices. The bean factory is more an implementation detail.

gnodet (changes requested):
Follow-up review after commits `08cc321` and `09c358d`.

**Compile error in `PluginHelper.loadKnownPlugins()`**

The web-editor commit (`08cc321`) appears to have mangled the method body. The current code has three issues that will prevent compilation:

1. **Duplicate variable declaration**: `String text` is declared twice — the original FQCN line (`java.nio.charset.StandardCharsets.UTF_8`) was not removed when the replacement line (`StandardCharsets.UTF_8`) was added.

2. **Missing `for` loop**: The `for (Object o : arr)` loop structure was lost — only the body (`if (o instanceof JsonObject jo) { ... }`) and closing braces remain.

3. **Double stream consumption**: `is.readAllBytes()` would be called twice on the same `InputStream`, which yields an empty byte array on the second call.

The corrected method should be:

```java
public static List<JsonObject> loadKnownPlugins() {
    try (InputStream is = PluginHelper.class.getClassLoader().getResourceAsStream("known-plugins.json")) {
        if (is != null) {
            String text = new String(is.readAllBytes(), StandardCharsets.UTF_8);
            JsonArray arr = (JsonArray) Jsoner.deserialize(text);
            List<JsonObject> result = new ArrayList<>(arr.size());
            for (Object o : arr) {
                if (o instanceof JsonObject jo) {
                    result.add(jo);
                }
            }
            return result;
        }
    } catch (Exception e) {
        // ignore
    }
    return List.of();
}
```

**Other observations on the new commits:**

- `known-plugins.json`: The updated forage description ("Configure components via opinionated bean factories") is more accurate — good improvement.
- The rest of the changeset (PluginGet, PluginAdd, PluginType vendor field, docs, tests) looks solid.

Please fix the `loadKnownPlugins()` method — it's the only blocking issue.

davsclaus:
Fixed in e11433274a5e — restored the `for` loop, removed the duplicate `String text` declaration, removed the stale FQCN line, and added the `StandardCharsets` import.

davsclaus:
@Croway 

⏺ Now I get it. Forage isn't really about "bean factories" from a user's perspective — that's an implementation detail. Here's what it actually does for you:

  The core value: You can connect Camel routes to AI models, databases, and vector stores using just properties/env vars — zero Java wiring code. Drop in a dependency, set a few
  properties, and your route can talk to OpenAI, PostgreSQL, Qdrant, etc.

  Here are elevator-pitch-style descriptions:

  1. "Connect to AI models, databases, and vector stores with zero Java code"
  2. "Add AI, JDBC, and vector database support to routes via properties"
  3. "Property-driven AI models, databases, and vector stores for Camel routes"
  4. "Wire AI models, databases, and vector stores using just configuration"
  5. "Drop-in AI, database, and vector store support — configure, don't code"
  6. "Use AI models and databases in routes without writing Java beans"

  The key shift: instead of saying what it is (bean factories), say what it does for you (connect to AI/databases without code). Which direction resonates?

gnodet (approved):
The compile errors in `loadKnownPlugins()` that I flagged in my previous review are now fixed — the duplicate `String text` declaration, missing `for (Object o : arr)` loop, and FQCN `java.nio.charset.StandardCharsets` are all resolved. The method now correctly reads the JSON, parses it, and returns the list of known plugins.

The overall feature looks good: known 3rd party plugin catalog with `known-plugins.json`, VENDOR column in `PluginGet`, name-based install via `PluginAdd`, and updated tests.

LGTM.

davsclaus:
@Croway any thought on a better forage description, if not then we will use the current (a bit confusing imho)

Croway:
> @Croway any thought on a better forage description, if not then we will use the current (a bit confusing imho)

Sorry, thanks for the ping, I'd make it more general, ala `Production ready Apache Camel components configurations via properties`

davsclaus:
description updated

PR #23450, opened 2026-05-22
Made Baggage mutable, to allow for extension.

beskow:
# Description

Made Baggage mutable, i.e. non-private and with a setter.

This allows for extending the Baggage with additional entries, which may be valuable for low-level programmatic baggage management when the generic solution (https://camel.apache.org/components/next/others/opentelemetry2.html#_baggage_customization) isn't sufficient.

For more details, see discussion in https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-23349

# Target

# Tracking


squakez (changes requested):
Sorry, we must decline it. This is against the design of the Span adapter that is meant to be immutable. We could break consistency in async Camel core logic processing if we let the final user to mess it up with the telemetry logic. The telemetry logic MUST be exclusive scope of core and we are limiting that with clear design decisions.

If you need something different of what it is proposed by the framework, I suggest to avoid using it at all and, instead, develop something specific to your used case. Camel is a large and complex framework and there are different ways to manage some specific problem, customization included.

beskow:
Ok, no problem. We can live with the sub-route workaround. Thanks anyway!

PR #23456, opened 2026-05-22
CAMEL-23600: camel-milo - expose the underlying milo OpcUaClient for custom DataType encoding/decoding

oscerd:
## Motivation

JIRA: https://issues.apache.org/jira/browse/CAMEL-23600

A user reported on the mailing list that the `camel-milo` OPC UA client works well for built-in OPC UA data types but cannot handle **custom (server-defined) data types**. Values of such nodes are returned as `ExtensionObject`, which can only be decoded — or, for writes, encoded — with an `EncodingContext` obtained from the underlying Eclipse Milo `OpcUaClient`.

Today the `OpcUaClient` is held privately inside `SubscriptionManager.Connected` and is unreachable from outside the component, so users have no way to obtain the encoding contexts.

## Change

Expose the active milo client through a new accessor:

- `MiloClientConnection.getOpcUaClient()` — delegates through `SubscriptionManager` to the live `OpcUaClient` of the current connection.

This is **purely additive**: no existing public API changes, no changed defaults, no new options. The component still does **not** auto-decode `ExtensionObject` values (it cannot reliably pick the right encoding context); instead it hands users the client so they can do it themselves:

```java
MiloClientEndpoint endpoint = context.getEndpoint("milo-client:opc.tcp://localhost:4334", MiloClientEndpoint.class);
MiloClientConnection connection = endpoint.createConnection();
try {
    OpcUaClient client = connection.getOpcUaClient();
    EncodingContext ctx = client.getDynamicEncodingContext(); // dynamic = server-defined custom types
    UaStructuredType decoded = extensionObject.decode(ctx);    // read path
    ExtensionObject toWrite = ExtensionObject.encode(ctx, myStructure); // write path
} finally {
    endpoint.releaseConnection(connection);
}
```

With the default caching connection manager, `createConnection()` returns the same connection the route uses, so this is the live client.

### Notes / contract
- The connection is established lazily and asynchronously, so `getOpcUaClient()` returns `null` until connected, and again briefly during a reconnect.
- The client is owned and managed by Camel; callers must not alter its lifecycle and should fetch it on demand rather than caching the reference (a reconnect replaces the instance). This is documented on the method and the component page.

## Testing
- New `MiloClientOpcUaClientAccessTest` waits (Awaitility) for the connection, asserts the client and its encoding context are reachable, and performs an `ExtensionObject` encode/decode round-trip of a built-in structured type through the exposed context.
- Full `camel-milo` module suite: 42 tests, 0 failures/errors.
- `mvn formatter:format impsort:sort` applied.

## Docs
- Added a "Custom data types: accessing the underlying milo OpcUaClient" section to `milo-client-component.adoc`. No upgrade-guide entry — the change is additive with no migration impact.

---
_Generated by Claude Code on behalf of Andrea Cosentino (@oscerd)._

### DIFF
--- a/components/camel-milo/src/main/docs/milo-client-component.adoc
+++ b/components/camel-milo/src/main/docs/milo-client-component.adoc
@@ -146,6 +146,40 @@ from("direct:start")
         }).to("mock:test1");
 ----
 
+=== Custom data types: accessing the underlying milo OpcUaClient
+
+Built-in OPC UA data types are read and written transparently. Values whose node has a *custom* (server-defined)
+data type are instead returned as an `org.eclipse.milo.opcua.stack.core.types.builtin.ExtensionObject`, which can only
+be decoded (or, for writes, encoded) with an `EncodingContext` obtained from the underlying milo client.
+
+The component does not decode these values automatically, because it cannot reliably determine which encoding context
+applies. Instead, the active `OpcUaClient` is exposed through `MiloClientConnection.getOpcUaClient()`, so you can obtain
+its encoding contexts and perform the encode/decode yourself:
+
+[source,java]
+----
+MiloClientEndpoint endpoint = context.getEndpoint("milo-client:opc.tcp://localhost:4334", MiloClientEndpoint.class);
+MiloClientConnection connection = endpoint.createConnection();
+try {
+    OpcUaClient client = connection.getOpcUaClient();
+    // dynamic context resolves custom, server-defined types; static context covers built-in/generated types
+    EncodingContext encodingContext = client.getDynamicEncodingContext();
+
+    // decode a value read as ExtensionObject
+    UaStructuredType decoded = extensionObject.decode(encodingContext);
+
+    // or encode a structure before writing it
+    ExtensionObject toWrite = ExtensionObject.encode(encodingContext, myStructure);
+} finally {
+    endpoint.releaseConnection(connection);
+}
+----
+
+NOTE: The connection is established lazily and asynchronously, so `getOpcUaClient()` returns `null` until the client is
+connected, and again briefly while a connection is being re-established. The client is owned and managed by Camel: do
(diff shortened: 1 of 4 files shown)


orpiske (approved):
Reviewed the changes — clean, well-documented, and well-tested additive API. Thread-safety is properly handled, the nullable contract is clearly documented, and the test covers the encode/decode round-trip. LGTM.

_Review by Claude Code on behalf of Otavio Rodolfo Piske (@orpiske)._

gnodet (approved):
Clean, well-scoped additive API. The change is backward-compatible — no existing signatures or defaults are modified.

**What I checked:**

- **Thread safety**: `SubscriptionManager.getOpcUaClient()` correctly acquires the `lock` before reading `this.connected`, consistent with all other methods in the class (`call`, `write`, `readValues`, `browse`). The Javadoc clearly documents the inherently racy nature of the returned reference (reconnect replaces the instance).

- **`checkInit()` call**: `MiloClientConnection.getOpcUaClient()` calls `checkInit()` before delegating to `this.manager`, which is consistent with every other public method in `MiloClientConnection` (`monitorValue`, `writeValue`, `readValues`, `call`, `browse`).

- **Nullable contract**: Both the `SubscriptionManager` and `MiloClientConnection` Javadocs document that the method returns `null` when not connected, and the test correctly handles this via Awaitility polling.

- **Test quality**: Uses Awaitility (no `Thread.sleep`) with an explicit `atMost` timeout. The encode/decode round-trip with `ReadValueId` through `ExtensionObject` exercises the actual use case. Extends `AbstractMiloServerTest` for the embedded OPC UA server, consistent with other tests in the module.

- **Documentation**: The new section in `milo-client-component.adoc` is clear and includes a working code example with proper resource cleanup (`try/finally` with `releaseConnection`). The `NOTE` block about connection lifecycle is well-placed.

- **No upgrade guide needed**: Agreed — this is purely additive with no migration impact.

- **Code style**: No Lombok, no FQCNs, imports are clean.

LGTM.

PR #23479, opened 2026-05-22
chore: Fix commands using System.console() to work inside camel shell

gnodet:
## Summary

Inside the JLine shell (`camel shell`), `System.console()` returns `null`, causing subcommands that read interactive input to silently fail. For example, `debug` prints nothing because its input-reading thread NPEs on `System.console().readLine()` in a background thread.

This PR:
- Moves the active terminal reference to `EnvironmentHelper` (in `core.common`) so all commands can detect and reuse the shell's JLine terminal for input
- Updates `isColorEnabled()` and `isInteractiveTerminal()` to check for the active terminal
- Adds a `readLine()` helper that uses the terminal when inside the shell, falling back to `System.console()`
- Fixes all commands that used `System.console()` directly:
  - **`Debug`** — user input (step, quit, etc.) now works inside shell
  - **`InfraRun`** — "press ENTER to stop" now works inside shell
  - **`Init`** — interactive template picker now works inside shell
  - **`CommandHelper.ReadConsoleTask`** — watch-mode "press enter to stop" now works inside shell
  - **`CommandHelper.confirmOperation`** — interactive confirmation now works inside shell

## Test plan

- [ ] `camel shell` then `debug hello.yaml` — verify step/quit commands work
- [ ] `camel shell` then `init` — verify interactive template picker appears
- [ ] `camel shell` then `infra run kafka` — verify "press ENTER to stop" works
- [ ] `camel debug hello.yaml` (outside shell) — verify still works as before
- [ ] `camel init` (outside shell) — verify interactive picker still works
- [ ] Build: `mvn install -B -pl dsl/camel-jbang/camel-jbang-core -DskipTests`

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/CommandHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/CommandHelper.java
@@ -17,6 +17,7 @@
 package org.apache.camel.dsl.jbang.core.commands;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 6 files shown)


davsclaus (approved):

davsclaus:
what about your other PR is there overlap between these 2 ?

PR #23491, opened 2026-05-24
CAMEL-21122: camel-test - release AvailablePortFinder.Port after each test for non-static @RegisterExtension fields

rkdfx:
# Description

`AvailablePortFinder.Port` leaked its global port-mapping entry whenever used as a non-static `@RegisterExtension` field under the default PER_METHOD lifecycle, because JUnit Jupiter does not fire `AfterAllCallback` in that case. Long camel-jetty runs accumulated stale entries and flaked. `Port` now also implements `BeforeAllCallback` + `AfterEachCallback`: `beforeAll` sets a flag that suppresses the new `afterEach` release, so method-scoped non-static fields self-release per test while static / PER_CLASS behaviour stays the same. Drops the `BaseJettyTest @AfterEach cleanupPorts()` workaround. Mirrored across `camel-test-junit5` and `camel-test-junit6`; documented in `camel-4x-upgrade-guide-4_21.adoc`. Lifecycle covered by a new `AvailablePortFinderLifecycleTest` driven through the JUnit Platform launcher.

Resolves https://issues.apache.org/jira/browse/CAMEL-21122

# Target

# Tracking

### DIFF
--- a/components/camel-jetty/src/test/java/org/apache/camel/component/jetty/BaseJettyTest.java
+++ b/components/camel-jetty/src/test/java/org/apache/camel/component/jetty/BaseJettyTest.java
@@ -23,7 +23,6 @@ import org.apache.camel.CamelContext;
 import org.apache.camel.http.common.HttpHeaderFilterStrategy;
 import org.apache.camel.test.AvailablePortFinder;
 import org.apache.camel.test.junit6.CamelTestSupport;
-import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 public abstract class BaseJettyTest extends CamelTestSupport {
@@ -35,13 +34,6 @@ public abstract class BaseJettyTest extends CamelTestSupport {
 
     @RegisterExtension
     protected AvailablePortFinder.Port port2 = AvailablePortFinder.find();
-
-    // Due to CAMEL-21122 ports are never released. So, force them to be released.
-    @AfterEach
-    void cleanupPorts() {
-        port1.release();
-        port2.release();
-    }
 
     @Override
     protected CamelContext createCamelContext() throws Exception {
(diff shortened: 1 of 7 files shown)


davsclaus (approved):
Clean, well-scoped fix for the `AvailablePortFinder.Port` lifecycle leak under non-static `@RegisterExtension`.

**Positives:**
- The `BeforeAllCallback` + `classScoped` flag approach correctly distinguishes static vs. non-static registration without reflection
- All three lifecycle combinations (PER_METHOD non-static, PER_CLASS non-static, static) are tested through the real JUnit Platform launcher — the right level to catch this bug
- Mirrored identically across camel-test-junit5 and camel-test-junit6
- BaseJettyTest workaround correctly removed
- Upgrade guide entry is clear

- `volatile` on `classScoped` is harmless but likely unnecessary since JUnit lifecycle callbacks have happens-before guarantees
- No double-release path exists: the `classScoped` flag cleanly partitions which callback calls `release()`

LGTM.

gnodet (approved):
Well-designed fix for CAMEL-21122. The approach is clean: use `BeforeAllCallback` as the signal that the `Port` is class-scoped (static `@RegisterExtension`), and release in `afterEach` only when non-class-scoped. This correctly handles all three JUnit lifecycle combinations:

1. **PER_METHOD + non-static field** — each test method gets its own `Port`, released after each method
2. **PER_CLASS + non-static field** — `beforeAll` fires (sets `classScoped`), port shared across methods, released on `afterAll`
3. **Static field** — same as PER_CLASS, `beforeAll` fires, released on `afterAll`

The test using `LauncherFactory` to drive nested test classes through the real JUnit Platform lifecycle is the right way to verify this — calling callbacks by hand would miss the exact lifecycle behavior that caused the leak.

The removal of the `BaseJettyTest.cleanupPorts()` workaround confirms this fixes the root cause. Both `camel-test-junit5` and `camel-test-junit6` copies updated consistently.

LGTM.

PR #23510, opened 2026-05-25
CAMEL-23608: camel-jbang - avoid JLine for terminal width detection

davsclaus:
## Summary

- Replace JLine `TerminalBuilder` with `COLUMNS` env var / `stty size` for terminal width detection in `TerminalWidthHelper`
- Fixes escape sequence leak (DA1 and CPR responses) into shell output after CLI commands like `camel ps`

## Root Cause

`TerminalWidthHelper.getTerminalWidth()` was introduced in CAMEL-23237 (#22219). It creates a JLine `Terminal` with `system(true)` which sends escape sequence queries to the terminal. The terminal responses arrive asynchronously and can leak into the shell after the `Terminal` is closed:

```
  PID   NAME             READY  STATUS   AGE   TOTAL  FAIL  INFLIGHT
 89933  circuit-breaker   1/1   Running  2m1s    119    32         0
^[[?64;1;2;4;6;17;18;21;22;52c^[[2;1R%
```

## Approach

Use `COLUMNS` env var (set by most shells) and `stty size` (POSIX) instead. These detect terminal width without sending escape sequences. Falls back to the default 120 columns if neither is available.

## Test plan

- [ ] Run `camel ps` — output should be clean, no trailing escape sequences
- [ ] Resize terminal, run `camel ps` — table should adapt to new width
- [ ] Pipe output (`camel ps | cat`) — should use default width (120)

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/TerminalWidthHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/TerminalWidthHelper.java
@@ -16,15 +16,16 @@
  */
 package org.apache.camel.dsl.jbang.core.common;
 
-import org.jline.terminal.Terminal;
-import org.jline.terminal.TerminalBuilder;
-
 /**
  * Helper for detecting the terminal width to adapt table and command output.
  *
  * <p>
- * Uses JLine to query the actual terminal size. Falls back to a default width when the terminal size cannot be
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Good fix. Replacing JLine `TerminalBuilder` with `COLUMNS` env var / `stty size` avoids the DA1/CPR escape sequence leak that was polluting shell output after `camel ps` and similar commands. The fallback chain (COLUMNS -> stty -> default 120) is sensible, and the `catch` block handles Windows where `stty` isn't available.

LGTM.

PR #23531, opened 2026-05-26
camel-openai: close OpenAI client on endpoint stop to prevent OkHttp thread leak

Croway:
## Summary

- Call `client.close()` in `OpenAIEndpoint.doStop()` before nulling the reference so that OkHttp's dispatcher executor, connection pool, and cache are properly shut down.
- Without this, OkHttp threads (`OkHttp Dispatcher`, `OkHttp ConnectionPool`) linger after route stop, especially visible under `exec:java` or Camel JBang dev mode where the JVM does not exit.

### DIFF
--- a/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
+++ b/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
@@ -138,6 +138,7 @@ public class OpenAIEndpoint extends DefaultEndpoint {
             returnDirectTools.clear();
         }
         if (client != null) {
+            client.close();
             client = null;
         }
         super.doStop();


components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
```diff
@@ -138,6 +138,7 @@ protected void doStop() throws Exception {
             returnDirectTools.clear();
         }
         if (client != null) {
+            client.close();
```
apupier, line 141: Should we enclose in a try catch as it is done for the mcp clients?

Croway:
```
Closes this client, relinquishing any underlying resources.
This is purposefully not inherited from AutoCloseable because the client is long-lived and usually should not be synchronously closed via try-with-resources.
```

this is from the .close() java doc, should be safe without a try catch

gnodet (approved):
Correct fix — calling `client.close()` before setting `client = null` in `OpenAIEndpoint.doStop()` ensures the HTTP client resources are properly released. Without this, the client would be leaked on endpoint stop.

LGTM.

PR #23532, opened 2026-05-26
CAMEL-23617: Add option to capture message payload size for observation

davsclaus:
## Summary

Adds message size tracking to Camel, capturing body and header sizes per endpoint for both incoming (IN) and outgoing (OUT) directions.

- Introduces `MessageSizeStrategy` SPI and `DefaultMessageSizeStrategy` for computing body/header byte sizes (supports `byte[]`, `String`, `StreamCache`, `WrappedFile`, `File`, `Path`, `Content-Length` header fallback)
- Tracks min/max/mean sizes per endpoint in `RuntimeEndpointRegistry` via exchange events (`ExchangeCreatedEvent` for IN, `ExchangeSendingEvent` for OUT)
- Resets stream cache after computing body size to ensure the body remains re-readable
- Size computation is wrapped in try-catch to gracefully handle any errors (returns -1)
- Exposes size statistics via JMX (`RuntimeEndpointRegistry` MBean), dev console (`EndpointDevConsole`), and CLI (`camel get endpoint`)
- CLI shows `BODY-SIZE`/`HDR-SIZE` columns (mean values) by default when data is available, with `--verbose` for min/max breakdown and `--sort size` support
- Sizes displayed as human-readable (`512 B`, `1.5 KB`, `2.3 MB`) in CLI table, raw bytes in JSON output
- Sets `CamelMessageBodySize` and `CamelMessageHeadersSize` exchange properties for IN direction, usable during routing
- Auto-enabled by the dev profile; requires `messageSizeEnabled=true` and `Extended` statistics level
- This is a global feature — no fine-grained per-route/per-endpoint control
- Streaming bodies (`InputStream`) require stream caching to be enabled
- Includes bundled `message-size` example for Camel JBang/TUI
- Includes documentation (`message-size.adoc`) and JMX docs update

## Test plan

- [x] `ManagedMessageSizeTest` verifies endpoint-level size statistics and exchange properties
- [x] `ManagedMessageSizeStreamCachingTest` verifies stream cached body is readable after size computation
- [x] Validated YAML DSL route via Camel MCP tools
- [x] Manual testing with `camel run` and `camel get endpoint` / `camel get endpoint --verbose` / `camel get endpoint --sort -size`


davsclaus:
1. The message size is not only per route, its also per endpoint and separated by IN and OUT direction.
2. The default code that calculates header size should be in a try .. catch to ignore any issues and return -1 if it fails to do so. The same for the message body.
3. For OUT then we need to reset stream caching afterwards if the body is stream caching, so when writing the OUT response it is re-readable
4. Update doc that streams (InputStream) is only support when using stream-caching
5. Also mention in docs that this is global feature if enabled, you cannot have fine grained control where to turn this on or off.

davsclaus:
Addressed all 5 items:

1. **Javadoc fix** — Updated `DefaultConfigurationProperties` javadoc from "per route" to "per endpoint (for both IN and OUT directions)"
2. **Try-catch** — Wrapped both `computeBodySize()` and `computeHeadersSize()` in try-catch, returning -1 on any exception
3. **Stream cache reset** — Added `StreamCache.reset()` after computing body size in all three event paths (IN, OUT sending, OUT reply)
4. **Streaming docs** — Added note that streaming bodies (`InputStream`) require stream caching to be enabled
5. **Global feature docs** — Added note that message size tracking is global, no per-route/per-endpoint control

Also added `ManagedMessageSizeStreamCachingTest` that sends a `ByteArrayInputStream` body through stream caching and verifies the body is still readable after size computation.

gnodet (approved):
Adds `messagePayloadSize` option to the observation configuration, allowing users to control whether message payload sizes are included in observability data. The bulk of the diff is generated catalog metadata, which is expected. The core logic change is straightforward.

LGTM.

PR #23534, opened 2026-05-26
feat(components): micrometer observability baggages

squakez:
Closes CAMEL-23415

# Description

# Target

# Tracking

### DIFF
--- a/components/camel-micrometer-observability/src/main/docs/micrometer-observability.adoc
+++ b/components/camel-micrometer-observability/src/main/docs/micrometer-observability.adoc
@@ -40,7 +40,7 @@ include::spring-boot:partial$starter.adoc[]
 
 === Spring Boot context propagation
 
-The starter is in charge to autoconfigure the component. Additionally you will need to specify the concrete Propagation implementation by adding the dependency you wish to use (for example, `io.micrometer:micrometer-tracing-bridge-otel`, 'io.micrometer:micrometer-tracing-bridge-brave' or any other technology you wish to use). If none is provided, a "no-op" implementation will be defined as default.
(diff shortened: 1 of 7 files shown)


components/camel-micrometer-observability/src/main/docs/micrometer-observability.adoc
```diff
@@ -173,3 +171,30 @@ public void process(Exchange exchange) throws Exception {
...
 ----
+
+=== Baggage customization
+
+`Baggage` is a way to attach key-value metadata to a request and carry it across service boundaries. In the context of telemetry technologies, baggage travels along with the context (like trace/span), but it's meant for custom data you define, not telemetry internals. Camel allows you to programmatically provide any `Baggage` information via header settings. Whenever the component finds an header defined as `MICROMETER_BAGGAGE_xyz` it will consider it as a baggage variable named `xyz`. For example, in Java DSL:
+
+[source,java]
+----
+                from("direct:start")
+                        .setHeader("MICROMETER_BAGGAGE_myValue", constant("1234"))
```
davsclaus: Hmm wont this cause baggage leaks over protocols as headers are part of the user data. Should this not be exchange properties instead. Those are not sent over the wire. Also a good idiom is to use CamelXXX style as reserved for Camel functionality

davsclaus (changes requested):

gnodet (approved):
Nice addition of baggage support to the Micrometer observation interceptor. The `MICROMETER_BAGGAGE_` header prefix convention is clean, and the `ScopedBaggages` AutoCloseable helper ensures proper lifecycle management of baggage scopes. Removing the old `BAGGAGE_CAMEL_FLAG` in favor of the new approach simplifies the design.

Test coverage in `BaggageSettingTest.java` validates the flow.

LGTM.

components/camel-micrometer-observability/src/main/docs/micrometer-observability.adoc
squakez: I changed to use a property now. Unfortunately we did not catch this requirement in the `camel-opentelemetry2` where we already have implemented the baggage via header instead. I am going to check if we can uniform that development over there as well.

davsclaus (approved):

PR #23535, opened 2026-05-26
CAMEL-23621: filter tool argument headers against declared parameters

Croway:
## Summary
- Filter LLM tool argument field names against the tool's declared parameter schema before setting them as Exchange headers
- Affects camel-langchain4j-tools, camel-langchain4j-agent, and camel-spring-ai-tools
- Undeclared field names are logged at WARN level and skipped
- Fix raw JsonNode values being set as headers in camel-langchain4j-agent — now extracts primitive values

## Behavioral change
Tools defined without parameter.* declarations will now have zero argument headers set from LLM tool calls. Previously, all LLM-provided argument field names were set as Exchange headers without filtering. This is a breaking change for routes that rely on implicit argument passthrough, e.g.: from("langchain4j-tools:test1?tags=user&description=Does not really do anything") Any LLM-provided arguments for such tools will be silently dropped. To restore argument flow, declare the expected parameters explicitly via parameter.<name>=<type>.

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
+++ b/components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
@@ -215,12 +215,25 @@ public class LangChain4jAgentProducer extends DefaultProducer {
                             JsonNode jsonNode = objectMapper.readValue(arguments, JsonNode.class);
                             jsonNode.fieldNames()
                                     .forEachRemaining(name -> {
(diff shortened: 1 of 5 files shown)


oscerd (approved):

gnodet (approved):
Good security hardening for the LangChain4j tool invocation path. Filtering incoming JSON fields against `toolSpecification.parameters().properties().keySet()` is the right approach — it ensures only declared tool parameters are forwarded as headers, preventing injection of Camel-internal headers like `CamelFileName` through LLM-generated tool arguments.

The WARN log for skipped undeclared fields is helpful for debugging without silently swallowing. Test case demonstrating that `CamelFileName` injection is blocked validates the fix well. Upgrade guide entry included.

LGTM.

davsclaus (approved):

PR #23544, opened 2026-05-27
CAMEL-22894: Extract sys/env, date, properties, ref, type, messageAs, pretty/toJson Simple functions into dedicated factories

ammachado:
# Description

Continues the CAMEL-22894 series, which extracts inline Simple language function dispatch from the monolithic `SimpleFunctionExpression` into dedicated `SimpleLanguageFunctionFactory` implementations.

This PR covers six new factory classes:

| Factory | Functions handled |
|---|---|
| `SystemFunctionFactory` | `${sys.name}`, `${sysenv.name}`, `${sysenv:name}`, `${env.name}`, `${env:name}` |
| `PropertiesFunctionFactory` | `${propertiesExist:key}`, `${properties:key}`, `${properties:key:default}`, `${ref:name}` |
| `TypeFunctionFactory` | `${type:fqn}`, `${type:fqn.FIELD}` |
| `DateFunctionFactory` | `${date:command}`, `${date:command:pattern}`, `${date-with-timezone:command:timezone:pattern}` (including `date:millis` code-gen special case) |
| `MessageFunctionFactory` | `${messageAs(type)}`, `${messageAs(type).OGNL}` |
| `OutputFunctionFactory` | `${pretty(exp)}`, `${toJson(exp)}`, `${toPrettyJson(exp)}` |

All factories follow the established pattern: `createFunction()` for runtime evaluation, `createCode()` (deprecated, csimple only) for code generation where applicable. `MessageFunctionFactory` and `OutputFunctionFactory` omit `createCode()` since csimple never supported those functions.

Also removes the two `@Deprecated(forRemoval = true)` methods `tryCreate()` and `tryCreateCode()` from `SimpleFunctionDispatcher`, which were confirmed unused.

`SimpleFunctionExpression` shrinks by ~160 lines after this PR.

## Test changes

Tests for the six factories are moved out of the monolithic `SimpleTest` into dedicated per-factory test classes under `o.a.c.language.simple.functions`. Notable refinements:

- `SystemFunctionFactoryTest` uses JUnit Pioneer's `@SetSystemProperty` (added `junit-pioneer` dependency to `camel-core`) to safely scope the property to the test method.
- `MessageFunctionFactoryTest` documents a subtle semantic: `messageAs(T)` returns `null` at the expression layer when the type conversion fails; the Simple predicate engine then coerces `null` to `false`. The original `SimpleTest` only tested the predicate layer, masking this.
- `OutputFunctionFactoryTest` documents the `XmlPrettyPrinter` behavior where text content directly inside the root element is silently dropped (only emitted when `indent > 1`).

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
@@ -24,13 +24,19 @@ import org.apache.camel.Expression;
 import org.apache.camel.language.simple.functions.BodyFunctionFactory;
 import org.apache.camel.language.simple.functions.CollateFunctionFactory;
 import org.apache.camel.language.simple.functions.CollectionFunctionFactory;
+import org.apache.camel.language.simple.functions.DateFunctionFactory;
 import org.apache.camel.language.simple.functions.HeaderFunctionFactory;
 import org.apache.camel.language.simple.functions.JoinFunctionFactory;
 import org.apache.camel.language.simple.functions.MathFunctionFactory;
+import org.apache.camel.language.simple.functions.MessageFunctionFactory;
 import org.apache.camel.language.simple.functions.MiscFunctionFactory;
+import org.apache.camel.language.simple.functions.OutputFunctionFactory;
+import org.apache.camel.language.simple.functions.PropertiesFunctionFactory;
 import org.apache.camel.language.simple.functions.RandomFunctionFactory;
 import org.apache.camel.language.simple.functions.SkipFunctionFactory;
 import org.apache.camel.language.simple.functions.StringFunctionFactory;
+import org.apache.camel.language.simple.functions.SystemFunctionFactory;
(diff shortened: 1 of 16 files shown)


davsclaus (approved):
Clean extraction of six more Simple function families into dedicated factories, continuing the CAMEL-22894 series. All factories faithfully reproduce the original dispatch logic from `SimpleFunctionExpression`.

Verified:
- `tryCreate()` / `tryCreateCode()` removals are safe — confirmed no external callers
- Dead `parts.length > 2` validation in `propertiesExist:` correctly dropped (split with limit 2 can never exceed 2 elements)
- `TypeFunctionFactory` preserves the `init()` + `cacheExpression()` wrapping
- `OutputFunctionFactory` prefix ordering is correct
- Test coverage is thorough — `MessageFunctionFactoryTest` documents the subtle null-vs-false semantic, `SystemFunctionFactoryTest` uses JUnit Pioneer `@SetSystemProperty` properly

Minor observation (non-blocking): `PropertiesFunctionFactory.createCode()` still has the same dead `parts.length > 2` check (line ~42 in the factory). Not a regression, just a cleanup opportunity.

LGTM.

PR #23546, opened 2026-05-27
CAMEL-23622: Exclude camel.jbang properties from startup summary

davsclaus:
## Summary

- Filter out `camel.jbang.*` properties from the property-placeholder startup summary log
- These are internal JBang properties (e.g. `camel.jbang.readmeFiles`) not meaningful to end users

## Test plan

- [ ] Manual: run `camel run` with a project that has a README.md and verify the summary no longer shows `camel.jbang.readmeFiles`


gnodet (approved):
Clean, minimal change. Filtering out `camel.jbang.*` properties from the startup summary log prevents internal jbang configuration from cluttering the user-visible output. The skip logic is consistent with the existing `templateId` filter pattern.

LGTM.

PR #23547, opened 2026-05-27
CAMEL-23615: camel-tui - Fix unsafe type casts and unguarded parsing

davsclaus:
## Summary

- Fix unsafe `(Boolean)` cast from JSON for `polling` field — use `Boolean.TRUE.equals()` instead
- Fix unsafe `(Boolean)` casts for `done`/`failed` fields in history entry parsing
- Guard `Long.parseLong(ctx.selectedPid)` with try-catch on tab switch to history

## Test plan

- [x] Builds clean
- [ ] Manual: TUI operates normally with consumer tab, history tab, and endpoint views


gnodet (approved):
Good defensive coding fixes for the TUI:

1. Wrapping `Long.parseLong(ctx.selectedPid)` in try-catch for `NumberFormatException` — prevents crash on unexpected PID values.
2. Replacing unsafe `(Boolean) json.get("done")` / `(Boolean) json.get("failed")` casts with `Boolean.TRUE.equals(...)` — avoids `ClassCastException` if the JSON value isn't a Boolean object.
3. Same fix for `cj.get("polling")` in `parseIntegration`.

All three changes follow the same pattern: replacing unsafe assumptions about runtime types with null-safe, cast-safe alternatives.

LGTM.

PR #23567, opened 2026-05-27
CAMEL-23627: fixed response error check in SlackProducer

Xadhoodin:
# Description

Currently a possible Slack erorr is checked and set only after `callback.done(true)` has been called. This makes the route complete before checking whether an exception has been set, leading said exception to bubble up and not get caught by any `onException` policy.

# Target

# Tracking

### DIFF
--- a/components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
+++ b/components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
@@ -95,7 +95,7 @@ public class SlackProducer extends DefaultAsyncProducer {
             callback.done(true);
         }
 
-        return false;
+        return true;
     }
 
     private ChatPostMessageResponse sendLegacySlackMessage(SlackMessage slackMessage) throws IOException, SlackApiException {
@@ -145,7 +145,7 @@ public class SlackProducer extends DefaultAsyncProducer {
             callback.done(true);
         }
 
-        return false;
+        return true;
     }
 
     private Message addEndPointOptions(Message slackMessage) {
--- a/components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
+++ b/components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
@@ -84,15 +84,15 @@ public class SlackProducer extends DefaultAsyncProducer {
                 slackMessage.setText(exchange.getIn().getBody(String.class));
                 response = sendLegacySlackMessage(slackMessage);
             }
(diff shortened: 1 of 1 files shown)


oscerd (approved):

davsclaus (approved):

Croway:
Hi @Xadhoodin , would it be possible to add a test in https://github.com/apache/camel/blob/8ae3d45d36df7b6fe785ffdf0d1070427dd5db31/components/camel-slack/src/test/java/org/apache/camel/component/slack/SlackProducerTest.java#L33 ?

components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
```diff
@@ -84,17 +84,17 @@ private boolean sendMessageByToken(Exchange exchange, AsyncCallback callback) {
...
             return true;
         } finally {
             callback.done(true);
         }
 
-        if (!response.isOk()) {
-            exchange.setException(new CamelExchangeException("Error POSTing to Slack API: " + response.toString(), exchange));
-        }
-
         return false;
```
davsclaus: This should be `return true` — `SlackProducer` is fully synchronous and `callback.done(true)` is called in the `finally` block above. Returning `false` incorrectly signals async completion.

```suggestion
        return true;
```

components/camel-slack/src/main/java/org/apache/camel/component/slack/SlackProducer.java
```diff
@@ -134,17 +134,17 @@ private boolean sendMessageByWebhookURL(Exchange exchange, AsyncCallback callbac
...
             return true;
         } finally {
             callback.done(true);
         }
 
-        if (response.getCode() < 200 || response.getCode() > 299) {
-            exchange.setException(new CamelExchangeException("Error POSTing to Slack API: " + response.toString(), exchange));
-        }
-
         return false;
```
davsclaus: Same here — should be `return true` to match the synchronous `callback.done(true)` above.

```suggestion
        return true;
```

davsclaus (changes requested):
Good catch on the exception ordering bug — the fix correctly moves the error check inside the `try` block so the exception is set before `callback.done(true)` fires.

One issue remains: both `sendMessageByToken` and `sendMessageByWebhookURL` return `false` at the end, but `SlackProducer` is fully synchronous and the `finally` block calls `callback.done(true)`. The return value should be `true` to match — `return false` signals async completion which is incorrect here.

davsclaus (approved):

PR #23568, opened 2026-05-27
CAMEL-23601: camel-jbang - export layered Docker packaging for Main and Spring Boot

Croway:
## Summary

- **Camel Main**: Dockerfiles now use layered packaging — dependencies are copied to `target/lib/` via `maven-dependency-plugin:copy-dependencies`, and the thin `.original` JAR is copied separately. On code-only changes, only the ~6 KB application layer is invalidated instead of the full 10+ MB fat JAR.
- **Spring Boot**: Dockerfiles use a multi-stage build with Spring Boot's `jarmode=tools extract --layers` to split the fat JAR into 4 Docker layers (dependencies, spring-boot-loader, snapshot-dependencies, application). Uses the new `jarmode=tools` syntax (replacing the deprecated `jarmode=layertools` removed in Spring Boot 4.1).
- **Quarkus**: Already uses fast-jar layered packaging. `--quarkus-package-type` is now deprecated and hidden.
- Runtime-specific Dockerfile templates: `Dockerfile21.ftl`/`Dockerfile25.ftl` for Camel Main, `Dockerfile-spring-boot21.ftl`/`Dockerfile-spring-boot25.ftl` for Spring Boot. Selection is driven by `ExportSpringBoot.getDockerfileTemplateName()`.
- `main-pom.ftl` adds `maven-jar-plugin` (classpath manifest) and `maven-dependency-plugin:copy-dependencies` to produce layered artifacts alongside the existing fat JAR.
- Base images: all bumped to ubi9 1.24. Java 25 switched from eclipse-temurin to ubi9/openjdk-25:1.24 / ubi9/openjdk-25-runtime:1.24

JIRA: https://issues.apache.org/jira/browse/CAMEL-23601

Credit to Simon Hartl for originally raising the concern.

## Test plan

- [x] Unit tests pass (`TemplateHelperTest` — 23 tests, `ExportTest` — 69 tests)
- [x] Exported Camel Main, Spring Boot, and Quarkus projects with `camel export`
- [x] Built all three with `mvn package`
- [x] Verified JARs run locally (`java -jar`)
- [x] Verified thin JAR + `lib/` runs for Camel Main
- [x] Built Docker images for all three runtimes
- [x] Ran all three Docker containers successfully


oscerd (approved):

davsclaus (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportQuarkus.java
```diff
@@ -318,16 +318,13 @@ protected String applicationPropertyLine(String key, String value) {
...
-            // For uber-jar, the generic Dockerfile works as-is
-            super.copyDockerFiles(buildDir);
-        } else {
-            // For fast-jar, use a Quarkus-specific JVM Dockerfile
-            Files.createDirectories(dockerSrc);
-            InputStream is
-                    = ExportQuarkus.class.getClassLoader().getResourceAsStream("quarkus-docker/Dockerfile.jvm");
+        Files.createDirectories(dockerSrc);
+
+        // Use Quarkus-specific layered Dockerfile (fast-jar)
```
davsclaus: Minor consistency concern: the uber-jar branch is removed here, but `quarkusPackageType` is still used at line 150 (to select the `appJar` path) and line 386 (passed to the POM template as `QuarkusPackageType`). If someone passes `--quarkus-package-type uber-jar` (deprecated but accepted), Quarkus will produce an uber-jar but this Dockerfile will expect fast-jar layout.

Since the option is deprecated and hidden, this is a low-risk edge case — but worth considering a warning or removing the remaining uber-jar references entirely.

davsclaus (commented):
Nice work on the layered Docker packaging — the per-runtime approach is well designed and the upgrade guide entry is clear.

One minor consistency note: `ExportQuarkus.copyDockerFiles()` now always generates a fast-jar Dockerfile, but the deprecated `--quarkus-package-type` option is still partially functional (line 150 selects the `appJar` path, line 386 passes `QuarkusPackageType` to the POM template which sets `quarkus.package.jar.type`). If a user passes `--quarkus-package-type uber-jar`, Quarkus will build an uber-jar but the generated Dockerfile will expect fast-jar layout. Since the option is deprecated and hidden this isn't blocking, but you might consider either emitting a warning when `uber-jar` is passed, or removing the remaining uber-jar logic entirely.

Also a minor question: the JDK 21 Spring Boot template uses `ubi9/openjdk-21:1.23` for the builder stage and `ubi9/openjdk-21-runtime:1.24` for the runtime stage — are these the intended versions, or should both use the same release?

Croway:
Good point @davsclaus , I'll update the docker base image to the latest, and use ubi9 for java25, for the quarkus part, I'll leave as it was, the property is deprecated, once we remove the property, we'll remove the uber-jar handling as well

PR #23566, opened 2026-05-27
CAMEL-22894: Extract jq/jsonpath/xpath/simpleJsonpath Simple functions into QueryLanguageFunctionFactory

ammachado:
# Description

Continues the [CAMEL-22894](https://issues.apache.org/jira/browse/CAMEL-22894) series of `SimpleFunctionExpression` refactoring.

This PR extracts the four embedded-language functions (`jq`, `jsonpath`, `xpath`, `simpleJsonpath`) that previously lived inline in `SimpleFunctionExpression.createSimpleCustomLanguage()` into a new dedicated `QueryLanguageFunctionFactory`. The factory is registered in `SimpleFunctionDispatcher.BUILT_INS` alongside the factories introduced in the prior PRs (#23263, #23382, #23388, #23432, #23445, #23544).

Also fixes a minor dead-code issue noted in the #23544 review: `PropertiesFunctionFactory.createCode()` had a `parts.length > 2` guard that can never be true because the preceding `split(":", 2)` call limits the result to at most 2 elements.

`SimpleFunctionExpression` is now ~664 lines, down from the original 3,713.

_Claude Code on behalf of Adriano Machado_

# Target

- [x] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

# Apache Camel coding standards and style

- [x] I checked that each commit in the pull request has a meaningful subject line and body.
- [x] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

PR #23569, opened 2026-05-27
CAMEL-22894: Extract bean: Simple function into BeanFunctionFactory

ammachado:
# Description

Continues the [CAMEL-22894](https://issues.apache.org/jira/browse/CAMEL-22894) series of `SimpleFunctionExpression` refactoring (follows #23566).

The `bean:` function previously had its ref/method/scope parsing logic duplicated in two private methods inside `SimpleFunctionExpression`: one for the runtime expression path and one for the CSimple code-generation path. This PR extracts both into a new `BeanFunctionFactory` with a shared `parseBeanRemainder` helper, eliminating the duplication and reducing `SimpleFunctionExpression` by ~100 more lines.

`SimpleFunctionExpression` is now ~560 lines, down from the original 3,713.

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
@@ -21,6 +21,7 @@ import java.util.function.Predicate;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.Expression;
+import org.apache.camel.language.simple.functions.BeanFunctionFactory;
 import org.apache.camel.language.simple.functions.BodyFunctionFactory;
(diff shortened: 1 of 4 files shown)


davsclaus (approved):
Clean extraction following the established CAMEL-22894 pattern. Verified behavioral equivalence: `createFunction` preserves the same `Language.createExpression()` call with identical properties layout, `parseBeanRemainder()` correctly handles all three syntaxes (dot, double-colon, query-string) including the `(` guard for parameter declarations, and the `type:` class resolution stays in `createFunction` only.

Ordering change is safe — `bean:` was the last check before `tryCreateBuiltIn()`, now it's the last item in the BUILT_INS list.

Minor improvement: `createCode` now correctly handles the `scope != null && method == null` case that was silently ignored in the original.

Tests cover all syntax variants, createCode branches, and unrecognized input.

PR #23576, opened 2026-05-27
CAMEL-22894: Extract OGNL/exchange-property and function(...) Simple functions into dedicated factories

ammachado:
# Description

Continues the ongoing extraction of `SimpleFunctionExpression` inline logic into dedicated `SimpleLanguageFunctionFactory` implementations (CAMEL-22894).

This PR extracts two groups into new factories:

**`ExchangeFunctionFactory`** — OGNL/exchange-property group:

- `${camelContext.OGNL}` — CamelContext OGNL navigation
- `${exception.OGNL}` — exception OGNL navigation
- `${exceptionAs(type).OGNL}` — typed exception cast with optional OGNL (CSimple code-gen only)
- `${exchangeProperty.name}`, `${exchangeProperty:name}`, `${exchangeProperty[name]}` — plain property access
- `${exchangePropertyAs(key, type)}` — typed property access (CSimple code-gen only)
- `${exchangePropertyAsIndex(key, type, index)}` — indexed typed property access (CSimple code-gen only)
- `${exchange.OGNL}` — raw Exchange OGNL navigation

**`CustomFunctionFactory`** — user-defined custom functions:

- `${function(name)}` — invoke registered custom function with body as input
- `${function(name, exp)}` — invoke registered custom function with explicit expression as input

All parsing and error messages are preserved exactly from the removed inline blocks. Prefix ordering within each factory is intentional to avoid incorrect prefix matches (e.g. `exchangePropertyAsIndex(` before `exchangePropertyAs(` before `exchangeProperty`).

`SimpleFunctionExpression` drops from ~560 to 295 lines after this extraction.

A post-review fix was also applied: the `exception` OGNL invalid-OGNL error message in `createCode` was incorrectly saying `${exceptionAs(type).OGNL}` instead of `${exception.OGNL}` (copy-paste from the branch above it). A `createCode` invalid-OGNL test was added to catch this class of regression.

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleFunctionDispatcher.java
@@ -26,6 +26,7 @@ import org.apache.camel.language.simple.functions.BodyFunctionFactory;
 import org.apache.camel.language.simple.functions.CollateFunctionFactory;
 import org.apache.camel.language.simple.functions.CollectionFunctionFactory;
 import org.apache.camel.language.simple.functions.DateFunctionFactory;
+import org.apache.camel.language.simple.functions.ExchangeFunctionFactory;
 import org.apache.camel.language.simple.functions.HeaderFunctionFactory;
 import org.apache.camel.language.simple.functions.JoinFunctionFactory;
 import org.apache.camel.language.simple.functions.MathFunctionFactory;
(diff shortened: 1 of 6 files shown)


davsclaus (approved):
Clean extraction that continues the established CAMEL-22894 pattern. Prefix ordering is preserved correctly, the `exception` OGNL error message bug fix is valid, and tests cover both `createFunction` and `createCode` paths.

**Optional suggestion:** `testCreateCodeExceptionInvalidOgnl` validates the exception is thrown but doesn't assert the message content. Since the bug was specifically a wrong error message (`exceptionAs` vs `exception`), asserting the message would be a stronger regression guard.

PR #23613, opened 2026-05-28
CAMEL-23639: Publish the Camel JBang MCP server as a Claude Code plugin

k-krawczyk:
## Summary

Adds a Claude Code plugin marketplace to the repository so users can install the Camel MCP server with two commands instead of hand-editing `.mcp.json`:

```
claude plugin marketplace add apache/camel
claude plugin install camel-mcp@camel-marketplace
```

## Changes

- `.claude-plugin/marketplace.json` — defines the `camel-marketplace` marketplace at the repo root
- `plugins/camel-mcp/.claude-plugin/plugin.json` — plugin manifest (Apache-2.0)
- `plugins/camel-mcp/.mcp.json` — launches the server over STDIO via JBang (`org.apache.camel:camel-jbang-mcp:LATEST:runner`); auto-loaded when the plugin is enabled
- docs: new "Install as a plugin (recommended)" subsection in `camel-jbang-mcp.adoc`; the manual `.mcp.json` setup is kept as an alternative. The existing setup examples were also switched to `LATEST`.

No Java/source changes — configuration and documentation only. JBang remains a prerequisite, which the docs call out.

## Versioning

Per review feedback, the server is referenced as `LATEST` rather than a pinned version. JBang resolves `LATEST` to the newest release on Maven Central, so the file does not need a bump on each Camel release. Verified locally that `jbang ...:camel-jbang-mcp:LATEST:runner` resolves and the MCP server starts and lists its tools.

## Testing

- `claude plugin validate` passes for both the marketplace and the plugin manifest
- Adding the marketplace from a local path loads the `camel-mcp` plugin
- The MCP server starts via JBang (`LATEST`) and `tools/list` returns the expected tools

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,0 +1,16 @@
+{
+  "name": "camel-marketplace",
+  "description": "Apache Camel plugins for Claude Code, including the Camel MCP server.",
+  "owner": {
+    "name": "Apache Camel",
+    "email": "dev@camel.apache.org",
+    "url": "https://camel.apache.org"
+  },
+  "plugins": [
+    {
+      "name": "camel-mcp",
+      "source": "./plugins/camel-mcp",
+      "description": "Apache Camel MCP server: exposes the Camel Catalog plus specialized tools (catalog exploration, route and YAML DSL validation, security hardening, error diagnosis, dependency check, OpenAPI contract-first scaffolding, and version-to-version migration) to AI coding assistants over the Model Context Protocol. Requires JBang on the PATH."
+    }
+  ]
+}
--- a/docs/user-manual/modules/ROOT/pages/camel-jbang-mcp.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-jbang-mcp.adoc
@@ -438,7 +438,7 @@ configuration):
       "command": "jbang",
       "args": [
         "-Dquarkus.log.level=WARN",
-        "org.apache.camel:camel-jbang-mcp:4.18.0:runner"
+        "org.apache.camel:camel-jbang-mcp:LATEST:runner"
       ]
     }
   }
@@ -457,7 +457,7 @@ Add the server to your MCP configuration:
       "command": "jbang",
       "args": [
         "-Dquarkus.log.level=WARN",
-        "org.apache.camel:camel-jbang-mcp:4.18.0:runner"
+        "org.apache.camel:camel-jbang-mcp:LATEST:runner"
       ]
     }
   }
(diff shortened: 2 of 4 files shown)


gnodet (approved):
Nice addition — makes the Camel MCP server installable as a Claude Code plugin via the marketplace mechanism. The structure follows the expected plugin layout (marketplace.json, plugin.json, .mcp.json).

One minor note: `.mcp.json` hardcodes `4.20.0:runner`. This will need updating with each release. Consider whether a SNAPSHOT or latest-resolution mechanism would be more appropriate for the development branch, or document the version-pinning policy.

Doc update to `camel-jbang-mcp.adoc` is clean — the recommended path via `claude plugin marketplace add` / `claude plugin install` is clearly documented, with manual configuration as a fallback.

LGTM.

davsclaus:
Can you use LATEST as the version number, or is that only something a maven downloader understands

davsclaus (approved):
Clean config-and-docs-only PR — well-structured marketplace definition, plugin manifest, and doc update. No Java changes, proper JIRA linkage, correct license.

The version pin to `4.20.0` in `.mcp.json` is the right call since JBang can't resolve SNAPSHOTs. Worth a follow-up issue to wire version bumps into the release process as the author suggested.

davsclaus:
Instead of pinning `4.20.0`, could we use `LATEST` here?

```json
"args": [
  "-Dquarkus.log.level=WARN",
  "org.apache.camel:camel-jbang-mcp:LATEST:runner"
]
```

JBang supports `LATEST` as a version marker — it resolves to the latest release on Maven Central. This would eliminate the need to bump the version on every Camel release. There's a small cold-start cost (JBang checks Maven Central on first run, then caches), but it avoids the version going stale.

davsclaus:
@k-krawczyk can you try to use LATEST if that works for you

oscerd (approved):
Generally this is a good PR. I would use latest instead of pinning.

zbendhiba (changes requested):
Thanks for your contribution.

LGTM! Let us know if LATEST work.

k-krawczyk:
Thanks all — switched to LATEST as suggested.
@zbendhiba   
I verified it works: jbang -Dquarkus.log.level=WARN org.apache.camel:camel-jbang-mcp:LATEST:runner
resolves to the latest release on Maven Central, the MCP server starts over STDIO, and tools/list
returns all 27 tools. Updated both the plugin .mcp.json and the documentation setup examples to use
LATEST, so the version no longer needs bumping on each Camel release.

zbendhiba (approved):

davsclaus:
should keywords include java
and should it also mention spring boot and quarkus as camel has first class use with these

k-krawczyk:
> should keywords include java and should it also mention spring boot and quarkus as camel has first class use with these
Good call — added java, spring-boot, and quarkus to the plugin keywords (now: camel, java, integration, mcp, model-context-protocol, jbang, spring-boot, quarkus, catalog). Pushed.

davsclaus:
LGTM

PR #23633, opened 2026-05-29
CAMEL-23643: Add service version to test-infra metadata

davsclaus:
## Summary

- Add `serviceVersion` field to `@InfraService` annotation and generated `metadata.json`
- Auto-detect service version from container image tags in `container.properties` (65 container-based services)
- Auto-detect service version from Maven dependency resolution via `META-INF/infra-service.properties` GA hints (6 embedded/special services: FTP, FTPS, SFTP, Hazelcast, Ignite, SMB)
- All 68 test-infra services now report their service version in metadata

## How it works

Three-tier version detection in `CamelTestInfraGenerateMetadataMojo`:

1. **Annotation override** — `@InfraService(serviceVersion = "...")` for manual specification
2. **Container image tag** — scans `container.properties` from dependency JARs, matches alias to image key, extracts version from Docker image tag (handles cross-module properties, multi-image modules like Kafka, RocketMQ's separate `.version` pattern)
3. **GA hint lookup** — reads `META-INF/infra-service.properties` files that map service alias to Maven `groupId:artifactId`, then resolves the artifact version from the dependency tree. Zero maintenance on version bumps.

## Test plan

- [x] Build `camel-package-maven-plugin` and `camel-test-infra-all`
- [x] Verify all 68 services have `serviceVersion` in generated `metadata.json`
- [x] Verify container-based services extract correct image tags (Kafka=4.3.0, Redpanda=v26.1.9, Redis=7.4.9-alpine, etc.)
- [x] Verify embedded services resolve from Maven dependencies (FTP=1.2.1, SFTP=2.18.0, Hazelcast=5.4.0, Ignite=2.18.0, SMB=0.14.0)
- [x] Verify multi-image modules match correct per-implementation version (Kafka 4 variants, HiveMQ 2 variants)

### DIFF
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/test-infra/metadata.json
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/test-infra/metadata.json
@@ -6,7 +6,8 @@
   "aliasImplementation" : [ ],
   "groupId" : "org.apache.camel",
   "artifactId" : "camel-test-infra-milvus",
-  "version" : "4.21.0-SNAPSHOT"
+  "version" : "4.21.0-SNAPSHOT",
+  "serviceVersion" : "v2.6.17"
 }, {
   "service" : "org.apache.camel.test.infra.aws.common.services.AWSInfraService",
   "description" : "Local AWS Services with LocalStack",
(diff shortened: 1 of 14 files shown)


apupier (commented):
I think we need to update the job which is creating the PR to upgrade the container

davsclaus:
> I think we need to update the job which is creating the PR to upgrade the container

I do not undertand

apupier:
> > I think we need to update the job which is creating the PR to upgrade the container
> 
> I do not undertand

When this job is running https://github.com/apache/camel/blob/main/.github/workflows/check-container-versions.yml , it is updating the value in the container.properties file (see fro instance https://github.com/apache/camel/pull/23619 ). With this PR, it should also update the value in the catalog/[...]/test-infra/metadata.json file, isn't it?

davsclaus:
Good point @apupier — tracked in [CAMEL-23648](https://issues.apache.org/jira/browse/CAMEL-23648) to update the `check-container-versions.yml` workflow to also regenerate `metadata.json` after bumping container image versions.

PR #23639, opened 2026-05-29
CAMEL-23646: Fix inverted BackgroundTask return values in sjms, pgevent and master

Croway:
## Summary

The `BackOff` → `BackgroundTask` migration (CAMEL-22206) preserved callback return values verbatim, but the two APIs have **opposite boolean semantics**: `BackOff` returns `false` to stop, `BackgroundTask` returns `true` to stop. This caused infinite retry loops after successful recovery in three components.

**camel-sjms/sjms2** — after a transient JMS connection interruption (reported with Oracle AQ), the consumer enters an infinite recovery loop. Each iteration destroys the working connection via `refreshConnection()` while `initConsumers()` skips re-creation (`consumers != null`), leaving consumers attached to closed sessions. Message consumption is permanently blocked.

**camel-pgevent** — after successful PostgreSQL reconnection, the task keeps calling `initConnection()` every `reconnectDelay`, leaking connections and duplicating notification listeners.

**camel-master** — after the delegated consumer starts, the task keeps calling `ServiceHelper.startService()` until budget exhaustion (default 10 iterations). Can revive a consumer after leadership loss.

### DIFF
--- a/components/camel-master/src/main/java/org/apache/camel/component/master/MasterConsumer.java
+++ b/components/camel-master/src/main/java/org/apache/camel/component/master/MasterConsumer.java
@@ -203,7 +203,7 @@ public class MasterConsumer extends DefaultConsumer implements ResumeAware<Resum
 
                 LOG.info("Leadership taken. Attempt #{} success. Consumer started: {}", leaderTask.iteration(),
                         delegatedEndpoint);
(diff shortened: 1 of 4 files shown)


davsclaus (approved):
50/50 % chance ;)

PR #23646, opened 2026-05-29
CAMEL-23648: Update container version workflow to regenerate metadata.json

davsclaus:
## Summary

- Add `update-metadata-version.py` script to update `serviceVersion` in metadata.json files when container image versions are bumped
- Update `check-container-versions.yml` workflow to call this script and include the regenerated metadata.json files in each commit

After CAMEL-23643 added `serviceVersion` to metadata.json (extracted from container image tags), the container version upgrade workflow needs to also update these generated files when bumping container.properties — otherwise CI detects stale generated files and fails the automated PRs.

The Python script matches entries by `artifactId` and `old serviceVersion` to update only the affected entries, and produces output matching Jackson's `DefaultPrettyPrinter` format so the result is identical to what the Maven build generates.

## Test plan

- [x] Round-trip formatting test: update + revert produces identical output to original
- [x] Multi-entry module test: AWS (16 entries) all updated, Kafka entries unaffected
- [x] Both metadata.json files (test-infra-all and catalog) tested

### DIFF
--- a/.github/actions/check-container-upgrade/update-metadata-version.py
+++ b/.github/actions/check-container-upgrade/update-metadata-version.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


apupier (approved):
personally not familiar with Python scripts but seems coherent.
Not sure how to test that before merging, so fine to merge and then monitor closely the result of container upgrade job

PR #23711, opened 2026-06-02
CAMEL-23063: Add camel-a2a component for Agent-to-Agent (A2A) protoco…

luigidemasi:
…l integration

Adds the `camel-a2a` component implementing the [A2A (Agent-to-Agent) protocol](https://a2a-protocol.org/latest/) for Apache Camel. The component provides both producer and consumer support, enabling Camel
  routes to participate in standardized agent-to-agent communication.

  Features

  - **Consumer** — exposes a Camel route as an A2A server with automatic agent card serving at `/.well-known/agent-card.json`
  - **Producer** — sends A2A messages (MESSAGE_SEND, MESSAGE_STREAM, TASK_GET, TASK_CANCEL, etc.) to remote agents
  - **Protocol bindings** — REST and JSON-RPC 2.0
  - **Streaming** — Server-Sent Events (SSE) for real-time task updates
  - **Authentication** — API key, Bearer token, OAuth 2.0, OpenID Connect
  - **Task management** — in-memory task store with TTL, push notifications, subscriber management
  - **Agent card** — discovery, resolution (file < bean < URI params), and auto-serving
  - **Type converters** — Message/Task to String, plus Simple DSL functions (`a2aText()`, `a2aData()`, `a2aFile()`)
  - **Progress API** — `A2AProgress.emit()` for status updates from any route

  **Demo**

  A multi-agent demo showcasing `camel-a2a` in action is available at:

[https://github.com/luigidemasi/camel-a2a-morning-routine](https://github.com/luigidemasi/camel-a2a-morning-routine)


components/camel-ai/camel-a2a/src/generated/resources/a2a.json
```diff
@@ -0,0 +1,15 @@
+{
+  "other": {
+    "kind": "other",
+    "name": "a2a",
+    "title": "A2a",
+    "description": "Camel A2A component for Agent-to-Agent protocol communication",
+    "deprecated": false,
+    "firstVersion": "4.21.0",
+    "label": "ai,a2a",
```
apupier: What is the reason for the a2a label? Plan to have several agent to agents components?

What about ai and agent labels? maybe we will have some components acting as agents or doing agent to something else communication?

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/auth/A2AAuthHandler.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return all;
+    }
+
+    private String resolvePreferredSchemeType() {
+        if (configuration.getOauthProfile() != null) {
+            return "oauth2";
+        }
+        if (configuration.getBearerToken() != null) {
+            return "http";
```
apupier: Is it for both http and https?

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/model/Task.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public static Builder builder(Task task) {
+        return new Builder()
+                .id(task.id)
+                .contextId(task.contextId)
+                .status(task.status)
+                .artifacts(task.artifacts != null ? new ArrayList<>(task.artifacts) : null)
+                .history(task.history != null ? new ArrayList<>(task.history) : null)
+                .metadata(task.metadata != null ? new HashMap<>(task.metadata) : null);
```
apupier, line 64: What about providing empty List or Map instead of null?

it avoids having to check for nullness

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AConsumer.java
```diff
@@ -0,0 +1,1278 @@
...
+                .status(task.status())
+                .build();
+    }
+
+    private void filterInboundHeaders(Exchange exchange) {
+        exchange.getMessage().getHeaders().entrySet().removeIf(entry -> {
+            String lower = entry.getKey().toLowerCase(Locale.ENGLISH);
+            return lower.startsWith("camela2a") || lower.startsWith("org.apache.camel.");
+        });
+    }
```
davsclaus, line 1742: **[Blocking — Security]** This filter only strips `CamelA2A*` and `org.apache.camel.*` headers but does **not** strip all `Camel*` headers. Per the project's [security model](https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/security-model.adoc) and the committer review checklist, consumers exposed to untrusted input must block `Camel*` / `camel*` / `org.apache.camel.*` **case-insensitively**.

The current filter leaves headers like `CamelBeanMethodName`, `CamelExecCommandExecutable`, etc. unfiltered — this is the exact attack vector from CVE-2025-27636 and five follow-on CVEs.

```suggestion
            return lower.startsWith("camel") || lower.startsWith("org.apache.camel.");
```

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/push/PushNotificationDispatcher.java
```diff
@@ -0,0 +1,171 @@
...
+    private final ScheduledExecutorService executor;
+
+    public PushNotificationDispatcher(HttpClient httpClient, A2ATaskStore store,
+                                      int maxRetries, long initialBackoffMs) {
+        this.httpClient = httpClient;
+        this.store = store;
+        this.maxRetries = maxRetries;
+        this.initialBackoffMs = initialBackoffMs;
+        this.executor = Executors.newScheduledThreadPool(4, r -> {
+            Thread t = new Thread(r, "A2APushDispatcher");
```
davsclaus: **[Medium]** This creates a `ScheduledExecutorService` via `Executors.newScheduledThreadPool()` directly, bypassing Camel's `ExecutorServiceManager`. The threads won't be tracked by JMX/monitoring, won't participate in graceful shutdown, and won't respect Camel's thread pool profile.

The `HttpClient` executor in `A2AEndpoint.doStart()` correctly uses `getCamelContext().getExecutorServiceManager()` — this should follow the same pattern. Since `PushNotificationDispatcher` doesn't have access to `CamelContext`, consider passing in a `ScheduledExecutorService` created by the endpoint instead.

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AConsumer.java
```diff
@@ -0,0 +1,1278 @@
...
+            exchange.getMessage().setHeader(Exchange.CONTENT_TYPE, A2AConstants.CONTENT_TYPE);
+        }
+    }
+
+    void handleListTasks(Exchange exchange) throws Exception {
+        Integer pageSize = exchange.getMessage().getHeader(A2AConstants.LIST_PAGE_SIZE, Integer.class);
+        String contextId = exchange.getMessage().getHeader(A2AConstants.LIST_CONTEXT_ID, String.class);
+
+        filterInboundHeaders(exchange);
+
```
davsclaus, line 974: **[Low]** `pageSize` and `contextId` are read from exchange headers **before** `filterInboundHeaders` is called (lines 734-735). An untrusted HTTP client could inject `CamelA2AListPageSize` and `CamelA2AListContextId` headers. While the impact is low (pageSize is capped at 100, contextId is just a filter key), the filter should run before any header reads — consistent with all other handlers in this class.

davsclaus (changes requested):
Thank you for this substantial contribution — the A2A component is well-structured, follows Camel conventions, and has impressive test coverage. I have one blocking security finding and a few additional items.

### Positive observations

- Good security defaults: `followRedirects=false`, agent card loader blocks redirects, `WebhookUrlValidator` has SSRF protection with private IP/IPv6 checks
- Comprehensive test coverage across operations, protocols, streaming, auth, and push notifications
- Proper `MojoHelper` registration, `SimpleFunctionDispatcher` integration, `secret = true` annotations on sensitive params
- No new external dependencies beyond Jackson (already managed by parent BOM)

### General (non-inline) findings

- **Missing upgrade guide entry** — per project conventions, new user-visible features should be documented in `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_21.adoc`
- **Dead code** — the private method `executeStreamProcessing` in `A2AConsumer` (line ~1077) appears unused and should be removed
- **CI** — one CI check ("build") is currently failing; the three matrix builds pass, so this may be infrastructure, but worth confirming

**Note:** This review evaluates the PR against Apache Camel's project rules and conventions. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AConsumer.java
```diff
@@ -0,0 +1,1278 @@
...
+                A2ATaskStore store = getEndpoint().getTaskStore();
+                store.updateStatusAndNotify(taskId, new TaskStatus(TaskState.FAILED));
+            } finally {
+                emitter.close();
+                releasePermit();
+            }
+        });
+    }
+
+    private DefaultStreamEmitter executeStreamProcessing(SendMessageRequest request) throws Exception {
```
orpiske: This one was flagged as unused. Please, can you double check @luigidemasi ?

orpiske (approved):
LGTM ... one minor thing about an unused method, but other than that looks fine.

Thanks @luigidemasi !

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/auth/A2AAuthHandler.java
luigidemasi: @apupier there's a misunderstanding here, the `"http"` here is **not** the transport protocol, it's the [OpenAPI 3.0 Security Scheme `type`](https://spec.openapis.org/oas/v3.0.3#security-scheme-object) constant, which the A2A protocol adopts for its `securitySchemes` field.

  The `type` enum has four values:

  | `type` value | Meaning |
  |---|---|
  | `"apiKey"` | API key (header, query, or cookie) |
  | `"http"` | RFC 7235 HTTP Authentication — Bearer, Basic, Digest, etc. |
  | `"oauth2"` | OAuth 2.0 flows |
  | `"openIdConnect"` | OpenID Connect discovery |

  So `"http"` means "this scheme uses the `Authorization` header" (e.g. `Authorization: Bearer <token>`), which works identically over both HTTP and HTTPS. The actual
  transport encryption is determined by the endpoint URI scheme, not by this field.

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/model/Task.java
luigidemasi: This is a protocol model class serialized by Jackson, `null` fields are omitted from JSON output while empty collections serialize as `[]` / `{}`. 
The A2A spec distinguishes between "field absent" (not provided) and "field present but empty" (explicitly no items), so the copy constructor preserves the original semantics.

components/camel-ai/camel-a2a/src/generated/resources/a2a.json
luigidemasi: changed to `"label": "ai,agent"`

davsclaus: these should be generated from code via the Category.XXX you set in the endpoint class

davsclaus: This ensures we have a consistent label(s) across the board.

luigidemasi: ok let's keep only `ai` then

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/auth/A2AAuthHandler.java
zbendhiba: Maybe we could some meaningful variables, declared in a section that explains this spec. WDYT ?

davsclaus (approved):
All findings from my previous `CHANGES_REQUESTED` review have been addressed:

1. **Security (was blocking)** — `filterInboundHeaders` now strips all `camel*` headers case-insensitively, matching the project's security model for consumer header filtering.
2. **Header read ordering** — `filterInboundHeaders` is called before reading untrusted headers (`pageSize`, `contextId`); `httpPath` is extracted before the filter in other handlers.
3. **ExecutorService** — `PushNotificationDispatcher` now receives the executor via injection from `getCamelContext().getExecutorServiceManager().newScheduledThreadPool()`, properly shut down in `doStop()`.
4. **Upgrade guide** — entry added.
5. **Category/labels** — fixed.

gnodet (commented):
This PR adds the camel-a2a component implementing the Agent-to-Agent protocol — a significant new feature. The diff exceeds 20,000 lines, which is beyond what can be reviewed automatically via the GitHub API diff endpoint.

The CI checks all pass (build on JDK 17, 21, and 25), and the PR description is comprehensive — covering consumer/producer support, protocol bindings, streaming, authentication, task management, and type converters.

This PR requires manual review given its size and the scope of the new component (REST + JSON-RPC 2.0 bindings, SSE streaming, OAuth/OIDC auth, in-memory task store). I'd recommend reviewers focus on:
- Security: authentication implementation and header filtering
- Protocol correctness: JSON-RPC 2.0 compliance
- Resource management: task store TTL and cleanup

oscerd (approved):

davsclaus:
LGTM @luigidemasi its a massive work so when you think its ready then go ahead and merge - we need it in 4.21 so we have both MCP and A2A

davsclaus:
Are we getting in shape for this to be included next week

luigidemasi:
> Are we getting in shape for this to be included next week

@davsclaus yes, it is almost ready and I think we are in good shape for next week. Before marking it ready, I want to finish and incorporate the OAuth authentication support for HTTP consumers from CAMEL-23685 (#23866) and CAMEL-23723 (#23926), so A2A lands with the right auth path for 4.21. 
I will update the PR later today.

PR #23713, opened 2026-06-02
CAMEL-23652: Fix route template local bean not found when param value matches bean name

davsclaus:
## Summary

- Fix bug where route template local beans (e.g., from kamelets) could not be referenced via `{{beanName}}` when a template parameter value coincidentally matched the bean name
- When `doStartRouteDefinitions()` processes local beans for uniqueness, the first loop detects "clashes" where a parameter value matches a bean key (e.g., `scheme=counter` where `counter` is a bean name). It removes the bean from the registry and renames it, but failed to add the original bean name as a template parameter — leaving `{{beanName}}` references unresolvable
- Added test `testLocalBeanWithParamValueClash` covering this scenario

## Test plan

- [x] Existing 72 route template tests pass (now 73 with the new test)
- [x] External reproducer project from JIRA issue passes
- [x] New test verifies the exact scenario: local bean referenced via `{{counter}}` with a parameter `scheme=counter` that coincidentally matches

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/builder/RouteTemplateLocalBeanTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/builder/RouteTemplateLocalBeanTest.java
@@ -949,4 +949,36 @@ public class RouteTemplateLocalBeanTest extends ContextTestSupport {
         return answer;
     }
 
+    @Test
+    public void testLocalBeanWithParamValueClash() throws Exception {
+        context.addRoutes(new RouteBuilder() {
+            @Override
+            public void configure() {
+                routeTemplate("myTemplate")
+                        .templateParameter("foo")
+                        .templateParameter("scheme")
+                        .templateBean("counter").typeClass(AtomicInteger.class).end()
+                        .from("direct:{{foo}}")
+                        .to("bean:{{counter}}?method=getAndIncrement");
+            }
+        });
+
+        context.start();
+
+        TemplatedRouteBuilder.builder(context, "myTemplate")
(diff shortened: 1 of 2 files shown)


oscerd (approved):

core/camel-core/src/test/java/org/apache/camel/builder/RouteTemplateLocalBeanTest.java
```diff
@@ -949,4 +949,36 @@ public static Processor createBuilderProcessorThree(String prefix) {
...
+        TemplatedRouteBuilder.builder(context, "myTemplate")
+                .parameter("foo", "one")
+                .parameter("scheme", "counter")
+                .routeId("myRoute")
+                .add();
+
+        assertEquals(1, context.getRoutes().size());
+
+        Object out = template.requestBody("direct:one", "World");
+        assertEquals(0, out);
```
apupier, line 977: to ease readability, it would be nice to add a message for assertion explaining why it expects 0

core/camel-core/src/test/java/org/apache/camel/builder/RouteTemplateLocalBeanTest.java
```diff
@@ -949,4 +949,36 @@ public static Processor createBuilderProcessorThree(String prefix) {
...
+        out = template.requestBody("direct:one", "World");
+        assertEquals(1, out);
```
apupier, line 979: to ease readability, it would be nice to add a message for assertion explaining why it expects 1

apupier (approved):

PR #23729, opened 2026-06-03
camel-test-junit6: Make more test configuration methods public for framework extensibility

jamesnetherton:
Motivation for this is to clean up and remove some old hacks in the Camel Quarkus test framework.

### DIFF
--- a/components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/TestExecutionConfiguration.java
+++ b/components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/TestExecutionConfiguration.java
@@ -163,7 +163,7 @@ public class TestExecutionConfiguration {
      * @return     <tt>true</tt> per class, <tt>false</tt> per test.
      */
     @Deprecated(since = "4.7.0")
-    protected TestExecutionConfiguration withCreateCamelContextPerClass(boolean createCamelContextPerClass) {
+    public TestExecutionConfiguration withCreateCamelContextPerClass(boolean createCamelContextPerClass) {
         this.createCamelContextPerClass = createCamelContextPerClass;
         return this;
     }
--- a/components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/CamelContextConfiguration.java
+++ b/components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/CamelContextConfiguration.java
@@ -319,7 +319,7 @@ public class CamelContextConfiguration {
     /**
      * A supplier that classes can use to create a {@link RouteBuilder} to define the routes for testing
      */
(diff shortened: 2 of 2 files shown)


davsclaus (approved):

components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/TestExecutionConfiguration.java
```diff
@@ -163,7 +163,7 @@ public boolean isCreateCamelContextPerClass() {
      * @return     <tt>true</tt> per class, <tt>false</tt> per test.
      */
     @Deprecated(since = "4.7.0")
-    protected TestExecutionConfiguration withCreateCamelContextPerClass(boolean createCamelContextPerClass) {
+    public TestExecutionConfiguration withCreateCamelContextPerClass(boolean createCamelContextPerClass) {
```
apupier, line 166: do we really want to make public a deprecated method?
Shouldn't we try to not use it at all on Camel Quarkus side?

jamesnetherton: I'm not sure of the history around it, but the API is somewhat incoherent. `withCreateCamelContextPerClass` is deprecated. But, `isCreateCamelContextPerClass` is not deprecated and is referenced in a few places that seem to be critical for the CQ test framework in some scenarios.

apupier: if I remember well, it was @orpiske which did the major work on the tests. @orpiske do you remember about this part? Any advice how to avoid the deprecated method?

gnodet (approved):
Small, focused change that makes `withRoutesSupplier()`, `withPostProcessor()`, and `withCreateCamelContextPerClass()` methods public in `CamelContextConfiguration` and `TestExecutionConfiguration`. This enables frameworks that extend Camel's test support (like Camel on Quarkus) to properly configure these options.

The change is backward-compatible — widening access from default/protected to public never breaks existing callers.

davsclaus (approved):

components/camel-test/camel-test-junit6/src/main/java/org/apache/camel/test/junit6/TestExecutionConfiguration.java
orpiske: I need to take a look at the my chat history, because I think the reason why I never removed this was indeed because it could break things on CEQ. I think, the right approach is either one of:

1. Undo this deprecation notice and add a comment stating that this is required for Quarkus or
2. Double check if it's not required in Camel Core and then, move what this API call does to the Quarkus equivalent (if possible). 

Option 1 might be easier, but 2 is likely a cleaner approach.

PR #23735, opened 2026-06-03
camel-jira: use non-deprecated maven url

JiriOndrusek:
Related to https://issues.apache.org/jira/browse/CAMEL-23389

# Description

The Atlassian Maven repository URL https://packages.atlassian.com/maven-external/ used in components/camel-jira/pom.xml was [deprecated as of February 1,
  2025](https://developer.atlassian.com/server/framework/atlassian-sdk/maven-package-mirrors/).

As of now, the old deprecated repo url started to fail in Camel Quarkus CI job, which builds Camel from source. The build fails during dependency resolution of camel-jira with:

```
  Failed to execute goal on project camel-jira: Could not collect dependencies for project org.apache.camel:camel-jira:jar:4.21.0-SNAPSHOT
  The following artifacts could not be resolved: com.atlassian.sal:sal-api:pom:7.2.1 (absent): Could not transfer artifact com.atlassian.sal:sal-api:pom:7.2.1 from/to atlassian
  (https://packages.atlassian.com/maven-external/): status code: 403, reason phrase: Forbidden (403)
```

  The solution should be to updated from https://packages.atlassian.com/maven-external/ to https://packages.atlassian.com/artifactory/maven-atlassian-all (which is used in suggested mvn setting)

(if the change does not help, we can change the url to https://maven.artifacts.atlassian.com/, which is the final
  destination all Atlassian repositorie where are URLs now redirect to)

# Target

# Tracking

### DIFF
--- a/components/camel-jira/pom.xml
+++ b/components/camel-jira/pom.xml
@@ -34,7 +34,7 @@
     <repositories>
         <repository>
             <id>atlassian</id>
-            <url>https://packages.atlassian.com/maven-external/</url>
+            <url>https://packages.atlassian.com/artifactory/maven-atlassian-all/</url>
             <name>atlassian external repo</name>
             <snapshots>
                 <enabled>false</enabled>


apupier:
see https://issues.apache.org/jira/browse/CAMEL-23389

davsclaus (approved):

apupier (changes requested):
we need to test the full build on a platform without any Maven cache. To be sure that we do not have third party depenencies which were previously hosted on https://packages.atlassian.com/maven-external are available on https://packages.atlassian.com/artifactory/maven-atlassian-all/ given that the change is:

```
We are changing how we provide customers and partners packages to build against our platforms. As of the 1st Feb 2025 we will no longer be providing third party packages at [packages.atlassian.com](http://packages.atlassian.com/maven-external). Customers and Partners will need to fetch these packages directly from upstream repositories.
```

apupier:
note that if it is blocking for something else, it will be safer to add the other repository without removing the old one as an intermediary step

apupier:
I launched a build locally after deleting .m2/repository content and removing my .m2/settings.xml

apupier (approved):
build success locally with:
* deleted .m2/repository
* deleted .m2/settings.xml
* ./mvnw clean install -Dquickly
* cd component/camel-jira then mvn clean install

PR #23738, opened 2026-06-03
CAMEL-23681: Optimize Exchange memory pressure: copy-on-write headers and lazy initialization

gnodet:
[CAMEL-23681](https://issues.apache.org/jira/browse/CAMEL-23681)

This PR reduces memory pressure during Exchange routing by implementing copy-on-write headers and targeted allocation optimizations.

### 1. Copy-on-Write Headers (Highest Impact)
Message copies now share the headers map until one side modifies it, avoiding expensive map copying on every `copyFrom()`.

**Implementation:**
- `CopyOnWriteHeadersMap` is a self-contained `Map` wrapper holding its own delegate, shared flag, and `HeadersMapFactory` — no back-reference to `DefaultMessage`
- `DefaultMessage.copyHeadersFrom()` wraps both source and target in `CopyOnWriteHeadersMap` sharing the same underlying map
- All mutation methods (`setHeader`, `removeHeader`, `removeHeaders`) delegate through `getHeaders()` which returns the COW wrapper transparently
- Read operations (get, containsKey, size, forEach, iteration) delegate directly without triggering a copy
- Write operations trigger a one-time clone via `HeadersMapFactory.newMap()`, then mutate the private copy
- COW-aware `keySet()`/`values()`/`entrySet()` views with lazy iterators — only `Iterator.remove()` triggers COW
- `reset()` nulls COW headers instead of clearing (avoids triggering a copy just to discard)
- Wrapper satisfies the full `Map` contract (`equals`, `hashCode`, `toString`, `forEach`)
- Not thread-safe — follows the same single-thread-per-Exchange model as `DefaultMessage`

**Impact:**
- Multicast 5-way with 20 headers: saves 4× header map allocations
- Splitter with 1000 items (read-only headers): saves ~999× header map allocations
- Only allocates a new map when headers are actually modified

### 2. Properties Map Reuse in Pooled Exchanges
`DefaultPooledExchange.done()` reuses cleared properties maps instead of discarding them. Maps larger than 50 entries are discarded to avoid holding oversized allocations.

## Testing

Unit tests in `DefaultMessageHeaderTest`:
- `testCopyOnWriteHeaders()` — verifies headers are shared until modification
- `testCopyOnWriteHeadersRemove()` — verifies removeHeader triggers COW
- `testCopyOnWriteHeadersRemoveMultiple()` — verifies removeHeaders triggers COW
- `testCopyOnWriteGetHeaders*()` — verifies mutations via `getHeaders().put/remove/clear/putAll` don't affect the original
- `testCopyOnWriteSetHeadersResetsShared()` — verifies `setHeaders()` breaks shared state
- `testCopyOnWriteMultipleCopiesIsolation()` — verifies multiple independent copies remain fully isolated
- `testLazyCopyOnWrite*()` — verifies read operations don't trigger COW, mutations via views do
- `testLazyPopulatedHeaders*()` — verifies lazy header population in subclasses (e.g., JmsMessage)

Internal-state tests in `CopyOnWriteHeadersTest` (same package as `DefaultMessage`):
- `testReadOperationsDoNotTriggerCow()` — reads and iteration preserve shared state
- `testForEachDoesNotTriggerCow()` — `forEach()` preserves shared state
- `testKeySetIterationDoesNotTriggerCow()` — keySet iteration preserves shared state, `iterator.remove()` triggers COW
- `testEntrySetIterationDoesNotTriggerCow()` — entrySet iteration preserves shared state, `iterator.remove()` triggers COW
- `testResetDoesNotCorruptSharedHeaders()` — reset drops the shared reference without clearing the shared map

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
@@ -53,7 +53,9 @@ public class DefaultMessage extends MessageSupport {
     @Override
     public void reset() {
         super.reset();
-        if (headers != null) {
+        if (headers instanceof CopyOnWriteHeadersMap) {
+            headers = null;
+        } else if (headers != null) {
             headers.clear();
         }
         removeTrait(MessageTrait.ATTACHMENTS);
@@ -62,7 +64,7 @@ public class DefaultMessage extends MessageSupport {
     @Override
     public Object getHeader(String name) {
         if (headers == null) {
-            // force creating headers
+            // force creating headers (supports lazy population in subclasses like JmsMessage)
             headers = createHeaders();
         }
 
@@ -204,22 +206,16 @@ public class DefaultMessage extends MessageSupport {
 
     @Override
     public void setHeader(String name, Object value) {
-        if (headers == null) {
-            headers = createHeaders();
-        }
-        headers.put(name, value);
+        getHeaders().put(name, value);
     }
 
     @Override
     public Object removeHeader(String name) {
-        if (headers == null) {
-            // force creating headers
-            headers = createHeaders();
-        }
-        if (headers.isEmpty()) {
+        Map<String, Object> h = getHeaders();
+        if (h.isEmpty()) {
             return null;
         }
-        return headers.remove(name);
+        return h.remove(name);
     }
 
     @Override
@@ -229,29 +225,25 @@ public class DefaultMessage extends MessageSupport {
 
     @Override
     public boolean removeHeaders(String pattern, String... excludePatterns) {
-        if (headers == null) {
-            // force creating headers
-            headers = createHeaders();
-        }
-        if (headers.isEmpty()) {
+        Map<String, Object> h = getHeaders();
+        if (h.isEmpty()) {
             return false;
         }
 
         // special optimized
         if (excludePatterns == null && "*".equals(pattern)) {
-            headers.clear();
+            h.clear();
             return true;
         }
 
-        final Set<String> toBeRemoved = PatternHelper.matchingSet(headers, pattern, excludePatterns);
+        final Set<String> toBeRemoved = PatternHelper.matchingSet(h, pattern, excludePatterns);
 
         if (toBeRemoved != null) {
-            if (toBeRemoved.size() == headers.size()) {
-                // special optimization when all should be removed
-                headers.clear();
+            if (toBeRemoved.size() == h.size()) {
+                h.clear();
             } else {
                 for (String key : toBeRemoved) {
-                    headers.remove(key);
+                    h.remove(key);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
```diff
@@ -36,7 +36,8 @@
  * implementation uses the {@link org.apache.camel.util.CaseInsensitiveMap CaseInsensitiveMap}.
  */
 public class DefaultMessage extends MessageSupport {
-    private Map<String, Object> headers;
+    protected Map<String, Object> headers;
```
davsclaus: Widening `headers` from `private` to `protected` exposes it to subclasses (`JmsMessage`, `GenericFileMessage`, `HttpMessage`, `NettyHttpMessage`, etc.) which could directly modify the field and bypass the COW invariant. Consider keeping it `private` and providing accessor methods instead.

Also, the `headersShared` flag should arguably be `private` too — subclasses should go through the COW-aware methods.

davsclaus (changes requested):
Thank you for this optimization work — the lazy headers initialization and pooled exchange property reuse are clean wins. However, the copy-on-write headers optimization has a fundamental safety gap that needs to be addressed before merge.

**The core issue:** `getHeaders()` (line 265) returns the shared map reference directly without triggering a COW copy. Production code widely uses `getHeaders().put(...)`, `.putAll(...)`, and `.clear()` to mutate headers — these bypass the COW guard entirely and would silently corrupt the shared map on both messages.

Examples of affected call sites:
- `ClaimCheckAggregationStrategy` — `.getHeaders().putAll()`, `.put()`, `.clear()`
- `DefaultFluentProducerTemplate` — `.getHeaders().putAll(headers)`
- `JmsBinding` — `.getHeaders().putAll(jmsHeaders)`
- `DefaultCxfBinding` — `.getHeaders().putAll()`, `.put()`
- And many more across components

These are exercised precisely in multicast, splitter, and recipient-list patterns — the high-fanout scenarios this optimization targets. To be safe, `getHeaders()` would also need to trigger a COW copy when `headersShared` is true — but that may negate most of the optimization benefit.

Additionally, `setHeaders()` (line 273) replaces the `headers` reference but doesn't reset `headersShared = false` — a minor inconsistency.

gnodet:
Thanks for the thorough review! All issues fixed:
1. getHeaders() now triggers COW copy when headersShared is true
2. setHeaders() resets headersShared = false
3. Fields reverted to private with package-private accessor methods for MessageSupport

The optimization still helps for read-only access patterns (getHeader(name), hasHeaders()) and lazy initialization.

core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
gnodet: Fixed — reverted both fields to private with package-private accessor methods for MessageSupport to access them safely.

davsclaus (approved):
The previous blocking concern (`getHeaders()` bypassing COW) is fully resolved. The new `CopyOnWriteHeadersMap` approach is the right design — it intercepts all mutation paths including `getHeaders().put()`, `.putAll()`, `.clear()`, and collection view mutations (`keySet().remove()`, `entrySet()` iteration, etc.).

The `headers` field is back to `private`, `setHeaders()` now resets the shared flag, and package-private accessors keep the COW internals contained. Test coverage is thorough — 17 new test methods including the exact bypass scenario I flagged and a concurrent-safety test with 10 threads.

Minor: `CopyOnWriteHeadersMap` uses FQCNs for `java.util.function.Function` / `BiFunction` instead of imports, but the build's OpenRewrite plugin will auto-fix this.

davsclaus (commented):
A couple of observations on the two new commits since my last approval. The core COW design remains sound.

**Likely test failure:** `testLazyCopyOnWriteKeySetIteration`, `testLazyCopyOnWriteValues`, and `testLazyCopyOnWriteEntrySet` assert that `isHeadersShared()` is still `true` after for-each iteration. But `CopyOnWriteSet.iterator()` and `CopyOnWriteCollection.iterator()` eagerly call `ensureWritable()`, which sets `headersShared = false`. These tests will likely fail in CI — either the tests need updating or the iterators should use a lazy COW wrapper instead of eagerly triggering the copy.

**Scope drift:** Commit `0df9d4c` changes `.github/actions/incremental-build/incremental-build.sh` (tail 50 → 500). This is unrelated to the Exchange memory optimization and would be better as a separate commit/PR.

davsclaus:
rebasing so CI will do full test again

davsclaus (approved):
All previously flagged issues are resolved. Iterator COW behavior is now correct — read-only iteration doesn't trigger a copy, only `iterator.remove()` does. Test assertions verify behavioral outcomes instead of internal state. CI script scope drift removed. CI is green across Java 17, 21, and 25 with 22 comprehensive tests.

PR #23769, opened 2026-06-04
CAMEL-23686: Fix dev profile overriding user properties

gnodet:

## Summary

- Fix dev profile unconditionally overriding user-configured properties like `camel.main.messageHistory=false`
- `ProfileConfigurer` now checks `autoConfiguredProperties` to skip settings the user explicitly set
- `DefaultConfigurationConfigurer` applies `messageHistory` and `sourceLocationEnabled` unconditionally (not only when true)

## Problem

The dev profile in `ProfileConfigurer.configureCommon()` unconditionally set `messageHistory=true` and other settings, running AFTER user properties were loaded. This meant `camel.main.messageHistory=false` in `application.properties` had no effect.

With pooled exchanges, this caused `DefaultMessageHistory` instances to accumulate unbounded — 2.95M instances consuming 236MB in a 60s test, ballooning heap from 75MB (prod profile) to 696MB (dev profile).

## Changes

- **`ProfileConfigurer.java`**: `setIfNotConfigured()` helper checks `autoConfiguredProperties` before applying dev profile defaults
- **`BaseMainSupport.java`**: Pass `autoConfiguredProperties` to `ProfileConfigurer.configureMain()`
- **`DefaultConfigurationConfigurer.java`**: Change asymmetric if-guards (`if (config.isX()) { set(true); }`) to unconditional application (`set(config.isX())`)

## Test plan

- [x] `MainTest` (11 tests) — pass
- [x] `MainSecurityPolicyTest` (35 tests) — pass, including dev profile security policy tests
- [x] All camel-main tests (230 tests) — pass

### DIFF
--- a/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
+++ b/core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
@@ -617,7 +617,8 @@ public abstract class BaseMainSupport extends BaseService {
         // configure the profile with pre-configured settings
         step = recorder.beginStep(BaseMainSupport.class, "configureMain", "Profile Configure");
         doInitFileConfigurations(camelContext, mainConfigurationProperties);
-        ProfileConfigurer.configureMain(camelContext, mainConfigurationProperties.getProfile(), mainConfigurationProperties);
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

davsclaus:
yeah this is more intuitive, otherwise you can use --profile=prod, and then all those dev settings wont be forced, but this PR is better

PR #23771, opened 2026-06-04
CAMEL-23686: Lighten DefaultUnitOfWork route stack allocation

gnodet:

## Summary

- Replace `ConcurrentLinkedDeque<Route>` with a single `Route` field + lazily-created `ArrayDeque` for nested routes
- Saves ~72 bytes per exchange in the common single-route case

## Problem

Every `DefaultUnitOfWork` eagerly allocated a `ConcurrentLinkedDeque<Route>` (24 bytes + Node objects) even though most exchanges only traverse a single route. In a high-throughput pipeline processing ~1M exchanges/sec, this adds up to significant GC pressure.

## Changes

- `Route route` field holds the current/top route (common case: single route)
- `Deque<Route> routeStack` is lazily created only when `pushRoute()` is called while a route is already set (nested routes via `direct:`, kamelets, etc.)
- `pushRoute()` / `popRoute()` / `getRoute()` / `routeStackLevel()` updated accordingly
- `reset()` clears both fields
- Extracted `countRoute()` helper for the filtered `routeStackLevel(boolean, boolean)` variant

## Test plan

- [x] 43 UnitOfWork-related tests pass
- [x] 230 camel-main tests pass
- [x] Verified nested route scenarios still work (direct, kamelets, route templates)

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultUnitOfWork.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultUnitOfWork.java
@@ -16,13 +16,13 @@
  */
 package org.apache.camel.impl.engine;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

davsclaus (approved):
Clean, well-targeted optimization. The single `Route` field for the common case with lazy `ArrayDeque` for nested routes is the right design — saves ~72 bytes per exchange with zero behavioral change. The `ConcurrentLinkedDeque` → `ArrayDeque` switch is safe given Camel's single-thread-per-exchange model. Push/pop/level semantics are correct, `reset()` handles both fields, and CI is green.

PR #23779, opened 2026-06-05
CAMEL-23693: Optimize CaseInsensitiveMap hash for ASCII keys

gnodet:
[CAMEL-23693](https://issues.apache.org/jira/browse/CAMEL-23693)

## Summary

Optimize the `CaseInsensitiveMap` hash function and eliminate redundant hash computations introduced in CAMEL-23691.

- **ASCII fast-path in hash**: `c += 32` for `A-Z` instead of `Character.toLowerCase(Character.toUpperCase(c))` per character. Full Unicode two-step fold only for non-ASCII (`c >= 128`)
- **Single-hash `put()`**: compute hash once, reuse for deduplication lookup, find, and bucket insertion (was 3 separate hash computations)
- **Single-hash `remove()`**: inline chain walk with one hash computation (was 2)

### Benchmark (18 typical headers, 2M iterations)

| Scenario | Before | After | vs TreeMap (old impl) |
|---|---|---|---|
| put/get/containsKey/remove | 9.0M ops/s | **33.3M ops/s** | **1.2x faster** |
| Exchange copy (create+populate+copy) | 490K copies/s | **1.9M copies/s** | **1.1x faster** |
| Get-heavy (10 lookups/iter) | 16.9M lookups/s | **39.8M lookups/s** | **1.7x faster** |

## Test plan

- [x] `CaseInsensitiveMapTest` — 32 tests pass
- [x] `DefaultMessageHeaderTest` — 38 tests pass

### DIFF
--- a/core/camel-util/src/main/java/org/apache/camel/util/CaseInsensitiveMap.java
+++ b/core/camel-util/src/main/java/org/apache/camel/util/CaseInsensitiveMap.java
@@ -84,12 +84,12 @@ public class CaseInsensitiveMap extends AbstractMap<String, Object> implements S
         knownTable = tbl;
     }
 
-    private static String deduplicateKey(String key) {
+    private static String deduplicateKey(String key, int hash) {
         int[] tbl = knownTable;
         if (tbl == null) {
             return key;
         }
-        int idx = tbl[caseInsensitiveHash(key) & knownMask];
+        int idx = tbl[hash & knownMask];
         while (idx != EMPTY) {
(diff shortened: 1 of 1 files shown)


davsclaus (approved):
Clean performance follow-up to #23766. The ASCII fast-path (`c += 32` for A-Z) is correct and produces identical hash values as the previous `Character.toLowerCase(Character.toUpperCase(c))` for ASCII, with full Unicode fallback for non-ASCII. Single-hash `put()` and `remove()` eliminate redundant computations on the hot path. CI is green, existing tests cover correctness.

PR #23780, opened 2026-06-05
Fix transformer name normalization for catalog lookup during export

smongiar:
## Summary

- Normalize transformer names before catalog lookup in `DependencyDownloaderTransformerResolver`
- Transformer names use colon-separated format (`aws2-ddb:application-json`) but the catalog stores them in dash-separated format (`aws2-ddb-application-json`)
- The `normalize(key)` method from `TransformerResolver` interface already handles this conversion via `StringHelper.sanitize()` — it just wasn't being called before the catalog lookup
- Also normalize the name for the "did you mean" suggestion to produce accurate suggestions

## Reproducer

```bash
# Create a route using a kamelet with data type transformers
cat > DdbSinkRoute.java << 'ROUTE'
public class DdbSinkRoute extends RouteBuilder {
    @Override
    public void configure() {
        from("timer:tick?period=5000")
            .setBody(constant("{\"key\": \"value\"}"))
            .to("kamelet:aws-ddb-sink");
    }
}
ROUTE

# This fails with: Cannot find transformer with name: aws2-ddb:application-json.
# Did you mean: aws2-ddb-application-json
camel export --runtime=spring-boot --gav=com.example:test:1.0.0 --dir=./app DdbSinkRoute.java
```

## Test plan

- [x] Added `catalogLookupShouldNormalizeColonSeparatedNames()` test verifying catalog lookup works with colon-separated names after normalization
- [x] Existing `stubTransformerShouldHaveNameSet()` test still passes

### DIFF
--- a/dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/DependencyDownloaderTransformerResolver.java
+++ b/dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/download/DependencyDownloaderTransformerResolver.java
@@ -48,7 +48,9 @@ public final class DependencyDownloaderTransformerResolver extends DefaultTransf
     @Override
     public Transformer resolve(TransformerKey key, CamelContext context) {
         String name = key.toString();
-        TransformerModel model = catalog.transformerModel(name);
+        // catalog stores transformer names in normalized (dash-separated) format
+        String normalizedName = normalize(key);
+        TransformerModel model = catalog.transformerModel(normalizedName);
         if (model != null) {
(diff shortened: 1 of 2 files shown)


Croway (approved):

davsclaus (approved):
Clean bug fix — the `normalize()` method was already available but wasn't called before the catalog lookup. The fix is minimal and correct: normalize once, reuse for both catalog lookup and "did you mean" suggestion, while preserving the original name in the error message. Good test and clear reproducer in the PR body. CI is green.

PR #23781, opened 2026-06-05
Fix camel export to auto-detect kamelets version from classpath

smongiar:
## Summary

- The `--kamelets-version` option in `camel export` was hardcoded to a compile-time default from `parent/pom.xml` (`RuntimeType.KAMELETS_VERSION`)
- `VersionHelper.extractKameletsVersion()` can detect the actual kamelets jar version from the classpath, but the null-check at `resolveDependencies()` never triggers because the field is always pre-initialized
- Remove the compile-time default so `extractKameletsVersion()` runs at runtime, detecting the actual version from the classpath and falling back to the compiled-in default only when no kamelets jar is found
- Add the same null-check in `runSilently()` which runs before `resolveDependencies()`

## Problem

When running `camel export --runtime=spring-boot` without explicit `--kamelets-version`, the generated `pom.xml` always uses the version compiled into `RuntimeType.KAMELETS_VERSION`. If the actual `camel-kamelets` jar on the classpath has a different version (e.g. in downstream/productized builds), the exported project gets the wrong kamelets dependency.

## Reproducer

```bash
# Install a camel CLI with kamelets 4.18.x on the classpath
# Then export without --kamelets-version:
camel export --runtime=spring-boot --gav=com.example:test:1.0.0 --dir=./app MyRoute.java

# Before fix: pom.xml has <version>4.17.0</version> for camel-kamelets (compiled-in default)
# After fix: pom.xml has <version>4.18.x</version> (detected from classpath)
```


Croway:
After these changes, is `RuntimeType.KAMELETS_VERSION` needed? I think so, I found it in `CatalogKamelet` and `CatalogDoc` can you investigate if these classes need to be updated as well?

smongiar:
> After these changes, is `RuntimeType.KAMELETS_VERSION` needed? I think so, I found it in `CatalogKamelet` and `CatalogDoc` can you investigate if these classes need to be updated as well?

`RuntimeType.KAMELETS_VERSION` is still needed, it's used by VersionHelper.extractKameletsVersion() when classpath detection fails. But I found another issue:  `defaultValue = RuntimeType.KAMELETS_VERSION`  eagerly set the field, making the classpath detection code path dead. 

So, I updated the fix removing the defaultValue and field initializer, adding a null check that calls `VersionHelper.extractKameletsVersion()` which tries classpath first, then falls back to `RuntimeType.KAMELETS_VERSION`

Croway (approved):

oscerd (approved):

apupier (changes requested):
regen required.

davsclaus (approved):
Well-scoped fix. The compile-time default for `kameletsVersion` masked the runtime classpath detection, causing exported projects to always use the compiled-in version rather than the actual kamelets jar version. The fix consistently removes the pre-initialization across all 4 commands (`ExportBaseCommand`, `CatalogKamelet`, `CatalogDoc`, `KameletTools`) and lets `VersionHelper.extractKameletsVersion()` detect the real version at runtime. Documentation updated across 7 command reference pages.

davsclaus:
There are uncommitted changes
HEAD detached at pull/23781/merge
	modified:   dsl/camel-jbang/camel-jbang-core/src/generated/resources/META-INF/camel-jbang-commands-metadata.json

PR #23784, opened 2026-06-05
CAMEL-23686: Avoid creating empty headers map on read access

gnodet:

## Summary

- `getHeader()`, `removeHeader()`, and `removeHeaders()` no longer force creation of a `CaseInsensitiveMap` when headers haven't been set
- Short-circuits with `null` return when `headers == null` and `isPopulateHeadersSupported()` is `false`
- JMS-style messages with lazy header population (`isPopulateHeadersSupported() == true`) are preserved

## Problem

Every `getHeader(name)` call on a message with no headers forced creation of a `CaseInsensitiveMap` (allocating 4 internal arrays, ~200 bytes) just to return `null`. This happened frequently in splitter/multicast sub-exchanges that never use headers.

Before:
```java
public Object getHeader(String name) {
    if (headers == null) {
        headers = createHeaders();  // allocates empty CaseInsensitiveMap
    }
    if (!headers.isEmpty()) {       // always false for new map
        return headers.get(name);
    } else {
        return null;                // returns null anyway
    }
}
```

After:
```java
public Object getHeader(String name) {
    if (headers == null) {
        if (!isPopulateHeadersSupported()) {
            return null;            // no allocation
        }
        headers = createHeaders();
    }
    return headers.get(name);
}
```

## Benchmark results

Baseline route (timer → split(1000) → setBody, no header access):
- 3,891 `DefaultMessage` instances but only **4 `CaseInsensitiveMap`** — split children don't allocate maps at all
- Without this fix: every sub-exchange would create an empty map on first `getHeader()` call

Pipeline route (with `setHeader` on every exchange):
- **24% throughput increase** measured by in-flight exchange count at snapshot time
- Map ratio stays at 0.57 (expected — every exchange writes headers)

## Test plan

- [x] `DefaultMessageHeaderTest` (38 tests) — pass, including lazy-populated headers test
- [x] `DefaultMessageTest` — pass

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/DefaultMessage.java
@@ -64,70 +64,37 @@ public class DefaultMessage extends MessageSupport {
     @Override
     public Object getHeader(String name) {
         if (headers == null) {
-            // force creating headers (supports lazy population in subclasses like JmsMessage)
+            if (!isPopulateHeadersSupported()) {
+                return null;
+            }
             headers = createHeaders();
         }
-
-        if (!headers.isEmpty()) {
-            return headers.get(name);
-        } else {
-            return null;
-        }
+        return headers.get(name);
     }
 
     @Override
     public Object getHeader(String name, Object defaultValue) {
-        Object answer = null;
-
-        if (headers == null) {
-            // force creating headers
-            headers = createHeaders();
-        }
-
-        if (!headers.isEmpty()) {
-            answer = headers.get(name);
-        }
+        Object answer = getHeader(name);
         return answer != null ? answer : defaultValue;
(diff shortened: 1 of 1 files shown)


apupier (approved):
astonishing improvements due to only empty Map instanciation removal

EDIT: I think it will be worthy to wait for Claus or Andrea feedback before merging as they might better know if there are corner cases

davsclaus (approved):
Clean optimization — eliminates unnecessary `CaseInsensitiveMap` allocation (~200 bytes) on read access when headers haven't been set. The `isPopulateHeadersSupported()` guard correctly preserves the JMS-style lazy population path. The DRY refactor of `getHeader` overloads delegating to the base method is a nice simplification (-37 net lines). CI is green.

Croway (approved):

PR #23796, opened 2026-06-05
CAMEL-23686: Replace ConcurrentHashMap with FlatMap for exchange properties

gnodet:

## Summary

- Replace `ConcurrentHashMap` with `FlatMap` (flat `Object[]` array) for exchange properties
- Make properties lazy in `DefaultPooledExchange` (was eagerly created in all constructors)

## Why ConcurrentHashMap is not needed

The `ConcurrentHashMap` was introduced in [CAMEL-715](https://issues.apache.org/jira/browse/CAMEL-715) (2008) to fix a `ConcurrentModificationException` in the old `thread()` DSL where one thread modified properties while another iterated them.

The test that reproduced the issue (`Camel715ThreadProcessorTest`) even says in its javadoc:
> *"An old unit test from CAMEL-715 which reproduced a problem which we don't have anymore in Camel threads EIP and the routing engine."*

The modern routing engine ensures single-threaded access per exchange. Message headers (`CaseInsensitiveMap`) are already explicitly documented as not thread-safe. Exchange properties don't need stronger guarantees.

## FlatMap design

A `Map<String, Object>` backed by a flat `Object[]` with alternating `{key, value, key, value, ...}` pairs:
- For ≤8 entries: faster than HashMap (linear scan beats hash + pointer chasing, single cache line)
- Zero per-entry object overhead (no `Node` objects)
- Default capacity: 4 entries (one `Object[8]` array = ~48 bytes vs ConcurrentHashMap's ~200+ bytes)
- Grows the array on demand for rare cases with many properties

## Benchmark expectations

- Typical exchange has 2-5 user properties (timer sets 4: COUNTER, NAME, PERIOD, FIRED_TIME)
- 325K `ConcurrentHashMap` instances at 64 bytes each = 20 MB → replaced by `FlatMap` at ~48 bytes = ~15 MB
- Pooled exchanges that never use properties: 0 bytes (was 200+ bytes from eager `ConcurrentHashMap`)

## Test plan

- [x] 7,127 camel-core tests pass (0 failures)
- [x] 370 exchange/property-related tests pass

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/AbstractExchange.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/AbstractExchange.java
@@ -22,7 +22,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.apache.camel.AsyncCallback;
@@ -285,7 +284,7 @@ abstract class AbstractExchange implements Exchange, ExchangeExtension {
         } else if (value != null) {
             // avoid the NullPointException
             if (properties == null) {
-                this.properties = new ConcurrentHashMap<>(8);
+                this.properties = new FlatMap<>(4);
             }
             properties.put(name, value);
         } else if (properties != null) {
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Well-designed optimization for the common case of small exchange property maps. `FlatMap` with alternating key/value pairs in a flat `Object[]` beats `ConcurrentHashMap` for the typical 2-5 property case via cache locality and zero per-entry overhead. The thread-safety justification is sound — `ConcurrentHashMap` dates from CAMEL-715 (2008) and the modern routing engine ensures single-threaded access. Lazy properties in `DefaultPooledExchange` saves ~200 bytes per pooled exchange. Iterator remove correctly handles the swap-with-last pattern.

Minor: `remove()` uses swap-with-last which changes insertion order — exchange properties have no ordering guarantee so this is fine, but worth a note in the class Javadoc.

PR #23797, opened 2026-06-05
CAMEL-23698: Do not append trailing ? to URI when query string is empty

davsclaus:
## Summary

- Fix `URISupport.createURIWithQuery()` to not append a bare `?` when the query string is empty
- When `camel-rest-openapi` produces a REST call from an OpenAPI spec with optional query parameters that are not set, the URL ended with a trailing `?` (e.g., `/users/myuser/repos?` instead of `/users/myuser/repos`)
- Some HTTP servers and proxies treat `/resource` and `/resource?` differently, so this can cause issues

## Root Cause

`URISupport.createURIWithQuery()` checked `if (query != null)` before appending `?` + query, but did not check if the query string was empty. Changed to `if (query != null && !query.isEmpty())`.

## Test plan

- [x] Added `testCreateURIWithQueryEmptyQueryString` in `URISupportTest` verifying no trailing `?` is appended for empty query strings
- [x] Existing `URISupportTest` tests pass (null query and non-empty query cases)

### DIFF
--- a/core/camel-util/src/main/java/org/apache/camel/util/URISupport.java
+++ b/core/camel-util/src/main/java/org/apache/camel/util/URISupport.java
@@ -443,7 +443,7 @@ public final class URISupport {
         if (before != null) {
             s = before;
         }
-        if (query != null) {
+        if (query != null && !query.isEmpty()) {
             s = s + "?" + query;
         }
         if (!s.contains("#") && uri.getFragment() != null) {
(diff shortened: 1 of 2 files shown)


davsclaus (commented):
Reviewed — looks good to merge.

- The `!query.isEmpty()` guard on `createURIWithQuery()` is the right fix. The original null-only check (from 2011) never handled empty strings, but `camel-rest-openapi` can produce them when all optional query params are unset.
- Checked all 6 production callers (`URISupport`, `ResourceHelper`, `HttpProducer`, `HttpHelper`, `NettyHttpHelper`, `UndertowHelper`) — none rely on appending a bare `?` for empty queries. They all benefit from this fix.
- Test coverage is appropriate: the new test complements the existing null and non-empty cases.
- No regression risk.

PR #23813, opened 2026-06-07
CAMEL-23680: Add group as scope to variables

davsclaus:
## Summary

- Adds a new built-in `group` variable scope alongside the existing `exchange`, `route`, and `global` scopes
- Group variables use the syntax `group:groupId:variableName` (e.g., `group:teamA:threshold`)
- Groups are user-managed by convention — the group name is picked in the variable key, no configuration needed
- Supports initial variable configuration via `application.properties`: `camel.variable.group.teamA.xxx = value`

## Changes

- **`VariableRepositoryFactory`** — added `GROUP_VARIABLE_REPOSITORY_ID` constant
- **`GroupVariableRepository`** (new) — in-memory group-scoped variable repository modeled on `RouteVariableRepository`, with `getGroupIds()` API and wildcard removal (`group:teamA:*`)
- **`DefaultVariableRepositoryFactory`** — registers group repo as a built-in alongside global and route
- **`MainSupportModelConfigurer`** — supports `camel.variable.group.` prefix in properties initialization
- **`variables.adoc`** — documented group scope in repository list, Java API examples, and properties configuration

CLI (`camel get variable`), developer console, JMX, and TUI pick up the group repository automatically via `BrowsableVariableRepository` — no changes needed in those modules.

## Test plan

- [x] `GroupVariableRepositoryTest` — 14 unit tests covering set/get, isolation, wildcards, `getGroupIds()`, error cases
- [x] `SetGroupVariableTest` — 4 integration tests covering route DSL, group isolation, Simple language access, cross-route sharing

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/GroupVariableRepository.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/GroupVariableRepository.java
@@ -1,0 +1,215 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.support;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.CamelContextAware;
+import org.apache.camel.StreamCache;
+import org.apache.camel.StreamCacheException;
+import org.apache.camel.spi.BrowsableVariableRepository;
+import org.apache.camel.spi.StreamCachingStrategy;
+import org.apache.camel.spi.VariableRepository;
+import org.apache.camel.support.service.ServiceSupport;
+import org.apache.camel.util.StringHelper;
+
+/**
+ * Group {@link VariableRepository} which stores variables in-memory per named group.
+ * <p>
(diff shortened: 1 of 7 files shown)


core/camel-support/src/main/java/org/apache/camel/support/GroupVariableRepository.java
```diff
@@ -0,0 +1,215 @@
...
+        } else {
+            // if the value is null, we just remove the key from the map
+            Map<String, Object> variables = groups.get(id);
+            if (variables != null) {
+                variables.remove(key);
+            }
+        }
+    }
+
+    public boolean hasVariables() {
```
davsclaus, line 107: Minor observation: when `value` is `null`, the key is removed from the inner map, but if the inner map becomes empty, the group entry remains in the `groups` outer map. This means `getGroupIds()` could return group names with zero variables.

This is the same behavior as `RouteVariableRepository`, so it's consistent — just noting it since `getGroupIds()` is a new user-facing API.

davsclaus (commented):
## Review: CAMEL-23680 — Add group as scope to variables

Clean, well-structured addition that follows the established `RouteVariableRepository` pattern closely. CI is green, test coverage is solid (14 unit + 4 integration tests).

### Findings

**1. Missing upgrade guide entry (Medium)**
Per project conventions, user-visible changes should be documented in the upgrade guide. This PR adds a new built-in variable scope (`group`) and reserves the id `"group"` for Camel's internal use. Anyone who happened to have a custom `VariableRepository` with id `"group"` would be affected since it now gets intercepted by the factory. An entry in `camel-4x-upgrade-guide-4_21.adoc` should mention:
- New `group` variable scope is available
- The id `group` is now reserved (along with the already-reserved `global`, `route`, `exchange`, `header`)

**2. Stale empty group maps after individual variable removal (Low — cosmetic, consistent with existing code)**
In `GroupVariableRepository.setVariable()`, setting a value to `null` removes the key from the inner map but does not remove the outer group entry when the inner map becomes empty. This means `getGroupIds()` could return group names that have no variables. This is the exact same behavior as `RouteVariableRepository`, so it's consistent and not a regression — just noting it since `getGroupIds()` is a new API and users may expect it to only return groups that actually have variables.

**3. No automatic lifecycle cleanup for groups (Question — likely intentional)**
`RouteVariableRepository` has a `LifecycleStrategySupport` that automatically removes route variables when routes are removed. Groups have no equivalent cleanup. This seems intentional since groups are user-managed by convention, but wanted to confirm.

### What looks good

- Implementation follows the exact same pattern as `RouteVariableRepository` — thread-safe, proper `StreamCache` handling
- `@since 4.21` Javadoc tags on new public API
- Documentation is clear with Java API and `application.properties` examples
- Group repo auto-registers via `BrowsableVariableRepository`, so CLI/console/JMX picks it up for free

This review does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

PR #23812, opened 2026-06-07
CAMEL-23688: Add extensive test coverage for camel-jbang CLI commands

ammachado:
# Description

Adds test coverage for `camel-jbang-core` CLI commands to prepare the codebase for a refactoring pass ([CAMEL-23688](https://issues.apache.org/jira/browse/CAMEL-23688)).

Before this PR, `commands/process/` had zero tests and `commands/action/` had only 2 test classes for 38+ command files.

**Approach for process commands** (`ListProcess`, `CamelRouteStatus`, `ListEndpoint`, etc.): mock the JDK static methods `ProcessHandle.allProcesses()` and `ProcessHandle.current()` with Mockito 5's `mockStatic`, then write hand-crafted `<pid>-status.json` fixtures via `CommandLineHelper.useHomeDir()`. This exercises the full table-rendering, sorting, filtering, and JSON-output paths without a running Camel integration.

**Approach for action commands**: pure unit tests against public/package-private utility methods, and an inner test-subclass pattern (already established in this module by `CamelRouteDiagramActionTest`) that overrides `findPids()` and `waitForOutputFile()` to inject a prepared JSON response.

**New files:**

`commands/` package:
- `CamelCommandHelperTest` — phase integer to human-readable status string mapping

`commands/action/` package:
- `CamelStartupRecorderActionTest` — step display, name parentheses, sort by duration, empty-pid error
- `LoggerActionTest` — logger list display, sort by context name across two processes
- `MessageTableHelperTest` — body/header visibility toggle, null root, exchange ID in output
- `CamelSourceActionTest` — `extractSourceName()` edge cases (null, plain file, scheme, path, multi-colon)

`commands/process/` package:
- `ProcessCommandTestSupport` — shared base: MockedStatic wiring, status-file helpers, JSON builders
- `CamelContextStatusTest` — Running/Starting/Stopped phase display, JSON output
- `CamelRouteStatusTest` — single/multiple routes, suspended route, filter, limit, JSON output
- `ListProcessTest` — table display, sort by pid/name/age, name filter, reload-error display, JSON output
- `ListEndpointTest` — URI display, direction filter (from/to), JSON output
- `ListHealthTest` — UP/DOWN health check display, readiness-only filter, JSON output
- `ListRestTest` — REST endpoint URL/method display, JSON output
- `ListConsumerTest` — consumer URI/state display, JSON output
- `ListErrorTest` — error display from `<pid>-error.json`, routeId, JSON output
[...]

PR #23822, opened 2026-06-07
CAMEL-23707: Fix row-mutation bug in ListVault when multiple vault types are configured

ammachado:
# Description

`ListVault.doProcessWatchCall()` builds its row list by mutating a single shared `Row` object across all vault-type blocks. The `row.copy()` call that isolates each row is guarded by `i > 0`, which means it only fires for the 2nd and subsequent secrets within a single vault. When a vault has exactly one secret, the row is added to the list while `row` still points to the same object. Every subsequent vault block then mutates that already-added object (`row.vault = "GCP"`, `row.vault = "Azure"`, etc.), so all rendered rows end up carrying the label of the last vault type processed. The same corruption surfaces in `--output json` mode, because the bug occurs during row collection before the `jsonOutput` branch is reached.

Fix: save a `baseRow` (pid + name only) before vault processing begins and reset `row = baseRow.copy()` at the start of each vault block. This ensures no already-added `Row` object is ever mutated by a subsequent vault block, and also prevents stale fields such as `region` from bleeding between vault types.

`ListVaultTest` and `ProcessCommandTestSupport` are ported from the companion PR #23812 (CAMEL-23688) to cover the fix. The test class includes regression tests for both table and JSON output with multiple vault types each having a single secret.

# Target

# Tracking

- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-23707) filed for the change (usually before you start working on it).

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-core/pom.xml
@@ -186,6 +186,12 @@
             <version>${wiremock-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/ListVault.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/process/ListVault.java
@@ -87,10 +87,12 @@ public class ListVault extends ProcessWatchCommand {
                         if ("CamelJBang".equals(row.name)) {
                             row.name = ProcessHelper.extractName(root, ph);
                         }
+                        final Row baseRow = row.copy();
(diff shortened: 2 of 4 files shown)


davsclaus (commented):
Nice catch on the row-mutation bug — the fix is correct and well-tested.

The `baseRow.copy()` approach properly isolates each vault block so that previously-added `Row` objects are never mutated by subsequent vault types. The test coverage is thorough (empty output, individual vault types, the multi-vault single-secret regression case, multi-secret within a vault, and JSON output).

**One coordination concern:** both this PR and the companion PR #23812 (CAMEL-23688) add the exact same two new files (`ListVaultTest.java` and `ProcessCommandTestSupport.java`). Whichever merges second will hit file-creation conflicts. Which PR is intended to merge first? If this bug-fix PR lands first, #23812 should rebase and deduplicate. If #23812 lands first, this PR will need rebasing.

ammachado:
I created separate PRs to allow the maintainers choose which to merge first. This is smaller and contains a bug fix, so I think that it should be merged first.

I'll rebase the other PR if this one is merged first. #23812 needs review anyway, to remove the `@Disabled` annotations.

The maintainers could consider enabling [GitHub Stacked PRs](https://github.github.com/gh-stack/) on this repository when available.

davsclaus:
okay so lets use this PR as is

davsclaus (approved):

PR #23886, opened 2026-06-09
CAMEL-23630: add HeaderFilterStrategy to camel-dapr

oscerd:
Implements [CAMEL-23630](https://issues.apache.org/jira/browse/CAMEL-23630).

## What this changes

- Adds a `DaprHeaderFilterStrategy` (extending `DefaultHeaderFilterStrategy`) and exposes it via `DaprComponent` / `DaprEndpoint` as the standard `headerFilterStrategy` URI option, default-instantiated. This brings camel-dapr into line with the rest of the messaging-component family (`camel-iggy`, `camel-kafka`, `camel-jms`, ...).
- Updates `DaprPubSubConsumer` to stop copying the inbound CloudEvent's `pubsubName` / `topic` into the `CamelDaprPubSubName` / `CamelDaprTopic` headers (this is CAMEL-23630 "option (b)"). Those two constants are **producer-direction routing headers**: `DaprConfigurationOptionsProxy#getPubSubName/getTopic` read them back and prefer them over the endpoint-configured destination, so carrying them out of a consumer made a route such as `from("dapr-pubsub:p:t").to("dapr-pubsub:p:other")` route to the **inbound** topic instead of the configured one. The remaining CloudEvent metadata headers (`CamelDaprID`, `CamelDaprSource`, `CamelDaprType`, `CamelDaprSpecificVersion`, `CamelDaprDataContentType`, `CamelDaprBinaryData`, `CamelDaprTime`, `CamelDaprTraceParent`, `CamelDaprTraceState`) are unchanged.
- Updates `DaprPubSubConsumerTest` to assert the two routing headers are no longer set on the consumed exchange (the remaining metadata headers are still asserted).
- Documents the user-visible consumer header change and the new strategy in the 4.21 upgrade guide.

## Design note

Following the `camel-iggy` reference implementation cited in the issue, the component's own `CamelDapr*` metadata headers are set **directly** (not run through the filter). The `HeaderFilterStrategy` is intended to filter an arbitrary, sender-controlled header map — which the Dapr SDK `CloudEvent<T>` POJO does not expose (it has a fixed, closed set of getters and no extension attributes). It is therefore added for consistency with the catalog and to provide the standard `headerFilterStrategy` knob, and is not invoked over the consumer's fixed-field mapping. See the review thread for the full rationale.

---

*The component implementation was done using the Gemini CLI and the Gemini 2.5 Pro model; the upgrade-guide note and review responses were prepared by Claude Code on behalf of Andrea Cosentino.*


components/camel-dapr/src/main/java/org/apache/camel/component/dapr/DaprHeaderFilterStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Default header filter strategy for Dapr endpoints.
+ */
+public class DaprHeaderFilterStrategy extends DefaultHeaderFilterStrategy {
+
+    public DaprHeaderFilterStrategy() {
+        setLowerCase(true);
+        setOutFilterStartsWith(CAMEL_FILTER_STARTS_WITH);
+        setInFilterStartsWith(CAMEL_FILTER_STARTS_WITH);
+    }
```
davsclaus, line 30: This filter uses `CAMEL_FILTER_STARTS_WITH` which blocks all headers starting with `Camel` / `camel`. Since every Dapr component header starts with `CamelDapr` (e.g., `CamelDaprID`, `CamelDaprSource`, `CamelDaprTopic`), this filter would block **all** of Dapr's own headers if it were actually applied in the consumer/producer.

Compare with `KafkaHeaderFilterStrategy` — Kafka's component headers don't use the `Camel` prefix, so the generic filter works. For Dapr, you'd need to either:
- Change Dapr's header naming to not use the `Camel` prefix, or
- Override `applyFilterToExternalHeaders()` to allow `CamelDapr*` headers through

components/camel-dapr/src/main/java/org/apache/camel/component/dapr/consumer/DaprPubSubConsumer.java
```diff
@@ -102,8 +102,6 @@ private Exchange createServiceBusExchange(final CloudEvent<byte[]> cloudEvent) {
         message.setBody(cloudEvent.getData());
 
         // set headers
-        message.setHeader(DaprConstants.PUBSUB_NAME, cloudEvent.getPubsubName());
-        message.setHeader(DaprConstants.TOPIC, cloudEvent.getTopic());
         message.setHeader(DaprConstants.ID, cloudEvent.getId());
```
davsclaus, line 105: These headers (`CamelDaprPubSubName` and `CamelDaprTopic`) were removed because the filter strategy would block them. But the consumer still sets `CamelDaprID`, `CamelDaprSource`, `CamelDaprType`, etc. — all of which would also be blocked by the same filter. The fix should be in the filter strategy (allow Dapr's own headers), not removing useful headers.

Users consuming from Dapr PubSub need to know which pubsub and topic a message came from.

davsclaus (changes requested):
> :robot: _Claude Code on behalf of Claus Ibsen_

Thanks for working on this Andrea — adding `HeaderFilterStrategy` to camel-dapr is the right direction. I found a few issues that need to be addressed before this can be merged.

Note: This review covers project conventions and correctness. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

---

### 1. (Critical) HeaderFilterStrategy is wired up but never used in message processing

`DaprPubSubConsumer.createServiceBusExchange()` sets headers directly via `message.setHeader()` without ever calling `headerFilterStrategy.applyFilterToExternalHeaders()`. The strategy is added to `DaprEndpoint` and `DaprComponent` but is dead code — it has no effect on actual message processing.

Sibling components actively use the strategy. For example, `Sqs2Consumer` (~line 275):
```java
HeaderFilterStrategy headerFilterStrategy = getEndpoint().getHeaderFilterStrategy();
if (!headerFilterStrategy.applyFilterToExternalHeaders(header, value, exchange)) {
    message.setHeader(header, value);
}
```

And Kafka's `KafkaRecordProcessor` calls `shouldBeFiltered()` on every header.

### 2. (Critical) Default DaprHeaderFilterStrategy would block ALL component headers

`DaprHeaderFilterStrategy` sets both `setInFilterStartsWith(CAMEL_FILTER_STARTS_WITH)` and `setOutFilterStartsWith(CAMEL_FILTER_STARTS_WITH)`, which filters headers starting with `Camel` / `camel`. But **every** Dapr header uses the `CamelDapr` prefix (`CamelDaprPubSubName`, `CamelDaprTopic`, `CamelDaprID`, `CamelDaprSource`, etc.).

If the filter were actually applied (fixing issue #1), it would filter out **all** of the component's own headers — including the ones the PR keeps (`CamelDaprID`, `CamelDaprSource`, `CamelDaprType`, etc.).

Compare with Kafka: Kafka component headers don't use the `Camel` prefix, so the `CAMEL_FILTER_STARTS_WITH` filter works correctly there. Dapr's header naming convention (`CamelDapr*`) conflicts with the generic `Camel*` filter.

### 3. (Major) Useful headers removed as workaround

The PR removes `PUBSUB_NAME` (`CamelDaprPubSubName`) and `TOPIC` (`CamelDaprTopic`) from being set in `DaprPubSubConsumer`. The CI fix section acknowledges this was because "the HeaderFilterStrategy applied to standard CamelDapr prefixed headers filtered them out."

But this is inconsistent — the consumer **still** sets `CamelDaprID`, `CamelDaprSource`, `CamelDaprType`, `CamelDaprSpecificVersion`, `CamelDaprDataContentType`, `CamelDaprBinaryData`, `CamelDaprTime`, `CamelDaprTraceParent`, `CamelDaprTraceState`. If the filter should block PubSubName and Topic, why not these?

Users consuming from Dapr PubSub likely need to know which pubsub and topic a message came from. Removing these headers breaks that use case.

### 4. (Minor) No producer-side header filtering

The PR adds the filter infrastructure but only touches the consumer side. The producer should also use the strategy when mapping Camel headers to outbound Dapr operations.

---

### Recommendation

The approach needs rework: either the filter strategy should be actively used in the consumer/producer (with proper allow-listing of Dapr's own `CamelDapr*` headers so they pass through), or the scope should be reconsidered.

oscerd:
> :robot: _Claude Code on behalf of Andrea Cosentino_

Thanks Claus for the careful review. I went through each point against CAMEL-23630, the camel-iggy reference implementation it mirrors, and the Dapr SDK, and I'd like to keep the current approach. Reasoning point by point:

**Context — CAMEL-23630 deliberately chose "option (b)".** The issue offers two alternatives for the consumer: (a) consult the `HeaderFilterStrategy` before setting each header, or (b) stop copying `PUBSUB_NAME`/`TOPIC` because they are producer-direction routing headers. This PR implements (b). The strategy is added for **consistency with the messaging-component family** and to expose the standard `headerFilterStrategy` option — exactly as camel-iggy (the cited reference) does.

**1. "Filter is never applied in the consumer."** Correct, and intentional. The siblings apply `applyFilterToExternalHeaders()` to an **arbitrary, sender-controlled header map**: `Sqs2Consumer` filters SQS message attributes, Kafka filters record headers, iggy filters `message.userHeaders()`. Note iggy sets its **own** fixed metadata headers (`MESSAGE_ID`, `MESSAGE_OFFSET`, …) **directly, bypassing the filter**, and runs only `userHeaders()` through it. The Dapr SDK `CloudEvent<T>` is a fixed POJO (`getId/getSource/getType/getPubsubName/...`) with **no extension/attribute map** (checked against `dapr-sdk` 1.16) — so there is no external header collection to filter, and every header name the consumer sets is a hardcoded, trusted `CamelDapr*` constant. There is no header-injection vector here to filter.

**2. "Default strategy would block all `CamelDapr*` headers if applied."** Agreed — and that's exactly why CAMEL-23630 chose (b) over (a). Running the component's own `CamelDapr`-prefixed metadata through a `Camel*` filter would strip them, so neither this PR nor iggy does that; the component's own headers are always set directly.

**3. "PubSubName/Topic removed but ID/Source/Type kept — inconsistent."** This asymmetry is the deliberate fix. `CamelDaprPubSubName`/`CamelDaprTopic` are **read back on the producer side** by `DaprConfigurationOptionsProxy#getPubSubName/getTopic` (header preferred over endpoint config; consumed by `DaprPubSubHandler`). Leaving them on a consumed exchange makes `from("dapr-pubsub:p:t").to("dapr-pubsub:p:other")` route to the **inbound** pubsub/topic instead of the configured one. `ID/Source/Type/...` are never read back by any producer operation, so they stay. A `dapr-pubsub` consumer also subscribes to a single fixed pubsub+topic, so those two headers were redundant with the endpoint config.

**4. "No producer-side filtering."** `DaprPubSubHandler#handle` reads only specific, typed headers via `getOption(...)` and builds the publish request from an explicit, route-author-supplied `CamelDaprMetadata` map — there is no generic "copy all Camel headers outbound" loop for an out-filter to act on. The out-filter is still exposed via the standard option for parity with the catalog.

I've pushed a follow-up documenting the consumer header change in the 4.21 upgrade guide. Happy to revisit if you'd prefer a different shape (e.g. exposing pubsub/topic under non-routing header names), but that would be a follow-up beyond this issue's scope.

components/camel-dapr/src/main/java/org/apache/camel/component/dapr/DaprHeaderFilterStrategy.java
oscerd: > :robot: _Claude Code on behalf of Andrea Cosentino_

Right — and that's why CAMEL-23630 stops copying the two routing headers (option b) rather than applying the filter per-header (option a). This mirrors `IggyHeaderFilterStrategy` and is exposed for consistency + the standard `headerFilterStrategy` option. It's never run over the component's own `CamelDapr*` metadata (those are set directly), so it won't strip them. A `Camel*` filter would only act on an arbitrary sender-controlled header map — which the Dapr `CloudEvent<T>` POJO doesn't have (no extension attributes), so there's nothing external to filter in the consumer. Overriding `applyFilterToExternalHeaders()` to allow `CamelDapr*` through isn't needed, since we never pass those (or any external map) through the filter.

components/camel-dapr/src/main/java/org/apache/camel/component/dapr/consumer/DaprPubSubConsumer.java
oscerd: > :robot: _Claude Code on behalf of Andrea Cosentino_

The asymmetry is intentional. `CamelDaprPubSubName`/`CamelDaprTopic` are read back on the **producer** side by `DaprConfigurationOptionsProxy#getPubSubName/getTopic` (header preferred over endpoint config, consumed by `DaprPubSubHandler`), so carrying them out of a consumer makes `from(dapr-pubsub).to(dapr-pubsub)` override the configured destination — the exact carry-over CAMEL-23630 targets. `CamelDaprID/Source/Type/...` are never read on the producer side, so they're harmless and stay. On knowing the source pubsub/topic: a `dapr-pubsub` consumer subscribes to a single fixed `pubSubName:topic`, so the value always equals the endpoint URI you configured — the header was redundant for the consumer. Added a 4.21 upgrade-guide note documenting the removal.

davsclaus (approved):
LGTM — clean, well-justified PR.

**Positive observations:**

1. **Bug fix is correct:** `PUBSUB_NAME` and `TOPIC` were producer-direction routing headers (read by `DaprConfigurationOptionsProxy`) that shouldn't have been set on consumed exchanges. Removing them fixes the routing loop described in CAMEL-23630.

2. **Security:** `DaprHeaderFilterStrategy` uses `setLowerCase(true)` with `CAMEL_FILTER_STARTS_WITH`, providing case-insensitive filtering of `Camel*` / `camel*` headers in both directions.

3. **Standard pattern:** `DaprEndpoint` correctly implements `HeaderFilterStrategyAware`, the strategy is exposed as a `@UriParam(label = "advanced")` option on both component and endpoint, and lazy initialization in `getHeaderFilterStrategy()` provides a sensible default. Follows the `camel-iggy` reference implementation.

4. **Test updated:** `DaprPubSubConsumerTest` asserts the two routing headers are now `null` on the consumed exchange.

5. **Upgrade guide:** Thorough entry in `camel-4x-upgrade-guide-4_21.adoc` with code example showing the routing problem and migration guidance.

PR #23902, opened 2026-06-10
CAMEL-19540: camel-pulsar - replace Thread.sleep in tests

anvithsg2004:
# Description

Replaces the only `Thread.sleep` in the camel-pulsar test tree with deterministic synchronization (CAMEL-19540).

`PulsarSharedSubscriptionMessageDistributionIT` used a fixed `Thread.sleep(1000)` inside the shared route `Processor` as a pacing delay: it slowed route 1 down so the late-started route 2 could join the Shared subscription while messages were still pending. A fixed sleep is timing-dependent — too short on loaded CI machines (flaky failures), wasted time on fast ones.

The sleep is replaced with a bounded Awaitility wait inside the processor (as suggested in review — thanks @oscerd):

```java
await().atMost(5, TimeUnit.SECONDS).until(() -> to2.getReceivedCounter() > 0);
```

- Route 1's processing threads hold each exchange until the late-starting consumer (route 2) has demonstrably received a message, so route 1 cannot drain the topic before route 2 comes online. Once route 2 receives its first message, everything proceeds immediately — no fixed delay left.
- Route 2 cannot block itself: its exchanges reach the `mock:result2` endpoint (incrementing the received counter) before the route advances to the polling processor, so the condition is already true.
- The 5s bound is deliberately well below the consumer's default `ackTimeoutMillis` (10s), so a held (unacknowledged) message cannot hit ack-timeout redelivery mid-test. The old 1s sleep was likewise safely below that boundary.
- On timeout the processor logs a warning and lets the exchange complete; the existing `MockEndpoint.assertIsSatisfied(10, TimeUnit.SECONDS, ...)` assertions remain the failure gate.

Mechanically: with `consumerQueueSize=1` and `numberOfConsumerThreads=2`, route 1 pins at most ~2 unacked in-flight messages while waiting, so most of the 10 produced messages remain at the broker for route 2 — the hand-off the sleep previously approximated by timing.

# Testing

- `./mvnw install -Psourcecheck -pl components/camel-pulsar` — BUILD SUCCESS; 23/23 unit tests pass; formatter/impsort source checks pass.
- `grep -rn "Thread.sleep" components/camel-pulsar/src/test` — no hits remain.
- The changed test is a Testcontainers-based IT; it could not be executed in the local dev environment (no Docker daemon, and camel-pulsar sets `skipITs.aarch64=true` so Pulsar ITs are skipped on arm64 hosts by design). CI on amd64 exercises it.

_Claude Code on behalf of Anvith SG_

### DIFF
--- a/components/camel-pulsar/src/test/java/org/apache/camel/component/pulsar/integration/PulsarSharedSubscriptionMessageDistributionIT.java
+++ b/components/camel-pulsar/src/test/java/org/apache/camel/component/pulsar/integration/PulsarSharedSubscriptionMessageDistributionIT.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.component.pulsar.integration;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.Endpoint;
@@ -59,6 +60,8 @@ public class PulsarSharedSubscriptionMessageDistributionIT extends PulsarITSuppo
     @EndpointInject("mock:result2")
     private MockEndpoint to2;
 
+    private final CountDownLatch lateConsumerReceived = new CountDownLatch(1);
+
     @Override
     protected RouteBuilder createRouteBuilder() {
         return new RouteBuilder() {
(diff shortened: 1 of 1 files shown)


oscerd:
You can use awaitility instead of a custom mechanism.

anvithsg2004:
Done — switched to Awaitility in c9a0ace6aa2: the processor now polls the late consumer's mock received counter (`await().atMost(5, SECONDS).until(() -> to2.getReceivedCounter() > 0)`) instead of the custom `CountDownLatch` + mock-callback wiring. The 5s bound stays below the default 10s `ackTimeoutMillis` so a held unacked message can't hit ack-timeout redelivery during the wait. Module build with `-Psourcecheck` is green (23/23 unit tests).

_Claude Code on behalf of Anvith SG_

davsclaus (approved):

oscerd (approved):

anvithsg2004:
The JDK 17 job failure is unrelated to this change — the failing test is `PulsarConsumerNoAcknowledgementIT#testAMessageIsConsumedMultipleTimesWithAckTimeoutBackoff`, while this PR only touches `PulsarSharedSubscriptionMessageDistributionIT` (different topic and subscription).

It looks pre-existing and inherently racy: both test methods in that class share the same topic and subscription, and by design nothing acknowledges the messages. The first method's unacked message can be redelivered into the second method's counting window, producing a 4th delivery where `expectedMessageCount(3)` demands exactly 3. The failed assertion then skips `producer.close()`, so the automatic rerun hits `ProducerBusyException` on the fixed producer name `camel-producer-1`.

Happy to file a separate JIRA (and a follow-up PR) for that flaky test. In the meantime, could a committer re-run the failed job?

_Claude Code on behalf of Anvith SG_

anvithsg2004:
Pushed an empty commit to retrigger CI, but the workflow runs for the new head are awaiting maintainer approval (fork PR policy). Could someone hit "Approve and run" on the pending workflows — or alternatively just re-run the failed JDK 17 job, whichever is easier?

_Claude Code on behalf of Anvith SG_

PR #23913, opened 2026-06-10
CAMEL-23651: camel-netty-http and camel-undertow - align muteException default with the other HTTP components

oscerd:
JIRA: [CAMEL-23651](https://issues.apache.org/jira/browse/CAMEL-23651)

## What

Aligns the `muteException` consumer option default in `camel-netty-http` and `camel-undertow` with the other HTTP server components, flipping it from `false` to `true`:

| Component | Before | After |
|---|---|---|
| camel-http / camel-jetty / camel-servlet (`HttpCommonComponent`) | `true` | `true` (unchanged) |
| camel-platform-http | `true` (since CAMEL-17826) | `true` (unchanged) |
| **camel-netty-http** | `false` | **`true`** |
| **camel-undertow** | `false` | **`true`** |

With `muteException=true`, a consumer-side processing failure returns an empty response body instead of the exception stack trace as `text/plain`.

## Changes

- `NettyHttpConfiguration` + `NettyHttpComponent`: field initialised to `true`, `defaultValue = "true"` on the annotations. The component-level field is the fallback for the URI parameter and is pushed onto the endpoint configuration, so both levels must change together.
- `UndertowComponent` + `UndertowEndpoint`: same default flip at both levels.
- `UndertowComponent.doCreateConsumer` (Rest DSL path): the `RestUndertowHttpBinding` is now given the endpoint's `muteException` value. Previously the binding was created with a hard-coded `false`, so the option (component- or endpoint-level) was silently ignored for Rest DSL consumers — the new default would not have been honoured there. `camel-netty-http` does not need this, as its binding reads the endpoint configuration directly.
- Regenerated metadata: component JSONs, catalog mirrors, endpoint/component DSL builders.
- Upgrade guide 4.21 entry (potential breaking change).

## Behaviour notes

`muteException` takes precedence over `transferException`, exactly as it already does in `camel-http`/`camel-jetty`/`camel-servlet` (`DefaultHttpBinding.doWriteExceptionResponse`). Routes using `transferException=true` on these two components must now also set `muteException=false` — the same combination the jetty/servlet tests have always used. This is called out in the upgrade guide.

## Tests

New:
- `NettyHttpMuteExceptionByDefaultTest` — consumer failure returns 500 + empty body with no option set
- `UndertowMuteExceptionByDefaultTest` — same for undertow
- `RestUndertowMuteExceptionByDefaultTest` — same for an undertow Rest DSL consumer (covers the binding propagation fix)

Updated to opt out explicitly (matching the existing jetty/servlet test pattern, e.g. `ServletTransferExceptionTest`):
- `NettyHttp500ErrorThrowExceptionOnServerTest` (`muteException=false`)
- `NettyHttpTransferExceptionTest` (`muteException=false&transferException=true`)
- `UndertowTransferExceptionTest` (`muteException=false&transferException=true`)

Results: `camel-netty-http` 258 tests, 0 failures; `camel-undertow` 165 tests, 0 failures. Full reactor build (`mvn clean install -DskipTests`) green; all regenerated files committed.

## Backports

Per the JIRA `fixVersions` (4.14.8 / 4.18.3 / 4.21.0), backport PRs to `camel-4.18.x` and `camel-4.14.x` will follow after this is merged, each with its own upgrade-guide entry.

---


davsclaus (approved):

ammachado:
The `muteException=true` default change in this PR broke `VertxHttpTransferExceptionTest` in `camel-vertx-http`. That test uses an Undertow consumer as the server-side endpoint with `transferException=true`, but now `muteException=true` takes precedence and suppresses exception serialization, so the client receives an empty `text/plain` 500 response instead of the serialized `IllegalStateException`.

A possible fix is to add `muteException=false` to the Undertow consumer URI in that test, consistent with the changes made here to `UndertowTransferExceptionTest`. It is a one-line change:

```java
// components/camel-vertx/camel-vertx-http/src/test/java/org/apache/camel/component/vertx/http/VertxHttpTransferExceptionTest.java
from(getTestServerUri() + "?transferException=true&muteException=false")
```

Worth checking whether any other cross-component tests that use an Undertow (or netty-http) consumer with `transferException=true` need the same opt-out.

PR #23926, opened 2026-06-11
CAMEL-23723: Add oauthProfile support to HTTP consumer components

luigidemasi:
Follow-up of [CAMEL-23685](https://issues.apache.org/jira/browse/CAMEL-23685), which introduced the public `OAuthTokenValidationFactory` SPI and `oauthProfile` support for `platform-http`.

JIRA: [CAMEL-23723](https://issues.apache.org/jira/browse/CAMEL-23723)

## What this PR does

Extends incoming `Authorization: Bearer` token validation to the main HTTP consumer components via the same SPI and `camel.oauth.<profile>.*` configuration — no JWT/JWKS/OIDC-discovery/introspection logic is duplicated into any component:

```java
from("undertow:http://0.0.0.0:8080/secure?oauthProfile=myprofile")
from("netty-http:http://0.0.0.0:8080/secure?oauthProfile=myprofile")
from("jetty:http://0.0.0.0:8080/secure?oauthProfile=myprofile")
```

| Component | Support |
|---|---|
| camel-jetty, camel-servlet | via shared `CamelServlet` hook |
| camel-netty-http | via per-consumer channel handler; requires `usingExecutorService=true` and `sync=true` (the defaults), rejects `nettySharedHttpServer` and mixed-executor shared ports at startup |
| camel-undertow | HTTP and WebSocket (validation at upgrade handshake, result propagated to message exchanges), REST DSL via `endpointProperty` |
| camel-rest-openapi | first-class `oauthProfile` endpoint option, forwarded explicitly to the consumer delegate; fails at startup when the delegate does not declare enforcement (see below) |
| camel-atmosphere-websocket | not supported — rejects the option at startup (documented) |
| camel-platform-http | refactored onto the new shared support (net deduplication, behavior aligned) |

## Behavioral contract (consistent with platform-http)

- Profile configuration resolved and validated at route startup (fail-fast)
- Missing token → `401` with `WWW-Authenticate: Bearer`; invalid token → `401` with `Bearer error="invalid_token"`
- Malformed `Authorization` header (RFC 6750 syntax, duplicates, oversize) → `400` with `Bearer error="invalid_request"`
- Validator infrastructure failure → `503` (rate-limited WARN log, fail-fast window for the dead IdP endpoint)
- `Authorization` header removed before route processing (also from native request objects where the transport allows)
- Validation result exposed as the `CamelOAuthTokenValidationResult` exchange property (authentication only — no authorization/role enforcement)

## Shared infrastructure

- New `OAuthHttpSecuritySupport` + `OAuthProfileAwareHttpEndpoint` in `camel-http-base` (RFC 6750 parsing, challenge generation, rejection handling) — single enforcement implementation for all components
- New `RestOpenApiConsumerFactory.supportsOAuthProfile()` SPI method in `camel-api` (default `false`, `@since 4.21`): `rest-openapi` fails closed at startup instead of creating an unprotected consumer when `oauthProfile` is set against a delegate factory that does not declare enforcement; `PlatformHttpComponent` declares support
- camel-oauth hardening along the way: JWKS serve-stale-on-error during IdP outages, per-endpoint fail-fast windows that record only transport-level failures and chain the root cause, lock scope fixes so no network I/O runs inside `ConcurrentHashMap.compute`

## Review feedback addressed

Per review discussion, `oauthProfile` on `rest-openapi` is now a declared `@UriParam` (catalog, endpoint DSL, IDE auto-complete) injected explicitly into the delegate parameters, and the `lenientProperties` annotation addition was reverted — runtime leniency via the pre-existing `isLenientProperties()` override is unchanged in both directions, so there is no behavioral change for existing `rest-openapi` users.

## Documentation

Per-component sections (including WebSocket and OPTIONS semantics, netty threading constraints, servlet container-security interplay), `oauth.adoc` shared contract and identity-provider failure behavior, REST DSL and contract-first OpenAPI pages, and a 4.21 upgrade-guide entry.

## Testing

Contract test suites per component (valid/missing/malformed/invalid/503, header stripping, exchange property, startup failure, duplicate headers on the wire, OPTIONS interplay, `matchOnUriPrefix`), WebSocket handshake/lifecycle tests including an unauthenticated-channel race reproduction, the rest-openapi fail-closed delegate check, and unit coverage for the RFC 6750 parser and JWT/JWKS validator paths (alg confusion, key rotation, rate limiting with injected clock).

## Note

Adds a clarifying comment to `VertxHttpTransferExceptionTest` (fixed on `main` by CAMEL-23734) documenting why `muteException=false` is required: CAMEL-23651 flipped the undertow default to `true`, and `muteException` takes precedence over `transferException`.

---
_Claude Code on behalf of Luigi De Masi_


components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
```diff
@@ -90,7 +90,8 @@
  * To call and expose REST services using OpenAPI specification as contract.
  */
 @UriEndpoint(firstVersion = "3.1.0", scheme = "rest-openapi", title = "REST OpenApi",
-             syntax = "rest-openapi:specificationUri#operationId", category = { Category.REST, Category.API })
+             syntax = "rest-openapi:specificationUri#operationId", category = { Category.REST, Category.API },
+             lenientProperties = true)
```
davsclaus: Adding `lenientProperties = true` enables `oauthProfile` pass-through to the delegate consumer, but it also means **all** `rest-openapi` endpoint URIs will now silently accept unknown/misspelled options. This is a behavioral change for existing `rest-openapi` users who relied on strict option validation at startup.

Consider adding an explicit upgrade guide entry noting that `rest-openapi` endpoint URIs are now lenient, separate from the `oauthProfile` documentation.

davsclaus (commented):
## Review: CAMEL-23723 — Add oauthProfile support to HTTP consumer components

89 files, +4361/−323 lines. Java builds pass (JDK 17, 25); doc validation fails.

Nice work extending the OAuth profile support across all HTTP consumer components with a clean shared SPI design. The centralized `OAuthHttpSecuritySupport` avoids duplicating JWT/JWKS/introspection logic and the RFC 6750 error responses are a solid improvement. A couple of items to address:

### Findings

**1. CI failure: Documentation validation** (Blocking)

The `build` job fails at the "Validate documentation" step (Antora). The Java builds pass. The Antora log shows a `section title out of sequence` warning. This could be related to heading levels in the new doc sections added to `rest-dsl.adoc`, `rest-dsl-openapi.adoc`, or `oauth.adoc`. Please investigate — if it's an infrastructure issue, re-running the job should resolve it; if it's a doc structure issue, the heading levels need adjustment.

**2. `rest-openapi` lenient properties — broader impact**

Adding `lenientProperties = true` to `RestOpenApiEndpoint` is needed for `oauthProfile` pass-through, but it affects **all** `rest-openapi` users — typos in any endpoint option will no longer be caught at startup. The upgrade guide documents the pass-through behavior for `oauthProfile` but doesn't mention that lenient properties are now enabled globally for `rest-openapi`. Consider adding an explicit upgrade guide entry for this behavioral change, so users who rely on strict option validation are aware.

### Questions

**3. Platform-http behavioral change for existing `oauthProfile` users**

The refactoring of `OAuthPlatformHttpSecurityHandler` changes error responses for users who already have `oauthProfile` on `platform-http`: malformed Bearer tokens now get HTTP 400 (was 401), and `WWW-Authenticate` headers are now included per RFC 6750. Since `oauthProfile` was introduced in the same 4.21 cycle (CAMEL-23685), this is acceptable for SNAPSHOT consumers — just confirming this behavioral delta is intentional.

### Verified (no issues)

- **Architecture**: `OAuthHttpSecuritySupport` centralizes validation; `OAuthProfileAwareHttpEndpoint` interface at `camel-http-base` level; `PlatformHttpConstants.OAUTH_TOKEN_VALIDATION_RESULT` delegates to the central constant (backward compatible)
- **Security**: Bearer token length limits (8192), leading whitespace limit (16), duplicate-header rejection, rate-limited infrastructure failure logging (30s), fail-closed in `CamelServlet` when security support is uninitialized
- **Component coverage**: Servlet/Jetty via `CamelServlet.authenticateOAuth()`, Netty in `HttpServerChannelHandler.beforeProcess()`, Undertow HTTP via `OAuthUndertowHttpHandler`, Undertow WebSocket during upgrade handshake, atmosphere-websocket explicitly rejects
- **Netty constraints**: Validates `usingExecutorService=true`, `sync=true`, rejects `nettySharedHttpServer` at startup
- **Undertow**: Dispatches from IO thread before blocking validation; WebSocket result attached to ONOPEN and subsequent exchanges
- **Tests**: Comprehensive `OAuthHttpSecuritySupportTest` (333 lines) plus per-component tests
- **Documentation**: Upgrade guide, REST DSL, OpenAPI, and component docs all updated with examples in Java/XML/YAML
- **`@since 4.21`** tags present on all new public types

---

**Note:** This review covers project conventions, architecture, and security-sensitive patterns. It does not replace specialized tools such as CodeRabbit or SonarCloud.

davsclaus:
## Suggestion: Add `oauthProfile` as a first-class `@UriParam` on `RestOpenApiEndpoint`

The `rest-openapi` endpoint is already lenient at runtime (`isLenientProperties()` returns `true` on `main`), so the `lenientProperties = true` annotation addition is metadata alignment rather than a behavioral change. However, `oauthProfile` would be better served as a proper `@UriParam` — it would appear in catalog metadata, endpoint DSL builders, IDE auto-complete, and component docs, rather than being a silent pass-through.

### Proposed changes

**1. `RestOpenApiEndpoint.java`** — add the field, getter/setter, and inject into delegate parameters:

```java
// Add field (alongside the other @UriParam fields, e.g. after apiContextPath)
@UriParam(label = "consumer,security", displayName = "OAuth Profile",
          description = "OAuth profile name passed to the HTTP consumer delegate for validating incoming "
                        + "Authorization: Bearer tokens. The selected consumer component must support the "
                        + "oauthProfile option; delegates that ignore unknown options will start without "
                        + "endpoint protection.")
private String oauthProfile;

// Add getter/setter
public String getOauthProfile() {
    return oauthProfile;
}

public void setOauthProfile(String oauthProfile) {
    this.oauthProfile = oauthProfile;
}
```

In `createConsumerFor()`, inject it into the parameters copy sent to the delegate factory — after `Map<String, Object> copy = new HashMap<>(parameters)` (line 325) and before the factory call (line 330):

```java
Map<String, Object> copy = new HashMap<>(parameters); // existing line
// pass oauthProfile to the delegate consumer
if (ObjectHelper.isNotEmpty(oauthProfile)) {
    copy.put("oauthProfile", oauthProfile);
}
```

**2. `@UriEndpoint` annotation** — revert the `lenientProperties = true` addition. The existing `isLenientProperties()` method override already handles runtime leniency for path/query parameter pass-through. No annotation change needed.

```java
// Keep the existing annotation without lenientProperties:
@UriEndpoint(firstVersion = "3.1.0", scheme = "rest-openapi", title = "REST OpenApi",
             syntax = "rest-openapi:specificationUri#operationId", category = { Category.REST, Category.API })
```

**3. Regenerate** — run `mvn clean install -DskipTests` in `components/camel-rest-openapi` so the generated URI factory, configurer, catalog JSON, and endpoint DSL builder pick up the new `@UriParam`.

**4. Upgrade guide** — the existing text about `rest-openapi` pass-through and lenient URIs can be simplified since `oauthProfile` is now a documented option. The caveat about verifying delegate support is still relevant and should stay.

### Why this is better

- `oauthProfile` appears in `rest-openapi` component docs, catalog JSON, and endpoint DSL builders
- IDE auto-complete works for the option
- No metadata change to lenient properties — avoids confusion about whether `rest-openapi` silently swallows typos (it already does via the method override, but the annotation didn't advertise it before)

luigidemasi:
## Re: Finding 1 — documentation validation failure

@davsclaus the failure is pre-existing on `main` and not related to this PR's doc changes.

From the Antora log of the failing run:
```
WARN (asciidoctor): section title out of sequence: expected level 1, got level 2
    file: docs/user-manual/modules/ROOT/pages/camel-jbang-configuration.adoc:6
```
and I think now has been fixed in commit: b0153b2adae9ceff4d52e7003e08c3af8ef3f2b6

### Re: first-class `@UriParam` for `oauthProfile`

Agreed — adopting this as proposed. It gives the option catalog/IDE/endpoint-DSL visibility and makes the pass-through intent explicit in code instead of riding the lenient leftovers.

One clarification for the record on the related inline comment: runtime behavior was already lenient before this PR — `RestOpenApiEndpoint.isLenientProperties()` has returned `true` since the component was introduced, and that override is what powers the existing delegate pass-through (`consumerComponentName`, `matchOnUriPrefix`, etc.). 

The annotation only aligned the generated metadata with that runtime reality, so there is no behavioral change for existing users either way, and no upgrade-guide entry is needed for it. 
Reverting it as suggested; if we want the catalog metadata to reflect the runtime leniency at some point, that can be a tiny separate JIRA.

#### Proposed addition: fail-closed delegate check for `oauthProfile`

While in there, I'd like to also close the fail-open this thread touches on (delegate silently ignoring a correctly-spelled `oauthProfile`). 
Since `oauthProfile` is a security option, a scoped capability method on the SPI can default closed without affecting anyone who doesn't use the option:

```java
// RestOpenApiConsumerFactory (camel-api)
/**
 * Whether consumers created by this factory enforce the oauthProfile option for
 * incoming bearer-token validation.
 *
 * @since 4.21
 */
default boolean supportsOAuthProfile() {
    return false;
}
```

PlatformHttpComponent (currently the only in-tree implementation and the only entry in
DEFAULT_REST_OPENAPI_CONSUMER_COMPONENTS) overrides it to true, and createConsumerFor() fails fast when oauthProfile is set but the resolved factory returns false. 

Result: setting oauthProfile against a delegate that does not enforce it becomes a startup error instead of an unprotected consumer — fail-closed by default, including for third-party factories, with zero impact on existing pass-through behavior. 
Typo'd option names remain subject to the documented lenient semantics; that is only addressable by the delegate's own strict binding.

Happy to include this in the PR, or move it to a follow-up JIRA if you prefer to keep the changeset as-is, let me know which you'd prefer.

### Re: Question 3 — platform-http error-response changes

Confirmed, intentional. The refactoring onto the shared `OAuthHttpSecuritySupport` aligned
platform-http with RFC 6750 error semantics: 
malformed `Authorization` syntax (including duplicate header values and oversized tokens) is now `400` with `WWW-Authenticate: Bearer
error="invalid_request"`, rejected tokens are `401` with `Bearer error="invalid_token"`, and missing credentials get the bare `Bearer` challenge. It also fixed two gaps in the original handler (single-value-only header reads, missing 400/invalid_token challenges).

Since `oauthProfile` was introduced by CAMEL-23685 in the same 4.21 cycle, no released behavior is affected — there are no existing users of the option outside SNAPSHOT. The new responses are documented in the component pages, `oauth.adoc`, and the 4.21 upgrade-guide entry.

davsclaus (approved):

PR #23927, opened 2026-06-11
CAMEL-21438: Fix six flaky tests and missing aarch64 image for TensorFlow Serving

ammachado:
# Description

Fix six test reliability issues:

**1. `DefaultConsumerBridgeErrorHandlerContinuedTest` (flaky assertion)**
The test asserted `mock:result` would receive **zero** messages when `continued(true)` is set on the error handler. In practice, `continued(true)` causes the exchange to continue through the main route after error handling, so `mock:result` does receive at least one message. Changed the expectation to `expectedMinimumMessageCount(1)`.

**2. `SimpleLRUCacheTest` (race condition after concurrent puts)**
After the concurrent-put latch completed, the test immediately asserted cache size and eviction counter. The eviction listener callback can fire asynchronously, causing intermittent failures. Wrapped the post-latch assertions in `Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(...)` so the assertions wait for the eviction to settle.

**3. TensorFlow Serving test-infra: missing aarch64 image**
The `bitnami/tensorflow-serving:2.18.0` image on Docker Hub is not available for `aarch64`. Replaced it with `mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1`, which is available on the Google mirror registry for that architecture, consistent with the project's preference for `mirror.gcr.io` over Docker Hub.

**4. `FileSortByExpressionTest`, `FileSortByNestedExpressionTest`, `FileSortByIgnoreCaseExpressionTest` (mock expectation race)**
`MockEndpoint.expectedBodiesReceived()` must be called before routes start delivering messages. With `initialDelay=0`, the file consumer fires on a background thread as soon as `addRoutes()` returns. If messages arrive before `expectedBodiesReceived()` is called, `expectedBodyValues` is still null so `performAssertions()` never populates `actualBodyValues`, causing body order assertions to compare each expected value against `null`.

Fix: move `getMockEndpoint()` and `expectedBodiesReceived()` before `context.addRoutes()` in all three test classes. Also removed no-op `context.start()` calls that widened the race window. The `@DisabledOnOs(ppc64le)` guard on `FileSortByIgnoreCaseExpressionTest` is retained until ppc64le CI can confirm the fix is sufficient on that platform.

**5. `FileProducerCharsetUTFtoISOConvertBodyToTest` (0-byte file race + cleanup masking)**
Two issues remained after two prior fixes (CAMEL-18608):

- `@BeforeEach writeTestData()` runs after the superclass `setUp()` starts the context and route. With `initialDelay=0` the consumer polls every 10 ms. A poll landing between `Files.newOutputStream()` (creates a 0-byte file) and `fos.close()` picks up an empty file, producing 0-byte output and causing the size assertion to never be satisfied.
  Fix: write to a `.tmp` file first then atomically rename via `Files.move(..., ATOMIC_MOVE)`. The consumer's `fileName` filter ignores the `.tmp` file and only ever sees the fully-written target.
- `@AfterEach` used `Files.delete()` instead of `Files.deleteIfExists()`: if the test fails before the output file is created, cleanup also throws, masking the original failure with a misleading "cannot run due to an error cleaning up" message.
  Fix: switch to `Files.deleteIfExists()`.

**6. `QueueProducerQoSTest` (ANYCAST/MULTICAST mismatch + zero-timeout mock assertion)**
Two issues caused non-deterministic failures:

- The expiry route consumed from `sjms:topic:ExpiryQueue` but Artemis routes expired messages to an ANYCAST address. A JMS Topic consumer creates a MULTICAST subscription, which does not receive ANYCAST messages. Whether the test passed depended on which subscriber created the ExpiryQueue address first. Fixed by switching to `sjms:queue:ExpiryQueue`.
- Both test methods called the static `MockEndpoint.assertIsSatisfied(context)` which delegates to each mock's own `resultWaitTime` (defaults to 0). With `requestTimeout=500ms` and `timeToLive=1000ms`, the expiry notification arrives after the assertion was already evaluated. Fixed by calling `mockExpiredAdvisory.assertIsSatisfied(10_000)` with an explicit 10-second timeout.

# Target

# Tracking

---

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/DefaultConsumerBridgeErrorHandlerContinuedTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/DefaultConsumerBridgeErrorHandlerContinuedTest.java
@@ -48,10 +48,9 @@ public class DefaultConsumerBridgeErrorHandlerContinuedTest extends ContextTestS
         getMockEndpoint("mock:onException").expectedMinimumMessageCount(1);
         getMockEndpoint("mock:subroute").expectedMinimumMessageCount(1);
 
-        // With continued(true), processing should continue after error handling
-        // However, since the consumer throws before creating a valid exchange,
-        // mock:result won't receive messages
-        getMockEndpoint("mock:result").expectedMessageCount(0);
+        // With continued(true), the exchange continues through the main route after error handling,
+        // so mock:result should receive messages.
+        getMockEndpoint("mock:result").expectedMinimumMessageCount(1);
 
         assertMockEndpointsSatisfied();
 
--- a/test-infra/camel-test-infra-tensorflow-serving/src/main/resources/org/apache/camel/test/infra/tensorflow/serving/services/container.properties
+++ b/test-infra/camel-test-infra-tensorflow-serving/src/main/resources/org/apache/camel/test/infra/tensorflow/serving/services/container.properties
@@ -15,7 +15,7 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 tensorflow.serving.container=mirror.gcr.io/tensorflow/serving:2.19.1
-tensorflow.serving.container.aarch64=bitnami/tensorflow-serving:2.18.0
+tensorflow.serving.container.aarch64=mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1
 tensorflow.serving.container.ppc64le=ibmcom/powerai:1.7.0-tensorflow-serving-ubuntu18.04-py37-ppc64le-21.035
 tensorflow.serving.container.ppc64le.version.include=ppc64le,ubuntu
 tensorflow.serving.container.ppc64le.version.exclude=x86_64,amd64,arm64
--- a/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/QueueProducerQoSTest.java
+++ b/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/QueueProducerQoSTest.java
@@ -91,7 +91,7 @@ public class QueueProducerQoSTest extends CamelTestSupport {
             template.requestBody(endpoint, "test message");
         });
 
-        MockEndpoint.assertIsSatisfied(context);
+        mockExpiredAdvisory.assertIsSatisfied(10_000);
 
         assertEquals(0, service.countMessages(TEST_INOUT_DESTINATION_NAME),
                 "There were unexpected messages left in the queue: " + TEST_INOUT_DESTINATION_NAME);
@@ -104,7 +104,7 @@ public class QueueProducerQoSTest extends CamelTestSupport {
         String endpoint = String.format("sjms:queue:%s?timeToLive=1000", TEST_INONLY_DESTINATION_NAME);
         template.sendBody(endpoint, "test message");
 
-        MockEndpoint.assertIsSatisfied(context);
+        mockExpiredAdvisory.assertIsSatisfied(10_000);
 
         assertEquals(0, service.countMessages(TEST_INONLY_DESTINATION_NAME),
                 "There were unexpected messages left in the queue: " + TEST_INONLY_DESTINATION_NAME);
@@ -124,7 +124,7 @@ public class QueueProducerQoSTest extends CamelTestSupport {
         return new RouteBuilder() {
             @Override
             public void configure() {
-                from("sjms:topic:ExpiryQueue")
+                from("sjms:queue:ExpiryQueue")
                         .routeId(EXPIRED_MESSAGE_ROUTE_ID)
                         .log("Expired message")
                         .to(MOCK_EXPIRED_ADVISORY);
--- a/core/camel-core/src/test/java/org/apache/camel/component/file/FileProducerCharsetUTFtoISOConvertBodyToTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/file/FileProducerCharsetUTFtoISOConvertBodyToTest.java
@@ -20,6 +20,7 @@ import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.ContextTestSupport;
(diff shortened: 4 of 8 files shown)


davsclaus (commented):
## Review: CAMEL-21438 — Fix flaky tests and missing aarch64 image

CI is green. The test reliability fixes are well-motivated and follow project conventions. A few notes below.

### Findings

**1. PR description is incomplete** (Minor)

The description mentions 3 changes, but the diff modifies 7 files. Four additional fixes are not covered:
- `FileProducerCharsetUTFtoISOConvertBodyToTest` — atomic file write + `deleteIfExists` cleanup
- `FileSortByExpressionTest` — move expectations before routes, remove hardcoded 1s timeout
- `FileSortByIgnoreCaseExpressionTest` — remove redundant `context.start()`, fix `initialDelay`/`delay`, typo fix
- `FileSortByNestedExpressionTest` — remove redundant `context.start()`, move expectations

Please update the PR description to cover all changes.

**2. `DefaultConsumerBridgeErrorHandlerContinuedTest` — assertion reversal**

This reverses the intent of the original CAMEL-22907 test (`fe9abb35c3b7`), which explicitly asserted `mock:result` should receive **0 messages**. The change to `expectedMinimumMessageCount(1)` is consistent with how `continued(true)` should work — after onException processing, the exchange continues through the route to `mock:result`. The original `expectedMessageCount(0)` was a timing-dependent assertion that passed only when the mock check completed before the continued exchange arrived.

This looks correct to me, but since it reverses an intentional assertion from CAMEL-22907, it would be good to get confirmation from @gnodet that this interpretation is right.

**3. `bitnamilegacy` image**

The `mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1` image uses a "legacy" namespace. Worth checking that it has ongoing maintenance for long-term reliability.

### Verified (no issues)

- **`SimpleLRUCacheTest`** — Awaitility usage is correct; eviction callbacks fire asynchronously. Follows the project's "no `Thread.sleep()`" convention.
- **File sort tests** — Moving mock expectations before `context.addRoutes()` and removing redundant `context.start()` (context is auto-started by `ContextTestSupport.setUp()`) is the right pattern.
- **`FileProducerCharsetUTFtoISOConvertBodyToTest`** — Atomic rename prevents file consumer from seeing a half-written file. Good defensive fix.
- **TensorFlow aarch64 image** — Switching from Docker Hub to `mirror.gcr.io` follows project conventions.

---

**Note:** This review covers project conventions, test patterns, and behavioral correctness. It does not replace specialized tools such as CodeRabbit or SonarCloud.

core/camel-core/src/test/java/org/apache/camel/processor/DefaultConsumerBridgeErrorHandlerContinuedTest.java
```diff
@@ -48,10 +48,9 @@ public void testDefaultConsumerBridgeErrorHandlerContinued() throws Exception {
         getMockEndpoint("mock:onException").expectedMinimumMessageCount(1);
         getMockEndpoint("mock:subroute").expectedMinimumMessageCount(1);
 
-        // With continued(true), processing should continue after error handling
-        // However, since the consumer throws before creating a valid exchange,
-        // mock:result won't receive messages
-        getMockEndpoint("mock:result").expectedMessageCount(0);
+        // With continued(true), the exchange continues through the main route after error handling,
+        // so mock:result should receive messages.
+        getMockEndpoint("mock:result").expectedMinimumMessageCount(1);
```
davsclaus, line 53: This reverses the original CAMEL-22907 assertion which expected 0 messages here, with the rationale that "the consumer throws before creating a valid exchange". With `continued(true)`, the bridge error handler creates an exchange, onException processes it, and `continued(true)` allows it to proceed through the route to `mock:result`. The original `expectedMessageCount(0)` was a race condition — it passed only when the mock assertion completed before the continued exchange arrived.

The new assertion looks correct, but since it contradicts the original test's explicit intent, confirmation from the CAMEL-22907 author would be valuable.

test-infra/camel-test-infra-tensorflow-serving/src/main/resources/org/apache/camel/test/infra/tensorflow/serving/services/container.properties
```diff
@@ -15,7 +15,7 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 tensorflow.serving.container=mirror.gcr.io/tensorflow/serving:2.19.1
-tensorflow.serving.container.aarch64=bitnami/tensorflow-serving:2.18.0
+tensorflow.serving.container.aarch64=mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1
```
davsclaus, line 18: Minor: the `bitnamilegacy` namespace suggests this image may be deprecated upstream. Worth verifying it has ongoing maintenance so this doesn't need another fix soon.

ammachado:
> The `mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1` image uses a "legacy" namespace. Worth checking that it has ongoing maintenance for long-term reliability.

Thanks for flagging this. The `bitnamilegacy` namespace is indeed deprecated: Bitnami restructured their container catalog in August 2025 and moved all versioned/pinned tags to `bitnamilegacy` with no further updates. The concern is valid.

However, there is currently no well-maintained `aarch64` image for TensorFlow Serving from any trusted registry:
- The official `tensorflow/serving` image is `amd64`-only (long-standing open issue, no `arm64` support planned).
- `emacski/tensorflow-serving`, the main community `aarch64` build, is last published at `2.6.0` and is not actively maintained.
- AWS Deep Learning Containers publish a `2.19` `aarch64` image, but it is SageMaker-specific and not suitable here.

`bitnamilegacy/tensorflow-serving:2.19.1` is frozen but functional. It is the only option that matches our required version on `aarch64`. We can revisit this when either the official TF Serving project adds `aarch64` support or a community project catches up to 2.19+.

The remaining alternative is to apply the `skipITs.aarch64` fix: add `<skipITs.aarch64>true</skipITs.aarch64>` to `camel-tensorflow-serving/pom.xml` and remove the `bitnamilegacy` line from `container.properties`. Please advice.

core/camel-core/src/test/java/org/apache/camel/processor/DefaultConsumerBridgeErrorHandlerContinuedTest.java
ammachado: @gnodet, you added this test in `CAMEL-22907` with `expectedMessageCount(0)` and the comment "since the consumer throws before creating a valid exchange, mock:result won't receive messages".

We found it flaky: after the `CAMEL-22907` fix, `continued(true)` causes the exchange to continue through the main route after `onException` handling, so `mock:result` does receive messages. The assertion was changed to `expectedMinimumMessageCount(1)`.

Could you confirm whether this matches the intended post-fix behavior, or whether `mock:result` was expected to remain at zero for a reason we may have missed?

davsclaus:
> > The `mirror.gcr.io/bitnamilegacy/tensorflow-serving:2.19.1` image uses a "legacy" namespace. Worth checking that it has ongoing maintenance for long-term reliability.
> 
> Thanks for flagging this. The `bitnamilegacy` namespace is indeed deprecated: Bitnami restructured their container catalog in August 2025 and moved all versioned/pinned tags to `bitnamilegacy` with no further updates. The concern is valid.
> 
> However, there is currently no well-maintained `aarch64` image for TensorFlow Serving from any trusted registry:
> 
> * The official `tensorflow/serving` image is `amd64`-only (long-standing open issue, no `arm64` support planned).
> * `emacski/tensorflow-serving`, the main community `aarch64` build, is last published at `2.6.0` and is not actively maintained.
> * AWS Deep Learning Containers publish a `2.19` `aarch64` image, but it is SageMaker-specific and not suitable here.
> 
> `bitnamilegacy/tensorflow-serving:2.19.1` is frozen but functional. It is the only option that matches our required version on `aarch64`. We can revisit this when either the official TF Serving project adds `aarch64` support or a community project catches up to 2.19+.
> 
> The remaining alternative is to apply the `skipITs.aarch64` fix: add `<skipITs.aarch64>true</skipITs.aarch64>` to `camel-tensorflow-serving/pom.xml` and remove the `bitnamilegacy` line from `container.properties`. Please advice.
> 
> _Claude Code on behalf of Adriano Machado_

Yeah though aarch64 is for mac developers to run on the laptop and not for production users. So it may be okay to allow them to test and validate tensorflow. Can we add a comment that about so we will not forget.

davsclaus (approved):

PR #23928, opened 2026-06-11
CAMEL-19528: camel-jpa - replace Thread.sleep in tests

rohfl:
Title:
CAMEL-19528: Replace Thread.sleep with Awaitility in camel-jpa tests

# Body:

## Motivation
The `camel-jpa` test suite relied on hard-coded `Thread.sleep()` calls to wait for asynchronous consumer activity (entity processing, deletion, file moves).
These sleeps are flaky and slow: they either wait longer than necessary on fast machines or, on slow ones, race ahead before the work has actually completed.

## Changes
Added the `org.awaitility:awaitility` test dependency to `camel-jpa`.

Migrated the `Thread.sleep()`-based waits to Awaitility in:
- `AbstractJpaMethodTest#consumeEntity` — wait until the consumer has deleted the entities (database is empty) before asserting, rather than sleeping and reading the entity while the consumer may still be mid-transaction.
- `JpaWithNamedQueryTest` — poll the database (with `ignoreExceptions()`) until the consumer's transaction has committed, instead of sleeping before reading.
- `FileConsumerJpaIdempotentTest` — wait until the consumed file has been moved to the `done` directory, and use `MockEndpoint.setAssertPeriod(...)` to assert the file is not re-consumed (replacing a trailing sleep).

### DIFF
--- a/components/camel-jpa/pom.xml
+++ b/components/camel-jpa/pom.xml
@@ -118,6 +118,12 @@
             <version>${jakarta-transaction-api-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>${awaitility-version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <profiles>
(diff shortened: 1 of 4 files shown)


davsclaus (approved):
Clean, well-done migration from `Thread.sleep()` to Awaitility — each replacement uses the appropriate technique for the specific wait condition:

- `untilAsserted()` for polling database state in `AbstractJpaMethodTest` and `JpaWithNamedQueryTest`
- `ignoreExceptions()` in `JpaWithNamedQueryTest` to handle the OpenJPA `InvalidStateException` while the consumer's transaction is still active — much more robust than a fixed 2-second sleep
- `until(file::exists)` for the file-move check in `FileConsumerJpaIdempotentTest`
- `mock.setAssertPeriod(2000)` for the negative assertion ("file not re-consumed") — the proper Camel MockEndpoint pattern

All timeouts are explicit, CI is green, and the PR is well-scoped to the three test files that needed it.

PR #23949, opened 2026-06-11
CAMEL-23735: Restore documented default aggregation strategy in camel-dynamic-router

Tholgue:
# Description

Since the CAMEL-20302 refactoring (Camel 4.4.0), the `dynamic-router` component uses a private
`NoopAggregationStrategy` as its default aggregation strategy, which keeps the **first** recipient's
result. Both the `aggregationStrategy` option documentation ("By default, Camel will use the last
reply as the outgoing message.") and the javadoc of
`DynamicRouterRecipientListHelper#createAggregationStrategy` ("If all else fails, then a
`UseLatestAggregationStrategy` is used.") still describe the pre-4.4 behavior — and the stale
`UseLatestAggregationStrategy` import was still present, suggesting the change was unintentional.

Practical impact: in `recipientMode=allMatch`, modifications made by any recipient other than the
first one (message body, exchange properties) are silently discarded when the exchange returns to
the calling route, so a `choice()` testing a property set by a subscriber behaves differently
depending on subscription priorities.

This PR restores `UseLatestAggregationStrategy` as the default, matching the documentation and the
pre-4.4 behavior (`DynamicRouterEndpoint#determineAggregationStrategy` in 4.3.0):

- `DynamicRouterRecipientListHelper`: fallback changed from `NoopAggregationStrategy` to
  `UseLatestAggregationStrategy`; `NoopAggregationStrategy` is kept but `@Deprecated` to avoid
  breaking binary compatibility on LTS backports.
- `DynamicRouterRecipientListHelperTest`: the fallback test now asserts
  `UseLatestAggregationStrategy`.
- New `DynamicRouterDefaultAggregationIT`: end-to-end reproducer pinning the documented behavior
  (two matching recipients in allMatch mode; the last recipient's body and exchange property must be
  visible in the calling route). This test fails before the fix and passes after.

Fixes [CAMEL-23735](https://issues.apache.org/jira/browse/CAMEL-23735). Since the regression is in
4.4.0 and later, a backport to the active LTS branches (e.g. camel-4.14.x) would be appreciated.

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a
  [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-23735) filed for the change

- [x] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all
  auto-generated changes.

### DIFF
--- a/components/camel-dynamic-router/src/main/java/org/apache/camel/component/dynamicrouter/routing/DynamicRouterRecipientListHelper.java
+++ b/components/camel-dynamic-router/src/main/java/org/apache/camel/component/dynamicrouter/routing/DynamicRouterRecipientListHelper.java
@@ -152,7 +152,7 @@ public final class DynamicRouterRecipientListHelper {
                                 .orElseThrow(() -> new IllegalArgumentException(
                                         "Cannot find AggregationStrategy in Registry with name: " +
                                                                                 cfg.getAggregationStrategy()))))
-                .orElse(new NoopAggregationStrategy());
+                .orElse(new UseLatestAggregationStrategy());
         CamelContextAware.trySetCamelContext(strategy, camelContext);
         return cfg.isShareUnitOfWork() ? new ShareUnitOfWorkAggregationStrategy(strategy) : strategy;
     }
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Well-investigated regression fix. The mismatch between the documentation ("By default, Camel will use the last reply as the outgoing message" in `DynamicRouterConfiguration.java:169`), the javadoc on `createAggregationStrategy` (references `UseLatestAggregationStrategy`), the stale import, and the actual `NoopAggregationStrategy` code is clear evidence this was unintentional in CAMEL-20302.

The fix is minimal and correct:
- One-line production change aligning code with docs
- `NoopAggregationStrategy` deprecated (not removed) for binary compat on LTS backports
- Unit test updated to assert the new default
- Good end-to-end integration test (`DynamicRouterDefaultAggregationIT`) that reproduces the issue with two matching recipients in `allMatch` mode

PR #23958, opened 2026-06-11
CAMEL-23738: camel-keycloak - always verify access token even without required roles/permissions

oscerd:
This implements [CAMEL-23738](https://issues.apache.org/jira/browse/CAMEL-23738).

`KeycloakSecurityPolicy` performs access-token verification (signature, issuer and expiry for local JWT, or active state and issuer for token introspection) only inside the role and permission checks in `KeycloakSecurityProcessor`. Those checks are skipped when `requiredRoles` and `requiredPermissions` are both empty — the default, and the documented "Basic Setup" — so the token was not verified in that configuration.

This change makes `beforeProcess()` always authenticate the token when one is present, independently of whether roles or permissions are configured. Role and permission checks now run only after authentication, and an invalid or unverifiable token is rejected the same way a missing token is.

### Changes
- `KeycloakSecurityProcessor`: add `authenticateToken()` and invoke it when no roles/permissions are configured; the existing role/permission code paths are unchanged (no double verification).
- New `KeycloakSecurityProcessorTest` covering local JWT and token-introspection modes with no roles/permissions configured.
- Component docs note and 4.21 upgrade-guide entry for the behavior change.

### Testing
- `camel-keycloak` module: unit tests + Testcontainers integration tests green.
- Full reactor `mvn clean install -DskipTests`: green.

### DIFF
--- a/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityProcessor.java
+++ b/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityProcessor.java
@@ -59,11 +59,23 @@ public class KeycloakSecurityProcessor extends DelegateProcessor {
                 throw new CamelAuthorizationException("Access token not found in exchange", exchange);
             }
 
-            if (!policy.getRequiredRolesAsList().isEmpty()) {
+            boolean rolesRequired = !policy.getRequiredRolesAsList().isEmpty();
+            boolean permissionsRequired = !policy.getRequiredPermissionsAsList().isEmpty();
+
+            // The token is always authenticated before the route runs (signature, issuer and expiry for
+            // local JWT verification, or active state and issuer for introspection). validateRoles() and
+            // validatePermissions() already authenticate the token, but they only run when roles or
+            // permissions are required; when neither is configured, authenticate here so that an invalid or
+            // unverifiable token is rejected instead of accepted. Authorization checks run after authentication.
+            if (!rolesRequired && !permissionsRequired) {
+                authenticateToken(accessToken, exchange);
+            }
+
+            if (rolesRequired) {
                 validateRoles(accessToken, exchange);
             }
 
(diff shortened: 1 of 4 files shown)


components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityProcessor.java
```diff
@@ -59,11 +59,23 @@ protected void beforeProcess(Exchange exchange) throws Exception {
                 throw new CamelAuthorizationException("Access token not found in exchange", exchange);
             }
 
-            if (!policy.getRequiredRolesAsList().isEmpty()) {
+            boolean rolesRequired = !policy.getRequiredRolesAsList().isEmpty();
+            boolean permissionsRequired = !policy.getRequiredPermissionsAsList().isEmpty();
+
+            // Always authenticate the token when one is present - signature, issuer and expiry for local
```
apupier: this comment seems to not correspond to what the code is doing.
The token is not always authenticated, it is only if roles and permission are not required

apupier:
requires rebase to benefit from fixed mai branch https://github.com/apache/camel/pull/23962

components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakSecurityProcessor.java
oscerd: Good catch, thanks — reworded the comment (force-pushed). The token *is* always authenticated before the route runs: `validateRoles()`/`validatePermissions()` already verify it (signature/issuer/expiry for local JWT, or active state/issuer for introspection) when roles or permissions are required, and the new `authenticateToken()` call covers the case where neither is configured. The comment now describes that overall flow instead of implying the `if`-block itself always authenticates.

I also rebased onto `main` to pick up `10003522d3f` (jbang doc-generation fix), which clears the unrelated "uncommitted changes" CI failure.

davsclaus (approved):

PR #23965, opened 2026-06-11
CAMEL-23740: Fix WrappedFile converter for FTP/SFTP remote files

Croway:
## Summary

- Fix `LangChain4jAgentConverter.toAiAgentBody(WrappedFile)` to handle remote file types (FTP, SFTP) by falling back to the exchange type converter when `getFile()` does not return a `java.io.File`
- MIME type is detected from `Exchange.FILE_NAME` header for remote files, with the same priority chain (agent header > file content type header > extension)
- Add unit tests for remote `WrappedFile` conversion with both extension-based and header-based MIME detection

## Test plan

- [x] Existing `LangChain4jAgentAutoConversionTest` tests pass (local `GenericFile<File>` path unchanged)
- [x] New `shouldConvertRemoteWrappedFileWithFileNameHeader` test verifies remote file → `ImageContent` via `Exchange.FILE_NAME`
- [x] New `shouldConvertRemoteWrappedFileWithMediaTypeHeader` test verifies header-based MIME override for remote files
- [x] Full IT suite passes (tools, multimodal, memory, guardrails)

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConverter.java
+++ b/components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConverter.java
@@ -36,6 +36,7 @@ import dev.langchain4j.data.video.Video;
 import org.apache.camel.Converter;
 import org.apache.camel.Exchange;
 import org.apache.camel.Message;
+import org.apache.camel.NoTypeConversionAvailableException;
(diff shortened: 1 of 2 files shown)


davsclaus (changes requested):
See how we detect WrappedFile and get its object, as when you use the mandatory convert to byte[] you end up reading the payload into memory, also a big FTP file, or a big file from camel-file that are in a WrappedFile / GenericFile.

There are some other components that has some code for handling this you should look at first

davsclaus (approved):

PR #23970, opened 2026-06-11
CAMEL-23739: camel-openai - support WrappedFile, byte[] and InputStream bodies for vision models

Croway:
## Description

Fixes [CAMEL-23739](https://issues.apache.org/jira/browse/CAMEL-23739).

The `chat-completion` operation only accepted `java.io.File` bodies for vision/image input. Bodies produced by file-based and cloud storage consumers (`WrappedFile`/`GenericFile` from file/FTP/SFTP, `byte[]`/`InputStream` from AWS S3, Azure Blob, etc.) fell through to text handling, corrupting binary image content. MIME detection also relied solely on `Files.probeContentType()`, which returns `null` on macOS for many paths.

### Changes

- `OpenAIProducer.buildUserMessage()` now dispatches on `WrappedFile`, `File`, `Path`, `byte[]` and `InputStream` bodies (same pattern as `OpenAIAudioTranscriptionProducer`)
- `WrappedFile` wrapping a local `java.io.File` is used by path directly; remote files (FTP/SFTP) resolve the MIME type from headers/file name **before** streaming the content via the type converter, so large unsupported payloads are rejected without being read into memory
- New `MimeTypeHelper` with header-priority MIME resolution: `CamelOpenAIMediaType` (new header) → `CamelFileContentType` / cloud storage content-type headers (S3, Azure Blob/Data Lake, GCS, MinIO, IBM COS) → `Content-Type` → file name extension fallback
- Extension detection uses `org.apache.camel.util.MimeTypeHelper.probeMimeType()` (per review feedback) instead of `Files.probeContentType()` — deterministic across platforms (previously a `.xml` file was a text prompt on macOS but rejected on Linux) and covers the full built-in MIME type table; `text/`-prefixed types plus `application/xml`/`application/json` are treated as text prompts
- Backward compatible: `byte[]`/`InputStream` bodies without an image MIME type keep the previous text behavior
- New `OpenAIVisionBodyTypesMockTest` (11 tests, including a real `file:` consumer route) asserting the base64 data URL in the outgoing request decodes to the original bytes
- Component docs updated (input handling, MIME detection priority, cloud storage examples) and 4.21 upgrade guide entry added


davsclaus:
btw see also MimeTyperHelper in camel-util - it avoid the files probe

davsclaus (approved):
Clean, well-structured enhancement. The body type dispatch in `buildUserMessage()` is straightforward, `MimeTypeHelper` has a sensible priority chain covering cloud storage headers, and the 9 tests provide good coverage including error paths. Documentation and upgrade guide entry are both present.

Nice fix on the existing doc bug (`?model=gpt-4.1-mini?userMessage=...` → `&userMessage=...`).

Croway:
> btw see also MimeTyperHelper in camel-util - it avoid the files probe

good point, I've updated the code, there's still room for improvement, in particular wrt components that interact with files, like the AI ones, soon or later we might refactor these components and provide a generic one with some shared utility classes

PR #23972, opened 2026-06-12
Revert testcontainer upgrade from 2.0.5 to 2.0.3

apupier:
testcontainer is not working on our Jenkins CI with 2.0.4+ https://issues.apache.org/jira/browse/CAMEL-23538

it was introduced when aligning dependencies with Spring Boot 4.1.0 https://github.com/apache/camel/commit/4a903e2b8f3c10014ded822d6fe5e08a65f638f0

# Description

# Target

# Tracking

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -506,7 +506,7 @@
         <stringtemplate-version>4.3.4</stringtemplate-version>
         <tahu-version>1.0.19</tahu-version>
         <tamboui-version>0.3.0</tamboui-version>
-        <testcontainers-version>2.0.5</testcontainers-version>
+        <testcontainers-version>2.0.3</testcontainers-version>
         <thymeleaf-version>3.1.5.RELEASE</thymeleaf-version>
         <tika-version>3.3.1</tika-version>
         <twilio-version>12.1.1</twilio-version>


oscerd (approved):

davsclaus (changes requested):
well we should not use old test container because of a ibm platform - end users of camel should be able to use the latest test container release.

davsclaus:
maybe the older version can be used only for that power platform via some maven profile stuff

apupier:
i think we should first fix the current main branch. 
Then we can discuss the strategy on a different issue:
* invest time to fix the upgrade on pcc64le
* invest time to introduce some ways to have different versions used (with all technical debts and complexity that it involves on the maintenance)

davsclaus:
you should maybe create an INFRA so ASF can be aware of this about their IBM servers

davsclaus:
The testcontainers issue only affects ppc64le on ASF Jenkins CI. Instead of downgrading globally, we can use a Maven profile that auto-activates on `ppc64le` and overrides the version to 2.0.3, keeping 2.0.5 as the default for all other platforms.

See #23985 for this alternative approach.

PR #23984, opened 2026-06-12
Configure Maven cache for Atlassian artifacts on GitHub Actions

apupier:
This is not ideal as we won't caught if the remote reposiotry completely vanish.
It is configured only for Linux based GitHub workflows for now (which are the one running the most often and more likely to hit this issue)

use the same configuration than what camel-quarkus is using.

so far no proposition or explanation come from Atlassian https://community.developer.atlassian.com/t/more-and-more-frequent-403-when-downloading-maven-artifacts/101078

Note that to be active, it must be ran successfully one time

# Description

# Target

# Tracking

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -51,6 +51,18 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
           cache: 'maven'
+      - name: Get Atlassian dependency version
+        id: atlassian-version
+        run: |
+          JIRA_VERSION=$(./mvnw help:evaluate -Dexpression=jira-rest-client-api-version -q -DforceStdout -N -f parent/pom.xml)
+          echo "version=${JIRA_VERSION}" >> $GITHUB_OUTPUT
+      - name: Restore Atlassian Maven Cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/.m2/repository/com/atlassian
+            ~/.m2/repository/io/atlassian
+          key: atlassian-maven-${{ steps.atlassian-version.outputs.version }}
       - name: maven build
         run: ./etc/scripts/regen.sh
       - name: archive logs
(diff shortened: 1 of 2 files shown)


oscerd (approved):

.github/workflows/main-build.yml
```diff
@@ -51,6 +51,18 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
           cache: 'maven'
+      - name: Get Atlassian dependency version
+        id: atlassian-version
+        run: |
+          JIRA_VERSION=$(./mvnw help:evaluate -Dexpression=jira-rest-client.version -q -DforceStdout -N)
```
jamesnetherton:
```suggestion
          JIRA_VERSION=$(./mvnw help:evaluate -Dexpression=jira-rest-client-api-version -q -DforceStdout -N)
```

.github/workflows/pr-build-main.yml
```diff
@@ -90,6 +90,18 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
           cache: 'maven'
+      - name: Get Atlassian dependency version
+        id: atlassian-version
+        run: |
+          JIRA_VERSION=$(./mvnw help:evaluate -Dexpression=jira-rest-client.version -q -DforceStdout -N)
```
jamesnetherton:
```suggestion
          JIRA_VERSION=$(./mvnw help:evaluate -Dexpression=jira-rest-client-api-version -q -DforceStdout -N)
```

.github/workflows/main-build.yml
```diff
@@ -59,6 +71,27 @@ jobs:
...
+        run: |
+            tar -I 'zstd -T0' -cf ${{ runner.temp }}/maven-repo.tar.zst -C ~ .m2/repository
+            ls -lh ${{ runner.temp }}/maven-repo.tar.zst
+            df -h /
+      - name: Persist Maven Repo
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: maven-repo
+          path: ${{ runner.temp }}/maven-repo.tar.zst
+          retention-days: 1
```
jamesnetherton: We don't need these steps for the Camel core. It's just for CQ because we have a ton of concurrent jobs.

jamesnetherton (approved):

jamesnetherton (commented):
Sorry, I might have approved too soon, I think you actually need to `cd parent` before trying to evaluate the maven property.

apupier:
> I might have approved too soon, I think you actually need to `cd parent` before trying to evaluate the maven property.

good point. We need to point to the pom of the parent module.
I chose to use -f parent/pom.xml instead of cd parent as mvnw is not available in parent module

davsclaus (approved):
CI-only change, no production code impact. Cache pins use full SHA, key is version-aware, `if: always()` on save ensures seeding even on failed builds. Consistent across both workflow files. Approach proven in camel-quarkus.

jamesnetherton (approved):

PR #23996, opened 2026-06-12
CAMEL-23697: Add responseType parameter for structured output in lang…

zbendhiba:
…chain4j-agent

# Description

# Target

# Tracking


components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -102,7 +121,13 @@ public void process(Exchange exchange) throws Exception {
 
         ToolProvider toolProvider = createComposedToolProvider(tags, exchange);
         String response = agent.chat(aiAgentBody, toolProvider);
-        exchange.getMessage().setBody(response);
+
+        Class<?> responseType = endpoint.getConfiguration().getResponseType();
```
Croway: **`responseType` parsing runs even when the schema was never applied to the model.** This lookup makes `process()` parse with `responseType` regardless of whether `resolveAndApplyStructuredOutput()` actually ran in `doInit()`. Two reachable configurations bypass it:

1. `agent` + `agentConfiguration` both set — both options are `autowired=true`, so a registry containing one bean of each auto-injects both. The `agent != null` branch wins and the `ResponseFormat` is never applied, yet the startup guard passes because `agentConfiguration != null`.
2. `agentFactory` + `agentConfiguration` + `responseType` — the factory-created agent replaces the internal agent on every exchange and never receives the `ResponseFormat` that `doInit()` applied to the (discarded) internal agent.

In both cases the model gets no JSON schema, answers in free-form prose, and `SERVICE_OUTPUT_PARSER.parseText` throws `OutputParsingException` on every message at runtime instead of failing fast at startup. Suggest extending the `doInit()` validation to reject `responseType` when `agent` or `agentFactory` is configured (or fixing it at the root — see the comment on `setResponseFormat` below).

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -72,15 +78,28 @@ public LangChain4jAgentProducer(LangChain4jAgentEndpoint endpoint) {
...
+                && endpoint.getConfiguration().getAgentConfiguration() == null) {
+            throw new IllegalArgumentException(
+                    "responseType requires agentConfiguration to be set. "
+                                               + "It only works when Camel creates the agent internally (inline agent creation mode).");
+        }
+
         if (endpoint.getConfiguration().getAgent() != null) {
             agent = endpoint.getConfiguration().getAgent();
         } else if (endpoint.getConfiguration().getAgentConfiguration() != null) {
+            if (ObjectHelper.isNotEmpty(endpoint.getConfiguration().getJsonSchema())
```
Croway: **The mutual-exclusion check is unreachable when an `agent` bean is also configured.** Because this check is nested inside the `else if (agentConfiguration != null)` branch, an endpoint with `agent=#myAgent&agentConfiguration=#cfg&jsonSchema=...&responseType=...` starts up cleanly with both "mutually exclusive" options set — `jsonSchema` is silently dropped while `responseType` still drives response parsing. Moving the check to the top-level validation in `doInit()` (alongside the `responseType`-requires-`agentConfiguration` guard) would make the documented contract hold on every path.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -305,41 +330,55 @@ private Map<String, CamelToolSpecification> discoverToolsByName(String tags) {
...
-        if (ObjectHelper.isEmpty(jsonSchema)) {
+        Class<?> responseType = endpoint.getConfiguration().getResponseType();
+
+        if (ObjectHelper.isEmpty(jsonSchema) && responseType == null) {
             return;
         }
 
-        String resolved = endpoint.getCamelContext().resolvePropertyPlaceholders(jsonSchema);
+        if (responseType != null) {
+            JsonSchema schema = JsonSchemas.jsonSchemaFrom(responseType)
```
Croway: **Non-POJO `responseType` values fail with a misleading error message.** In langchain4j 1.16.2, `JsonSchemas.jsonSchemaFrom()` returns `Optional.empty()` for any type whose output parser is not `PojoOutputParser` — `String`, `Integer`, enums, `List`/`Set`, primitives. So e.g. `responseType=java.lang.Integer` fails startup with "Ensure the class has public fields or getters" even though `Integer` has getters; the real restriction is POJO-only. Suggest rewording the message (e.g. "responseType must be a POJO class; simple types, enums and collections are not supported") so users aren't sent chasing the wrong fix.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -305,41 +330,55 @@ private Map<String, CamelToolSpecification> discoverToolsByName(String tags) {
...
+                    .orElseThrow(() -> new IllegalArgumentException(
+                            "Cannot derive JSON schema from responseType: " + responseType.getName()
+                                                                    + ". Ensure the class has public fields or getters and is not a raw generic type."));
+            ResponseFormat derivedFormat = ResponseFormat.builder()
+                    .type(ResponseFormatType.JSON)
+                    .jsonSchema(schema)
```
Croway: **Schema name diverges from the documented `camel_schema` convention.** `JsonSchemas.jsonSchemaFrom()` names the schema `rawClass.getSimpleName()` (verified in langchain4j 1.16.2 sources), while the `jsonSchema` branch below deliberately uses the fixed name `camel_schema` with a comment citing cross-component consistency with camel-openai. Migrating a route from `jsonSchema` to `responseType` therefore silently changes the `json_schema.name` sent to OpenAI-compatible providers from `camel_schema` to e.g. `CarRentalRecommendation`. Consider rebuilding the derived schema with `.name("camel_schema")` to keep both options consistent.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -102,7 +121,13 @@ public void process(Exchange exchange) throws Exception {
 
         ToolProvider toolProvider = createComposedToolProvider(tags, exchange);
         String response = agent.chat(aiAgentBody, toolProvider);
-        exchange.getMessage().setBody(response);
+
+        Class<?> responseType = endpoint.getConfiguration().getResponseType();
+        if (responseType != null) {
+            exchange.getMessage().setBody(SERVICE_OUTPUT_PARSER.parseText(responseType, response));
```
Croway: **No error handling around parsing — raw response is lost on failure.** If the model returns a null/blank/unparseable response (content filter, guardrail rewrite, provider quirk), `parseText` throws a raw `dev.langchain4j.service.output.OutputParsingException` and the original response text is gone — `onException` handlers can't log it or route a fallback. Consider wrapping in a Camel-friendly exception that carries the raw response (or preserving it in a header) so routes keep an escape hatch.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -56,6 +58,10 @@
 
 public class LangChain4jAgentProducer extends DefaultProducer {
     private static final Logger LOG = LoggerFactory.getLogger(LangChain4jAgentProducer.class);
+    // ServiceOutputParser is used intentionally: beyond JSON parsing, it strips markdown code fences
+    // (e.g. ```json ... ```) that some LLMs emit even when ResponseFormat is explicitly set to JSON.
+    // Plain Jackson would throw a JsonParseException on such responses.
+    private static final ServiceOutputParser SERVICE_OUTPUT_PARSER = new ServiceOutputParser();
```
Croway: Heads-up: `ServiceOutputParser` is annotated `@Internal` in langchain4j 1.16.2 (verified via `javap`; `JsonSchemas` by contrast is public API), and this is the only place in Camel that uses it. The code-fence-stripping rationale in the comment is a fair tradeoff, but its signature can change without deprecation on any langchain4j upgrade — worth keeping in mind when bumping the dependency.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -305,41 +330,55 @@ private Map<String, CamelToolSpecification> discoverToolsByName(String tags) {
...
+        if (responseType != null) {
+            JsonSchema schema = JsonSchemas.jsonSchemaFrom(responseType)
+                    .orElseThrow(() -> new IllegalArgumentException(
+                            "Cannot derive JSON schema from responseType: " + responseType.getName()
+                                                                    + ". Ensure the class has public fields or getters and is not a raw generic type."));
+            ResponseFormat derivedFormat = ResponseFormat.builder()
+                    .type(ResponseFormatType.JSON)
+                    .jsonSchema(schema)
+                    .build();
+            ((AbstractAgent<?>) agent).setResponseFormat(derivedFormat);
```
Croway: **Design suggestion:** rather than the cast+setter, a `responseFormat` field on `AgentConfiguration` consumed by `AbstractAgent.configureBuilder()` (which already sources guardrails, tools, RAG and memory from the configuration) would let factory-created agents and user-provided `AbstractAgent` subclasses receive the format too — removing the "requires agentConfiguration" restriction at the root and fixing the agentFactory gap noted above. The cast pattern pre-dates this PR (CAMEL-23642 introduced it for `jsonSchema`), but this PR adds a second cast site, so it may be the right moment to generalize.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -305,41 +330,55 @@ private Map<String, CamelToolSpecification> discoverToolsByName(String tags) {
...
 
     /**
-     * Resolves the jsonSchema endpoint property: loads from classpath/filesystem if it is a resource URI, otherwise
-     * treats it as inline JSON. Converts the raw JSON string into a langchain4j ResponseFormat and sets it on the
-     * agent. Only called when the agent is created internally from agentConfiguration.
+     * Resolves and applies structured output configuration (jsonSchema or responseType) to the agent. Only called when
+     * the agent is created internally from agentConfiguration.
      */
-    private void resolveAndApplyJsonSchema() throws Exception {
+    private void resolveAndApplyStructuredOutput() throws Exception {
```
Croway, line 334: Minor: both branches of this method duplicate the `ResponseFormat.builder().type(JSON).jsonSchema(...).build()` + `((AbstractAgent<?>) agent).setResponseFormat(...)` epilogue, and the feature's validation is spread over three places (top of `doInit`, inside the `agentConfiguration` branch, and the early-return here). Deriving the `JsonSchema` per branch and building/applying the format once — plus one consolidated validation block — would prevent drift; the two branches have already diverged on schema naming.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -102,7 +121,13 @@ public void process(Exchange exchange) throws Exception {
 
         ToolProvider toolProvider = createComposedToolProvider(tags, exchange);
```
Croway, line 126: Pre-existing issue adjacent to this change (not introduced by this PR, noting for a follow-up JIRA): a few lines above, `agent = agentFactory.createAgent(exchange)` writes the shared instance field from `process()` on a singleton producer, so concurrent exchanges race — thread A can end up chatting with thread B's factory-created agent (cross-tenant memory/credential leakage with per-tenant factories). The fix is a local variable: `Agent agent = agentFactory != null ? agentFactory.createAgent(exchange) : this.agent;`.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentConfiguration.java
```diff
@@ -58,10 +58,18 @@ public class LangChain4jAgentConfiguration implements Cloneable {
...
+                            + "Mutually exclusive with responseType.",
               supportFileReference = true, largeInput = true, inputLanguage = "json")
     private String jsonSchema;
 
+    @UriParam
+    @Metadata(description = "Java class to use as response format for structured outputs. "
+                            + "Camel will automatically derive the JSON schema from the class structure and unmarshal the response. "
+                            + "This option works only when using agentConfiguration (inline agent creation mode). "
+                            + "Mutually exclusive with jsonSchema.")
+    private Class<?> responseType;
```
Croway: Consistency note: camel-openai exposes the equivalent feature as `outputClass` — a `String` resolved at request time via `ClassResolver`, overridable per message via the `CamelOpenAIOutputClass` header, with the body left as the raw JSON string. This option differs in name, type, dynamicity and body semantics (auto-unmarshalled POJO), even though the PR borrows camel-openai's `camel_schema` convention. A `Class<?>` bound at startup also bypasses the `ClassResolver` indirection, which matters in classloader-partitioned runtimes. Worth aligning the surface (or at least documenting the difference) so users moving between camel-ai components aren't surprised.

Croway (commented):
Automated code review of this PR — _Claude Code on behalf of Federico Mariani (@croway)_.

The feature works well on the tested path (inline `agentConfiguration` + POJO `responseType`). The main theme of the inline findings: the new validation only covers one of the three agent-creation paths (`agent` bean, `agentConfiguration`, `agentFactory`), so several supported configurations end up with `responseType` parsing enabled while the JSON schema was never applied to the model — failing per-message at runtime instead of fast at startup. There are also two smaller correctness/consistency issues in the schema-derivation branch (misleading error for non-POJO types, schema-name divergence from the documented `camel_schema` convention) and a few design/cleanup notes.

Details inline. One finding (producer field race) is pre-existing and only noted for a follow-up JIRA.

zbendhiba:
Thanks for the detailed review @Croway.

Here's what I addressed:

**Renamed `responseType` to `outputClass`** to align with `camel-openai`. The body is now a raw JSON string. No auto-unmarshalling. Users add `.unmarshal().json(MyClass.class)` in their route. This also removed `ServiceOutputParser` (which was `@Internal`) and the parse-failure issue where the original response was lost on exception.

**Fixed validation in `doInit()`**. The guard now enforces all three conditions when `outputClass` or `jsonSchema` is set: `agentConfiguration` must be present, `agent` must be absent, `agentFactory` must be absent. The mutual-exclusion check moved to the top so it fires on all paths.

**Fixed the error message** for non-POJO types. It now says "outputClass must be a POJO class with public fields or getters. Simple types, enums, and collections are not supported."

**Fixed schema naming**. I rebuild the derived schema with `.name("camel_schema")` so both `outputClass` and `jsonSchema` send the same name to the model.

**Removed the code duplication** in `resolveAndApplyStructuredOutput()`. Both branches now derive only their `JsonSchema`. The `ResponseFormat` build and `setResponseFormat` call happen once at the bottom.

**Fixed the race condition** on the `agent` field. `agentFactory.createAgent(exchange)` was writing a shared field from `process()` on a singleton producer. Fixed with a local variable.

**Added two unit tests** for the newly guarded configs: `outputClass` + `agent` bean, and `outputClass` + `agentFactory`.

**Updated the docs** to reflect the rename, raw JSON body semantics, and the three-condition constraint.

On the `AgentConfiguration.configureBuilder()` suggestion: I kept `outputClass` intentionally consistent with `jsonSchema`. Both have the same restriction. Generalizing both through `configureBuilder()` is a broader change. I think it deserves its own JIRA rather than expanding this PR.

One remaining difference from `camel-openai`: `outputClass` is a `Class<?>` bound at startup, not a `String` resolved per-request via `ClassResolver` with a header override. That's a deliberate choice for now. Happy to open a follow-up if you think full alignment is worth it.

_Claude Code on behalf of Zineb Bendhiba_

davsclaus (commented):
Thanks for adding the `outputClass` option, Zineb — nice feature that makes structured output much more accessible without hand-writing JSON schemas.

A few observations:

**Naming inconsistency — test names say "responseType" but the parameter is `outputClass`**

The new integration test class is named `LangChain4jAgentResponseTypeIT`, and several new unit test methods in `LangChain4jAgentStructuredOutputTest` use "responseType" in their names (e.g., `testResponseTypeWithoutAgentConfigurationThrowsIllegalArgumentException`, `testNonPojoResponseTypeThrowsIllegalArgumentException`). The actual endpoint option is `outputClass`. Looks like a rename happened during development — would be good to align the test names for maintainability.

**Thread-safety fix in `process()` bundled but undocumented**

The change from `this.agent = agentFactory.createAgent(exchange)` (mutating the shared instance field) to `Agent agent = agentFactory != null ? ... : this.agent` (local variable) fixes a real thread-safety issue under concurrent exchange processing. This is a good improvement but it's a separate bug fix — worth mentioning in the PR description so reviewers and future git-blame readers understand the intent.

**PR description**

The body is currently just the template checkboxes. A brief description of the feature, its constraints (inline-agent-only, mutually exclusive with jsonSchema), and the thread-safety fix would help reviewers.

Everything else looks solid: validation logic is thorough, documentation is clear with good examples, generated files are committed, and test coverage is good.

components/camel-ai/camel-langchain4j-agent/src/test/java/org/apache/camel/component/langchain4j/agent/integration/LangChain4jAgentOutputClassIT.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration test for outputClass parameter with the langchain4j-agent component.
+ */
+@DisabledIfSystemProperty(named = "ci.env.name", matches = ".*", disabledReason = "Requires too much network resources")
```
davsclaus, line 39: Nit: The class name uses `ResponseType` but the actual endpoint parameter is `outputClass`. Consider renaming to `LangChain4jAgentOutputClassIT` (or similar) to match the parameter name and avoid confusion for future maintainers.

components/camel-ai/camel-langchain4j-agent/src/main/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducer.java
```diff
@@ -94,9 +119,7 @@ public void process(Exchange exchange) throws Exception {
         // tags for Camel Routes as Tools
```
davsclaus, line 119: Good fix — the old code mutated the shared `agent` field on every exchange when using `agentFactory`, which is a thread-safety issue under concurrent processing. Using a local variable is the right approach.

This is a separate bug fix from the `outputClass` feature though — worth calling out in the PR description so it doesn't get lost.

oscerd (approved):
Looks good — the new optional `outputClass` parameter is cleanly implemented and fully wired: all generated files (catalog JSON, component docs, endpoint-dsl and component-dsl builders, configurers) are regenerated and committed, the option has a thorough description, CI is green, and there's solid coverage (the validation branches as unit tests plus the Ollama ITs gated off CI). I also like that it quietly fixes a real concurrency bug — moving the per-request `agent` from the shared producer field to a method-local.

Minor, non-blocking: the PR title and the `LangChain4jAgentResponseTypeIT` filename still say "responseType", but the shipped param is `outputClass` (renamed for consistency with camel-openai) — worth aligning the title/IT name to avoid confusion.

Nice work 👍

---

zbendhiba:
Thanks @davsclaus @davsclaus  

I pushed a fix ro rename test class and its methods / variables

davsclaus (approved):

zbendhiba:
I think we are fine for this PR, I can merge this! Many thanks for all these reviews!

PR #23997, opened 2026-06-12
CAMEL-21438: Fix and re-enable flaky tests on s390x

ammachado:
# Description

Fix and re-enable tests that were disabled on s390x (and some on all CI environments) due to flakiness. Fixes fall into four categories: replacing \`Thread.sleep\` with deterministic waits, tightening timeouts, removing stale architecture exclusions, and fixing a structural permit-leak bug.

## Commits in this PR

**1. Replace \`Thread.sleep\` with deterministic waits (original batch)**

- \`MainListenerTest\`: background \`main.run()\` thread was polled with a 100ms sleep; replaced with a \`CountDownLatch\` that fires on the \`beforeInitialize\` event.
- \`ManagedThrottlingExceptionRoutePolicyTest\`: 200ms sleep before asserting JMX state replaced with \`Awaitility.await().atMost(10s)\`.
- \`AsyncCompletionServiceTest\`: redundant 300ms sleep before a blocking \`take()\` call removed entirely.

**2. Fix s390x-specific timeouts and races**

- \`AggregateCompleteAllOnStopTest\`: fixed a race between route stop and mock assertion.
- \`FileNoOpLockFileTest\`: 1s polling timeout too tight; increased.
- \`FileConsumerResumeFromOffsetStrategyTest\`: 2s wait replaced with Awaitility.
- \`BarcodeDataFormatSpringTest\`: flaky due to timing; fixed with deterministic wait.

**3. Re-enable \`XsltCustomizeEntityResolverTest\` on s390x**

The \`@DisabledOnOs(s390x)\` annotation was a workaround for a file-consumer race condition that CAMEL-23189 already fixed (replaced \`file:\` route with \`direct:\`). Annotation removed. Also simplified the \`EntityResolver\` anonymous class to a lambda.

**4. Fix \`ThrottlerTest\` flakiness and re-enable on s390x**

Three root causes fixed:
- \`testSendLotsOfMessagesButOnly3GetThroughWithin2Seconds\`: raised \`setResultWaitTime\` from 2s to 10s and added an explicit counter assertion to decouple the delivery timeout from the throttle-window check.
- \`assertThrottlerTiming\`: dropped the minimum elapsed-time bound, which tests machine speed rather than throttle correctness.
- \`sendMessagesWithHeaderExpression\`: start the elapsed timer from a \`CountDownLatch\` fired by the first executing thread, not from task submission, to avoid inflating elapsed with thread pool scheduling overhead.

**5. Fix \`ConcurrentRequestsThrottlerTest\` permit leak and re-enable on s390x**

The test semaphore acquire and release lived in two separate processors with a \`delay(100)\` between them. If an exchange failed between them, the release never ran and the permit leaked, causing later \`tryAcquire\` calls to fail spuriously. Fixed by registering an \`onCompletion\` callback inside the acquire processor so the permit is released on both success and failure. The standalone release processor is removed from all three routes and from the Spring XML counterpart (\`SpringThrottlerTest\` / \`throttler.xml\`).

# Target

# Tracking

---

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
@@ -23,9 +23,11 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.ContextTestSupport;
+import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.processor.ThrottlerRejectedExecutionException;
+import org.apache.camel.support.SynchronizationAdapter;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
@@ -35,10 +37,9 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 // time-bound that does not run well in shared environments
 @EnabledOnOs(value = { OS.LINUX, OS.MAC, OS.FREEBSD, OS.OPENBSD },
-             architectures = { "amd64", "aarch64", "ppc64le" },
-             disabledReason = "This test does not run reliably multiple platforms (see CAMEL-21438)")
+             architectures = { "amd64", "aarch64", "ppc64le", "s390x" },
+             disabledReason = "This test does not run reliably on all platforms (see CAMEL-21438)")
 public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
-    private static final int INTERVAL = 500;
     private static final int MESSAGE_COUNT = 9;
     private static final int CONCURRENT_REQUESTS = 2;
     protected static Semaphore semaphore;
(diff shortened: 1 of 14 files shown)


core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
```diff
@@ -199,32 +190,59 @@ public void configure() {
...
+                                    s.release();
+                                }
+
+                                @Override
+                                public void onFailure(Exchange ex) {
+                                    s.release();
+                                }
+                            });
                         })
                         .delay(100)
```
davsclaus: `SynchronizationAdapter.onComplete()` and `onFailure()` both delegate to `onDone()` (see `SynchronizationAdapter.java:29-38`), so overriding just `onDone()` achieves the same result with less code. Same applies to the other two routes in this file and `SpringThrottlerTest.IncrementProcessor`.

```suggestion
                            exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                @Override
                                public void onDone(Exchange ex) {
                                    s.release();
                                }
                            });
```

davsclaus (commented):
Nice work fixing these flaky tests — the permit-leak fix in the throttler tests is a particularly good catch.

One minor simplification suggestion below.

This review checks project rules and conventions only. It does not replace specialized review tools or static analysis.

PR #24005, opened 2026-06-14
CAMEL-23640: camel-jbang - add restart command to infra service

sk-reddy17:
## Summary

  - Adds a new `camel infra restart <service>` command that stops any running instance of the named external service and starts a fresh one
  - Supports the same options as `camel infra run`: `--background`, `--port`, `--log`, and `--kill` (for forcibly stopping the existing instance)
  - Fixes `InfraBaseCommand.findPids` to handle a missing `~/.camel` directory gracefully (was crashing with `NoSuchFileException` on a clean machine)
  - Regenerates metadata and docs to include the new subcommand
  - Adds upgrade guide entry

  Fixes https://issues.apache.org/jira/browse/CAMEL-23640

  ## Usage

  ```bash
  # Foreground (blocks until stopped)
  camel infra restart kafka

  # Background (returns immediately)
  camel infra restart kafka --background

  # Force-kill the existing instance before restarting
  camel infra restart kafka --kill

  When no instance is running, restart simply starts the service (same as camel infra run).
  ```                                                                                                                                                                                                            
  Test plan

  - [x] InfraRestartTest.restartService — restart with nothing running → starts the service
  - [x] InfraRestartTest.restartStopsExistingService — restart with a fake running PID file → prints stopping message then starts fresh
  - [x] Manual end-to-end test: started FTP via camel infra run ftp, then ran camel infra restart ftp — confirmed "Stopping external service ftp (PID: ...)" followed by a fresh start


davsclaus:
the upgrade guide should not include this - as this is a new feature (the guide is big already)

sk-reddy17:
> the upgrade guide should not include this - as this is a new feature (the guide is big already)

Done, removed it. Thanks for the feedback!

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraRestart.java
```diff
@@ -0,0 +1,124 @@
...
+        for (Long pid : pids.keySet()) {
+            ProcessHandle.of(pid).ifPresent(ph -> {
+                try {
+                    ph.onExit().get(10, TimeUnit.SECONDS);
+                } catch (Exception e) {
+                    // ignore and proceed to start the service again
+                }
+            });
+        }
+
```
davsclaus, line 86: The background path here re-implements the command-list building and process spawning that already exists in `InfraRun.runBackground()`. The foreground path nicely delegates to `InfraRun.doCall()` — could the background path do the same by setting `infraRun.background = true` and letting `InfraRun` handle it?

This would avoid drift if `InfraRun.runBackground()` evolves (e.g., new options, different command construction).

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/infra/InfraBaseCommand.java
```diff
@@ -103,6 +103,8 @@ protected static Map<Long, Path> findPids(String name) throws Exception {
                     pids.put(Long.valueOf(pid), pidFile);
                 }
             }
+        } catch (IOException e) {
```
davsclaus: Good fix for the missing `~/.camel` directory crash. Consider narrowing to `NoSuchFileException` instead of all `IOException` — a permission error or disk I/O failure would be silently swallowed here.

```suggestion
        } catch (java.nio.file.NoSuchFileException e) {
            // ignore - camel directory does not exist yet
```

davsclaus (commented):
Thanks for the contribution, nice addition to the infra tooling!

A few observations below — nothing blocking, mostly about code reuse and robustness.

davsclaus (approved):

PR #24030, opened 2026-06-15
CAMEL-23759: camel-spring-ws - apply a HeaderFilterStrategy to inbound SOAP headers

oscerd:
## Description

The `spring-ws` consumer maps inbound SOAP header element/attribute names straight into the Camel `Exchange` header map. This aligns `camel-spring-ws` with the rest of the component catalog (`camel-cxf`, `camel-mail`, `camel-coap`, ...) by applying a `HeaderFilterStrategy` on the inbound path.

- Adds a configurable `headerFilterStrategy` endpoint option (`common` label).
- Defaults to a new `SpringWebserviceHeaderFilterStrategy` that filters the `Camel*` / `camel*` namespace (case-insensitive) in both the inbound and outbound directions, mirroring `CoAPHeaderFilterStrategy` / `IggyHeaderFilterStrategy`.
- Inbound SOAP header element/attribute names in that namespace are no longer promoted to `Exchange` headers.

## Breaking change

Documented in the 4.21 upgrade guide. A custom `headerFilterStrategy` restores the previous behaviour.

## Testing

Adds `ConsumerSoapHeaderFilterTest`, which drives the consumer with a crafted SOAP header and asserts a regular header element is still propagated while an internal `Camel*` header is filtered. The full `camel-spring-ws` module suite passes (92 tests, 0 failures).


davsclaus (commented):
Good security-hardening work — this aligns `camel-spring-ws` with the header-filtering pattern already in place across the component catalog (Kafka, CXF, mail, CoAP, etc.).

A couple of observations below, nothing blocking.

Note: this review does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud.

components/camel-spring-parent/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/SpringWebserviceConfiguration.java
```diff
@@ -420,6 +424,19 @@ public void setMessageIdStrategy(MessageIdStrategy messageIdStrategy) {
...
+     * To use a custom HeaderFilterStrategy to filter headers mapped to and from the Camel message. By default the
+     * internal {@code Camel*} and {@code org.apache.camel.*} header namespace is filtered out from inbound SOAP
+     * headers.
+     */
+    public void setHeaderFilterStrategy(HeaderFilterStrategy headerFilterStrategy) {
+        this.headerFilterStrategy = headerFilterStrategy;
+    }
+
     public boolean isAllowResponseHeaderOverride() {
         return allowResponseHeaderOverride;
```
davsclaus, line 441: Minor: the description says _"the internal `Camel*` and `org.apache.camel.*` header namespace is filtered out"_ — this reads like two separate prefix filters, but `CAMEL_FILTER_STARTS_WITH` is `["Camel", "camel"]`. With `lowerCase=true`, this does cover `org.apache.camel.*` (because it starts with `camel` when lowercased), so the behavior is correct. But the description could be clearer — something like:

```suggestion
     * To use a custom HeaderFilterStrategy to filter headers mapped to and from the Camel message. By default the
     * internal {@code Camel} and {@code camel} header namespace (case-insensitive) is filtered out from inbound SOAP
     * headers.
```

components/camel-spring-parent/camel-spring-ws/src/test/java/org/apache/camel/component/spring/ws/ConsumerSoapHeaderFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+        webServiceTemplate = applicationContext.getBean("webServiceTemplate", WebServiceTemplate.class);
+    }
+
+    @Test
+    public void internalCamelHeaderFromInboundSoapHeaderIsFiltered() throws Exception {
+        result.expectedMessageCount(1);
+
+        Source source = new StreamSource(new StringReader(xmlRequest));
+        webServiceTemplate.sendSourceAndReceive(source, message -> {
+            SoapMessage soap = (SoapMessage) message;
```
davsclaus, line 62: The test covers SOAP header *elements* (`addHeaderElement`). The consumer code also filters SOAP header *attributes* (`getAllAttributes` path at line 188-195). Consider adding a test case that sets a `CamelFoo` attribute via `soapHeader.addAttribute(new QName(..., "CamelFoo"), "...")` to cover both code paths.

components/camel-spring-parent/camel-spring-ws/src/main/java/org/apache/camel/component/spring/ws/SpringWebserviceConfiguration.java
oscerd: Applied — updated the javadoc to reference the `Camel` / `camel` (case-insensitive) namespace, matching `CAMEL_FILTER_STARTS_WITH`. Thanks for the catch. — Claude Code on behalf of Andrea Cosentino

components/camel-spring-parent/camel-spring-ws/src/test/java/org/apache/camel/component/spring/ws/ConsumerSoapHeaderFilterTest.java
oscerd: Good catch — added `internalCamelAttributeFromInboundSoapHeaderIsFiltered` covering the `getAllAttributes` path: a `CamelFileName` attribute is asserted filtered while a normal attribute is still propagated. — Claude Code on behalf of Andrea Cosentino

davsclaus (approved):

oscerd:
Pushed two follow-ups for the red CI: adding the `headerFilterStrategy` `@UriParam` regenerates the catalog mirror **and** the endpoint-DSL builder (`SpringWebserviceEndpointBuilderFactory`), neither of which a module-scoped build touches — both are now committed, so the "uncommitted changes" check should pass. — Claude Code on behalf of Andrea Cosentino

PR #24031, opened 2026-06-15
CAMEL-23757: Add Java DSL model writer for full-circle DSL transformation

davsclaus:
## Summary

- Add a new `core/camel-java-io` module that generates fluent Java DSL source code from Camel model definitions
- Enables full-circle DSL transformation: XML/YAML → Model → Java DSL (and back via existing parsers)
- Follows the same code-generation pattern as existing `camel-xml-io` and `camel-yaml-io` modules

## Key features

- **101 compile-verified test cases** — every generated Java DSL snippet compiles against the real Camel API
- **`@DslArg` annotation** on model fields drives primary argument rendering (position, type, enum/class literals)
- **Idiomatic output** for all major EIP patterns:
  - `from("direct:start")` / `to("mock:result")` / `to(ExchangePattern.InOut, "uri")`
  - `bean("myRef", "method")` / `claimCheck(ClaimCheckOperation.Set, "key")` / `threads(5, 10)`
  - `log(LoggingLevel.WARN, "message")` / `throwException(IllegalArgumentException.class, "msg")`
  - `filter(simple("..."))` / `choice().when(...)` / `split(body())` / block EIPs with `.end()`
  - `marshal(dataFormat().json().library("Jackson").end())` — full data format builder pattern
  - `circuitBreaker().resilience4jConfiguration().failureRateThreshold("30").end()` — sub-builders
  - `saga().option("key", simple("..."))` — PropertyExpressionDefinition handling
  - Toggle attributes: `.streaming()` / `.eagerCheckCompletion()` (no-arg when true, omitted when false)
  - Boolean attributes: `.parallelProcessing(true)` (unquoted)
  - Error handlers, intercepts, onException, onCompletion, load balancers
- **Property placeholder support** — String overloads added to Resilience4j, FaultTolerance, and Sampling config classes

## Test plan

- [x] All 101 XML to Java DSL compile tests pass
- [x] Tests cover: simple routes, expressions, EIPs, data formats, error handlers, intercepts, circuit breakers, saga, throttle, threads, bean, claimCheck, log with level, to with pattern


core/camel-core-model/src/main/java/org/apache/camel/model/ToDefinition.java
```diff
@@ -40,6 +41,7 @@ public class ToDefinition extends SendDefinition<ToDefinition> {
     private String variableReceive;
     @XmlAttribute
     @Metadata(label = "advanced", javaType = "org.apache.camel.ExchangePattern", enums = "InOnly,InOut")
+    @DslArg(position = 1, renderType = "enumString", typeName = "ExchangePattern")
```
gnodet: 🔴 **Bug: `to()` parameter ordering is reversed**

With `uri` at position 0 (inherited from `SendDefinition`) and `pattern` at position 1, the generated code produces:
```java
.to("direct:foo", ExchangePattern.InOut)
```

But `ProcessorDefinition` only has `to(ExchangePattern pattern, String uri)` — pattern first, uri second. There is no `to(String, ExchangePattern)` overload, so this would fail to compile.

Fix: either swap positions (`pattern` = 0, `uri` = 1), or — since `to(uri)` is by far the most common form and `to(pattern, uri)` is rare — consider whether the pattern should even be a primary arg at all (it could just be a chained attribute: `.to("uri").pattern(ExchangePattern.InOut)`).

Not caught by tests because no XML test file has `<to uri="..." pattern="..."/>`.

core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
```diff
@@ -42,6 +43,7 @@ public abstract class SendDefinition<Type extends ProcessorDefinition<Type>> ext
 
     @XmlAttribute
     @Metadata(required = true)
+    @DslArg
```
gnodet: Note: this `@DslArg` (position=0 by default) is inherited by `ToDefinition`, `ToDynamicDefinition`, `WireTapDefinition`, etc. The position interplay with `ToDefinition.pattern` (position=1) produces a reversed argument order — see comment on `ToDefinition.java`.

gnodet (commented):
## Review of `camel-java-io` Java DSL Model Writer

Nice work on the overall architecture — the annotation-driven approach with `@DslArg` is clean, the Velocity template is well structured, and `JavaDslCompileTest` (parameterized over all XML test files) is an excellent quality gate.

A few observations below, keeping in mind this is still a draft.

### `@DslArg` naming
The name is good. It follows the existing `{domain}{role}` convention in `spi-annotations` (`YamlIn`, `YamlOut`, `YamlProperty`). Alternatives like `@JavaDslArg` over-specialize, `@DslParameter` collides with URI parameters. 👍

### `position` attribute
The `position` attribute is justified — field reordering is not a safe alternative because:
- Inherited fields (e.g. `SendDefinition.uri`) can't be reordered relative to subclass fields without breaking the abstraction
- JAXB `@XmlAccessorType(FIELD)` without explicit `propOrder` may use declaration order — reordering fields could change XML output
- `position` makes the DSL contract explicit and resilient to refactors

### Missing EIPs
Several commonly-used EIPs fall back to `beginStep()` (empty parens) and produce less-than-idiomatic output:
- **`bean`** (high priority): generates `.bean().ref("myRef")` instead of `.bean("myRef", "method")`
- **`claimCheck`**: generates `.claimCheck().operation("Set")` instead of `.claimCheck(ClaimCheckOperation.Set)`
- **`threads`**: generates `.threads().poolSize("5")` instead of `.threads(5, 10)`

These could be addressed in follow-up PRs.

### Boolean attributes
Boolean-like attributes are rendered as `.streaming("true")` instead of the idiomatic `.streaming()` (no-arg toggle). This compiles (there are `String` overloads) but isn't what a human would write. Low priority.

### Test gaps
- No test exercises `to()` with `pattern` set — would have caught the bug below
- No test for `throwException()` with both `exceptionType` and `message`
- No test for `log()` with `loggingLevel` set
- No test for error handler rendering

### Thread safety
The writer is not thread-safe (mutable `indentLevel` / `handledAttributes`), which is fine for its use case. Worth a Javadoc note on the class.

core/camel-core-model/src/main/java/org/apache/camel/model/ToDefinition.java
davsclaus: Fixed in 6991fa5 — swapped positions: `pattern` → position 0, `uri` → position 1. Now generates `.to(ExchangePattern.InOut, "mock:result")` matching `ProcessorDefinition.to(ExchangePattern, String)`. Added `toWithPattern.xml` test to cover this case (97/97 tests pass).

core/camel-core-model/src/main/java/org/apache/camel/model/SendDefinition.java
davsclaus: Fixed in the same commit (6991fa5) — `SendDefinition.uri` is now `@DslArg(position = 1)` so subclasses like `ToDefinition` that add a position-0 arg (pattern) get the correct ordering.

PR #24032, opened 2026-06-15
CAMEL-23760: camel-oauth - require a JWK set to verify token signatures in UserProfile

oscerd:
## Description

The `camel-oauth` `UserProfile` token verification skipped the JWS signature check when the configured JWK set was missing or empty, leaving the signature unverified in that case. This makes the signature check mandatory:

- When no JWK set is available (`null` or empty), the token is now rejected with an `OAuthException` rather than accepted.
- Deployments with a correctly resolved JWK set are unaffected.
- This aligns the legacy `UserProfile` path with the `JwtTokenValidator` SPI path, which already fails closed on this condition.

## Behaviour change

Documented in the 4.21 upgrade guide.

## Testing

Adds `UserProfileTest` with three cases: rejection on an empty JWK set, rejection on a missing JWK set, and acceptance when the signature verifies against a matching key.

### DIFF
--- a/components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
+++ b/components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
@@ -178,17 +178,18 @@ public class UserProfile {
             var signedJWT = SignedJWT.parse(token);
             var keyID = signedJWT.getHeader().getKeyID();
 
-            // Fetch Keycloak public key
+            // Resolve the signing key from the configured JWK set and verify the token signature
             var jwkSet = config.getJWKSet();
-            if (!jwkSet.isEmpty()) {
-                var rsaKey = (RSAKey) jwkSet.getKeyByKeyId(keyID);
-                if (rsaKey == null) {
-                    throw new OAuthException("No matching key found for: " + keyID);
-                }
-                RSAPublicKey publicKey = rsaKey.toRSAPublicKey();
-                if (!signedJWT.verify(new RSASSAVerifier(publicKey))) {
-                    throw new OAuthException("Invalid token signature");
-                }
+            if (jwkSet == null || jwkSet.isEmpty()) {
+                throw new OAuthException("Cannot verify token signature: no JWK set available");
+            }
+            var rsaKey = (RSAKey) jwkSet.getKeyByKeyId(keyID);
+            if (rsaKey == null) {
+                throw new OAuthException("No matching key found for: " + keyID);
+            }
+            RSAPublicKey publicKey = rsaKey.toRSAPublicKey();
+            if (!signedJWT.verify(new RSASSAVerifier(publicKey))) {
+                throw new OAuthException("Invalid token signature");
             }
 
             // Decode the payload into a JsonObject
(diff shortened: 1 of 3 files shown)


components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
```diff
@@ -180,15 +180,16 @@ private static JsonObject verifyToken(OAuthConfig config, String token, boolean
 
             // Fetch Keycloak public key
```
davsclaus: Nit: this comment is a leftover from when the code was Keycloak-specific. The verification is now provider-agnostic (works with any OIDC IdP). Consider updating:

```suggestion
            // Verify JWS signature against configured JWK set
```

davsclaus (commented):
Nice security hardening — the fail-closed approach aligns `UserProfile` with the `JwtTokenValidator` SPI path and closes a real gap where tokens could be accepted without signature verification.

No blocking issues found. Two minor suggestions inline.

Note: CI checks have not reported yet on this branch — please verify green before merging.

components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
```diff
@@ -180,15 +180,16 @@ private static JsonObject verifyToken(OAuthConfig config, String token, boolean
...
-            if (!jwkSet.isEmpty()) {
-                var rsaKey = (RSAKey) jwkSet.getKeyByKeyId(keyID);
-                if (rsaKey == null) {
-                    throw new OAuthException("No matching key found for: " + keyID);
-                }
-                RSAPublicKey publicKey = rsaKey.toRSAPublicKey();
-                if (!signedJWT.verify(new RSASSAVerifier(publicKey))) {
-                    throw new OAuthException("Invalid token signature");
-                }
+            if (jwkSet == null || jwkSet.isEmpty()) {
```
davsclaus, line 183: Good catch adding the `null` check here — the old code would have thrown an NPE if `getJWKSet()` returned `null` (which it can, since `OAuthConfig.jwkSet` has no default). This is a correctness fix beyond the stated scope of the PR and worth calling out in the description.

luigidemasi (approved):
LGTM

components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
oscerd: Updated — the comment is now provider-agnostic: `// Resolve the signing key from the configured JWK set and verify the token signature`. Thanks for the catch. — Claude Code on behalf of Andrea Cosentino

components/camel-oauth/src/main/java/org/apache/camel/oauth/UserProfile.java
oscerd: Thanks! Right — `OAuthConfig.jwkSet` has no default so `getJWKSet()` can return null; the guard keeps verification fail-closed instead of NPE-ing. — Claude Code on behalf of Andrea Cosentino

PR #24041, opened 2026-06-16
chore: downgrade protobuf-version to 4.34.2

jamesnetherton:
Relates to:

* https://github.com/apache/camel-quarkus/issues/8681
* The Salesforce project in camel-spring-boot-examples fails with [errors](https://issues.apache.org/jira/browse/CAMEL-23758) similar to those described in the above issue

Protobuf gencode version must be <= the runtime version used by consumers. Spring Boot uses 4.34.2 and Quarkus uses 4.35.0, so aligning to 4.34.2 ensures generated stubs are compatible with both frameworks without errors.

Excluded com.google.protobuf from Dependabot to manage this manually and avoid inadvertent bumps that break framework compatibility. Perhaps protobuf is something we bump when we do Spring Boot dependency alignment.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,6 +34,8 @@ updates:
       - dependency-name: "org.apache.logging.log4j:*"
       # This is a virtual plugin which doesn't really exist
       - dependency-name: "org.eclipse.m2e:lifecycle-mapping"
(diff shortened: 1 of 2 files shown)


apupier:
relates to https://issues.apache.org/jira/browse/CAMEL-23758

davsclaus (approved):

PR #24046, opened 2026-06-16
chore: camel-spring-ws - exclude transitive opensaml dependencies

davsclaus:
## Summary

- Exclude `org.opensaml:*` transitive dependencies from `spring-ws-security` in camel-spring-ws
- The OpenSAML artifacts are hosted on `build.shibboleth.net`, a third-party Maven repository that has proven unreliable, causing intermittent CI and build failures (see #24042)
- Camel's Spring-WS component does not use OpenSAML — the dependency is pulled in transitively through `wss4j → opensaml-saml-impl`

## Test plan

- [x] Verified no camel-spring-ws source code imports or references `org.opensaml`
- [ ] CI passes with the exclusion in place

### DIFF
--- a/components/camel-spring-parent/camel-spring-ws/pom.xml
+++ b/components/camel-spring-parent/camel-spring-ws/pom.xml
@@ -116,6 +116,10 @@
                     <groupId>xalan</groupId>
                     <artifactId>xalan</artifactId>
                 </exclusion>
+                <exclusion>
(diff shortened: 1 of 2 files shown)


oscerd (approved):

apupier (commented):
Should we not remove the repository declaration from pom.xml and the .rrf configuration?

davsclaus:
opensaml still used in camel-cxf

PR #24050, opened 2026-06-16
CAMEL-23772: dataformat component honors global camel.dataformat.* configuration

oscerd:
## Summary

`dataformat:` component endpoints (e.g. `to("dataformat:beanio:unmarshal")`) ignored the global `camel.dataformat.<name>.*` configuration from `application.properties`, while the equivalent fluent DSL (`unmarshal().beanio(...)`) already honored it. The two paths behaved inconsistently, and the component path could throw at runtime (e.g. `IllegalArgumentException: Stream name not configured` for BeanIO).

Reported while analysing [camel-quarkus#7651](https://github.com/apache/camel-quarkus/issues/7651); the root cause is in Camel core.

## Root cause

`DataFormatComponent.createEndpoint()` calls `createDataFormat(name)`, which returns a fresh, **unconfigured** instance, and then binds only the endpoint URI parameters. It never copies the options of the auto-configured "template" data format (the one set up from `camel.dataformat.<name>.*`, resolvable via `resolveDataFormat(name)` and listed in `getDataFormatNames()`).

`DataFormatReifier.configureDataFormat()` already does this for the `marshal()`/`unmarshal()` DSL path (CAMEL-22352), but the component uses a different code path that was never updated.

## Fix

When the component creates a brand-new data format instance, it now harvests the auto-configured template's option values (via the same `ExtendedPropertyConfigurerGetter` mechanism the reifier uses) and binds them onto the new instance **before** binding the endpoint URI parameters — so URI parameters keep precedence over the global configuration. The shared template instance itself is never mutated.

The change is intentionally contained to `camel-dataformat`. `AbstractCamelContext.createDataFormat()` is deliberately left unchanged, since it must keep returning a clean instance for the reifier and other callers.

## Tests

Added `DataFormatComponentConfigTest` in `camel-csv` (mirroring the CAMEL-22352 `MainCsvTest`): a `Main`-based route using `to("dataformat:csv:unmarshal")` with `camel.dataformat.csv.ignore-surrounding-spaces=true` in `application.properties`. It asserts the global option is applied (surrounding spaces stripped) — this fails without the fix and passes with it. The existing `MainCsvTest` (reifier path) still passes.

## Docs

Added a `camel-dataformat` entry to the 4.21 upgrade guide describing the behavior alignment.

## Build

Full-reactor `mvn clean install -Dquickly` passes. No generated/catalog/DSL files changed (there is no `@UriParam` surface change).

---

### DIFF
--- a/components/camel-csv/pom.xml
+++ b/components/camel-csv/pom.xml
@@ -53,6 +53,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.camel</groupId>
+            <artifactId>camel-dataformat</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
             <artifactId>camel-spring-xml</artifactId>
             <scope>test</scope>
         </dependency>
--- a/components/camel-csv/src/test/java/org/apache/camel/dataformat/csv/component/DataFormatComponentConfigTest.java
+++ b/components/camel-csv/src/test/java/org/apache/camel/dataformat/csv/component/DataFormatComponentConfigTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


davsclaus (approved):
Nice work on this fix — the inconsistency between the `dataformat:` component path and the `marshal()`/`unmarshal()` DSL was a real gap, and this closes it cleanly.

**What looks good:**

- The approach correctly mirrors the established pattern in `DataFormatReifier.configureDataFormat()` — using `ExtendedPropertyConfigurerGetter` to read options from the auto-configured template.
- The CAMEL-16032 invariant (fresh instances, no shared-instance mutation) is preserved.
- URI parameter precedence is maintained: global config first, then endpoint params override.
- The `created` flag correctly gates the copy logic.
- The test mirrors the existing `MainCsvTest` pattern and uses `try-finally` for cleanup.
- Upgrade guide entry is accurate and well-placed.

**One suggestion (non-blocking):**

The test validates that global config *is applied*, but doesn't test the *precedence* case where a URI parameter overrides a global property. A second test case like `to("dataformat:csv:unmarshal?ignoreSurroundingSpaces=false")` (overriding the global `true`) would confirm endpoint params retain precedence.

PR #24062, opened 2026-06-16
CAMEL-23704: Use isolated exchange copy per tool in langchain4j-tools

k-krawczyk:
### CAMEL-23704

`LangChain4jToolsProducer.invokeTools()` reused a single `Exchange` across all tool invocations within a request. As a result, one tool's message body, accumulated argument headers, and exceptions leaked into subsequent tool invocations.

#### Changes
- Snapshot the original incoming exchange once (`toolsChat`) and reuse it as a clean **baseline** across all LLM iterations.
- Invoke each tool on its **own independent copy** of that baseline (`ExchangeHelper.createCopy`), mirroring the multicast/splitter isolation pattern.
- Copy each tool's outcome back onto the parent exchange with `ExchangeHelper.copyResults`, so the producer output still reflects the (last) tool's result body and declared argument headers — no accumulation across tools ("last tool wins").
- Exceptions are kept on the tool's own exchange and propagated to the parent without affecting sibling tools.

#### Behavior change
A tool route no longer receives the previous tool's output as its body, nor the previous tool's argument headers. Documented in the 4.21 upgrade guide.

#### Note on the test
`LangChain4jToolMultipleCallsTest` previously asserted that the second tool's body contained the first tool's output — i.e. it codified the exact leak this ticket reports (the loop and the test both landed in `050af78ca`). It has been updated to assert isolation: each tool sees its own argument headers, does **not** inherit the previous tool's `name` header, and receives the original incoming body.

#### Relation to CAMEL-21937
This change was rebased on top of CAMEL-21937 (*Reset route stop flag between tool invocations*). The per-tool exchange isolation introduced here subsumes the **per-iteration** `ROUTE_STOP` reset from that fix — since every tool now runs on its own copy of the baseline exchange, a `stop()` in one tool route can no longer reach sibling tools, so that per-iteration reset was removed as dead code. The **post-loop** `exchange.setRouteStop(false)` is **kept**, because `ExchangeHelper.copyResults` propagates the last tool's `routeStop` onto the parent exchange and it must not leak into the calling route. `LangChain4jToolStopEipTest` continues to pass.

#### Testing
`mvn test -pl components/camel-ai/camel-langchain4j-tools` — all green (39 tests, including `LangChain4jToolStopEipTest`).

### DIFF
--- a/components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsProducer.java
+++ b/components/camel-ai/camel-langchain4j-tools/src/main/java/org/apache/camel/component/langchain4j/tools/LangChain4jToolsProducer.java
@@ -46,6 +46,7 @@ import org.apache.camel.TypeConverter;
 import org.apache.camel.component.langchain4j.tools.spec.CamelToolExecutorCache;
 import org.apache.camel.component.langchain4j.tools.spec.CamelToolSpecification;
 import org.apache.camel.support.DefaultProducer;
+import org.apache.camel.support.ExchangeHelper;
 import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -119,6 +120,8 @@ public class LangChain4jToolsProducer extends DefaultProducer {
             return null;
         }
 
+        final Exchange baseline = ExchangeHelper.createCopy(exchange, true);
+
         // First talk to the model to get the tools to be called
         int i = 0;
         do {
@@ -129,7 +132,7 @@ public class LangChain4jToolsProducer extends DefaultProducer {
             }
 
             // Only invoke the tools ... the response will be computed on the next loop
-            invokeTools(chatMessages, exchange, response, toolPair);
+            invokeTools(chatMessages, exchange, response, toolPair, baseline);
             LOG.debug("Finished iteration {}", i);
             i++;
         } while (true);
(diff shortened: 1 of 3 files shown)


k-krawczyk:
cc @Croway @davsclaus for review — you're the most active committers on `camel-langchain4j-tools` (and @davsclaus is the reporter). Would appreciate your feedback on the per-tool exchange isolation approach.

k-krawczyk:
Rebased onto current `main` to resolve conflicts. Notably integrated with CAMEL-21937: per-tool isolation makes the per-iteration `ROUTE_STOP` reset redundant (removed), while the post-loop reset is retained because `copyResults` propagates `routeStop` to the parent. All module tests pass, including `LangChain4jToolStopEipTest`.

davsclaus (approved):
Clean bug fix using the established multicast/splitter isolation pattern (`ExchangeHelper.createCopy` + `copyResults`). The baseline capture, per-tool isolation, result propagation, and post-loop `routeStop` cleanup are all correct. The test update properly asserts isolation instead of codifying the leak, and the upgrade guide entry is well-written.

This review covers project rules, conventions, and code correctness visible from the diff. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

oscerd (approved):
Independent review (I hadn't looked at this one yet) — this is correct and well-scoped. Replacing the shared-exchange mutation with a per-tool isolated `exchange.copy()` from an immutable baseline is the right pattern, and the regression test genuinely proves isolation (the second tool no longer inherits the first tool's body/headers, and the "last tool wins" semantics on the parent are documented). CI is green, the upgrade-guide entry on `main` is clear, and `ExchangeHelper.copyResults` correctly propagates state including the routeStop flag — so the post-loop reset is still needed and is correctly retained.

One tiny nit: the PR body cites commit `050af78ca` as where the original loop/test landed, but that hash is an unrelated TUI commit — worth correcting or dropping.

LGTM 👍

---

PR #24065, opened 2026-06-17
CAMEL-23775: Add classpath-discovered YAML DSL deserializer resolvers

luigidemasi:
## Summary
Adds automatic discovery for YAML DSL `YamlDeserializerResolver` implementations so optional modules can contribute custom YAML route step deserializers when they are present on the classpath.

This change also defines and tests deterministic resolver precedence across built-in, registry-provided, and classpath-discovered resolvers.

  ## Changes

  - Discover resolver providers from `META-INF/services/org/apache/camel/YamlDeserializerResolver`.
  - Include Camel application/context classloaders when looking up resolver service resources.
  - Preserve built-in YAML DSL resolver behavior unless an external resolver explicitly uses higher precedence.
  - Apply deterministic ordering for registry and classpath resolvers, including class-name and registry-name tie-breaks.
  - Keep resolver loading fail-fast with clearer diagnostics for broken provider entries.
  - Clear cached constructors when resolver membership changes.
  - Document the extension point, ordering rules, schema limitations, and trusted-input expectations.
  - Add tests for custom route steps, nested `steps`, route execution, registry resolvers, Kamelets, parser usage, resolver collisions, ordering, inline comments, and broken providers.
  - Also updates stale YAML normalization command references from `camel yaml normalize` to `camel validate normalize` in the compact-notation warning/docs/
  generated metadata.

  ## Tests

  - `mvn -pl dsl/camel-yaml-dsl/camel-yaml-dsl-common,dsl/camel-yaml-dsl/camel-yaml-dsl -DskipITs -Dtest=YamlDeserializerResolverDiscoveryTest test`
  - `mvn -pl dsl/camel-yaml-dsl/camel-yaml-dsl-common,dsl/camel-yaml-dsl/camel-yaml-dsl,dsl/camel-yaml-dsl/camel-yaml-dsl-validator -DskipITs test`
  - `mvn -pl dsl/camel-yaml-dsl/camel-yaml-dsl-validator-maven-plugin -DskipTests install`
  - `git diff --check`

_🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Luigi De Masi._


oscerd (changes requested):
Thanks for this — the classpath-discovered resolver SPI is a clean, well-tested addition, and I appreciate the explicit security note in `yaml-dsl.adoc` clarifying that resolver providers and the YAML itself are trusted route-author/deployment inputs (consistent with Camel's threat model).

A couple of things before this can go in:

**Blocking**
- CI is red on `build (25)` at the *"Fail if there are uncommitted changes"* gate. It looks like the regenerated user-manual mirror of the validator-plugin doc (`docs/user-manual/modules/ROOT/pages/camel-yaml-dsl-validator-maven-plugin.adoc`) wasn't committed — the module source copy under `dsl/camel-yaml-dsl/.../src/main/docs/` was edited but not its generated mirror. Re-running the validator-plugin doc generation and committing the result should clear it.

**Minor**
- The PR also carries an unrelated `camel yaml normalize` → `camel validate normalize` rename across ~6 files. It looks like a legitimate catch-up to the CLI rebrand, but it's outside CAMEL-23775's scope and isn't called out in the description — worth a one-line note for reviewers, or splitting out.
- `newYamlDeserializationContext()` now calls `ctx.start()` eagerly so registry resolvers load up front. That seems *more* correct than before, but it is a behavioral change in when registry resolvers are discovered — please confirm it's intended (the new tests do pass).

Test coverage and docs (upgrade guide on `4_21`, the `yaml-dsl.adoc` "Extending YAML route steps" section) are solid. Nice work.

---

dsl/camel-yaml-dsl/camel-yaml-dsl-common/src/main/java/org/apache/camel/dsl/yaml/common/YamlDeserializationContext.java
```diff
@@ -120,14 +141,204 @@ protected Optional<ConstructNode> findConstructorFor(Node node) {
...
+
+    private void loadResolvers() {
         ObjectHelper.notNull(camelContext, "camel context");
 
-        this.resolvers.addAll(getCamelContext().getRegistry().findByType(YamlDeserializerResolver.class));
+        if (resolversLoaded) {
+            return;
+        }
+
+        addResolverEntries(loadResolversFromClasspath());
```
jamesnetherton: Could we perhaps have some way for runtimes to suppress loading resolvers from the classpath?

On Camel Quarkus it's not really needed because we can discover and configure all `YamlDeserializerResolver` impls at build time.

luigidemasi: @jamesnetherton I agree that Camel Quarkus should not need this runtime classpath discovery path if it can discover and configure the resolvers at build time.

  **I think the critical bit is that this should not be a user-facing “suppress classpath loading” switch.**
  A cleaner fix would be to introduce a Camel-owned discovery/provider SPI, for example `YamlDeserializerResolverProvider`, registered as a Camel context plugin.

  The YAML deserialization context would then:

  1. Ask the provider for automatically discovered resolvers.
  2. Load explicit registry resolvers separately, as it does today.

  The default Camel provider would preserve the current behavior by reading:

  ```text
  META-INF/services/org/apache/camel/YamlDeserializerResolver
  ```

  and instantiating those classes through the Camel injector.

  Camel Quarkus could register its own provider at build time, returning the resolvers discovered during augmentation, without doing runtime ClassLoader.getResources(...) or service-resource scanning. 

  This should avoid breaking existing users because:

  - YamlDeserializerResolver itself remains unchanged.
  - Existing service descriptor based discovery continues to work in plain Camel.
  - Explicit registry resolvers remain supported.
  - Quarkus/native gets an authoritative build-time integration point instead of a runtime suppression flag.

  So I propose to adjust the PR in that direction unless you see a better Camel Quarkus integration point.

_AI-generated draft by Claude on behalf of Luigi De Masi._

luigidemasi:
@oscerd:
>
> **Blocking**
> 
> * CI is red on `build (25)` at the _"Fail if there are uncommitted changes"_ gate. It looks like the regenerated user-manual mirror of the validator-plugin doc (`docs/user-manual/modules/ROOT/pages/camel-yaml-dsl-validator-maven-plugin.adoc`) wasn't committed — the module source copy under `dsl/camel-yaml-dsl/.../src/main/docs/` was edited but not its generated mirror. Re-running the validator-plugin doc generation and committing the result should clear it.

It should be fixed now ✅

> **Minor**
> 
> * The PR also carries an unrelated `camel yaml normalize` → `camel validate normalize` rename across ~6 files. It looks like a legitimate catch-up to the CLI rebrand, but it's outside [CAMEL-23775](https://issues.apache.org/jira/browse/CAMEL-23775)'s scope and isn't called out in the description — worth a one-line note for reviewers, or splitting out.

the rename is tiny, legitimate, and partly generated/doc catch-up. Splitting it would add process overhead for a low-risk wording correction, I'll update the pr description for the reviewers.


> * `newYamlDeserializationContext()` now calls `ctx.start()` eagerly so registry resolvers load up front. That seems _more_ correct than before, but it is a behavioral change in when registry resolvers are discovered — please confirm it's intended (the new tests do pass).

Yes, this is intentional.
Previously `YamlDeserializationContext.start()` was the place where registry-provided `YamlDeserializerResolver` instances were loaded, but the route loader did not explicitly start the context after creating it. With the new classpath discovery path, resolver discovery has to happen before parsing starts, otherwise a module-provided YAML step may be encountered before its resolver has been loaded.

dsl/camel-yaml-dsl/camel-yaml-dsl-common/src/main/java/org/apache/camel/dsl/yaml/common/YamlDeserializationContext.java
jamesnetherton: > Introduce a Camel-owned discovery/provider SPI, for example YamlDeserializerResolverProvider, registered as a Camel context plugin

That sounds ok. Lets follow that direction.

I can always follow with future PRs if additional tweaks are needed for CQ.

oscerd (approved):

PR #24072, opened 2026-06-17
feat(dsl): camel jbang export jib

squakez:
Added a Jib profile to enable the publishing of the container via Jib profile

NOTE: this is separate from the existing jib flags which may be used in kubernetes plugin (via jkube and others).

Closes CAMEL-23780

# Description

# Target

# Tracking


davsclaus (commented):
Can we use the same jib replacholder as in the k8s 

```
              <version>[=JibMavenPluginVersion]</version>
```

squakez:
> JibMavenPluginVersion

I don't think so. That variable is enabled when the jkube plugin is enabled. The goal of this PR is separate from the Jkube one as it wants to include the plugin regardless of the presence of other tools. The idea is that the user can launch a `mvn jib:build` OOTB and have the same experience with all the runtimes we support as explained also in the readme.md files of each export.

davsclaus:
Yeah but we should not hardcode the version, so it does not go stale

squakez:
> Yeah but we should not hardcode the version, so it does not go stale

Ah, okey. Sure, I'll see how to make it a variable then. Thanks!

davsclaus (approved):

oscerd (commented):
Reviewed this independently as part of a batch pass. Nicely scoped, low-risk feature — thanks for the work here. 👍

**What looks good**
- Purely additive at the template level: the `jib` profile is `activeByDefault=false`, so existing export output is unchanged.
- Tests added (`ExportTest.shouldContainJibProfile` + the updated `TemplateHelperTest` cases), README templates updated, `camel-jbang-configuration.adoc` and the catalog metadata regenerated, and `Closes CAMEL-23780` referenced.
- `jib-maven-plugin` is injected only into the *generated* end-user project POMs (a Maven plugin), so it doesn't enter camel-jbang's own dependency tree; `dependency-review` passed and `3.5.1` is the current release.
- The one Java behavior change (moving the `JibMavenPluginVersion` model entry out of the kubernetes-conditional path into the always-run POM creation) is the right fix so the new profile can resolve the placeholder, and it reuses the existing `[=JibMavenPluginVersion]` convention as requested earlier in the thread.

**Minor, non-blocking nits (take or leave)**
- Every generated `jib` profile hardcodes `<allowInsecureRegistries>true</allowInsecureRegistries>`. Handy for local/dev registries, but it ships into every exported project. A one-line note in the README that it can be removed for production pushes would help users who don't expect a permissive default.
- `ExportTest.shouldContainJibProfile` asserts the rendered POM contains `<id>jib</id>`; it might be worth also asserting the plugin coordinate / resolved version rendered (`jib-maven-plugin`, `3.5.1`) so a future regression in the model wiring fails this test directly rather than only at FreeMarker render time. (`TemplateHelperTest` does cover the version, so this is belt-and-suspenders.)

**Status:** CI is fully green on the current HEAD, and the PR already carries an approval from @davsclaus. The branch was rebased onto newer `main` after that approval — I confirmed the jib feature code itself is unchanged by the rebase, so the existing approval still stands. No action needed from my side; the notes above are optional.

---

squakez:
> <allowInsecureRegistries>true</allowInsecureRegistries>

The tool is mainly used for prototyping, so no need to provide any further documentation at this time about it.

PR #24079, opened 2026-06-17
Clean jbang-camel-it/target/data folder

apupier:
On Jenkins CI, this folder has "root" as the owner of this file. It is most likely causing Jenkins workspace cleanup plugin failing to clean the workspace, thus leading to no more disk space.

Note that I'm unable to reproduce locally. The folder is created with my account (but i'm also an admin, maybe the reason?). That said i'm unable to play all the tests locally due to various errors (most of them seems to search for a /home/jbang/.m2 which obviously doesn't exists)

# Description

# Target

# Tracking

### DIFF
--- a/dsl/camel-jbang/camel-jbang-it/pom.xml
+++ b/dsl/camel-jbang/camel-jbang-it/pom.xml
@@ -238,6 +238,18 @@
                                     <goal>run</goal>
                                 </goals>
                             </execution>
+                            <execution>
+                                <id>post-integration-test</id>
+                                <phase>post-integration-test</phase>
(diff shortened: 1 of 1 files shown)


apupier:
@mcarlett as you created the initial Camel JBang IT tests in case, does it ring a bell for this data folder for different potential rights for it?
Are you able to launch the tests locally?

davsclaus (approved):

mcarlett:
thanks @apupier we adapted the CI since it depends on the os user, anyway I think it saves space at least, thanks!

PR #24094, opened 2026-06-17
Configure Mockito to avoid self attachment warning

apupier:
it is in link with https://openjdk.org/jeps/451

the warnings were:
```
Mockito is currently self-attaching to enable the inline-mock-maker.
This will no longer work in future releases of the JDK. Please add
Mockito as an agent to your build as described in Mockito's
documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot
loader classes because bootstrap classpath has been appended
WARNING: A Java agent has been loaded dynamically
(/home/apupier/.m2/repository/net/bytebuddy/byte-buddy-agent/1.18.10/byte-buddy-agent-1.18.10.jar)
WARNING: If a serviceability tool is in use, please run with
-XX:+EnableDynamicAgentLoading to hide this warning
WARNING: If a serviceability tool is not in use, please run with
-Djdk.instrument.traceUsage for more information
WARNING: Dynamic loading of agents will be disallowed by default in a
future release
```

it remains just `OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended` which has a different root cause

note that I have not found a JEP for planned removal of the Dynamic Loading of Agents. Anyway, it will allow to be ready and it is mentioned as a bit safer.

# Description

# Target

# Tracking

### DIFF
--- a/catalog/camel-catalog/pom.xml
+++ b/catalog/camel-catalog/pom.xml
@@ -35,6 +35,7 @@
         <!-- use by camel-catalog -->
         <firstVersion>2.15.0</firstVersion>
         <label>tooling</label>
+        <argLine></argLine>
     </properties>
 
     <dependencies>
@@ -202,6 +203,18 @@
                         <delimiter>@</delimiter>
                     </delimiters>
                 </configuration>
+            </plugin>
+            <!-- Workaround: redeclaring without argLine to avoid error with Mockito specified as an agent -->
+             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <phase>none
+                        </phase><!-- unbound here to speedup -Dquickly build; bound to a phase in the full profile -->
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
--- a/pom.xml
+++ b/pom.xml
@@ -217,6 +217,17 @@
                     </execution>
                 </executions>
             </plugin>
+             <plugin>
+                 <groupId>org.apache.maven.plugins</groupId>
+                 <artifactId>maven-dependency-plugin</artifactId>
+                 <executions>
+                     <execution>
+                         <goals>
+                            <goal>properties</goal>
+                         </goals>
+                     </execution>
+                 </executions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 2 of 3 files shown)


davsclaus (approved):

davsclaus:
Nice if we can remove this warning

apupier:
hum not the kind of errors that i would have expected with this kind of change...:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-remote-resources-plugin:3.3.0:process (default) on project bundles-pom: Error loading supplemental data models: Could not find resource 'target/legal/legal.xml'. -> [Help 1]
```

EDIT: which of course i do not reproduce locally when building directly the maven module in error
EDIT2: reproducing when building from top-level

davsclaus:
⏺ The build fails on bundles-pom with:

  Error loading supplemental data models: Could not find resource 'target/legal/legal.xml'

  This is a build infrastructure issue in the bundles-pom module — the maven-remote-resources-plugin can't find target/legal/legal.xml. This is unrelated to the PR's Mockito changes. It failed
  on Java 25 only (the Java 17 job also failed, likely the same cause).

  This is a known pattern in the Camel build — the legal.xml supplemental resource is generated by a prior build step, and if that step is skipped or the build order is off (especially with mvnd
  parallelism), bundles-pom can't find it. The PR itself (configuring Mockito self-attachment) doesn't touch anything near bundles-pom or the legal/resource plugin setup, so this looks like a
  flaky CI issue rather than a problem introduced by the PR.

apupier:
I doubt that it is flaky as I reproduced every time locally when built from top-level.
Bob suggested very different solutions (mainly to use ids for Maven execution goals or link it to a specific phase) but none worked so far.

apupier:
going further but new error:

```
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd '/home/runner/work/camel/camel/catalog/camel-catalog' && '/usr/lib/jvm/temurin-25-jdk-amd64/bin/java' '-javaagent:${org.mockito:mockito-core:jar}' '-jar' '/home/runner/work/camel/camel/catalog/camel-catalog/target/surefire/surefirebooter-20260618071746273_3.jar' '/home/runner/work/camel/camel/catalog/camel-catalog/target/surefire' '2026-06-18T07-17-45_562-jvmRun1' 'surefire-20260618071746273_1tmp' 'surefire_0-20260618071746273_2tmp'
```

(more expected as they mention it in the documentation

apupier:
hum, the workaround for maven-surefire-plugin is not enough. In fact, it seems that we should declare the new parameter only if mockito is part of the dependencies, otherwise all others are failing with similar error

apupier:
closing for now to avoid polluting the opened PR list

I won't havetime to di more on this part very soon

PR #24118, opened 2026-06-18
CAMEL-23796: Fix wrong token in logical operator right-hand-side error message

ammachado:
# Description

When the simple expression parser encounters an invalid right-hand side in a logical operator expression (`&&` / `||`), it was reporting the wrong token in the error message: it used `left.getToken()` instead of `right.getToken()`.

This PR fixes the one-line bug and makes the error branch reachable by teaching `LogicalExpression.acceptLeftNode` and `acceptRightNode` to enforce that only `BinaryExpression` or `LogicalExpression` nodes are valid on either side of a logical operator. Previously both methods unconditionally returned `true`, so the error path was dead code and the wrong-token bug was unobservable.

A new test (`SimplePredicateParserLogicalTest`) covers:
- Happy-path `&&` and `||` combinations (7 tests)
- A negative case that verifies the error message names the right-hand offending token, not the left-hand one (1 test)

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
@@ -552,7 +552,7 @@ public class SimplePredicateParser extends BaseSimpleParser {
                 }
                 if (!logical.acceptRightNode(right)) {
                     throw new SimpleParserException(
-                            "Logical operator " + operator + " does not support right hand side token " + left.getToken(),
+                            "Logical operator " + operator + " does not support right hand side token " + right.getToken(),
                             token.getToken().getIndex());
                 }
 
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Correct bug fix — `left.getToken()` → `right.getToken()` on line 555 is a clear copy-paste error from `4ea9e6c35737` (2020). The `instanceof` restriction in `acceptLeftNode`/`acceptRightNode` is necessary to make the error path reachable and is semantically correct: only `BinaryExpression` (predicates like `${body} == 'hello'`) and `LogicalExpression` (chained `&&`/`||`) are valid operands of logical operators. The only unary operators (`++`/`--`) are value modifiers, not predicates — rejecting them is the right behavior.

Good test coverage: 7 happy-path tests plus a negative test that verifies the error message contains the right-hand token (`42`) and not the left-hand one (`==`).

PR #24120, opened 2026-06-18
CAMEL-23798: Replace AtomicBoolean/AtomicInteger with boolean[]/int[] lambda capture holders

ammachado:
# Description

`SimpleExpressionParser` and `SimplePredicateParser` used `AtomicBoolean` and `AtomicInteger` solely as mutable holders to capture state inside lambda expressions. Since both parsers are single-threaded, using `Atomic*` types is semantically misleading (it implies concurrent access) and carries unnecessary overhead.

This commit replaces them with single-element `boolean[]` and `int[]` arrays, which are the standard Java idiom for capturing mutable state in a lambda without implying thread safety.

A new `SimpleParserMutableStateTest` covers the state-tracking paths in both parsers: balanced/unbalanced single quotes, balanced/unbalanced function blocks, and compound predicate expressions.

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
@@ -19,7 +19,6 @@ package org.apache.camel.language.simple;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.Expression;
@@ -242,8 +241,8 @@ public class SimpleExpressionParser extends BaseSimpleParser {
     protected void parseAndCreateAstModel() {
         // we loop the tokens and create a sequence of ast nodes
 
-        // counter to keep track of number of functions in the tokens
-        AtomicInteger functions = new AtomicInteger();
+        // single-element array used as a mutable counter (no concurrency; AtomicInteger is not needed here)
+        int[] functions = { 0 };
 
         LiteralNode imageToken = null;
         for (SimpleToken token : tokens) {
@@ -288,15 +287,15 @@ public class SimpleExpressionParser extends BaseSimpleParser {
         }
     }
 
-    private SimpleNode createNode(SimpleToken token, AtomicInteger functions) {
+    private SimpleNode createNode(SimpleToken token, int[] functions) {
         // expression only support functions, unary operators, operators, and other operators
         if (token.getType().isFunctionStart()) {
             // starting a new function
-            functions.incrementAndGet();
+            functions[0]++;
             return new SimpleFunctionStart(token, cacheExpression, skipFileFunctions);
-        } else if (functions.get() > 0 && token.getType().isFunctionEnd()) {
+        } else if (functions[0] > 0 && token.getType().isFunctionEnd()) {
             // there must be a start function already, to let this be an end function
-            functions.decrementAndGet();
+            functions[0]--;
             return new SimpleFunctionEnd(token);
         } else if (token.getType().isUnary()) {
             // there must be an end function as previous, to let this be a unary function
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
@@ -23,7 +23,6 @@ import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.Expression;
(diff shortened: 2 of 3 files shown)


core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
```diff
@@ -242,8 +241,8 @@ protected void removeIgnorableWhiteSpaceTokens() {
     protected void parseAndCreateAstModel() {
         // we loop the tokens and create a sequence of ast nodes
 
-        // counter to keep track of number of functions in the tokens
-        AtomicInteger functions = new AtomicInteger();
+        // single-element array used as a mutable counter (no concurrency; AtomicInteger is not needed here)
```
davsclaus, line 244: This comment explains the implementation mechanism (`single-element array`, `no concurrency`, `AtomicInteger is not needed`) rather than the domain intent. The original comment `// counter to keep track of number of functions in the tokens` was more useful — it tells the reader *what* is being tracked, not *how* the holder works.

If keeping the `int[]` change, consider restoring the original comment.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
```diff
@@ -334,25 +333,25 @@ private void addImageToken(LiteralNode imageToken) {
      * Creates a node from the given token
      *
      * @param  token         the token
-     * @param  startSingle   state of single quoted blocks
-     * @param  startDouble   state of double quoted blocks
-     * @param  startFunction state of function blocks
+     * @param  startSingle   state of single quoted blocks (single-element boolean array used as a mutable holder)
+     * @param  startDouble   state of double quoted blocks (single-element boolean array used as a mutable holder)
+     * @param  startFunction state of function blocks (single-element boolean array used as a mutable holder)
      * @return               the created node, or <tt>null</tt> to let a default node be created instead.
```
davsclaus, line 339: The added `(single-element boolean array used as a mutable holder)` in each `@param` line duplicates what the type signature `boolean[]` already conveys. The original Javadoc (`state of single quoted blocks`) documented the *semantic meaning* — what the parameter represents — which is more valuable than describing its Java type.

davsclaus (commented):
Thanks for the contribution, Adriano.

The refactoring is functionally correct — behavior is preserved. A few observations:

**PR description accuracy:** The title and description state these `Atomic*` types were used "solely as mutable holders to capture state inside lambda expressions." However, there are no lambdas capturing these variables in either file. The `AtomicInteger` and `AtomicBoolean` instances are passed as regular method parameters to private helper methods (`createNode()`). The only lambda in these files is `tokens.removeIf(t -> ...)` which does not reference any of the replaced variables. The actual pattern is pass-by-reference via method parameters — a different (and legitimate) use of mutable wrappers.

**Comments:** The replacement comment `// single-element array used as a mutable counter (no concurrency; AtomicInteger is not needed here)` describes the implementation mechanism rather than the domain intent. The original `// counter to keep track of number of functions in the tokens` was more informative. Similarly the Javadoc additions like `(single-element boolean array used as a mutable holder)` duplicate what the type signature already says.

**Test overlap:** Several of the new tests duplicate existing coverage — for example `testPredicateUnclosedSingleQuoteThrows` covers the same scenario as `SimpleParserPredicateInvalidTest.testSimpleUnevenSingleQuote`, and the happy-path tests are covered by `SimpleTest` and `SimplePredicateParserNodesTest`. The tests verify parser behavior (which would pass identically with `AtomicInteger` or `int[]`), not the mutable state holder mechanism specifically.

**Net value:** Both `AtomicInteger` and `int[]` serve the same purpose as mutable holders in this single-threaded context. `AtomicInteger` has been the established idiom in this code since 2012 and is self-documenting. The change trades one pattern for another without a clear readability or performance win.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
ammachado: Correct on both counts: this counter is passed by reference into createNode(), not captured by a lambda, and the comment described the mechanism rather than what is tracked. Withdrawing the PR; see closing comment.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimplePredicateParser.java
ammachado: Agreed: the added text restates the boolean[] type and the original @param documented the semantic meaning. Withdrawing the PR; see closing comment.

ammachado:
Withdrawing this PR.

The review is correct on the substance. Two factual issues confirmed against the code:

1. The change was mislabeled as "lambda capture holders." The `int[]`/`boolean[]` values are passed by reference as parameters into `createNode()`; the only lambda (`tokens.removeIf`) does not reference them.
2. The new inline comment and `@param` additions described the Java mechanism rather than the domain intent, and the added tests largely duplicate existing coverage (e.g. `SimpleParserPredicateInvalidTest.testSimpleUnevenSingleQuote`, which additionally asserts the error index).

On net value: `AtomicBoolean`/`AtomicInteger` as single-threaded mutable holders has been the established idiom in these parsers since 2012 and is self-documenting. Swapping it for single-element arrays is lateral churn in long-stable code with no clear readability or performance win, so this is not worth pursuing. Closing and resolving CAMEL-23798 as Won't Do.

PR #24121, opened 2026-06-18
CAMEL-23799: Sort init-block substitution keys by length to prevent prefix collision

ammachado:
# Description

When `SimpleExpressionParser` and `SimplePredicateParser` substitute init-block variable names into the expression string, a shorter key that is a prefix of a longer one (e.g. `$n` and `$name`) can match first and corrupt the substitution: `$name` becomes `${variable.n}ame`.

The fix sorts the keys by descending length before iterating, so longer (more specific) keys are always replaced first.

A new `SimpleInitBlockOverlappingKeysTest` demonstrates the bug: both fixtures declare the short key first (so insertion order is shortest-first, the corrupting order), and the tests fail without the sort and pass with it.

# Target

# Tracking

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/SimpleExpressionParser.java
@@ -17,6 +17,7 @@
 package org.apache.camel.language.simple;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Clean bug fix — the prefix collision in init-block variable substitution is a real issue, and the descending-length sort is the right approach.

The regression tests are well-designed: they declare the shorter key first to force the problematic `LinkedHashSet` insertion order, so they fail without the fix and pass with it.

One minor note: the function key sort in `SimpleExpressionParser` is defensive but not strictly necessary — the function substitution patterns include delimiters (`$key()` and `${key(`) that prevent shorter keys from matching inside longer ones. Harmless either way.

PR #24122, opened 2026-06-18
CAMEL-23800: Prefix function cache keys to enforce keyspace separation from expression cache

ammachado:
# Description

`SimpleFunctionExpression` shares the same `Map<String, Expression>` cache instance with the
expression-level cache (`SimpleExpressionParser`, `SimplePredicateParser`). Without key namespacing,
a function name like `body` could collide with any other entry stored under the same string,
causing the wrong `Expression` to be returned silently.

This commit introduces a `@FUNC@` prefix constant in `SimpleFunctionExpression` and applies it
on every cache read and write. The prefix is intentional: `@` is not a valid character in a
Simple function name, so `@FUNC@body` can never collide with any expression-level key.

This issue was discovered during the work on CAMEL-22894.

# Target

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-23800) filed for the change (usually before you start working on it).

### DIFF
--- a/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
+++ b/core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
@@ -56,15 +56,16 @@ public class SimpleFunctionExpression extends LiteralExpression {
     @Override
     public Expression createExpression(CamelContext camelContext, String expression) {
         String function = text.toString();
+        String cacheKey = FUNCTION_CACHE_KEY_PREFIX + function;
 
-        Expression answer = cacheExpression != null ? cacheExpression.get(function) : null;
+        Expression answer = cacheExpression != null ? cacheExpression.get(cacheKey) : null;
         if (answer == null) {
             answer = createSimpleExpression(camelContext, function, true);
             if (answer != null) {
                 answer.init(camelContext);
             }
             if (cacheExpression != null && answer != null) {
-                cacheExpression.put(function, answer);
+                cacheExpression.put(cacheKey, answer);
             }
         }
         return answer;
@@ -96,6 +97,11 @@ public class SimpleFunctionExpression extends LiteralExpression {
         }
         return exp;
     }
+
+    // Prefix that separates function-level cache entries from the top-level expression cache entries
+    // written by SimpleLanguage (which uses "@SIMPLE@" + fullExpression). The two keyspaces share the
+    // same LRUCache instance, so distinct prefixes are needed to prevent accidental collisions.
+    private static final String FUNCTION_CACHE_KEY_PREFIX = "@FUNC@";
 
     private static final DirectFunctionFactory DIRECT_FACTORY = new DirectFunctionFactory();
 
--- a/core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleFunctionCacheKeyTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleFunctionCacheKeyTest.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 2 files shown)


core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
```diff
@@ -97,6 +98,11 @@ private Expression createSimpleExpression(CamelContext camelContext, String func
         return exp;
     }
 
+    // Prefix that separates function-level cache entries from the top-level expression cache entries
+    // written by SimpleLanguage (which uses "@SIMPLE@" + fullExpression). The two keyspaces share the
+    // same LRUCache instance, so distinct prefixes are needed to prevent accidental collisions.
+    private static final String FUNCTION_CACHE_KEY_PREFIX = "@FUNC@";
```
davsclaus, line 104: The `@SIMPLE@` prefix on expression-level keys (added in CAMEL-17073) already prevents collision with raw function names stored here. The `@FUNC@` prefix adds a second layer of separation that is redundant with the existing design.

Also, this constant is declared after the methods — Java convention typically places `private static final` constants near the top of the class, next to the `cacheExpression` field at line 42.

core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleFunctionCacheKeyTest.java
```diff
@@ -0,0 +1,96 @@
...
+     * function would find the poison entry and return "WRONG" instead of the actual body value.
+     */
+    @Test
+    public void testFunctionCacheKeyIsPrefixed() {
+        Map<String, Expression> sharedCache = new HashMap<>();
+        sharedCache.put("body", ExpressionBuilder.constantExpression("WRONG")); // raw key: collision target without the fix
+
+        exchange.getIn().setBody("correct");
+
+        SimpleExpressionParser parser = new SimpleExpressionParser(context, "${body}", true, sharedCache);
```
davsclaus, line 50: This test seeds a raw `"body"` entry into the cache manually, but no code path in the framework writes raw unprefixed keys to the shared cache except `SimpleFunctionExpression` itself — which would write the *correct* body expression, not a poison value. The scenario being tested cannot occur through normal framework execution.

davsclaus (changes requested):
Thanks for the effort on this, Adriano.

However, the collision described in the PR does not exist in practice. This was already fixed in CAMEL-17073 (commit `72ebeb85bdf`, Oct 2021), which added the `@SIMPLE@` prefix to expression-level cache keys in `SimpleLanguage`:

```java
private static final String CACHE_KEY_PREFIX = "@SIMPLE@";
```

After that fix, the two keyspaces in the shared cache are:
- **Expression-level entries:** `@SIMPLE@${body}`, `@SIMPLE@header.foo`, etc.
- **Function-level entries:** `body`, `header.foo`, etc.

These cannot collide — a raw function name like `"body"` can never equal `"@SIMPLE@${body}"`. The only code path that writes raw (unprefixed) keys to the cache is `SimpleFunctionExpression` itself, which always writes the *correct* expression for that function name.

The test (`testFunctionCacheKeyIsPrefixed`) demonstrates the collision by manually seeding a raw `"body"` entry into the cache, but no code path in the framework does this — the scenario is artificial.

If the intent is defensive hardening (protecting against hypothetical future writers to the cache), that could be reasonable — but the PR description and JIRA should reflect that framing rather than describing it as fixing an active bug.

core/camel-core-languages/src/main/java/org/apache/camel/language/simple/ast/SimpleFunctionExpression.java
ammachado: Confirmed. Expression-level keys are always written with the `@SIMPLE@` prefix (`SimpleLanguage` lines 117/179/203, since CAMEL-17073) and function-level keys are raw, so the two keyspaces in the shared cache are already disjoint. A raw function name can never equal `@SIMPLE@${...}`, so the `@FUNC@` prefix is redundant. Closing the PR.

core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleFunctionCacheKeyTest.java
ammachado: Confirmed. `SimpleFunctionExpression` is the only writer of raw (unprefixed) keys to the shared cache, and it always writes the correct expression for that function name. No framework path seeds a poison `"body"` entry, so the scenario this test exercises cannot occur in practice. Removing it along with the rest of the change.

ammachado:
Closing this PR. @davsclaus is correct: the collision this change targets does not exist.

CAMEL-17073 (commit `72ebeb85bdf`) already prefixes every expression-level cache key with `@SIMPLE@`, while `SimpleFunctionExpression` writes raw function names. Those are the only two writers to the shared `cacheExpression` map, so a raw function name can never equal `@SIMPLE@${...}` and the keyspaces are already disjoint. `SimpleFunctionExpression` also always writes the correct expression for a given function name, so function-to-function reuse is correct as well.

The `@FUNC@` prefix is therefore redundant, and the regression test reproduced an artificial scenario (a manually seeded poison entry) that no framework path produces. CAMEL-23800 will be resolved as "Not a Problem".

PR #24123, opened 2026-06-18
CAMEL-21438: Fix flaky file-based tests (Thread.sleep, missing Awaitility guards, discarded matchesWaitTime return)

ammachado:
# Description

Fix a set of flaky tests in the file component area. Each fix addresses a specific non-determinism pattern:

**1. `PollEnrichFileCustomAggregationStrategyTest`**
Replace `Thread.sleep(500)` with `getMockEndpoint("mock:start").await(10, TimeUnit.SECONDS)`. The sleep was guessing that the route had consumed the first file and reached the `pollEnrich` step; the `await` waits for the observable proof of that (the `mock:start` latch reaching zero) before writing the enrichdata file.

**2. `FileConsumeHiddenFilesTest`**
Add explicit `atMost(10, TimeUnit.SECONDS)` to a bare `Awaitility.await()` call. Without it the test relies on Awaitility's opaque library default, violating the project guideline that every `await()` must have an explicit timeout.

**3. `FileProducerFileExistFailTest`**
Add `mock.await(10, TimeUnit.SECONDS)` before `assertMockEndpointsSatisfied()`. The `doAssertIsSatisfied` only calls `waitForCompleteLatch()` if `expectedCount != getReceivedCounter()` at the moment it is entered; on a fast machine the count is already met and the latch wait is skipped entirely, leaving `runTests()` (which evaluates `expectedFileExists`) with no stable-state guarantee.

**4. `GenericFileConverterTest`**
Same root cause as #3. The test uses `isUseRouteBuilder()=false` so each of the 7 test methods manually starts the context; `assertMockEndpointsSatisfied()` is the only timing gate but has no explicit latch wait before `runTests()`. Add `mock.await(10, TimeUnit.SECONDS)` before `assertMockEndpointsSatisfied()` in every method.

**5. `FileProducerCharsetUTFtoUTFTest`, `FileProducerCharsetUTFOptimizedTest`, `FileProducerCharsetUTFtoISOConfiguredTest`**
After `oneExchangeDone.matchesWaitTime()` confirms the exchange completed, `Files.readAllBytes` on the output file can still return 0 bytes because Camel's file producer writes via a temp file + atomic rename, and the OS may not have flushed the page to the reader yet. `FileProducerCharsetUTFtoISOTest` already carries an explicit comment about this and uses an Awaitility guard; apply the same pattern to the three missing tests.
Additionally, `FileProducerCharsetUTFOptimizedTest` was silently discarding the boolean return value of `matchesWaitTime()`, masking any timeout before the file assertions ran.

**6. Review feedback (oscerd, davsclaus)**
- All `MockEndpoint.await(10, TimeUnit.SECONDS)` calls in `FileProducerFileExistFailTest`, `GenericFileConverterTest`, and `PollEnrichFileCustomAggregationStrategyTest` now assert the returned boolean with `assertTrue(...)` so a timeout causes an explicit, descriptive failure instead of silently falling through to `assertMockEndpointsSatisfied()`.
- `ConcurrentRequestsThrottlerTest.java` was reverted to match `main` exactly (commit `0c03b30d0f3` had already fixed it). Prior commits in this branch introduced an out-of-scope `onDone` refactor and accidentally dropped `.delay(100)` from `direct:expressionConstant`; both are now restored.

# Target

# Tracking

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
@@ -194,12 +194,7 @@ public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
                             assertTrue(s.tryAcquire(), "'direct:a' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                                public void onComplete(Exchange ex) {
-                                    s.release();
-                                }
-
-                                @Override
-                                public void onFailure(Exchange ex) {
+                                public void onDone(Exchange ex) {
                                     s.release();
                                 }
                             });
@@ -213,12 +208,11 @@ public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
                             assertTrue(s.tryAcquire(), "'direct:expressionConstant' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                            exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
-                                @Override
                                 public void onDone(Exchange ex) {
                                     s.release();
                                 }
                             });
+                        })
                         .to("log:result", "mock:result");
 
                 from("direct:expressionHeader").throttle(header("throttleValue")).concurrentRequestsMode()
@@ -227,12 +221,7 @@ public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
                             assertTrue(s.tryAcquire(), "'direct:expressionHeader' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                                public void onComplete(Exchange ex) {
-                                    s.release();
-                                }
-
-                                @Override
-                                public void onFailure(Exchange ex) {
+                                public void onDone(Exchange ex) {
                                     s.release();
                                 }
                             });
--- a/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
@@ -194,7 +194,12 @@ public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
                             assertTrue(s.tryAcquire(), "'direct:a' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                                public void onDone(Exchange ex) {
+                                public void onComplete(Exchange ex) {
+                                    s.release();
+                                }
+
+                                @Override
+                                public void onFailure(Exchange ex) {
                                     s.release();
                                 }
                             });
@@ -208,11 +213,17 @@ public class ConcurrentRequestsThrottlerTest extends ContextTestSupport {
                             assertTrue(s.tryAcquire(), "'direct:expressionConstant' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                                public void onDone(Exchange ex) {
+                                public void onComplete(Exchange ex) {
+                                    s.release();
+                                }
+
+                                @Override
+                                public void onFailure(Exchange ex) {
                                     s.release();
                                 }
                             });
                         })
+                        .delay(100)
                         .to("log:result", "mock:result");
 
                 from("direct:expressionHeader").throttle(header("throttleValue")).concurrentRequestsMode()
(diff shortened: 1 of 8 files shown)


davsclaus (approved):
Clean PR — all five fixes follow established codebase patterns and align with the project's testing guidelines (no `Thread.sleep`, explicit Awaitility timeouts, checked return values).

**Highlights:**
- `PollEnrichFileCustomAggregationStrategyTest`: correct synchronization point (`mock:start.await`) replaces `Thread.sleep(500)`
- `FileConsumeHiddenFilesTest`: bare `Awaitility.await()` now has explicit `atMost(10, TimeUnit.SECONDS)`
- `FileProducerFileExistFailTest` + `GenericFileConverterTest`: `mock.await()` before `assertMockEndpointsSatisfied()` ensures latch-based stability for `expectedFileExists` assertions
- Charset tests (`UTFtoUTF`, `UTFOptimized`, `UTFtoISOConfigured`): Awaitility guard for file flush, matching the existing pattern in `FileProducerCharsetUTFtoISOTest`
- `FileProducerCharsetUTFOptimizedTest`: `assertTrue(oneExchangeDone.matchesWaitTime())` now captures the previously discarded return value

oscerd (changes requested):
Thanks for tackling the file-test flakiness — the seven intended fixes here are solid:

- `Thread.sleep(500)` is removed from `PollEnrichFileCustomAggregationStrategyTest` and replaced with `getMockEndpoint("mock:start").await(10, SECONDS)`, which waits on the real proof point (`mock:start` receiving the message, immediately before `pollEnrich`) instead of guessing — a strictly better synchronization point.
- The Awaitility conversions in `FileConsumeHiddenFilesTest` (`atMost(10s)`) and the three `FileProducerCharset*` tests (`atMost(5s)`) all carry explicit timeouts.
- The discarded `matchesWaitTime()` return in `FileProducerFileExistFailTest` is now actually asserted.

No `Thread.sleep` is added anywhere in the intended scope. 👍

**Blocking issue — the branch does not compile at the current head.**

The head commit `092bb65fa` ("Update …ConcurrentRequestsThrottlerTest.java") leaves `core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java` in a malformed state. In the `from("direct:expressionConstant")` route (around lines 214-222 at head):

- an `@Override` annotation is immediately followed by a statement (`exchange.getExchangeExtension().addOnCompletion(...)`) rather than a method;
- the first `new SynchronizationAdapter() { ... }` is opened but its body contains another `addOnCompletion(...)` call where the method declarations should be;
- the enclosing `.process(exchange -> { ... })` lambda is never closed, leaving `.to("log:result", "mock:result");` dangling.

This will fail `camel-core` test compilation. The file is also outside this PR's stated scope (the description covers the file-based tests + `matchesWaitTime`), and the change looks like an incomplete edit rather than an intended part of CAMEL-21438.

A couple of related observations:

- No CI has run on this branch yet (`no checks reported`), so the breakage wasn't caught automatically. It would be worth letting the build go green before merge — especially for a flaky-test fix.
- The existing approval predates this head commit.

Suggested resolution: drop/restore `ConcurrentRequestsThrottlerTest.java` (or complete the intended edit) so the branch compiles, then let CI run. The seven file-test fixes themselves look ready.

Optional nit (non-blocking): a few of the new `MockEndpoint.await(...)` calls discard the returned `boolean` (e.g. `FileProducerFileExistFailTest.java:55`, the `GenericFileConverterTest` methods, `PollEnrichFileCustomAggregationStrategyTest.java:44`). It's harmless since each is followed by `assertMockEndpointsSatisfied()`, but the explicit `assertTrue(...)` pattern you used for `matchesWaitTime` would be more consistent.

---

core/camel-core/src/test/java/org/apache/camel/processor/throttle/concurrent/ConcurrentRequestsThrottlerTest.java
```diff
@@ -213,17 +208,11 @@ public void onFailure(Exchange ex) {
                             assertTrue(s.tryAcquire(), "'direct:expressionConstant' too many requests");
                             exchange.getExchangeExtension().addOnCompletion(new SynchronizationAdapter() {
                                 @Override
-                                public void onComplete(Exchange ex) {
-                                    s.release();
-                                }
-
-                                @Override
-                                public void onFailure(Exchange ex) {
+                                public void onDone(Exchange ex) {
```
davsclaus: `.delay(100)` is removed here from `direct:expressionConstant` but kept on the other two routes (`direct:a` and `direct:expressionHeader`). All three had the delay before this PR — the inconsistency looks unintentional.

Also, this file already compiles correctly on `main` (fixed by commit `0c03b30d0f3`). The `onDone` refactoring and delay removal effectively revert that fix. Consider dropping this file from the PR.

davsclaus (commented):
The 7 file-test fixes are solid — they follow project conventions (no `Thread.sleep`, explicit Awaitility timeouts, `assertTrue` on return values), and oscerd's feedback has been addressed.

One concern with the `ConcurrentRequestsThrottlerTest` changes:

**This file already compiles correctly on `main`** — commit `0c03b30d0f3` (June 13) restored it after the prior commit `5ce7e9e6a39` left it malformed. The changes in this PR go beyond restoration: they refactor `onComplete`/`onFailure` to `onDone` and remove `.delay(100)` from `direct:expressionConstant`. This effectively reverts parts of `0c03b30d0f3`.

The `.delay(100)` removal is also inconsistent — it's removed from `direct:expressionConstant` but kept on `direct:a` and `direct:expressionHeader`. All three routes had the delay before; removing it from only one seems unintentional.

Suggestion: drop the `ConcurrentRequestsThrottlerTest` changes from this PR since the file is already correct on `main`. The 7 file-test fixes are ready as-is.

davsclaus (approved):

PR #24137, opened 2026-06-19
CAMEL-23767: camel-platform-http-main - warn when authentication is enabled but no mechanism is configured

oscerd:
Fixes [CAMEL-23767](https://issues.apache.org/jira/browse/CAMEL-23767).

## Problem

`DefaultMainHttpServerFactory.configureAuthentication` wires basic or JWT authentication only when a basic-auth properties file (`basicPropertiesFile`) or a JWT keystore (`jwtKeystoreType`) is configured:

```java
if (configuration.getBasicPropertiesFile() != null) {        // basic
} else if (configuration.getJwtKeystoreType() != null) {     // JWT
}
// else: nothing
```

When `authenticationEnabled=true` but neither is configured, the embedded HTTP server (and the management server) started **without any authentication and without any warning** — silently exposing an unprotected server while the operator believes it is protected.

## Change

Both `configureAuthentication` overloads now log a clear **warning** at startup when authentication is enabled but no mechanism is configured:

> Authentication is enabled (authenticationEnabled=true) but no authentication mechanism is configured: neither a basic-auth properties file (basicPropertiesFile) nor a JWT keystore (jwtKeystoreType) is set. The HTTP server will start WITHOUT authentication. Configure an authentication mechanism, or set authenticationEnabled=false to disable authentication.

The server still starts (behaviour otherwise unchanged), so the change is **non-breaking**. Warning rather than failing fast keeps it safe for the patch-release backports (per the Jira `fixVersions`); note the JWT branch already fails fast for a *partial* keystore configuration, so warning for a *missing* mechanism is consistent.

## Tests

- `AuthenticationConfigurationMainHttpServerTest#testAuthenticationEnabledWithoutMechanism` — captures the log and asserts the server starts **and** the warning is emitted (new `auth-no-mechanism.properties`).
- All existing module tests pass (17).
- Full reactor build (`mvn clean install -DskipTests`, 1854 modules) green, no drift. No generated-file impact (internal factory, no `@UriParam`).

## Documentation

- `camel-4x-upgrade-guide-4_21.adoc` — note appended to the existing `camel-platform-http-main` section.

## Compatibility / backport

Non-breaking; suitable for backport to `camel-4.18.x` and `camel-4.14.x`. The matching 4_18/4_14 upgrade-guide entries will be added on `main` with the backports.

---

### DIFF
--- a/components/camel-platform-http-main/src/main/java/org/apache/camel/component/platform/http/main/DefaultMainHttpServerFactory.java
+++ b/components/camel-platform-http-main/src/main/java/org/apache/camel/component/platform/http/main/DefaultMainHttpServerFactory.java
@@ -28,9 +28,13 @@ import org.apache.camel.main.MainHttpServerFactory;
 import org.apache.camel.spi.annotations.JdkService;
 import org.apache.camel.support.TempDirHelper;
 import org.apache.camel.util.ObjectHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @JdkService(MainConstants.PLATFORM_HTTP_SERVER)
 public class DefaultMainHttpServerFactory implements CamelContextAware, MainHttpServerFactory {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DefaultMainHttpServerFactory.class);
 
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

luigidemasi (approved):
LGTM!

PR #24139, opened 2026-06-19
CAMEL-23804: Fix SSE streaming on camel-platform-http-vertx

jamesnetherton:
AsyncInputStream greedily filled its 16KB buffer before emitting data, which blocked SSE events behind the next queue poll. Add an eagerFlush mode that emits each chunk immediately, and activate it automatically for text/event-stream responses. Also slice the output buffer to the actual bytes written to prevent trailing zeroes on partial reads.

Update camel-a2a docs to remove the Vert.x SSE streaming limitation now that platform-http delivers events eagerly.


davsclaus (approved):
Nice fix — clean, focused, and well-tested.

The `eagerFlush` mode correctly addresses the greedy-fill issue for SSE streams, and the `writeBuff.slice(0, written)` fix is a welcome bonus that prevents trailing zero bytes on partial reads for all `InputStream` responses.

The test design (PipedOutputStream + CountDownLatch proving eager delivery) is solid.

oscerd (approved):

PR #24154, opened 2026-06-20
Fix Jenkins matrix filters

fallintoplace:
## Summary

- Require the platform filter and JDK filter to both match before a matrix cell runs.
- Apply the same filter logic to the JDK 26 Jenkinsfile.

## Root cause

The matrix `when` block used one broad `anyOf`, so selecting a platform could also run every JDK on that platform, and selecting a JDK could also run that JDK across every platform. The filters are independent dimensions, so each dimension needs its own wildcard-or-match check.

## Validation

- `git diff --check -- Jenkinsfile Jenkinsfile.jdk26`
- Ran a small truth-table sanity check: selecting one platform and one JDK now allows exactly one cell instead of five.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,11 +63,15 @@ pipeline {
                 options {
                     throttle(['camel'])
                 }
-                when { anyOf {
-                    expression { params.PLATFORM_FILTER == 'all' }
-                    expression { params.PLATFORM_FILTER == env.PLATFORM }
-                    expression { params.JDK_FILTER == 'all' }
-                    expression { params.JDK_FILTER == env.JDK_NAME }
+                when { allOf {
+                    anyOf {
+                        expression { params.PLATFORM_FILTER == 'all' }
+                        expression { params.PLATFORM_FILTER == env.PLATFORM }
+                    }
+                    anyOf {
+                        expression { params.JDK_FILTER == 'all' }
+                        expression { params.JDK_FILTER == env.JDK_NAME }
+                    }
                 } }
                 axes {
                     axis {
--- a/Jenkinsfile.jdk26
+++ b/Jenkinsfile.jdk26
@@ -68,11 +68,15 @@ pipeline {
                 options {
                     throttle(['camel'])
                 }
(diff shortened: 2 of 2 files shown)


davsclaus (approved):
Correct fix — the `anyOf` flattening meant selecting one dimension (platform or JDK) would match all values in the other dimension. The `allOf { anyOf { ... } anyOf { ... } }` nesting properly makes the two filters independent. Both Jenkinsfiles updated consistently. LGTM.

apupier (approved):

oscerd (commented):
Independently verified — this is a correct, minimal fix. 👍

The original single flat `anyOf` OR-combined all four expressions across both filter dimensions, so a wildcard (or match) in *either* dimension was enough to run a matrix cell. Concretely, with `PLATFORM_FILTER=ppc64le` and `JDK_FILTER=all`, the `params.JDK_FILTER == 'all'` term is true for every cell, so all platforms ran regardless of the platform selection (and symmetrically for the JDK dimension). Wrapping each dimension in its own `anyOf` (wildcard-or-match) and combining them with `allOf` makes the two filters independent `AND`-ed predicates, which is the intended behavior. The "5 cells instead of 1" example in the description checks out (3 platform-matches ∪ 3 JDK-matches − 1 overlap = 5).

A few confirmations against the current `main`:
- The filter choices (`Jenkinsfile:53-54`) line up exactly with the matrix axis values for `PLATFORM`/`JDK_NAME`, so `params.PLATFORM_FILTER == env.PLATFORM` and `params.JDK_FILTER == env.JDK_NAME` are sound comparisons.
- The nested `allOf { anyOf{...} anyOf{...} }` is valid declarative-pipeline `when` syntax.
- `Jenkinsfile` and `Jenkinsfile.jdk26` receive the identical fix — nicely consistent.

Nit (non-blocking): the commit could carry a `CAMEL-XXXXX` reference per the commit convention, though that's commonly waived for CI-only changes.

No concerns from me. Thanks for the clear root-cause writeup, @fallintoplace.

---

PR #24155, opened 2026-06-20
Fix metadata updater JSON handling

fallintoplace:
## Summary

- Use Python JSON serialization when formatting metadata keys, string values, and array items.
- Exit with an error when a requested metadata file has no matching target to update.

## Root cause

The metadata updater manually wrapped strings in quotes while preserving the generated Jackson-like layout. That could emit invalid JSON when existing metadata values or new service versions contained characters that need escaping. The script also ignored update results, so a missing match could leave generated metadata stale while the workflow continued.

## Validation

- `python3 -m py_compile .github/actions/check-container-upgrade/update-metadata-version.py`
- `git diff --check`
- Ran temp-file checks covering quoted/backslash/newline values, successful updates across both metadata files, and the no-match failure path.

### DIFF
--- a/.github/actions/check-container-upgrade/update-metadata-version.py
+++ b/.github/actions/check-container-upgrade/update-metadata-version.py
@@ -42,16 +42,12 @@ def jackson_format(data):
                 if not value:
                     formatted = "[ ]"
                 else:
-                    items = ", ".join(f'"{v}"' for v in value)
+                    items = ", ".join(json.dumps(v, ensure_ascii=False) for v in value)
                     formatted = f"[ {items} ]"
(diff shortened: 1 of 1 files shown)


oscerd (approved):

oscerd (approved):

davsclaus (approved):
Clean fix — replacing manual string formatting with `json.dumps()` properly handles escaping for special characters, and the no-match error exit is the right behavior for a CI script. LGTM.

Minor note: commit title doesn't follow the `CAMEL-XXXX:` convention, but this is a CI infrastructure fix so it's not a blocker.

PR #24156, opened 2026-06-20
CAMEL-23807: render the route-diagram dev console with the HTML web component

rkdfx:
# Description

The route-diagram developer console (used by `camel run --console`) now renders its HTML output with the interactive <camel-route-diagram> web component instead of a static base64 PNG, fed by the sibling route-structure dev console. Request ?format=png for the legacy inline image; ascii/unicode themes are unchanged.

The web component now requests application/json explicitly, since the dev console content-negotiates on Accept and otherwise returns its plain-text rendering.

# Target

- [x] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

# Apache Camel coding standards and style

- [x] I checked that each commit in the pull request has a meaningful subject line and body.

- [x] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

PR #24160, opened 2026-06-21
CAMEL-23807: make the route-diagram web component render in the dev console (follow-up)

rkdfx:
# Description

Follow-up after the merge (https://github.com/apache/camel/pull/24156): the diagram wasn't actually rendering when running with `--console`.

The page pulled the web component from `/camel/diagram/camel-route-diagram.js`, but that path isn't served when only the developer console is up, so it 404'd and nothing showed. It's now embedded straight into the console's HTML response, so there's no separate file to fetch.

# Target

# Tracking

### DIFF
--- a/components/camel-diagram/src/main/java/org/apache/camel/diagram/DiagramDevConsole.java
+++ b/components/camel-diagram/src/main/java/org/apache/camel/diagram/DiagramDevConsole.java
@@ -17,6 +17,9 @@
 package org.apache.camel.diagram;
 
 import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.StringJoiner;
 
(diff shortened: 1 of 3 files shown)


rkdfx:
Tested locally in browser:

Screenshot 2026-06-21 at 8 09 36 PM

rkdfx:
Separate from this PR, but I hit a dev-console bug while testing and couldn't find a ticket for it, so flagging it here in case you want to file one.

If you open the console index with a trailing slash (`http://localhost:8080/q/dev/`) and click any console, you land on `/q/dev/dev/<id>` (eg `/q/dev/dev/route-diagram`) and get "Developer console not found: dev". Without the trailing slash (`/q/dev`) it works. It's every console link, not just the diagram.

The index renders the links as relative `href="dev/<id>"` (ManagementHttpServer, around line 1060), so from `/q/dev` they resolve to `/q/dev/<id>`, but from `/q/dev/` they double up to `/q/dev/dev/<id>`.

@davsclaus

davsclaus (approved):

davsclaus (approved):
Clean, well-scoped follow-up fix. The approach of inlining the JS via `Class.getResourceAsStream()` at class-load time is correct for the `--console` scenario where static resource serving isn't available. The defensive unwrapping logic in the JS handles both direct and wrapped dev console responses without breaking either case. Tests updated to verify the new behavior.

One minor observation (non-blocking): the `loadWebComponentJs()` catch branch silently returns `""` on `IOException` — a `LOG.warn()` there could help operators debug deployment issues where the JAR is repackaged or shaded incorrectly.

davsclaus:
yeah @rkdfx we have seen the / problem before - you are welcome to look into camel-platform-http-main if you want to hunt it down and improve it

PR #24180, opened 2026-06-22
CAMEL-23765: remote-file consumers - contain localWorkDirectory downloads within the work directory

oscerd:
Fixes [CAMEL-23765](https://issues.apache.org/jira/browse/CAMEL-23765).

## Problem

When `localWorkDirectory` is enabled, the remote-file consumers build the local work file path from the remote file name (`target.getRelativeFilePath()`) **without ensuring the result stays within the configured work directory**:

```java
File local = new File(normalizePath(localWorkDirectory));
String relativeName = target.getRelativeFilePath();   // untrusted remote name
local = new File(local, relativeName);                // "../" escapes the work dir
local.mkdirs(); // ... writes outside localWorkDirectory
```

A remote file name containing `../` sequences could therefore resolve to a path **outside** the work directory (arbitrary local file write) — unlike the file producer, which already jails writes via `FileUtil.compactPath` + `startsWith` when `jailStartingDirectory` is enabled. Per the security model, *path traversal in file/FTP consumers* is in scope.

## Change

- New shared **`GenericFileHelper.jailToLocalWorkDirectory(target, workDir)`** (camel-file) — throws `GenericFileOperationFailedException` if `compactPath(target)` does not `startsWith` `compactPath(workDir)`, mirroring the producer's `jailedCheck`.
- Applied (to both the in-progress temp file and the final file), **before `mkdirs`**, in the `localWorkDirectory` download path of:
  - `FtpOperations` + `SftpOperations` (camel-ftp)
  - `MinaSftpOperations` (camel-mina-sftp)
  - `FilesOperations` (camel-azure-files)
  - `SmbOperations` (camel-smb)
- **Reuses the existing `jailStartingDirectory` option** (default `true`, inherited by all file endpoints) — secure by default, consistent with the producer, opt-out via `jailStartingDirectory=false`. No new config surface.
- The `jailStartingDirectory` option is now `label = "common"` (was `"producer"`) with an expanded description, since it now governs consumer-side `localWorkDirectory` downloads in addition to producer writes. Catalog/component metadata and endpoint-dsl factories are regenerated accordingly for all file-based components (file, ftp, ftps, sftp, smb, azure-files, mina-sftp, scp).

## Tests

- New `GenericFileHelperTest` — verifies files within the work directory (incl. `../` that still resolves inside) are allowed, and escaping paths are rejected.
- Existing localWorkDirectory ITs (camel-smb `SmbConsumerLocalWorkDirectoryIT`, camel-mina-sftp feature IT) cover legitimate-download regression.
- Full reactor build (`mvn clean install -DskipTests`, 1874 modules) green. The only generated-file changes are the `jailStartingDirectory` `producer → common` relabel (catalog + component JSON + endpoint-dsl factories for the file components); no other drift.

## Documentation

- `camel-4x-upgrade-guide-4_21.adoc` — note added for the remote-file consumers.

## Compatibility / backport

Default-secure with an opt-out (`jailStartingDirectory=false`), so suitable for backport to `camel-4.18.x` and `camel-4.14.x` (per the Jira `fixVersions`: 4.14.8, 4.18.3, 4.21.0). Matching 4_18/4_14 guide entries will be added on `main` with the backports.

---


components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
```diff
@@ -16,16 +16,39 @@
...
 
     private GenericFileHelper() {
     }
 
+    /**
+     * Ensures the resolved local work file stays within the configured local work directory. The remote file name used
+     * to build the local work file path may contain {@code ../} sequences that would otherwise resolve to a path
+     * outside the work directory.
+     *
+     * @param  target                              the resolved local work file (or its in-progress temp file)
```
davsclaus, line 36: Minor (non-blocking): The `jailStartingDirectory` option on `GenericFileEndpoint` is declared with `label = "producer"`, so it only appears in the **Producer Options** table in generated component docs. This PR extends its meaning to also govern consumer-side `localWorkDirectory` downloads. Since the default is `true`, users only need to find the option if they want to *disable* it (and the upgrade guide entry names it), so practical impact is low. A follow-up to change the label to `"common"` and expand the description would improve discoverability.

components/camel-smb/src/main/java/org/apache/camel/component/smb/SmbOperations.java
```diff
@@ -337,12 +338,19 @@ private boolean retrieveFileToFileInLocalWorkDirectory(String name, Exchange exc
...
 
+            java.io.File localWorkDir = local;
             temp = new java.io.File(local, relativeName + ".inprogress");
 
             // create directory to local work file
             local.mkdirs();
             local = new java.io.File(local, relativeName);
 
+            // ensure the local work file stays within the local work directory (CAMEL-23765)
+            if (endpoint.isJailStartingDirectory()) {
```
davsclaus, line 346: Style (non-blocking): In the other four implementations (`FtpOperations`, `SftpOperations`, `MinaSftpOperations`, `FilesOperations`), the jail check is placed **before** `mkdirs()`. Here it comes **after** `local.mkdirs()` (line 343). This is safe because `mkdirs()` at that point runs on the base work directory (before `relativeName` is appended), so no escaped path is created. But reordering to match the other implementations would improve consistency for future reviewers.

davsclaus (approved):
Solid, well-scoped security fix. The `jailToLocalWorkDirectory` check mirrors the existing `GenericFileProducer.jailedCheck` pattern, covers all five `retrieveFileToFileInLocalWorkDirectory` implementations, and is secure by default via the existing `jailStartingDirectory` option. No blocking issues found — two minor observations below.

components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
oscerd: Good call — applied here rather than as a follow-up (commit 9709484). `jailStartingDirectory` is now `label = "common"` (was `"producer"`) with an expanded description noting that, for consumers using a `localWorkDirectory`, it also restricts downloads to within that directory. It now appears in the common options for all file-based components, and the catalog/component metadata + endpoint-dsl factories are regenerated in this PR.

Since this PR introduces the consumer-side semantic, documenting it here also keeps the 4.18.x / 4.14.x backports consistent (the option governs consumer downloads on those branches too).

components/camel-smb/src/main/java/org/apache/camel/component/smb/SmbOperations.java
oscerd: Done (commit 9709484) — reordered so the jail check runs before directory creation, matching `FtpOperations` / `SftpOperations` / `MinaSftpOperations` / `FilesOperations`. I kept SMB's existing `mkdirs()` on the base work directory (now `localWorkDir.mkdirs()`, equivalent to the previous `local.mkdirs()` call) rather than switching to `local.mkdirs()` on the full path, to avoid changing pre-existing behavior unrelated to this fix.

davsclaus (approved):

PR #24185, opened 2026-06-22
CAMEL-23812: camel-milo better username & password handling

robMate:
# Description

The Camel Milo client component previously embedded authentication credentials directly in the endpoint URI using the format user:password@host. 
To prevent Milo from receiving credentials (which it doesn't handle), the component used string manipulation with `discoveryUri.replaceFirst(user + "@", "")` to strip credentials from the URL before connection. This caused issues because the `replaceFirst()` function uses regex patterns that fail with regex special characters like "$.*", and passwords containing characters like "%" cause issues with `java.net.URLDecoder.decode()` due to percent-encoding in URLs.

This approach has limitations: it fails when credentials contain special characters commonly found in auto-generated passwords, such as:

@ (breaks URI parsing)
?, & (URI query parameter delimiters)
/, # (URI structure delimiters)
$, % (encoding/variable expansion) 

For example, a password like pass@$?&/#% would cause the discovery URI reconstruction to fail or produce incorrect results.

# Target

# Tracking

### DIFF
--- a/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientConfiguration.java
+++ b/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientConfiguration.java
@@ -116,7 +116,7 @@ public class MiloClientConfiguration implements Cloneable {
     @UriParam(label = "security")
     private String username;
 
-    @UriParam(label = "security", secret = true)
+    @UriParam(label = "security", security = "secret")
     private String password;
 
     public MiloClientConfiguration() {
--- a/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientConfiguration.java
+++ b/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientConfiguration.java
@@ -113,6 +113,12 @@ public class MiloClientConfiguration implements Cloneable {
     @UriParam(label = "client", defaultValue = "1_000.0")
     private Double requestedPublishingInterval = DEFAULT_REQUESTED_PUBLISHING_INTERVAL;
 
+    @UriParam(label = "security")
+    private String username;
+
+    @UriParam(label = "security", secret = true)
+    private String password;
+
     public MiloClientConfiguration() {
     }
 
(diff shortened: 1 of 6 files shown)


components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientConfiguration.java
```diff
@@ -113,6 +113,12 @@ public class MiloClientConfiguration implements Cloneable {
     @UriParam(label = "client", defaultValue = "1_000.0")
     private Double requestedPublishingInterval = DEFAULT_REQUESTED_PUBLISHING_INTERVAL;
 
+    @UriParam(label = "security")
+    private String username;
+
+    @UriParam(label = "security", secret = true)
```
Croway: as per https://issues.apache.org/jira/browse/CAMEL-23250 it should be `secret = "secret"`

components/camel-milo/src/main/java/org/apache/camel/component/milo/client/MiloClientEndpoint.java
```diff
@@ -203,4 +203,25 @@ public boolean isOmitNullValues() {
     public void setOmitNullValues(boolean omitNullValues) {
         this.omitNullValues = omitNullValues;
     }
+
+
```
davsclaus, line 207: Nit: double blank line here — should be a single blank line per code style.

davsclaus (changes requested):
Nice work tackling the special-character credential handling — the `replaceFirst(user + "@", "")` regex approach was genuinely broken and the explicit-parameter alternative is a clean solution.

One blocking item and a couple of minor observations:

**Blocking: missing generated files** — the new `@UriParam` fields (`username`, `password`) in `MiloClientConfiguration` should trigger regeneration of the component metadata (`src/generated/`). The current `milo-client.json` doesn't contain these params, and the diff doesn't include any generated-file changes. CI will fail with uncommitted generated changes. Please run `mvn clean install -DskipTests` in the `components/camel-milo` directory and commit the regenerated files.

- The endpoint-level `getUsername/setUsername/getPassword/setPassword` delegation methods are not annotated with `@UriParam`. Since `MiloClientConfiguration` already declares these as `@UriParam` and `configuration` itself is a `@UriParam` on the endpoint, Camel's property binding should resolve them through the nested config object — the endpoint-level methods may be redundant.
- Discovery URI reconstruction now always runs (even without user info), which is correct but subtly changes from the prior conditional-only modification.

davsclaus:
replaced by other PR closing this

PR #24193, opened 2026-06-23
CAMEL-23812: camel-milo better username & password handling

davsclaus:
## Summary

- Adds explicit `username` and `password` options to `MiloClientConfiguration` so credentials no longer need to be embedded in the endpoint URI
- Fixes issues with special characters in passwords (e.g. `@`, `$`, `?`, `%`) that broke URI parsing and `replaceFirst()` regex matching
- Rebuilds discovery URI from parsed URI parts instead of string manipulation
- Includes regenerated metadata files (configurers, URI factories, JSON) that were missing from the original PR

Based on the work from @robMate in #24185, with regenerated metadata and formatting fixes added on top.

## Test plan

- [x] `ExplicitCredentialsTest` added to verify explicit username/password handling
- [ ] Build camel-milo module: `cd components/camel-milo && mvn verify`


robMate:
Cool thank you for the quick resolution of this issue and for fixing up my code 😊👍

PR #24201, opened 2026-06-23
Use a shallow git clone on Jenkins when possible for main build

apupier:
there are currently issues with disk space available on the Jenkins nodes.
Camel is one of the most consuming one. Each of its workspace for Camel core is taking 4.4G, it includes 1.4G for the .git folder. Using a shallow clone should reduce the 1.4G to 63M and gain few minutes to the build too.

Co-authored-by: IBM Bob IDE 1.0.3

# Description

# Target

# Tracking

### DIFF
--- a/Jenkinsfile.jdk26
+++ b/Jenkinsfile.jdk26
@@ -96,7 +96,14 @@ pipeline {
                         steps {
                             cleanWs()
                             sh 'rm -rvf /home/jenkins/.m2/repository/org/apache/camel'
-                            checkout scm
+                            checkout([
+                                $class: 'GitSCM',
+                                branches: scm.branches,
+                                extensions: [
+                                    [$class: 'CloneOption', depth: 1, noTags: true, shallow: true]
+                                ],
+                                userRemoteConfigs: scm.userRemoteConfigs
+                            ])
                         }
                     }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -133,7 +133,24 @@ pipeline {
                         steps {
                             cleanWs()
                             sh 'rm -rvf /home/jenkins/.m2/repository/org/apache/camel'
-                            checkout scm
+                            script {
+                                // Use full clone for JDK 21 on ubuntu-avx (needed for Sonar analysis)
+                                // Use shallow clone for all other combinations
+                                if ("${PLATFORM}" == "ubuntu-avx" && "${JDK_NAME}" == "jdk_21_latest") {
+                                    echo "Using full clone for ${PLATFORM}-${JDK_NAME} (required for code coverage and Sonar)"
+                                    checkout scm
+                                } else {
+                                    echo "Using shallow clone for ${PLATFORM}-${JDK_NAME}"
+                                    checkout([
+                                        $class: 'GitSCM',
+                                        branches: scm.branches,
+                                        extensions: [
+                                            [$class: 'CloneOption', depth: 1, noTags: true, shallow: true]
+                                        ],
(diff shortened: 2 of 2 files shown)


apupier:
before merging I would like to test it on the CI by changing manually. I will likely do it tomorrow

Jenkinsfile.jdk26
```diff
@@ -96,7 +96,14 @@ pipeline {
...
-                            checkout scm
+                            checkout([
+                                        $class: 'GitSCM',
+                                        branches: scm.branches,
+                                        extensions: [
+                                            [$class: 'CloneOption', depth: 1, noTags: true, shallow: true]
+                                        ],
+                                        userRemoteConfigs: scm.userRemoteConfigs
+                                    ])
                         }
```
davsclaus, line 107: Nit: the indentation here is slightly off compared to the surrounding `steps` block and the equivalent change in the main `Jenkinsfile`. The `$class` and its children have extra leading spaces.

davsclaus (commented):
Thanks for working on this, @apupier — reducing the `.git` footprint from 1.4GB to ~63MB per workspace is a meaningful improvement for the Jenkins nodes.

A few observations from reviewing the diff:

**Sonar/coverage case is correctly handled** — the full clone exception for `ubuntu-avx` + `jdk_21_latest` ensures SonarCloud's SCM blame sensor (which uses `git blame`) will continue to work. Good catch.

**No git-history-dependent Maven plugins found** — I checked the root `pom.xml` for plugins like `git-commit-id-plugin` or `buildnumber-maven-plugin` and found none, so `depth: 1` won't break the Maven build itself.

**Minor observation on `Jenkinsfile.jdk26`** — this file unconditionally uses shallow clone for all platforms, which is safe since there's no Sonar/coverage stage. If Sonar analysis is ever added for JDK 26 builds, someone would need to add the full-clone exception there too. A brief comment noting this could help future maintainers.

**Indentation in `Jenkinsfile.jdk26`** — the `checkout([...])` block has slightly inconsistent indentation compared to the surrounding code and the main `Jenkinsfile` change. Very minor/cosmetic.

Overall the approach looks sound. I see you want to validate on Jenkins first before merging — makes sense given this is infrastructure.

apupier:
tested with JDk 26, the exact git commands used on Jenkisn are:
git init
git fetch --no-tags --force --progress --depth=1 -- https://github.com/apache/camel.git +refs/heads/*:refs/remotes/origin/*

which is fecthing all branches, causing the .git to be at 376M, instead of 63M with a pure `git clone --depth=1 https://github.com/apache/camel` which is no downloading branches. To see if we can configure that safely. Can also be improved in another iteration.

remains the conditional to test too on the Camel main repo

Jenkinsfile.jdk26
apupier: updated

apupier:
> remains the conditional to test too on the Camel main repo

tested with https://ci-builds.apache.org/job/Camel/job/Camel%20Core%20(Build%20and%20test)/job/main/1943/

and validating that all .git in saved workspaces get down to 365M (no more 376M as I cleaned some old branches on apache/camel fork) vs 1.4G previously and for the remaining JDK 21/Ubuntu

I propose to merge this version and look to reduce even more to the 63M potential target in another iteration (CI is currently crowded so complicated to test too many things in //) it will surely be by using honorRefSpec parameter.
The current state already allows to save 1.1x (4 (matrix config) + 1 (JDK26)) x 5 (number of history build kept) = 27.5G of disk space

davsclaus (approved):

PR #24203, opened 2026-06-23
CAMEL-19531: Remove unnecessary Thread.sleep calls from leveldb tests

anvithsg2004:
### Motivation

Some leveldb tests contain unnecessary `Thread.sleep()` calls that increase test execution time without contributing to the behavior being verified.

### Modifications

* Removed unnecessary `Thread.sleep()` calls from leveldb tests.
* Kept the existing test assertions and behavior unchanged.

### Result

The tests continue to pass while avoiding unnecessary waiting, resulting in faster and more efficient test execution.


davsclaus (approved):
LGTM — all six `Thread.sleep()` removals are safe.

**`Thread.sleep(3)` / `Thread.sleep(5)` in concurrent tests (4 files):** These trivial delays inside `Callable` tasks provide no meaningful simulation. The tests already use `MockEndpoint.assertIsSatisfied()` with proper timeouts.

**`Thread.sleep(1000)` between stop/start (2 files):** `LevelDBFile.stop()` and `start()` are both fully synchronous — `stop()` calls `IOHelper.close(db)` and nulls the reference, `start()` opens a new connection via `DBFactory.open()`. No async cleanup, so no delay needed.

Awaitility replacement is not necessary here since the sleeps were genuinely superfluous rather than masking race conditions.

PR #24212, opened 2026-06-24
Reflect output change in DependencyListITCase

jvrubel:
The command `camel dependency list --runtime=spring-boot` no longer outputs the artifact version. This change reflects that in the test


Croway:
Do you happen to know why dependency list spring-boot no longer outputs the artifact version? is that something we want, or a regression on main?

davsclaus:
must be the BOM import so all versions are managed there

davsclaus:
Thanks for looking into this. However, we'd like the version to always be included in the output for all runtimes (Spring Boot, Quarkus, etc.), since tooling that consumes `camel dependency list` needs full GAV coordinates (groupId:artifactId:version) to fetch JARs.

The `DependencyList` code already has logic to resolve BOM-managed versions (lines 184-188), so the fix should be to investigate why the version resolution isn't working rather than removing it from the assertion.

See: https://issues.apache.org/jira/browse/CAMEL-23824

jvrubel:
I assumed it was intentional, but after some digging I found that the behaviour of the command has not changed and since the dependencies are BOM-managed, this is probably a bug. I'm now investigating the root cause. I apologize for the rushed "solution". I'll let this serve as a lesson to not take Claude up on every suggestion.

PR #24218, opened 2026-06-24
CAMEL-23819 fix(camel-milo): prevent NullPointerException by filtering null browse results

robMate:
# Description

When browsing the OPC UA node tree of a Siemens S7 PLC, a NullPointerException occurred in MiloBrowseProducer. The S7's OPC UA server returns null for BrowseResult.getReferences() on certain nodes (e.g., leaf nodes or nodes without child references), rather than returning an empty array. The stream pipeline in the browse result processing did not account for this, causing a NPE when Stream::of was called on a null reference array.

# Root Cause
The code at MiloBrowseProducer.java:114 assumed that BrowseResult.getReferences() always returns a non-null array. The Siemens S7 OPC UA server implementation violates this assumption by returning null references for nodes that have no children or for which browsing is not applicable.

# Solution
A .filter(Objects::nonNull) step was added to the stream pipeline between .map(BrowseResult::getReferences) and .flatMap(Stream::of). This filters out any null reference arrays before they are passed to Stream::of, gracefully skipping nodes without references instead of crashing.

# Target

# Tracking

_Created with help from Claude Code on behalf of Robin Mattes_

### DIFF
--- a/components/camel-milo/src/main/java/org/apache/camel/component/milo/browse/MiloBrowseProducer.java
+++ b/components/camel-milo/src/main/java/org/apache/camel/component/milo/browse/MiloBrowseProducer.java
@@ -112,6 +112,7 @@ public class MiloBrowseProducer extends DefaultAsyncProducer {
 
                     final List<String> expandedNodes = browseResults.values().stream()
                             .map(BrowseResult::getReferences)
+                            .filter(Objects::nonNull)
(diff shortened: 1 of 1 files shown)


apupier (approved):

oscerd (approved):

apupier:
Thanks for the contribution!

PR #24224, opened 2026-06-24
CAMEL-23821: fix not finding any files in validation

orpiske:


orpiske:
Likely this needs to be backported to 4.18.x as well

davsclaus (approved):

PR #24230, opened 2026-06-24
[backport camel-4.18.x] CAMEL-23828: Honor explicit --quarkus-version during export

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-23828

When the user specifies --quarkus-version on the CLI, via application.properties, or via system property, the export command should use that exact version instead of resolving from the Quarkus Extension Registry. Previously the registry resolution was unconditional and would overwrite the user's version with the latest patch for that stream.

The fix removes the default value from the --quarkus-version option so the field starts as null. ExportQuarkus now only applies the build-time default and registry resolution when no explicit version was provided.

# Description

# Target

# Tracking


davsclaus:
How does this work on the main branch? We cannot make special code on 4.18.x if its not similar/same as the main branch

JiriOndrusek:
> How does this work on the main branch? We cannot make special code on 4.18.x if its not similar/same as the main branch

In upstream, there was a big change https://issues.apache.org/jira/browse/CAMEL-23560 and part of it is QuarkusPlatformMixin, which provides the same functionality https://github.com/apache/camel/blob/8cb42d2840e65207d2de40d8f5cbd6900e622c4e/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/QuarkusPlatformMixin.java#L70

(I mentioned the upstream fix in the ticket for 4.18.x, but I should mention it in the PR as well)

davsclaus:
Ah okay, then this is fine

davsclaus (approved):

PR #24246, opened 2026-06-25
CAMEL-23817: Embed test plugin in Camel launcher

christophd:
# Description

- Use Citrus JBang commands directly from launcher instead of spawning a new JBang process
- Update to Citrus 5.0.0-M2 for Jackson 3 support and Camel 4.20+

# Target

# Tracking

Fixes CAMEL-23817

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-test/src/main/java/org/apache/camel/dsl/jbang/core/commands/test/TestPlugin.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-test/src/main/java/org/apache/camel/dsl/jbang/core/commands/test/TestPlugin.java
@@ -21,9 +21,6 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
 import java.util.Optional;
 
 import org.apache.camel.RuntimeCamelException;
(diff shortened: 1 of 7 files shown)


parent/pom.xml
```diff
@@ -102,7 +102,7 @@
         <jta-api-1.2-version>1.2</jta-api-1.2-version>
         <chicory-version>1.7.5</chicory-version>
         <chunk-templates-version>3.6.2</chunk-templates-version>
-        <citrus-version>4.10.1</citrus-version>
+        <citrus-version>5.0.0-M2</citrus-version>
```
davsclaus, line 105: Citrus `4.10.1` (GA) → `5.0.0-M2` (milestone). Milestone releases may have API instability. Is there a timeline for Citrus 5.0.0 GA relative to Camel 4.21?

dsl/camel-jbang/camel-jbang-plugin-test/src/main/java/org/apache/camel/dsl/jbang/core/commands/test/TestPlugin.java
```diff
@@ -139,50 +163,53 @@ private int executeInitCommand(JBangSupport citrus, List<String> args) {
                             .formatted(CitrusSettings.getApplicationPropertiesFile(), citrusApplicationProperties));
                 }
             }
+        }
 
-            return execute(citrus, "init", args);
+        /**
+         * Evaluate the current runtime Camel JBang or Camel Launcher. By default, assume Camel JBang.
+         */
```
davsclaus, line 172: Minor readability note: the double-negative logic (`"false".equals(System.getProperty("camel.launcher", "false"))`) is a bit subtle. A comment or inverting the check (e.g. `isCamelLauncherRuntime()` returning `"true".equals(...)`) might be clearer, though this is just a style observation.

davsclaus (commented):
Thanks for the PR @christophd — nice to see the test plugin embedding becoming viable with Citrus 5.

**Context:** This is the second attempt at embedding. The first (CAMEL-23147) was reverted in `9c452bf4bb7` due to classpath/dependency resolution issues. The Citrus 5.0.0-M2 redesign (direct picocli commands instead of spawning a JBang subprocess) is what makes this feasible now.

A few observations and questions:

1. **Milestone dependency** — Citrus moves from `4.10.1` (GA) to `5.0.0-M2` (milestone). Is there a timeline for Citrus 5.0.0 GA? If it lands before Camel 4.21, the risk is low. If not, Camel would ship with a pre-release dependency.

2. **No tests** — Given this is a retry of a previously-reverted change, even a basic test verifying the plugin loads and the subcommand structure registers correctly would add confidence.

3. **`run` command test directory navigation** — The old `executeRunCommand()` automatically navigated to the `test/` subdirectory and stripped the `test/` prefix from file paths. This logic is removed. Does Citrus 5's `Run` command handle this internally, or is this a behavioral change for users?

4. **`System.setProperty` cleanup** — The version properties (`citrus.jbang.version`, `citrus.camel.jbang.version`, `CitrusSettings.RESOURCES_WORKDIR_PROPERTY`) are set in the current JVM but never cleaned up. For a CLI tool this is fine, just noting for awareness.

parent/pom.xml
christophd: Citrus 5.0.0 GA is planned to be released within the next weeks in July. The M2 milestone is very stable already and if any there are only minor API changes to be expected on the road to GA.

christophd:
> 3. run command test directory navigation — The old executeRunCommand() automatically navigated to the test/ subdirectory and stripped the test/ prefix from file paths. This logic is removed. Does Citrus 5's Run command handle this internally, or is this a behavioral change for users?

yes, this is now being handled properly in Citrus run command

dsl/camel-jbang/camel-jbang-plugin-test/src/main/java/org/apache/camel/dsl/jbang/core/commands/test/TestPlugin.java
christophd: good catch! thanks! I have adjusted the method

parent/pom.xml
christophd: I am checking if the solution is also working with Citrus 4.10.1 ...

davsclaus:
Using M2 is fine, no problem, good to know its GA before Camel 4.22 LTS

davsclaus (approved):

PR #24247, opened 2026-06-25
CAMEL-19549: Replace Thread.sleep() with proper synchronization in camel-core tests

tmielke:
CAMEL-19549: Replace Thread.sleep() with proper synchronization in camel-core tests

Replaced or removed 17 Thread.sleep() calls across 15 test files to improve test reliability and reduce flakiness:
- Replaced with Awaitility for async completion waiting (5 files)
- Replaced with CountDownLatch for thread coordination (4 files)
- Removed unnecessary sleeps where operations are synchronous (6 files)

This eliminates race conditions from fixed timeouts and saves a few seconds of test execution time. Tests with legitimate timing-based behavior were kept unchanged.
All unit tests pass.

Made with help from AI tools.

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/direct/DirectProducerBlockingTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/direct/DirectProducerBlockingTest.java
@@ -16,8 +16,10 @@
  */
 package org.apache.camel.component.direct;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.CamelExchangeException;
 import org.apache.camel.CamelExecutionException;
@@ -74,12 +76,14 @@ public class DirectProducerBlockingTest extends ContextTestSupport {
     public void testProducerBlocksResumeTest() throws Exception {
         context.getRouteController().suspendRoute("foo");
 
+        CountDownLatch producerReady = new CountDownLatch(1);
         ExecutorService executor = Executors.newSingleThreadExecutor();
         executor.submit(new Runnable() {
             @Override
             public void run() {
                 try {
-                    Thread.sleep(200);
+                    // Wait for producer to start blocking
+                    assertTrue(producerReady.await(2, TimeUnit.SECONDS));
                     log.info("Resuming consumer");
                     context.getRouteController().resumeRoute("foo");
                 } catch (Exception e) {
@@ -90,6 +94,8 @@ public class DirectProducerBlockingTest extends ContextTestSupport {
 
         getMockEndpoint("mock:result").expectedMessageCount(1);
 
+        // Signal that we're about to send (producer will block)
+        producerReady.countDown();
         template.sendBody("direct:suspended?block=true&timeout=1000", "hello world");
 
         assertMockEndpointsSatisfied();
--- a/core/camel-core/src/test/java/org/apache/camel/component/seda/FileSedaShutdownCompleteAllTasksTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/seda/FileSedaShutdownCompleteAllTasksTest.java
@@ -52,16 +52,9 @@ public class FileSedaShutdownCompleteAllTasksTest extends ContextTestSupport {
                         .shutdownRunningTask(ShutdownRunningTask.CompleteAllTasks).to("log:delay").to("seda:foo");
 
                 from("seda:foo").routeId("route2").to("log:bar").to("mock:bar").process(new Processor() {
-                    boolean first = true;
-
                     @Override
                     public void process(Exchange exchange) throws Exception {
                         latch.countDown();
-                        if (first) {
-                            // only sleep on first
-                            Thread.sleep(100);
-                        }
-                        first = false;
                     }
                 });
             }
--- a/core/camel-core/src/test/java/org/apache/camel/processor/aggregator/DistributedCompletionIntervalTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/aggregator/DistributedCompletionIntervalTest.java
@@ -36,8 +36,6 @@ public class DistributedCompletionIntervalTest extends AbstractDistributedTest {
         MockEndpoint mock2 = getMockEndpoint2("mock:result");
         mock2.expectedMessageCount(0);
 
-        // ensure messages are send after the 1s
-        Thread.sleep(2000);
         sendMessages();
 
         mock.assertIsSatisfied();
@@ -51,8 +49,6 @@ public class DistributedCompletionIntervalTest extends AbstractDistributedTest {
         MockEndpoint mock2 = getMockEndpoint2("mock:result");
         mock2.expectedBodiesReceived("Message 19");
 
-        // ensure messages are send after the 1s
-        Thread.sleep(2000);
         sendMessages();
 
         mock2.assertIsSatisfied();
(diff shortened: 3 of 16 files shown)


core/camel-core/src/test/java/org/apache/camel/component/direct/DirectProducerBlockingTest.java
```diff
@@ -90,6 +94,8 @@ public void run() {
 
         getMockEndpoint("mock:result").expectedMessageCount(1);
 
+        // Signal that we're about to send (producer will block)
+        producerReady.countDown();
```
orpiske, line 98: **[Critical — Race condition]**

`producerReady.countDown()` fires here *before* `template.sendBody()` on the next line. The background thread can wake up, call `resumeRoute("foo")`, and complete before `sendBody()` enters its internal blocking loop.

If the route is already resumed when the producer tries to send, the test passes trivially (send to a running route) rather than testing recovery from a blocked producer — which was the original test's purpose.

The original `Thread.sleep(200)` was crude but gave `sendBody()` time to enter the blocking state. The fix needs a signal from *inside* the blocking logic, or at minimum an observable check (e.g., via Awaitility polling for the producer thread being in `WAITING` state or the endpoint showing inflight exchanges).

If no clean signal point exists, keeping the original sleep with a `// TODO` comment would be preferable to a latch that doesn't synchronize on the right event.

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/DistributedCompletionIntervalTest.java
```diff
@@ -36,8 +36,6 @@ public void testCamelContext1Wins() throws Exception {
         MockEndpoint mock2 = getMockEndpoint2("mock:result");
         mock2.expectedMessageCount(0);
```
orpiske, line 38: **[Critical — Timing dependency removed]**

The removed `Thread.sleep(2000)` was ensuring that the aggregator's completion interval timer (5s / 10s) had started ticking before messages arrived. This test uses a shared `MemoryAggregationRepository` between two CamelContexts and relies on the interval timer firing on only one of them.

Without the sleep, message arrival races with timer initialization, which could cause non-deterministic behavior.

Suggestions:
- Keep the sleep with a comment explaining it's for timer synchronization (not a synchronization artifact), or
- Use Awaitility to wait for a timer-started signal if one is observable, or
- Restructure the test to combine `completionSize` with `completionInterval` for determinism.

core/camel-core/src/test/java/org/apache/camel/component/seda/FileSedaShutdownCompleteAllTasksTest.java
```diff
@@ -52,16 +52,9 @@ public void configure() {
                         .shutdownRunningTask(ShutdownRunningTask.CompleteAllTasks).to("log:delay").to("seda:foo");
 
                 from("seda:foo").routeId("route2").to("log:bar").to("mock:bar").process(new Processor() {
-                    boolean first = true;
-
                     @Override
                     public void process(Exchange exchange) throws Exception {
```
orpiske, line 56: **[Important — Test semantics changed]**

The removed `Thread.sleep(100)` (and the `first` boolean) simulated a slow processor on the first message. The test's purpose is to verify that shutdown completes *all* tasks even when one is in-flight and slow. Without the sleep, all 5 messages process near-instantly and this scenario is no longer exercised.

This sleep was load-bearing for the test's intent, not a synchronization artifact. If the goal is to avoid `Thread.sleep`, consider using a `CountDownLatch` that the processor blocks on until the test thread signals it:

```java
CountDownLatch processorGate = new CountDownLatch(1);
// In processor: if (first) { processorGate.await(2, TimeUnit.SECONDS); }
// In test: processorGate.countDown(); then context.stop();
```

core/camel-core/src/test/java/org/apache/camel/processor/enricher/PollEnricherNoResourceTest.java
```diff
@@ -39,8 +39,6 @@ public void testNoResourceA() throws Exception {
     public void testResourceA() throws Exception {
         template.sendBody("seda:foo", "Bye World");
```
orpiske, line 40: **[Minor — Subtle behavior change]**

The seda send is asynchronous. The removed `Thread.sleep(250)` ensured the message was fully enqueued in the seda queue before `pollEnrich("seda:foo", 1000)` started polling. While `pollEnrich` has a 1-second timeout that should accommodate the async enqueue, the original test was deliberately ensuring the message was *present* before polling started — a subtly different test scenario (polling with data already available vs. polling and waiting for data to arrive).

This is likely safe in practice, but worth acknowledging the intent change.

orpiske (changes requested):
Thank you for this cleanup, Torsten — eliminating `Thread.sleep()` from tests is valuable work and directly aligned with the project's documented policy. Several of the changes here are genuinely good improvements (especially the `DefaultAsyncProcessorAwaitManagerTest` Awaitility rework and the `StopTimeoutRouteTest` dead-code removal).

However, a few changes introduce race conditions or silently weaken test coverage where the original sleep was load-bearing. I've left inline comments on the specific files.

**Summary of concerns (by severity):**

1. **[Critical]** `DirectProducerBlockingTest` — latch fires before the producer enters its blocking state, creating a race condition
2. **[Critical]** `DistributedCompletionIntervalTest` — removing 2-second sleeps likely breaks timing-dependent aggregator test
3. **[Important]** `FileSedaShutdownCompleteAllTasksTest` — removing sleep changes the test's purpose (no longer simulates slow in-flight processing during shutdown)
4. **[Minor]** `PollEnricherNoResourceTest` — removing sleep from async seda send subtly changes the tested scenario
5. **[Minor]** `XPathRouteConcurrentTest` — an entire test method (`testXPathTwoMessagesNotSameTime`) is deleted, not just the sleep. Worth noting explicitly in the PR description.

This review covers project rules and conventions only. It does not replace specialized review tools such as CodeRabbit or SonarCloud.

_This review was generated by an AI agent (Claude Code on behalf of Otavio Piske) and may contain inaccuracies. Please verify all suggestions before applying._

core/camel-core/src/test/java/org/apache/camel/component/direct/DirectProducerBlockingTest.java
tmielke: Agreed, I will refactor the test.

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/DistributedCompletionIntervalTest.java
orpiske: This is an incorrect assessment from the AI review. We can ignore it.

orpiske (approved):
Just re-reviewed this manually and some of these issues are not entirely correct. The PR should be OK to merge. Thanks @tmielke

core/camel-core/src/test/java/org/apache/camel/processor/aggregator/DistributedCompletionIntervalTest.java
tmielke: After another careful review and discussion with Otavio, we agree the removal of `Thread.sleep()` here is safe. 
Both `CamelContexts` and routes will have fully initialised before the test methods even start.
The aggregator completion timer will start once the route starts and should also be fully initialised by the time the `setUp()` method completes (defined in base class `AbstractDistributedTest`)
Hence, it is not required to sleep in the test methods and sending messages immediately should be just fine.
Otavio and I agreed to keep this change and remove the call to `Thread.sleep(2000)` in both test methods.

PR #24252, opened 2026-06-25
CAMEL-23463: camel-aws-bedrock - reject non-text models in invokeTextModel with a clear error

oscerd:
## What

`BedrockProducer#setResponseText` (reached only via the `invokeTextModel` operation) `switch`es on `modelId` to extract text from the model response. Several model IDs advertised as valid in `BedrockConfiguration.modelId`'s `@UriParam(enums=...)` are **not** text-generation models (image, embedding, rerank, video/speech). They fell through to the `default` branch and threw a confusing `IllegalStateException("Unexpected model: <modelId>")`, even though the model is well known — it simply isn't a text model.

Fixes [CAMEL-23463](https://issues.apache.org/jira/browse/CAMEL-23463).

## How

Replace the catch-all `default` for these declared-but-non-text models with explicit, categorized cases that fail fast with an actionable `IllegalArgumentException`:

| Category | Models | Message points to |
|---|---|---|
| Image | `amazon.titan-image-generator-v1`, `amazon.titan-image-generator-v2:0`, `amazon.nova-canvas-v1:0`, all `stability.*` | `invokeImageModel` |
| Embedding | `amazon.titan-embed-text-v1`, `amazon.titan-embed-image-v1`, `cohere.embed-english-v3`, `cohere.embed-multilingual-v3` | `invokeEmbeddingsModel` |
| Rerank | `amazon.rerank-v1:0`, `cohere.rerank-v3-5:0` | (not supported by `invokeTextModel`) |
| Video/Speech | `amazon.nova-reel-v1:0`, `amazon.nova-reel-v1:1`, `amazon.nova-sonic-v1:0` | (not supported by `invokeTextModel`) |

A genuinely unknown `modelId` still throws `IllegalStateException("Unexpected model: ...")`.

### Why a clear error instead of auto-routing

The Jira issue offered two options (auto-route to the right operation, or validate/document). This PR takes the clear-error route because:

- The operation is explicitly chosen by the user (URI or `CamelAwsBedrockOperation` header); silently overriding it would be surprising.
- Each operation has a different response contract (`invokeTextModel` extracts text, `invokeImageModel` extracts the `images` node, `invokeEmbeddingsModel` returns the raw response) — auto-routing would break the contract of the operation the user asked for.
- Rerank/video/speech models have no corresponding operation, so auto-routing cannot cover them anyway.

The check stays at response time (where the old exception was thrown) rather than endpoint creation, because the operation can be overridden per message via `CamelAwsBedrockOperation`, so the model/operation pairing isn't reliably known at endpoint creation.

## Tests

`BedrockInvokeTextModelUnsupportedModelTest` (mocked `BedrockRuntimeClient`, no AWS access):
- Parameterized over all 16 non-text models — asserts an `IllegalArgumentException` naming the offending model and the `invokeTextModel` operation, and pointing to `invokeImageModel` / `invokeEmbeddingsModel` where a dedicated operation exists.
- Happy-path test confirming a text model (`amazon.titan-text-express-v1`) still works.

## Docs

Added a short note to `aws-bedrock-component.adoc` clarifying that the operation must match the model type.

---

### DIFF
--- a/components/camel-aws/camel-aws-bedrock/src/main/docs/aws-bedrock-component.adoc
+++ b/components/camel-aws/camel-aws-bedrock/src/main/docs/aws-bedrock-component.adoc
@@ -83,6 +83,12 @@ Camel-AWS Bedrock component provides the following operation on the producer sid
 - converseStream
 - applyGuardrail
 
+The operation must match the model type. Text-generation models are used with `invokeTextModel`, image-generation
+models (for example `amazon.titan-image-generator-v1`, `amazon.nova-canvas-v1:0`, the `stability.*` models) with
+`invokeImageModel`, and embedding models (for example `amazon.titan-embed-text-v1`, `cohere.embed-english-v3`) with
+`invokeEmbeddingsModel`. Invoking `invokeTextModel` with an image, embedding, rerank or video/speech model fails fast
+with an `IllegalArgumentException` that names the model and points to the operation to use instead.
+
 === Streaming Support
 
 The streaming operations (`invokeTextModelStreaming`, `invokeImageModelStreaming`, `invokeEmbeddingsModelStreaming`) enable real-time streaming of model responses, providing lower latency and better user experience for interactive applications.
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Clean, well-scoped PR. All 16 non-text models from `BedrockModels` are covered by the new rejection cases with correct model ID strings. Tests are thorough (parameterized over all 16 + happy-path), documentation is clear, and the design rationale (clear error over auto-routing) is sound.

One minor observation: `amazon.titan-embed-text-v2:0` remains in the text-model section (line 282, from CAMEL-20765) while `titan-embed-text-v1` is now rejected as an embedding model. This is consistent with existing behavior and likely intentional, but worth a note if so.

PR #24269, opened 2026-06-26
CAMEL-23462: camel-aws-bedrock - reject wrong-type body in invoke*Model POJO mode with a clear error

oscerd:
## What this PR does

Fixes [CAMEL-23462](https://issues.apache.org/jira/browse/CAMEL-23462).

The `invokeTextModel`, `invokeImageModel`, and `invokeEmbeddingsModel` operations in `BedrockProducer` only acted when `pojoRequest=true` **and** the message body was an `InvokeModelRequest`. For any other body type (e.g. `String`, `Map`), the `if (payload instanceof InvokeModelRequest ...)` block was skipped with no `else`, so the producer returned silently — leaving the caller with a `null` body and no indication of what went wrong.

The `converse`, `converseStream`, and `applyGuardrail` operations already throw an `IllegalArgumentException` for the same misconfiguration. This PR makes the invoke operations consistent.

## Changes

- **`BedrockProducer`**: add an `else throw new IllegalArgumentException(...)` branch to:
  - `invokeTextModel`
  - `invokeImageModel`
  - `invokeEmbeddingsModel`
  - `invokeTextModelStreaming` — the same silent no-op existed on the streaming path; this also covers `invokeImageModelStreaming` and `invokeEmbeddingsModelStreaming`, which delegate to it.
- **New `BedrockInvokeModelPojoRequestTest`**: 5 mock-based unit tests (no AWS credentials / network) verifying each operation fails fast with the expected `IllegalArgumentException` type and message when given a wrong-type POJO body. The exception is raised before any client call, so a mocked client suffices.

## Behaviour change

| | Before | After |
|---|--------|-------|
| Wrong-type body, `pojoRequest=true` | silent `null` output, no error | `IllegalArgumentException` naming the required body type |

This only affects already-broken/misconfigured routes (a wrong body type produced no usable output before). No default, option, header, or public API signature changes, so no upgrade-guide entry is required.

## Testing

`mvn clean install -Psourcecheck` in `components/camel-aws/camel-aws-bedrock` → **BUILD SUCCESS**, 122 tests pass (including the 5 new ones), formatter/impsort clean.

---

### DIFF
--- a/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/runtime/BedrockProducer.java
+++ b/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/runtime/BedrockProducer.java
@@ -142,6 +142,9 @@ public class BedrockProducer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 setResponseText(result, message);
+            } else {
+                throw new IllegalArgumentException(
+                        "InvokeTextModel operation requires InvokeModelRequest in POJO mode");
             }
         } else {
             InvokeModelRequest.Builder builder = InvokeModelRequest.builder();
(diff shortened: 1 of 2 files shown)


davsclaus (approved):
Straightforward fail-fast fix that makes the `invoke*Model` operations consistent with the existing `converse`, `converseStream`, and `applyGuardrail` operations — all of which already throw `IllegalArgumentException` for wrong-type POJO bodies. The error messages follow the same `"<Operation> operation requires <Type> in POJO mode"` convention. Well-tested with 5 new unit tests covering each operation path. No behavioral regression — the `else` branches only fire on already-broken configurations that silently returned null.

PR #24276, opened 2026-06-27
CAMEL-23654: camel-jbang - group commands in --help output by categor…

sk-reddy17:
The `camel --help` output lists 40+ commands in one flat alphabetical list,
  which is hard to scan. This groups them into categories (Running, Monitoring,
  Actions, Development, Configuration, Catalog, AI) so it's easier to find what
  you need.

  Done with a small picocli `IHelpSectionRenderer` wired into `CamelJBangMain`.
  Anything not mapped to a group falls into "Other", so no command gets dropped.

  Fixes https://issues.apache.org/jira/browse/CAMEL-23654

  ## Example

  Commands:

    Running:
      run          Run as local Camel integration
      dev          Run in dev mode with live reload
      stop         Shuts down running Camel integrations

    Monitoring:
      get          Get status of Camel integrations
      top          Top status of Camel integrations

    Configuration:
      config       Get and set user configuration values

    Catalog:
      catalog      List artifacts from Camel Catalog

    AI:
      ask          Ask a question about a running Camel application using AI
    ...

  ## Test plan

  - [x] Test checks all the category headers show up in `--help`
  - [x] Test checks the key commands (run, get, config, catalog) are still listed
  - [x] Ran `camel --help` by hand to confirm it looks right

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/GroupedCommandHelpRenderer.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/GroupedCommandHelpRenderer.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.dsl.jbang.core.commands;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import picocli.CommandLine;
+
+/**
+ * Renders the camel CLI subcommand list grouped by functional category, replacing the default flat alphabetical list.
+ */
+public class GroupedCommandHelpRenderer implements CommandLine.IHelpSectionRenderer {
+
+    private static final Map<String, List<String>> GROUPS = new LinkedHashMap<>();
+
+    static {
+        GROUPS.put("Running", List.of("run", "dev", "stop", "restart", "ps", "log", "shell", "script"));
+        GROUPS.put("Monitoring", List.of("get", "top", "trace", "hawtio", "jolokia"));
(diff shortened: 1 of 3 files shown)


davsclaus (commented):
Thanks for working on this, @sk-reddy17 — grouping the help output is a welcome UX improvement. The categorizations look well thought out and all 37 built-in commands are covered.

A couple of minor suggestions for robustness (not blockers):

1. **Hidden-command filtering**: The renderer iterates over all `subcommands()` without checking `usageMessage().hidden()`. The default picocli renderer filters hidden commands out. While there are no hidden top-level commands today, if a plugin registers one in the future it would unexpectedly appear in help. A one-line guard in both loops would future-proof this:
   ```java
   if (sub.getCommandSpec().usageMessage().hidden()) continue;
   ```

2. **Maintenance guardrail**: The static `GROUPS` map needs manual updating when new commands are added. The "Other" fallback handles this gracefully at runtime, but a test asserting that no *built-in* commands fall into "Other" would catch drift early — so developers adding a new command get a test failure reminding them to update the map.

3. **Test coverage**: The tests check group headers and a few key commands, but don't verify that *all* registered commands appear in the output. A count-based assertion would be a stronger safety net.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/GroupedCommandHelpRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+        Map<String, CommandLine> subcommands = help.commandSpec().subcommands();
+        if (subcommands.isEmpty()) {
+            return "";
+        }
+
+        int nameWidth = subcommands.keySet().stream().mapToInt(String::length).max().orElse(10);
+
+        List<String> assigned = new ArrayList<>();
+        StringBuilder sb = new StringBuilder();
+
```
davsclaus, line 55: Minor: the default picocli renderer skips commands with `hidden = true`. Consider adding a guard here and in the ungrouped loop below:
```suggestion
                if (sub != null) {
                    if (!sub.getCommandSpec().usageMessage().hidden()) {
                        present.add(sub);
                    }
                    assigned.add(name);
```
This way hidden commands are still tracked as "assigned" (so they don't leak into "Other") but won't be rendered.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/GroupedCommandHelpRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+                CommandLine sub = subcommands.get(name);
+                if (sub != null) {
+                    present.add(sub);
+                    assigned.add(name);
+                }
+            }
+            if (!present.isEmpty()) {
+                sb.append(String.format("%n  %s:%n", entry.getKey()));
+                for (CommandLine sub : present) {
+                    appendCommand(sb, sub, nameWidth);
```
davsclaus, line 72: Same hidden-command concern here — ungrouped commands from plugins could be hidden:
```suggestion
            if (!assigned.contains(entry.getKey()) && !entry.getValue().getCommandSpec().usageMessage().hidden()) {
```

davsclaus (approved):

PR #24283, opened 2026-06-28
CAMEL-23840: camel-core: pollEnrich with cacheSize(-1) does not disable consumer cache (dynamic endpoints)

msnijder30:
# Description

I found out about this issue with pollEnrich because we had a memory issue with the SFTP component because we did not set the cache size so it defaulted to 1000. So I read the documentation and set it to cachesize to -1, but the memory still kept going up.

Our code was using dynamic endpoints that triggered this.

pollEnrich().cacheSize(-1) is documented to disable the consumer cache entirely - polling consumers should be created fresh for each call and discarded immediately after use. However, this does not appear to work.

Root cause: PollEnricher.doBuild() unconditionally creates a DefaultConsumerCache regardless of cacheSize. DefaultConsumerCache then normalizes any cacheSize <= 0 to the CamelContext maximum cache pool size (default 1000):

```java
// DefaultConsumerCache line 55
this.maxCacheSize = cacheSize <= 0 ? CamelContextHelper.getMaximumCachePoolSize(camelContext) : cacheSize;
```

So instead of discarding consumers, up to 1000 polling consumers are retained in a cache. 

For resource-backed components (e.g., SFTP), each retained consumer represents an open connection that is never cleaned up until the route stops.

## Reproducer
You can find my reproducer here: https://github.com/msnijder30/reproduce-camel-pollenrich-cache-issue
Also attached as zip to the JIRA ticket.

## The fix
This mirrors how DefaultProducerCache (via EmptyProducerCache) already works on the producer side for cacheSize(-1).

Didn't create a DefaultConsumerCache class because it's only used in this class, unlike DefaultProducerCache which is used in multiple places.

Let me know what you think 👍 it was interesting to run into this bug

# Target

# Tracking
- [X] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue]
- https://issues.apache.org/jira/browse/CAMEL-23840

# AI-assisted contributions

- Used OpenCode with DeepSeek V4 Pro

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
@@ -168,7 +168,7 @@ public class PollEnricher extends BaseProcessorSupport implements IdAware, Route
     }
 
     public EndpointUtilizationStatistics getEndpointUtilizationStatistics() {
-        return consumerCache.getEndpointUtilizationStatistics();
+        return consumerCache != null ? consumerCache.getEndpointUtilizationStatistics() : null;
     }
 
     public AggregationStrategy getAggregationStrategy() {
@@ -313,8 +313,13 @@ public class PollEnricher extends BaseProcessorSupport implements IdAware, Route
                 prototype = false;
             }
 
-            // acquire the consumer from the cache
-            consumer = consumerCache.acquirePollingConsumer(endpoint);
+            // acquire the consumer, either from cache or create a new one
+            if (consumerCache != null) {
+                consumer = consumerCache.acquirePollingConsumer(endpoint);
+            } else {
+                consumer = endpoint.createPollingConsumer();
+                ServiceHelper.startService(consumer);
+            }
         } catch (Exception e) {
             if (isIgnoreInvalidEndpoint()) {
                 if (LOG.isDebugEnabled()) {
@@ -361,8 +366,13 @@ public class PollEnricher extends BaseProcessorSupport implements IdAware, Route
             callback.done(true);
             return true;
         } finally {
-            // return the consumer back to the cache
-            consumerCache.releasePollingConsumer(endpoint, consumer);
+            if (consumerCache != null) {
+                // return the consumer back to the cache
+                consumerCache.releasePollingConsumer(endpoint, consumer);
+            } else {
+                // consumer cache disabled, stop the consumer immediately
+                ServiceHelper.stopAndShutdownService(consumer);
+            }
             // and stop prototype endpoints
             if (prototype) {
                 ServiceHelper.stopAndShutdownService(endpoint);
@@ -543,7 +553,7 @@ public class PollEnricher extends BaseProcessorSupport implements IdAware, Route
 
     @Override
     protected void doBuild() throws Exception {
-        if (consumerCache == null) {
+        if (consumerCache == null && cacheSize >= 0) {
             // create consumer cache if we use dynamic expressions for computing the endpoints to poll
             consumerCache = new DefaultConsumerCache(this, camelContext, cacheSize);
             LOG.debug("PollEnrich {} using ConsumerCache with cacheSize={}", this, cacheSize);
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
@@ -49,6 +49,7 @@ import org.apache.camel.support.EndpointHelper;
 import org.apache.camel.support.EventDrivenPollingConsumer;
 import org.apache.camel.support.ExchangeHelper;
 import org.apache.camel.support.cache.DefaultConsumerCache;
+import org.apache.camel.support.cache.EmptyConsumerCache;
 import org.apache.camel.support.service.ServiceHelper;
 import org.apache.camel.util.URISupport;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


Croway (approved):

core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
```diff
@@ -168,7 +168,7 @@ public Expression getExpression() {
     }
 
     public EndpointUtilizationStatistics getEndpointUtilizationStatistics() {
-        return consumerCache.getEndpointUtilizationStatistics();
+        return consumerCache != null ? consumerCache.getEndpointUtilizationStatistics() : null;
```
davsclaus: This null check would not be needed if an `EmptyConsumerCache` class existed — analogous to `EmptyProducerCache` in `core/camel-support/src/main/java/org/apache/camel/support/cache/EmptyProducerCache.java`.

The `EmptyConsumerCache` would override `acquirePollingConsumer()` to always create a new consumer (via `endpoint.createPollingConsumer()` + `ServiceHelper.startService()`) and `releasePollingConsumer()` to always `ServiceHelper.stopAndShutdownService()`. It should also override `getCapacity()` to return `0`.

Then `PollEnricher.doBuild()` would become:
```suggestion
        return consumerCache.getEndpointUtilizationStatistics();
```

core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
```diff
@@ -543,7 +553,7 @@ protected static String resolveScheme(Exchange exchange, String uri) {
 
     @Override
     protected void doBuild() throws Exception {
-        if (consumerCache == null) {
+        if (consumerCache == null && cacheSize >= 0) {
```
davsclaus: Instead of skipping cache creation, use the `EmptyConsumerCache` pattern (like `SendDynamicProcessor` does with `EmptyProducerCache`):

```java
if (consumerCache == null) {
    if (cacheSize < 0) {
        consumerCache = new EmptyConsumerCache(this, camelContext);
        LOG.debug("PollEnrich {} is not using ConsumerCache", this);
    } else {
        consumerCache = new DefaultConsumerCache(this, camelContext, cacheSize);
        LOG.debug("PollEnrich {} using ConsumerCache with cacheSize={}", this, cacheSize);
    }
}
```

See `SendDynamicProcessor.doBuild()` for the equivalent producer-side pattern.

core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
```diff
@@ -313,8 +313,13 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
...
+            // acquire the consumer, either from cache or create a new one
+            if (consumerCache != null) {
+                consumer = consumerCache.acquirePollingConsumer(endpoint);
+            } else {
+                consumer = endpoint.createPollingConsumer();
+                ServiceHelper.startService(consumer);
+            }
         } catch (Exception e) {
             if (isIgnoreInvalidEndpoint()) {
                 if (LOG.isDebugEnabled()) {
```
davsclaus, line 321: With `EmptyConsumerCache`, the acquire/release contract stays the same regardless of cache mode — this entire `if/else` block would collapse back to:
```java
consumer = consumerCache.acquirePollingConsumer(endpoint);
```

core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
```diff
@@ -361,8 +366,13 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
...
+            if (consumerCache != null) {
+                // return the consumer back to the cache
+                consumerCache.releasePollingConsumer(endpoint, consumer);
+            } else {
+                // consumer cache disabled, stop the consumer immediately
+                ServiceHelper.stopAndShutdownService(consumer);
+            }
             // and stop prototype endpoints
             if (prototype) {
                 ServiceHelper.stopAndShutdownService(endpoint);
```
davsclaus, line 369: Same here — with `EmptyConsumerCache` this collapses back to:
```java
consumerCache.releasePollingConsumer(endpoint, consumer);
```

davsclaus (changes requested):
Thanks for the fix and the clear root-cause analysis — the bug is real and the test is well done.

However, the approach of setting `consumerCache = null` and scattering `if (consumerCache != null)` checks diverges from the established pattern on the producer side. `SendDynamicProcessor`, `RecipientList`, and `RoutingSlip` all use `EmptyProducerCache` (extends `DefaultProducerCache`) when `cacheSize < 0` — it overrides `acquireProducer()` to always create new, and `releaseProducer()` to always stop/shutdown. No null checks needed in the calling code.

Please create an `EmptyConsumerCache` (extends `DefaultConsumerCache`) in `core/camel-support/src/main/java/org/apache/camel/support/cache/` that mirrors `EmptyProducerCache`, and use it in `PollEnricher.doBuild()` so that `consumerCache` is never null. This keeps the codebase consistent and avoids the null checks.

The test additions are great — please keep those.

core/camel-core-processor/src/main/java/org/apache/camel/processor/PollEnricher.java
msnijder30: Thanks for the review, good points, better to keep it similar with the producer side. 

I have implemented the comments, some changes also required in DefaultConsumerCache.java so it does not create the service pool (basically identical changes to the producer).

The EmptyConsumerCache is also basically identical to the producer with the required changes to make it work for the consumer.

Let me know what you think - this solution still solves the issue also, tested with my reproducer application.

davsclaus (approved):
Thanks for the update — this now correctly mirrors the `EmptyProducerCache` / `DefaultProducerCache` pattern. The `DefaultConsumerCache` changes match `DefaultProducerCache`'s null-pool handling, `EmptyConsumerCache` mirrors `EmptyProducerCache`, and `PollEnricher.doBuild()` follows `SendDynamicProcessor.doBuild()`. Clean, consistent, and well tested.

PR #24284, opened 2026-06-28
CAMEL-23841: camel-jbang - fix TUI layout overflow and add minimum size guard

ammachado:
# Description

Fix TUI layout overflow issues visible at 120×43 (the primary target terminal size) and add a minimum-size guard for smaller terminals.

**Tab bar overflow**: the full tab bar needs ~126 chars and clips `0 More▾` to a bare `0` at 120 columns. Two display modes are now used: full labels (` 1 Overview | 2 Log | ...`) at ≥126 cols, compact labels (`1 Overview|2 Log|...`, ~88 chars) at 88–125 cols. Badge X-position calculation was updated to match the active divider width.

**Footer overflow**: some tabs have footers that exceed terminal width by 50–80 chars (Diagram: ~170–186, Errors: ~169–199), clipping mid-word. On overflow the footer now drops content in priority order: decorative right-side spans (MCP status, recording) first, then secondary F-key hints (F2/F3/F6), and only then tab-specific action hints (stop/kill/restart). F1 help and the first tab hint are always preserved.

The secondary F-key drop logic is extracted into a package-private static `dropFKeyHints` helper so it can be unit tested. It removes hint pairs from the tail (F6 first, then F3, F2) and stops before the first pair, so F1 (help) always survives even on a narrow terminal.

**Minimum size guard**: when the terminal is below 88×24, all rendering is skipped and a btop-style centered message is shown with the current and required dimensions (out-of-range values in orange). Normal rendering resumes automatically on resize — no new event-handling code is needed since TamboUI already re-invokes `render(Frame)` with current dimensions on every resize.

# Target

# Tracking

JIRA: https://issues.apache.org/jira/browse/CAMEL-23841

# AI-assisted contributions

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
@@ -99,6 +99,12 @@ public class CamelMonitor extends CamelCommand {
     private static final int MAX_TRACES = 200;
     private static final int NUM_TABS = 10;
 
+    // Compact tab bar (10 labels + 9 "|" dividers) needs 88 chars — that is the true minimum
+    private static final int MIN_WIDTH = 88;
+    private static final int MIN_HEIGHT = 24;
+    // Full tab bar (10 labels + 9 " | " dividers) needs 126 chars; use compact below that
+    private static final int TABS_FULL_MIN_WIDTH = 126;
+
     // Tab indices
     private static final int TAB_OVERVIEW = 0;
     private static final int TAB_LOG = 1;
@@ -968,6 +974,11 @@ public class CamelMonitor extends CamelCommand {
     private void render(Frame frame) {
         Rect area = frame.area();
 
+        if (area.width() < MIN_WIDTH || area.height() < MIN_HEIGHT) {
+            renderTooSmall(frame, area);
+            return;
+        }
+
         // Layout: header (1 row) + spacer (1 row) + tabs (2 rows) + spacer (1 row) + content (fill) + footer (1 row)
         List<Rect> mainChunks = Layout.vertical()
                 .constraints(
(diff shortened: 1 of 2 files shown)


dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
```diff
@@ -1733,9 +1810,34 @@ private void renderFooter(Frame frame, Rect area) {
...
+        if (hintsWidth + rightWidth + minGap > area.width()) {
+            // Drop decorative right-side content first
+            rightSpans.clear();
+            rightWidth = 0;
+            minGap = 0;
+            // Drop secondary F-key hints (F2/F3/F6) before tab-specific action hints.
+            // They are inserted at position 2 (after the first tab hint), so the last
+            // secondary F-key pair sits at index (2 + secondaryFKey - 2).
+            while (secondaryFKey > 0 && hintsWidth > area.width()) {
+                int dropIdx = 2 + secondaryFKey - 2;
```
davsclaus: Bug: when F1 is present (the common case — most tabs have help text), this formula points to F3 instead of F6, and the loop ends up removing F1 while keeping F6.

`insertFKeyHints` inserts [F1_k, F1_l, F2_k, F2_l, F3_k, F3_l, F6_k, F6_l] at position 2. The secondary spans (F2/F3/F6) start at index **4**, not 2. But `2 + secondaryFKey - 2` = `secondaryFKey` = 6, which points to F3 (index 6), not F6 (index 8). After three iterations, F1 at [2,3] gets removed while F6 survives — the opposite of the stated intent.

**Suggested fix:** have `insertFKeyHints` return `fKeySpans.size()` (total count including F1) instead of `fKeySpans.size() - (hasHelp ? 2 : 0)`. Then use `dropIdx = fKeyTotal` here, with loop condition `fKeyTotal > 2`. This naturally drops from the end (F6 first) and stops before F1:

```java
while (fKeyTotal > 2 && hintsWidth > area.width()) {
    int dropIdx = fKeyTotal;
    Span labelSpan = spans.remove(dropIdx + 1);
    Span keySpan = spans.remove(dropIdx);
    hintsWidth -= keySpan.width() + labelSpan.width();
    fKeyTotal -= 2;
}
```

Same change needed in `renderOverviewFooter`.

davsclaus (changes requested):
Nice work on the compact tab bar and the priority-based footer truncation — the approach is solid and the minimum-size guard is a good addition.

One bug in the footer overflow drop loop: the index formula removes F1 (which should be preserved) while keeping F6, when `hasHelp=true` (the common case). Details in the inline comment.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
ammachado: Good catch, you're exactly right. I traced it: with help text present, `insertFKeyHints` returned the secondary-only count (`size - 2 = 6`), and `dropIdx = 2 + 6 - 2 = 6` pointed at the F3 pair, so the loop removed F3, F2, then **F1** while keeping F6, the opposite of the intent.

Applied your suggested fix in `40b4c8b`:

- `insertFKeyHints` now returns the total F-key span count (`fKeySpans.size()`).
- Extracted the drop logic into a testable static `dropFKeyHints(spans, fKeyTotal, hintsWidth, available)` that removes pairs from the tail (F6, then F3, F2) and stops at 2, so the first pair (F1 help when present) is always preserved. Renamed `secondaryFKey` to `fKeyTotal` for accuracy.

Added 4 unit tests in `CamelMonitorTest` covering the help, no-help, no-overflow, and heavy-overflow cases. I verified they fail against the original index formula and pass with the fix.

davsclaus (approved):

PR #24318, opened 2026-06-29
fix: make TUI charts fill available width on wide terminals

gnodet:
## Summary

Charts previously used a fixed 60-point history buffer which left significant empty space on wide terminals. This PR makes charts dynamically size to fill the available width.

### Changes
- **MetricsCollector**: Increase history buffer caps from 60 to 300 points so enough data accumulates for wider terminals
- **OverviewTab**: Increase `MAX_SPARKLINE_POINTS` from 60 to 300 (bar chart `renderPoints` was already computed dynamically from available width)
- **EndpointsTab**: Increase `MAX_CHART_POINTS` from 60 to 300 and compute `renderPoints` dynamically from the chart area width in all three chart methods (`renderEndpointFlow`, `renderSingleEndpointChart`, `renderPayloadSizeChart`)

## Test plan
- [x] Build passes locally
- [x] Verified visually — sparklines and bar charts now fill their panels on wide terminals
- [ ] CI passes

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/EndpointsTab.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/EndpointsTab.java
@@ -51,7 +51,7 @@ import static org.apache.camel.dsl.jbang.core.commands.tui.MonitorContext.*;
 class EndpointsTab implements MonitorTab {
 
     private static final String[] SORT_COLUMNS = { "component", "route", "dir", "total", "body", "hdr", "uri" };
-    private static final int MAX_CHART_POINTS = 60;
+    private static final int MAX_CHART_POINTS = 300;
     private static final int CHART_ALL = 0;
     private static final int CHART_SINGLE = 1;
     private static final int CHART_OFF = 2;
@@ -426,7 +426,7 @@ class EndpointsTab implements MonitorTab {
         LinkedList<Long> inHist = inHistMap.getOrDefault(pid, new LinkedList<>());
         LinkedList<Long> outHist = outHistMap.getOrDefault(pid, new LinkedList<>());
 
-        int renderPoints = MAX_CHART_POINTS;
+        int renderPoints = Math.min(MAX_CHART_POINTS, Math.max(2, hSplit.get(1).width() - 6));
         long[] inArr = new long[renderPoints];
         long[] outArr = new long[renderPoints];
         for (int i = 0; i < renderPoints; i++) {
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/EndpointsTab.java
```diff
@@ -51,7 +51,7 @@
 class EndpointsTab implements MonitorTab {
 
     private static final String[] SORT_COLUMNS = { "component", "route", "dir", "total", "body", "hdr", "uri" };
-    private static final int MAX_CHART_POINTS = 60;
+    private static final int MAX_CHART_POINTS = 300;
```
davsclaus, line 54: `CircuitBreakerTab` also has `MAX_CHART_POINTS = 60` and a hardcoded `renderPoints = MAX_CHART_POINTS` (line 355) — same pattern this PR fixes here. The MetricsCollector buffer is already bumped (CB history uses `MAX_ENDPOINT_CHART_POINTS` via `recordEndpointSample`), so only the render side needs the same `Math.min(MAX, Math.max(2, width - 6))` treatment. Could be a follow-up.

davsclaus (commented):
Nice cleanup — the dynamic sizing logic is correct and the buffer increase is well-scoped.

One thing I noticed: `CircuitBreakerTab` has the same fixed-60 chart sizing (`MAX_CHART_POINTS = 60` at line 52, `renderPoints = MAX_CHART_POINTS` at line 355) that this PR fixes in `EndpointsTab`. The CB history buffer in `MetricsCollector` is already bumped to 300 (it goes through `recordEndpointSample` → `MAX_ENDPOINT_CHART_POINTS`), but the render side is still hardcoded. Could be addressed in a follow-up.

Also a pre-existing nit: `OverviewTab` declares its own `MAX_SPARKLINE_POINTS` instead of referencing `MetricsCollector.MAX_SPARKLINE_POINTS` — they could drift.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/EndpointsTab.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good catch! Applied the same treatment to `CircuitBreakerTab`: bumped `MAX_CHART_POINTS` to 300, computed `renderPoints` dynamically from `vSplit.get(1).width() - 6`, and made the x-axis labels dynamic too. See 398034a.

davsclaus (approved):

PR #24320, opened 2026-06-29
CAMEL-23338: added OpenSearchClient as parameter to pass custom instance

renjth-81:
# Description

Added OpenSearchClient as a parameter to allow custom OpenSearchClient instance to be passed.

# Target

- [ ✓] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking

- [ ✓] I checked that each commit in the pull request has a meaningful subject line and body.

- [ ✓] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

# AI-assisted contributions

- [ x] If this PR includes AI-generated code, commits have proper co-authorship attribution (e.g., `Co-authored-by` trailers) and the PR description identifies the AI tool used.


davsclaus (approved):

davsclaus:
you need to also rebuild camel-catalog, and some code in component and endpoint dsl

davsclaus:
There are uncommitted changes
HEAD detached at pull/24320/merge
	modified:   catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/opensearch.json
	modified:   dsl/camel-componentdsl/src/generated/java/org/apache/camel/builder/component/dsl/OpensearchComponentBuilderFactory.java
	modified:   dsl/camel-endpointdsl/src/generated/java/org/apache/camel/builder/endpoint/dsl/OpensearchEndpointBuilderFactory.java

oscerd (changes requested):
Thanks for the contribution, @renjth-81 — being able to pass a custom `OpenSearchClient` is a useful addition, and the generated catalog/DSL files are all correctly regenerated. However, there's a blocking correctness problem in the current implementation that CI is also catching, so I'm requesting changes.

## Blocker: the cached transport captures a `null` RestClient → NPE on every exchange

`OpensearchProducer` now builds and caches `openSearchClient` in its **constructor**:

- the constructor sets `this.client = endpoint.getClient();` then `this.openSearchClient = createDefaultOpenSearchClient();`
- `createDefaultOpenSearchClient()` builds `new OpenSearchClient(new RestClientTransport(client, ...))`, but only calls `startClient()` when `configuration.isDisconnect() && client == null`.

In the standard configuration (a `RestClient` built from `hostAddresses`, no autowired client), `endpoint.getClient()` is **null at construction time** — the component creates the endpoint with a null client by default (`OpensearchComponent.java:91`), and the real `RestClient` is created lazily later in `doStart() → startClient()` (`OpensearchProducer.java`, the double-checked-locking block around lines 452–467). `disconnect` defaults to `false` (`OpensearchConfiguration.java:60`), so `createDefaultOpenSearchClient()` does **not** start the client and ends up wrapping a `RestClientTransport` around a `null` `RestClient`.

`process()` then does `OpenSearchTransport transport = openSearchClient._transport();` and reuses that cached transport, so every exchange dereferences the null `RestClient`:

```
java.lang.NullPointerException: Cannot invoke
  "org.opensearch.client.RestClient.performRequestAsync(...)" because "this.restClient" is null
```

This is exactly what CI is failing on — `build (17, false)` and `build (25, false)` are both red, with `OpensearchPingIT.testPing` and all `OpensearchBulkIT.*` failing with this NPE. The user-supplied-client path (`endpoint.getOpenSearchClient() != null`) works, but the default path — which all of the component's own integration tests exercise — is broken.

### Suggested direction

Resolve the transport lazily rather than in the constructor: keep using the user-supplied `OpenSearchClient` when one is provided, but for the default case build the transport in `process()` / after `doStart()` once `client` is non-null (as the original code did). Please also keep the `disconnect=true` semantics in mind — the old `process()` rebuilt the transport per call and `cleanup()` closes and nulls the client after each exchange (`OpensearchProducer.java`, around lines 437–445), so a single cached transport would also break disconnect mode after the first exchange.

## Other items

- **Tests:** please add a test proving a custom `OpenSearchClient` instance is actually honored at runtime — the new option currently has no coverage, and the existing integration tests regressed.
- **Jira:** apache/camel changes should be tracked by a `CAMEL-XXXXX` issue with the commit/PR title prefixed accordingly. Could you create one (component `camel-opensearch`) and update the title? The `added missing files` commit subject could also be made more descriptive.

For context: this PR currently shows as approved, but that review was on the very first commit (`62b5af1`), before the generated-files commit and before CI ran — so it predates the failures above.

Thanks again for working on this — happy to re-review once the transport init is made lazy and CI is green.

---

davsclaus:
creating a new client should ideally not be done in constructors but in doInit or doStart

renjth-81:
@davsclaus @oscerd I have pushed the fixes and integration test. Please could you review

components/camel-opensearch/src/test/java/org/apache/camel/component/opensearch/integration/CustomOpensearchClientPingIT.java
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.opensearch.integration;
+
+import java.io.IOException;
+
+import org.apache.camel.builder.RouteBuilder;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
```
davsclaus, line 23: **Missing dependency:** `mockito-core` is not declared as a test dependency in `components/camel-opensearch/pom.xml` and is not transitively available. This import (and all Mockito usage in this file) will cause a compilation failure.

Please add to `pom.xml`:
```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <version>${mockito-version}</version>
    <scope>test</scope>
</dependency>
```

components/camel-opensearch/src/main/java/org/apache/camel/component/opensearch/OpensearchProducer.java
```diff
@@ -154,12 +156,17 @@ private OpensearchOperation resolveOperation(Exchange exchange) {
...
+                final ObjectMapper mapper = new ObjectMapper();
+                mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+                transport = new RestClientTransport(client, new JacksonJsonpMapper(mapper));
+            } else {
+                transport = openSearchClient._transport();
             }
-            final ObjectMapper mapper = new ObjectMapper();
-            mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
-            OpenSearchTransport transport = new RestClientTransport(client, new JacksonJsonpMapper(mapper));
             // 2. Index and type will be set by:
```
davsclaus, line 170: **Transport lifecycle bug with `disconnect=true`:** When a custom `OpenSearchClient` is provided and `disconnect=true`, `cleanup()` (line ~438) calls `IOHelper.close(ctx.transport())`. In this branch, `transport` is `openSearchClient._transport()` — the user's own transport. Closing it will corrupt the user's client.

The default path creates a fresh `RestClientTransport` per call, so closing it is safe. But the custom client's transport should not be closed by the component.

Suggested fix: pass a flag into `ActionContext` indicating whether the transport is user-owned, and skip `IOHelper.close(ctx.transport())` in `cleanup()` when it is.

components/camel-opensearch/src/main/java/org/apache/camel/component/opensearch/OpensearchProducer.java
```diff
@@ -188,6 +195,7 @@ public boolean process(Exchange exchange, AsyncCallback callback) {
             }
```
davsclaus, line 195: Nit: this blank line was not present before and is not needed — please remove to keep the diff clean.

davsclaus (changes requested):
Thanks for addressing the NPE blocker from the previous review, @renjth-81 — the lazy transport resolution in `process()` is the right approach. However, there are a couple of remaining issues that need fixing before this can be merged.

## Blocker: Missing Mockito dependency

The new test `CustomOpensearchClientPingIT.java` uses `Mockito.mock()`, `Mockito.when()`, and `RETURNS_SMART_NULLS`, but `components/camel-opensearch/pom.xml` does **not** declare `mockito-core` as a test dependency. No transitive path brings it in either (`camel-test-junit6`, `camel-test-infra-core`, and `camel-test-infra-opensearch` do not depend on Mockito). The module will fail to compile.

Please add to `pom.xml`:
```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <version>${mockito-version}</version>
    <scope>test</scope>
</dependency>
```

## Bug: Transport closed on user-provided client when `disconnect=true`

In `OpensearchProducer.cleanup()`, `IOHelper.close(ctx.transport())` is called when `disconnect=true`. For the default path this is safe — a fresh `RestClientTransport` is created per `process()` call. But when a custom `OpenSearchClient` is provided, `ctx.transport()` is `openSearchClient._transport()` — the user's own transport. Closing it will corrupt the user's client for future use.

The test exercises this path (`disconnect=true` + custom client), but since the mock's `close()` is a no-op, the bug is masked.

Suggested fix: skip closing the transport and client in `cleanup()` when a custom `OpenSearchClient` was provided.

## Minor: `doStart()` logs spurious warning with custom client

`doStart()` unconditionally calls `startClient()` when `disconnect=false`. When a user provides only an `OpenSearchClient` (no `hostAddresses`), `startClient()` logs: *"Incorrect ip address and port parameters settings for OpenSearch cluster"*. Consider skipping `startClient()` when a custom `OpenSearchClient` is set.

## Minor: Commit hygiene

- Per project conventions, commits should be prefixed with `CAMEL-23338:`.
- There's a merge commit and duplicate commit messages — please squash/clean up before merge.
- A spurious blank line was added in `OpensearchProducer.java` between `Integer from = message.getHeader(...)` and `if (from == null)` — please remove.

---

Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

oscerd (commented):
Re-review after the latest push — good progress, the Round 1 items are addressed:

- The transport is now resolved lazily in `process()` (not in the constructor), so the default path no longer NPEs, and `OpensearchPingIT` / `BulkIT` pass.
- `CustomOpensearchClientPingIT` proves a custom `OpenSearchClient` is actually used.
- The title now carries the `CAMEL-23338:` prefix.

Remaining items (matching @davsclaus's review):

- **`disconnect=true` + a custom client closes the caller's transport.** When `openSearchClient` is provided, `process()` sets `transport = openSearchClient._transport()` and passes it into the `ActionContext`; `cleanup()` then does `IOHelper.close(ctx.transport())` whenever `configuration.isDisconnect()` is true (`OpensearchProducer.java:437-438`). So with `disconnect=true` and a user-supplied client, the component closes the caller's transport after the first exchange, corrupting their client. The custom-client path should not close a transport it does not own.
- Commit hygiene: the individual commits are not `CAMEL-23338:`-prefixed and there is a merge commit + duplicate messages — a squash at merge would tidy this up.

On the "missing Mockito dependency" point: `mockito-core` is pulled in transitively via `camel-test-junit6`, which is why the module compiles and CI is green — an explicit declaration is optional rather than required.

Thanks for iterating on this!

renjth-81:
@davsclaus  @oscerd Please could you review the changes

components/camel-opensearch/src/main/java/org/apache/camel/component/opensearch/OpensearchProducer.java
```diff
@@ -91,11 +91,19 @@ class OpensearchProducer extends DefaultAsyncProducer {
...
+    private final OpenSearchClient openSearchClient;
+    private boolean isCustomClient;
 
     public OpensearchProducer(OpensearchEndpoint endpoint, OpensearchConfiguration configuration) {
         super(endpoint);
         this.configuration = configuration;
         this.client = endpoint.getClient();
+        this.openSearchClient = endpoint.getOpenSearchClient();
+        if(this.openSearchClient != null) {
+            isCustomClient = true;
```
davsclaus: The `isCustomClient` field is always equivalent to `openSearchClient != null` — both are set once in the constructor and never modified. This field adds unnecessary complexity.

The if/else block also has formatting issues (missing space after `if`, missing spaces around `else`).

Suggestion: remove the `isCustomClient` field entirely and use `openSearchClient != null` directly at the 3 call sites.

```suggestion
        this.openSearchClient = endpoint.getOpenSearchClient();
```

components/camel-opensearch/src/main/java/org/apache/camel/component/opensearch/OpensearchProducer.java
```diff
@@ -154,12 +162,17 @@ private OpensearchOperation resolveOperation(Exchange exchange) {
     @Override
     public boolean process(Exchange exchange, AsyncCallback callback) {
         try {
-            if (configuration.isDisconnect() && client == null) {
-                startClient();
+            OpenSearchTransport transport;
+            if (openSearchClient == null) {
+                if (configuration.isDisconnect() && client == null && !isCustomClient) {
```
davsclaus: Inside this `if (openSearchClient == null)` branch, the `&& !isCustomClient` check is always true (since `isCustomClient` is only true when `openSearchClient != null`). This is redundant and can be simplified to:

```suggestion
                if (configuration.isDisconnect() && client == null) {
```

components/camel-opensearch/pom.xml
```diff
@@ -90,6 +90,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>org.mockito</groupId>
+          <artifactId>mockito-core</artifactId>
+          <version>${mockito-version}</version>
```
davsclaus: Minor: the child elements use 2-space indentation here, but the rest of the POM uses 4-space indentation for dependency children. Running `mvn formatter:format` should fix this automatically.

```suggestion
          <groupId>org.mockito</groupId>
            <artifactId>mockito-core</artifactId>
            <version>${mockito-version}</version>
            <scope>test</scope>
```

Actually, please just run `mvn formatter:format` on this module to align with the project style.

davsclaus (changes requested):

## Review Summary

Thank you for this contribution — the ability to inject a custom `OpenSearchClient` is a valuable addition, especially for AWS OpenSearch Service users who need custom transport options. The overall approach is sound. A few items need attention before this can be merged.

> **Note:** This review covers project rules and conventions. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

### Documentation

The PR adds a new user-visible endpoint option (`openSearchClient`) but does not include a documentation update. Per project guidelines, any user-visible change must be documented. Please update the opensearch component docs page to describe the new option and when to use it (e.g., AWS OpenSearch Service with custom transport/credentials).

### Precedence Behavior

When both `client` (RestClient) and `openSearchClient` (OpenSearchClient) are configured — possible via autowiring since both are marked `autowired = true` — the `openSearchClient` silently takes precedence. Consider documenting this precedence in the option description, or logging a warning when both are set.

### Test Coverage

The integration test only covers the `Ping` operation with disconnect=true/false. Other operations (Index, Search, Bulk, etc.) are not tested with the custom client path. Consider adding at least one test for an operation that exercises the data path (e.g., Index or Search).

oscerd (commented):
Re-reviewing after your squash + latest changes — the items from my earlier note are all resolved: the `disconnect=true` + custom-client case now guards the close (`!isCustomClient`), the spurious "incorrect ip/port" warning is skipped for a custom client, and the history is squashed into a single `CAMEL-23338:` commit. Nice.

The one thing that will still turn CI red is the formatting gate — `if(this.openSearchClient != null)` / `}else{` need spaces (`mvn formatter:format impsort:sort` fixes it). That, plus the rest of @davsclaus's latest list, are the remaining items. Thanks for iterating on this!

gnodet (changes requested):

## Review: CAMEL-23338 — Add OpenSearchClient as parameter

The core approach is sound — lazy transport resolution in `process()`, skipping client lifecycle for custom client. However, several items need attention before merging, and prior review feedback from @davsclaus still needs to be addressed:

### 1. BLOCKER: Formatting violations fail CI

In `OpensearchProducer.java`, the constructor has `if(this.openSearchClient != null)` and `}else{` — missing spaces after `if` and around `else`. Please run:
```bash
mvn formatter:format impsort:sort
```

### 2. BLOCKER: Redundant `isCustomClient` field

As @davsclaus already noted, the `isCustomClient` boolean field is always equivalent to `openSearchClient != null` — both are set once in the constructor and never mutated. This adds unnecessary state. On line 167, the check `if (configuration.isDisconnect() && client == null && !isCustomClient)` is inside a branch that already established `openSearchClient == null`, making `!isCustomClient` always true — dead code.

Please remove the field and use `openSearchClient != null` directly.

### 3. Missing component documentation update

No `.adoc` files were modified. The new `openSearchClient` endpoint option should be documented in `components/camel-opensearch/src/main/docs/opensearch-component.adoc` with its use case (e.g., AWS OpenSearch Service with custom transport/credentials).

### 4. POM indentation mismatch

The added `mockito-core` dependency in `pom.xml` uses 2-space indentation while the rest of the POM uses 4-space. Also, `mockito-core` is already available transitively via `camel-test-junit6` — the explicit dependency may be redundant.

### 5. Precedence ambiguity when both clients are set

Both `RestClient client` and `OpenSearchClient openSearchClient` are `autowired = true`. If a user's registry contains both, `openSearchClient` silently takes precedence. This should be documented in the `@UriParam` description (e.g., "When set, this takes precedence over the RestClient-based client option").

Thanks for this contribution — the feature is useful and the approach is correct. Please address the formatting, the `isCustomClient` removal, and the documentation, and this should be ready.

renjth-81:
@davsclaus @gnodet @oscerd please could you review

components/camel-opensearch/src/main/docs/opensearch-component.adoc
```diff
@@ -406,3 +406,29 @@ MsearchRequest.Builder builder = new MsearchRequest.Builder().index("twitter").s
...
+
+getContext().getRegistry().bind("customClient", customClient);
+
+from("direct:deleteIndex")
+  .to("opensearch://foo?openSearchClient=#customClient");
+
+DeleteIndexRequest.Builder builder = new DeleteIndexRequest.Builder().index("twitter");
+Boolean status = template().requestBody("direct:deleteIndex", builder, Boolean.class);
+----
+
```
davsclaus: **Missing closing `====`:** This `[tabs]` block (opened at line 412) is never closed. Every other tabs block in this file has a matching `====`. Without it, AsciiDoc rendering breaks from this point onward.

Please add `====` after this line:
```suggestion
----
====
```

components/camel-opensearch/src/main/docs/opensearch-component.adoc
```diff
@@ -406,3 +406,29 @@ MsearchRequest.Builder builder = new MsearchRequest.Builder().index("twitter").s
...
+Java::
++
+[source,java]
+----
+SdkHttpClient httpClient = ApacheHttpClient.builder().build();
+final ObjectMapper mapper = new ObjectMapper();
+mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+OpenSearchTransport transport = new AwsSdk2Transport(httpClient, host, Region.of(region),AwsSdk2TransportOptions.builder().setCredentials(buildCredentialsProvider()).setMapper(new JacksonJsonpMapper(mapper)).build());
+OpenSearchClient customClient = new OpenSearchClient(transport);
+
```
davsclaus, line 430: This line is ~200 chars and will overflow in the rendered docs. Please break it across multiple lines for readability, e.g.:
```java
OpenSearchTransport transport = new AwsSdk2Transport(
    httpClient, host, Region.of(region),
    AwsSdk2TransportOptions.builder()
        .setCredentials(buildCredentialsProvider())
        .setMapper(new JacksonJsonpMapper(mapper))
        .build());
```

davsclaus (changes requested):

## Review: CAMEL-23338 — Add OpenSearchClient as parameter

All prior blocking issues from @oscerd, @davsclaus, and @gnodet have been addressed in the latest push — the lazy transport resolution in `process()`, the `disconnect=true` guard in `cleanup()`, the `doStart()` skip, and the `isCustomClient` removal are all correctly implemented. Nice work iterating on this.

One documentation issue remains that will break the rendered page.

### Documentation: Missing closing `====` for tabs block

The new "Custom OpenSearchClient Example" section opens a `[tabs]` / `====` block but never closes it. Every other tabs block in this file has a matching `====` closure (lines 172, 226, 291, 355, 395). Without it, AsciiDoc rendering will break from this point downward on the published page.

Fix: add `====` after the closing `----`.

### Documentation: Long line in example

The `AwsSdk2Transport` constructor call is ~200 chars and will overflow in the rendered docs. Please break it into multiple lines for readability.

### Documentation: Example mixes route DSL with test invocation

The example shows `getContext().getRegistry().bind(...)`, route `from()/to()`, and `template().requestBody(...)` together. This is valid test-style code, but for a user-facing doc example, it would be clearer to separate the route definition from the programmatic invocation — similar to how other examples in this file show just the route DSL or just the request body invocation.

---

Note: this review covers project rules and conventions. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

renjth-81:
@davsclaus Please could you review

davsclaus:
I already did recently, did you look at the comments.

components/camel-opensearch/src/main/docs/opensearch-component.adoc
renjth-81: fixed

renjth-81 (commented):
added missing === at the end ,  removed the java test invocation in the example.

davsclaus (commented):

## Review: CAMEL-23338 — Add OpenSearchClient as parameter

Thank you for your persistence through multiple review rounds — the core design is solid and this is a valuable addition for AWS OpenSearch Service users.

### Remaining items

**1. Missing trailing newline in `opensearch-component.adoc`**
The file ends without a final newline (`\ No newline at end of file`). Please add one.

**2. Missing explicit `mockito-core` test dependency**
Both new test files import `org.mockito.Mockito`. While it works transitively today (via `camel-test-infra-core`), it should be declared explicitly as a `<scope>test</scope>` dependency in `components/camel-opensearch/pom.xml` — if the transitive path changes, these tests break silently.

**3. Pre-existing: `ObjectMapper` per `process()` call** (non-blocking)
In the `openSearchClient == null` branch, a new `ObjectMapper` + `JacksonJsonpMapper` is created on every `process()` invocation. This is pre-existing, not introduced by your PR — just noting it as a follow-up opportunity.

### What looks good

- Lazy transport resolution in `process()` avoids the NPE from the earlier iteration
- Lifecycle guards for `disconnect`, `doStart()`, and `cleanup()` are correct
- Generated files are properly regenerated
- Tests cover Index, DeleteIndex, Exists, and Ping with both `disconnect=true/false`
- Documentation example demonstrates the AWS SDK transport use case

components/camel-opensearch/src/main/docs/opensearch-component.adoc
```diff
@@ -406,3 +406,32 @@ MsearchRequest.Builder builder = new MsearchRequest.Builder().index("twitter").s
...
+        .build());
+OpenSearchClient customClient = new OpenSearchClient(transport);
+
+getContext().getRegistry().bind("customClient", customClient);
+
+from("direct:deleteIndex")
+  .to("opensearch://foo?operation=DeleteIndex&indexName=twitter&openSearchClient=#customClient");
+
+----
+====
```
davsclaus, line 437: Missing trailing newline at end of file. Please add a blank line after `====` so the file ends with a newline character.

components/camel-opensearch/src/test/java/org/apache/camel/component/opensearch/integration/CustomOpensearchClientIndexIT.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
```
davsclaus, line 7: `mockito-core` is not declared as an explicit test dependency in `components/camel-opensearch/pom.xml`. It works today via transitive dependency from `camel-test-infra-core`, but should be declared directly to avoid silent breakage if that transitive path changes.

Please add to `pom.xml`:
```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <scope>test</scope>
</dependency>
```

davsclaus (approved):

components/camel-opensearch/src/main/java/org/apache/camel/component/opensearch/OpensearchProducer.java
```diff
@@ -451,7 +458,7 @@ private void cleanup(ActionContext ctx) {
     @Override
     protected void doStart() throws Exception {
         super.doStart();
-        if (!configuration.isDisconnect()) {
+        if (!configuration.isDisconnect() && openSearchClient == null) {
```
davsclaus, line 461: This guard correctly skips `startClient()` when a custom `OpenSearchClient` is set — good.

However, `doStop()` (line 550 in the current file) is missing the same guard. It unconditionally closes `client` if non-null, which is inconsistent with the lifecycle pattern established here, in `process()`, and in `cleanup()`.

```suggestion
        if (!configuration.isDisconnect() && openSearchClient == null) {
```

And `doStop()` should similarly be guarded:
```java
if (client != null && openSearchClient == null) {
```

components/camel-opensearch/src/main/docs/opensearch-component.adoc
davsclaus, line 437: Missing trailing newline at end of file.

davsclaus (changes requested):
Thanks for the contribution — useful feature for users who need custom transports (e.g. AWS SDK transport). One blocking issue and a couple of minor items.

**Blocking:**

`doStop()` (`OpensearchProducer.java:550`) is not guarded with `openSearchClient == null`, unlike `doStart()`, `process()`, and `cleanup()` which all correctly skip RestClient lifecycle management when a custom `OpenSearchClient` is provided. When both a `RestClient` and an `OpenSearchClient` are in the registry, `doStop()` would unconditionally close the `RestClient` even though the producer never used it.

**Non-blocking observations:**

- `opensearch-component.adoc` ends without a trailing newline (`\ No newline at end of file`) — most editors and git tooling expect one.
- New test code uses JUnit assertions (`assertEquals`, `assertTrue`, etc.) — project conventions prefer AssertJ (`assertThat(...)`) for new tests.
- Since `openSearchClient` is `autowired = true`, users who already have an `OpenSearchClient` bean in their registry for other purposes would see it silently auto-wired into the component. Consider whether an upgrade guide note is warranted.
- CAMEL-23338 is currently unassigned in JIRA — per project contribution rules, the contributor should assign it before working on it.

This review covers project rules and conventions. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

davsclaus:
done via another PR

PR #24329, opened 2026-06-30
CAMEL-23857: TUI allow to choose runtime when running from folder or examples

davsclaus:
## Summary

- TUI run options form now lets users choose runtime (Camel Main, Spring Boot, Quarkus) when running from folder or examples
- Export passes `exportBaseDir` so Spring Boot and Quarkus exports work correctly from any directory
- Profile (dev/prod) toggle added to run options form; exported `application.properties` includes `camel.main.profile`
- Move `profile` from `MainConfigurationProperties` to `DefaultConfigurationProperties` so Spring Boot config generator picks it up and all runtimes can use it
- `ProfileConfigurer.configureCommon()` now sets `tracingStandby` so SB/Quarkus also get dev-mode tracing support
- TUI Browse Files supports navigating subdirectories (enter/esc to navigate in/out, `..` to go up)
- TUI restart disabled for Spring Boot and Quarkus until reliable; error notifications persist 15 seconds

**Note:** Companion Spring Boot changes (calling `ProfileConfigurer.configureCommon`, dev profile support in tracer/error-registry/security auto-configurations) will be in a separate camel-spring-boot PR.

## Test plan

- [ ] Run TUI, select a folder-based example, pick Spring Boot runtime, verify it exports and runs
- [ ] Run TUI, select a folder-based example, pick Quarkus runtime, verify it exports and runs  
- [ ] Verify dev profile toggle works: set profile=dev, check that tracing/health/dev-console are enabled
- [ ] Browse Files on a Maven project: navigate into `src/main/java/...` subdirectories and back out
- [ ] Verify restart works for Camel Main and shows "not supported" notification for SB/Quarkus


oscerd (commented):
Thanks @davsclaus — nice feature. I went through the logic and it checks out: the `FilesBrowser` `..` navigation is null-safe (parent is only dereferenced when `dir != rootDir`), the `CamelMonitor` restart guard matches the platform strings emitted by `LocalCliConnector`, `RunOptionsForm.buildArgs()` maps the runtime values correctly, and the regenerated `camel-main-configuration-metadata.json` (core + catalog copy) is consistent. Moving `setTracingStandby(true)` into `configureCommon` is safe since `configureMain` calls `configureCommon` at its end. Commits follow the `CAMEL-23857:` convention with attribution. Just a couple of non-blocking notes:

**1. Minor binary-compat edge on `withProfile`.** This PR moves `getProfile`/`setProfile`/`withProfile` from `MainConfigurationProperties` up to `DefaultConfigurationProperties`. `getProfile`/`setProfile` are descriptor-identical, so they remain binary-safe via inheritance. But `withProfile` changes from a concrete return type (`MainConfigurationProperties`, as shipped in 4.20.0) to the generic `T` (erased to `DefaultConfigurationProperties`). Source compatibility is fully preserved and this is consistent with every other `withXxx` builder — but code compiled against 4.20.0 calling `mainConfig.withProfile("dev")` could hit a `NoSuchMethodError` at runtime if not recompiled, since the erased descriptor no longer matches and there's no covariant bridge (`MainConfigurationProperties` no longer declares the method). `profile` was only introduced in 4.20.0 so the blast radius is small; you might still want to keep a covariant `withProfile` override in `MainConfigurationProperties`, or call it out in the upgrade guide.

**2. Tests / docs.** No tests in the changeset — `FilesBrowser` sorting/`..` logic and `RunOptionsForm.buildArgs()` are unit-testable in principle even though the interactive rendering isn't. The new user-visible TUI features (runtime chooser, dev/prod toggle, Browse Files subdirectory navigation) aren't documented in the camel-jbang TUI page. Both minor for TUI tooling — just flagging.

CI: the matrix builds were still running at review time, so worth confirming green before merge.

Otherwise looks good.

---

PR #24348, opened 2026-07-01
CAMEL-23842: camel-pqc - use authenticated encryption (AEAD) in PQCDataFormat instead of ECB

oscerd:
## CAMEL-23842

The `pqc` data format encrypted payloads with a bare cipher name (e.g. `AES`), which resolves to `AES/ECB/PKCS5Padding`. ECB leaks plaintext block structure and there is no IV or integrity protection, so the ciphertext is malleable and the DEM layer of the KEM-DEM construction is not IND-CCA2 secure.

### Changes
- The symmetric (data-encapsulation) layer now uses **authenticated encryption**: 128-bit block ciphers use **GCM**, and the ChaCha20 stream cipher uses **ChaCha20-Poly1305**.
- A random 12-byte nonce is written to the wire format; decryption verifies the authentication tag via `Cipher.doFinal` — **not** `CipherInputStream`, which can silently swallow an AEAD tag failure and return truncated plaintext. Tampered or corrupted messages are now rejected.
- Only AEAD-capable symmetric algorithms are accepted: `AES`, `ARIA`, `CAMELLIA`, `CAST6`, `DSTU7624`, `GOST3412-2015`, `SEED`, `SM4` (GCM) and `CHACHA7539` (ChaCha20-Poly1305). The non-AEAD legacy/stream ciphers (`RC2`, `RC5`, `CAST5`, `GOST28147`, `DESEDE`, `GRAIN128`, `HC128`, `HC256`, `SALSA20`) are rejected at startup.
- A malformed-input guard bounds the encapsulation length read from untrusted data.
- **Removed the now-dead `bufferSize` option** — it only configured the previous streaming implementation, which AEAD replaced (payloads are processed in memory so the tag can be verified before releasing plaintext).
- Updated the DSL model enum and regenerated all metadata (module + catalog dataformat JSON, catalog model JSON, XSD and YAML DSL schemas, DSL writers/parsers), refreshed the component docs, and added the 4.22 upgrade guide.

### Wire format
```
[4 bytes: encapsulation length] [N: encapsulation] [12 bytes: nonce] [M: ciphertext + auth tag]
```

### ⚠️ Breaking change
- Data encrypted by Camel 4.21 or earlier (unauthenticated ECB, no nonce) **cannot be decrypted** by this version — it must be re-encrypted.
- The non-AEAD symmetric algorithms above are no longer accepted by the data format.
- The `bufferSize` data-format option was removed.

All documented in the new `camel-4x-upgrade-guide-4_22` entry.

### Tests
Added `PQCDataFormatAeadTest` covering AES-GCM, ARIA-GCM and ChaCha20-Poly1305 round-trips, tamper detection, and rejection of non-AEAD algorithms (at start and at marshal). Existing dataformat tests are unchanged and pass; the full-reactor build is green.

---


davsclaus (commented):
Good security fix — replacing ECB with GCM/ChaCha20-Poly1305 AEAD is the right call. The core cryptographic implementation looks correct: 12-byte random nonce per NIST SP 800-38D, 128-bit GCM tag (no truncation), `Cipher.doFinal()` instead of `CipherInputStream` (which silently swallows AEAD tag failures), and fail-fast validation in `doStart()` with defense-in-depth in `marshal()`.

A few minor observations for a possible follow-up:

1. **`bufferSize` is now dead code** — the two lines that used it are removed, but the field, getter/setter, and generated metadata remain with the now-misleading description _"The size of the buffer used for streaming encryption/decryption."_ Consider deprecating or removing it (with an upgrade-guide note).

2. **`PQCSymmetricAlgorithms` enum retains rejected values** — `RC2`, `RC5`, `CAST5`, `GOST28147`, `DESEDE`, `GRAIN128`, `HC128`, `HC256`, `SALSA20` are still in the Java enum but rejected at runtime by `checkAeadSupported()`. Deprecating or removing these entries would keep the enum consistent with the actual runtime behavior.

3. **Upgrade guide could mention the memory behavior change** — the guide covers the wire format change and algorithm restrictions, but omits that payloads are now loaded entirely into memory (`byte[]`) rather than streamed. The component docs mention it, but users reading only the upgrade guide might miss this for large-payload use cases.

None of these are blocking — the PR is a clear improvement.

apupier:
conflict to resolve

davsclaus (approved):

apupier:
some parts to regen:

```

diff --git a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
index a3a4ab3e..79c888ab 100644
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-spring.xsd
@@ -9452,7 +9452,9 @@ The Post-Quantum KEM algorithm to use for key encapsulation. Default value: MLKE
           <xs:annotation>
 <![CDATA[
-The symmetric encryption algorithm to use with the shared secret. Default value: AES
+The symmetric encryption algorithm to use with the shared secret. Only algorithms that support authenticated encryption
+(AEAD) are allowed: AES, ARIA, CAMELLIA, CAST6, DSTU7624, GOST3412_2015, SEED and SM4 are encrypted with GCM, and
+CHACHA7539 with ChaCha20-Poly1305. Default value: AES
 ]]>
             </xs:documentation>
           </xs:annotation>
diff --git a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd
index ea145f08..c1bc3be1 100644
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/schemas/camel-xml-io.xsd
@@ -8574,7 +8574,9 @@ The Post-Quantum KEM algorithm to use for key encapsulation. Default value: MLKE
           <xs:annotation>
 <![CDATA[
-The symmetric encryption algorithm to use with the shared secret. Default value: AES
+The symmetric encryption algorithm to use with the shared secret. Only algorithms that support authenticated encryption
+(AEAD) are allowed: AES, ARIA, CAMELLIA, CAST6, DSTU7624, GOST3412_2015, SEED and SM4 are encrypted with GCM, and
+CHACHA7539 with ChaCha20-Poly1305. Default value: AES
 ]]>
             </xs:documentation>
           </xs:annotation>

```

oscerd:
Thanks for the review @davsclaus, and @apupier for flagging the conflict.

Pushed 60093de addressing the feedback:

**1. `bufferSize` dead code → removed.** Since authenticated encryption replaced the streaming implementation, the option no longer had any effect, so I removed it entirely — the field/getter/setter on the component and the `core-model` model (+ `Builder`), the hand-written reifier, and all regenerated metadata (configurer, catalog, XSD/YAML schemas, DSL writers/parsers). Documented in the component docs and the 4.22 upgrade guide.

**2. `PQCSymmetricAlgorithms` retaining the "rejected" values → kept intentionally.** That enum is shared with the PQC *component* (the KEM operations in `PQCProducer`), not only the dataformat. `RC2`, `RC5`, `CAST5`, `GOST28147`, `DESEDE`, etc. are still valid there — they are used as key-derivation labels for `KEMGenerateSpec`/`KEMExtractSpec` and do not require AEAD (there are dedicated round-trip tests, e.g. `PQCMLKEMGenerateEncapsulationCryptoRoundTripDESedeTest`, `...GOST28147Test`, `...RC5Test`). Removing the entries would break the component. The dataformat instead restricts to the AEAD subset at runtime via `checkAeadSupported()`, and the dataformat's own DSL enum now lists only the 9 AEAD-capable algorithms.

**3. Upgrade-guide memory-behavior note → added.** The 4.22 guide now mentions that payloads are processed in memory (the AEAD tag is verified before releasing plaintext) rather than streamed, and to enable stream caching for very large payloads.

**Conflict** (@apupier): resolved by rebasing onto `main` (the concurrent 4_22 upgrade-guide addition for `camel-fory`); both entries are now in the guide and the PR is mergeable. The full-reactor build passes, and the stale `camel-spring.xsd` / `camel-xml-io.xsd` that were failing CI are regenerated.

davsclaus (approved):

PR #24349, opened 2026-07-01
Fix ConcurrentModificationException in TUI MemoryTab

gnodet:
## Summary

Fixes `ConcurrentModificationException` in TUI `MemoryTab.renderSparkline()` caused by a race condition between the background data-refresh thread and the UI render thread.

**Root cause**: `MetricsCollector` stores history data as `LinkedList` values inside `ConcurrentHashMap` maps. The background refresh thread mutates these lists in place (`.add()`, `.remove(0)`), while the UI render thread iterates them via `.stream()` and `.get(i)`. `ConcurrentHashMap` protects the map itself but not the individual `LinkedList` values.

**Fix**: Use copy-on-write semantics in `MetricsCollector` — instead of mutating history lists in place, create a new copy, modify it, and atomically swap it into the map via `ConcurrentHashMap.compute()`. Readers always iterate an unmodified list snapshot. This fixes **all** affected tabs at once (MemoryTab, OverviewTab, EndpointsTab, CircuitBreakerTab).

- Introduce `addToHistory()` helper that encapsulates the copy-on-write pattern
- Refactor all 8 history update sites in `MetricsCollector` to use it
- Add concurrency regression tests exercising concurrent read/write

**Stack trace that was observed:**
```
java.util.ConcurrentModificationException
  at java.base/java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1254)
  at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
  ...
  at org.apache.camel.dsl.jbang.core.commands.tui.MemoryTab.renderSparkline(MemoryTab.java:244)
  at org.apache.camel.dsl.jbang.core.commands.tui.MemoryTab.render(MemoryTab.java:116)
```

## Test plan

- [x] New `MetricsCollectorConcurrencyTest` with 2 tests exercising concurrent read/write on heap and throughput history maps
- [x] All 469 existing TUI tests pass
- [ ] Manual verification: run `camel monitor` with a live integration and navigate to the Memory tab

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/MetricsCollector.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/MetricsCollector.java
@@ -176,16 +176,8 @@ class MetricsCollector {
             Long lastTime = previousExchangesTime.get(pid);
             if (lastTime == null || now - lastTime >= 1000) {
                 previousExchangesTime.put(pid, now);
-                LinkedList<Long> hist = throughputHistory.computeIfAbsent(pid, k -> new LinkedList<>());
-                hist.add(tp);
-                while (hist.size() > MAX_SPARKLINE_POINTS) {
-                    hist.remove(0);
-                }
-                LinkedList<Long> fhist = failedHistory.computeIfAbsent(pid, k -> new LinkedList<>());
-                fhist.add(fp);
-                while (fhist.size() > MAX_SPARKLINE_POINTS) {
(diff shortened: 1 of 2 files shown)


davsclaus (approved):
Clean, correct fix for a real race condition. The copy-on-write approach via `ConcurrentHashMap.compute()` is idiomatic and covers all history maps exposed to the UI render thread.

1. The concurrency tests validate the COW pattern directly on the map rather than exercising the production methods (`updateHeapHistory`, etc.) due to their time throttles. This is acceptable — it proves the pattern prevents CME.

2. No Jira ticket referenced (branch is `fix/tui-concurrent-modification`). Minor guideline deviation; the fix is self-contained.

PR #24356, opened 2026-07-01
CAMEL-23865: Use EWMA smoothing for throughput and fix TUI sparkline scaling

davsclaus:
## Summary

Switch the throughput metric in LoadThroughput from raw instantaneous rate to EWMA (exponentially weighted moving average) smoothing with a 1-minute decay window. This produces stable, converging throughput readings instead of oscillating 0/spike values — consistent with how LoadTriplet already computes CPU load averages.

Also fixes the TUI sparkline charts to properly display sub-1.0 msg/s rates by scaling throughput values by 100x internally.

## Changes

**camel-management (core):**
- LoadThroughput: EWMA smoothing with Math.exp(-1/60) decay, negative delta clamping, reset() stops the StopWatch
- LoadThroughputTest: Awaitility-based tests verifying EWMA convergence, smoothing bounds, and reset behavior

**camel-jbang-plugin-tui:**
- MetricsCollector: reads EWMA throughput from status JSON, stores scaled by THROUGHPUT_SCALE=100; shared niceMax() and formatThroughput() utility methods
- OverviewTab / EndpointsTab: sparkline charts use scaled values with 1-2-5 Y-axis snapping; delegate to MetricsCollector for formatting

**docs:**
- Upgrade guide entry for throughput MBean semantics change (instantaneous to EWMA)

### DIFF
--- a/core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
+++ b/core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
@@ -48,7 +48,7 @@ public final class LoadThroughput {
             if (time > 0) {
                 long delta = currentReading - last;
                 // instantaneous rate in exchanges/second for this interval
-                double instantRate = (1000d / time) * delta;
+                double instantRate = Math.max(0, (1000d / time) * delta);
                 // apply EWMA smoothing
                 thp = instantRate + EXP_1 * (thp - instantRate);
             }
@@ -61,6 +61,7 @@ public final class LoadThroughput {
     }
 
     public void reset() {
+        watch.stop();
         last = 0;
         thp = 0;
     }
--- a/core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
+++ b/core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
@@ -19,18 +19,25 @@ package org.apache.camel.management.mbean;
 import org.apache.camel.util.StopWatch;
 
 /**
- * Holds the load throughput messages/second
+ * Holds the throughput (messages/second) using EWMA (exponentially weighted moving average) smoothing, modeled after
+ * Unix load averages (same approach as {@link LoadTriplet}).
+ *
+ * The instantaneous rate from each 1-second sampling interval is smoothed with a 1-minute decay window so that the
+ * reported value converges to the true average rate instead of oscillating between 0 and spike values.
  */
 public final class LoadThroughput {
+
+    // EWMA exponent for a 1-minute decay window, sampled every 1 second
+    private static final double EXP_1 = Math.exp(-1.0 / 60.0);
 
     private final StopWatch watch = new StopWatch(false);
     private long last;
     private double thp;
 
     /**
-     * Update the load statistics
+     * Update the throughput statistics
      *
-     * @param currentReading the current reading
+     * @param currentReading the current cumulative exchange count
      */
     public void update(long currentReading) {
         if (!watch.isStarted()) {
@@ -40,14 +47,10 @@ public final class LoadThroughput {
             long time = watch.takenAndRestart();
             if (time > 0) {
                 long delta = currentReading - last;
-                if (delta > 0) {
-                    // need to calculate with fractions
-                    thp = (1000d / time) * delta;
-                } else {
-                    thp = 0;
-                }
-            } else {
-                thp = 0;
+                // instantaneous rate in exchanges/second for this interval
+                double instantRate = (1000d / time) * delta;
+                // apply EWMA smoothing
+                thp = instantRate + EXP_1 * (thp - instantRate);
             }
         }
         last = currentReading;
--- a/core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
+++ b/core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
@@ -16,15 +16,17 @@
  */
 package org.apache.camel.management;
 
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
 import org.apache.camel.management.mbean.LoadThroughput;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
 
+import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@DisabledOnOs(OS.AIX)
 public class LoadThroughputTest {
 
     @Test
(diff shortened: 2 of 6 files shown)


core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
```diff
@@ -40,14 +47,10 @@ public void update(long currentReading) {
...
-                if (delta > 0) {
-                    // need to calculate with fractions
-                    thp = (1000d / time) * delta;
-                } else {
-                    thp = 0;
-                }
-            } else {
-                thp = 0;
+                // instantaneous rate in exchanges/second for this interval
+                double instantRate = (1000d / time) * delta;
```
gnodet: The old code guarded `if (delta > 0)` and set `thp = 0` for non-positive deltas. With EWMA, a negative `instantRate` (from `currentReading < last`, e.g. after a route restart that resets the exchange counter) bleeds into `thp` and takes ~60 seconds to decay back toward zero.

Consider clamping:
```suggestion
                double instantRate = Math.max(0, (1000d / time) * delta);
```

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @Test
+    public void testConvergesToSteadyRate() throws Exception {
+        LoadThroughput t = new LoadThroughput();
+
+        // simulate 1 exchange per second for 120 seconds (well past 1-minute EWMA window)
+        long total = 0;
+        t.update(total);
+        Thread.sleep(10);
+        for (int i = 0; i < 120; i++) {
+            total++;
```
gnodet: The test introduces multiple `Thread.sleep()` calls. Per project conventions:

> *"New test code MUST NOT introduce `Thread.sleep()` calls. Use the Awaitility library instead."*

Since `LoadThroughput` relies on `StopWatch` internally, it requires real elapsed time. A better long-term approach would be an injectable time source for fully deterministic tests. At minimum, the sleep-based timing should be replaced with Awaitility polling assertions.

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        // should converge close to the steady rate
+        assertTrue(t.getThroughput() > 0, "Throughput should be positive for steady input");
+    }
+
+    @Test
+    public void testSmoothing() throws Exception {
+        LoadThroughput t = new LoadThroughput();
+
```
gnodet, line 60: This assertion (`> 0`) would also pass with the old raw throughput calculation. Consider tightening it to verify that EWMA smoothing is actually converging toward the expected steady-state rate (e.g., within a reasonable tolerance of the true rate).

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.camel.management.mbean.LoadThroughput;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@DisabledOnOs(OS.AIX)
+public class LoadThroughputTest {
```
gnodet, line 30: This test is pure Java math + `Thread.sleep` — no JMX interaction. The `@DisabledOnOs(OS.AIX)` annotation may have been cargo-culted from other management tests that interact with platform MBeans. Is there a specific AIX reason to skip this test? If not, consider removing it.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/OverviewTab.java
```diff
@@ -637,6 +644,34 @@ private void renderInfoPanel(Frame frame, Rect area) {
         frame.renderWidget(Paragraph.builder().text(Text.from(lines)).build(), inner);
     }
 
+    /**
+     * Snap to a nice Y-axis ceiling using a 1-2-5 sequence (scaled by THROUGHPUT_SCALE). For example with scale=100:
+     * 0.20 → 1, 1.5 → 2, 3 → 5, 7 → 10, 15 → 20, 35 → 50, 80 → 100, etc.
+     */
+    private static long niceMax(long rawMax) {
+        long s = MetricsCollector.THROUGHPUT_SCALE;
```
gnodet: `niceMax` and `formatThroughput` are duplicated between `OverviewTab` and `EndpointsTab` with **different** implementations:

- **Here** (`OverviewTab`): hardcoded array of 10 steps up to `1000 * THROUGHPUT_SCALE`; falls back to `rawMax` (no rounding) beyond that.
- **`EndpointsTab`**: uses a `while(true)` loop with 1-2-5 step sequence that scales indefinitely.

The same data could get different Y-axis ceilings depending on which chart displays it. Consider extracting both methods to a shared utility (e.g., on `MetricsCollector` where `THROUGHPUT_SCALE` already lives) with a single correct implementation.

gnodet (changes requested):
Thanks for this improvement, Claus — the EWMA smoothing approach is sound and consistent with the existing `LoadTriplet` pattern, and the TUI sparkline scaling is a nice enhancement. The formula is correct and the overall direction is good.

A few items to address before merging (details in inline comments):

**Blocking:**
1. The new test introduces `Thread.sleep()` calls — project conventions require Awaitility instead.
2. Negative deltas are no longer guarded, which can produce persistent negative throughput via EWMA memory.

**Suggestions:**
3. `reset()` doesn't stop the watch — the first post-reset update can produce a spike that persists ~60s via EWMA.
4. `niceMax` and `formatThroughput` are duplicated between `OverviewTab` and `EndpointsTab` with divergent implementations.
5. The throughput JMX MBean attribute changed semantics (instantaneous → EWMA) — this should get an upgrade guide entry per project conventions.

This review covers project conventions and code correctness. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

core/camel-management/src/main/java/org/apache/camel/management/mbean/LoadThroughput.java
davsclaus: Done — added `Math.max(0, ...)` clamping on the instantaneous rate. Also added `watch.stop()` in `reset()` so the next `update()` re-enters the initialization branch cleanly, avoiding post-reset spikes.

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
davsclaus: Done — rewrote all tests using Awaitility's `pollInterval` + `untilAsserted` pattern. Since `LoadThroughput` relies on real elapsed time via `StopWatch`, the Awaitility polling drives the updates at 10ms intervals while the assertions verify convergence.

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
davsclaus: Done — tightened the convergence assertion to check throughput > 5.0 (verifying actual EWMA convergence toward the steady-state rate rather than just > 0). The smoothing test now also checks the value is bounded within a reasonable range.

core/camel-management/src/test/java/org/apache/camel/management/LoadThroughputTest.java
davsclaus: Good catch — removed. This is a pure Java math test with no JMX interaction, so the AIX exclusion was cargo-culted from other management tests.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/OverviewTab.java
davsclaus: Done — extracted both `niceMax` and `formatThroughput` to `MetricsCollector` as package-private static utility methods. Used the more robust `EndpointsTab` version (1-2-5 loop that scales indefinitely) as the single implementation. Both tabs now delegate to `MetricsCollector.niceMax()` and `MetricsCollector.formatThroughput()`.

oscerd (commented):
Thanks for the throughput smoothing work, Claus — the EWMA approach reads cleanly and the new `LoadThroughputTest` is a welcome addition.

I went back through @gnodet's earlier `CHANGES_REQUESTED` items, and they all look addressed as of `54b6cab`:

1. ✅ `Thread.sleep` → Awaitility in the test (`await().pollInterval(10, MILLISECONDS).untilAsserted(...)`).
2. ✅ Negative-delta guard: `double instantRate = Math.max(0, (1000d / time) * delta);` in `LoadThroughput`.
3. ✅ `reset()` now calls `watch.stop()` so the next `update()` re-initialises through the clean init branch.
4. ✅ `niceMax` / `formatThroughput` de-duplicated into `MetricsCollector` (no leftover copies in `EndpointsTab` / `OverviewTab`).
5. ✅ Upgrade-guide note added — correctly in `camel-4x-upgrade-guide-4_22.adoc`.

So the standing `CHANGES_REQUESTED` looks stale now — @gnodet, would you mind re-reviewing to clear it?

Two minor, non-blocking observations (both essentially unreachable in practice — take or leave):

- `MetricsCollector.formatThroughput` returns `String.valueOf(Math.round(v))` for `v >= 10` with no unit suffix, while the callers pad to a fixed width (`in:%-4s`, `out:%-4s`, and `%4s` on the sparkline axis). A rate ≥ 10000 msg/s renders 5 digits and overflows the 4-char column, nudging the alignment. Purely cosmetic.
- `MetricsCollector.niceMax` grows `multiplier *= 10` inside an unbounded `while (true)` loop. For a pathological `rawMax` near `Long.MAX_VALUE`, `multiplier` would overflow negative. Not reachable with the real ×100-scaled rates, but a defensive upper bound would make it obviously safe.

Nothing here blocks — the substantive review items are all resolved.

---

PR #24363, opened 2026-07-01
improving vertx websocket for end to end tracing in camel - POC

JinyuChen97:
This commit enables WebSocket message spans to be linked back to their originating HTTP upgrade request spans using OpenTelemetry span links.

Key changes:

1. VertxWebsocketHost: Capture HTTP upgrade span context using reflection
   - Avoids compile-time dependency on OpenTelemetry
   - Stores span context in RoutingContext during WebSocket upgrade

2. VertxWebsocketPeer: Store handshake span context
   - New field to hold the HTTP upgrade span context
   - Persists for the lifetime of the WebSocket connection

3. VertxWebsocketSpanDecorator: Custom span decorator for vertx-websocket
   - Collects span contexts from target WebSocket peers (Producer)
   - Supports multiple span links (sendToAll scenarios)
   - Uses reflection to avoid component dependencies

4. OpenTelemetryTracer: Create span links from collected contexts
   - Extract span contexts from exchange headers
   - Support both direct SpanContext objects and serialized format
   - Add links to spans during creation

5. VertxWebsocketConstants: Add HANDSHAKE_SPAN_CONTEXT_KEY constant

Benefits:
- Complete distributed tracing across HTTP and WebSocket boundaries
- Zero compile-time dependencies between components
- Supports both single and multiple span links

Related to: https://github.com/apache/camel-quarkus/issues/8280

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetryTracer.java
+++ b/components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetryTracer.java
@@ -17,12 +17,15 @@
 package org.apache.camel.opentelemetry2;
 
 import java.lang.management.ManagementFactory;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
@@ -180,6 +183,11 @@ public class OpenTelemetryTracer extends org.apache.camel.telemetry.Tracer {
     private class OpentelemetrySpanLifecycleManager implements SpanLifecycleManager {
 
         private final static String BAGGAGE_VAR_PREFIX = "OTEL_BAGGAGE_";
+        /**
+         * Exchange property name for storing a span context to create span links. This is used by components like
+         * vertx-websocket to link WebSocket message spans back to the original HTTP upgrade request span.
+         */
+        private final static String SPAN_LINK_CONTEXT_PROPERTY = "CamelVertxWebsocketHandshakeSpanContext";
 
         private final Tracer tracer;
         private final ContextPropagators contextPropagators;
@@ -193,6 +201,14 @@ public class OpenTelemetryTracer extends org.apache.camel.telemetry.Tracer {
         public Span create(String spanName, String spanKind, Span parent, SpanContextPropagationExtractor extractor) {
             SpanBuilder builder = tracer.spanBuilder(spanName);
             Baggage baggage = Baggage.current();
+
+            // Extract span link contexts (can be multiple for scenarios like sendToAll)
+            List<SpanContext> linkContexts = extractSpanLinkContexts(extractor);
+            for (SpanContext linkContext : linkContexts) {
+                if (linkContext != null && linkContext.isValid()) {
+                    builder = builder.addLink(linkContext);
+                }
+            }
 
             if (parent != null) {
                 OpenTelemetrySpanAdapter otelParentSpan = (OpenTelemetrySpanAdapter) parent;
@@ -229,6 +245,49 @@ public class OpenTelemetryTracer extends org.apache.camel.telemetry.Tracer {
             }
 
             return new OpenTelemetrySpanAdapter(builder.startSpan(), baggage);
+        }
+
+        /**
+         * Extracts span contexts from the exchange properties to create span links. This allows linking spans across
+         * asynchronous boundaries, such as WebSocket messages back to the HTTP upgrade request.
+         * <p>
+         * Supports multiple span links for scenarios like broadcasting to multiple WebSocket connections from different
+         * HTTP requests.
+         *
+         * @param  extractor the span context propagation extractor (usually the Exchange)
+         * @return           list of span contexts to link to (empty if none present)
+         */
+        private List<SpanContext> extractSpanLinkContexts(SpanContextPropagationExtractor extractor) {
+            List<SpanContext> result = new ArrayList<>();
+            if (extractor == null) {
+                return result;
+            }
+
+            Object value = extractor.get(SPAN_LINK_CONTEXT_PROPERTY);
+
+            if (value instanceof SpanContext) {
+                result.add((SpanContext) value);
+            } else if (value instanceof String) {
+                try {
+                    String str = (String) value;
+                    // Format: "traceId1:spanId1,traceId2:spanId2,..."
+                    // Split by comma to support multiple span links
+                    for (String part : str.split(",")) {
+                        String[] components = part.trim().split(":");
+                        if (components.length == 2) {
+                            SpanContext ctx = SpanContext.createFromRemoteParent(
+                                    components[0], components[1],
+                                    io.opentelemetry.api.trace.TraceFlags.getSampled(),
+                                    io.opentelemetry.api.trace.TraceState.getDefault());
+                            result.add(ctx);
+                        }
+                    }
+                } catch (Exception e) {
+                    LOG.warn("Failed to parse span link contexts from string: {}", value, e);
+                }
+            }
+
+            return result;
         }
 
         // We inspect the exchange in order to find any baggage variable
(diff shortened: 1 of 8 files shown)


components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetryTracer.java
```diff
@@ -180,6 +183,11 @@ protected void doShutdown() {
     private class OpentelemetrySpanLifecycleManager implements SpanLifecycleManager {
 
         private final static String BAGGAGE_VAR_PREFIX = "OTEL_BAGGAGE_";
+        /**
+         * Exchange property name for storing a span context to create span links. This is used by components like
+         * vertx-websocket to link WebSocket message spans back to the original HTTP upgrade request span.
```
davsclaus, line 188: This hardcodes a vertx-websocket-specific constant (`CamelVertxWebsocketHandshakeSpanContext`) inside the generic OpenTelemetry tracer. The `camel-opentelemetry2` module should not know about individual component internals.

The span link concept should be added to the `camel-telemetry` abstraction layer (e.g., `SpanLifecycleManager.create()`) so any component can benefit from it, not just vertx-websocket.

components/camel-opentelemetry2/src/main/java/org/apache/camel/opentelemetry2/OpenTelemetryTracer.java
```diff
@@ -231,6 +247,49 @@ public String get(SpanContextPropagationExtractor carrier, String key) {
...
+            Object value = extractor.get(SPAN_LINK_CONTEXT_PROPERTY);
+
+            if (value instanceof SpanContext) {
+                result.add((SpanContext) value);
+            } else if (value instanceof String) {
+                try {
+                    String str = (String) value;
+                    // Format: "traceId1:spanId1,traceId2:spanId2,..."
+                    // Split by comma to support multiple span links
+                    for (String part : str.split(",")) {
```
davsclaus, line 275: The custom `traceId:spanId` serialization format is fragile and non-standard. If span links become a first-class feature in the telemetry abstraction, consider using W3C Trace Context (`traceparent` header format) for serialization, which is already the standard used by the context propagation layer.

components/camel-telemetry/src/main/java/org/apache/camel/telemetry/decorators/VertxWebsocketSpanDecorator.java
```diff
@@ -0,0 +1,185 @@
...
+                if (endpoint == null) {
+                    return;
+                }
+
+                // Get all peers
+                Method findPeersMethod = endpoint.getClass().getMethod("findPeerObjectsForHostPort");
+                @SuppressWarnings("unchecked")
+                List<Object> allPeers = (List<Object>) findPeersMethod.invoke(endpoint);
+
+                if (allPeers == null || allPeers.isEmpty()) {
```
davsclaus, line 96: `exchange.getContext().hasEndpoint(...)` is unreliable for endpoint lookup — it only returns an endpoint if it happens to be registered in the `EndpointRegistry` by that exact URI. This may return `null` in many valid scenarios (e.g., if the URI doesn't match exactly, or the registry has been cleaned up).

components/camel-telemetry/src/main/java/org/apache/camel/telemetry/decorators/VertxWebsocketSpanDecorator.java
```diff
@@ -0,0 +1,185 @@
...
+                    return;
+                }
+
+                // Determine which peers will actually receive the message
```
davsclaus, line 100: This reflection call assumes the endpoint has a method `findPeerObjectsForHostPort()`. The vertx-websocket PR adds this as a `public` method, but the decorator in `camel-telemetry` cannot compile-time verify it exists. If the method is ever renamed, this will silently fail (caught by the empty `catch` on line 170). Consider whether a different integration approach would be more robust.

components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketEndpoint.java
```diff
@@ -250,9 +251,10 @@ protected ServerWebSocket findPeerForConnectionKey(String connectionKey) {
     }
 
     /**
-     * Finds all WebSockets associated with a host matching this endpoint configured port and resource path
+     * Finds all VertxWebsocketPeer objects associated with a host matching this endpoint configured port and resource
+     * path
```
davsclaus, line 255: This changes API visibility from `protected` to `public` and changes the return type from `Map<String, ServerWebSocket>` to `List<VertxWebsocketPeer>`. The existing `findPeersForHostPort()` is kept but now delegates to this new method.

Per project standards, public API changes require justification. If the only consumer is the reflection-based span decorator, consider whether a different design would avoid the need for this public API.

components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketHost.java
```diff
@@ -85,6 +86,10 @@ public void connect(VertxWebsocketConsumer consumer) {
         }
 
         route.handler(routingContext -> {
+            // Capture HTTP request span context BEFORE WebSocket upgrade
```
davsclaus, line 89: Capturing span context in the HTTP upgrade handler is the right idea — this is the correct point in the lifecycle where the HTTP span is still active. However, the reflection approach here (while avoiding compile-time dependency) means this code won't produce compile errors if the OTel API changes. Consider whether a conditional import / optional dependency pattern would be more maintainable.

davsclaus (changes requested):
Thank you for working on WebSocket span linking — it's a valuable feature for distributed tracing across HTTP/WebSocket boundaries. The approach needs some architectural changes before it can be merged. Here are the key findings:

### Blocking Issues

**1. Component-specific code in the generic tracer**

The `OpenTelemetryTracer` now hardcodes `CamelVertxWebsocketHandshakeSpanContext` — a vertx-websocket-specific constant inside the generic OTel tracer. This violates the separation of concerns established in [`design/tracing.adoc`](https://github.com/apache/camel/blob/main/design/tracing.adoc), which explicitly states that implementation-specific details must not leak into the abstraction.

The `SpanLifecycleManager` interface currently has no concept of span links. If span links are needed (and they are a useful feature), this should be designed as an enhancement to the `camel-telemetry` abstraction layer first — e.g., extending `SpanLifecycleManager.create()` — rather than bolted onto the OTel implementation for a single component. I'd recommend discussing this with @squakez (the tracing redesign author).

**2. No tests**

The PR adds significant functionality across 3 modules (`camel-vertx-websocket`, `camel-telemetry`, `camel-opentelemetry2`) with zero test coverage. Per project rules, every PR must include tests for new functionality.

**3. CI build failure**

The Javadoc comment in `VertxWebsocketConstants.java` gets reformatted by the build's formatter step, producing an uncommitted change that fails CI. Run `mvn formatter:format impsort:sort` in the affected module and commit the result.

**4. Public API changes without justification**

- `VertxWebsocketEndpoint.findPeerObjectsForHostPort()` — new `public` method. Per project standards, public API changes need justification.
- `VertxWebsocketProducer.getConnectedPeers()` — changed from `private` to `protected`.

These visibility changes exist solely to support the reflection-based span decorator. The design should avoid exposing internal API.

### Other Concerns

**5. Fragile reflection in `VertxWebsocketSpanDecorator`**

The decorator uses extensive reflection (~185 lines, mostly reflection) to call `findPeerObjectsForHostPort()`, `getConfiguration()`, `isSendToAll()`, `getHandshakeSpanContext()`, etc. This will silently break if any method is renamed. Compare with `VertxHttpSpanDecorator` (~20 lines, no reflection). Additionally, `exchange.getContext().hasEndpoint(...)` is unreliable as an endpoint lookup.

**6. No JIRA issue**

The PR references `camel-quarkus/issues/8280` but has no CAMEL JIRA ticket. Per project conventions, commits should follow `CAMEL-XXXX: Brief description` format.

### Questions

- The PR title says "POC" — is this intended for discussion or as a production-ready contribution?
- Has the tracing redesign owner (@squakez) been consulted on adding span link support?

### Suggestion

1. Open a CAMEL JIRA issue to track this work
2. Discuss with @squakez about adding span link support to `SpanLifecycleManager` at the abstraction level
3. Design a generic span link mechanism rather than a vertx-websocket-specific one

JinyuChen97:
@davsclaus Thanks for the comments, this POC was only a quick show and verify what the improved linked span should looks like in Jagger UI. I will discuss with @squakez when I restart these improvements. Would you want me to close this pull request or leave it as draft pr?

davsclaus:
you are welcome to close

PR #24367, opened 2026-07-01
CAMEL-23872: Add JFR Old Object Sample panel for memory leak diagnosis

davsclaus:
## Summary

- Adds JFR-based old object sampling for memory leak diagnosis across all Camel tooling layers
- Dev console manages JFR recording lifecycle and aggregates samples by class + stack trace fingerprint
- CLI command (`camel cmd jfr-old-objects`) with `--stacktrace`, `--min-size`, `--top` options
- TUI tab with recording progress, sortable/filterable table, detail panel with reference chains and stack traces
- MCP tool for AI agent access to JFR old object data

### Features

- **Aggregation**: Samples from the same class and allocation site are grouped together with count and total size
- **Readable names**: JVM array descriptors (`[B`, `[Ljava.lang.String;`) converted to human-readable form
- **Min-size filter**: TUI cycles through preset sizes (1 KB to 100 MB), CLI accepts `--min-size 1MB`
- **Stack trace dimming**: JDK/Jakarta framework frames dimmed in TUI detail panel
- **Recording age**: Title shows how old the cached results are (in minutes)

## Test plan

- [x] Build all affected modules (camel-console, camel-cli-connector, camel-jbang-core, camel-jbang-plugin-tui, camel-jbang-mcp, catalog)
- [x] Run leak test app and verify samples are captured and aggregated
- [x] Verify TUI tab: recording progress, results table, detail panel, sorting, min-size filter, PgUp/PgDn scrolling
- [x] Verify CLI: `--query`, `--stacktrace`, `--min-size` options
- [ ] CI checks


dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/action/CamelJfrOldObjects.java
```diff
@@ -0,0 +1,534 @@
...
+
+        // run 1
+        printer().printf("Recording 1 of 2 (%ds)...%n", dur1);
+        JsonObject startResp = sendAction(pid, "start", dur1);
+        if (startResp == null || !"recording".equals(startResp.getString("status"))) {
+            printer().println("Error: Failed to start recording 1");
+            return 1;
+        }
+
+        Thread.sleep((dur1 + 3) * 1000L);
```
gnodet: This `Thread.sleep((dur1 + 3) * 1000L)` blocks the calling thread for the full recording duration (up to 63+ seconds with defaults). The dev console already auto-stops via its `scheduler.schedule()` and reports completion through the `status` command.

Consider polling status instead — similar to how `ActionBaseCommand.getJsonObject()` polls with a `StopWatch` loop:

```java
// Poll status until recording completes (or times out)
StopWatch watch = new StopWatch();
long timeout = (dur1 + 30) * 1000L;
while (watch.taken() < timeout) {
    Thread.sleep(2000);
    JsonObject sts = sendAction(pid, "status", 0);
    if (sts != null && !"recording".equals(sts.getString("status"))) {
        break;
    }
}
```

This would detect early completion/failure immediately and eliminate the `+3` second padding that could be too short on a loaded system. Same applies to the second recording sleep below.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RuntimeTools.java
```diff
@@ -327,6 +327,140 @@ public JsonObject camel_runtime_heap_histogram(
...
+        if (r1 == null || !"recording".equals(r1.getString("status"))) {
+            throw new ToolCallException("Failed to start recording 1", null);
+        }
+
+        try {
+            Thread.sleep((dur1 + 3) * 1000L);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new ToolCallException("Interrupted during recording 1", null);
+        }
```
gnodet: The MCP tool blocks for `(dur1 + 3) * 1000L` here and again below for recording 2 — potentially 96+ seconds total with 30s default. MCP clients may have their own timeouts, and a blocked tool call gives no feedback to the caller.

Same suggestion as in the CLI command: poll the `status` action every 2–3 seconds instead. This also applies to the second sleep at line ~437.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/JfrOldObjectSampleTab.java
```diff
@@ -0,0 +1,1568 @@
...
+        }
+    }
+
+    private void scheduleResultsPoll(String pid, int dur) {
+        if (ctx.runner == null) {
+            return;
+        }
+        startDaemonThread("jfr-poll-" + pid, () -> {
+            try {
+                Thread.sleep((dur + 3) * 1000L);
```
gnodet: Polling status at shorter intervals (e.g., every 2–3 seconds) instead of sleeping for the full `(dur + 3)` seconds would also let the TUI refresh the progress bar with actual server-reported elapsed/remaining time, rather than relying on locally-calculated estimates.

core/camel-console/src/main/java/org/apache/camel/impl/console/JfrMemoryLeakDevConsole.java
```diff
@@ -0,0 +1,808 @@
...
+            case "stop" -> doStop(options);
+            case "status" -> doStatus();
+            case "query" -> doQuery(options);
+            case "compare" -> doCompare(options);
+            default -> errorJson("Unknown command: " + command);
+        };
+    }
+
+    private JsonObject doStart(Map<String, Object> options) {
+        if (activeRecording != null) {
```
gnodet, line 100: Minor: `doStart()` is not synchronized, but `doStopRecordingAndParse()` is. The check-then-act on `activeRecording` here has a theoretical race — two concurrent callers could both pass the null check and create two `Recording` instances, leaking the first. Making `doStart` synchronized (or using `compareAndSet` on an `AtomicReference`) would close the gap. Low severity since concurrent calls to this console are unlikely in practice.

gnodet (commented):
Nice feature — the dual recording with trend comparison is a really thoughtful addition for memory leak diagnosis, and the JFR API handling (multi-JDK field fallbacks, reference chain traversal, readable array names) is solid work.

A few items worth considering:

**Thread.sleep for recording waits (CLI, MCP, TUI):** The long `Thread.sleep((dur + 3) * 1000L)` calls (up to ~123 seconds in dual mode) could be replaced with status polling. The dev console already has auto-stop via `scheduler.schedule()` and a `doStatus()` endpoint that reports `recording` vs `completed`. Polling status every 2–3 seconds would eliminate the fragile `+3` padding, detect early failures, and avoid blocking threads for extended periods — especially relevant for the MCP tool where callers may have their own timeouts. The existing `ActionBaseCommand.getJsonObject()` poll pattern (100ms loop with `StopWatch` timeout) could serve as a model. The 500ms sleeps after `System.gc()` in the dev console are fine — standard JFR diagnostic practice.

**Tests:** The `readableClassName()` method is package-visible and has non-trivial edge cases (multi-dimensional arrays, primitive descriptors, `L...;` class descriptors) — a unit test would be easy to add and would protect against regressions. The `applyFilters()` and aggregation logic are also good candidates.

See inline comments for specifics.

This review does not replace specialized static analysis tools such as SonarCloud, CodeRabbit, or Sourcery.

_This review was generated by an AI agent (Claude Code on behalf of forgebot) and may contain inaccuracies. Please verify all suggestions before applying._

oscerd (commented):
Really nice, complete feature — dev-console + CLI + MCP + TUI, catalog/metadata fully regenerated, and the generated CLI doc + index are in place. A few things worth addressing:

- **Tests**: there are no tests in the PR (0 of 23 files). Most of it (live JFR + terminal rendering) is genuinely hard to unit-test, and I see the same no-test pattern in the precedent CAMEL-23837 — but two pieces are pure and cheap to cover: `readableClassName` (`JfrOldObjectSampleDevConsole.java:857-887`) and the trend classification in `doCompare` (`:474-495`). `camel-console` already has `AbstractDevConsoleTest` + ~19 `*DevConsoleTest`, so at least a small unit test for those two would be valuable.
- **MCP dual-mode default duration mismatch**: `doDualJfrRecording` uses `int dur = 30` and clamps `dur <= 0` → 30 (`RuntimeTools.java:1805, 1809-1810`), but the `@ToolArg` description says *"default 60, use 0 for manual stop"* (`:1777`) and the CLI `--duration` defaults to 60. So the MCP dual default (30s + 60s) diverges from the CLI (60s + 120s), and "0 for manual stop" doesn't hold in dual mode (0 becomes 30). Worth aligning the doc and behavior.
- **CI**: no checks have run on this branch yet (0 reported) — just make sure the formatter / impsort / generated-sources gates go green before merge.

(For anyone grepping the diff: the `Thread.sleep(...)` calls and `System.gc()` here are in the JFR recording paths, not tests, and are intentional — no Awaitility concern.)

---

PR #24377, opened 2026-07-02
CAMEL-23868: camel-file - make local work directory / starting directory containment checks path-boundary aware

oscerd:
## What

`GenericFileHelper.jailToLocalWorkDirectory` (added in CAMEL-23765 to keep remote-file
`localWorkDirectory` downloads inside the configured work directory) and
`GenericFileProducer.jailedCheck` (the `jailStartingDirectory` producer check) both decided
containment with a bare `String.startsWith` prefix test.

A bare prefix test ignores path-segment boundaries. When the compacted directory string has no
trailing separator — which is what `jailToLocalWorkDirectory` gets from `File.getPath()` — a
sibling directory whose name merely *extends* the configured directory's name (for example
`.../localwork` vs `.../localworkEVIL`) also satisfies the prefix and is wrongly treated as
contained. `jailedCheck` avoids this today only incidentally, because the producer always hands it
a `baseDir` that carries a trailing separator.

## Change

- Add a shared, boundary-aware `GenericFileHelper.isWithinDirectory(compactTarget, compactDir)` and
  route both containment checks through it, so the two can no longer drift apart.
- The helper compares on path-segment boundaries (`equals`, or `startsWith(dir + File.separator)`),
  tolerates a trailing separator on the directory (the producer supplies one), and treats an empty
  directory as "no boundary" — preserving the existing behaviour for all legitimate cases.

## Tests

- `GenericFileHelperTest`: a name-prefixed sibling (`../localworkEVIL/...`) is now rejected, and a
  new test locks the shared helper's boundary, trailing-separator and empty-directory behaviour.

Continues the containment work from CAMEL-23765.

Issue: https://issues.apache.org/jira/browse/CAMEL-23868

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
@@ -41,12 +41,36 @@ public final class GenericFileHelper {
         // compact first as the remote relative name can use ../ etc
         String compactTarget = FileUtil.compactPath(target.getPath());
         String compactWork = FileUtil.compactPath(localWorkDirectory.getPath());
-        if (!compactTarget.startsWith(compactWork)) {
+        if (!isWithinDirectory(compactTarget, compactWork)) {
             throw new GenericFileOperationFailedException(
                     "Cannot retrieve file to local work file: " + compactTarget
                                                           + " as it is jailed to the local work directory: "
                                                           + compactWork);
         }
+    }
+
+    /**
+     * Determines whether a compacted target path is contained within a compacted directory path, using a path-segment
+     * boundary comparison. Unlike a bare string prefix test, a sibling directory whose name merely extends the
+     * directory name (for example {@code /work} versus {@code /workspace}) is not considered contained. A trailing
+     * separator on the directory is tolerated, and an empty directory imposes no boundary.
+     *
+     * @param  compactTarget the compacted target path (see {@link FileUtil#compactPath(String)})
+     * @param  compactDir    the compacted directory the target must stay within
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
Reviewed against the project's contribution rules and conventions.

**Correctness — the fix is sound:**
- The new `isWithinDirectory` correctly handles: exact match, child paths via `startsWith(dir + File.separator)`, trailing separators on the directory, and empty directory (no boundary).
- Both call sites (`jailToLocalWorkDirectory` and `jailedCheck`) route through the same helper, eliminating the prior risk of the two checks drifting apart.
- `FileUtil.compactPath` normalises separators before the comparison reaches `isWithinDirectory`, so mixed-separator edge cases are handled.

**Tests — adequate:**
- The new `isWithinDirectoryRespectsPathBoundaries` test covers exact match, children, trailing separator tolerance, name-prefix sibling rejection, and empty directory — good boundary coverage.
- The existing `shouldRejectFilesEscapingLocalWorkDirectory` test gets an additional assertion for the `localworkEVIL` sibling scenario, locking the fix at the integration level.

**Project conventions — all met:**
- Commit message follows `CAMEL-23868: <description>` format.
- JIRA issue linked in PR body.
- AI attribution present.
- Single commit, focused scope, no unrelated changes.

**Prior history — no conflicts:**
- `jailToLocalWorkDirectory` was introduced in CAMEL-23765 by the same author. This PR is a direct continuation fixing a gap identified in that work.
- `jailedCheck` in `GenericFileProducer` was incidentally safe because the producer always appends a trailing separator to `baseDir` — but that's fragile. Routing through the shared helper is the right fix.

PR #24384, opened 2026-07-02
CAMEL-20206: refactor complex method getVersion()

fgalaf:
# Description

This PR is one step further for issue [CAMEL-20206](https://issues.apache.org/jira/browse/CAMEL-20206).
Refactored method getVersion() from VersionHelper.java class. Replaced try block with try-with-resources for readability and also introduced helper method since both try blocks were the same.

# Target

# Tracking


oscerd (approved):

catalog/camel-catalog/src/main/java/org/apache/camel/catalog/VersionHelper.java
```diff
@@ -92,4 +57,17 @@ public synchronized String getVersion() {
         return version;
     }
 
+    private String getVersionFromProperties(String properties) {
```
davsclaus: [nit] The parameter name `properties` is a bit ambiguous — it reads like a `Properties` object rather than a classpath resource path. Consider renaming to `resourcePath`:

```suggestion
    private String getVersionFromProperties(String resourcePath) {
```

(and update the usage on line 62 accordingly)

davsclaus (approved):
Nice cleanup — the extract-method refactoring is clean and behaviorally identical to the original. try-with-resources is a good improvement over the manual close blocks.

One minor nit: the parameter name `properties` in `getVersionFromProperties(String properties)` reads as if it's a `Properties` object rather than a classpath resource path. Something like `resourcePath` would be slightly clearer, but this is non-blocking.

catalog/camel-catalog/src/main/java/org/apache/camel/catalog/VersionHelper.java
fgalaf: Suggestion applied.

Croway (approved):

PR #24386, opened 2026-07-02
CAMEL-23237: camel-jbang catalog tables fill the terminal width

ammachado:
# Description

The `camel catalog` commands (`component`, `dataformat`, `language`, `transformer`, `kamelet`, ...) previously truncated the `DESCRIPTION` column at a fixed 80 characters regardless of terminal size, and each command hand-rolled its column definitions with inconsistent widths (`NAME` was 60 for `transformer` and 30 elsewhere).

This PR sizes the last free-text column to fill the detected terminal width and standardizes the shared column definitions.

**Changes:**

- New `common/CamelTableColumns.java` toolkit: `name()`, `since()`, `lastText()`, `measure()`, `lastColumnWidth()`. `measure()` computes a column's actual rendered width (header vs. longest cell, capped); `lastColumnWidth()` gives the remaining terminal width to the last column and truncates it with an ellipsis so each row stays on one line.
- `TerminalWidthHelper`:
  - Added `fillWidth()` (no upper cap, unlike `flexWidth()`).
  - Terminal width is now also detected on Windows (`cmd`/PowerShell) via `mode con`, in addition to the existing `COLUMNS` / `stty size` detection. Column parsing is positional (second integer) so it survives localized Windows output where the `Columns:` label is translated.
  - `DEFAULT_WIDTH` stays at 120 (the fallback when no terminal can be detected, e.g. piped/redirected output).
- Migrated `CatalogBaseCommand`, `CatalogTransformer` (dropped its `nameWidth()` override), `CatalogKamelet`, `UpdateList`, and `InfraBaseCommand` (`camel infra list`) onto the toolkit.
- Documented in the 4.22 upgrade guide. For full, untruncated output suitable for scripting, use `--json`.

**Follow-up fix included in this PR (found during self-review):** `noBorderOverhead(n)` returned `(n-1)*2`, but `AsciiTable` `NO_BORDERS` pads every column with one leading and one trailing space, so the real overhead is `n*2`. This was harmless while every table used capped `flexWidth`, but the new uncapped fill sizes the last column to the exact remainder, so the 2-column underestimate made the widest rows render at `terminalWidth + 2` and soft-wrap. Fixed at the root (safe for all existing `flexWidth` callers, which clamp to `[min, max]`) and covered by a new rendering-level regression test.

**Review updates:**

- An earlier revision lowered `DEFAULT_WIDTH` from 120 to 80; that shared-default change is reverted (80 is too narrow for modern terminals, and it narrowed the no-TTY fallback for every not-yet-migrated `NO_BORDERS` table). `UpdateList` is now migrated onto the toolkit rather than left as follow-up, which is the structural fix for the previous `UpdateListTest` CI failure: its `DESCRIPTION` column is sized from measured column widths instead of the shared fallback constant.
- `camel infra list` (`InfraBaseCommand`) is now migrated too: `DESCRIPTION` fills the terminal, and `IMPLEMENTATION` / `SERVICE_DATA` truncate with an ellipsis instead of letting the raw JSON service data overflow the terminal. The `--json` service-data handling was extracted into a testable `parseServiceData()` helper (embedded as nested JSON, not an escaped string; unparseable data is omitted rather than emitted, and no longer risks an NPE on missing data), covered by two new `InfraTest` cases.

# Target

# Tracking

CAMEL-23237

# AI-assisted contributions

_AI-assisted with Claude Code (Claude Opus 4.8) on behalf of Adriano Machado._

### DIFF
--- a/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
@@ -132,5 +132,5 @@ example 60 for `transformer` and 30 elsewhere).
 
 Terminal width is now detected on Windows (`cmd` / PowerShell) via `mode con`, in addition to the
 existing `COLUMNS` / `stty size` detection. When no terminal can be detected (for example when the
-output is piped or redirected), the width falls back to 80 columns. For full, untruncated output
+output is piped or redirected), the width falls back to 120 columns. For full, untruncated output
 suitable for scripting, use the `--json` option.
--- a/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
@@ -134,3 +134,7 @@ Terminal width is now detected on Windows (`cmd` / PowerShell) via `mode con`, i
 existing `COLUMNS` / `stty size` detection. When no terminal can be detected (for example when the
 output is piped or redirected), the width falls back to 120 columns. For full, untruncated output
 suitable for scripting, use the `--json` option.
+
+The `camel infra list` table now sizes its `DESCRIPTION` column to the terminal width, and truncates
+the `IMPLEMENTATION` and `SERVICE_DATA` columns with an ellipsis instead of letting the raw service
+data overflow the terminal. The complete, structured service data remains available via `--json`.
--- a/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
+++ b/docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
@@ -120,3 +120,17 @@ discarding the message with no exception or log.
 Scalar value nodes are now sent as a single record. Only container nodes (`ArrayNode`, `ObjectNode`)
 are still split, which is unchanged. Any `convertBodyTo(...)` previously used as a workaround is no
 longer required.
+
+=== camel-jbang catalog tables fill the terminal width
+
+The `camel catalog` commands (`camel catalog component`, `camel catalog dataformat`,
+`camel catalog language`, `camel catalog transformer`, `camel catalog kamelet`, ...) now size the
+`DESCRIPTION` column to the detected terminal width instead of the previous fixed 80-character cap,
+so wide terminals show more of the description before it is truncated with an ellipsis. The `NAME`
+column width is also standardized across these commands (it previously differed per command, for
+example 60 for `transformer` and 30 elsewhere).
+
+Terminal width is now detected on Windows (`cmd` / PowerShell) via `mode con`, in addition to the
+existing `COLUMNS` / `stty size` detection. When no terminal can be detected (for example when the
+output is piped or redirected), the width falls back to 80 columns. For full, untruncated output
+suitable for scripting, use the `--json` option.
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/catalog/CatalogBaseCommand.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/catalog/CatalogBaseCommand.java
@@ -20,18 +20,19 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import com.github.freva.asciitable.AsciiTable;
 import com.github.freva.asciitable.Column;
 import com.github.freva.asciitable.HorizontalAlign;
-import com.github.freva.asciitable.OverflowBehaviour;
 import org.apache.camel.catalog.CamelCatalog;
 import org.apache.camel.catalog.DefaultCamelCatalog;
 import org.apache.camel.dsl.jbang.core.commands.CamelCommand;
 import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;
 import org.apache.camel.dsl.jbang.core.commands.MavenResolverMixin;
 import org.apache.camel.dsl.jbang.core.commands.QuarkusPlatformMixin;
+import org.apache.camel.dsl.jbang.core.common.CamelTableColumns;
 import org.apache.camel.dsl.jbang.core.common.CatalogLoader;
 import org.apache.camel.dsl.jbang.core.common.RuntimeCompletionCandidates;
 import org.apache.camel.dsl.jbang.core.common.RuntimeType;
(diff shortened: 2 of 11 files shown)


gnodet (commented):
Nice work, @ammachado — this is a well-structured improvement to the catalog table rendering. The new `CamelTableColumns` toolkit centralizes what was previously hand-rolled in every command with differing widths, and the `noBorderOverhead` bug fix (from `(n-1)*2` to `n*2`) is correct — `AsciiTable` `NO_BORDERS` pads every column with one leading and one trailing space.

The test coverage is good: `CamelTableColumnsTest` covers `measure()`, `lastColumnWidth()`, and the rendering-level regression test that verifies tables don't overflow the terminal width. The `TerminalWidthHelperTest` additions for `parseColumns`, `fillWidth`, and the localized Windows `mode con` output are thorough.

A few observations:

**1. [observation] `DEFAULT_WIDTH` change from 120 to 80**

This is documented in the upgrade guide, which is good. Worth noting that this changes the behavior for all existing callers of `getTerminalWidth()` that fall back to the default — not just the catalog commands. Any table/command that previously got 120 as the fallback will now get 80, which may truncate more aggressively when output is piped or redirected. The PR description says this is intentional ("the standard fallback when no terminal can be detected"), and 80 is indeed the standard terminal width assumption. Just flagging it since it's a cross-cutting default change.

**2. [nit] `parseColumns` assumes column count is always the second integer**

This works for both `stty size` ("rows cols") and English/French `mode con` output. However, some Windows locales or non-standard `mode con` output might have the integers in a different order (e.g., if a header line contains a number before the "Lines" field). The positional approach is pragmatic — label parsing would be fragile across locales — but worth documenting this assumption in the Javadoc (which it already does, so this is fine).

**3. [positive] Bug fix in `noBorderOverhead` is backward-compatible**

The fix changes the formula but all existing callers used `flexWidth()` which clamps to `[min, max]`, so the 2-column shift only matters for the new uncapped `fillWidth()`. Nice that the PR identified and fixed this proactively.

**CI:** The failures (`UpdateListTest.listUpdateVersions`) are in version-update listing, unrelated to this PR's catalog table changes. Appears to be a pre-existing flaky test.

_Reviewed with Claude Code on behalf of Guillaume Nodet (@gnodet). This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

oscerd (changes requested):
Thanks @ammachado — filling the DESCRIPTION column to the terminal width is a real improvement, the off-by-2 `noBorderOverhead` fix is correct, and I like that it's now backed by a rendering-level regression test (`CamelTableColumnsTest.lastColumnWidthKeepsRenderedTableWithinTerminal`) instead of an arithmetic-only assertion. There's a blocking CI regression though, so I'm requesting changes.

## Blocker: the shared width change breaks `UpdateListTest` (deterministic, in CI)

This PR changes two things in the shared `TerminalWidthHelper`:

- `DEFAULT_WIDTH` 120 → 80 (the no-TTY / piped fallback)
- `noBorderOverhead(n)` from `(columnCount - 1) * 2` to `columnCount * 2`

But `UpdateList` (`commands/update/UpdateList.java`) was intentionally left for follow-up and still calls `flexWidth(terminalWidth(), 45, noBorderOverhead(4), 20, 80)`. In CI there is no TTY, so `terminalWidth()` returns the fallback, and the DESCRIPTION column width collapses:

- before: `max(20, min(80, 120 − 45 − 6)) = 69` → the 43-char description fits on one line
- after:  `max(20, min(80, 80 − 45 − 8)) = 27` → the description word-wraps

`UpdateListTest.listUpdateVersions` (line 37) asserts `.contains("Migrates Camel 4 application to Camel 4.9.0")`, and that contiguous string no longer exists once the column wraps. The CI log confirms it: `build (17, false)` fails deterministically (through all surefire reruns) at `UpdateListTest.listUpdateVersions:37`, with the rendered table showing the wrapped `Migrates Camel 4 …` description. `build (25, false)` is the matrix cancel-cascade of the same failure.

### Suggested fix

Either migrate `UpdateList` in this PR as well, or make `UpdateListTest` width-robust (e.g. pin `COLUMNS`, or assert against a wrap-tolerant substring).

### Blast-radius note (non-blocking)

Lowering `DEFAULT_WIDTH` 120 → 80 narrows the piped/redirected fallback for **every** not-yet-migrated `NO_BORDERS` table, not only the catalog commands — `UpdateList` is just the one with a test pinning the old width. Worth a quick scan of the other process/action snapshot tests for the same coupling before the follow-up migration lands.

Everything else looks good — the upgrade-guide entry is clear and the new tests are strong. Once CI is green this should be an easy re-review.

---

davsclaus (changes requested):
80 width is way to small today

ammachado:
The rationale for 80-column width is that this is (or was) the default width for the Windows Command Prompt terminal. 

I'll revert this change.

davsclaus (commented):
## Review: CAMEL-23237 — camel-jbang catalog tables fill the terminal width

### Blocking Issue

**1. Merge conflicts** — The PR currently has `CONFLICTING` merge state. It needs a rebase onto `main` before CI can run and the changes can be verified.

### Prior Review Status

The two prior `CHANGES_REQUESTED` reviews from @oscerd and @davsclaus appear to be addressed by commit `a70d811` (reverts `DEFAULT_WIDTH` back to 120 and migrates `UpdateList` onto the `CamelTableColumns` toolkit). However, because of the merge conflict, CI cannot verify this.

### Confirmed Findings

**2. `noBorderOverhead` formula change is cross-cutting** (non-blocking, informational)

The fix from `(columnCount - 1) * 2` to `columnCount * 2` is correct — AsciiTable `NO_BORDERS` pads every column with 1 leading + 1 trailing space. However, this affects **~30 existing callers** across `CamelMemoryLeak`, `CamelThreadDump`, `CamelSpanAction`, `LoggerAction`, `CamelHistoryAction`, `CamelBrowseAction`, `CamelBeanDump`, `CamelProcessorStatus`, `CamelRouteStatus`, `MessageTableHelper`, etc.

The PR description states this is safe because all existing callers use `flexWidth()` which clamps to `[min, max]`. This is correct — the 2-column shift only becomes visible at mid-range terminal widths (where flex columns may render 2 chars narrower), and is actually a fix for the previous 2-char overflow. Worth noting for posterity but not blocking.

### Questions / Observations

**3. `InfraBaseCommand` still uses old pattern** — `InfraBaseCommand.java` (line 213-214) calls `flexWidth(...)` with hand-rolled fixed column widths but was not migrated to the new `CamelTableColumns` toolkit. This is fine as follow-up work, but if the contributor wants this PR to be the single source of truth for catalog-style tables, `InfraBaseCommand` follows the same pattern.

**4. Test coverage is good** — `CamelTableColumnsTest` covers `measure()`, `lastColumnWidth()`, and importantly includes a rendering-level regression test that renders an actual AsciiTable and verifies all lines fit within the terminal width. `TerminalWidthHelperTest` additions for `parseColumns`, `fillWidth`, and localized Windows `mode con` output are thorough.

**5. Upgrade guide entry** — Present, clear, and appropriately scoped under the `camel-jbang` section. Documents the behavior change and recommends `--json` for scripting.

**6. AI attribution** — Properly disclosed in PR body and commit trailers.

### Summary

This is a well-structured improvement. The `CamelTableColumns` toolkit centralizes duplicated column definitions, the `noBorderOverhead` fix is correct, and the adaptive column measurement via `measure()` is better than the previous hardcoded width guessing. The main action needed is resolving the merge conflict and getting CI green.

ammachado:
Following up on @davsclaus's review, specifically the `InfraBaseCommand` observation:

> **3. `InfraBaseCommand` still uses old pattern** — `InfraBaseCommand.java` (line 213-214) calls `flexWidth(...)` with hand-rolled fixed column widths but was not migrated to the new `CamelTableColumns` toolkit. This is fine as follow-up work, but if the contributor wants this PR to be the single source of truth for catalog-style tables, `InfraBaseCommand` follows the same pattern.

`camel infra list` (`InfraBaseCommand`) is now migrated onto the toolkit (commit 3766f13). The infra table isn't quite the same shape as the catalog tables (its flex column was a middle column with `NEWLINE` wrapping, and it has two trailing unbounded columns), so the migration also made a small layout change: `DESCRIPTION` fills the terminal width, and `IMPLEMENTATION` / `SERVICE_DATA` truncate with an ellipsis instead of letting the raw JSON service data overflow the terminal. The full, structured service data remains available via `--json` (now covered by tests confirming it is embedded as nested JSON rather than an escaped string).

One design question for reviewers before considering this settled:

- **Does `SERVICE_DATA` belong in the table at all?** It currently renders raw JSON, which is not very readable in a terminal. Should the column be dropped from the table view (keeping the data available via `--json` only), or should it instead surface user-friendly, actionable information about the service, e.g. the port and credentials for a database?

_Claude Code on behalf of Adriano Machado._

davsclaus (approved):
Nice work on this PR — the measured-column approach and the `CamelTableColumns` toolkit are a solid improvement over the hand-rolled per-command widths.

A few observations (none blocking):

- The `noBorderOverhead` fix from `(n-1)*2` to `n*2` is correct per AsciiTable's padding model. All ~50 existing `flexWidth` callers are safe because they clamp to `[min, max]`, absorbing the 2-column reduction.
- The positional `parseColumns()` approach ("take the second integer") is robust for both `stty size` and localized `mode con` output — good call avoiding label-based parsing.
- The rendering-level regression test (`lastColumnWidthKeepsRenderedTableWithinTerminal`) directly validates the claim that sized tables don't overflow — that's the right test to have.
- Scope is well-bounded: migrating `process/` and `action/` commands can be follow-up work.

davsclaus (approved):
Nice work, @ammachado — the measured-column approach and the `CamelTableColumns` toolkit are a solid improvement over the hand-rolled per-command widths.

A few observations (none blocking):

- The `noBorderOverhead` fix from `(n-1)*2` to `n*2` is correct per AsciiTable's padding model. All ~50 existing `flexWidth` callers are safe because they clamp to `[min, max]`, absorbing the 2-column reduction. The rendering-level regression test (`lastColumnWidthKeepsRenderedTableWithinTerminal`) is the right guard for this.
- The positional `parseColumns()` approach ("take the second integer") is robust for both `stty size` and localized `mode con` output. The fallback to `DEFAULT_WIDTH` (120) handles any unexpected format gracefully.
- The `InfraBaseCommand.parseServiceData` extraction into a testable static method with two new test cases (nested JSON and malformed-data) is a clean improvement.
- Upgrade guide entry is present and well-written, pointing users to `--json` for scripting.

All prior `CHANGES_REQUESTED` items have been addressed. Ready to merge once CI is green.

PR #24390, opened 2026-07-02
CAMEL-21438: Fix flaky JMS consumer startup races in IBM MQ and Artemis tests

ammachado:
# Description

Fixes flaky JMS tests caused by sending messages before JMS consumer routes have registered listeners on the broker.

**IBM MQ test infra (`camel-test-infra-ibmmq`)**
- Wait for both the primary listener port and the queue-manager startup log message (combined via `WaitAllStrategy` with individual timeouts, since chained `waitingFor()` calls replace rather than combine strategies)
- Verify a real JMS connection succeeds before tests run, using Awaitility (replacing the previous `TestUtils.waitFor` sleep-retry loop) for clearer timeout diagnostics

**Shared JMS test helper (`JmsTestHelper`)**
- Add a standalone `JmsTestHelper` with `waitForJmsConsumerRoutes(CamelContext, String...)` (100ms default route uptime threshold) plus a `minUptimeMillis` overload
- Keeping it standalone (rather than on a base class) gives a single static API reachable identically from all three JMS test lineages (`AbstractJMSTest`, `AbstractPersistentJMSTest`, and plain `CamelTestSupport` tests such as `TwoConsumerOnSameQueueIT`), removing the earlier dual static/instance-wrapper surface
- Null-check the route so a missing/removed id times out clearly instead of throwing `NullPointerException`

**Affected tests**
- `JmsComponentIbmMQTest` — wait on the JMS consumer route (not the direct producer), `JmsReplyToIbmMQTest`
- `SingleMessageSameTopicIT` — merge the two `@BeforeEach` methods into one so the route wait always precedes the send (JUnit does not honor `@Order` on lifecycle methods); topic consumers keep the 200ms threshold
- `MultipleMessagesSameTopicIT` — migrate inline uptime waits to the shared helper (200ms)
- `TwoConsumerOnSameQueueIT` — wait before sends and after route restart
- `TemporaryQueueRouteTest`, `JmsReplyToLoopIssueTest`

Route uptime is only a heuristic for broker listener registration. Queue consumers use the 100ms default; topic subscriptions take a little longer to register, so the topic tests keep the longer 200ms threshold.

# Target

# Tracking

https://issues.apache.org/jira/browse/CAMEL-21438

# AI-assisted contributions

Prepared with AI assistance (Cursor and Claude Code, on behalf of Adriano Machado). Commits include `Co-authored-by` trailers.

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/AbstractJMSTest.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/AbstractJMSTest.java
@@ -17,12 +17,9 @@
 
 package org.apache.camel.component.jms;
 
-import java.util.concurrent.TimeUnit;
-
 import jakarta.jms.ConnectionFactory;
 
 import org.apache.camel.CamelContext;
-import org.apache.camel.Route;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.test.infra.artemis.common.ConnectionFactoryHelper;
 import org.apache.camel.test.infra.artemis.services.ArtemisService;
@@ -32,7 +29,6 @@ import org.apache.camel.test.infra.core.annotations.RouteFixture;
 import org.apache.camel.test.infra.core.api.CamelTestSupportHelper;
 import org.apache.camel.test.infra.core.api.ConfigurableContext;
 import org.apache.camel.test.infra.core.api.ConfigurableRoute;
-import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Tags;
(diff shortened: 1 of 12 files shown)


davsclaus (approved):

davsclaus (approved):

orpiske (approved):
Nice work fixing these JMS consumer startup races — the centralized `waitForJmsConsumerRoutes()` utility is a solid improvement over the scattered inline waits, and the `WaitAllStrategy` fix in the IBM MQ infra is a good catch.

### Potential improvements for the future

**Utility placement**: `AbstractPersistentJMSTest` extends `CamelTestSupport`, not `AbstractJMSTest` — they are sibling hierarchies. This creates a dual API surface (static calls vs instance wrappers). Extracting to a standalone `JmsTestHelper` utility class would provide a single consistent API reachable from all three test lineages (`AbstractJMSTest`, `AbstractPersistentJMSTest`, and standalone `CamelTestSupport` tests like `TwoConsumerOnSameQueueIT`).

**IBM MQ connection check**: The `TestUtils.waitFor()` call in `IbmMQLocalContainerInfraService` uses a sleep-retry loop internally. Replacing it with Awaitility would be more consistent with the rest of the PR and would provide better timeout diagnostics.

Neither of these block the PR — the current approach is correct and pragmatic.

_This review was generated by an AI agent (Claude Code on behalf of Otavio Piske) and may contain inaccuracies. Please verify all suggestions before applying._

orpiske:
### Follow-up: `@Order` on `@BeforeEach` does not control lifecycle method execution order

After further analysis, `@Order(1)` on `waitForConnections()` and `@Order(2)` on `prepare()` in `SingleMessageSameTopicIT` **will not guarantee execution order**. JUnit Jupiter (including 5.14.x used by this project) only supports `@Order` for:

- **Test methods** via `@TestMethodOrder(MethodOrderer.OrderAnnotation.class)`
- **Test classes** via `@TestClassOrder`
- **Extension fields** via `@RegisterExtension`

For `@BeforeEach` lifecycle methods, [execution order within a single class is explicitly non-deterministic](https://docs.junit.org/6.1.0/writing-tests/test-execution-order.html), and the [feature request to support it (junit-team/junit5#2990)](https://github.com/junit-team/junit5/issues/2990) remains open.

This means `prepare()` (which sends the message) may still execute before `waitForConnections()`, preserving the original race condition.

**Suggested fix:** Merge the two `@BeforeEach` methods into one:

```java
@BeforeEach
void waitAndPrepare() {
    waitForJmsConsumerRoutes(TOPIC_ROUTE_UPTIME_MILLIS, "a", "b");

    getMockEndpoint("mock:a").expectedBodiesReceived("Hello World");
    getMockEndpoint("mock:b").expectedBodiesReceived("Hello World");
    template.sendBody("activemq:topic:SingleMessageSameTopicIT", "Hello World");
}
```

This ensures the wait always runs before the send. The rest of the PR looks good — this is the only spot where the fix relies on `@BeforeEach` ordering.

_This comment was generated by an AI agent (Claude Code on behalf of Otavio Piske) and may contain inaccuracies. Please verify before applying._

PR #24406, opened 2026-07-03
CAMEL-23891: camel-mail - filter Camel internal headers in MimeMultipartDataFormat unmarshal

oscerd:
This aligns `MimeMultipartDataFormat` unmarshalling (`headersInline=true`) with the camel-mail consumer: internal Camel headers (`Camel*`, matched case-insensitively) present in the external MIME headers are no longer copied verbatim onto the Camel message.

### Background

Previously `copyNonStandardHeaders()` copied every MIME header (except `Message-ID`, `MIME-Version`, `Content-Type`) straight into the message header map with no `HeaderFilterStrategy`. This is inconsistent with:

- the camel-mail **consumer**, which applies `MailHeaderFilterStrategy` (extends `DefaultHeaderFilterStrategy`) and filters the `Camel*` namespace case-insensitively on the inbound direction;
- the data format's own **marshal** path, where inline headers are only copied when they match an explicit `includeHeaders` pattern.

### Change

`copyNonStandardHeaders()` now delegates to `DefaultHeaderFilterStrategy.applyFilterToExternalHeaders(...)`, so `Camel*` headers arriving in MIME content are filtered on the inbound path. Ordinary application headers (e.g. `X-...`) continue to pass through unchanged.

- Added a unit test covering prefix + case-insensitive filtering and application-header passthrough.
- Added a 4.22 upgrade-guide entry (potential behavior change).

JIRA: https://issues.apache.org/jira/browse/CAMEL-23891

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Andrea Cosentino

### DIFF
--- a/components/camel-mail/src/main/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormat.java
+++ b/components/camel-mail/src/main/java/org/apache/camel/dataformat/mime/multipart/MimeMultipartDataFormat.java
@@ -50,8 +50,10 @@ import org.apache.camel.NoTypeConversionAvailableException;
 import org.apache.camel.attachment.Attachment;
 import org.apache.camel.attachment.AttachmentMessage;
 import org.apache.camel.attachment.DefaultAttachment;
+import org.apache.camel.spi.HeaderFilterStrategy;
 import org.apache.camel.spi.annotations.Dataformat;
 import org.apache.camel.support.DefaultDataFormat;
+import org.apache.camel.support.DefaultHeaderFilterStrategy;
 import org.apache.camel.support.ExchangeHelper;
 import org.apache.camel.support.MessageHelper;
 import org.apache.camel.util.IOHelper;
@@ -72,6 +74,7 @@ public class MimeMultipartDataFormat extends DefaultDataFormat {
     private String includeHeaders;
     private Pattern includeHeadersPattern;
     private boolean binaryContent;
+    private final HeaderFilterStrategy headerFilterStrategy = new DefaultHeaderFilterStrategy();
 
     public String getMultipartSubType() {
         return multipartSubType;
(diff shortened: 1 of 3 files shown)


davsclaus (approved):
**APPROVE** — Clean, well-scoped security-hardening change.

`MimeMultipartDataFormat.copyNonStandardHeaders()` previously copied every non-standard MIME header onto the Camel message with no filtering, allowing `Camel*` headers from external MIME content to be injected into the exchange. The fix adds a `DefaultHeaderFilterStrategy` on the inbound path, filtering `Camel*`/`camel*` prefixes case-insensitively — consistent with the mail consumer's own inbound filtering.

Verified:
- `DefaultHeaderFilterStrategy` with defaults correctly catches all case variants (`CamelFoo`, `camelBar`, `CAMELBaz`) via `inFilterStartsWith` + `lowerCase=true`
- Using `DefaultHeaderFilterStrategy` (not `MailHeaderFilterStrategy`) is correct — the data format operates on MIME content generically, not JavaMail sessions
- Test covers prefix matching, case-insensitivity, and normal header passthrough
- Upgrade guide entry documents the behavior change with a clear workaround
- Branch naming, commit format, JIRA linkage, AI attribution all follow project conventions

gnodet (approved):
Clean, well-scoped security hardening. Verified independently — agreed with the existing review.

**Analysis:**

1. **Correct strategy choice** — Using `DefaultHeaderFilterStrategy` (not `MailHeaderFilterStrategy`) is appropriate. The data format operates on generic MIME content independent of the mail component. `MailHeaderFilterStrategy` additionally filters `mail.smtp.*` / `mail.smtps.*` prefixes which are specific to the mail consumer's JavaMail session properties (CAMEL-23522) and don't apply to the data format.

2. **Case-insensitive filtering works correctly** — Traced through `DefaultHeaderFilterStrategy.doFiltering()`: the `inFilterStartsWith` defaults to `{ "Camel", "camel" }` and with `lowerCase=true` (default), all case variants are caught. For `CAMELBaz`: first `tryHeaderMatch("CAMELBaz", {"Camel", "camel"})` → false, then `lower = "camelbaz"` → `tryHeaderMatch("camelbaz", {"Camel", "camel"})` → true (matches `"camel"` prefix). Confirmed by the test's `assertNull` on all three case variants.

3. **Scope is correct** — Only the `headersInline=true` unmarshal path (`unmarshalWithInlineHeaders` → `copyNonStandardHeaders`) is affected. The `unmarshalWithExternalHeaders` path doesn't call `copyNonStandardHeaders`.

4. **Thread safety** — `DefaultHeaderFilterStrategy` is effectively immutable after construction (default configuration, no mutations). The `private final` field is safe for concurrent access.

5. **Upgrade guide** — Accurate, clear workaround documented.

This is consistent with the security model in `CLAUDE.md`: external message senders are untrusted, and `Camel*` headers from untrusted sources should be filtered to prevent header-injection attacks (the same class as CVE-2025-27636).

_Claude Code on behalf of Jiri Ondrusek_

PR #24412, opened 2026-07-03
CAMEL-23720: Add emoji icons to camel-jbang TUI tab headers and centralize glyphs

ammachado:
# Description

Adds emoji icons to the `camel-jbang-plugin-tui` monitor and centralizes all glyph handling:

- Introduce `TuiIcons` as the single source of truth for emoji and UI symbols. Add icons to every primary tab header, the More submenu, and the Go to popup; replace hardcoded footer navigation hints with `TuiIcons` constants.
- Consolidate the More submenu into a single `TabRegistry.MoreTab` record (icon, name, label, tab), replacing the fragile parallel arrays and the `MORE_TAB_KEY_INDEX` offset table. `PopupManager`, `McpFacade`, and `ActionsPopup` now read from this one list.
- Derive keyboard shortcuts from Windows-style `&` mnemonic markers embedded in each label, so the underline offset and the trigger key can no longer drift apart. Mnemonic triggers are case-insensitive.
- Icons use plain 2-column emoji without VS16 variation selectors (CAMEL-23818) to avoid width artifacts.

Review feedback addressed:

- Strip VS16 variation selectors from `DEV_PROFILE` and `WARN` so all `TuiIcons` constants are consistently plain 2-column emoji without VS16, matching the class Javadoc contract.
- Migrate `TuiHelper.fileEmoji()`, `fileEmojiByName()`, and `detectPomEmoji()` to use `TuiIcons` constants instead of raw emoji literals, completing the glyph centralization.

_AI-generated PR description on behalf of Adriano Machado, via Claude Code._

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FilesBrowser.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FilesBrowser.java
@@ -98,7 +98,7 @@ class FilesBrowser {
                     .forEach(p -> {
                         String name = p.getFileName().toString();
                         if (Files.isDirectory(p) && !name.startsWith(".")) {
-                            dirs.add(new FileEntry("📁", name, -1, p.toString(), true));
+                            dirs.add(new FileEntry(TuiIcons.FOLDER, name, -1, p.toString(), true));
                         } else if (Files.isRegularFile(p)) {
                             String emoji = TuiHelper.fileEmoji(p);
                             long size = 0;
@@ -118,7 +118,7 @@ class FilesBrowser {
 
         List<FileEntry> found = new ArrayList<>();
         if (!dir.equals(rootDir)) {
-            found.add(new FileEntry("📁", "..", -1, dir.getParent().toString(), true));
+            found.add(new FileEntry(TuiIcons.FOLDER, "..", -1, dir.getParent().toString(), true));
         }
         found.addAll(dirs);
         found.addAll(files);
@@ -267,7 +267,7 @@ class FilesBrowser {
         if (sourceViewer.isVisible()) {
             sourceViewer.renderFooter(spans);
         } else {
-            TuiHelper.hint(spans, "↑↓", "navigate");
+            TuiHelper.hint(spans, TuiIcons.HINT_SCROLL, "navigate");
             TuiHelper.hint(spans, "Enter", "open");
             TuiHelper.hint(spans, "Esc", "close");
         }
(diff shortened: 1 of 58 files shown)


dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/TuiIcons.java
```diff
@@ -0,0 +1,220 @@
...
+ * Single source of truth for emoji and symbolic icons used across the Camel TUI.
+ * <p/>
+ * Tab/menu icons use plain 2-column-wide emoji without VS16 variation selectors (see CAMEL-23818). Doctor and legacy
+ * status glyphs may still use mixed-width symbols until migrated.
+ */
+final class TuiIcons {
+
+    // ---- Brand & runtime ----
+    static final String CAMEL = "🐪";
+    static final String SPRING_BOOT = "🍃";
```
davsclaus, line 31: Minor: `DEV_PROFILE = "🛠️"` and `WARN = "⚠️"` (lines 35, 39) contain VS16 variation selectors, while the class Javadoc says to use "plain 2-column-wide emoji without VS16". The `primaryTabEmojisHaveNoVariationSelector` test only checks `PRIMARY_TAB_ICONS`. These were pre-existing values being centralized — consider either stripping VS16 or adding a doc note that these are intentional exceptions.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FilesBrowser.java
```diff
@@ -117,7 +117,7 @@ private boolean loadDirectory(Path dir) {
```
davsclaus, line 118: The `switch` in `fileType()` (line 315) now matches against `TuiIcons.CAMEL`, `TuiIcons.JAVA`, etc., but `TuiHelper.fileEmoji()` which produces the values being compared still returns raw emoji literals. Works because the values are identical, but a follow-up migrating `TuiHelper.fileEmoji()`/`fileEmojiByName()` to use `TuiIcons` would complete the centralization.

davsclaus (approved):
## Review: CAMEL-23720 — Add emoji icons to TUI tab headers and centralize glyphs

Well-structured refactoring across 5 commits. `TuiIcons` as single source of truth, `MoreTab` record consolidation, and mnemonic-derived shortcuts are all sound. Excellent test coverage (icon width, VS16 absence, shortcut uniqueness, historical sequence oracle, Browse/Configuration regression guard).

### Minor suggestions (non-blocking)

1. **`TuiHelper.fileEmoji()` / `fileEmojiByName()` still return raw emoji literals** — `FilesBrowser.fileType()` now switches on `TuiIcons` constants, but the methods producing the values it compares against haven't been migrated. Works because the values are identical, but contradicts the "single source of truth" principle this PR establishes. Good follow-up candidate.

2. **VS16 in some constants** — `DEV_PROFILE = "🛠️"` and `WARN = "⚠️"` contain VS16 (U+FE0F), while the class Javadoc and CAMEL-23818 reference say to avoid it. The `primaryTabEmojisHaveNoVariationSelector` test only checks `PRIMARY_TAB_ICONS`. These were pre-existing values; consider cleaning them up or documenting as exceptions.

3. **Icon alias naming** — `TuiIcons.OTEL` (🔗) is reused for MCP connection doctor checks (not OTel-related), and `TAB_STARTUP` reuses `QUARKUS` (🚀) though Startup isn't Quarkus-specific. Naming nitpicks, not bugs.

None of these are blocking. Nice contribution!

---
_This review does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud._

davsclaus (approved):

davsclaus (approved):

PR #24426, opened 2026-07-04
CAMEL-23909: Fix flaky CxfConsumerPayLoadFaultMessageTest

gnodet:
## Summary

Fix flaky `CxfConsumerPayLoadFaultMessageTest` in `camel-cxf-soap` reported in [CAMEL-23909](https://issues.apache.org/jira/browse/CAMEL-23909).

- **Root cause**: `CxfConsumer.extractFromBody()` only checked `body instanceof Throwable` to detect faults. When a route sets a `CxfPayload` containing a raw `<soap:Fault>` XML element directly on the message body (as in `CxfConsumerPayLoadFaultMessageTest`), the fault was not detected and went through CXF's normal response serialization path instead of the fault handling path, leading to intermittent failures depending on CXF/JDK internal serialization behavior.
- **Fix**: Enhanced `extractFromBody()` to also detect `CxfPayload` bodies containing SOAP Fault elements. Added `extractFaultFromPayload()` which checks for a single Element body with localName "Fault" in SOAP 1.1 or 1.2 namespace, extracts `faultcode`/`faultstring`/`detail`, and constructs a proper `SoapFault` that CXF can process through its standard fault handling pipeline.
- **Copilot review fix**: Changed `parseFaultCode()` to look up the namespace prefix from the `<faultcode>` element itself rather than the parent `<Fault>` element, so that namespace declarations on the code element are also visible. `lookupNamespaceURI` walks up the tree, so ancestor declarations are still found.
- **Testing**: All 161 tests in `camel-cxf-soap` pass (0 failures, 2 skipped). The specific test was run 5 additional times in isolation with 0 failures.

## Test plan

- [x] `CxfConsumerPayLoadFaultMessageTest` passes consistently (5/5 isolated runs)
- [x] `CxfConsumerPayloadFaultTest` (parent, uses `setException`) still passes
- [x] `CxfConsumerPayLoadMarshalFaultTest` (uses `SoapFault` as body) still passes
- [x] Full `camel-cxf-soap` test suite passes (161 tests, 0 failures)
- [x] CI build passes (JDK 17 ✅, JDK 25 ✅)

### DIFF
--- a/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfConsumer.java
+++ b/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfConsumer.java
@@ -18,11 +18,16 @@ package org.apache.camel.component.cxf.jaxws;
 
 import java.lang.reflect.Method;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import jakarta.xml.ws.WebFault;
 
+import javax.xml.namespace.QName;
+
 import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 import org.apache.camel.AsyncCallback;
 import org.apache.camel.ExchangePattern;
@@ -30,11 +35,13 @@ import org.apache.camel.ExchangeTimedOutException;
 import org.apache.camel.Processor;
 import org.apache.camel.Suspendable;
 import org.apache.camel.component.cxf.common.CxfBinding;
+import org.apache.camel.component.cxf.common.CxfPayload;
 import org.apache.camel.component.cxf.common.DataFormat;
 import org.apache.camel.component.cxf.common.UnitOfWorkCloserInterceptor;
 import org.apache.camel.component.cxf.common.message.CxfConstants;
 import org.apache.camel.support.DefaultConsumer;
 import org.apache.camel.util.ObjectHelper;
+import org.apache.cxf.binding.soap.SoapFault;
 import org.apache.cxf.continuations.Continuation;
 import org.apache.cxf.continuations.ContinuationProvider;
 import org.apache.cxf.endpoint.Server;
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

ammachado:
@gnodet the failed CI run log mentioned above is related to a LevelDB test, not to CXF. Also, extracting the fault as a `SoapFault` isn't a behavior change worth mentioning on the migration guide?

gnodet:

@ammachado Thank you for the review and good questions!

1. **CI run link**: You're right — the investigation confirmed that the linked CI run was actually cancelled (not failed due to CXF), and the `CxfConsumerPayLoadFaultMessageTest` test **did pass** in that run. The flakiness was reported in [CAMEL-23909](https://issues.apache.org/jira/browse/CAMEL-23909) based on observed intermittent failures on JDK 25. The root cause analysis identified that the fault detection path was fragile — it relied on CXF's internal response serialization behavior to handle a `CxfPayload` containing a raw `<soap:Fault>` element through the normal (non-fault) path, which could behave inconsistently depending on JDK/CXF internals.

2. **Migration guide**: This is a bug fix rather than a behavior change — when a route sets a `CxfPayload` containing a `<soap:Fault>` element as the message body, the clear intent is to return a SOAP fault to the client. The old code only detected `Throwable` bodies, so the fault XML was serialized through CXF's normal response path instead of the fault handling path. The fix makes this pattern behave consistently with the other two existing fault patterns:
   - `exchange.setException(soapFault)` — detected by `checkFailure()` (works correctly)
   - `exchange.getMessage().setBody(soapFault)` — detected by `extractFromBody()` since `SoapFault` is a `Throwable` (works correctly)
   - `exchange.getMessage().setBody(cxfPayload)` — **was not detected**, now fixed

   Since the end result for the client is the same SOAP fault response (just now going through the correct CXF fault pipeline), I don't think a migration guide entry is needed. No user would intentionally rely on a `CxfPayload<soap:Fault>` body *not* being treated as a fault.

squakez:
> 
> _Claude Code on behalf of @squakez_
> 

Why this is posting on behalf of myself? FYI @gnodet

PR #24448, opened 2026-07-06
CAMEL-23918: Fix flaky tests in camel-reactive-streams

gnodet:
## Summary

Fixes flaky tests in `camel-reactive-streams` by:

1. **Replacing `Thread.sleep()` with Awaitility and proper async patterns** across 5 test files:
   - `DelayedMonoPublisherTest`: Replaced `Thread.sleep` delays with `ScheduledExecutorService.schedule()` for delayed operations and Awaitility `await()` for condition-based waiting
   - `BackpressureStrategyTest`: Replaced `Thread.sleep(1000)` with `await().atMost().until()` using `AtomicInteger` counter to track timer event completion; replaced `Thread.sleep(200)` + `assertEquals` with `await().during().until()` for stability verification
   - `BackpressureSubscriberTest`: Replaced `Thread.sleep` in route processors with Camel's `delay()` EIP
   - `BackpressurePublisherRoutePolicyTest`: Replaced `Thread.sleep(250)` + assertion with `await().atMost().untilAsserted()`
   - `BasicPublisherTest`: Removed unnecessary `Thread.sleep(60)` between dispose and subscribe

2. **Fixing a race condition in BackpressureStrategyTest** where the async flush mechanism in `CamelSubscription` caused exact-value assertions to fail non-deterministically. The flush is scheduled on a worker pool, and if the next timer event arrives before the flush runs, the backpressure strategy modifies the buffer, changing which items get delivered. Replaced fragile exact-sum assertions with invariant-based assertions that verify the actual backpressure strategy behavior:
   - OLDEST strategy: first item is always 1 (oldest kept in buffer), second is an early buffered item
   - LATEST strategy: second item is always 20 (most recent always kept)

## Test plan

- [x] All 65 tests pass in `camel-reactive-streams` module
- [x] `BackpressureStrategyTest` passes without surefire auto-retry (`-Dsurefire.rerunFailingTestsCount=0`)
- [x] CI passes

### DIFF
--- a/components/camel-reactive-streams/src/test/java/org/apache/camel/component/reactive/streams/BackpressureStrategyTest.java
+++ b/components/camel-reactive-streams/src/test/java/org/apache/camel/component/reactive/streams/BackpressureStrategyTest.java
@@ -17,6 +17,8 @@
 package org.apache.camel.component.reactive.streams;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
@@ -116,8 +118,15 @@ public class BackpressureStrategyTest extends BaseReactiveTest {
         await().during(Duration.ofMillis(500))
                 .atMost(Duration.ofSeconds(2))
                 .until(() -> queue.size() == 2);
-        int sum = queue.stream().reduce((i, j) -> i + j).get();
-        assertEquals(3, sum); // 1 + 2 = 3
+        // Verify OLDEST strategy behavior:
+        // - First item is always 1 (OLDEST keeps the first buffered item)
+        // - Second item is an early value (the first item buffered after the initial flush)
+        // The exact second value depends on flush timing vs timer events,
+        // but it must be > 1 and much less than 20 (the last timer event)
+        List<Integer> items = new ArrayList<>(queue);
+        assertEquals(1, items.get(0).intValue(), "OLDEST strategy should deliver item 1 first");
+        assertTrue(items.get(1) > 1 && items.get(1) < 20,
+                "OLDEST strategy should keep an early buffered item, got: " + items.get(1));
 
         subscriber.cancel();
     }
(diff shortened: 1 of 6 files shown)


components/camel-reactive-streams/src/test/java/org/apache/camel/component/reactive/streams/BackpressureStrategyTest.java
```diff
@@ -145,12 +155,14 @@ public void onNext(Integer o) {
...
+        await().atMost(5, TimeUnit.SECONDS).until(() -> timerCount.get() >= 20);
 
         subscriber.request(19);
-        assertTrue(latch2.await(1, TimeUnit.SECONDS));
-        Thread.sleep(200); // add other time to ensure no other items arrive
-        assertEquals(2, queue.size());
+        assertTrue(latch2.await(2, TimeUnit.SECONDS));
+        // Verify exactly 2 items received and no more arrive
+        await().atMost(2, TimeUnit.SECONDS)
+                .untilAsserted(() -> assertEquals(2, queue.size()));
```
apupier: this is no more waiting for potential new messages erroneously to arrive

gnodet:
@copilot review

components/camel-reactive-streams/src/test/java/org/apache/camel/component/reactive/streams/BackpressureStrategyTest.java
gnodet: Good catch! Fixed in 2b503ed — replaced `await().untilAsserted()` with `await().during(500ms).until(() -> queue.size() == 2)` which verifies that the queue size remains stable at 2 for 500ms, ensuring no extra items erroneously arrive.

gnodet:

📋 **Process note**: Before undrafting, the test(s) modified in this PR should be run 100 times (e.g. with `mvn test -Dsurefire.rerunFailingTestsCount=0 -Dtest=<TestClass> -pl <module> -Dsurefire.runCount=100`) to confirm they are actually flaky. If they pass 100/100 without the fix, the change may not be needed.

Also ensure:
- MockEndpoint helpers (`assertIsSatisfied()` already waits 10s, `setResultWaitTime()`, `setAssertPeriod()`) are used where appropriate instead of Awaitility wrappers
- Valid `Thread.sleep()` calls (UDP pacing, simulation delays, "wait to confirm nothing happens") are preserved

gnodet:

Closing — this fix was not validated by running the test 100 times first. Will re-dispatch with a validation-first approach: run the test 100x, and only create a JIRA + fix if it actually fails.

gnodet:
### Third commit: Fix race condition in BackpressureStrategyTest

Investigation revealed a pre-existing race condition in `BackpressureStrategyTest` that caused the exact-sum assertions to fail non-deterministically (even without `Thread.sleep`).

**Root cause**: `CamelSubscription.checkAndFlush()` schedules the flush asynchronously on a worker pool thread. With the timer firing every 20ms, the next item can arrive and modify the buffer (via the backpressure strategy) before the flush actually runs:

- **OLDEST**: Item 1 enters buffer → Item 2 arrives before flush → OLDEST discards item 2 (keeps oldest) → flush delivers item 1 → item 3 becomes the new oldest → result: [1, 3] sum=4 instead of [1, 2] sum=3
- **LATEST**: Item 1 enters buffer → Item 2 arrives before flush → LATEST replaces item 1 with item 2 → flush delivers item 2 → item 20 ends up as latest → result: [2, 20] sum=22 instead of [1, 20] sum=21

**Fix**: Replace fragile exact-sum assertions with invariant-based assertions that verify the actual backpressure strategy behavior:
- **OLDEST**: first item is always 1 (oldest kept in buffer), second item is an early value (>1, <20)
- **LATEST**: second item is always 20 (most recent always kept by LATEST strategy)

gnodet:
@copilot review

davsclaus (approved):

davsclaus (approved):

PR #24449, opened 2026-07-06
chore: use non-blinking software block cursor in TUI

gnodet:
## Summary

- Replace the hardware terminal cursor in ShellPanel with a software-rendered reversed-video cursor that does not blink. The hardware cursor's blink timer resets on every TamboUI frame redraw, causing it to appear half-shown on each keypress.
- Replace the blinking half-block (`▌`) cursor in CaptionOverlay with a non-blinking full-block (`█`) character.
- The software cursor renders the character at the cursor position with reversed style, matching the approach already used by AiPanel and similar to how Claude Code renders its cursor.

## Changes

- **ShellPanel**: Add `convertRow(buffer, offset, width, cursorCol)` overload that emits the cursor column as a single reversed-video span. Pass cursor coordinates from `renderLiveView` instead of using `frame.setCursorPosition()`. Remove `lastCursorX`/`lastCursorY` tracking fields.
- **CaptionOverlay**: Replace `▌`/space blink toggle with static `█`.
- **ShellPanelTest**: Add 6 tests for cursor rendering (middle, start, end, null codepoint, no cursor, bold run split).

## Test plan

- [x] All 525 TUI module tests pass
- [x] New cursor rendering tests cover edge cases (cursor at start/middle/end, null cell, bold text, no-cursor sentinel)
- [ ] Manual: open `camel monitor` TUI, open shell panel (F6), verify cursor is a solid non-blinking reversed block
- [ ] Manual: verify cursor moves correctly with arrow keys, Home/End
- [ ] Manual: test CaptionOverlay inline input shows non-blinking block cursor

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CaptionOverlay.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CaptionOverlay.java
@@ -190,8 +190,7 @@ class CaptionOverlay {
     private void renderInline(Frame frame, Rect area) {
         Style style = Style.EMPTY.fg(Color.WHITE).bold();
         String text = inlineBuffer != null ? inlineBuffer.toString() : "";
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

davsclaus (approved):

davsclaus:
LGTM

PR #24450, opened 2026-07-06
CAMEL-23924: Fix flaky ServletStreamingChunkedTest

gnodet:
## Summary

Fix flaky `ServletStreamingChunkedTest` by replacing `Thread.sleep(500)` calls with proper `CountDownLatch`-based synchronization:

- **First sleep**: replaced with a `requestStarted` CountDownLatch that the main thread counts down after the HTTP request has been processed and response headers confirmed (200, chunked). This ensures the writer only starts producing data once the servlet is actively streaming the response.
- **Second and third sleeps** (after `latch1.await`/`latch2.await`): removed entirely as the existing latches already provide the necessary synchronization between reader and writer for subsequent chunks.

The fixed-delay `Thread.sleep(500)` was unreliable on slow CI environments (not enough time) and wasteful on fast machines (unnecessary waiting).

## Test plan

- [x] `mvn test -pl components/camel-servlet -Dtest=ServletStreamingChunkedTest` — passes
- [x] Ran test 3 additional times in a loop to verify stability — all green
- [x] `mvn formatter:format impsort:sort` — no formatting issues

### DIFF
--- a/components/camel-servlet/src/test/java/org/apache/camel/component/servlet/ServletStreamingChunkedTest.java
+++ b/components/camel-servlet/src/test/java/org/apache/camel/component/servlet/ServletStreamingChunkedTest.java
@@ -42,28 +42,25 @@ public class ServletStreamingChunkedTest extends ServletCamelRouterTestSupport {
     @Test
     public void testStreaming() throws Exception {
 
+        final CountDownLatch requestStarted = new CountDownLatch(1);
         final CountDownLatch latch1 = new CountDownLatch(1);
         final CountDownLatch latch2 = new CountDownLatch(1);
 
-        // use background thread to write to stream that camel-servlet uses as reponse
+        // use background thread to write to stream that camel-servlet uses as response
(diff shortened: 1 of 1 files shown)


gnodet:

📋 **Process note**: Before undrafting, the test(s) modified in this PR should be run 100 times (e.g. with `mvn test -Dsurefire.rerunFailingTestsCount=0 -Dtest=<TestClass> -pl <module> -Dsurefire.runCount=100`) to confirm they are actually flaky. If they pass 100/100 without the fix, the change may not be needed.

Also ensure:
- MockEndpoint helpers (`assertIsSatisfied()` already waits 10s, `setResultWaitTime()`, `setAssertPeriod()`) are used where appropriate instead of Awaitility wrappers
- Valid `Thread.sleep()` calls (UDP pacing, simulation delays, "wait to confirm nothing happens") are preserved

gnodet:

Closing — this fix was not validated by running the test 100 times first. Will re-dispatch with a validation-first approach: run the test 100x, and only create a JIRA + fix if it actually fails.

PR #24451, opened 2026-07-06
CAMEL-23914: handle @@CamelMagicValue@@ property conversion

mcarlett:
# Description

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/component/PropertyConfigurerSupport.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/component/PropertyConfigurerSupport.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.support.component;
 
+import java.time.Duration;
 import java.util.List;
 
 import org.apache.camel.CamelContext;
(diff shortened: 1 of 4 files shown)


davsclaus (changes requested):
you should not include the jbang metadata.json file as its not related to this and you have an out of sync local code base

core/camel-support/src/main/java/org/apache/camel/support/component/PropertyConfigurerSupport.java
```diff
@@ -95,10 +98,24 @@ public static <T> T property(CamelContext camelContext, Class<T> type, Object va
...
+                } else if (double.class == type || Double.class == type) {
+                    return (T) Double.valueOf(1d);
+                } else if (float.class == type || Float.class == type) {
+                    return (T) Float.valueOf(1f);
+                } else if (short.class == type || Short.class == type) {
+                    return (T) Short.valueOf((short) 1);
+                } else if (byte.class == type || Byte.class == type) {
+                    return (T) Byte.valueOf((byte) 0);
                 }
+            }
```
davsclaus, line 120: Please also add `Duration.class` handling here, matching what `ExportTypeConverter` does:
```suggestion
                } else if (byte.class == type || Byte.class == type) {
                    return (T) Byte.valueOf((byte) 0);
                } else if (java.time.Duration.class == type) {
                    return (T) java.time.Duration.ofMillis(1);
                }
```
And once this is in place, the duplicate MAGIC_VALUE handling in `ExportTypeConverter` and `DummyTypeConverter` can be removed.

davsclaus (commented):
Thanks for the fix Marco — the approach of handling `MAGIC_VALUE` in `PropertyConfigurerSupport.property()` is correct and puts it at the right layer (before `mandatoryConvertTo()`).

A couple of things to address:

1. **Add `Duration.class` handling** — `ExportTypeConverter` and `DummyTypeConverter` both handle `Duration.class → Duration.ofMillis(1)` for MAGIC_VALUE. Please add the same to `PropertyConfigurerSupport` so the fix is complete for all types.

2. **Remove the duplicate from `ExportTypeConverter`** — now that `PropertyConfigurerSupport.property()` handles MAGIC_VALUE for all primitive/wrapper types (and Duration), the corresponding code in `ExportTypeConverter` is redundant. Please remove the MAGIC_VALUE handling from `ExportTypeConverter` to avoid the duplication. Same for `DummyTypeConverter` in `camel-yaml-dsl-validator`.

Otherwise looks good — tests are comprehensive.

core/camel-support/src/main/java/org/apache/camel/support/component/PropertyConfigurerSupport.java
mcarlett: Done — added `Duration.class → Duration.ofMillis(1)` handling and removed the duplicate MAGIC_VALUE code from both `ExportTypeConverter` and `DummyTypeConverter`. Also added a `testMagicValueToDuration` test case.

mcarlett:
Thanks for the review @davsclaus. Addressed both points:

1. Added `Duration.class → Duration.ofMillis(1)` handling in `PropertyConfigurerSupport.property()` with a test case
2. Removed the duplicate MAGIC_VALUE handling from both `ExportTypeConverter` and `DummyTypeConverter` — they now only handle regular value conversion

davsclaus (approved):
Thanks Marco — all feedback addressed cleanly. Duration handling added, duplicate MAGIC_VALUE code removed from `ExportTypeConverter` and `DummyTypeConverter`. LGTM.

PR #24467, opened 2026-07-06
CAMEL-23883: Kafka producer no longer silently drops scalar Jackson ValueNode

matteominin:
# Description
Fixes [CAMEL-23883](https://issues.apache.org/jira/browse/CAMEL-23883).

With `useIterator=true`, the Kafka producer splits an `Iterable` body into one record per element. A Jackson `JsonNode` implements `Iterable`, so a scalar value node (e.g. the `IntNode` from `transform().jq(".my-value")`) was seen as an empty iterable: the loop ran zero times, `send()` was never called, and the message was silently dropped with no exception or log.

This narrows `isIterable()` so a `JsonNode` counts as a batch only when it is a container node (`isContainerNode()`). Scalar nodes now take the single-message path, so they are either delivered or fail with a clear serialization error, never silently dropped. Container nodes (`ArrayNode`/`ObjectNode`) are still split, unchanged.

## Open question / possible follow-up

This fixes the reported Jackson case, but the same silent drop could recur for any body  that is `Iterable` yet yields an empty iterator. An exception isn't safe (an empty `Collection` is a legitimate no-op).  
Would be worth adding a DEBUG log when the iterable path produces zero records?  
I'm happy to add it here or in a follow-up PR.

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaProducer.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaProducer.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.camel.AsyncCallback;
 import org.apache.camel.Exchange;
 import org.apache.camel.Message;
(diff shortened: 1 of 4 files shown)


davsclaus (approved):
Nice fix — clean, well-scoped, and properly tested.

The `JsonNode` check before the general `Iterable` check is the right approach since `JsonNode` implements `Iterable<JsonNode>`. Container nodes continue to be split; scalars now take the single-message path instead of being silently dropped. Unit tests cover both scalar and container cases, the integration test verifies end-to-end delivery, and the upgrade guide documents the behavior change.

`jackson-databind` is already a compile-scope dependency of `camel-kafka`, so no new coupling is introduced.

One minor observation (non-blocking): the follow-up idea in the PR description about adding a DEBUG log when the iterable path produces zero records would be a nice complementary improvement to catch similar silent-drop scenarios for other `Iterable` types.

PR #24468, opened 2026-07-06
CAMEL-22616: camel-main - Add support for camel.cluster.xxx configurations

davsclaus:
## Summary

Add declarative `camel.cluster.<type>.*` property support to camel-main/camel-jbang, matching Spring Boot's existing auto-configuration.

- **Runtime support**: `BaseMainSupport` routes `camel.cluster.*` properties, groups by type sub-prefix (file, kubernetes, zookeeper, consul, jgroups-raft, infinispan), resolves the cluster service via FactoryFinder (`@JdkService` annotations), binds properties, and adds as CamelContext service
- **Tooling metadata**: `PrepareCamelMainMojo` loads existing bean JSON metadata from component directories to generate 75 cluster configuration options in `camel-main-configuration-metadata.json` — enabling IDE auto-complete and validation
- **Auto-download**: camel-jbang auto-downloads the correct component JAR when `camel.cluster.<type>` properties are detected (via both property prefix and FactoryFinder mappings)

### Example usage
```properties
# application.properties
camel.cluster.file.root = /tmp/cluster
camel.cluster.file.acquireLockInterval = 5

camel.cluster.kubernetes.namespace = default
camel.cluster.kubernetes.podName = my-pod
```

## Test plan
- [ ] Build `tooling/maven/camel-package-maven-plugin`, `core/camel-main`, verify `camel-main-configuration-metadata.json` contains `camel.cluster.*` groups and properties
- [ ] Test with camel-jbang: create route with `camel.cluster.file.root=/tmp/cluster` in `application.properties`, verify `FileLockClusterService` starts
- [ ] Verify auto-download: use `camel.cluster.zookeeper.nodes=...` without explicit dependency, confirm camel-zookeeper JAR is downloaded


dsl/camel-kamelet-main/src/generated/resources/camel-factoryfinder-known-dependencies.properties
```diff
@@ -17,6 +17,11 @@
 
 META-INF/services/org/apache/camel/bean-processor-factory=camel:bean
 META-INF/services/org/apache/camel/bean-proxy-factory=camel:bean
+META-INF/services/org/apache/camel/cluster-service-consul=camel:consul
```
Croway, line 21: camel-infinispan is missing, does it need to be added in this file?

davsclaus: Good catch! The factory finder generation in `PrepareKameletMainMojo` was using `findCamelDirectory()` which walks **up** parent directories instead of scanning into nested modules. For `camel-infinispan/camel-infinispan/`, it would find an empty directory at the `components/` level and return no results.

Fixed by replacing `findCamelDirectory()` with direct path resolution and adding nested module scanning. This also picked up several other previously-missed factory finders from nested modules (AWS/Azure/Google/IBM secret manager periodic tasks, avro-rpc-jetty, spring-xml, microprofile-config).

Croway (approved):

PR #24471, opened 2026-07-06
CAMEL-23223: Fix flaky ManagedMessageHistoryAutoConfigIT

gnodet:
## Summary

_Claude Code on behalf of apupier_

Fix flaky `ManagedMessageHistoryAutoConfigIT` which has a **70% flaky rate** (113/161 runs fail) per the [Develocity dashboard](https://develocity.apache.org/scans/tests?search.rootProjectNames=camel&tests.sortField=FLAKY).

### Root cause

The OTel periodic metric exporter was configured with a very short export interval (300ms). On slow CI machines, the exporter fires during message processing, and the **last exported `MetricData`** may contain incomplete point data. The test's `assertPointDataForRouteId` assertion then fails because the last metric snapshot doesn't yet include all route data.

PR #24425 previously added `CopyOnWriteArrayList` thread-safety and Awaitility-based polling, but did **not** address the core timing issue — the export interval was still 300ms.

### Fix

Increase the `otel.metric.export.interval` from **300ms → 5000ms** so the first periodic export fires well after all messages have been processed (message processing takes <1s). This eliminates the race where an intermediate export captures partial data that becomes the "last" Camel metric seen by the Awaitility assertion.

### Validation

Ran the test **12+ consecutive times** locally — all passed. With the prior 70% failure rate, the probability of 12 consecutive passes by chance is 0.3^12 ≈ 5.3e-7, providing strong statistical evidence that the fix resolves the flakiness.


gnodet:
@copilot review

davsclaus (approved):

davsclaus (approved):

PR #24473, opened 2026-07-06
CAMEL-23382: camel-ai-tool - create unified AI tool component

zbendhiba:
Step 1: Create the new camel-ai-tool module with AiToolSpec, AiToolRegistry, consumer endpoint, and full tests. No changes to existing modules.

# Description

## Summary

Create the new `camel-ai-tool` module with the unified AI tool abstraction layer:
  - `AiToolSpec` — vendor-neutral tool descriptor (name, description, parameters, tags)
  - `AiToolRegistry` — central registry for discovering and managing AI tools
  - `AiToolConsumer` — endpoint that registers Camel routes as AI-callable tools
  - Full unit test coverage

No changes to existing modules in this PR.

 ## Roadmap
Technical design details : design/aiTool.adoc file

This work is split into several sequential PRs to keep reviews focused:

  - [x] **Step 1** (this PR): Create the new `camel-ai-tool` module with `AiToolSpec`, `AiToolRegistry`, consumer endpoint, and full tests. No changes to existing modules.
  - [ ] **Step 2**: Wire `camel-langchain4j-agent` producer to read tools from `AiToolRegistry` instead of the old cache, with upgrade guide entry.
  - [ ] **Step 3**: Wire `camel-spring-ai-chat` producer to discover tools from `AiToolRegistry`, with upgrade guide entry.
  - [ ] **Step 4**: Wire `camel-langchain4j-tools` producer to read from `AiToolRegistry`, keeping the agentic loop intact.
  - [ ] **Step 5**: Deprecate `camel-langchain4j-tools` consumer side and remove `camel-spring-ai-tools` entirely, with upgrade guide entries for both.
  - [ ] **Step 6**: Add tool support to `camel-openai`. Resolve tools from `AiToolRegistry` and implement the dispatch loop
  (separate Jira issue).


Croway:
### Discussion: passing tool arguments as an exchange variable instead of headers

This PR changes how tool arguments reach the route compared to `camel-langchain4j-tools`: instead of one header per argument, everything is wrapped in a single `AiToolArguments` object under the `AiTool.TOOL_ARGUMENTS` exchange variable. We think this is the right call and want to make the reasoning and consequences explicit before steps 2–5 build on it:

**Behavioral changes**
1. **Behavioral break in steps 2–5**: every existing tool route reading `${header.city}` or using sql `:#param` binding will fail at runtime, not at startup. The upgrade guide entries promised in the roadmap need to spell out the migration pattern very explicitly.
2. **Lost binding convenience**: the documented replacement patterns should be either simple OGNL (`:#${variable.AiToolArguments.parameters[id]}`), or a one-line processor extracting arguments into route-author-named headers (safe, since the names are then trusted).

**Why variables are safer than headers**
- LLM output is untrusted input under Camel's threat model, and headers are exactly the surface behind the CVE-2025-27636 header-injection family (the `allowedParams` filter in `LangChain4jToolsProducer` is a hardening patch over this very risk).
- Headers propagate downstream: a tool route doing `.to("http:...")` or `.to("jms:...")` forwards LLM-controlled values as HTTP headers / JMS properties under default filter strategies. Variables are never auto-propagated and never drive component dispatch.
- Wrapping all arguments under one constant key means even argument *names* can't pollute any namespace.

**Open question**: keep the single wrapper variable (most conservative), or additionally set one variable per *declared* argument (`${variable.city}`) for ergonomics? Declared names are route-author-chosen, so per-argument variables would not reopen the injection surface.

_Claude Code on behalf of @Croway_

@zbendhiba @orpiske @davsclaus @oscerd this is something worth discussing, from a user perspective the header approach is more elegant, `${header.city}` or using sql `:#param` compared to `:#${variable.AiToolArguments.parameters[id]}`

Croway (commented):
Reviewed the new module in depth — the per-context registry and the security posture around argument passing are solid (see separate discussion comment). A handful of findings below, two of which I'd consider pre-merge: the broken SQL doc example and the executor's uncaught body-conversion path. Also one structural question best settled now: the package is `org.apache.camel.component.ai.tools` (plural) while the artifact/scheme are `ai-tool` (singular) — once steps 2–6 wire other modules against these classes, renaming becomes a breaking change.

_Claude Code on behalf of @Croway_

components/camel-ai/camel-ai-tool/src/main/docs/ai-tool-component.adoc
```diff
@@ -0,0 +1,241 @@
...
+[source,java]
+----
+// Define tools with tags
+from("ai-tool:weather?tags=weather,external-api&description=Get weather for a city" +
+    "&parameter.city=string&parameter.city.description=The city name")
+    .to("bean:weatherService");
+
+from("ai-tool:queryUser?tags=users&description=Query database by user ID" +
+    "&parameter.id=integer&parameter.id.description=The user ID&parameter.id.required=true")
+    .to("sql:SELECT name FROM users WHERE id = :#id");
```
Croway: This example won't work: `:#id` binds from headers/body, but tool arguments are in the `AiToolArguments` exchange variable. Either use `:#${variable.AiToolArguments.parameters[id]}` or add a processor step that extracts `args.getInteger("id")` into a header first. Worth fixing since it's the pattern users will copy (the XML and YAML tabs have the same issue).

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+
+        if (exchange.getException() != null) {
+            Exception routeError = exchange.getException();
+            LOG.error("Error executing tool '{}': {}", toolName, routeError.getMessage(), routeError);
+            return new AiToolResult.ExecutionError(
+                    String.format("Error executing tool '%s': %s", toolName, routeError.getMessage()), routeError);
+        }
+
+        String result = exchange.getMessage().getBody(String.class);
+        LOG.debug("Tool '{}' execution completed successfully", toolName);
```
Croway: This sits outside the try/catch, so a `TypeConversionException` (route returns a non-convertible body) propagates to the adapter — contradicting the documented contract that all errors come back as typed `AiToolResult` variants. Suggest wrapping it and returning an `ExecutionError`.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolComponent.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    protected Endpoint createEndpoint(String uri, String remaining, Map<String, Object> parameters) throws Exception {
+        if (ObjectHelper.isEmpty(remaining)) {
+            throw new IllegalArgumentException(
+                    "A toolName must be provided: ai-tool:<toolName>?description=<desc>");
+        }
+
+        final String toolName = StringHelper.before(remaining, "/", remaining);
```
Croway: `ai-tool:myTool/anything` silently drops `/anything`. Since the syntax is `ai-tool:toolName` with no path segments, better to fail on a `/` (or use `remaining` as-is) so typos surface instead of registering a truncated tool name.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tool/AiToolConsumer.java
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            registeredTags = null;
+            registeredInDefaultPool = true;
+            LOG.debug("Registering tool '{}' in default pool (no tags)", toolName);
+            registry.putDefault(registeredSpec);
+        }
+    }
+
+    @Override
+    protected void doStop() throws Exception {
```
Croway, line 101: Route **suspend** doesn't deregister the tool (no `doSuspend` override), so a suspended route stays in the registry and the executor will still dispatch into its processor. Either deregister on suspend or have `AiToolExecutor` check the consumer's state before dispatching — and document whichever is chosen.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolRegistry.java
```diff
@@ -0,0 +1,171 @@
...
+
+    public void put(String tag, AiToolSpec spec) {
+        tools.compute(tag, (k, set) -> {
+            if (set == null) {
+                set = Collections.synchronizedSet(new LinkedHashSet<>());
+            }
+            synchronized (set) {
+                for (AiToolSpec existing : set) {
+                    if (existing.getName().equals(spec.getName()) && existing != spec) {
+                        LOG.warn("Duplicate toolName '{}' under tag '{}' -- the LLM adapter will see both "
```
Croway: Warn-only on duplicate tool names means both specs stay registered, and most LLM APIs (OpenAI included) reject duplicate function names — so the failure surfaces later and far from the cause. Since this registry is the foundation for steps 2–6, consider failing fast at registration, or documenting last-wins semantics.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tool/AiToolConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+ * @since 4.22
+ */
+@Configurer
+@UriParams
+public class AiToolConfiguration implements Cloneable {
+
+    @UriParam(description = "Comma-separated list of tags used to group tools. "
+                            + "Producers filter the registry by these tags to select which tools to expose to the LLM. "
+                            + "When omitted, the tool goes into a default pool available to all producers.")
+    private String tags;
```
Croway, line 41: Missing `@Metadata(label = "consumer")` — the generated catalog JSON shows `"label": ""` for `tags` while `description`/`parameters` have `"consumer"`.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tool/AiToolRegistry.java
```diff
@@ -0,0 +1,171 @@
...
+ * <p>
+ * Each {@link CamelContext} gets its own registry instance, registered as a context plugin. Use
+ * {@link #getOrCreate(CamelContext)} to obtain the instance for a given context.
+ * <p>
+ * Replaces the duplicated {@code CamelToolExecutorCache} singletons from {@code camel-langchain4j-tools} and
+ * {@code camel-spring-ai-tools}.
+ *
+ * @since 4.22
+ */
+public final class AiToolRegistry {
```
Croway, line 40: This class relies on `synchronized` throughout — `synchronized (context)` in `getOrCreate`, `Collections.synchronizedSet`, and `synchronized (set)` blocks in `put`/`getToolsByTag`/`getAllTools`/`getTools`/`getDefaultTools`. Camel supports virtual threads, and the project has been systematically replacing `synchronized` with `ReentrantLock` for virtual-thread compatibility (CAMEL-20199, e.g. #21703) — new code shouldn't reintroduce it. A `ReentrantLock`-guarded `LinkedHashSet` keeps the insertion ordering, or `CopyOnWriteArraySet` would fit this read-heavy registry with no locking at all.

zbendhiba:
> ### Discussion: passing tool arguments as an exchange variable instead of headers
> This PR changes how tool arguments reach the route compared to `camel-langchain4j-tools`: instead of one header per argument, everything is wrapped in a single `AiToolArguments` object under the `AiTool.TOOL_ARGUMENTS` exchange variable. We think this is the right call and want to make the reasoning and consequences explicit before steps 2–5 build on it:
> 
> **Behavioral changes**
> 
> 1. **Behavioral break in steps 2–5**: every existing tool route reading `${header.city}` or using sql `:#param` binding will fail at runtime, not at startup. The upgrade guide entries promised in the roadmap need to spell out the migration pattern very explicitly.
> 2. **Lost binding convenience**: the documented replacement patterns should be either simple OGNL (`:#${variable.AiToolArguments.parameters[id]}`), or a one-line processor extracting arguments into route-author-named headers (safe, since the names are then trusted).
> 
> **Why variables are safer than headers**
> 
> * LLM output is untrusted input under Camel's threat model, and headers are exactly the surface behind the [CVE-2025-27636](https://github.com/advisories/GHSA-2c2h-2855-mf97) header-injection family (the `allowedParams` filter in `LangChain4jToolsProducer` is a hardening patch over this very risk).
> * Headers propagate downstream: a tool route doing `.to("http:...")` or `.to("jms:...")` forwards LLM-controlled values as HTTP headers / JMS properties under default filter strategies. Variables are never auto-propagated and never drive component dispatch.
> * Wrapping all arguments under one constant key means even argument _names_ can't pollute any namespace.
> 
> **Open question**: keep the single wrapper variable (most conservative), or additionally set one variable per _declared_ argument (`${variable.city}`) for ergonomics? Declared names are route-author-chosen, so per-argument variables would not reopen the injection surface.
> 
> _Claude Code on behalf of @Croway_
> 
> @zbendhiba @orpiske @davsclaus @oscerd this is something worth discussing, from a user perspective the header approach is more elegant, `${header.city}` or using sql `:#param` compared to `:#${variable.AiToolArguments.parameters[id]}`

I assume this change. Keep in mind that this is internal to components using the camel ai tool. I've already started implementing for camel langchain4j agent and camel spring ai chat components. There is 0 impact for users. there's no reason to propagate headers as those are the internal LLM calls.

I didn't spent time on langchain4j-tools internal. I marked in the spec this component as to deprecate.

components/camel-ai/camel-ai-tool/src/main/docs/ai-tool-component.adoc
zbendhiba: so I guess your call on headers comes from this. which is a valid thing.

I guess this one should be revised.

zbendhiba: Let me review that. I've deleted the headers for security when we move those to the Executor, however this adds issues for execution

Croway: exactly!

zbendhiba:
> > ### Discussion: passing tool arguments as an exchange variable instead of headers
> > This PR changes how tool arguments reach the route compared to `camel-langchain4j-tools`: instead of one header per argument, everything is wrapped in a single `AiToolArguments` object under the `AiTool.TOOL_ARGUMENTS` exchange variable. We think this is the right call and want to make the reasoning and consequences explicit before steps 2–5 build on it:
> > **Behavioral changes**
> > 
> > 1. **Behavioral break in steps 2–5**: every existing tool route reading `${header.city}` or using sql `:#param` binding will fail at runtime, not at startup. The upgrade guide entries promised in the roadmap need to spell out the migration pattern very explicitly.
> > 2. **Lost binding convenience**: the documented replacement patterns should be either simple OGNL (`:#${variable.AiToolArguments.parameters[id]}`), or a one-line processor extracting arguments into route-author-named headers (safe, since the names are then trusted).
> > 
> > **Why variables are safer than headers**
> > 
> > * LLM output is untrusted input under Camel's threat model, and headers are exactly the surface behind the [CVE-2025-27636](https://github.com/advisories/GHSA-2c2h-2855-mf97) header-injection family (the `allowedParams` filter in `LangChain4jToolsProducer` is a hardening patch over this very risk).
> > * Headers propagate downstream: a tool route doing `.to("http:...")` or `.to("jms:...")` forwards LLM-controlled values as HTTP headers / JMS properties under default filter strategies. Variables are never auto-propagated and never drive component dispatch.
> > * Wrapping all arguments under one constant key means even argument _names_ can't pollute any namespace.
> > 
> > **Open question**: keep the single wrapper variable (most conservative), or additionally set one variable per _declared_ argument (`${variable.city}`) for ergonomics? Declared names are route-author-chosen, so per-argument variables would not reopen the injection surface.
> > _Claude Code on behalf of @Croway_
> > @zbendhiba @orpiske @davsclaus @oscerd this is something worth discussing, from a user perspective the header approach is more elegant, `${header.city}` or using sql `:#param` compared to `:#${variable.AiToolArguments.parameters[id]}`
> 
> I assume this change. Keep in mind that this is internal to components using the camel ai tool. I've already started implementing for camel langchain4j agent and camel spring ai chat components. There is 0 impact for users. there's no reason to propagate headers as those are the internal LLM calls.
> 
> I didn't spent time on langchain4j-tools internal. I marked in the spec this component as to deprecate.

revised, let me look at this again

orpiske:
ACK ... This one is on my TODO list for tomorrow morning

orpiske (approved):
**High level review from the perspective of downstream consumers**

_Claude Code on behalf of Otavio Rodolfo Piske_

### Overall Assessment: Approve (Step 1)

This PR introduces a clean, well-scoped abstraction layer. Step 1 is purely additive — no existing modules are modified. The analysis below focuses on how this new module interacts with downstream bean-factory systems like Forage.

### Compatibility

The `AiToolRegistry` and `AiToolSpec` abstractions operate at a different layer than bean-based agent creation. Agent beans (ChatModel + Memory + RAG + Guardrails) are created and bound to the CamelContext registry via `registry.bind()`. Tools are registered in the separate `AiToolRegistry` plugin. These are complementary — the agent is the "brain," the tools are the "hands." The `?bean=` URI syntax for agent lookup is unaffected.

**Beans still work. No conflicts.**

### Extension Points (positive)

The module exposes useful hooks for downstream consumers:
- `AiToolRegistry.getOrCreate(CamelContext)` — clean programmatic access for registering tools outside of route DSL
- `AiToolSpec` constructor is public — third parties can build and register tool descriptors
- Tag-based discovery maps well to multi-agent configurations

### Concerns for Steps 2-5

1. **Argument passing change**: The shift from exchange headers (`exchange.getMessage().setHeader(name, value)`) to exchange variables (`AiToolArguments`) in Step 2 will be a breaking change for existing tool routes that read parameters via `${header.city}`. This deserves a clear upgrade guide entry when it lands.

2. **Missing SPI for third-party tool sources**: There is no `AiToolProvider` SPI for external tool catalogs (e.g., MCP server catalogs, property-driven tool definitions). All tools must come through either `AiToolConsumer` routes or programmatic `AiToolRegistry.put()`. Consider whether a provider SPI would be valuable for Step 6 or beyond.

3. **URI migration path**: When `langchain4j-tools:` consumer is deprecated (Step 5), users will need to migrate to `ai-tool:`. A smooth deprecation period where both work simultaneously would reduce friction.

### Summary

Step 1 is safe to merge. The architecture is sound and the separation of concerns is correct. The risks noted above are all in future steps and are flagged early so they can be addressed as the roadmap progresses.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiTool.java
```diff
@@ -14,15 +14,22 @@
...
+public final class AiTool {
+
+    public static final String SCHEME = "ai-tool";
 
-/** Picocli tab-completion candidates for {@code --theme}. */
-public class ThemeModeCompletionCandidates implements Iterable<String> {
+    /**
+     * Exchange variable name under which {@link AiToolArguments} is set by the executor.
+     */
+    public static final String TOOL_ARGUMENTS = "AiToolArguments";
```
JiriOndrusek: We should probably use CamelXxx format.  The proper variable name should be `CamelAiToolArguments.`

orpiske:
> ### Discussion: passing tool arguments as an exchange variable instead of headers
> 
> This PR changes how tool arguments reach the route compared to `camel-langchain4j-tools`: instead of one header per argument, everything is wrapped in a single `AiToolArguments` object under the `AiTool.TOOL_ARGUMENTS` exchange variable. We think this is the right call and want to make the reasoning and consequences explicit before steps 2–5 build on it:
> 
> **Behavioral changes**
> 
>     1. **Behavioral break in steps 2–5**: every existing tool route reading `${header.city}` or using sql `:#param` binding will fail at runtime, not at startup. The upgrade guide entries promised in the roadmap need to spell out the migration pattern very explicitly.
> 
>     2. **Lost binding convenience**: the documented replacement patterns should be either simple OGNL (`:#${variable.AiToolArguments.parameters[id]}`), or a one-line processor extracting arguments into route-author-named headers (safe, since the names are then trusted).
> 
> 
> **Why variables are safer than headers**
> 
>     * LLM output is untrusted input under Camel's threat model, and headers are exactly the surface behind the [CVE-2025-27636](https://github.com/advisories/GHSA-2c2h-2855-mf97) header-injection family (the `allowedParams` filter in `LangChain4jToolsProducer` is a hardening patch over this very risk).
> 
>     * Headers propagate downstream: a tool route doing `.to("http:...")` or `.to("jms:...")` forwards LLM-controlled values as HTTP headers / JMS properties under default filter strategies. Variables are never auto-propagated and never drive component dispatch.
> 
>     * Wrapping all arguments under one constant key means even argument _names_ can't pollute any namespace.
> 
> 
> **Open question**: keep the single wrapper variable (most conservative), or additionally set one variable per _declared_ argument (`${variable.city}`) for ergonomics? Declared names are route-author-chosen, so per-argument variables would not reopen the injection surface.
> 
> _Claude Code on behalf of @Croway_
> 
> @zbendhiba @orpiske @davsclaus @oscerd this is something worth discussing, from a user perspective the header approach is more elegant, `${header.city}` or using sql `:#param` compared to `:#${variable.AiToolArguments.parameters[id]}`

I am OK with this change, although: 

* it does concern me a bit that processing the tool arguments get harder. 
* headers propagating downstream may be a two-edged sword if that leads to situations where integrations inadvertently leak credentials/PII/etc. 

Whether to keep a single wrapper or per variable ... I don't have a strong opinion. I think the one variable per argument approach would solve one of my concerns above, but I am unsure how much more complicated it is to implement.

oscerd (commented):
Nice work, Zineb — reviewed mainly from a security and integration-correctness angle.

The security posture is the part I want to call out as well-designed: untrusted LLM tool-call arguments are wrapped under a single constant key into an exchange **variable** (`exchange.setVariable(AiTool.TOOL_ARGUMENTS, ...)`) rather than spread into Exchange headers. That deliberately avoids the raw-name→header mapping behind the `camel-langchain4j-tools` header-injection family (CVE-2025-27636 and follow-ons) — argument *names* can't pollute the header namespace, and variables don't auto-propagate or drive dispatch. `AiToolResult` also carries the note about not returning raw exception text to the LLM. Good hardening. MojoHelper registration, generated files and conventions all check out, and CI is green.

Concurring with @Croway's two pre-merge items:

- The executor's body conversion `exchange.getMessage().getBody(String.class)` sits *after* the try/catch, so a `TypeConversionException` would escape the "all errors returned as a typed `AiToolResult`" contract instead of becoming an `ExecutionError`.
- The flagship doc example binds `:#id` in the SQL, but arguments now live in the `AiToolArguments` variable, so the copy-paste example won't bind at runtime (functional only — it stays parameterized, so no injection concern).

Thanks for taking the hardened approach here.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+                         + "a parameter that is declared as required in the tool specification",
+                        entry.getKey(), toolName);
+                IllegalArgumentException cause = new IllegalArgumentException(
+                        String.format("Missing required argument '%s' for tool '%s'", entry.getKey(), toolName));
+                return new AiToolResult.ArgumentError(cause.getMessage(), cause);
+            }
+        }
+
+        // Defensive copy so callers cannot mutate arguments during route execution
+        Map<String, Object> argsCopy = arguments != null ? new HashMap<>(arguments) : Map.of();
```
oscerd: The warn above says the undeclared argument is being *ignored*, but `argsCopy` copies **all** entries — undeclared arguments included — so they remain accessible to the route via `AiToolArguments`. The generated schema also advertises `additionalProperties: false` to the model.

Two ways to make this consistent:

1. Actually filter undeclared arguments out of `argsCopy` when the tool declares parameters (defense-in-depth: extra parameters an LLM hallucinates — or is prompted into sending — never reach the route), or
2. Keep the pass-through behavior and reword the log message/Javadoc (the design doc says "warned about but not rejected").

Option 1 seems preferable given the schema already tells the model `additionalProperties: false`. Either way, a test asserting the chosen behavior would pin it down — `testExecuteIgnoresUndeclaredArguments` currently passes under both semantics (it never checks whether `extraParam` is present in `AiToolArguments`).

components/camel-ai/camel-ai-tool/src/main/docs/ai-tool-component.adoc
oscerd: This example won't bind as written: camel-sql resolves `:#id` from the body-as-Map, the headers, or an exchange **variable named `id`** (see `SqlHelper.lookupParameter`), but with this design the arguments live only inside the single `AiToolArguments` variable — there is no `id` variable or header, and the body is not the arguments map. (The pattern worked with the old header-based tools components, which is probably where the example stems from.)

Options: use a simple-expression binding (e.g. `:#${variable.AiToolArguments.parameters[id]}` — worth a quick verification), switch the example to a processor/bean that reads `AiToolArguments`, or — if `:#name` ergonomics are a goal for tool routes — consider having the executor additionally expose declared arguments individually. The same applies to the XML tab (line 188) and the YAML tab (line 226).

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolConsumer.java
```diff
@@ -0,0 +1,100 @@
...
+        String jsonSchema = !parameterDefs.isEmpty()
+                ? AiToolParameterHelper.buildJsonSchemaFromDefs(parameterDefs)
+                : null;
+
+        registeredSpec = new AiToolSpec(
+                toolName, configuration.getDescription(), parameterDefs, jsonSchema, this);
+
+        AiToolRegistry registry = AiToolRegistry.getOrCreate(getEndpoint().getCamelContext());
+        String tags = configuration.getTags();
+        if (tags != null && !tags.isBlank()) {
```
oscerd: Edge case: a `tags` value that is non-blank but contains only blank items (e.g. `tags=,`) passes this check, but `splitTags` then returns an empty array — so the loop below registers the tool **nowhere** (neither under a tag nor in the default pool), silently. Similarly, `tags=,research` registers the tool under the empty-string tag `""`.

Suggestion: filter blank entries in `splitTags`, and if nothing remains fall back to the default pool here, matching the documented "when omitted" semantics.

components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tools/AiToolComponent.java
oscerd: Question: this silently drops anything after the first `/` (`ai-tool:foo/bar` → tool name `foo`). Since the syntax is just `ai-tool:toolName`, silently truncating user input seems surprising — is the `/` handling intentional (future-proofing?), or would rejecting / keeping the full `remaining` be clearer? A short comment would help either way.

components/camel-ai/camel-ai-tool/src/main/docs/ai-tool-component.adoc
```diff
@@ -0,0 +1,241 @@
...
+
+// LangChain4j agent uses only weather-tagged tools
+from("direct:chat")
+    .to("langchain4j-agent:assistant?agent=#myAgent&tags=weather");
+----
```
oscerd, line 166: Minor: this section shows the `langchain4j-agent` producer consuming registry tools, but that wiring only lands in Step 2 — a reader of the nightly docs could reasonably expect it to work today. The See Also section already says "future release"; a short NOTE admonition here as well would avoid confusion until Steps 2–4 land.

oscerd (commented):
### Step 1 review — conventions & quality pass

Really nice groundwork: the design doc (`design/aiTool.adoc`) plus the sequential-PR plan make this easy to review, and the Step 1 scoping (no changes to existing modules) is exactly right. Checked against the repo conventions:

- ✅ MojoHelper registration, `camel-ai` module list, parent/BOM/allcomponents, catalog + component/endpoint DSL regeneration, docs symlinks and `nav.adoc` — all present and consistent
- ✅ Only `camel-support` as runtime dependency; no new external dependencies
- ✅ Comprehensive unit tests (including a registry concurrency test); no `Thread.sleep`
- ✅ Passing tool arguments as a single typed exchange **variable** instead of per-name headers avoids header-namespace collisions entirely — good call, and in line with the committer security checklist for consumers
- ✅ `additionalProperties: false` in the generated schema and the sanitization note on `AiToolResult` are thoughtful touches

I left 5 inline comments: 3 small correctness/consistency items (undeclared-argument handling vs. its log message, a `sql` doc example that won't bind with variable-based arguments, a blank-tags edge case) and 2 questions.

One module-wide naming question that doesn't fit a single line: the Java package is `org.apache.camel.component.ai.tools` (plural) while the artifact and scheme are `camel-ai-tool` / `ai-tool` (singular). Since Steps 2–4 will make other modules depend on these classes, and package renames after a release are breaking, it may be worth aligning on one form now (e.g. `org.apache.camel.component.ai.tool`).

None of this is blocking — looking forward to the follow-up steps.

_This review is a conventions/quality pass and does not replace dedicated review tooling or static analysis (e.g. SonarCloud)._

davsclaus:
> exchange.setVariable(AiTool.TOOL_ARGUMENTS, ...))

EIPs and Camel components should frankly use exchange properties. We made variables only for end users to have full namespace for their own variables that dont get mixed up with internal camel state.

gnodet (approved):
Re-reviewed after the latest commits (f9bfe307, adebbe7f). The follow-up changes substantively address prior review feedback:

- Package rename from `ai.tools` (plural) to `ai.tool` (singular)
- `synchronized` replaced with `ReentrantLock`
- Body conversion moved inside try/catch
- `/` rejection in tool names
- Suspend/resume lifecycle support
- Blank-tag edge case handled
- Duplicate tool names now fail-fast
- Doc examples fixed to use bean/header pattern
- `AiToolArguments` exchange variable removed in favor of per-header arguments with Camel-prefix filtering

Security posture is sound: Camel-prefixed header names are rejected (CVE-2025-27636 family), undeclared arguments are filtered out, and `AiToolResult` documents sanitization requirements. Test coverage is thorough (lifecycle, executor edge cases, registry concurrency, parameter parsing, Camel-prefix rejection).

LGTM.

zbendhiba:
Thanks everyone for the reviews and the feedback !
I believe I've addressed most of the feedback. 
I've also analysed if this works in Camel Quarkus.

oscerd (approved):
Re-reviewing after the latest push — this looks ready to me. Both pre-merge blockers are resolved: the executor's `getBody(String.class)` is now inside the try/catch (so a conversion failure becomes a typed `ExecutionError` per the contract), and the flagship doc example was rewritten to a binding that actually works at runtime.

On security: the design deliberately changed from "arguments in an exchange variable" to "arguments as headers, filtered." The new form guards it two ways — undeclared arguments are dropped, and argument names matching `camel` / `org.apache.camel.` (case-insensitive) are rejected, citing the CVE-2025-27636 family. @gnodet already blessed this approach and it's sound; the header form is what makes `${header.x}` / bean / SQL binding natural. One small defense-in-depth note, non-blocking: the undeclared-argument filter is skipped when a tool declares zero parameters (`AiToolExecutor` ~line 100), so for a no-parameter tool extra LLM-supplied names still reach `setHeader` filtered only by the `Camel*` prefix check — the dangerous internal family stays blocked, so this is hardening rather than a hole.

The red CI is unrelated `camel-vertx-websocket` flakiness (a known-flaky module this PR doesn't touch); a re-run should clear it. Nice work.

davsclaus (approved):

zbendhiba:
Many thanks for the latest reviews.
With 3 committer approvals and all blockers addressed, I'll go ahead and merge this.
@oscerd I suggest addressing your non-blocking comment in a separate PR.

zbendhiba:
sorry there were 4 approvals, not 3

oscerd:
> Many thanks for the latest reviews. With 3 committer approvals and all blockers addressed, I'll go ahead and merge this. @oscerd I suggest addressing your non-blocking comment in a separate PR.

Sounds good.

PR #24485, opened 2026-07-07
CAMEL-23928: Add AiServices builder customizer and tool-calling options to AgentConfiguration

davsclaus:
## Summary

- Expose key LangChain4j `AiServices` builder tool-calling options directly on `AgentConfiguration`: `maxToolCallingRoundTrips`, `hallucinatedToolNameStrategy`, `toolExecutionErrorHandler`, `toolArgumentsErrorHandler`, `compensateOnToolErrors`
- Add `aiServicesCustomizer` (`Consumer<AiServices<?>>`) escape hatch for configuring any `AiServices` builder option not directly exposed
- Wire all new options into `AbstractAgent.configureBuilder()`, with the customizer invoked last

This addresses a gap where users needing tool-calling error handling (e.g. handling hallucinated tool names) had to implement `AgentFactory` and reimplement all of `AbstractAgent.configureBuilder()` from scratch.

Follow-up [CAMEL-23929](https://issues.apache.org/jira/browse/CAMEL-23929) will expose the simple scalar options as `@UriParam` on `LangChain4jAgentConfiguration`.

## Test plan

- [x] Unit tests added for all new fluent setters and chaining (`AgentConfigurationTest`)
- [x] All 189 tests pass in `camel-langchain4j-agent-api`
- [x] Downstream `camel-langchain4j-agent` module compiles cleanly

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent-api/src/test/java/org/apache/camel/component/langchain4j/agent/api/AgentConfigurationTest.java
+++ b/components/camel-ai/camel-langchain4j-agent-api/src/test/java/org/apache/camel/component/langchain4j/agent/api/AgentConfigurationTest.java
@@ -19,6 +19,7 @@ package org.apache.camel.component.langchain4j.agent.api;
 import java.io.Serializable;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
 
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.data.message.ToolExecutionResultMessage;
@@ -275,7 +276,7 @@ public class AgentConfigurationTest {
         AgentConfiguration config = new AgentConfiguration();
         assertNull(config.getHallucinatedToolNameStrategy());
 
-        java.util.function.Function<ToolExecutionRequest, ToolExecutionResultMessage> strategy
+        Function<ToolExecutionRequest, ToolExecutionResultMessage> strategy
                 = request -> ToolExecutionResultMessage.from(request, "Unknown tool: " + request.name());
         AgentConfiguration result = config.withHallucinatedToolNameStrategy(strategy);
 
--- a/components/camel-ai/camel-langchain4j-agent-api/src/test/java/org/apache/camel/component/langchain4j/agent/api/AgentConfigurationTest.java
+++ b/components/camel-ai/camel-langchain4j-agent-api/src/test/java/org/apache/camel/component/langchain4j/agent/api/AgentConfigurationTest.java
@@ -18,11 +18,18 @@ package org.apache.camel.component.langchain4j.agent.api;
 
 import java.io.Serializable;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.service.tool.ToolArgumentsErrorHandler;
+import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class AgentConfigurationTest {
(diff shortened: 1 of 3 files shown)


components/camel-ai/camel-langchain4j-agent-api/src/test/java/org/apache/camel/component/langchain4j/agent/api/AgentConfigurationTest.java
```diff
@@ -249,4 +256,97 @@ public void testWithOutputGuardrailClassesArray_WithInvalidClasses() {
...
+    @Test
+    public void testMaxToolCallingRoundTrips() {
+        AgentConfiguration config = new AgentConfiguration();
+        assertEquals(0, config.getMaxToolCallingRoundTrips());
+
+        AgentConfiguration result = config.withMaxToolCallingRoundTrips(10);
+
+        assertSame(config, result);
+        assertEquals(10, config.getMaxToolCallingRoundTrips());
+    }
```
davsclaus, line 272: Nit: FQCN usage — the project import style rule requires an `import` statement rather than inline qualification. Add `import java.util.function.Function;` to the imports and use the simple name here.

```suggestion
        Function<ToolExecutionRequest, ToolExecutionResultMessage> strategy
```

davsclaus (commented):
Clean, well-structured PR that follows existing patterns. The `aiServicesCustomizer` escape hatch and tool-calling options are a good addition to avoid users having to reimplement `configureBuilder()` from scratch.

One minor style nit inline.

orpiske (approved):

zbendhiba:
I'm analysing this, please hold on before merge.

zbendhiba (changes requested):
Hey @davsclaus , thanks for the PR!

I'm happy you opened this. These are real concerns from users, and I want to address them. I'll definitely work on some of these for the camel-ai-tool (CAMEL-23382).

That said, I looked at how these options work in the context of our current agent component, and I have a few concerns.

Our ToolExecutor already catches errors In LangChain4jAgentProducer, our Camel ToolExecutor catches all exceptions and returns them as strings to the LLM. LangChain4j never sees the exception. So toolExecutionErrorHandler and toolArgumentsErrorHandler would never fire for Camel route tools.
They could work for customTools (raw @Tool objects). But we need integration tests to prove it.

Error handling defaults are flipping in langchain4j 2.0
Worth noting: langchain4j is changing the default behavior of both error handlers in 2.0.
- ToolExecutionErrorHandler currently returns the exception message to the LLM. In 2.0, it will throw instead. The current default is actually a security concern because it can leak internal data to the model.
- ToolArgumentsErrorHandler currently throws. In 2.0, it will return a text message to the LLM.
So the defaults are swapping. If we expose these options now without tests, we'll also need to handle the breaking change when we bump to langchain4j 2.0. One more reason to wait and get this right.

compensateOnToolErrors can't work with route tools
This feature relies on @CompensateFor annotations on Java methods. Our tools are Camel routes, not annotated methods. So this option has no effect in the Camel tool model.

Again, it could work with customTools. But we need an integration test.

aiServicesCustomizer vs agent=#myBean

Camel creates the AiServices and manages the builder. If a user wants full control, they can build their own Agent and inject it via agent=#myBean. That already bypasses configureBuilder() entirely.

The customizer runs last and can silently undo what configureBuilder() sets up. I'd rather point users to the existing agent bean injection path. Probably we need to sync on architecture design approval for this.
 
What I need before we can merge

Integration tests covering both Camel route tools and annotated @Tool objects for each option:
  - maxToolCallingRoundTrips (this one I'm pretty confident works)
  - toolExecutionErrorHandler with Camel route tools. Does it actually fire?
  - toolArgumentsErrorHandler. Same question.
  - hallucinatedToolNameStrategy with our dynamic ToolProvider
  - compensateOnToolErrors without @CompensateFor annotations
  - aiServicesCustomizer not breaking the rest of the configuration

If the tests show these options don't work with Camel route tools, I'd rather wait. CAMEL-23382 (camel-ai-tool) is about to land and will
replace the Camel tool executor. Anything touching that area now will be rewritten in a few days. Better to revisit on top of the new tool  model.

maxToolCallingRoundTrips is the one option I'm confident about. We could scope this PR to that and move the rest to a follow-up after camel-ai-tool lands.

davsclaus:
yeah lets park this

atiaomar1978-hub:
**Question: why was this PR closed, and is there a replacement?**

_AI-generated comment on behalf of Omar Atie_

PR #24485 was parked after @zbendhiba's review (integration tests needed; wait for `camel-ai-tool`), and the branch was closed on Aug 13. However, the **CAMEL-23928 changes were already merged into `main` via PR #24492** (Jul 7) — `AgentConfiguration` tool-calling options and `aiServicesCustomizer` are on `main` today.

**`camel-ai-tool` does not replace this PR** — it replaces the old tool execution model (`langchain4j-tools` / `spring-ai-tools`) with `ai-tool:` + `AiToolRegistry` + `AiToolExecutor`. The #24485 options configure the LangChain4j **agent** (`AiServices` builder); `camel-ai-tool` defines and runs **route tools**. They work together.

Gaps identified in review were addressed separately:
- **CAMEL-23944** — route tool errors now propagate so `ToolExecutionErrorHandler` / `compensateOnToolErrors` can fire
- **CAMEL-23929** (open) — `@UriParam` exposure for scalar options like `maxToolCallingRoundTrips`

Remaining limitations vs the original #24485 intent:
- `compensateOnToolErrors` still only applies to annotated `@Tool` methods, not Camel route tools
- `toolArgumentsErrorHandler` may not fire for route-tool argument validation (errors returned as strings)

Closing #24485 makes sense as housekeeping; the functionality lives on `main` via #24492 + follow-ups. Anyone needing URI-level config should track **CAMEL-23929**.

PR #24498, opened 2026-07-07
CAMEL-23850: camel-jbang route-diagram and route-topology can dump from source files

ammachado:
# Description

`route-diagram` and `route-topology` now accept a set of Camel route source files instead of only a running integration's name/pid, so diagrams and inter-route topology can be generated at source time without starting the application. `camel.main.dumpRoutes=json` additionally writes a `route-topology.json` alongside the per-route structure files when topology dumping is enabled.

Follow-up fixes included in this PR:
- `DefaultDumpRoutesStrategy` always writes `route-topology.json` (with empty nodes/edges if there are no routes), so its presence reliably signals the dump completed instead of being ambiguous with a failed/timed-out dump.
- `CamelRouteDiagramAction` waits for that `route-topology.json` marker (written last) with a freshness check against boot start, instead of returning as soon as any file appears; fixes a race that could silently merge a partial multi-file batch and guards against a stale file from a previous `--watch` iteration.
- `CamelRouteTopologyAction` gates `--metric` on `pid > 0` like `CamelRouteDiagramAction` already does, so source-file dumps no longer render a bogus empty metrics line.
- Both actions give a more actionable message when a dump never completes, instead of a bare "No routes found".
- Corrected `renderSourceToFile` javadoc (it does not actually skip PID lookup).
- Added tests for the PID-vs-source-file dispatch boundary, the batch-completion race, stale-marker handling, `topologyExternal=false`, and the always-write empty-topology behavior.

Addressed review feedback:
- `CamelRouteDiagramAction#readRoutesFromFolder` now returns `null` (with an actionable error message) instead of merging whatever partial per-file JSON exists, if the `route-topology.json` marker never becomes fresh within the wait timeout; previously a dump that never completed would still be merged and rendered as if it were complete. Added a regression test for this path.
- Extracted the duplicated transient-boot source-dump logic (file validation, work-dir setup, `Run.runTransform`) from `CamelRouteDiagramAction` and `CamelRouteTopologyAction` into a shared `ActionBaseCommand#dumpRoutesFromSource` helper that returns the work dir and boot-start instant via a `SourceDump` record.
- Dropped the misleading `volatile` qualifier on the single-threaded `pid` field in both actions.
- Annotated the picocli completion-only `filePaths` field with `@SuppressWarnings("unused")` (it only exists so picocli offers file-path completion; values are consumed into `files`) to silence static-analysis noise.
- Kept plain `Thread.sleep` in `CamelRouteDiagramActionMultiFileTest`'s background writer threads (with a clarifying comment that they simulate delayed dump-file writes), since the no-`Thread.sleep` guideline targets test assertion polling, not arbitrary timing in helper threads; the `InterruptedException` is handled and the now-unused imports removed.

# Target

# Tracking

# AI-assisted contributions


davsclaus (commented):
Nice work on this PR — the completion-marker design and the test coverage are solid. A few observations below, none blocking.

### Findings

**1. `await().pollDelay(...).until(() -> true)` is a disguised `Thread.sleep` (minor)**

In `CamelRouteDiagramActionMultiFileTest`, the two `await().pollDelay(300, MILLISECONDS).atMost(5, SECONDS).until(() -> true)` calls inside helper `Thread` runners are effectively `Thread.sleep(300)` — the predicate is always true, so `pollDelay` does all the work. Since these are inside background threads simulating delayed file writes (not the test's main assertion path), using `Thread.sleep(300)` directly would be more readable and honest. The Awaitility rule targets test assertion polling, not arbitrary delays in background threads.

**2. `volatile long pid` in `CamelRouteTopologyAction` is unnecessary (cosmetic)**

The `pid` field is only accessed from `doCall()` on the same thread. The `volatile` qualifier is harmless but misleading — it suggests concurrent access that doesn't exist.

**3. Duplicated `doCallSource()` boot logic (maintainability)**

Both `CamelRouteDiagramAction.doCallSource()` and `CamelRouteTopologyAction.doCallSource()` implement nearly identical file-validation, work-directory setup, and transient-Camel-boot code (~30 lines each), differing only in what they read from the dump output. A shared helper would reduce the maintenance surface.

**4. `filePaths` field declared but never read (static analysis)**

The `Path[] filePaths` field exists solely for picocli tab completion. SonarCloud may flag it as unused — a `@SuppressWarnings("unused")` or brief Javadoc would prevent noise.

### What's well done

- Thorough test coverage (4 new test classes) covering dispatch, multi-file merge, marker timing, stale-marker handling, and timeout
- Completion-marker design (`route-topology.json` always written last, even with empty arrays) cleanly solves the race condition
- Backward-compatible single-argument mode
- Upgrade guide and command reference documentation included
- Clear, actionable error messages for failure paths

_This review does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud)._

dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/action/CamelRouteDiagramActionMultiFileTest.java
```diff
@@ -0,0 +1,175 @@
...
+        // DefaultDumpRoutesStrategy) land shortly after; if readRoutesFromFolder returned as soon as any file
+        // appeared (the pre-fix behavior), it would race ahead and merge only route1, silently dropping route2
+        Thread writer = new Thread(() -> {
+            await().pollDelay(300, TimeUnit.MILLISECONDS).atMost(5, TimeUnit.SECONDS).until(() -> true);
+            try {
+                Files.writeString(tempDir.resolve("route2.json"), """
+                        {"routes":[{"routeId":"route2","code":[{"type":"from","code":"direct:linked","level":0}]}]}
+                        """);
+                Files.writeString(tempDir.resolve("route-topology.json"), """
+                        {"nodes":[],"edges":[]}
```
davsclaus, line 108: Minor: `await().pollDelay(300, MILLISECONDS).atMost(5, SECONDS).until(() -> true)` is effectively `Thread.sleep(300)` — the predicate is always true so `pollDelay` does all the work. Since this is inside a helper `Thread` simulating a delayed file write (not the test's main assertion path), a plain `Thread.sleep(300)` would be more readable. The Awaitility-over-`Thread.sleep` rule targets test assertion polling, not arbitrary timing delays in background threads.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/action/CamelRouteTopologyAction.java
```diff
@@ -52,11 +58,17 @@
                  "  camel cmd route-topology --json",
                  "  camel cmd route-topology --theme=unicode",
                  "  camel cmd route-topology --theme=dark",
-                 "  camel cmd route-topology --theme=dark --output=topology.png" })
+                 "  camel cmd route-topology --theme=dark --output=topology.png",
+                 "  camel cmd route-topology hello.yaml bye.yaml" })
 public class CamelRouteTopologyAction extends ActionBaseCommand {
```
davsclaus, line 60: Cosmetic: `pid` is only accessed from `doCall()` on the same thread — `volatile` is unnecessary and suggests concurrent access that doesn't exist. Consider dropping the qualifier.

davsclaus (approved):

PR #24506, opened 2026-07-08
CAMEL-23936: Fix logHttpActivity losing Content-Encoding on gzip responses

davsclaus:
## Summary

When `logHttpActivity` is enabled and the HTTP server returns a gzip-encoded response, the response body is not decompressed, causing downstream processing (e.g., Jackson JSON parsing) to fail with `JsonParseException: Illegal character (CTRL-CHAR, code 31)`.

**Root cause:** In `LoggingHttpActivityListener`, when the response entity's stream is consumed for logging and a replacement `ByteArrayEntity` is created, the original `Content-Encoding` (e.g., `gzip`) was not passed to the new entity. This caused Apache HttpClient's `ContentCompressionExec` to skip decompression, leaving the raw gzip bytes in the response body.

**Fix:** Pass the original content encoding as the third argument to the `ByteArrayEntity` constructor, preserving the encoding metadata for downstream decompression.

- `LoggingHttpActivityListener.java:208` — `new ByteArrayEntity(arr, ct)` → `new ByteArrayEntity(arr, ct, ce)`

## Test plan

- [x] `HttpLoggingActivityGzipTest` extends `HttpCompressionTest` — verifies gzip-encoded responses are correctly decompressed when `logHttpActivity` is enabled
- [x] Inherited tests from `HttpCompressionTest` also pass with logging enabled (bonus coverage)
- [x] Verified the test fails without the fix (raw gzip bytes reach the body)
- [x] All existing `camel-http` tests pass

### DIFF
--- a/components/camel-http/src/main/java/org/apache/camel/component/http/LoggingHttpActivityListener.java
+++ b/components/camel-http/src/main/java/org/apache/camel/component/http/LoggingHttpActivityListener.java
@@ -205,7 +205,7 @@ public class LoggingHttpActivityListener extends ServiceSupport implements Camel
                                 byte[] arr = bos.toByteArray();
                                 // ByteArrayEntity close() is a NOOP. Additionally
                                 // the stream must be closed by client (request/response in our case).
-                                e = new ByteArrayEntity(arr, ct); // NOSONAR
+                                e = new ByteArrayEntity(arr, ct, ce); // NOSONAR
                                 if (request instanceof HttpEntityContainer ec) {
                                     ec.setEntity(e);
                                 } else if (response instanceof HttpEntityContainer ec) {
--- a/components/camel-http/src/test/java/org/apache/camel/component/http/HttpCompressionTest.java
+++ b/components/camel-http/src/test/java/org/apache/camel/component/http/HttpCompressionTest.java
@@ -59,7 +59,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class HttpCompressionTest extends BaseHttpTest {
 
-    private HttpServer localServer;
+    protected HttpServer localServer;
 
     @Override
     public void setupResources() throws Exception {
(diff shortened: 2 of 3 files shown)


davsclaus (commented):
Clean, well-targeted bug fix. The one-line change at `LoggingHttpActivityListener.java:208` correctly preserves the `Content-Encoding` when re-creating the `ByteArrayEntity` after logging consumes the entity stream. Without this, `ContentCompressionExec` skips gzip decompression and downstream processing fails.

The `ce` variable correctly holds the original content encoding in both request and response paths, and the `ByteArrayEntity(byte[], ContentType, String)` constructor handles `null` safely for the non-compressed case.

Test coverage is good — `HttpLoggingActivityGzipTest` exercises the exact bug scenario with a local gzip-compressing HTTP server.

**Minor suggestion (non-blocking):** The new test duplicates ~130 lines of server infrastructure (inner interceptor classes and entity wrappers) from `HttpCompressionTest`. It could extend `HttpCompressionTest` instead, overriding only `createCamelContext()` and adding the logging-specific test method. But the current approach is consistent with the self-contained test pattern used elsewhere in the module.

oscerd (approved):
Thanks for the fix, Claus — reviewed against `main` and it looks correct.

- The one-line change in `LoggingHttpActivityListener` (`new ByteArrayEntity(arr, ct)` → `new ByteArrayEntity(arr, ct, ce)`) correctly re-attaches the `Content-Encoding` to the re-created entity. `arr` holds the raw gzip bytes and the separate local `data` decompression used only for logging is untouched, so there is no double-decode and the non-logging path is unaffected (the listener runs only when `logHttpActivity=true`).
- Good regression coverage: `HttpLoggingActivityGzipTest` extends `HttpCompressionTest`, no `Thread.sleep`, and CI is green.
- Commit convention and attribution/sign-off trailers are correct.

LGTM.

PR #24507, opened 2026-07-08
[4.18.x] CAMEL-23936: Fix logHttpActivity losing Content-Encoding on gzip responses

davsclaus:
## Summary

Backport of #24506 to `camel-4.18.x`.

When `logHttpActivity` is enabled and the HTTP server returns a gzip-encoded response, the response body is not decompressed, causing downstream processing (e.g., Jackson JSON parsing) to fail with `JsonParseException: Illegal character (CTRL-CHAR, code 31)`.

**Fix:** Pass the original content encoding to the `ByteArrayEntity` constructor when replacing the consumed response entity in `LoggingHttpActivityListener`.

Note: On 4.18.x the `ce` variable is a `Header` (not `String` as on `main`), so the fix extracts `ce.getValue()`.

## Test plan

- [x] New `HttpLoggingActivityGzipTest` — verifies gzip-encoded responses are correctly decompressed when `logHttpActivity` is enabled
- [x] All existing `camel-http` tests pass

### DIFF
--- a/components/camel-http/src/main/java/org/apache/camel/component/http/LoggingHttpActivityListener.java
+++ b/components/camel-http/src/main/java/org/apache/camel/component/http/LoggingHttpActivityListener.java
@@ -204,7 +204,7 @@ public class LoggingHttpActivityListener extends ServiceSupport implements Camel
                                 byte[] arr = bos.toByteArray();
                                 // ByteArrayEntity close() is a NOOP. Additionally
                                 // the stream must be closed by client (request/response in our case).
-                                e = new ByteArrayEntity(arr, ct); // NOSONAR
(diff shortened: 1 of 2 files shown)


oscerd (approved):
Backport LGTM. The fix is faithfully carried to `camel-4.18.x`: since `ce` is a `Header` on this branch (vs a `String` on main), the adaptation `new ByteArrayEntity(arr, ct, ce != null ? ce.getValue() : null)` yields the same effective content-encoding. CI is green.

One minor, non-blocking note: this cherry-picks only the first commit of #24506, so it carries the self-contained version of the test rather than main's second-commit refactor that extends `HttpCompressionTest`. That is fine (self-contained tests are an accepted pattern here) — just flagging in case you want parity.

PR #24508, opened 2026-07-08
CAMEL-21438: Harden container-exception retry logic and isolate IggyConsumerOffsetIT

ammachado:
# Description

Two related fixes to integration-test reliability, tracked under CAMEL-21438.

**1. Fix `isRetryableContainerException` short-circuiting on `NoHttpResponseException`**

`TestServiceUtil.isRetryableContainerException` walked the exception cause chain and, in the same loop pass, checked for both `NoHttpResponseException` and `ContainerFetchException`/`ContainerLaunchException`. Because testcontainers always wraps docker-java failures as the *cause* of a `ContainerFetchException`/`ContainerLaunchException`, the outer container-exception check matched first and returned `true` before the loop ever reached the nested `NoHttpResponseException`. As a result a dead or unresponsive Docker daemon was retried instead of failing fast. The fix walks the chain for `NoHttpResponseException` in a separate pass first, so that condition wins. The test is updated to wrap `NoHttpResponseException` inside a `ContainerLaunchException`, matching how testcontainers actually throws it.

**2. Isolate `IggyConsumerOffsetIT` on a dedicated topic/stream**

`IggyConsumerOffsetIT` reused the shared `TOPIC`/`STREAM`/`CONSUMER_GROUP` constants that other IT classes share through the same static `IggyService`/client, which allowed cross-test interference. It now uses dedicated `OFFSET_TOPIC`/`OFFSET_STREAM`/`OFFSET_CONSUMER_GROUP` names and asserts the exact ordered message bodies received after each batch, confirming the offset-based consumption behaves correctly.

# Target

# Tracking

# AI-assisted contributions

---

_AI-assisted with Claude Code on behalf of ammachado._

### DIFF
--- a/components/camel-iggy/src/test/java/org/apache/camel/component/iggy/IggyConsumerOffsetIT.java
+++ b/components/camel-iggy/src/test/java/org/apache/camel/component/iggy/IggyConsumerOffsetIT.java
@@ -16,15 +16,22 @@
  */
 package org.apache.camel.component.iggy;
 
+import java.util.stream.IntStream;
+
 import org.apache.camel.RoutesBuilder;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
 @DisabledIfSystemProperty(named = "ci.env.name", matches = ".*",
                           disabledReason = "Iggy 0.6.0+ requires io_uring which is not available on CI environments")
 public class IggyConsumerOffsetIT extends IggyTestBase {
+
+    private static final String OFFSET_TOPIC = "test-topic-offset";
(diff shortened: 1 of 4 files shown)


davsclaus (approved):
Reviewed against project rules and conventions — this is a focused, well-tested reliability fix.

**TestServiceUtil fix:** The two-pass approach in `isRetryableContainerException` correctly resolves the short-circuit bug where `NoHttpResponseException` wrapped inside `ContainerLaunchException` was being retried instead of failing fast. The string-based class check avoids an unnecessary compile-time dependency on `docker-java-zerodep`. The new test correctly models real-world testcontainers wrapping.

**Iggy test isolation:** Dedicated `OFFSET_TOPIC`/`OFFSET_STREAM`/`OFFSET_CONSUMER_GROUP` constants eliminate shared-state interference. The `sendMessage(stream, topic, message)` overload in `IggyTestBase` is backward-compatible. The added exact-order body assertions are a meaningful improvement over count-only checks.

No rule violations, no scope drift, no regressions visible in the diff.

PR #24526, opened 2026-07-08
CAMEL-19548: Replace Thread.sleep with Awaitility in camel-zookeeper-master tests

tmielke:
Replaced `Thread.sleep(1000)` with Awaitility-based synchronization in `MasterEndpointFailoverIT` to eliminate flakiness. The test now waits for actual ZooKeeper master/standby state transitions instead of using fixed delays.

Made with help from AI tools.

### DIFF
--- a/components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
+++ b/components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
@@ -16,14 +16,17 @@
  */
 package org.apache.camel.component.zookeepermaster.integration;
 
+import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.camel.CamelContext;
+import org.apache.camel.Consumer;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.component.seda.SedaComponent;
 import org.apache.camel.component.zookeepermaster.CuratorFactoryBean;
+import org.apache.camel.component.zookeepermaster.MasterConsumer;
 import org.apache.camel.impl.DefaultCamelContext;
 import org.apache.camel.support.SimpleRegistry;
 import org.apache.camel.test.infra.zookeeper.services.ZooKeeperService;
@@ -36,11 +39,15 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.awaitility.Awaitility.await;
+
 public class MasterEndpointFailoverIT {
     @RegisterExtension
     static ZooKeeperService service = ZooKeeperServiceFactory.createService();
 
     private static final transient Logger LOG = LoggerFactory.getLogger(MasterEndpointFailoverIT.class);
+
+    private static final Duration FAILOVER_TIMEOUT = Duration.ofSeconds(5);
 
     protected ProducerTemplate template;
     protected CamelContext producerContext;
@@ -94,7 +101,7 @@ public class MasterEndpointFailoverIT {
             }
         });
         consumerContext2.addComponent("seda", sedaComponent);
-        // Need to start at less one consumerContext to enable the seda queue for producerContext
+        // Need to start at least one consumerContext to enable the seda queue for producerContext
         producerContext.start();
         consumerContext1.start();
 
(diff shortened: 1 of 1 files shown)


components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
```diff
@@ -36,12 +39,16 @@
...
     @RegisterExtension
     static ZooKeeperService service = ZooKeeperServiceFactory.createService();
 
     private static final transient Logger LOG = LoggerFactory.getLogger(MasterEndpointFailoverIT.class);
 
+    private static final Duration FAILOVER_TIMEOUT = Duration.ofSeconds(5);
+
     protected ProducerTemplate template;
     protected CamelContext producerContext;
     protected CamelContext consumerContext1;
```
davsclaus, line 54: The `FAILOVER_TIMEOUT` of 5 seconds may be tight on heavily loaded CI nodes. Consider bumping to 10–15 seconds to avoid flakiness — the project's own guidelines use `atMost(20, TimeUnit.SECONDS)` as an example.

```suggestion
    private static final Duration FAILOVER_TIMEOUT = Duration.ofSeconds(15);
```

davsclaus (commented):
Thanks for the contribution @tmielke — replacing `Thread.sleep` with Awaitility is the right approach and aligns with project conventions.

A few minor observations:

- The Javadoc blocks on the three new helper methods use `/* */` (single-star) instead of `/** */`, so the `@param`/`@return`/`@throws` tags won't be picked up by tooling. That said, for test-internal `protected` helpers, the Javadoc is heavier than typical Camel test code — consider removing the doc blocks entirely or keeping a one-liner at most.

- The 5-second `FAILOVER_TIMEOUT` may be tight on heavily loaded CI nodes. The project's own guidelines suggest `atMost(20, TimeUnit.SECONDS)` as an example. Consider bumping to 10–15 seconds for safety — it's still far better than the old fixed 1-second sleep.

- The `// Idempotent re-start for test flow clarity` comment on line 121 draws attention to something that isn't confusing — consider removing it.

- Minor: `getMasterConsumer()` could use Java 17+ pattern matching (`if (consumer instanceof MasterConsumer mc)`) instead of the classic cast, but this is purely stylistic.

Overall this looks good. Nice cleanup!

---

components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
```diff
@@ -113,22 +120,67 @@ public void afterRun() {
...
-        consumerContext1.start();
+        consumerContext1.start(); // Idempotent re-start for test flow clarity
+        awaitMaster(consumerContext1);
         assertMessageReceived(result1Endpoint, result2Endpoint);
 
         LOG.info("Starting consumerContext2");
         consumerContext2.start();
+        awaitMaster(consumerContext1);
+        awaitStandby(consumerContext2);
         assertMessageReceivedLoop(result1Endpoint, result2Endpoint, 3);
```
davsclaus, line 131: This uses `/* */` (single-star) instead of `/** */`, so the `@param` tag won't be processed by Javadoc tooling. That said, for a test-internal helper, the doc block could simply be removed.

components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
```diff
@@ -113,22 +120,67 @@ public void afterRun() {
...
+                    MasterConsumer masterConsumer = getMasterConsumer(context);
+                    return masterConsumer.isMaster() && masterConsumer.isConnected();
+                });
+    }
+
+    /*
+     * Wait fixed time for the master consumer to be the standby and connected.
+     * @param context the CamelContext
+     */
+    protected void awaitStandby(CamelContext context) {
```
davsclaus, line 155: Minor: could use Java 17+ pattern matching to simplify:

```suggestion
    protected MasterConsumer getMasterConsumer(CamelContext context) {
        if (context.getRoutes().isEmpty()) {
            throw new IllegalStateException("No routes registered in context");
        }
        Consumer consumer = context.getRoutes().get(0).getConsumer();
        if (consumer instanceof MasterConsumer mc) {
            return mc;
        }
        throw new IllegalStateException("Expected MasterConsumer but got: " + consumer.getClass());
    }
```

components/camel-zookeeper-master/src/test/java/org/apache/camel/component/zookeepermaster/integration/MasterEndpointFailoverIT.java
tmielke: Zookeper service runs embedded inside the same JVM so master election will be very fast and the 5 seconds timeout should be sufficient, even on busy CI nodes.

oscerd (approved):
Thanks Torsten. Clean migration off `Thread.sleep` in `MasterEndpointFailoverIT`:

- The `Thread.sleep(1000)` is removed and the failover waits now use `await().atMost(FAILOVER_TIMEOUT).until(...)` with an explicit 5s timeout and real predicates — no busy-wait, and I confirmed no `Thread.sleep` remains anywhere in the file.
- CI is green.

LGTM.

PR #24527, opened 2026-07-08
CAMEL-23768: camel-keycloak - select the JWKS verification key by the token kid

oscerd:
Fixes [CAMEL-23768](https://issues.apache.org/jira/browse/CAMEL-23768).

## Problem

`KeycloakPublicKeyResolver` ignored the JWT header `kid` and returned the **first** key in the JWKS, and `KeycloakSecurityProcessor` never passed the kid through (`getPublicKey(null)`). During key rotation (multiple keys present) this could pick the wrong key and reject an otherwise-valid token. The token is still cryptographically verified against a real key, so this is a **correctness/availability** matter, not a bypass.

## Change

- `KeycloakSecurityHelper.extractKeyId(token)` — pulls the `kid` from the JWS header using Keycloak's own `JWSInput` (existing dependency); returns `null` if absent/unparseable.
- `KeycloakSecurityProcessor` — extracts the kid and passes it to `resolver.getPublicKey(kid)`.
- `KeycloakPublicKeyResolver.getPublicKey` — forces one refresh when the kid is unknown (to pick up rotated keys), then delegates to a new package-visible `selectKey` that **fails closed** (throws) when a kid is given but no key matches. The first-key fallback is kept only when the token carries no kid.

## Tests

- `KeycloakPublicKeyResolverTest` — matching kid selects that key (not the first); unknown kid fails closed; null kid falls back to the first key (JWKS built from generated RSA keys, no container).
- `KeycloakSecurityHelperTest` — `extractKeyId` returns the header kid / `null` for a malformed token.
- Full reactor build (`mvn clean install -DskipTests`, 1871 modules) green; no generated-file impact (plain Java).

## Compatibility / backport

Correctness fix, `fixVersions` 4.18.4 / 4.22.0 (camel-keycloak does not exist on the 4.14.x line). Default behaviour for a single-key JWKS is unchanged.

---

### DIFF
--- a/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakPublicKeyResolver.java
+++ b/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakPublicKeyResolver.java
@@ -57,7 +57,8 @@ public class KeycloakPublicKeyResolver {
     /**
      * Gets the public key for verifying JWT signatures. Keys are cached and refreshed periodically.
      *
-     * @param  kid         the key ID from the JWT header (optional, uses first key if null)
+     * @param  kid         the key ID from the JWT header; the matching key is selected, and resolution fails closed
+     *                     when a kid is given but no key matches (a {@code null} kid falls back to the first key)
      * @return             the public key
      * @throws IOException if fetching keys fails
      */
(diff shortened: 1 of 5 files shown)


davsclaus:
there is a merge conflict

davsclaus (approved):
## Review: CAMEL-23768 — Select JWKS verification key by token kid

**No blocking issues.** Clean correctness fix for a real key-rotation problem.

### What was checked

- **Git history:** `KeycloakPublicKeyResolver` introduced in CAMEL-22854 (c1ed776) by the same author. The `getPublicKey(null)` call has been there since inception — this fixes an original oversight, not reverting intentional behavior.
- **Correctness:** `selectKey(kid)` throws `IOException` when a kid is given but no key matches. Right security posture: never verify a token against an unrelated key during rotation.
- **Backward compatibility:** When the token carries no `kid` (null), the first-key fallback is preserved. Single-key JWKS deployments are unaffected.
- **Tests:** Good coverage of the three key scenarios (matching kid, unknown kid fails closed, null kid fallback). Tests build their own JWKS from generated RSA keys — no container needed.
- **AI attribution:** Present in both the PR body and the commit trailer.

- `getPublicKey()` can call `refreshKeys()` twice in quick succession (cache expiry + unknown kid). Harmless (idempotent, synchronized) but costs an extra HTTP round-trip in that edge case.
- `parseJwks` widened from `private` to package-visible for testability. Common pattern.
- Broad `Exception` catch in `extractKeyId` is pragmatic since `JWSInput` can throw various exceptions from untrusted input, and a null return is safe because `TokenVerifier` rejects bad tokens downstream.

oscerd:
Rebased onto `main` and resolved the conflict. It was with CAMEL-23875 (optional audience validation), which added `@Test` methods at the same spot in `KeycloakSecurityHelperTest`; the two additions are independent, so I kept both — the audience tests and the new `extractKeyId` tests now coexist. The production files (`KeycloakSecurityHelper`, `KeycloakSecurityProcessor`) merged automatically (kid-selection and audience validation are orthogonal). All camel-keycloak unit tests pass (HelperTest 23, ResolverTest 3, ProcessorTest 9) and a full reactor build is green. CI is re-running.

PR #24541, opened 2026-07-08
CAMEL-23971: fix ordered properties containsKey support

ammachado:
# Description

Fixes `OrderedProperties` containment support by delegating `containsKey`, `containsValue`, and legacy `contains` to the backing `LinkedHashMap` in `BaseOrderedProperties`, and adds regression coverage for those semantics plus remove behavior.

`BaseOrderedProperties` moved storage into a backing map but never overrode the inherited `Hashtable` containment methods, so property presence checks could return false even when the ordered properties instance actually contained the key.

Additionally, `CamelCaseOrderedProperties` overrides `get`/`getProperty` to look up keys as-is, then `dashToCamelCase`, then `camelCaseToDash`, but left `containsKey` with the plain exact-match behavior. This override now makes `containsKey` apply the same camelCase/dash fallback so it stays consistent with `get`/`getProperty`. Only `containsKey` needs this; values are never transformed, so `containsValue`/`contains` remain the plain base-class delegation.

The Camel JBang export path in `ExportCamelMain` uses `containsKey(...)` for profile-based property presence checks, so explicitly configured keys are treated as present even when their value is empty. With the `CamelCaseOrderedProperties.containsKey` override, dash-style keys (e.g. `camel.management.metrics-enabled`) are still matched.

All `BaseOrderedProperties` descendants were reviewed: `OrderedProperties` (no read overrides) and `OrderedLocationProperties` (only `clear`/`remove` overrides, no key transformation on read) need no containment changes.

_AI-generated by Codex/Claude Code on behalf of ammachado._

# Target

- [x] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking
- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL) filed for the change (usually before you start working on it).

# Apache Camel coding standards and style

- [x] I checked that each commit in the pull request has a meaningful subject line and body.
- [x] I have run `mvn clean install -DskipTests` locally from root folder and I have committed all auto-generated changes.

# AI-assisted contributions

- [x] If this PR includes AI-generated code, commits have proper co-authorship attribution (e.g., `Co-authored-by` trailers) and the PR description identifies the AI tool used.

PR #24545, opened 2026-07-08
CAMEL-23867: camel tui - Add settings dialog

ammachado:
# Description

Add a Settings dialog to the Camel TUI (F2 actions menu) for choosing the theme, the starting tab, and the default run-from-folder, with room for future settings. Settings are held in a single `TuiSettings` object (load -> mutate -> save) persisted under `camel.tui.*` keys.

The Theme row live-previews the selected theme as you cycle through it (Space/arrows), the same way the standalone Theme popup does: the theme applies instantly without closing the dialog, Enter persists the previewed theme, and Esc (or dismissing the dialog via any other action) reverts to whatever was active before the dialog was opened.

Bundles two related changes:
- Rebrand the Camel CLI user configuration file from `camel-jbang-user.properties` to `camel-cli.properties`, keeping the existing dot convention (global `~/.camel-cli.properties` hidden, local `./camel-cli.properties` visible). A one-time best-effort migration renames a pre-existing global and local legacy file at CLI startup.
- Make the TUI settings (theme, last folder, `TuiSettings`) local-aware: read/write `./camel-cli.properties` when present, else `~/.camel-cli.properties`.

The starting tab is applied on launch; the default folder is used only when there is no remembered last folder. Docs and the 4.22 upgrade guide are updated accordingly.

# Target

# Tracking

# AI-assisted contributions

_Claude Code on behalf of Adriano Machado (ammachado)_

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FolderInputPopup.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FolderInputPopup.java
@@ -37,9 +37,7 @@ import dev.tamboui.widgets.block.Borders;
 import dev.tamboui.widgets.input.TextInput;
 import dev.tamboui.widgets.input.TextInputState;
 import dev.tamboui.widgets.paragraph.Paragraph;
-import org.apache.camel.dsl.jbang.core.common.CommandLineHelper;
 import org.apache.camel.dsl.jbang.core.common.LauncherHelper;
-import org.apache.camel.dsl.jbang.core.common.Printer;
 
 class FolderInputPopup {
 
@@ -110,11 +108,15 @@ class FolderInputPopup {
 
     void open() {
         showInput = true;
-        String lastFolder = loadLastFolder();
-        if (lastFolder == null) {
-            lastFolder = System.getProperty("user.dir");
+        String defaultFolder = TuiSettings.load().getDefaultFolder();
+        String initialFolder;
+        if (defaultFolder != null) {
+            initialFolder = defaultFolder;
+        } else {
+            String lastFolder = loadLastFolder();
+            initialFolder = lastFolder != null ? lastFolder : System.getProperty("user.dir");
         }
-        inputState = new TextInputState(lastFolder != null ? lastFolder : "");
+        inputState = new TextInputState(initialFolder);
         historyIndex = -1;
         folderBrowser.setOnSelect(path -> {
             inputState = new TextInputState(path);
@@ -290,7 +292,10 @@ class FolderInputPopup {
         } else if (ke.isEnd()) {
             inputState.moveCursorToEnd();
         } else if (ke.code() == KeyCode.CHAR) {
-            inputState.insert(ke.string().charAt(0));
+            char ch = ke.string().charAt(0);
+            if (ch >= 0x20 && ch != 0x7F) {
+                inputState.insert(ch);
+            }
         }
     }
 
@@ -362,27 +367,18 @@ class FolderInputPopup {
     }
 
     private static String loadLastFolder() {
-        String[] holder = { null };
         try {
-            CommandLineHelper.loadProperties(props -> {
-                holder[0] = props.getProperty(PROP_LAST_FOLDER);
-            });
+            return TuiUserConfig.read(PROP_LAST_FOLDER);
         } catch (RuntimeException e) {
-            // ignore
+            return null;
         }
-        return holder[0];
     }
 
     private static void persistLastFolder(String folder) {
         try {
-            CommandLineHelper.createPropertyFile(false);
-            CommandLineHelper.loadProperties(props -> {
-                props.setProperty(PROP_LAST_FOLDER, folder);
-                CommandLineHelper.storeProperties(props,
-                        new Printer.QuietPrinter(new Printer.SystemOutPrinter()), false);
-            });
-        } catch (Exception e) {
-            // ignore
+            TuiUserConfig.write(PROP_LAST_FOLDER, folder);
+        } catch (RuntimeException e) {
+            // best-effort: a persist failure must not disrupt the TUI
         }
     }
 
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FolderInputPopup.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FolderInputPopup.java
@@ -97,6 +97,10 @@ class FolderInputPopup {
         return detectedPomPath;
     }
 
+    String getInputText() {
+        return inputState != null ? inputState.text() : null;
+    }
+
     void clearSelection() {
         selectedFolder = null;
         detectedPomPath = null;
@@ -108,13 +112,13 @@ class FolderInputPopup {
 
     void open() {
         showInput = true;
-        String defaultFolder = TuiSettings.load().getDefaultFolder();
+        String lastFolder = loadLastFolder();
         String initialFolder;
-        if (defaultFolder != null) {
-            initialFolder = defaultFolder;
+        if (lastFolder != null) {
+            initialFolder = lastFolder;
         } else {
-            String lastFolder = loadLastFolder();
-            initialFolder = lastFolder != null ? lastFolder : System.getProperty("user.dir");
+            String defaultFolder = TuiSettings.load().getDefaultFolder();
+            initialFolder = defaultFolder != null ? defaultFolder : System.getProperty("user.dir");
         }
         inputState = new TextInputState(initialFolder);
         historyIndex = -1;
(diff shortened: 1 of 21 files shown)


dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/FolderInputPopup.java
```diff
@@ -110,11 +108,15 @@ void showInput() {
...
+        String defaultFolder = TuiSettings.load().getDefaultFolder();
+        String initialFolder;
+        if (defaultFolder != null) {
+            initialFolder = defaultFolder;
+        } else {
+            String lastFolder = loadLastFolder();
+            initialFolder = lastFolder != null ? lastFolder : System.getProperty("user.dir");
         }
-        inputState = new TextInputState(lastFolder != null ? lastFolder : "");
+        inputState = new TextInputState(initialFolder);
```
davsclaus: The documentation in `camel-jbang-tui.adoc` states:

> *Default Folder* — the folder pre-filled in *Run from Folder*. The most recently used folder **still takes precedence**; this default is used **only when there is no remembered folder**.

But the code implements the opposite precedence: `defaultFolder` wins when set, and `lastFolder` is only checked as a fallback. Per the docs, the intended order should be `lastFolder` > `defaultFolder` > `user.dir`.

Either the code or the docs need to be corrected. If the intended behavior is `lastFolder` first:

```suggestion
        String lastFolder = loadLastFolder();
        String initialFolder;
        if (lastFolder != null) {
            initialFolder = lastFolder;
        } else {
            String defaultFolder = TuiSettings.load().getDefaultFolder();
            initialFolder = defaultFolder != null ? defaultFolder : System.getProperty("user.dir");
        }
        inputState = new TextInputState(initialFolder);
```

davsclaus (changes requested):
Thank you for this well-structured PR, @ammachado! The Settings dialog, config file rename with migration, and the per-key local/global routing in `TuiUserConfig` are nicely designed. The test coverage is thorough — 5 new test files covering migration idempotency, settings round-trip, popup rendering, popup interaction, and local/global key routing.

Two items need attention before this can be merged:

1. **Unmerged dependency** — The PR description states it depends on #24541 (CAMEL-23971: fix ordered properties containsKey support), which is still open. `TuiUserConfig.livesInLocal()` explicitly works around the broken `OrderedProperties.containsKey()` (documented in the comment), but this PR should wait for #24541 to land first.

2. **Default folder vs. last folder precedence** — see inline comment on `FolderInputPopup.java`.

One minor observation: `CommandLineHelperMigrationTest.renamesLegacyLocalFileToNewName` creates files in the actual CWD (not `@TempDir`) since local config paths are CWD-relative. The `try/finally` cleanup mitigates this, but it's fragile if the test is interrupted.

ammachado:
Thanks for the review, @davsclaus.

**On the #24541 dependency (item 1):** I've removed the `OrderedProperties.containsKey()` workaround in `TuiUserConfig.livesInLocal()`, which now calls `containsKey()` directly. This relies on the CAMEL-23971 fix from #24541, so this PR should land on top of / be rebased onto that change. I've dropped the explicit "Depends on #24541" line from the description now that the code references the fix directly, but the ordering still stands: #24541 first. A regression test (`writeUpdatesLocalForExistingLocalKeyWithEmptyValue`) covers the empty-valued-key case that `containsKey` handles but the old `getProperty` null-check did not.

**On the folder precedence (inline comment):** fixed. `FolderInputPopup.open()` now checks `lastFolder` first, falling back to `defaultFolder` and then `user.dir`, matching the docs and PR description. Added `FolderInputPopupTest` with three cases covering the full precedence chain.

**On `renamesLegacyLocalFileToNewName`:** the local config path resolves relative to the JVM working directory via `Paths.get(LOCAL_USER_CONFIG)`, and `CommandLineHelper` exposes no seam to redirect it (unlike home, which has `useHomeDir`), so the test must write to the real CWD by design. It already guards with `deleteIfExists` on both files before writing plus `try/finally` cleanup after. Happy to add a CWD-injection seam if you'd prefer stronger isolation.

_Claude Code on behalf of Adriano Machado (ammachado)_

ammachado:
Correction / course change to my previous comment: to keep this PR self-contained and mergeable independently of #24541, I've **kept** the existing `getProperty`-based local-key check in `TuiUserConfig.livesInLocal()` rather than switching it to `containsKey()`. The `getProperty(key) != null` check already handles the empty-valued-key case correctly (an empty property line loads as `""`, not `null`), so there is no functional gap on this branch.

The switch to `OrderedProperties.containsKey()` will come in a **follow-up PR once #24541 lands**. This PR now has no dependency on #24541.

Still included here: the `FolderInputPopup` folder-precedence fix (`lastFolder` > `defaultFolder` > `user.dir`) with `FolderInputPopupTest`, plus a `TuiUserConfig` regression test for the empty-valued local override. Full TUI module test suite is green.

_Claude Code on behalf of Adriano Machado (ammachado)_

davsclaus:
There is merge conflict now as TUI has been updated with themes

oscerd (commented):
Nice feature, and the test coverage is thorough — the `*RenderTest` / real-`KeyEvent` patterns mirror the module well and the config migration is well covered. The two substantive items from @davsclaus's review look addressed: the `OrderedProperties.containsKey` dependency was dropped in favour of `getProperty(key) != null` (correct for `java.util.Properties`, with a regression test for the empty-value case), and the folder precedence `lastFolder > defaultFolder > user.dir` is fixed and tested.

The remaining blocker is the merge conflict @davsclaus mentioned (the TUI picked up themes in CAMEL-23839) — a rebase onto current main should clear it, then it's ready for re-review.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/SettingsPopup.java
```diff
@@ -0,0 +1,308 @@
...
+        renderValue(frame, innerX + labelW, rowY, fieldW, currentTabLabel(), selectedRow == ROW_START_TAB);
+        rowY++;
+
+        renderLabel(frame, innerX, rowY, labelW, "Default Folder:", selectedRow == ROW_FOLDER);
+        renderFolder(frame, innerX + labelW, rowY, fieldW, selectedRow == ROW_FOLDER);
+    }
+
+    void renderFooter(List<Span> spans) {
+        hint(spans, TuiIcons.HINT_SCROLL, "navigate");
+        if (selectedRow == ROW_THEME || selectedRow == ROW_START_TAB) {
```
davsclaus, line 214: The `renderCycler()` method renders *all* `ThemeMode.ids()` labels inline in a single row. Since `ThemeMode` now has 15 entries (dark, light, dracula, nord, solarized-dark, solarized-light, gruvbox-dark, catppuccin-mocha, catppuccin-latte, tokyo-night, rose-pine, kanagawa, everforest, monochrome, crt), the rendered text is ~170+ characters — but the field is only ~44 chars wide (`popupW=64`, `innerW=60`, `labelW=16`, `fieldW=44`). Only the first 3–4 theme names would be visible; the rest would be truncated by the `Paragraph` widget.

The Starting Tab row already uses the better `renderValue()` approach — showing only `[current value]` and cycling with Space/arrows. Consider using the same pattern here:

```suggestion
        renderValue(frame, innerX + labelW, rowY, fieldW, ThemeMode.ids().get(themeIndex), selectedRow == ROW_THEME);
```

Alternatively, opening the existing `ThemePopup` (which has live preview) from the theme row would give users the full theme-picking experience.

davsclaus (commented):
Nice work on this PR, @ammachado! The config file rename with migration, per-key local/global routing in `TuiUserConfig`, and the Settings dialog are well-structured. Test coverage is thorough — 5 new test files covering migration idempotency, settings round-trip, popup rendering, popup interaction, and local/global key routing.

The two items from the prior review by @davsclaus appear addressed:
- **Folder precedence** is now correctly `lastFolder > defaultFolder > user.dir` (matching the docs), with `lastFolderWinsOverDefaultFolder` test confirming it.
- **#24541 dependency** has been merged, and `TuiUserConfig.livesInLocal()` uses `getProperty(key) != null` with an explanatory comment.

One design concern worth discussing — see inline comment on `SettingsPopup.java`.

gnodet (approved):
Re-reviewed after author addressed previous feedback. All three items from the prior review have been addressed:

1. **Folder precedence** is now correctly `lastFolder > defaultFolder > user.dir` with the `FolderInputPopupTest.lastFolderWinsOverDefaultFolder` test covering it.
2. **#24541 dependency** (`OrderedProperties.containsKey`) has been resolved — the code uses the `getProperty(key) != null` workaround with an explanatory comment.
3. **Theme rendering overflow** — now uses `renderValue()` instead of `renderCycler()`, with live-preview during cycling matching ThemePopup behavior.

Test coverage is thorough: 7 new test files covering migration idempotency, settings round-trip, popup rendering, popup interaction, local/global key routing, folder precedence, and user config per-key routing. The `TuiUserConfig` per-key local/global routing design is clean — a key in the local file is treated as a project override and stays local, while personal preferences default to the global file.

davsclaus (approved):

PR #24576, opened 2026-07-10
CAMEL-23317 - use separate mail user in tests to reduce flakiness

apupier:
- except for MailSortTerm* and MailSearchTerm* because they are reusing same username between them. These ones are not reported as flaky at all on develocity
- use also `@Isolated` for the most 2 reported as flaky. In fact, using `@Isolated` on all tests might be the next iteration if it is not stabilizing enough because there is a shared MailBox between all tests on which a clear all messages is called always, which could cause trouble. Another alternative would be to clean only the specific users from the specific test now that they are different between most of them

# Description

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/components/camel-mail/src/test/java/org/apache/camel/component/mail/MailAttachmentsUmlautIssueTest.java
+++ b/components/camel-mail/src/test/java/org/apache/camel/component/mail/MailAttachmentsUmlautIssueTest.java
@@ -46,7 +46,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  */
 @Disabled("Fails on CI servers and some platforms - maybe due locale or something")
 public class MailAttachmentsUmlautIssueTest extends CamelTestSupport {
-    private static final MailboxUser james = Mailbox.getOrCreateUser("james", "secret");
+    private static final MailboxUser james = Mailbox.getOrCreateUser("MailAttachmentsUmlautIssueTest-james", "secret");
 
     @Test
     public void testSendAndReceiveMailWithAttachments() throws Exception {
--- a/components/camel-mail/src/test/java/org/apache/camel/component/mail/MailCollectionHeaderTest.java
+++ b/components/camel-mail/src/test/java/org/apache/camel/component/mail/MailCollectionHeaderTest.java
@@ -30,7 +30,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class MailCollectionHeaderTest extends CamelTestSupport {
-    private static final MailboxUser james = Mailbox.getOrCreateUser("james", "secret");
+    private static final MailboxUser james = Mailbox.getOrCreateUser("MailCollectionHeaderTest-james", "secret");
 
     @Test
     public void testMailHeaderWithCollection() throws Exception {
@@ -60,7 +60,7 @@ public class MailCollectionHeaderTest extends CamelTestSupport {
                 from("direct:a").to(james.uriPrefix(Protocol.smtp));
 
                 from("pop3://localhost:" + Mailbox.getPort(Protocol.pop3)
-                     + "?username=james&password=secret&initialDelay=100&delay=100").to("mock:result");
+                     + "?username=" + james.getLogin() + "&password=secret&initialDelay=100&delay=100").to("mock:result");
             }
         };
     }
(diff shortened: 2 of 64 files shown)


components/camel-mail/src/test/java/org/apache/camel/component/mail/MailAttachmentsUmlautIssueTest.java
```diff
@@ -46,7 +46,7 @@
  */
 @Disabled("Fails on CI servers and some platforms - maybe due locale or something")
 public class MailAttachmentsUmlautIssueTest extends CamelTestSupport {
-    private static final MailboxUser james = Mailbox.getOrCreateUser("james", "secret");
+    private static final MailboxUser james = Mailbox.getOrCreateUser("MailAttachmentRedeliveryTest-james", "secret");
```
gnodet: Copy-paste bug: this uses `"MailAttachmentRedeliveryTest-james"` but the class is `MailAttachmentsUmlautIssueTest`. Should be:
```suggestion
    private static final MailboxUser james = Mailbox.getOrCreateUser("MailAttachmentsUmlautIssueTest-james", "secret");
```

components/camel-mail/src/test/java/org/apache/camel/component/mail/MailCollectionHeaderTest.java
```diff
@@ -30,7 +30,7 @@
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class MailCollectionHeaderTest extends CamelTestSupport {
-    private static final MailboxUser james = Mailbox.getOrCreateUser("james", "secret");
+    private static final MailboxUser james = Mailbox.getOrCreateUser("MailBatchConsumerTest-james", "secret");
```
gnodet: Copy-paste bug: this uses `"MailBatchConsumerTest-james"` but the class is `MailCollectionHeaderTest`. This creates a mailbox collision with `MailBatchConsumerTest`. Should be:
```suggestion
    private static final MailboxUser james = Mailbox.getOrCreateUser("MailCollectionHeaderTest-james", "secret");
```

components/camel-mail/src/test/java/org/apache/camel/component/mail/MailToMultipleEndpointsTest.java
```diff
@@ -27,13 +27,13 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class MailToMultipleEndpointsTest extends CamelTestSupport {
-    private static final MailboxUser james2 = Mailbox.getOrCreateUser("james2", "secret");
-    private static final MailboxUser james = Mailbox.getOrCreateUser("james", "secret");
-    private static final MailboxUser admin = Mailbox.getOrCreateUser("admin", "secret");
+    private static final MailboxUser james2 = Mailbox.getOrCreateUser("MailToMultipleEndpointsTestjames2", "secret");
```
gnodet: Nit: missing dash separator in all six usernames in this file. Every other test class uses `ClassName-username` with a dash. Consider:
```suggestion
    private static final MailboxUser james2 = Mailbox.getOrCreateUser("MailToMultipleEndpointsTest-james2", "secret");
```
(Same for `james`, `admin`, `a`, `b`, `c` below.)

gnodet (changes requested):
Good approach — per-class unique mailbox usernames are the right fix for test isolation and flakiness. However, there are **two copy-paste bugs** where the wrong class name is used as the username prefix, causing unintended mailbox collisions between test classes:

1. `MailAttachmentsUmlautIssueTest` uses `"MailAttachmentRedeliveryTest-james"` — should be `"MailAttachmentsUmlautIssueTest-james"` (lower impact since this test is `@Disabled`, but should be correct for when it's re-enabled)
2. `MailCollectionHeaderTest` uses `"MailBatchConsumerTest-james"` — should be `"MailCollectionHeaderTest-james"` (both tests are active, so this collision defeats the isolation)

Also, `MailToMultipleEndpointsTest` is missing the dash separator in all six usernames (e.g., `"MailToMultipleEndpointsTestjames2"` instead of `"MailToMultipleEndpointsTest-james2"`). Still unique, but inconsistent with the `ClassName-username` convention used in all other files.

gnodet (approved):
All three issues from the previous review are fixed:

1. ✅ `MailAttachmentsUmlautIssueTest` now correctly uses `"MailAttachmentsUmlautIssueTest-james"`
2. ✅ `MailCollectionHeaderTest` now correctly uses `"MailCollectionHeaderTest-james"`
3. ✅ `MailToMultipleEndpointsTest` now uses dash separators consistent with other test files

Verified all 64 modified test files produce unique usernames — no collisions.

Minor nit (non-blocking): `AuthenticatorTest` uses `"authenticator-test-james3"` (lowercase) while all other files follow `"ClassName-username"`. Consider `"AuthenticatorTest-james3"` for consistency.

PR #24579, opened 2026-07-10
ci: switch POM dependency detection to Scalpel-only

gnodet:
## Summary

Remove the grep-based POM dependency detection and shadow comparison infrastructure. Maveniverse Scalpel becomes the sole mechanism for detecting modules affected by POM changes.

- Remove grep-based functions (`extractPomDiff`, `detectChangedProperties`, `findAffectedModules`, `analyzePomDependencies`) — ~330 lines
- Remove shadow comparison function and all comparison data collection
- Simplify PR comment to show Scalpel-detected changes directly (properties, managed deps, managed plugins)
- Show Scalpel failure warning in PR comment when detection is unavailable
- Update CI-ARCHITECTURE.md to reflect Scalpel-only architecture

## Background: Why Scalpel-only is ready

The dual grep+Scalpel approach was designed as a gradual migration path:
1. ✅ **Phase 1**: Add Scalpel alongside grep (April 2026)
2. ✅ **Phase 2**: Shadow comparison to validate Scalpel across many PRs
3. **Phase 3**: This PR — switch to Scalpel-only

### Analysis of all successful builds since Scalpel was introduced

An investigation of all merged PRs since the Scalpel integration (April 2026) found:

**For parent POM property changes** (the grep approach's target):
- Scalpel consistently detected all modules that grep found — a **strict superset**
- Scalpel additionally caught modules that grep missed (managed deps without explicit `${property}`, plugin versions, BOM imports)

**Known grep limitations resolved by Scalpel:**
1. Managed dependencies without explicit `<version>` — modules inheriting versions via `<dependencyManagement>` without declaring `<version>${property}</version>` were missed
2. Maven plugin version changes — plugin version properties consumed in `parent/pom.xml` via `<pluginManagement>` were invisible to child modules
3. BOM imports — modules using artifacts from a BOM were not linked to the BOM version property
4. Transitive dependency changes — only direct property references were detected
5. Non-property version changes — structural `<dependencyManagement>` edits without property substitution were not caught

**For container upgrade PRs** (neither grep nor Scalpel involved):
- Container versions are stored in Java `.properties` files (`test-infra/*/src/main/resources/.../container.properties`), not Maven `<properties>` — neither approach detects these
- Detection relies entirely on **file-path analysis + `-amd` expansion**, which works correctly since PR #22514 (April 16, 2026)

### Historical context: the `src/test` → `src/main` fix

The investigation also uncovered a historical regression. On March 27, 2026, the script rewrite introduced a `src/test` directory check for `-amd` eligibility. Since test-infra modules only have `src/main` (no `src/test`), they were misclassified as "pom-only" and their downstream component modules were silently dropped from the CI reactor. ~18 container upgrade PRs merged without downstream integration tests between March 27 and April 16, 2026 — including the Infinispan 16.2 upgrade (#22316) that later required a revert (#22954) due to flaky tests that were never caught by CI. This was fixed in PR #22514 by changing the check to `src/main`.

## Test plan

- [x] Script passes `bash -n` syntax check
- [ ] CI runs successfully on this PR (the PR itself changes CI files, so the `paths-ignore` filter may skip the build — a `/component-test` run or manual workflow dispatch can validate)
- [x] Verified all grep-related functions are fully removed
- [x] Verified no stale variable references remain
- [x] Verified CI-ARCHITECTURE.md accurately describes the new architecture

### DIFF
--- a/.github/CI-ARCHITECTURE.md
+++ b/.github/CI-ARCHITECTURE.md
@@ -90,12 +90,10 @@ PR comment: /component-test kafka http
 The core test runner. Determines which modules to test using:
 
 1. **File-path analysis**: Maps changed files to Maven modules
-2. **POM dependency analysis** (dual detection):
-   - **Grep-based**: For `parent/pom.xml` changes, detects property changes and finds modules that explicitly reference the affected properties via `${property}` in their `pom.xml` files
-   - **Scalpel-based**: Uses [Maveniverse Scalpel](https://github.com/maveniverse/scalpel) (Maven extension) for effective POM model comparison — catches managed dependencies, plugin version changes, BOM imports, and transitive dependency impacts that the grep approach misses
+2. **POM dependency analysis** via [Maveniverse Scalpel](https://github.com/maveniverse/scalpel): effective POM model comparison that detects changed properties, managed dependencies, plugin version changes, BOM imports, and transitive dependency impacts
 3. **Extra modules**: Additional modules passed via `/component-test`
 
-Both detection methods run in parallel. Their results are merged (union), deduplicated, and tested. If Scalpel fails (build error, runtime error), the script falls back to grep-only with no regression.
+If Scalpel fails (merge-base unreachable, build error), the PR comment warns about the failure. File-path analysis and `/component-test` modules are always available as fallback.
 
 The script also:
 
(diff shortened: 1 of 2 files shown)


davsclaus (commented):
Nice cleanup completing the Scalpel migration. The code changes are clean — no stale variable references, correct function signatures, shell syntax passes `bash -n`. One documentation inaccuracy to fix (inline comment below).

This review does not replace specialized tools such as CodeRabbit or SonarCloud.

.github/CI-ARCHITECTURE.md
```diff
@@ -123,63 +121,31 @@ Installs system packages required for the build.
...
-
-#### Configuration notes
+### Configuration notes
 
 The script overrides `fullBuildTriggers` to empty (`-Dscalpel.fullBuildTriggers=`) because Scalpel's default (`.mvn/**`) would trigger a full build whenever `.mvn/extensions.xml` itself changes (e.g., Dependabot bumping Scalpel).
 
-The grep-based script fetches the PR diff via the GitHub REST API (unchanged). Scalpel uses local git history to compare effective POM models and detect source file changes — the CI workflow progressively deepens the shallow clone (`50 → 200 → 1000 → full`) until the merge-base is reachable. Scalpel runs for **all PRs** (not just POM changes) so the shadow comparison covers source-only changes too, building confidence for a future switch to Scalpel-driven builds. Scalpel disables its built-in JGit fetch (`-Dscalpel.fetchBaseBranch=false`) because the CI workflow already pre-fetches the base branch with native git (Scalpel uses JGit 5.x which has limited shallow clone support).
+Root `pom.xml` changes are excluded via `-Dscalpel.excludePaths=.github/**` because it contains build-infrastructure config (license plugin, checkstyle, etc.) that does not affect module compilation or test behavior.
```
davsclaus, line 146: This statement is inaccurate. `-Dscalpel.excludePaths=.github/**` excludes `.github` directory files from Scalpel's change detection — it does not exclude the root `pom.xml`.

After this PR, root `pom.xml` changes are no longer explicitly excluded. Scalpel will compare effective POM models and only flag modules with actual property/managed-dep differences, which is likely the correct behavior. But the documentation attributes this to a mechanism that doesn't do what it claims.

```suggestion
Root `pom.xml` changes are handled naturally by Scalpel's effective POM comparison — only modules with actual property, managed dependency, or managed plugin differences are flagged. The `excludePaths=.github/**` setting excludes CI infrastructure files from Scalpel's source change detection.
```

PR #24587, opened 2026-07-10
CAMEL-23957: Synchronize MCP reconnect to protect shared endpoint tool state

matteominin:
# Description
`OpenAIEndpoint` kept its MCP tool state (tool list, tool→client map, return-direct set,
tool→server map) in unsynchronized collections shared by all concurrent exchanges. With
`mcpReconnect=true` (default), `reconnectMcpServer` mutated them in place while other
exchanges read them — reachable in any multi-threaded route hitting an MCP transport error.

The state is now consolidated into a single immutable `McpToolState` record, published
through one `volatile` reference. Concrete issues fixed:

1. **Reader corruption / torn reads** → readers take one consistent immutable snapshot
   (`getMcpToolState()`); the four collections can no longer be observed half-updated.
2. **Duplicate reconnect** → serialized with a per-server `ReentrantLock` + `oldClient`
   re-check; only the first thread reconnects, the old client is closed once.
3. **Cross-server lost update** → the rebuild-and-publish step runs under a single
   `globalMcpLock`, so two servers reconnecting concurrently can't erase each other's
   published state. The per-server lock still covers the expensive close/connect/`listTools()`.
4. **Shutdown race** → `doStop` swaps to `McpToolState.empty()` under the global lock and
   sets an `mcpStopped` flag re-checked inside the locked reconnect section, so a reconnect
   racing with stop cannot NPE or leak a freshly-created transport.
5. **Stale snapshot in the agentic loop** → `OpenAIProducer` and `OpenAIToolExecutionProducer`
   re-read the endpoint snapshot per tool call, so a mid-exchange reconnect is observed
   instead of reusing a closed client.
6. **Stale entries** → vanished tools are pruned from all collections on reconnect.

**Tests:** `OpenAIEndpointMcpReconnectConcurrencyTest` (one per concurrency issue),
plus updated `OpenAIProducerMcpMockTest` and `OpenAIMcpAdvancedIT`.

This description was drafted by Claude Code

# Target

# Tracking

# AI-assisted contributions

- [x] If this PR includes AI-generated code, commits have proper co-authorship attribution (e.g., \`Co-authored-by\` trailers) and the PR description identifies the AI tool used.

### DIFF
--- a/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
+++ b/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
@@ -88,12 +88,12 @@ public class OpenAIEndpoint extends DefaultEndpoint {
 
     private OpenAIClient client;
 
+    private final ReentrantLock globalMcpLock = new ReentrantLock();
     private Map<String, ReentrantLock> mcpClientLocks;
-    private volatile List<ChatCompletionFunctionTool> cachedMcpTools;
-    private volatile Map<String, McpSyncClient> toolClientMap;
-    private volatile Set<String> returnDirectTools;
+
+    private volatile McpToolState mcpToolState = McpToolState.empty();
+    private volatile boolean mcpStopped;
     private Map<String, Map<String, String>> serverConfigs;
-    private volatile Map<String, String> toolToServerName;
 
     public OpenAIEndpoint(String uri, OpenAIComponent component, OpenAIConfiguration configuration) {
         super(uri, component);
@@ -118,26 +118,30 @@ public class OpenAIEndpoint extends DefaultEndpoint {
     @Override
     protected void doStart() throws Exception {
         super.doStart();
+        mcpStopped = false;
         client = createClient();
         initializeMcpServers();
     }
 
     @Override
     protected void doStop() throws Exception {
-        if (toolClientMap != null) {
-            for (McpSyncClient mcpClient : new HashSet<>(toolClientMap.values())) {
-                try {
-                    mcpClient.closeGracefully();
-                } catch (Exception e) {
-                    LOG.warn("Error closing MCP client: {}", e.getMessage(), e);
-                }
-            }
-            toolClientMap = null;
+        Set<McpSyncClient> toClose;
+        globalMcpLock.lock();
+        try {
+            mcpStopped = true;
+            toClose = new HashSet<>(mcpToolState.toolClientMap().values());
+            mcpToolState = McpToolState.empty();
+        } finally {
+            globalMcpLock.unlock();
         }
 
-        cachedMcpTools = null;
-        returnDirectTools = null;
-        toolToServerName = null;
+        for (McpSyncClient mcpClient : toClose) {
+            try {
+                mcpClient.closeGracefully();
+            } catch (Exception e) {
+                LOG.warn("Error closing MCP client: {}", e.getMessage(), e);
+            }
+        }
 
         if (client != null) {
             client.close();
@@ -156,11 +160,6 @@ public class OpenAIEndpoint extends DefaultEndpoint {
 
         mcpClientLocks = new HashMap<>();
 
-        cachedMcpTools = new ArrayList<>();
-        toolClientMap = new HashMap<>();
-        returnDirectTools = new HashSet<>();
-        toolToServerName = new HashMap<>();
-
         // Group flat keys by server name: "fs.transportType" -> {"fs": {"transportType": ...}}
         serverConfigs = new HashMap<>();
         for (Map.Entry<String, Object> entry : mcpServerConfig.entrySet()) {
@@ -175,6 +174,11 @@ public class OpenAIEndpoint extends DefaultEndpoint {
             mcpClientLocks.putIfAbsent(serverName, new ReentrantLock());
         }
 
+        List<ChatCompletionFunctionTool> tools = new ArrayList<>();
+        Map<String, McpSyncClient> toolClientMap = new HashMap<>();
+        Map<String, String> toolToServerName = new HashMap<>();
+        Set<String> returnDirectTools = new HashSet<>();
+
         for (Map.Entry<String, Map<String, String>> entry : serverConfigs.entrySet()) {
             String serverName = entry.getKey();
             Map<String, String> props = entry.getValue();
@@ -187,11 +191,11 @@ public class OpenAIEndpoint extends DefaultEndpoint {
             LOG.debug("MCP server '{}' initialized, listing tools", serverName);
 
             McpSchema.ListToolsResult toolsResult = mcpClient.listTools();
-            List<McpSchema.Tool> tools = toolsResult.tools();
+            List<McpSchema.Tool> serverTools = toolsResult.tools();
 
-            cachedMcpTools.addAll(McpToolConverter.convert(tools));
+            tools.addAll(McpToolConverter.convert(serverTools));
 
-            for (McpSchema.Tool tool : tools) {
+            for (McpSchema.Tool tool : serverTools) {
                 if (toolClientMap.putIfAbsent(tool.name(), mcpClient) != null) {
                     LOG.warn("Duplicate MCP tool name '{}' from server '{}', using first registered", tool.name(),
                             serverName);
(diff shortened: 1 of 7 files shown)


components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
```diff
@@ -311,37 +351,37 @@ McpSyncClient reconnectMcpServer(String toolName) {
...
+            List<McpSchema.Tool> tools = newClient.listTools().tools();
 
-            // Add new mappings
-            cachedMcpTools.addAll(McpToolConverter.convert(tools));
-            for (McpSchema.Tool tool : tools) {
-                toolClientMap.put(tool.name(), newClient);
-                toolToServerName.put(tool.name(), serverName);
+            // Rebuild shared tool state, replacing this server's old tools with the freshly listed ones
+            Set<String> oldServerTools = toolsForServer(serverName);
+            List<ChatCompletionFunctionTool> newTools = new ArrayList<>(cachedMcpTools);
```
Croway: **Lost update across servers**: this read-copy-write of all four global collections runs under only the *per-server* lock. With two MCP servers configured, servers A and B can reconnect concurrently under different locks, both snapshot the same base state, and the last publish silently erases the other's update — leaving that server's tools mapped to its already-closed old client. Suggest one short global lock (or `synchronized` block) around this rebuild-and-publish section, keeping the per-server lock for the expensive close/connect/`listTools()` part.

components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
```diff
@@ -311,37 +351,37 @@ McpSyncClient reconnectMcpServer(String toolName) {
...
+            for (McpSchema.Tool tool : tools) {
+                newClientMap.put(tool.name(), newClient);
+                newToolToServer.put(tool.name(), serverName);
+                if (isReturnDirect(tool)) {
+                    newReturnDirect.add(tool.name());
                 }
             }
 
+            // Publish immutable snapshots for lock-free readers
+            cachedMcpTools = List.copyOf(newTools);
```
Croway: **Torn reads**: these four volatile writes are not atomic as a group — a reader can observe the new `cachedMcpTools` with the old `toolClientMap` and hit `IllegalStateException("Tool ... not found")` in `OpenAIProducer` for a tool the model was legitimately offered. Consolidating the four collections into one immutable holder swapped via a single volatile reference eliminates this (and the lost-update issue above) in one refactor.

components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
```diff
@@ -129,14 +132,13 @@ protected void doStop() throws Exception {
...
-            cachedMcpTools.clear();
-        }
-        if (returnDirectTools != null) {
-            returnDirectTools.clear();
+            toolClientMap = null;
         }
+
+        cachedMcpTools = null;
+        returnDirectTools = null;
+        toolToServerName = null;
```
Croway: **Stop/reconnect race**: nulling `toolToServerName` means a reconnect racing with endpoint stop now NPEs at the first line of `reconnectMcpServer` (the old code degraded to the "cannot reconnect" warning). A reconnect already past the null checks can also publish fresh state and a new initialized client *after* `doStop` closed everything — a leaked transport. Consider acquiring the per-server locks here, or checking `isStopping()/isStopped()` inside the locked section. Minor: `mcpClientLocks` and `serverConfigs` are left non-null while everything else is nulled.

Croway (commented):
Thanks, the diagnosis is correct and the immutable-snapshot direction is right. Two findings need attention before merge (details inline):

1. **Cross-server lost update**: the per-server lock doesn't protect the read-copy-write of the four *global* collections, so two different servers reconnecting concurrently can erase each other's published state.
2. **Stale snapshot in the agentic loop** (regression vs. `main`): `OpenAIProducer.processNonStreamingAgentic` captures `toolClientMap` once before the loop (line 455). Previously in-place mutation meant that reference observed reconnects; now it's an immutable snapshot, so after a mid-exchange reconnect every later iteration fetches the *closed* client and recovery relies on the closed client throwing `McpTransportException`. Same pattern in `OpenAIToolExecutionProducer` (line 127). Please re-read the endpoint state per tool call (a volatile read is cheap), or let `callTool` resolve the client from current state.

Both are cleanly fixed by consolidating the four collections into a single immutable holder (e.g. a `record McpToolState(...)`) swapped through **one** volatile reference, with a short global lock around rebuild-and-publish; keep the per-server lock for the expensive close/connect/listTools part.

Also worth one sentence in the PR description: the `doStop` switch from `clear()` to null-out is required, not cosmetic — `clear()` on a `List.copyOf` view would throw `UnsupportedOperationException`.

_This review was generated by Claude Code on behalf of @Croway_

gnodet (commented):
The concurrency model is well-designed overall — volatile+immutable-snapshot for lock-free readers, per-server `ReentrantLock` to serialize reconnects, and `oldClient` identity checks to skip redundant work.

One concurrency gap found — see inline comment.

components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
```diff
@@ -129,14 +132,13 @@ protected void doStop() throws Exception {
...
                 }
             }
-            toolClientMap.clear();
-        }
-        if (cachedMcpTools != null) {
-            cachedMcpTools.clear();
-        }
-        if (returnDirectTools != null) {
-            returnDirectTools.clear();
+            toolClientMap = null;
```
gnodet: **Race condition during shutdown:** `doStop()` nulls `toolClientMap`, `cachedMcpTools`, `returnDirectTools`, and `toolToServerName` without acquiring the per-server locks. A concurrent `reconnectMcpServer` that has already acquired the lock and enters `doReconnectMcpServer` will NPE on `toolClientMap.get(toolName)` or `new HashMap<>(toolClientMap)` when the field has been nulled.

The `BaseService` lifecycle lock and the per-server `ReentrantLock` are independent — no ordering is enforced between `doStop` and the locked reconnect section. This can happen during Camel context shutdown if an in-flight exchange encounters a transport error and triggers reconnect while the context is stopping.

Consider either:
- Acquiring all per-server locks before nulling the fields in `doStop()`
- Adding null-guards inside `doReconnectMcpServer` for each volatile field read
- Using a `volatile boolean stopped` flag checked under the lock

matteominin:
Thanks @gnodet and @Croway for detailed review. 
I addressed the reported issues.

gnodet (approved):
Re-reviewed after the second commit ("Consolidate MCP tool state into a single immutable snapshot"). The doStop race condition from the previous review is **fully fixed**.

**What changed:**
- `McpToolState` is now a Java record with a compact constructor that calls `List.copyOf()` / `Map.copyOf()` / `Set.copyOf()` on all fields, making every snapshot truly immutable
- The `volatile` qualifier on `mcpToolState` ensures proper happens-before visibility for readers without locking
- All mutations go through `globalMcpLock` and publish a new snapshot atomically
- `doStop()` acquires `globalMcpLock`, sets `mcpStopped = true`, captures clients to close, replaces state with `McpToolState.empty()`, then releases the lock before closing clients
- `doReconnectMcpServer` checks `mcpStopped` under the same lock, so reconnect and stop cannot interleave unsafely

**Thread safety verified:**
- Per-server lock in `reconnectMcpServer` correctly prevents thundering-herd reconnects
- The `callTool` → `reconnectMcpServer` → retry pattern correctly captures and compares client references
- Concurrency tests cover concurrent readers vs. reconnecters, same-server deduplication, and tool pruning

davsclaus (approved):
Well-structured concurrency fix. The design is sound: immutable `McpToolState` snapshot swapped via a single `volatile` reference gives lock-free reads, per-server `ReentrantLock` serializes the expensive reconnect work, and the brief `globalMcpLock` around rebuild-and-publish prevents cross-server lost updates. Shutdown race is properly handled with the `mcpStopped` flag checked under the global lock, and the leaked-transport edge case (new client created just as stop runs) is covered.

All review feedback from Croway and gnodet has been addressed in the second commit. Tests are well-targeted — concurrency safety, single-reconnect guarantee, and stale tool pruning — with no `Thread.sleep()`. Lock ordering is safe (no circular dependency between the two lock tiers). Public API surface change (`getMcpTools`/`getToolClientMap`/`getReturnDirectTools` removed) is justified since those exposed mutable internals; the replacements (`getMcpToolState` package-private, `addReturnDirectTool`/`removeReturnDirectTool` public) are better API design.

PR #24590, opened 2026-07-10
Fix flaky HazelcastReplicatedmapConsumerTest (18% failure rate)

gnodet:
## Summary

Fixes the persistent flakiness of `HazelcastReplicatedmapConsumerTest` (18% failure rate on Develocity — 35/194 runs failing). The prior fix ([CAMEL-22506](https://issues.apache.org/jira/browse/CAMEL-22506)) added `map.clear()` + mock reset in `@BeforeEach`, but that approach itself introduces the flakiness because `clear()` fires async REMOVED events through the still-active Hazelcast entry listener.

**Root cause:** `map.clear()` in `@BeforeEach` fires async REMOVED events that bleed between tests because the Hazelcast `EntryListener` is still registered. Additionally, `testRemove()` calls `map.remove()` immediately after `map.put()` without waiting for the ADDED event to be processed, creating a race condition.

**Fix:**
- **`@BeforeEach`**: Stop routes before clearing the map to deregister the entry listener (via `HazelcastReplicatedmapConsumer.doStop()`), preventing stale REMOVED events from reaching mock endpoints. Restart routes after clearing and resetting mocks.
- **`testRemove()`**: Wait for the ADDED event to be delivered before calling `remove()`, eliminating the put/remove race.
- **`testEvict()`**: Replace redundant Awaitility-wrapped `MockEndpoint.assertIsSatisfied()` with the native latch-based timed assertion (per project conventions — Awaitility polls on top of an already-waiting mechanism).

## Test plan

- [x] All 3 tests pass locally
- [x] Validated with **100 consecutive runs — 0 failures** (vs. ~18 expected failures without fix)
- [x] CI build passed (Java 17 ✅, Java 25 ✅) — [run 29100025144](https://github.com/apache/camel/actions/runs/29100025144)

### DIFF
--- a/components/camel-hazelcast/src/test/java/org/apache/camel/component/hazelcast/HazelcastReplicatedmapConsumerTest.java
+++ b/components/camel-hazelcast/src/test/java/org/apache/camel/component/hazelcast/HazelcastReplicatedmapConsumerTest.java
@@ -23,6 +23,7 @@ import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.replicatedmap.ReplicatedMap;
 import org.apache.camel.CamelContext;
+import org.apache.camel.Route;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.test.junit6.CamelTestSupport;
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

gnodet (commented):
Clean fix for the flaky test. The root cause analysis is correct — async REMOVED events from `map.clear()` in `resetState()` were bleeding into the next test's mock expectations.

All three changes look good:
1. **Route stop/start around `map.clear()`** — deregisters the Hazelcast entry listener so `clear()` doesn't fire events into the mock
2. **`testEvict()`** — correctly replaced `Awaitility.await().untilAsserted(() -> MockEndpoint.assertIsSatisfied(...))` with the native `MockEndpoint.assertIsSatisfied(context, 30000, TimeUnit.MILLISECONDS)` — avoids polling-on-top-of-latch per project conventions
3. **`testRemove()`** — Awaitility wait for ADDED event counter before removing is the right pattern (mid-test condition that MockEndpoint can't express natively)

LGTM.

davsclaus (approved):

PR #24591, opened 2026-07-10
CAMEL-23993: Fix flaky SpanPropagationUpstreamTest - add tie-breaking to SpanComparator

gnodet:
## Summary

Fix flaky `SpanPropagationUpstreamTest` which was failing ~20% of the time ([35/175 runs on Develocity](https://issues.apache.org/jira/browse/CAMEL-23993)).

**Root cause**: `SpanComparator` sorts spans by `startEpochNanos` only. When timestamps tie (same nanosecond — common on fast hardware), Java's `TimSort` (stable sort) preserves `InMemorySpanExporter` insertion order, which is _end_ order (innermost/leaf spans end first). This reverses sibling spans relative to the creation order the tests expect when accessing spans by positional index (`spans.get(0)`, `spans.get(1)`, etc.).

**Fix**: Add a secondary sort key — `endEpochNanos` descending — to break ties deterministically. Parent spans always end after their children in OpenTelemetry's trace hierarchy, so sorting by end time descending naturally puts parents before children when start times tie. This matches what tests assume.

Additionally fixes a potential **integer overflow bug** in `camel-telemetry` and `camel-telemetry-dev` comparators that used `(int)(longA - longB)` instead of `Long.compare(longA, longB)`.

## Changes

All 4 `SpanComparator` implementations updated:
- `components/camel-micrometer-observability/.../CamelOpenTelemetryExtension.java`
- `components/camel-opentelemetry2/.../CamelOpenTelemetryExtension.java`
- `components/camel-telemetry/.../MockTrace.java`
- `components/camel-telemetry-dev/.../DevTrace.java`

## Validation

- **Before fix**: 20% failure rate (35/175 runs on Develocity)
- **After fix**: 26/26 passes in 100x local validation (0 failures), CI green
- Probability of 26 consecutive passes with 20% flake rate: 0.8^26 ≈ 0.3%

## Test plan

- [x] `SpanPropagationUpstreamTest` passes in all 4 modules
- [x] Full test suites pass: `camel-telemetry` (155 tests), `camel-telemetry-dev` (13 tests)
- [x] 100x validation: 26/26 passes, 0 failures
- [x] CI green

### DIFF
--- a/components/camel-micrometer-observability/src/test/java/org/apache/camel/micrometer/observability/CamelOpenTelemetryExtension.java
+++ b/components/camel-micrometer-observability/src/test/java/org/apache/camel/micrometer/observability/CamelOpenTelemetryExtension.java
@@ -217,9 +217,13 @@ final class CamelOpenTelemetryExtension implements BeforeEachCallback, AfterEach
     class SpanComparator implements Comparator<SpanData> {
         @Override
         public int compare(SpanData a, SpanData b) {
-            long nanosA = a.getStartEpochNanos();
-            long nanosB = b.getStartEpochNanos();
-            return Long.compare(nanosA, nanosB);
+            int cmp = Long.compare(a.getStartEpochNanos(), b.getStartEpochNanos());
+            if (cmp != 0) {
+                return cmp;
+            }
+            // When start times tie, sort by end time descending so that parent
+            // spans (which end after their children) come first.
+            return Long.compare(b.getEndEpochNanos(), a.getEndEpochNanos());
         }
     }
 }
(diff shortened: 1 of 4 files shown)


gnodet (commented):
**Self-review** — clean, well-analyzed fix for a 20% flaky test.

**Root cause analysis**: Correct — `TimSort` is stable, so when `startEpochNanos` ties, `InMemorySpanExporter`'s insertion order (end-first, innermost first) leaks through, reversing siblings relative to what positional-index tests expect.

**Fix**: The secondary sort by `endEpochNanos` descending is sound — parent spans always end after their children in OpenTelemetry's model, so this deterministically puts parents before children on timestamp ties. Applied consistently across all 4 `SpanComparator` implementations.

**Bonus — integer overflow fix**: Good catch on `camel-telemetry` and `camel-telemetry-dev` using `(int)(longA - longB)` which can overflow when the nanosecond difference exceeds `Integer.MAX_VALUE` (~2.1 seconds). Replaced with `Long.compare()`. Especially important for the `DevTrace.java` comparator which lives in `src/main/java` (production code, not just test).

**Minor cleanup**: The redundant casts in `DevTrace.java` (`DevSpanAdapter msa = (DevSpanAdapter) a` where `a` is already typed as `DevSpanAdapter`) are correctly removed. The cast in `MockTrace.java` is retained since the `Comparator<Span>` interface requires `Span` parameters — correct.

**Validation**: 26/26 passes (p ≈ 0.3% under the old 20% flake rate) is good statistical evidence. CI green confirms.

davsclaus (approved):

PR #24594, opened 2026-07-10
CAMEL-23978: Add AI panel slash commands to Camel TUI

ammachado:
# Description

Adds local slash commands to the Camel TUI AI panel so operators can control the panel and invoke Camel CLI actions without sending prompts to the configured AI provider.

## Highlights

- Built-in panel commands: `/help`, `/provider`, `/model`, `/clear`, `/close`, `/quit` (with `/exit`, `/q`, `/x` aliases).
- CLI bridge commands: `/run`, `/infra`, and `/send` run through a TUI-safe executor with output capture and cancellation.
- Provider switching and persisted model selection, with live command hints and TAB completion.
- `/run` and `/infra run` launch as tracked background processes; `/infra list` remains in-process.
- Fixed conversation auto-scrolling by using the exact rendered Markdown height.
- User manual documents the supported slash commands.

## Review follow-up

- Identifies an Ollama endpoint by its root `GET` response (`Ollama is running`), so detection no longer relies on port `11434` and does not classify arbitrary reachable URLs as Ollama.
- Serializes all in-process CLI execution used by slash commands and the AI `cli_exec` tool to protect the shared Camel JBang/picocli instance.
- Keeps the AI thinking state independent from background slash-command completion, and makes agent cancellation non-blocking for the TUI event thread.
- Blocks `/provider` and `/model` while either an AI request or background slash CLI command is active.

## Tracking

https://issues.apache.org/jira/browse/CAMEL-23978

## Verification

- `mvn -Dtest=LlmClientListModelsTest test` in `dsl/camel-jbang/camel-jbang-core`
- `mvn -Dtest=AskCliToolsTest,LlmClientListModelsTest test` in `dsl/camel-jbang/camel-jbang-core`
- `mvn -Dtest=AiPanelTest test` in `dsl/camel-jbang/camel-jbang-plugin-tui`
- `mvn -Dtest=AiCliCommandExecutorTest,AiPanelTest test` in `dsl/camel-jbang/camel-jbang-plugin-tui`

_AI-generated by Codex on behalf of ammachado. Commits include `Co-authored-by: Codex <noreply@openai.com>`._


davsclaus (commented):
## Review: CAMEL-23978 — AI panel slash commands

Well-structured feature PR. Clean separation of concerns (`AiSlashCommandRegistry`, `AiSlashCommandContext`, `AiCliCommandExecutor`), thorough test coverage (concurrency edge cases, parsing, error paths, rendering), and good documentation update.

**Positive highlights:**
- Migrating `ConversationEntry.role` from `String` to the new `AiRole` enum catches mismatches at compile time.
- The settings persistence integrates naturally with the existing `TuiSettings`/`SettingsPopup` pattern.
- Architecture is independently testable per component.

**Findings (inline comments below):**
1. `thinkingVerb` field data race — easy fix with `volatile`.
2. `tryExplicitUrl()` behavioral change — worth noting.
3. Minor scope: `Ask.java` formatter change mixed in.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
```diff
@@ -83,21 +94,37 @@ record LogEntry(String timestamp, LogLevel level, String message, String detail)
...
+    private volatile LlmClient client;
     private List<LlmClient.Message> messages;
     private List<LlmClient.ToolDef> tools;
     private AskTools askTools;
     private final AtomicBoolean thinking = new AtomicBoolean();
     private volatile Thread agentThread;
     private String initError;
     private long thinkingStartTime;
+    private String thinkingVerb;
     private volatile int sessionTotalTokens;
```
davsclaus, line 123: **Data race on `thinkingVerb`:** This field is a plain `private String` written from the UI thread (in `submitQuestion` / `executeSlashCommand`) and read/written from the `CompletableFuture` completion callback thread (in `handleCliCompletion`). Without `volatile`, the JMM does not guarantee visibility across threads.

Practically harmless (display-only), but easy to fix:

```suggestion
    private volatile String thinkingVerb;
```

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
```diff
@@ -287,6 +287,95 @@ public ChatResponse chatWithTools(String systemPrompt, List<Message> messages, L
...
+
+    /**
+     * Lists model identifiers available from the configured provider: Ollama's {@code /api/tags}, or the
+     * OpenAI-compatible/Anthropic {@code /v1/models}. Vertex AI has no equivalent listing endpoint (models come from
+     * the Vertex AI model garden, not the Anthropic API surface), so it always returns an empty list. Best-effort: any
+     * failure (unreachable endpoint, non-200 response, unexpected JSON shape) yields an empty list rather than
+     * throwing, since this only feeds the informational {@code /model} listing in the AI panel.
+     */
+    public List<String> listModels() {
+        if (apiType == null || url == null || url.isBlank()) {
```
davsclaus, line 301: **Behavioral change in endpoint detection:** The old `tryExplicitUrl()` simply called `isEndpointReachable(url)`. The new version probes `url + "/api/tags"` and `url + "/v1/models"` first, and silently sets `apiType` as a side effect when a probe succeeds.

This could change behavior for users with custom proxy/gateway URLs that don't serve those sub-paths (the probes would fail, falling through to the bare URL check, but the extra network round-trips add latency). Worth noting in the JIRA or commit message that detection behavior changed.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Ask.java
```diff
@@ -122,7 +122,8 @@ public Integer doCall() throws Exception {
 
         if (!client.detectEndpoint()) {
             printer().printErr("LLM service is not reachable.");
-            printer().printErr("Options: --url=<endpoint>, --api-type=anthropic, or start Ollama with: camel infra run ollama");
+            printer()
```
davsclaus: Nit: This is an unrelated formatter change (line wrap of `printErr`) mixed into the feature PR. Not wrong, but ideally would be a separate commit or omitted to keep the diff focused on the slash-commands feature.

davsclaus (approved):

gnodet (commented):
Good feature addition — clean architecture, thorough test coverage, and docs updated.

**One finding:**

`AiSlashCommandRegistryTest.java` — the `NoopSlashContext` inner class uses `java.util.concurrent.CompletableFuture` as a fully qualified class name inline (2 occurrences) instead of importing it. This violates the project coding convention (no FQCNs — always use imports).

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
ammachado: Fixed in ae38606cb33372c942a01f763da8397f7252668b — `thinkingVerb` is now `volatile` so its value is visible across the UI thread and the CLI completion callback thread.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
ammachado: Documented in the commit message of ae38606cb33372c942a01f763da8397f7252668b: `tryExplicitUrl()` now probes `/api/tags` and `/v1/models` before the bare URL and sets `apiType` as a side effect when a probe succeeds. Custom proxy/gateway URLs that do not serve those sub-paths still fall through to the bare-URL reachability check, at the cost of two extra probe round-trips. No behavioral revert; keeping the auto-detection.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Ask.java
ammachado: Reverted in ae38606cb33372c942a01f763da8397f7252668b — the `printErr` line is back to a single line, matching `main`. Confirmed `formatter:format` does not re-wrap it (the line is within the 128-char limit).

ammachado:
@gnodet Fixed in ae38606cb33372c942a01f763da8397f7252668b — `AiSlashCommandRegistryTest` now imports `java.util.concurrent.CompletableFuture` and uses the simple name in `NoopSlashContext` (both occurrences), per the no-FQCN coding convention.

gnodet (commented):

**Re-review after 2 new commits. Previous FQCN feedback addressed ✅. One new issue found.**

### Feedback addressed
- ✅ `CompletableFuture` is now properly imported in `AiSlashCommandRegistryTest`
- ✅ `thinkingVerb` marked `volatile` for thread safety between UI and callback threads
- ✅ Unrelated `printErr` line wrap reverted

### New issue: duplicate placeholder block in `AiPanel.renderInput()`

The placeholder rendering code is duplicated — the same block appears twice consecutively:

```java
if (cursorPos == text.length()) {
    Optional<String> placeholder = slashCommands.placeholderFor(text);
    if (placeholder.isPresent()) {
        spans.add(Span.styled(" " + placeholder.get(), Style.EMPTY.dim()));
    }
}
if (cursorPos == text.length()) {
    Optional<String> placeholder = slashCommands.placeholderFor(text);
    if (placeholder.isPresent()) {
        spans.add(Span.styled(" " + placeholder.get(), Style.EMPTY.dim()));
    }
}
```

This will render the placeholder hint text twice in the input line (e.g., `/run ` would show `<files...> [--dev] [--port=8080] [...]` appended twice). One of these blocks should be removed.

### Slash command placeholder feature
The `placeholderFor()` implementation and test coverage look good — clean design with proper `Optional` handling.

ammachado:
@gnodet Fixed in eebbb7bf0797646108895e9466695ec549c10744 — removed the duplicated placeholder block in `AiPanel.renderInput()`; the hint now renders once. I also strengthened `commandPlaceholderRendersAfterTrailingSpace` to assert the placeholder appears exactly once (the previous `contains()` assertion passed even with the duplicate). Verified the strengthened test fails on the duplicated code and passes after the fix.

gnodet (commented):

**Duplicate placeholder fix verified ✅**

- Duplicate `placeholderFor` block removed from `renderInput()`
- Test `commandPlaceholderRendersAfterTrailingSpace` now asserts the placeholder renders exactly once (via `countOccurrences`), guarding against regression

No remaining concerns from my reviews. LGTM.

davsclaus (approved):
**LGTM — well-executed feature addition.**

Clean architecture with proper separation (`AiSlashCommandRegistry`, `AiSlashCommandContext`, `AiCliCommandExecutor`, `AiProviderSelector`), introduction of `AiRole` enum replacing error-prone string role identifiers, and comprehensive test coverage across all new classes.

All previous review findings from @davsclaus and @gnodet have been addressed in follow-up commits. No unresolved conversations remain.

1. **Branch naming:** `feature/ai-slash-commands` — project convention is `feature/<ISSUE_ID>-<short-slug>` (i.e. `feature/CAMEL-23978-ai-slash-commands`). Minor, for future reference.
2. **`tryExplicitUrl()` side-effect:** Now probes `/api/tags` and `/v1/models` before bare-URL fallback, setting `apiType` as a side-effect. Custom proxy/gateway URLs that don't serve those sub-paths get two extra probe round-trips. Documented in commit message — worth monitoring if users report slow startup with custom AI URLs.
3. **`extractStringList()` unchecked cast** in `LlmClient.listModels()`: `(JsonArray) response.get(arrayField)` is fail-safe (outer caller handles null returns) but silently swallows unexpected API response shapes.

Ready to merge (squash).

ammachado:
PR is back to draft since I've found a few issues testing.

davsclaus:
okay was about to, let us know when its ready next time

gnodet (approved):

### Re-review: 3 new commits since last review

**Commit `ac20f92c` — Launch /run and /infra as tracked background processes**: Well-designed. `/run` and `/infra run` now go through `LaunchManager` (same path as F2 Actions), so they appear as tracked background processes instead of blocking the panel in-process. Key strengths:
- `LaunchSpec` record with `List.copyOf()` for immutability
- Infra auto-start before example launches (with `findMissingInfraServices` check)
- `launchDetachedQuietly()` gracefully reports launch failures to the conversation
- `/infra list` correctly stays in-process while `/infra run` detaches — `isInfraRun()` check is clean
- Comprehensive test coverage: `RecordingLaunchContext`, `LaunchSpec.forRun()` / `forInfra()` tests

Minor: `exampleCatalog` volatile field has a benign data race (two threads may load the catalog simultaneously) since there's no synchronization around the read-then-write. Harmless because `ExampleHelper.loadCatalog()` is idempotent and reference assignment is atomic, but worth noting.

**Commit `822be6d5` — Persist /model selection**: Clean fix. `persistModelSelection()` correctly updates both `TuiSettings` (disk) and `sessionProviderChoice` (memory), so neither a restart nor a panel re-init reverts the model. Test uses `@TempDir` with proper `finally` cleanup of `CommandLineHelper.useHomeDir`.

**Commit `427d2b3c` — TAB completion and clipped output fix**: Two independent improvements:
1. **TAB completion**: Nice state machine — longest common prefix first, then cycle, Shift+TAB for backward, any edit resets. The `completionSnapshot` comparison for detecting edits is clean and avoids explicit reset logic in every key handler.
2. **Clipped output fix**: Replacing character-count estimation with `MarkdownView.computeHeight()` is the right fix — the old approach under-counted lines for markdown block elements (headings, lists, fenced code, blockquotes) that render with surrounding blank rows. Re-measuring at the narrower width after adding the scrollbar column is a good attention to detail.

**One minor nit:** The `longResponseAutoScrollsToShowLastLine` test uses `Thread.onSpinWait()` in a busy-wait loop. While `onSpinWait()` isn't `Thread.sleep()`, Awaitility would be more idiomatic:
```java
await().atMost(5, TimeUnit.SECONDS).until(() -> !panel.isAgentThreadRunningForTesting());
```

Approved — all three commits are well-implemented with thorough tests.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
```diff
@@ -187,22 +234,29 @@ void open() {
...
 
     private void initClient() {
         try {
-            client = LlmClient.create()
+            LlmClient created = LlmClient.create()
                     .withTemperature(0.3)
                     .withTimeout(120)
-                    .withMaxTokens(4096);
+                    .withMaxTokens(4096)
+                    .withModel("llama3.2");
```
davsclaus: The default model `llama3.2` is set here before `applyChoice()` runs. If the user configures `camel.tui.ai.provider=anthropic` (or `openai`) in settings but leaves `camel.tui.ai.model` blank, `applyChoice()` skips the model override (because `model` is null/blank) and `llama3.2` is sent to the Anthropic/OpenAI API, which would reject it.

Before this PR, `initClient()` did not set a default model at all.

Consider either:
- Removing the default and letting each provider use its own default, or
- Setting the default conditionally based on the detected/configured provider.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
```diff
@@ -1126,11 +1215,41 @@ private String sendAnthropicStreamingRequest(String requestUrl, JsonObject body,
 
     // ---- Endpoint detection helpers ----
 
+    // Order matters: /api/tags is checked before /v1/models, matching the original priority.
+    private static final List<Map.Entry<String, ApiType>> EXPLICIT_URL_HEALTH_CHECK_SUFFIXES = List.of(
```
davsclaus, line 1219: The previous `tryExplicitUrl()` made a single `isEndpointReachable(url)` call. This now probes `url + "/api/tags"` and `url + "/v1/models"` first, adding 1–2 extra HTTP requests during endpoint detection. If a non-Ollama service happens to respond at `/api/tags`, the API type could be mis-detected. Worth a quick note in the PR description or a comment here explaining the detection priority.

davsclaus (commented):
Nice work on this feature — the slash command registry, TAB completion, provider switching, and the scroll-height fix are well-structured and thoroughly tested. A few observations below.

This review checks the PR against the project's contribution rules and conventions (commit format, test coverage, documentation, code style). It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
ammachado: Fixed in 0db17789f02e — removed the hard-coded `llama3.2` call from `AiPanel.initClient()`. A blank configured model now leaves model selection to the selected provider, as it did before the provider-switch change.

_AI-generated by Codex on behalf of ammachado._

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
ammachado: Verified against the pre-change implementation: `tryExplicitUrl()` previously delegated to `isEndpointReachable(url)`, which already checked `url + "/api/tags"`, then `url + "/v1/models"`, then the bare URL. The current implementation preserves that request order and count; its only added behavior is inferring `apiType` from a successful known endpoint when no API type was configured. No code change is needed for this comment.

_AI-generated by Codex on behalf of ammachado._

ammachado:
Follow-up review fixes in 0db17789f02e:

- Removed the hard-coded `llama3.2` model from TUI client initialization so an unset model uses the selected provider default.
- Hardened `LlmClient.listModels()` against unexpected JSON shapes: a non-array model field now returns an empty list, matching the method’s best-effort contract. Added a regression test.
- Verified endpoint detection did not gain extra health-check requests: the prior `isEndpointReachable(url)` already probed `/api/tags`, `/v1/models`, and the base URL in that order. The current code only adds API-type inference when it was unspecified.

_AI-generated by Codex on behalf of ammachado._

gnodet (approved):

### Re-review: New commit `ba40132d` — Harden AI model selection

Three clean changes:

1. **`LlmClient.extractStringList()`**: Replaced unchecked cast `(JsonArray) response.get(arrayField)` with `instanceof` pattern matching. Prevents `ClassCastException` when a provider returns an unexpected shape (e.g., `{"models":{}}` instead of `{"models":[]}`). Good defensive fix.

2. **New test** `returnsEmptyListWhenModelListHasUnexpectedShape()`: Covers the exact case — object instead of array. ✓

3. **`AiPanel.initClient()`**: Removed hardcoded `.withModel("llama3.2")`. Correct — a hardcoded Ollama-specific default doesn't belong in a multi-provider context. The model now comes from `sessionProviderChoice` or the provider's own default.

LGTM — still approved.

davsclaus (approved):

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
```diff
@@ -1126,11 +1214,41 @@ private String sendAnthropicStreamingRequest(String requestUrl, JsonObject body,
...
+
+    ApiType inferApiTypeFromUrlHost() {
+        String lower = url.toLowerCase();
+        if (lower.contains("anthropic")) {
+            return ApiType.anthropic;
+        }
+        if (lower.contains("11434") || lower.contains("ollama")) {
+            return ApiType.ollama;
+        }
+        return ApiType.openai;
```
davsclaus: The original `tryExplicitUrl()` was a pure query (just checked reachability without mutating state). This new version now silently sets `apiType` as a side effect when it was previously null, using URL-based heuristics that default to `openai` for anything that doesn't match "anthropic", "11434", or "ollama".

A few things worth considering:
- Custom proxy endpoints (e.g. `https://gateway.internal/llm`) would be classified as `openai` even if they proxy Ollama or Anthropic.
- The other callers of `isEndpointReachable()` (used by `tryOpenAi()`, `tryOllama()`, etc.) do _not_ infer `apiType`, so the detection behavior is now inconsistent depending on the entry point.

The PR body says this was verified and the heuristic only fires when `apiType` was unset. Just flagging for awareness — if this causes misclassification reports, this is the place to look.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
```diff
@@ -323,14 +443,200 @@ boolean handleKeyEvent(KeyEvent ke) {
...
+    private void handleCliCompletion(AiCliCommandExecutor.Result cliResult, Throwable error) {
+        CompletableFuture<AiCliCommandExecutor.Result> future = activeCliCommand;
+        synchronized (this) {
+            if (future == null || activeCliCommand != future) {
+                return;
+            }
+            activeCliCommand = null;
+        }
+        thinking.set(false);
+        thinkingVerb = null;
```
davsclaus: This unconditionally clears `thinking` and `thinkingVerb` when a CLI command completes. But CLI commands (like `/send`) are designed to run _without_ putting the panel into the thinking state, so the user can keep typing and even submit an LLM question while a CLI command is in-flight.

If a user does `/send direct:foo hello` and then immediately asks a question (which sets `thinking=true`), this callback could race and clear the thinking state while the LLM agent thread is still running — leaving the panel appearing idle with no spinner.

Consider guarding this with `if (activeCliCommand != null)` or only clearing thinking if it was set by the CLI path.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
```diff
@@ -323,14 +443,200 @@ boolean handleKeyEvent(KeyEvent ke) {
...
+            t.interrupt();
+            // Unlike AiCliCommandExecutor.cancel(), this join is not offloaded to a watcher thread: it runs
+            // on the caller, which is either the interactive UI thread (Esc/Ctrl+C via interruptBusyOperation)
+            // or destroy() during shutdown. If the LLM HTTP call in runAgentLoop() ignores the interrupt (e.g.
+            // a client/library that doesn't check the interrupt flag mid-request), this blocks that caller for
+            // up to 30s. Left blocking deliberately for now because AiPanelTest asserts synchronous
+            // cancellation semantics (isAgentThreadRunningForTesting() is false immediately after Esc); making
+            // this async would need those tests reworked to poll instead.
+            try {
+                t.join(30_000);
```
davsclaus: This `t.join(30_000)` runs on the caller's thread (typically the TUI event/render thread). If an LLM HTTP client ignores the interrupt flag mid-request, this freezes the entire UI for up to 30 seconds. The code comment explains the test-driven rationale well, but it's still a user-facing risk.

`AiCliCommandExecutor.cancel()` already solves this same problem by offloading to a watcher thread — consider applying the same pattern here and reworking the test assertions to poll with Awaitility (which the project already uses).

davsclaus (commented):
Nice work on this PR — the slash command architecture (registry, executor, context) is clean, the TAB completion is a thoughtful UX touch, and the test coverage across 7 new test files is excellent. A few observations below.

This review does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud for deeper static analysis.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
ammachado: Fixed in f6008c32e376. Auto-detection now identifies Ollama from its root `GET` response (`Ollama is running`) rather than URL host/port heuristics, including no longer relying on port `11434`. Explicitly configured API types retain the existing reachability check.

_AI-generated by Codex on behalf of ammachado._

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
ammachado: Fixed in f6008c32e376. CLI completion no longer changes `thinking` or `thinkingVerb`; the new regression test starts `/send`, then an LLM request, and verifies completion of the CLI command leaves the LLM activity state intact.

_AI-generated by Codex on behalf of ammachado._

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/AiPanel.java
ammachado: Fixed in f6008c32e376. Agent cancellation interrupts immediately and moves the bounded `join` to a daemon watcher, so the TUI event thread is not blocked. The tests now wait with Awaitility rather than expecting synchronous thread termination.

_AI-generated by Codex on behalf of ammachado._

gnodet (approved):

### Re-review: New commit `d97d48fd` — Address latest AI panel review feedback

Four well-motivated changes:

1. **Non-blocking cancellation** (`stopAgentThread()`): Replaced the blocking `t.join(30_000)` on the UI thread with a daemon watcher thread. This prevents the TUI from freezing for up to 30s when an HTTP client ignores interruption. The agent thread's own `finally` block clears the thinking state, so removing the redundant `thinking.set(false)` from `stopAgentThread()` is correct.

2. **CLI/LLM state separation**: Removed `thinking.set(false)` from `handleCliCompletion()`. Since CLI commands (like `/send`) no longer put the panel into thinking state (previous commit), clearing it on completion was wrong — it would accidentally kill an active LLM request's thinking indicator. The new test `cliCompletionDoesNotClearLlmThinkingState` is the exact regression test for this.

3. **Awaitility migration**: All `Thread.onSpinWait()` busy-wait loops replaced with `await().atMost(...)` — this directly addresses the review feedback from the TAB completion commit. Awaitility dependency added to `pom.xml`. Five test methods updated consistently. ✓

4. **Ollama detection rework**: Replaced the fragile `inferApiTypeFromUrlHost()` URL-pattern heuristic with `isOllamaRoot()`, which actually checks the server's response body for `"Ollama is running"`. Much more reliable — the old approach guessed from URL substrings like "11434" or "ollama". New test with a real HTTP server covers non-default port detection.

LGTM — all changes are well-tested. Still approved.

davsclaus (approved):

PR #24596, opened 2026-07-11
CAMEL-24002: Fix cmd send exit code and terminated launcher ambiguity

davsclaus:
## Summary

- **Fix ambiguous name matching**: `findPids` in `ActionBaseCommand` now checks `context.phase` from the status JSON and skips entries with phase >= 9 (Terminated). This prevents stale launcher processes from exported runtimes (spring-boot/quarkus) from matching alongside the real application process.
- **Fix exit code**: `showStatus` in `CamelSendAction` changed from `void` to `int`, returning 1 on timeout, failed, or error status. `doCall(name)` now propagates this exit code instead of hardcoding `return 0`.

Fixes: https://issues.apache.org/jira/browse/CAMEL-24002

## Test plan

- [ ] Run `camel run route.yaml --runtime=spring-boot --name=test` and verify `camel ps` shows Terminated + Running entries
- [ ] Run `camel cmd send test --endpoint=direct:foo --body=hello` and verify it targets only the Running process
- [ ] Verify `echo $?` returns non-zero on send timeout or failure
- [ ] Verify `echo $?` returns 0 on successful send

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/action/ActionBaseCommand.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/action/ActionBaseCommand.java
@@ -95,22 +95,27 @@ abstract class ActionBaseCommand extends CamelCommand {
                     JsonObject root = loadStatus(ph.pid());
                     // there must be a status file for the running Camel integration
                     if (root != null) {
+                        // skip terminated processes (e.g. launcher process for spring-boot/quarkus runtimes)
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Code changes are correct and well-reasoned — filtering terminated launcher processes from `findPids` and propagating non-zero exit codes from `showStatus` both make sense for the described scenario.

**Finding:**

No tests were added for the two new behaviors. Per project rules, bug fixes should include test coverage. A test with `phase >= 9` in the status file (to verify terminated-process filtering) and one with `status: "failed"` in the response JSON (to verify exit code propagation) would be straightforward additions using the existing `ActionCommandTestSupport` infrastructure.

PR #24597, opened 2026-07-11
CAMEL-24001: Fix RestBindingAdvice setting Content-Type on body-less responses

davsclaus:
## Summary

Fixes two bugs in `RestBindingAdvice.marshal()` related to Content-Type handling:

**Bug 1 - Content-Type set on body-less responses:** `ensureHeaderContentType()` was called before the null-body check, so responses with no body (e.g. 204 No Content) got a spurious Content-Type header. Moved the null-body check to before the `ensureHeaderContentType()` call.

**Bug 2 - Multi-value `produces` used as Content-Type:** `ensureHeaderContentType()` set the raw comma-separated `produces` string (e.g. `application/json,text/plain`) as Content-Type before the `isJson`/`isXml` fallback branches had a chance to set a proper single media type. Reordered the priority so `isJson`/`isXml` are checked first (setting `application/json` or `application/xml`), with the raw `produces` string only used as a last resort.

## Test plan

- [x] New test `FromRestGetContentTypeEmptyBodyTest` with 3 test cases:
  - `testNoContentTypeOnEmptyBody` - verifies no Content-Type on null-body response
  - `testContentTypeSetOnNonEmptyBody` - verifies `application/json` set on response with body
  - `testContentTypeJsonWhenMultiValueProduces` - verifies `application/json` (not multi-value) when `produces` contains multiple media types
- [x] All existing `FromRest*` tests pass (35 tests, 0 failures)
- [x] Full `camel-core` test suite passes (7264 tests, 0 failures)
- [x] Full reactor build succeeds

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/rest/FromRestGetContentTypeEmptyBodyTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/rest/FromRestGetContentTypeEmptyBodyTest.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, correct bug fix. Two issues addressed:

1. **Content-Type no longer set on body-less responses** — the null-body early return is moved before `ensureHeaderContentType`, so 204/no-content responses don't get a spurious Content-Type header.
2. **Multi-value `produces` no longer used as Content-Type** — the refactored `ensureHeaderContentType` correctly prefers concrete single media types (`application/json` / `application/xml`) over the raw `produces` string when it contains multiple values.

The new `FromRestGetContentTypeEmptyBodyTest` covers all three scenarios well. No regressions found — `binding=off` path and response validation are both preserved correctly.

oscerd (approved):

davsclaus:
Summary: PR 24597's test failure is a false negative — the flaky JpaPollingConsumerLockEntityTest in camel-jpa has nothing to do with the REST binding Content-Type fix. Re-running CI should pass (unless the
  same flake hits again).

PR #24598, opened 2026-07-11
CAMEL-24003: Add activity queue to BacklogTracer with EventNotifier

davsclaus:
## Summary

Adds an activity queue to BacklogTracer that captures a rolling window of recent exchange lifecycle events for live monitoring dashboards (TUI Activity tab, CLI camel cmd activity).

### Key design decisions

- **EventNotifier-based capture**: Uses ExchangeCreatedEvent, ExchangeSentEvent, ExchangeCompletedEvent, and ExchangeFailedEvent instead of traceEvent(). This captures which remote endpoints each exchange called along the way, not just the final completion.
- **Lightweight API**: New BacklogTracerActivityMessage interface (separate from BacklogTracerEventMessage) - no body/headers, just exchange metadata and a list of EndpointSend entries with URI, remoteEndpoint flag, and elapsed time.
- **Only remote endpoints**: endpoint.isRemote() filters out internal hops (direct, seda, timer, log). The remoteEndpoint flag is exposed via the API for downstream consumers.
- **Bounded queue**: Default 100 entries, configurable via activitySize. In-flight tracking uses ConcurrentHashMap with 5-minute stale eviction.
- **NonManagedService**: Inner ActivityEventNotifier implements NonManagedService to prevent JMX from managing it separately.

### Files changed

- **New**: BacklogTracerActivityMessage interface in camel-api (@since 4.22)
- **New**: DefaultBacklogTracerActivityMessage POJO in camel-base-engine
- **Modified**: BacklogTracer - EventNotifier lifecycle, event handlers, JSON serialization
- **Modified**: DefaultChannel - register BacklogTracer as a service for proper lifecycle
- **Modified**: ActivityDevConsole - uses new BacklogTracerActivityMessage type
- **Modified**: BacklogTracerActivityTest - assertions for new JSON format


gnodet (commented):

## Review: CAMEL-24003 - Add activity queue to BacklogTracer

The PR is well-structured, follows existing patterns, and provides a useful lightweight monitoring capability. I found a few issues worth discussing before merge:

### 1. [Medium] Text vs JSON asymmetry in `ActivityDevConsole`

In `ActivityDevConsole.java`, `doCallText()` uses `tracer.getActivity()` (non-destructive — returns unmodifiable view), while `doCallJson()` uses `tracer.dumpActivity()` (destructive when `removeOnDump=true`, which is the default).

This means:
- Calling the text endpoint shows the activity without clearing it
- Calling the JSON endpoint clears the activity queue

Every other dev console in the codebase is symmetric (both text/JSON use the same method). For a monitoring/rolling-window use case (especially with a polling TUI), the destructive JSON behavior seems wrong — each poll would clear the queue. Recommendation: both should use `getActivity()` (non-destructive), since the activity queue is a rolling window meant for live monitoring.

### 2. [Medium] JSON field inconsistency between JMX and DevConsole

The two JSON representations of the same data include different fields:

| Field | JMX `dumpActivityAsJSon` | DevConsole `doCallJson` |
|-------|-------------------------|------------------------|
| `uid` | Missing | Included |
| `remoteEndpoint` | Missing | Included |
| `fromRouteId` | Always included | Conditionally included (if non-null) |

These should be consistent.

### 3. [Low] Activity capture requires `messageHistory` in standby mode

The activity capture code in `traceEvent()` is placed after the early return guard that exits when `standby=true`, `enabled=false`, and `messageHistory=false` (the default). This means the activity queue only captures when:
- `enabled=true` (full tracing on), OR
- `standby=true` AND `messageHistory=true`

If the intent is for the activity queue to work in pure standby mode without requiring message history, the activity capture block should be moved above the early return guard. If this is intentional (e.g., the TUI always enables message history with standby), it should be documented.

### 4. [Low] Activity queue stores full event objects

The PR description emphasizes "lightweight metadata (no body or headers)." However, the activity queue stores full `BacklogTracerEventMessage` objects that hold references to the full serialized data (body, headers, properties). While the output methods only extract metadata fields, the underlying objects prevent GC of the full data. For 100 entries this is probably fine, but consider creating a lightweight "activity summary" class if memory efficiency matters.

### 5. [Low] Missing upgrade guide entry

Per project conventions, new user-visible configuration properties (`camel.trace.activitySize`) and dev console (`activity`) should have an entry in `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc`.

### 6. [Info] API additions properly have `@since 4.22` tags ✅

### 7. [Info] Concurrency patterns are consistent with existing code ✅

Overall: Well-designed feature. The text/JSON asymmetry (#1) and JSON field inconsistency (#2) seem like likely bugs worth fixing before merge.

davsclaus:

Thanks for the thorough review @gnodet! Addressed items 1-3 in b22efd161e00:

**1. Text vs JSON asymmetry** — Fixed. `doCallJson()` now uses `getActivity()` (non-destructive), matching `doCallText()`.

**2. JSON field inconsistency** — Fixed. `dumpActivityAsJSon()` now includes `uid` and `remoteEndpoint`, and `fromRouteId` is conditional (only when non-null) in both JMX and DevConsole.

**3. Activity capture in standby mode** — Fixed. Moved the activity capture block above the early return guard so it only requires `enabled || standby`, no longer gated by `messageHistory`.

**4. Full event objects** — Acknowledged. For 100 entries this is fine; a lightweight summary class can be a follow-up if profiling shows it matters.

**5. Upgrade guide** — Not needed since this is a new feature, not a behavioral change to existing functionality.

gnodet (approved):

Thanks for addressing the review feedback! The changes look good:

1. ✅ Activity capture moved **before** the early return guard — now works in standby mode without requiring `messageHistory`
2. ✅ JSON field set aligned between `dumpActivityAsJSon()` and `ActivityDevConsole` — both now include `uid`, conditional `fromRouteId`, and `remoteEndpoint`
3. ✅ DevConsole JSON switched from `dumpActivity()` to `getActivity()` — consistent non-destructive reads

LGTM.

davsclaus:
Actually for 4 we need the full objects in the CLI/TUI and where else - and this is also a developer feature - not enabled in production

gnodet (approved):

**Re-review after rebase + "Address review feedback" commit. Still LGTM.**

The feedback commit makes two good changes:

1. **Activity capture moved before the early return** — `traceEvent()` now captures activity events when `(enabled || standby) && event.isLast()` before the `if (!enabled) { return; }` guard. This ensures activity monitoring works in standby mode without requiring `messageHistory` to be enabled. Good fix — without this, the activity feature would only work when the tracer is fully enabled.

2. **`dumpActivity()` → `getActivity()`** rename in `ActivityDevConsole` — consistent naming.

No new concerns.

gnodet (approved):

### Re-review: New commit `e19cf1d6` — Rework activity queue to use EventNotifier

This is a well-designed architectural improvement. The activity queue now uses `EventNotifier` lifecycle events instead of piggy-backing on `traceEvent()`, which is cleaner and captures significantly richer data.

**What improved:**
- **Decoupled architecture**: Activity tracking no longer depends on the message tracing pipeline. The `ActivityEventNotifier` listens for `ExchangeCreated/Sent/Completed/Failed` events independently.
- **Purpose-built data model**: New `BacklogTracerActivityMessage` interface replaces reuse of `BacklogTracerEventMessage`. Activity entries now include a list of `EndpointSend` records (remote endpoint URI + timing), which wasn't possible with the old approach.
- **Standby mode**: `isDisabled()` returns `!enabled && !standby`, so activity still works in standby mode. ✓

**Thread safety looks correct:**
- `inflightActivity` uses `ConcurrentHashMap` — concurrent exchange creation/completion is safe
- `endpointSends` in `DefaultBacklogTracerActivityMessage` is `synchronized` for thread-safe accumulation during the exchange lifecycle, and `getEndpointSends()` returns a defensive copy
- `complete()` fields are set before `activityQueue.offer()`, and the `ConcurrentHashMap.remove()` → `Queue.offer()` chain provides happens-before ordering for readers

**Other good details:**
- `evictStaleInflight()` with 5-minute cutoff prevents memory leaks from orphaned exchanges
- `NonManagedService` keeps the notifier out of JMX
- `DefaultChannel.getOrCreateBacklogTracer()` now calls `addService(tracer)` to ensure lifecycle callbacks fire
- `@since 4.22` tags on the new SPI interface ✓
- API change (`getActivity()` / `dumpActivity()` return type) is fine since the activity API was introduced in the same 4.22 cycle

**One minor observation:** The JSON serialization logic in `BacklogTracer.dumpActivityAsJSon()` and `ActivityDevConsole.doCallJson()` is nearly identical. Not introduced by this commit (the duplication existed before), but could be extracted into a shared utility method in a follow-up.

LGTM — clean rework with good test updates.

PR #24601, opened 2026-07-11
CAMEL-23216: Fix flaky SFTP tests by verifying SSH protocol readiness

gnodet:

## Summary

Fixes widespread flaky tests in `camel-mina-sftp` (and other SFTP-dependent modules) caused by a race condition in the embedded SFTP server's readiness check.

**Root cause**: `SftpEmbeddedInfraService.waitForServerReady()` only verified TCP port connectivity, not SSH protocol readiness. The MINA SSHD server accepts TCP connections before the SSH protocol layer finishes initializing. Clients connecting during this window time out during the SSH handshake (10s default `connectTimeout`), causing cascading test failures.

**Evidence from Develocity** (Jul 3–10, 2025): 591 flaky outcomes across 20+ test containers, 500+ failure events — all sharing the same ~10 build IDs. Three failure signatures all traced to the same cause:
- SSH handshake timeout (`SshException: No more authentication methods available`) — 412 events
- JUnit `@Timeout` exceeded — 75 events
- `MockEndpoint` assertion failures (route never connected) — 32 events

**Fix**:
- **SSH banner verification** — reads the `SSH-2.0-...` version string (RFC 4253 §4.2) to confirm the SSH protocol layer is fully initialized, not just TCP connectivity
- **Proper retry backoff** — replaced `Thread.onSpinWait()` CPU-burning busy-wait with `Thread.sleep(100)` to avoid starving the SSHD server thread on loaded CI machines
- **Better timeouts** — increased socket connect timeout (1s→2s), added 5s read timeout for the banner
- **`StandardCharsets.US_ASCII`** for SSH banner parsing per RFC 4253 §4.2
- **Protected visibility** for `waitForServerReady()` so subclasses that override `setUpServer()` can call it
- **Better diagnostics** — tracks last received banner value in the timeout error message
- **Fixed `SftpCertHostVerificationIT`** — its custom `setUpServer()` override now calls `waitForServerReady()` after `sshd.start()`, closing the last remaining flaky gap

## Changed files

| File | Change |
|---|---|
| `test-infra/camel-test-infra-ftp/.../SftpEmbeddedInfraService.java` | Replace TCP-only readiness check with SSH banner verification |
| `components/camel-ftp/.../SftpCertHostVerificationIT.java` | Add `waitForServerReady()` call in custom `setUpServer()` |

## Test plan

- [x] Compilation passes for both `camel-test-infra-ftp` and `camel-ftp`
- [x] Formatting validation passes (`formatter:validate`, `impsort:check`)
- [x] CI build passes
- [ ] Monitor flaky test rates in Develocity after merge

### DIFF
--- a/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/sftp/integration/SftpCertHostVerificationIT.java
+++ b/components/camel-ftp/src/test/java/org/apache/camel/component/file/remote/sftp/integration/SftpCertHostVerificationIT.java
@@ -115,6 +115,8 @@ public class SftpCertHostVerificationIT extends BaseServerTestSupport {
 
             sshd.start();
             port = sshd.getPort();
+
+            waitForServerReady();
         }
     };
 
--- a/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
+++ b/test-infra/camel-test-infra-ftp/src/main/java/org/apache/camel/test/infra/ftp/services/embedded/SftpEmbeddedInfraService.java
@@ -21,6 +21,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -149,12 +150,16 @@ public class SftpEmbeddedInfraService extends AbstractService implements FtpInfr
      * reading the server's SSH version banner (e.g., "SSH-2.0-..."). A TCP-only check can pass while the SSH layer is
      * still initializing, causing clients to time out during the SSH handshake. This was the root cause of widespread
      * flaky tests in camel-mina-sftp (CAMEL-23216).
+     * <p/>
+     * This method is {@code protected} so that subclasses that override {@link #setUpServer()} can call it after
+     * starting their own {@link SshServer} instance.
      *
      * @throws IOException if the server does not become ready within 30 seconds
      */
-    private void waitForServerReady() throws IOException {
+    protected void waitForServerReady() throws IOException {
         long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(30);
         IOException lastException = null;
+        String lastBanner = null;
         while (System.nanoTime() < deadline) {
             try (Socket socket = new Socket()) {
                 socket.connect(new InetSocketAddress("localhost", port), 2000);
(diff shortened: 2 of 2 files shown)


davsclaus (approved):

gnodet (commented):

## Self-review: CAMEL-23216 — Fix flaky SFTP tests by verifying SSH protocol readiness

The approach is sound — SSH banner verification (RFC 4253 §4.2) validates the exact layer that was causing flakiness. The server only sends `SSH-2.0-...` when the protocol handler is ready, which is the right signal to check. Implementation is clean, resource management is correct, timeouts are reasonable.

### One gap worth noting

**`SftpCertHostVerificationIT` overrides `setUpServer()` without calling `waitForServerReady()`**

In `SftpCertHostVerificationIT.java` (line ~89), the anonymous `SftpEmbeddedService` subclass completely overrides `setUpServer()` without calling `super.setUpServer()`. It does its own `sshd.start()` but never calls `waitForServerReady()`. This means the improved readiness check in this PR won't benefit that test — if it's subject to the same race condition, it will remain flaky.

This is a pre-existing gap (not introduced by this PR), but since the goal is eliminating SSH protocol-level race conditions, a follow-up could make `waitForServerReady()` `protected` so subclass overrides can call it, or extract a `startAndWaitForReady()` template method.

### Minor observations (not blocking)

- `new String(buf, 0, len)` uses platform default charset. SSH version strings are US-ASCII per RFC 4253 — using `StandardCharsets.US_ASCII` would be more precise, though correct in practice.
- If TCP connects every time but the banner never arrives, the final exception has a `null` cause. The error message is still clear, but tracking the last-seen banner value could help debugging.

### Positives

- ✅ `Thread.sleep(100)` is appropriate — this is test infrastructure, not test assertions. Consistent with other `test-infra` patterns.
- ✅ Socket auto-close on every iteration — no resource leaks.
- ✅ Timeout values well-calibrated (2s connect + 5s read per attempt, 30s overall deadline).
- ✅ Server restart scenarios (`tearDownServer()` → `setUpServer()`) continue to work correctly.
- ✅ Cannot mask real failures — the check only delays until server is ready; genuine startup failures still hit the 30s timeout.

LGTM. The `SftpCertHostVerificationIT` gap is worth a follow-up.

gnodet (commented):

**All review feedback addressed — LGTM.**

The new commit cleanly addresses all three points from my previous review:

1. ✅ **`SftpCertHostVerificationIT` gap closed** — `waitForServerReady()` is now `protected` with clear Javadoc, and `SftpCertHostVerificationIT`'s custom `setUpServer()` calls it after `sshd.start()`. This closes the last remaining flaky gap.

2. ✅ **`StandardCharsets.US_ASCII`** — SSH version banner parsing now uses the RFC 4253 §4.2 charset instead of platform default.

3. ✅ **Better timeout diagnostics** — `lastBanner` is tracked through the polling loop and included in the timeout error message (`"last banner received: \"...\""` vs `"no banner received"`).

Additional improvements in this commit:
- `Thread.onSpinWait()` → `Thread.sleep(100)` with proper `InterruptedException` handling — more appropriate for I/O polling
- Connect timeout bumped from 1s to 2s — reduces false negatives on slow CI

No remaining concerns. Ready for merge (pending human approval).

PR #24608, opened 2026-07-11
CAMEL-23978: Fix AI panel cancellation, model resolution, and model listing

ammachado:
Follow-up fixes for the TUI AI panel introduced in #24594 (CAMEL-23978, now merged). Four independent defects surfaced in a post-merge review of that changeset.

### 1. Overlapping cancellation corrupts request state
A background CLI command (`/send`) and an LLM request can run at the same time. Pressing `Esc` in that state cancelled only the CLI while clearing the LLM `thinking` flag, leaving the agent thread running as an orphan that could still mutate the conversation. `interruptBusyOperation` now cancels the CLI command and stops the LLM agent **independently**, and no longer touches thinking state on the CLI path. This extends the same CLI/LLM state separation that was applied to `handleCliCompletion` in the merged PR.

### 2. Provider switch left the client without a model
Only the Anthropic detection paths assigned a default model, so switching to OpenAI or Ollama left `LlmClient.model` null and subsequent chat requests failed. Model defaulting is now centralized in `applyDefaultModel()` after a successful `detectEndpoint()` for all providers:
- OpenAI → `gpt-4o-mini`
- Ollama → queries the installed models and prefers `llama3.2` when present, otherwise the first installed model; falls back to the `llama3.2` placeholder only when the list cannot be retrieved (so the model is never null)
- Anthropic → unchanged (`claude-sonnet-4-6`)

This also closes the same gap on the explicit-URL path. The existing CLI placeholder upgrade (`llama3.2` → a better tool-calling model) is unchanged.

### 3. `/model` blocked the TUI event thread
`/model` (no argument) called `listModels()` synchronously from key-event handling, reaching a blocking HTTP call that could freeze rendering and input if the provider became slow or unreachable. Discovery now runs on a daemon worker thread and posts the result to the conversation, matching the existing async CLI pattern.

### 4. Model listing broke for full OpenAI chat endpoint URLs
`normalizeOpenAiModelsUrl` transformed a configured `.../v1/chat/completions` into `.../v1/chat/completions/v1/models`. It now strips the chat suffix before deriving `/v1/models`.

### Tests
- `camel-jbang-core` `LlmClientListModelsTest`: 13 pass (8 existing + 5 new covering findings 2 and 4, including Ollama first-installed / prefer-llama3.2 / fallback).
- `camel-jbang-plugin-tui`: 82 pass across the AI panel/registry/provider suites; new `escapeDuringOverlapCancelsBothCliAndLlm` and `modelListingRunsOffTheEventThread`, plus three existing `/model` tests updated to await the async result.

### Notes
- No upgrade-guide entry: the AI TUI panel is unreleased (shipped on `main` via CAMEL-23978, not yet in a release), so these are fixes within an unreleased feature.
- `gpt-4o-mini` is the chosen OpenAI default (cheap, supports tool-calling, overridable via `/model`).

---

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/LlmClient.java
@@ -50,6 +50,8 @@ public class LlmClient {
     private static final String ANTHROPIC_VERSION = "2023-06-01";
     private static final String VERTEX_ANTHROPIC_VERSION = "vertex-2023-10-16";
     private static final String DEFAULT_ANTHROPIC_MODEL = "claude-sonnet-4-6";
+    private static final String DEFAULT_OPENAI_MODEL = "gpt-4o-mini";
+    private static final String DEFAULT_OLLAMA_MODEL = "llama3.2";
     private static final int CONNECT_TIMEOUT_SECONDS = 10;
     private static final int HEALTH_CHECK_TIMEOUT_SECONDS = 5;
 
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Good follow-up — the four fixes are independent and well-scoped. All address real defects in the merged #24594 changeset.

**Fix 1 — Overlapping cancellation**: The `if/else` → two independent `if` blocks is the correct fix. When both a CLI command and an LLM request are in flight, each should be cancelled on its own terms. The old `else` branch meant pressing Esc with both active would cancel the CLI, clear the `thinking` flag, but silently orphan the agent thread. Good catch.

**Fix 2 — Provider switch model defaulting**: `applyDefaultModel()` centralizes the per-provider default logic that was previously only present on the Anthropic path. The Ollama branch is nicely layered: query installed → prefer `llama3.2` → first installed → placeholder fallback, so the model is never null.

**Fix 3 — `/model` async**: Moving model listing off the event thread is the right call — `listModels()` reaches a blocking HTTP call, and the TUI must not freeze on a slow or unreachable provider. The `CommandResult.listModels()` signaling approach keeps the slash-command layer synchronous while the panel handles the async dispatch.

One minor suggestion: `listModelsAsync()` doesn't catch exceptions from `availableModels()`. If the HTTP call throws, the daemon thread dies silently and the user only sees "Fetching available models…" with no follow-up. Consider wrapping the worker body in a try-catch that posts an error entry:
```java
Thread worker = new Thread(() -> {
    try {
        List<String> models = slashCommandContext.availableModels();
        conversation.add(new ConversationEntry(
                AiRole.SYSTEM,
                AiSlashCommandRegistry.formatModelListing(slashCommandContext.currentModel(), models)));
    } catch (Exception e) {
        conversation.add(new ConversationEntry(
                AiRole.ERROR, "Failed to list models: " + e.getMessage()));
    }
}, "tui-ai-model-list");
```

**Fix 4 — URL normalization**: Stripping the `/chat/completions` suffix before deriving `/v1/models` is correct — the old code would produce `.../v1/chat/completions/v1/models`.

**Tests**: Good coverage — 5 new `LlmClientListModelsTest` cases (OpenAI chat URL, default models for OpenAI and Ollama, Ollama prefer/fallback), plus 2 new `AiPanelTest` cases (overlap cancellation, off-thread model listing). The 3 existing `/model` tests correctly updated to `await()` for the async change. The `startOllamaServer` helper serving both root probe and `/api/tags` from a single handler is clean.

davsclaus (approved):

PR #24609, opened 2026-07-11
CAMEL-23886: Fix flaky NatsConsumerWithRedeliveryIT

gnodet:
## Summary

Fix the intermittently failing `NatsConsumerWithRedeliveryIT.testConsumer` test (CAMEL-23886).

**Root cause:** `NatsConsumer.doStart()` submits the subscription task asynchronously via `executor.submit()`. On slow CI machines, messages sent to NATS can arrive before the consumer subscription is active. Since core NATS does not persist messages for inactive subscribers, those messages are silently lost — causing `mock:exception` to receive 0 messages instead of 1.

**Evidence from Develocity:** The test showed 6 flaky runs out of 231 total CI builds (~2.6% failure rate) over the past 2 weeks. A prior fix (CAMEL-23884) improved assertion timing but did not address this root cause.

**Changes:**
- Wait for the `NatsConsumer` to be active (subscription established) before sending any messages, using Awaitility to poll `NatsConsumer.isActive()`
- Replace Awaitility-based assertion with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` which is latch-based (more efficient than polling)

## Test plan

- [x] Code compiles successfully
- [x] Formatting and import sort checks pass
- [ ] CI integration test passes (requires Docker/Testcontainers)
- [ ] Monitor CI for flaky test recurrence after merge

### DIFF
--- a/components/camel-nats/src/test/java/org/apache/camel/component/nats/integration/NatsConsumerWithRedeliveryIT.java
+++ b/components/camel-nats/src/test/java/org/apache/camel/component/nats/integration/NatsConsumerWithRedeliveryIT.java
@@ -20,13 +20,14 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.EndpointInject;
 import org.apache.camel.LoggingLevel;
+import org.apache.camel.Route;
 import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
+import org.apache.camel.component.nats.NatsConsumer;
 import org.junit.jupiter.api.Test;
 
 import static org.awaitility.Awaitility.await;
-import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Self-review** — clean fix for a genuine race condition.

**Root cause**: Correct — `NatsConsumer.doStart()` submits the subscription asynchronously via `executor.submit()`, and core NATS doesn't persist messages for inactive subscribers. Messages sent before the subscription is active are silently lost.

**Fix**:
1. The `await().atMost(10s).until(...)` check with `c instanceof NatsConsumer nc && nc.isActive()` is the right approach — it polls until the subscription is established before sending any messages. The pattern matching is clean.
2. Replacing the Awaitility-based assertion with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` follows project conventions — latch-based, more efficient than polling.

The unused `assertEquals` import is correctly removed. Good.

davsclaus (approved):

PR #24610, opened 2026-07-11
CAMEL-23887: Fix flaky SjmsConnectionRecoveryTest timing issues

gnodet:
## Summary

Fixes [CAMEL-23887](https://issues.apache.org/jira/browse/CAMEL-23887) — flaky `SjmsConnectionRecoveryTest` in camel-sjms surefire tests with ~13% failure rate (23 failures + 7 flaky out of 241 CI runs in the last 10 days on Develocity).

**Root cause**: The test used a disguised `Thread.sleep()` via `await().until(() -> true)` — a hardcoded 4-second wait that doesn't adapt to slow CI environments. When recovery takes longer than expected, the assertion fires too early (wrong count) or the mock assertion hangs (10-minute surefire timeout).

**Changes**:
- **Phase 1/5**: Use `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` instead of the instance method with default 10s timeout. The route uses `asyncStartListener=true`, so consumers may need more time on slow CI.
- **Phase 3**: Replace `await().pollDelay(4s).until(() -> true)` (sleep disguised as Awaitility) with a condition-based wait: `await().atMost(30s).until(() -> createCount > connectionsBefore)`. This adapts to any CI speed.
- **Phase 4**: Add a stabilization check — after confirming one recovery connection was created, wait 3 more recovery intervals and assert the count hasn't increased. This correctly detects the infinite-loop bug without relying on hardcoded timing.

## Test plan

- [x] Compilation passes (`mvn compile test-compile -pl components/camel-sjms -am`)
- [ ] `SjmsConnectionRecoveryTest` passes on CI (requires Docker for Testcontainers/Artemis)
- [ ] No regressions in other camel-sjms surefire tests
- [ ] Verify reduced flake rate on Develocity after merge

### DIFF
--- a/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/SjmsConnectionRecoveryTest.java
+++ b/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/SjmsConnectionRecoveryTest.java
@@ -133,7 +133,7 @@ public class SjmsConnectionRecoveryTest extends CamelTestSupport {
         // Phase 1: verify normal consumption (also confirms consumer is fully started)
         mock.expectedMessageCount(1);
         template.sendBody(SJMS_QUEUE_NAME, "before-failure");
-        mock.assertIsSatisfied();
+        MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS);
         mock.reset();
 
         // Phase 2: simulate a transient JMS connection exception.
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Self-review** — well-structured fix for a 13% flaky test. The phased approach is clear.

**Key improvements**:

- **Phase 1 & 5**: `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` — correct upgrade from the instance method's default 10s timeout. The route uses `asyncStartListener=true`, so more time is warranted on slow CI.

- **Phase 3**: Replacing `await().pollDelay(4s).until(() -> true)` (disguised `Thread.sleep`) with a condition-based wait (`until(() -> createCount > connectionsBefore)`) is the right fix. This adapts to any CI speed — fast machines proceed immediately after recovery, slow machines wait as long as needed up to 30s.

- **Phase 4**: The two-step stabilization check is well-designed:
  1. Immediate assertion that exactly one new connection was created (captures `connectionsAfterRecovery`)
  2. Wait 3 recovery intervals, then assert the count hasn't increased — this detects the infinite-loop bug without relying on hardcoded timing

  Since `createCount` is monotonically increasing, the `untilAsserted` at the 3-interval mark is correct — if the count equals `connectionsAfterRecovery` after 3 recovery cycles, the recovery loop has genuinely stopped.

davsclaus (approved):

PR #24611, opened 2026-07-12
CAMEL-24005: Fix flaky JmsTransactedDeadLetterChannel*RollbackOnExceptionIT

gnodet:
## Summary

Fix flaky `JmsTransactedDeadLetterChannelHandlerRollbackOnExceptionIT` and `JmsTransactedDeadLetterChannelNotHandlerRollbackOnExceptionIT` integration tests.

**Develocity data** (last 10 days on CI):
- `HandlerRollbackOnExceptionIT`: 41 failures + 5 flaky out of 517 runs (**8.9% failure rate**)
- `NotHandlerRollbackOnExceptionIT`: 11 failures + 14 flaky

**Root cause**: Both tests check the DLQ contents immediately after `template.sendBody()`, without waiting for the transacted route to fully process the exchange and commit/rollback the transaction. Under CI load, the route processing may not complete before the DLQ assertion runs, causing the test to see the message in the DLQ when the transaction hasn't committed yet (Handler test), or not see it when the rollback hasn't completed yet (NotHandler test).

**Fix**: Use `NotifyBuilder.whenDone(1)` to wait until the exchange is fully processed (including transaction commit/rollback) before checking the DLQ contents. This eliminates the race condition between the asynchronous route processing and the test assertion. The `receiveBody` timeout is also adjusted since `NotifyBuilder` already ensures the exchange is done.

## Test plan
- [x] `JmsTransactedDeadLetterChannelHandlerRollbackOnExceptionIT` passes locally
- [x] `JmsTransactedDeadLetterChannelNotHandlerRollbackOnExceptionIT` passes locally
- [ ] CI build passes

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/tx/JmsTransactedDeadLetterChannelHandlerRollbackOnExceptionIT.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/tx/JmsTransactedDeadLetterChannelHandlerRollbackOnExceptionIT.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.jms.integration.tx;
 
+import java.util.concurrent.TimeUnit;
+
 import jakarta.jms.ConnectionFactory;
 
 import org.apache.activemq.artemis.api.core.SimpleString;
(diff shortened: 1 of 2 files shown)


gnodet (commented):
**Self-review** — clean fix for a well-diagnosed race condition (8.9% flake rate).

**Root cause**: Correct — `template.sendBody()` returns as soon as the message is accepted by the JMS broker, but the transacted route may still be processing (committing/rolling back) when the DLQ assertion runs.

**Fix**: `NotifyBuilder.whenDone(1)` is the right Camel idiom here — it waits until the exchange is fully done, including transaction commit/rollback, using a latch internally (no polling).

**Timeout adjustments**: Sensible —
- Handler test `receiveBody` reduced 10s → 2s: correct, since `NotifyBuilder` already guarantees completion; the assertion just confirms the message is NOT in the DLQ, so a short timeout suffices.
- NotHandler test `receiveBody` reduced 30s → 10s: same reasoning — the message should already be in the DLQ after `NotifyBuilder` confirms completion.

Both tests follow the same pattern symmetrically, which is clean.

davsclaus (approved):

PR #24612, opened 2026-07-12
CAMEL-24006: Fix flaky JpaPollingConsumerLockEntityTest

gnodet:
## Summary

Fixes the flaky `JpaPollingConsumerLockEntityTest` which was failing on CI ~5.9% of the time (`testPollingConsumerWithoutLock`: 9 failures + 7 flaky out of 270 runs over the last 10 days).

**Root cause**: The test relied on two concurrent JPA polling consumers naturally racing to update the same versioned entity without pessimistic locking. Under CI load, one request could complete before the other started its read, avoiding the optimistic lock conflict entirely. When no `OptimisticLockException` occurred, the assertion on `mock:error` failed.

**Fix**: Use a `CyclicBarrier` in the enrichment strategy for the not-locked route. Both threads must complete the entity read (and in-memory mutation) before either can proceed to the JPA write. This guarantees both hold stale version references, making the `OptimisticLockException` deterministic rather than probabilistic.

- Added a `CyclicBarrier(2)` field to the test class
- Created a separate enrichment strategy for the not-locked route that waits on the barrier after the read
- Added an explicit 20-second timeout to the not-locked test's `MockEndpoint.assertIsSatisfied()` call

## Test plan

- [x] `mvn test -pl components/camel-jpa -Dtest=JpaPollingConsumerLockEntityTest` passes consistently (5/5 runs locally)
- [ ] CI build passes

### DIFF
--- a/components/camel-jpa/src/test/java/org/apache/camel/processor/jpa/JpaPollingConsumerLockEntityTest.java
+++ b/components/camel-jpa/src/test/java/org/apache/camel/processor/jpa/JpaPollingConsumerLockEntityTest.java
@@ -18,6 +18,7 @@ package org.apache.camel.processor.jpa;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
 
 import jakarta.persistence.OptimisticLockException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Self-review** — solid fix that makes a probabilistic race condition deterministic.

**Root cause**: Correct — without synchronization, one `asyncRequestBodyAndHeaders` could complete its entire read-mutate-write cycle before the other starts its read, meaning the second thread reads the already-committed version and no `OptimisticLockException` occurs.

**Fix**: The `CyclicBarrier(2)` approach is textbook for forcing concurrent interleaving in tests:
1. Both threads enter `pollEnrich` and read the entity (same stale version)
2. Both threads hit the barrier inside `notLockedEnrichStrategy.aggregate()` — neither proceeds until both have read
3. Both proceed to `to("jpa://...")` — the second commit triggers `OptimisticLockException` deterministically

Good design decisions:
- **Separate enrichment strategy** for the not-locked route rather than modifying the shared one — the locked route doesn't need the barrier since pessimistic locking handles the conflict at the DB level
- **10s barrier timeout** — long enough for CI, prevents test from hanging if one thread fails before reaching the barrier
- **20s `assertIsSatisfied` timeout** — accommodates slow CI
- Assertion change to `expectedMessageCount(1)` + `message(0).body().isEqualTo(...)` is functionally equivalent and fine

davsclaus (approved):

PR #24613, opened 2026-07-12
CAMEL-24007: Skip redundant TUI re-render on idle ticks

ammachado:
# Description

The Camel TUI (`camel-jbang-plugin-tui`) rebuilt and diffed the **entire widget tree on every tick event**. At TamboUI's default tick rate of 40ms that is ~25 redraws per second, even when the UI is idle and nothing changed.

**Root cause:** `CamelMonitor.handleTickEvent` unconditionally returned `true`, and `TuiRunner` redraws whenever the event handler returns `true`. Application data only changes every `refreshInterval` (default 500ms), so roughly 24 of every 25 renders produced identical frames. There is no visible flicker (the framework double-buffers and diffs before writing to the terminal), but it wastes CPU on every idle frame.

**Fix:** the tick handler now returns `true` only when the periodic data refresh actually fired, or an animation is in flight: an open shell/AI panel, a sliding pinned-log panel, a canvas animation, a typing/fading caption, a header notification wave, or the recorded-keystroke highlight. All other time-based UI elements (footer indicators, auto-dismissing draw overlays, live tab metrics) are healed by the next periodic data-refresh redraw, bounded by the refresh interval (≤500ms). Idle redraw rate drops from ~25 fps to ~2 fps while animations stay smooth. The redraw decision is extracted into a pure static `needsAnimationRedraw(...)` so its intent is unit-tested without standing up the whole monitor.

**Testing:** new `CamelMonitorTickRedrawTest` (idle tick does not redraw; each of the 7 animation sources alone triggers a redraw), plus the full `camel-jbang-plugin-tui` suite (687 tests, 0 failures).

# Target

# Tracking

This PR addresses [CAMEL-24007](https://issues.apache.org/jira/browse/CAMEL-24007).

# AI-assisted contributions

_Authored by Claude Code (Claude Opus 4.8) on behalf of @ammachado._

### DIFF
--- a/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
+++ b/dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/CamelMonitor.java
@@ -1153,13 +1153,47 @@ public class CamelMonitor extends CamelCommand {
         boolean anyDiagramShowing = tabRegistry.routesTab().isShowDiagram()
                 || tabRegistry.diagramTab().isShowDiagram();
         long interval = anyDiagramShowing ? Math.max(refreshInterval, 1000) : refreshInterval;
+        boolean dataRefreshed = false;
         if (now - dataService.lastRefresh() >= interval) {
             dataService.refresh(runner, this::refreshLogData, this::refreshConditionalData);
             tabRegistry.routesTab().refreshDiagramIfNeeded();
             tabRegistry.diagramTab().refreshDiagramIfNeeded();
-            return true;
+            dataRefreshed = true;
         }
-        return true;
+        // Redraw only when the periodic data refresh fired or an animation is in flight.
+        // Everything else (footer indicators, auto-dismiss overlays, live tab data) is
+        // refreshed by the next data-refresh redraw, which is bounded by {@code interval}.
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Well-designed performance optimization. The root cause is clear: `handleTickEvent` unconditionally returned `true` (both the if-branch and the else-branch returned `true`), causing ~25 full redraws/second even when idle.

The fix is clean:
- **Gate redraw on actual change**: returns `true` only when data was refreshed OR an animation is in flight
- **Extracted `needsAnimationRedraw`**: pure static method with boolean params — easy to reason about and test independently
- **Comprehensive animation source coverage**: shell panel, AI panel, log-pin slide, canvas overlay, caption typing/fading, notification wave, recorded-keystroke highlight — I verified these against the field declarations and they cover all the time-sensitive visual elements
- **Deliberate omissions are reasonable**: `drawOverlay` (auto-dismiss bounded by refresh interval ≤500ms) and `helpOverlay` (static, no tick-based animation) are correctly excluded
- **Good test coverage**: idle-tick-no-redraw and each-individual-flag tests verify the core invariant

One minor observation: the `drawOverlay.tick(now)`, `captionOverlay.tick(now)`, and `recordingManager.tickRecentKeys(now)` calls still happen every tick (updating internal animation state), but the frame is only repainted when `needsAnimationRedraw` returns true. This is correct — the internal state machine advances, and the visible check ensures repainting happens when there's something to show.

Nice reduction from ~25 fps to ~2 fps when idle while keeping animations smooth. LGTM ✅

davsclaus (approved):

PR #24616, opened 2026-07-12
CAMEL-24013: Fix flaky RawMailMessageTest

gnodet:
## Summary

Fix flaky `RawMailMessageTest` (12.4% failure rate in CI) caused by two race conditions:

1. **Wrong mailbox user in IMAP raw route**: The route for raw IMAP messages was consuming from `jonesImap` instead of `jonesRawImap`. This meant two routes (raw IMAP and normal IMAP) competed for the same single message in the `jonesImap` mailbox, both with `delete=true`. Whichever route polled first consumed and deleted the message, leaving the other route with an empty mailbox and a failing assertion.

2. **Premature `Mailbox.clearAll()` in `testGetRawJavaMailMessage`**: This call wiped all mailbox contents (including messages prepared by `setupResources()`) before the consumer routes for other tests had necessarily consumed their messages, causing intermittent failures depending on test execution order.

## Changes

- Fixed the IMAP raw message route to consume from `jonesRawImap` (matching the POP3 raw route which correctly uses `jonesRawPop3`)
- Removed the `Mailbox.clearAll()` call from `testGetRawJavaMailMessage` (cleanup is already handled by `cleanupResources()`)

## Test plan

- [x] `mvn test -pl components/camel-mail -Dtest=RawMailMessageTest` — all 5 tests pass
- [ ] CI build passes

### DIFF
--- a/components/camel-mail/src/test/java/org/apache/camel/component/mail/RawMailMessageTest.java
+++ b/components/camel-mail/src/test/java/org/apache/camel/component/mail/RawMailMessageTest.java
@@ -69,8 +69,6 @@ public class RawMailMessageTest extends CamelTestSupport {
 
     @Test
     public void testGetRawJavaMailMessage() throws Exception {
-        Mailbox.clearAll();
-
         Map<String, Object> map = new HashMap<>();
         map.put("To", davsclaus.getEmail());
         map.put("From", "jstrachan@apache.org");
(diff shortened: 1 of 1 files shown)


gnodet (commented):

### Review: CAMEL-24013 — Fix flaky RawMailMessageTest

**Changes** (1 file, +1/−3):
1. `jonesImap` → `jonesRawImap` in the IMAP raw message route
2. Removed premature `Mailbox.clearAll()` from `testGetRawJavaMailMessage`

**Assessment:**
This is a genuine bug fix, not just a timing adjustment:

1. **Wrong mailbox**: The IMAP raw route was consuming from `jonesImap` instead of `jonesRawImap`, causing two routes (raw IMAP and normal IMAP) to compete for the same single message with `delete=true`. Whichever polled first won, leaving the other empty. The POP3 raw route correctly used `jonesRawPop3`, making this clearly a copy-paste error. At 12.4% failure rate, this was a very real issue.

2. **Premature cleanup**: `Mailbox.clearAll()` at the start of `testGetRawJavaMailMessage` could wipe messages that other test methods' consumer routes hadn't yet consumed. `cleanupResources()` already handles cleanup at the right lifecycle point.

Both fixes are minimal and correct. The mailbox user fix is the key change — it eliminates the race between competing consumers entirely.

Looks good ✓

davsclaus (approved):

PR #24618, opened 2026-07-12
CAMEL-24015: Fix flaky VertxPlatformHttpOAuthProfileTest

gnodet:
## Summary

- Replace the low-level Vert.x `HttpClient` with `RestAssured` in the `rejectsMissingBearerTokenBeforeBodyHandler` test method (consistent with the other three tests in this class)
- The Vert.x client test used `sendHead()` with `Content-Length: 1024` to send only HTTP headers without a body; the server correctly rejected with 401, but the response body "Unauthorized" was sometimes received as empty by the client due to a race condition in Vert.x response body delivery on half-open connections
- This produced intermittent `expected: <Unauthorized> but was: <>` failures (7% failure rate -- 7 failed, 21 flaky out of 247 CI runs in the last 2 weeks)

## Test plan

- [x] `mvn test -pl components/camel-platform-http-vertx -Dtest=VertxPlatformHttpOAuthProfileTest` passes (4/4 tests, 0 failures)
- [x] Verified stability with 5 consecutive local runs (all pass)

### DIFF
--- a/components/camel-platform-http-vertx/src/test/java/org/apache/camel/component/platform/http/vertx/VertxPlatformHttpOAuthProfileTest.java
+++ b/components/camel-platform-http-vertx/src/test/java/org/apache/camel/component/platform/http/vertx/VertxPlatformHttpOAuthProfileTest.java
@@ -16,12 +16,8 @@
  */
 package org.apache.camel.component.platform.http.vertx;
 
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import io.vertx.core.http.HttpClient;
-import io.vertx.core.http.HttpMethod;
 import org.apache.camel.CamelContext;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.platform.http.PlatformHttpConstants;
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

gnodet (commented):

### Review: CAMEL-24015 — Fix flaky VertxPlatformHttpOAuthProfileTest

**Changes** (1 file, +9/−42):
- Replace low-level Vert.x `HttpClient` + `sendHead()` + `CompletableFuture` machinery with RestAssured `given().body("hello").post("/secure").then().statusCode(401).body(equalTo("Unauthorized"))`
- Remove 4 unused imports and the `client.close()` cleanup in `finally`

**Assessment:**
- Root cause is well-identified: `sendHead()` with `Content-Length: 1024` created a half-open connection where the 401 response body was sometimes delivered as empty by Vert.x — 7% flake rate confirms the race
- RestAssured is already used by the other 3 tests in this class — good consistency
- **Test intent preserved**: The test still validates that OAuth rejection (401) happens before the route body handler. `assertEquals(0, routeInvocations.get())` confirms the route was never invoked, even though RestAssured now sends a complete body (the server rejects at the OAuth layer before the body reaches the route)
- Net -33 lines — significant simplification while maintaining the same coverage

Looks good ✓

PR #24621, opened 2026-07-12
CAMEL-24020: Fix flaky SjmsConnectionRecoveryTest

gnodet:
## Summary

Fix flaky `SjmsConnectionRecoveryTest.testRecoveryStopsAfterSuccessfulReconnection` which has a high failure rate on CI: **24 failures and 7 flaky runs out of 254 total runs (12% failure rate)** in the last 12 days ([Develocity test dashboard](https://develocity.apache.org)).

**Root cause:** The test uses `asyncStartListener=true`, which means the SJMS consumer starts asynchronously in the background. Two timing issues cause flaky failures:

1. **Messages lost before consumer subscribes** — Messages sent before the async consumer is subscribed to the queue are lost (delivered to a queue with no active subscriber). This causes the test to hang until the 10-minute Surefire timeout or fail with `JMSException: Failed to create session factory` when the connection itself isn't ready yet.

2. **Fixed-delay wait in Phase 3** — The previous code used `await().pollDelay(4s).until(() -> true)` which is effectively a disguised `Thread.sleep()`. Under CI load, recovery may take longer or shorter than 4 seconds, making the assertion unreliable.

**Failure modes observed on CI:**
- `TimeoutException` after 10 minutes — the test hangs because the message sent in Phase 1 was lost (consumer not yet subscribed)
- `JMSException: Failed to create session factory` — the send itself fails because the broker connection isn't fully established yet

**Fix:**
- **Phases 1 & 5:** Wrap the send+assert cycles in `Awaitility.untilAsserted()` to retry until the async consumer is ready. This is NOT redundant Awaitility wrapping of `MockEndpoint` — we are retrying the _send_ operation itself because messages sent to a queue with no active subscriber are lost.
- **Phase 3:** Replace the fixed-delay wait with proper Awaitility polling: first wait for recovery (connection count increases), then verify stability (connection count stays constant for 3 recovery intervals, proving the recovery loop has stopped).

## Test plan

- [ ] CI passes for `camel-sjms` module
- [ ] `SjmsConnectionRecoveryTest` passes reliably (no more 10-minute timeouts)
- [ ] Recovery loop detection still works (Phase 4 assertion would fail if the bug regresses)

### DIFF
--- a/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/SjmsConnectionRecoveryTest.java
+++ b/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/SjmsConnectionRecoveryTest.java
@@ -130,10 +130,20 @@ public class SjmsConnectionRecoveryTest extends CamelTestSupport {
     public void testRecoveryStopsAfterSuccessfulReconnection() throws Exception {
         MockEndpoint mock = getMockEndpoint(MOCK_RESULT);
 
-        // Phase 1: verify normal consumption (also confirms consumer is fully started)
-        mock.expectedMessageCount(1);
-        template.sendBody(SJMS_QUEUE_NAME, "before-failure");
-        mock.assertIsSatisfied();
+        // Phase 1: verify normal consumption (also confirms consumer is fully started).
+        // With asyncStartListener=true, the consumer starts in the background and may not be
+        // subscribed yet when the first message is sent. The message would be lost (delivered
+        // to a queue with no subscriber). Use Awaitility to retry the send+assert cycle until
+        // the async consumer is ready.
+        // NOTE: This is NOT redundant Awaitility wrapping of MockEndpoint — we are retrying
+        // the send operation itself, not just the assertion wait.
+        await().atMost(30, TimeUnit.SECONDS)
+                .untilAsserted(() -> {
+                    mock.reset();
+                    mock.expectedMessageCount(1);
+                    template.sendBody(SJMS_QUEUE_NAME, "before-failure");
+                    mock.assertIsSatisfied();
+                });
         mock.reset();
 
         // Phase 2: simulate a transient JMS connection exception.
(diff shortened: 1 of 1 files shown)


gnodet (commented):

### Review: CAMEL-24020 — Fix flaky SjmsConnectionRecoveryTest

**Changes** (1 file, +20/−7):
- Wrap Phase 1 and Phase 5 send+assert cycles with `Awaitility.untilAsserted()`
- `mock.reset()` at the start of each retry clears previous state

**Assessment:**
- Root cause is well-explained: `asyncStartListener=true` means the SJMS consumer starts in the background. The prior fix (PR #24610 / CAMEL-23887) correctly removed redundant Awaitility wrapping of MockEndpoint assertions, but inadvertently removed the implicit retry that was masking the async startup race.
- **This is NOT redundant Awaitility wrapping** — the `untilAsserted` block retries the *entire operation* including `mock.reset()` + `template.sendBody()`, not just the assertion. If the consumer isn't subscribed, either the send throws (`JMSException: Failed to create session factory`) or the message is lost. The retry loop covers both failure modes.
- High failure rate (23 failures, 8 flaky / 271 runs) confirms a real issue.

**One observation** (not blocking): The inner `mock.assertIsSatisfied()` uses the default 10s latch wait, while the outer `atMost` is also 10s. For the "send throws immediately" case, Awaitility retries quickly and the pattern works well. For the "message lost" case (mock waits full 10s for a message that never arrives), the outer 10s may be tight for a second attempt. If this proves insufficient in practice, consider either reducing the inner mock wait time or increasing the outer atMost.

Looks good ✓

gnodet (commented):

### Re-review: force-pushed commit `6c285034`

Improvements over the previous version:

1. **Phases 1 & 5**: `atMost` increased from 10s to 30s — addresses my earlier observation about the inner MockEndpoint 10s latch wait potentially consuming the entire outer Awaitility timeout. With 30s, there's room for 2–3 retries even with 10s inner waits.

2. **Phase 3** (major improvement): The previous code had `await().pollDelay(4 * interval).until(() -> true)` — effectively a disguised `Thread.sleep`. Now properly refactored into:
   - **Condition-based wait**: `until(() -> countingFactory.createCount.get() > connectionsBefore)` — waits for recovery to actually create a new connection
   - **Stability assertion**: captures `countAfterRecovery`, then verifies the count stays stable over 3 recovery intervals via `untilAsserted(() -> assertEquals(countAfterRecovery, countingFactory.createCount.get(), ...))` — detects the bug (extra connections) with a clear failure message

3. Added explicit `NOTE` comment about Awaitility not being redundant wrapping.

4. Used `3L * RECOVERY_INTERVAL_MS` — `long` literals prevent potential `int` overflow.

All concerns from my previous review are addressed. Looks good ✓

gnodet:
Recreating this PR from the main repo branch — the fork-based branch wasn't triggering the CI build workflow.

PR #24625, opened 2026-07-12
CAMEL-24026: Fix flaky XsltFromFileExceptionTest

gnodet:
## Summary

Fix flaky `XsltFromFileExceptionTest` by reordering assertions to wait for the file consumer to complete before checking mock expectations.

**Root cause**: The test called `assertMockEndpointsSatisfied()` before `oneExchangeDone.matchesWaitTime()`. On slow CI machines, the file consumer may not have picked up and processed the file yet when the mock assertion starts, causing intermittent timeout failures.

**Fix**:
- Wait for `oneExchangeDone` latch first (confirms file consumer processed the file)
- Then assert mock expectations (which now pass immediately since the exchange already completed)
- Assert the `oneExchangeDone` result with `assertTrue()` for better failure diagnostics

**Develocity data**: 6 failures out of recent runs.

## Test plan

- [x] `XsltFromFileExceptionTest` passes locally (2 tests, 0 failures)
- [ ] CI build passes

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
@@ -21,6 +21,8 @@ import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 /**
  *
  */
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Self-review: clean fix for the assertion ordering race condition.

**What was wrong**: `assertMockEndpointsSatisfied()` was called *before* `oneExchangeDone.matchesWaitTime()`, so the mock assertion could time out waiting for messages that the file consumer hadn't processed yet. The `oneExchangeDone.matchesWaitTime()` call was also dead code — its boolean return value was ignored.

**What the fix does right**:
- Waits for `oneExchangeDone` first (exchange completion latch), then asserts mock expectations — correct ordering since mock expectations pass instantly once the exchange is done
- Wraps `matchesWaitTime()` in `assertTrue()` for better failure diagnostics instead of silently ignoring the return value
- Adds clear comments explaining the wait rationale

LGTM ✅

davsclaus (approved):

PR #24626, opened 2026-07-12
CAMEL-24027: Fix flaky InOutQueueProducerAsyncLoadTest

gnodet:
## Summary

Fix flaky `InOutQueueProducerAsyncLoadTest` by reducing the message count and improving synchronization.

**Root cause**: The test sent 5000 InOut JMS request-reply messages via a 2-thread executor with only a 30-second Awaitility timeout. On slow CI machines, processing 5000 round-trip messages through an embedded ActiveMQ broker frequently exceeded the timeout. Additionally, test failures inside the executor threads were silently swallowed.

**Fix**:
- Reduce message count from 5000 to 500 (still validates concurrent InOut load behavior)
- Replace `Awaitility.await().until(executor::isTerminated)` with a `CountDownLatch` for precise completion tracking
- Add `AtomicInteger` to count and assert zero failures (previously swallowed)
- Use `executor.awaitTermination()` with explicit 10s timeout instead of polling
- Increase overall completion timeout to 60 seconds
- Modernize anonymous `Runnable` to lambda

## Test plan

- [x] `InOutQueueProducerAsyncLoadTest` passes locally (1 test, 0 failures, 4.7s)
- [ ] CI build passes

### DIFF
--- a/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/InOutQueueProducerAsyncLoadTest.java
+++ b/components/camel-sjms/src/test/java/org/apache/camel/component/sjms/producer/InOutQueueProducerAsyncLoadTest.java
@@ -16,8 +16,10 @@
  */
 package org.apache.camel.component.sjms.producer;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import jakarta.jms.JMSException;
 import jakarta.jms.Message;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Self-review: solid fix addressing multiple issues in the load test.

**Key improvements**:
1. **Message count 5000→500**: Still validates concurrent InOut request-reply behavior. 5000 round-trip JMS messages through an embedded broker with 2 threads was unrealistic for CI timeouts.
2. **CountDownLatch over Awaitility**: More precise — counts individual message completions rather than polling `executor.isTerminated()`. The latch fires the instant all 500 messages are done.
3. **Failure tracking via AtomicInteger**: Critical fix — the old code caught exceptions inside the Runnable but never surfaced them. Tests could "pass" with 100% failure rate inside the executor threads. Now the assertion at the end catches that.
4. **`executor.awaitTermination()` with explicit timeout**: Clean shutdown with a proper bound instead of Awaitility polling.
5. **Lambda modernization**: Minor but welcome cleanup.

The timeout structure is well-layered: 60s for message processing (latch), then 10s for executor shutdown — clear separation of concerns.

LGTM ✅

davsclaus (approved):

PR #24632, opened 2026-07-12
CAMEL-24031: Fix flaky JmsAddAndRemoveRouteManagementIT

gnodet:
## Summary

Fix flaky `JmsAddAndRemoveRouteManagementIT` identified by [Develocity](https://develocity.apache.org).

The test dynamically adds a JMS route, sends a message, then removes the route and verifies thread pool cleanup via JMX. On slow CI, both JMX registration and cleanup can take longer than the original 5s timeouts.

Changes:
- Increase Awaitility timeouts from 5s to 30s for JMX thread pool registration and cleanup checks
- Set 30s result wait time on MockEndpoint
- Wait for JMS consumer to actually subscribe before sending the test message

## Test plan

- [x] Test passes locally with `mvn verify -pl components/camel-jms -Dtest=JmsAddAndRemoveRouteManagementIT -Dit.test=JmsAddAndRemoveRouteManagementIT`
- [ ] CI passes

### DIFF
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/JmsAddAndRemoveRouteManagementIT.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/JmsAddAndRemoveRouteManagementIT.java
@@ -90,9 +90,10 @@ public class JmsAddAndRemoveRouteManagementIT extends AbstractJMSTest {
             }
         });
 
-        // Wait for the new route's thread pool to be registered and identify it
+        // Wait for the new route's thread pool to be registered and identify it.
+        // JMX registration can be slow on CI, so use a generous timeout.
         AtomicReference<Set<ObjectName>> duringRef = new AtomicReference<>();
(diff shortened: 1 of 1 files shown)


gnodet (commented):

**LGTM** — straightforward Awaitility timeout increase for JMX operations on slow CI.

The test dynamically adds/removes a JMS route and verifies thread pool registration/cleanup via JMX MBean queries. Both Awaitility blocks increased from 5s to 30s — Awaitility still exits as soon as the condition is met, so the longer upper bound only matters on slow CI environments where JMX registration or cleanup can lag.

Minor note: the PR description mentions "Set 30s result wait time on MockEndpoint" and "Wait for JMS consumer to actually subscribe before sending the test message" but those changes don't appear in the diff. Consider updating the description to match the actual changeset (Awaitility timeout increases only).

Good fix — the test's core assertions are unchanged, only the tolerance for CI slowness is improved.

davsclaus (approved):

PR #24635, opened 2026-07-13
CAMEL-24037: Fix flaky camel-core tests using timed assertions

gnodet:
## Summary

Fix three flaky camel-core tests identified by Develocity:

- **TwoSchedulerTest** (4 flaky/70 builds = 6%): Replace `assertMockEndpointsSatisfied()` with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)`. Also fix `assertSame(tn1, tn2)` → `assertEquals(tn1, tn2)` since `assertSame` checks reference identity, not value equality, for String objects.
- **DistributedTimeoutTest** (4 flaky/70 builds = 6%): Remove Awaitility wrapping around `MockEndpoint.assertIsSatisfied()` (anti-pattern per project guidelines — MockEndpoint already waits internally via CountDownLatch). Use direct `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` for both contexts. Increase Awaitility `atMost` from 2s to 5s for the aggregation timeout wait.
- **ThrottlingExceptionRoutePolicyOpenViaConfigTest** (11 flaky/275 builds = 4%): Replace three `assertMockEndpointsSatisfied()` calls with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)`.

## Test plan

- [x] All three tests pass locally after changes
- [ ] CI build passes

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/scheduler/TwoSchedulerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/scheduler/TwoSchedulerTest.java
@@ -16,11 +16,14 @@
  */
 package org.apache.camel.component.scheduler;
 
+import java.util.concurrent.TimeUnit;
+
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class TwoSchedulerTest extends ContextTestSupport {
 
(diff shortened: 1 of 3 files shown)


gnodet (commented):

**LGTM** — three well-targeted flaky-test fixes.

- **TwoSchedulerTest**: Standard timed assertion switch. The `assertSame` → `assertEquals` fix is a good catch — `assertSame` checks reference identity (`==`), but String values retrieved from exchange headers may not be the same object reference even when equal by value. `assertEquals` using `.equals()` is correct here.

- **DistributedTimeoutTest**: Correctly removes the Awaitility wrapping of `MockEndpoint.assertIsSatisfied()` — this is the anti-pattern called out in project guidelines (MockEndpoint already waits via CountDownLatch; Awaitility adds unnecessary polling overhead). The direct `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` for both contexts is the right approach. The Awaitility timeout increase from 2s to 5s for the `invoked.get() == 1` aggregation timeout check is also sensible for slow CI.

- **ThrottlingExceptionRoutePolicyOpenViaConfigTest**: Clean replacement of three `setResultWaitTime(X)` + `assertMockEndpointsSatisfied()` blocks with consistent `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)`. Simpler and more consistent.

davsclaus (approved):

PR #24636, opened 2026-07-13
CAMEL-24038: Fix flaky AsyncWiretapTest in telemetry modules

gnodet:
## Summary

Fix flaky `AsyncWiretapTest` in both `camel-telemetry` and `camel-telemetry-dev` modules. Both tests had ~5-7% flaky rate in CI.

**Root cause:** The tests send 10 messages through an async wiretap route with a 2-second delay. Traces are written asynchronously after exchange processing. The assertions on trace data need to wait for the async writing to complete.

**Changes:**
- **camel-telemetry AsyncWiretapTest**: Replace `Thread.sleep(10000)` with Awaitility polling until all 10 traces with 7 spans each arrive. Replace `mock.setAssertPeriod(5000)` + `mock.assertIsSatisfied(1000)` with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)`.
- **camel-telemetry-dev AsyncWiretapTest**: Add Awaitility polling for trace completion (previously had no wait at all). Same mock assertion fix.

## Test plan

- [x] Both tests pass locally after changes
- [ ] CI build passes

### DIFF
--- a/components/camel-telemetry-dev/src/test/java/org/apache/camel/telemetrydev/AsyncWiretapTest.java
+++ b/components/camel-telemetry-dev/src/test/java/org/apache/camel/telemetrydev/AsyncWiretapTest.java
@@ -16,9 +16,9 @@
  */
 package org.apache.camel.telemetrydev;
 
-import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.CamelContextAware;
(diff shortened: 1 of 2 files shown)


gnodet (commented):

**LGTM** — good fix eliminating `Thread.sleep(10000)` and improving assertion reliability in both telemetry test modules.

**camel-telemetry `AsyncWiretapTest`**: Correctly replaces the 10-second `Thread.sleep` with Awaitility polling until all 10 traces with 7 spans each arrive. This is the right pattern — traces are written asynchronously after exchange processing, so polling for trace completion is appropriate (NOT wrapping MockEndpoint — the Awaitility here waits on trace data, not mock expectations). Also fixes the mock assertion from `setAssertPeriod(5000)` + `assertIsSatisfied(1000)` to `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)`.

**camel-telemetry-dev `AsyncWiretapTest`**: Same pattern applied. Previously had no explicit wait for async trace completion — the original timing-based mock assertions likely masked the race condition. Adding the Awaitility trace-completion wait makes the test deterministic.

Both changes follow project conventions correctly: Awaitility for condition-based waits on non-MockEndpoint state, timed MockEndpoint assertions for mock expectations.

davsclaus (approved):

PR #24637, opened 2026-07-13
CAMEL-24003: Fix thread-safety in BacklogTracer message history queue

davsclaus:
## Summary

- Fix race condition in `BacklogTracer.traceEvent()` where concurrent calls from multiple threads (e.g. timer thread and Kafka consumer threads) could corrupt the message history queue state
- The compound queue operations (peek/offer/addAll/clear) on `provisionalHistoryQueue` and `completeHistoryQueue` must be atomic — without synchronization, interleaving threads can cause events to be dropped and the complete history to become permanently stale

## Root Cause

When a route topology includes downstream routes connected via Kafka/SEDA, the originating route's thread and the downstream consumer threads call `traceEvent()` concurrently. The race window:

1. E0 completes on timer thread → transfer to complete queue → provisional cleared
2. Downstream route (Kafka consumer thread) starts → its first event enters provisional as head
3. E1 fires on timer thread → head belongs to downstream route (different breadcrumbId) → E1 events fail match check → **all E1 events dropped**
4. Complete history queue never gets updated with new exchange data

This was observed as the TUI Inspect tab showing 20-hour-old data from the first successful exchange, despite thousands of subsequent exchanges completing.

## Fix

Add `synchronized (historyLock)` around the message history section of `traceEvent()`. Only the history queue management is synchronized — the main backlog queue and activity queue processing remain lock-free. The lock scope is minimal (local variable reads and queue operations).

## Test plan

- [x] All 21 `BacklogTracer*Test` tests pass
- [ ] Manual verification: run a multi-route topology with Kafka/SEDA downstream routes and confirm the Inspect tab shows recent exchange data

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogTracer.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogTracer.java
@@ -78,6 +78,7 @@ public class BacklogTracer extends ServiceSupport implements org.apache.camel.sp
     private final ActivityEventNotifier activityEventNotifier = new ActivityEventNotifier();
     private int activitySize = 100;
     private static final long INFLIGHT_EVICTION_MILLIS = 5 * 60 * 1000;
+    private final Object historyLock = new Object();
     private volatile String lastCompletedBreadcrumbId;
     private boolean removeOnDump = true;
     private int bodyMaxChars = 32 * 1024;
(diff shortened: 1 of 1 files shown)


gnodet (approved):

**LGTM — Approved.**

Well-analyzed race condition and clean fix. The root cause is clearly explained: concurrent `traceEvent()` calls from different threads (timer + Kafka/SEDA consumer threads) interleave the compound peek → check → offer → addAll → clear sequence, causing events to be dropped and the complete history to go stale.

The fix is well-scoped:
- **Dedicated lock object** (`private final Object historyLock`) — good practice, avoids exposing the lock via `synchronized(this)`.
- **Minimal synchronization scope** — only the message history queue management is synchronized. The main backlog queue and activity queue processing remain lock-free, so there's no contention impact on the hot path.
- **Reader-side safety is already provided** by `LinkedBlockingQueue` — the snapshot operations at lines 478 and 547 (`Collections.unmodifiableCollection(completeHistoryQueue)` and `new ArrayList<>(completeHistoryQueue)`) use the queue's own thread-safe iteration, so they don't need the lock. The lock only needs to protect the compound write sequence.
- **`volatile` on `lastCompletedBreadcrumbId`** is retained — harmless belt-and-suspenders since all accesses are now inside the lock, but protects against future reads outside it.

Logic is unchanged — purely a synchronization wrapper addition.

PR #24650, opened 2026-07-13
CAMEL-24042: Fix flaky tests in camel-core (batch 10)

gnodet:
## Summary

Fix flaky tests identified by [Develocity](https://develocity.apache.org) in camel-core:

- **ThrottlingExceptionRoutePolicyOpenViaConfigTest**: Increase Awaitility timeouts from 5s to 10s
- **DistributedTimeoutTest**: Increase Awaitility timeout from 5s to 10s
- **MockEndpointTimeClauseTest**: Replace 8× bare `assertMockEndpointsSatisfied()` with `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` (2 intentional bare calls left: assertThrows and timing-critical test)
- **FileProducerNoForcedWritesTest**: Replace `assertMockEndpointsSatisfied()` with timed variant
- **StopTimeoutRouteTest**: Replace `assertMockEndpointsSatisfied()` with timed variant
- **ForegroundTaskTest**: Increase `@Timeout` from 10s to 20s (slow test predicates sleep 2s/iteration, tight under CI load)
- **ForegroundTimeTaskTest**: Increase `@Timeout` from 10s to 20s

## Test plan

- [x] All 28 test methods pass locally (0 failures)

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/file/FileProducerNoForcedWritesTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/file/FileProducerNoForcedWritesTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.file;
 
+import java.util.concurrent.TimeUnit;
+
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
@@ -33,7 +35,7 @@ public class FileProducerNoForcedWritesTest extends ContextTestSupport {
 
         template.sendBodyAndHeader(fileUri(), "Hello World", Exchange.FILE_NAME, "hello.txt");
 
-        assertMockEndpointsSatisfied();
+        MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS);
 
         assertFileExists(testFile("output.txt"));
         assertEquals("Hello World",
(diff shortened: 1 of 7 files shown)


davsclaus (approved):

oscerd (commented):
Reviewed each changed test for the project's async-testing rules (no `Thread.sleep`, condition-based waits, MockEndpoint timed assertions, no Awaitility-wrapping of `assertIsSatisfied`) — all clean, no change requests. Notes from the verification:

- The `assertMockEndpointsSatisfied()` → `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` migrations are semantically the same latch-based wait with the implicit 10s default raised to an explicit 30s — exactly the pattern the repo guidelines prescribe. The two bare `assertIsSatisfied()` calls deliberately left in `MockEndpointTimeClauseTest` are correct as-is (one is inside `assertThrows`, the other gates a latch on the default wait).
- The `@Timeout(10)` → `@Timeout(20)` bumps in `ForegroundTaskTest`/`ForegroundTimeTaskTest` are pure hang-guards: task time-boundedness is still asserted by each task's own `withMaxDuration` budget and iteration-count assertions, so nothing is weakened. The arithmetic checks out (slow-predicate tests have ~6s nominal runtime, leaving only ~4s headroom under the old value).
- No assertion is weakened anywhere in the batch: expected message counts, exactly-once invocation checks (`invoked.get() == 1` in `DistributedTimeoutTest`), and upper/lower-bound duration asserts are all unchanged.
- One observation for the series rather than this PR: the whole batch is timeout-headroom on waits that prior batches already made condition-based — which is the right remaining lever. `DistributedTimeoutTest:54` is now on its second escalation (2s→5s in CAMEL-24037/#24635, 5s→10s here) for an aggregator with a 200ms completion timeout; if it flakes a third time, it may deserve a look at what actually stalls the timeout checker on loaded CI instead of another bump.

CI is green on both JDK 17 and 25 builds (which run the camel-core tests).

PR #24652, opened 2026-07-13
[backport camel-4.18.x] CAMEL-23768: camel-keycloak - select the JWKS verification key by the token kid

oscerd:
## Backport of #24527

Cherry-pick of #24527 onto `camel-4.18.x` (clean pick, no conflicts; camel-keycloak module tests green on 4.18.x; full reactor `mvn clean install -DskipTests` green).

**Original PR:** #24527 - CAMEL-23768: camel-keycloak - select the JWKS verification key by the token kid
**Original author:** @oscerd
**JIRA:** https://issues.apache.org/jira/browse/CAMEL-23768 (fixVersions already lists 4.18.4)

### Original description

Fixes [CAMEL-23768](https://issues.apache.org/jira/browse/CAMEL-23768).

## Problem

`KeycloakPublicKeyResolver` ignored the JWT header `kid` and returned the **first** key in the JWKS, and `KeycloakSecurityProcessor` never passed the kid through (`getPublicKey(null)`). During key rotation (multiple keys present) this could pick the wrong key and reject an otherwise-valid token. The token is still cryptographically verified against a real key, so this is a **correctness/availability** matter, not a bypass.

## Change

- `KeycloakSecurityHelper.extractKeyId(token)` — pulls the `kid` from the JWS header using Keycloak's own `JWSInput` (existing dependency); returns `null` if absent/unparseable.
- `KeycloakSecurityProcessor` — extracts the kid and passes it to `resolver.getPublicKey(kid)`.
- `KeycloakPublicKeyResolver.getPublicKey` — forces one refresh when the kid is unknown (to pick up rotated keys), then delegates to a new package-visible `selectKey` that **fails closed** (throws) when a kid is given but no key matches. The first-key fallback is kept only when the token carries no kid.

## Tests

- `KeycloakPublicKeyResolverTest` — matching kid selects that key (not the first); unknown kid fails closed; null kid falls back to the first key (JWKS built from generated RSA keys, no container).
- `KeycloakSecurityHelperTest` — `extractKeyId` returns the header kid / `null` for a malformed token.
- Full reactor build (`mvn clean install -DskipTests`, 1871 modules) green; no generated-file impact (plain Java).

## Compatibility / backport

Correctness fix, `fixVersions` 4.18.4 / 4.22.0 (camel-keycloak does not exist on the 4.14.x line). Default behaviour for a single-key JWKS is unchanged.

---

### DIFF
--- a/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakPublicKeyResolver.java
+++ b/components/camel-keycloak/src/main/java/org/apache/camel/component/keycloak/security/KeycloakPublicKeyResolver.java
@@ -57,7 +57,8 @@ public class KeycloakPublicKeyResolver {
     /**
      * Gets the public key for verifying JWT signatures. Keys are cached and refreshed periodically.
      *
-     * @param  kid         the key ID from the JWT header (optional, uses first key if null)
+     * @param  kid         the key ID from the JWT header; the matching key is selected, and resolution fails closed
+     *                     when a kid is given but no key matches (a {@code null} kid falls back to the first key)
      * @return             the public key
      * @throws IOException if fetching keys fails
      */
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean cherry-pick of merged PR #24527 (main) onto camel-4.18.x. The fix correctly selects the JWKS verification key by the token's `kid` header instead of always using the first key, and fails closed when a `kid` is present but no matching key exists.

Security analysis:
- Tokens are now verified against the correct JWKS key matching the kid claim, preventing verification against the wrong key during key rotation.
- Fail-closed behavior (IOException when kid present but no match after forced refresh) is the correct security posture.
- Null-kid fallback to first key preserves backward compatibility for older tokens or IdPs that omit kid.
- Single forced refresh on unknown kid handles key rotation without introducing DoS risk against the JWKS endpoint.
- Test coverage is thorough: three selectKey scenarios + two extractKeyId edge cases.

PR #24653, opened 2026-07-13
chore: Adapt camel-pqc to Bouncy Castle 1.85 algorithm renames

davsclaus:
## Summary

_Claude Code on behalf of davsclaus_

Bouncy Castle 1.85 removed the pre-standardization PQC algorithm names (DILITHIUM, KYBER, SPHINCSPLUS) from the BCPQC provider. This PR adapts camel-pqc to use the NIST standardized names:

- **DILITHIUM** → `ML-DSA` (BC provider) — FIPS 204
- **KYBER** → `ML-KEM` (BC provider) — FIPS 203
- **SPHINCSPLUS** → `SLH-DSA` (BC provider) — FIPS 205

The Camel enum constants (`signatureAlgorithm=DILITHIUM`, `keyEncapsulationAlgorithm=KYBER`, etc.) are preserved unchanged for backward compatibility — existing routes continue to work transparently.

Changes:
- Updated `PQCSignatureAlgorithms.DILITHIUM` and `.SPHINCSPLUS` to use new JCA names and BC provider
- Updated `PQCKeyEncapsulationAlgorithms.KYBER` to use new JCA name and BC provider
- Updated default parameter specs from old BC classes (`DilithiumParameterSpec`, `KyberParameterSpec`, `SPHINCSPlusParameterSpec`) to standardized ones (`MLDSAParameterSpec`, `MLKEMParameterSpec`, `SLHDSAParameterSpec`)
- Updated all lifecycle managers (File, AWS, Hashicorp) and crypto material classes
- Updated all tests

This is a companion to PR #24648 (Bouncy Castle 1.84 → 1.85 version bump).

## Test plan

- [x] All 168 camel-pqc tests pass (including `PQCEndToEndIntegrationTest` which was failing)
- [x] Code compiles and formats correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24654, opened 2026-07-13
CAMEL-23846: Camel-PQC - make the NIST parameter set configurable on the endpoint

oscerd:
# CAMEL-23846: Camel-PQC — make the NIST parameter set configurable on the endpoint

## Motivation

The default key material classes hardcode a single parameter set per algorithm — `PQCDefaultMLDSAMaterial`
uses **ML-DSA-65**, ML-KEM defaults to **ML-KEM-512**, Falcon to 512 — and the lifecycle
`generateKeyPair` path only called the 2-arg `KeyLifecycleManager.generateKeyPair(algorithm, keyId)`.
To use any other NIST security level (ML-DSA-44/87, ML-KEM-768/1024, the SLH-DSA and Falcon variants)
users had to register their own `KeyPair`/`Signature`/`KeyGenerator` beans. There was no declarative
endpoint option.

## What this does

Adds a **`parameterSpec`** endpoint option (`@UriParam`, label `advanced`).

- **`PQCParameterSpecResolver`** (new) maps an algorithm + parameter-set name to the BouncyCastle
  `AlgorithmParameterSpec`, covering **16 algorithms**: MLDSA, SLHDSA, FALCON, DILITHIUM, SPHINCSPLUS,
  PICNIC, MLKEM, KYBER, NTRU, NTRULPRime, SNTRUPrime, BIKE, HQC, CMCE, FRODO, SABER.
- **`PQCComponent`** generates the key material with the configured parameter set instead of the
  hardcoded default material, and **caches it per (algorithm, parameterSpec)** so that endpoints
  sharing the same configuration — e.g. a `sign` and a `verify` endpoint — use the **same key**,
  mirroring the semantics of the shared default material. (Without this, each endpoint would get a
  fresh key and verification would fail.)
- **`generateKeyPair`** lifecycle operation now uses the **parameterSpec-aware** `KeyLifecycleManager`
  overload.

### Name handling

Values are BouncyCastle's parameter-set names, resolved **case-insensitively**; the underscore form of
the BC constants (`ml_dsa_87`) is accepted as an alias of the canonical `ML-DSA-87`. Note BC is not
uniform here — the modern specs *throw* on an unknown name while the older ones *return `null`* — both
are normalised into one clear `IllegalArgumentException`.

Rebased on the **BouncyCastle 1.85** rename (#24653): `DILITHIUM`, `SPHINCSPLUS` and `KYBER` are now
backed by the standardized **ML-DSA / SLH-DSA / ML-KEM** algorithms, so they take those standardized
parameter sets (e.g. `ML-DSA-65` for `DILITHIUM`) rather than the removed pre-standardization ones.
There is a regression test for exactly this.

### Out of scope (rejected with a clear error)

`parameterSpec` is not supported for the **stateful** hash-based signatures (XMSS, XMSSMT, LMS, HSS),
for **MAYO/SNOVA** (no name-addressable BC parameter spec), nor for the **hybrid** operations (they pair
a fixed classical key set with a matching PQC key set). Those still use registry-supplied beans.

## Testing

- `PQCParameterSpecResolverTest` (8 tests): resolves signature and KEM parameter sets, case-insensitive
  and underscore-alias handling, the BC-1.85 standardized aliases for DILITHIUM/SPHINCSPLUS/KYBER,
  `isSupported`, unsupported-algorithm rejection, and unknown-name rejection for **both** the throwing
  and the `null`-returning BC spec classes.
- `PQCParameterSpecTest` (6 tests): `parameterSpec` drives the actual key size (ML-DSA-44 < 65 < 87, and
  ML-KEM-512 < 1024), endpoints with the same algorithm+spec **share the key**, a full sign→verify round
  trip with a non-default parameter set, and rejection of XMSS and of hybrid operations.

All 14 green; full reactor build green (catalog + endpoint-DSL + component-DSL regenerated).

## Documentation

`pqc-component.adoc` gains a *Parameter Sets (NIST Security Levels)* section with an example, the
key-sharing semantics, a per-algorithm table of parameter-set names, and the unsupported cases.

## Backport

None — new feature, `main` only (4.22.0).

---


gnodet (approved):
Well-structured new feature adding a `parameterSpec` endpoint option to camel-pqc for selecting NIST parameter sets. The code is correct, properly tested (13 tests), documented, and follows existing project conventions.

**Scanner coverage:** No standalone static analysis tools (PMD, Checkstyle, semgrep) available in the environment. Rely on CI for comprehensive static analysis.

Minor notes:
- The `configureParameterSpecMaterial` method registers BouncyCastle providers via `Security.addProvider()` (a JVM-global side effect) — consistent with the existing default material classes.
- `SUPPORTED_ALGORITHMS` uses mixed-case entries matching enum constants exactly, which is correct since algorithm names always come from `valueOf()`.
- The `resolveKeyPair` caching uses a correct check-then-`putIfAbsent` pattern on `ConcurrentHashMap`.
- Test coverage is good: `PQCParameterSpecResolverTest` (7 tests) + `PQCParameterSpecTest` (6 tests).
- Documentation in `pqc-component.adoc` includes parameter-set table, examples, key-sharing semantics.
- No upgrade guide entry needed (additive feature, no breaking changes).

davsclaus (approved):

oscerd:
Rebased onto `main` after the BouncyCastle 1.85 rename (#24653).

That commit re-points Camel's `DILITHIUM` → **ML-DSA**, `SPHINCSPLUS` → **SLH-DSA** and `KYBER` → **ML-KEM** (provider `BC`). `PQCParameterSpecResolver` was still handing those three the pre-standardization `DilithiumParameterSpec`/`SPHINCSPlusParameterSpec`/`KyberParameterSpec` — which still compiles, but would have been the wrong spec type for the new generators at runtime. They now resolve via `MLDSAParameterSpec`/`SLHDSAParameterSpec`/`MLKEMParameterSpec`, with a regression test (`testPreStandardizationAliasesUseStandardizedParameterSets`) pinning it, and the docs table updated.

14 tests green, full reactor build green.

PR #24655, opened 2026-07-13
CAMEL-24044: camel-sql - Fix Postgres aggregation repositories binding version without a placeholder

Croway:
JIRA: https://issues.apache.org/jira/browse/CAMEL-24044

Since the CAMEL-7810 optimistic-locking refactor (Camel 3.4), `PostgresAggregationRepository.insert()` built `INSERT INTO t(exchange, id, ...) ON CONFLICT DO NOTHING` without the `version` column while the base `insertHelper` binds blob, key **and version** — one more bound parameter than placeholders, so every `add()` failed with `The column index is out of range` (verified against real PostgreSQL). With `storeBodyAsText=true` the version was silently bound into the `body` placeholder before failing. `ClusteredPostgresAggregationRepository` had the same defect in its completed-table insert (every `remove()` failed at completion) and never wrote the `instance_id` column, so instance-scoped recovery could never match any row.

Changes:
- Include the `version` column in both Postgres `INSERT` statements and pass the actual version instead of the hardcoded `1L`
- Clustered variant additionally writes `instance_id` when `completed && recoveryByInstance`
- New integration tests against real PostgreSQL (`camel-test-infra-postgres`) covering the full add/get/update/remove/recover/confirm lifecycle, `storeBodyAsText`, and instance-scoped recovery with two nodes — these fail on current `main` and pass with the fix
- Fixed the stale `INSERT` example in the component docs and added an upgrade-guide entry (the aggregation tables need the `version BIGINT` column, as `JdbcAggregationRepository` already required for reads/updates)

Run the new ITs with:
`mvn verify -Pit -Dit.test='PostgresAggregationRepositoryIT,ClusteredPostgresAggregationRepositoryIT' -Dtest=NoSuchTest -Dsurefire.failIfNoSpecifiedTests=false`

### DIFF
--- a/components/camel-sql/pom.xml
+++ b/components/camel-sql/pom.xml
@@ -94,6 +94,19 @@
             <version>${mariadb-version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- PostgreSQL only used for the aggregation repository integration tests -->
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>${pgjdbc-driver-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-infra-postgres</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/components/camel-sql/src/main/docs/sql-component.adoc
+++ b/components/camel-sql/src/main/docs/sql-component.adoc
@@ -1040,7 +1040,7 @@ uses special `INSERT .. ON CONFLICT ..` statement to provide optimistic locking 
 This statement is (with default aggregation table definition):
 [source,sql]
 ----
-INSERT INTO aggregation (id, exchange) values (?, ?) ON CONFLICT DO NOTHING
+INSERT INTO aggregation (exchange, id, version) VALUES (?, ?, ?) ON CONFLICT DO NOTHING
 ----
 
 Details can be found https://www.postgresql.org/docs/9.5/sql-insert.html[in PostgreSQL documentation].
--- a/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/ClusteredPostgresAggregationRepository.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/ClusteredPostgresAggregationRepository.java
@@ -56,13 +56,14 @@ public class ClusteredPostgresAggregationRepository extends ClusteredJdbcAggrega
             final CamelContext camelContext, final String correlationId, final Exchange exchange, String repositoryName,
             final Long version, final boolean completed)
             throws Exception {
-        // The default totalParameterIndex is 2 for ID and Exchange. Depending on logic this will be increased
-        int totalParameterIndex = 2;
+        // The default totalParameterIndex is 3 for ID, Exchange and version. Depending on logic this will be increased
+        int totalParameterIndex = 3;
         StringBuilder queryBuilder = new StringBuilder(256)
                 .append("INSERT INTO ").append(repositoryName)
                 .append('(')
                 .append(EXCHANGE).append(", ")
-                .append(ID);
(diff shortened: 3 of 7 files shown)


gnodet (approved):
Correct fix for a long-standing bug where the Postgres aggregation repositories' INSERT statements were missing the `version` column (and the clustered variant also missed `instance_id`), causing parameter-count mismatches on every `add()` call.

**Scanner coverage:** No static analysis tools (PMD, Checkstyle, semgrep) available. Rely on CI.

Key observations:
- The fix correctly aligns the Postgres subclasses' INSERT column lists with the base `insertHelper()` binding order (exchange blob, key, version, and optionally instance_id).
- Passing `version` instead of the hardcoded `1L` is correct: the caller provides `1L` for new inserts and the actual version from `get()` for completed-table inserts.
- The `ClusteredPostgresAggregationRepository` fix for `instance_id` is also correct: `insertHelper()` conditionally binds `instanceId` when `completed && recoveryByInstance`, so the SQL must include it under the same condition.
- No impact on other database dialects — fix is confined to Postgres subclasses.
- No SQL injection risks — all values are parameterized via `PreparedStatement`.
- Integration tests are thorough, covering full lifecycle (add/get/update/remove/scan/recover/confirm), `storeBodyAsText` mode, and multi-node instance-scoped recovery against real PostgreSQL.
- Documentation and upgrade guide updates are appropriate.

davsclaus (approved):

oscerd (commented):
Reviewed with focus on the parameter/placeholder arithmetic, both repository flavors, and the recent deserialization-hardening lineage in these classes — the fix is correct and I have no change requests. Summary of the verification:

**Root cause confirmed against main.** `JdbcAggregationRepository.insertHelper` (`JdbcAggregationRepository.java:336-340`) always binds three base parameters (exchange blob → 1, id → 2, `version` → 3) before the optional body/header ones, while the Postgres subclasses' SQL builders emit only two base columns (`EXCHANGE, ID`) — one more bound parameter than `?` placeholders in every configuration, so `add()` could never succeed on Postgres. The new column order (`EXCHANGE, ID, VERSION, [BODY], [headers...], [instance_id]`) exactly matches the binder's positional order in both the plain and clustered `insertHelper`, and passing the real `version` (instead of a hardcoded value) matches base-class semantics for the completed-table insert path. The clustered variant's added `instance_id` column also makes `scan()`'s `WHERE INSTANCE_ID = ?` able to match rows for the first time. The regression traces to the Camel 3.4-era CAMEL-7810 rework, which introduced the version binding without adding the placeholder — nothing intentional is being reverted.

**Since the broken path could never succeed, there is no behavioral-compat risk** — and the documented default DDL already includes `version BIGINT NOT NULL`, so users following the docs need no schema change. The stale SQL example in `sql-component.adoc` is corrected too (the old line had the wrong column order *and* count).

**Tests**: the two new real-PostgreSQL ITs (plain + clustered, covering add/get/update/remove/scan/recover/confirm, `storeBodyAsText`, and two-node instance-scoped recovery) are exactly what this fix needed — the H2-based unit tests structurally couldn't catch a Postgres-only SQL defect. No `Thread.sleep`; new dependencies are test-scoped only (`org.postgresql:postgresql` via the existing parent property + project-internal `camel-test-infra-postgres`), matching the established `camel-pg-replication-slot` pattern.

**Security cross-check**: the ObjectInputFilter/deserialization-filter work in these classes is untouched (write path only), and values remain `?`-bound — only constant identifiers are concatenated into the SQL.

Minor notes, none blocking:
1. `ClusteredPostgresAggregationRepository` appends the `"instance_id"` string literal because the base-class `INSTANCE_ID` constant is `private` (`ClusteredJdbcAggregationRepository.java:48`) — fine as-is; widening the constant to `protected` would be a reasonable tiny follow-up.
2. The PR body's `mvn verify -Pit` references a profile that doesn't exist in this build — failsafe ITs are bound via the always-active `full` profile, so plain `mvn verify` runs them; description-only inaccuracy.
3. The PR body declares "Generated with Claude Code" but the commit lacks the AI co-authorship trailer — per the repo's AI guidelines, worth adding to the squash-merge message.

The two matrix build jobs were still pending at review time — worth confirming green before merge.

PR #24656, opened 2026-07-13
CAMEL-24045: camel-sql - Fix connection leak with outputType=StreamList when the statement is not a query

Croway:
JIRA: https://issues.apache.org/jira/browse/CAMEL-24045

`SqlProducer.processStreamList()` takes a raw connection from the DataSource so the result set can stay open while the exchange is routed. When `PreparedStatement.execute()` returns `false` (INSERT/UPDATE/DELETE/DDL), the result-set branch was skipped: no `ResultSetIteratorCompletion` was registered and neither the connection nor the statement was ever closed — the catch block only cleaned up on exception. Since the executed SQL can vary per exchange via the `CamelSqlQuery` header or `useMessageBodyForSql=true`, each such exchange leaked one pooled connection, eventually exhausting the pool.

Changes:
- Close the statement and return the connection to the pool when the executed statement produced no result set
- Set the `CamelSqlUpdateCount` header from `getUpdateCount()` for the non-query case, consistent with the non-query path in `processInternal()` (review follow-up)
- New test with a connection-counting DataSource proving no connection stays open after the exchange completes and asserting the update-count header (fails on current `main`, passes with the fix); existing StreamList producer/consumer tests unaffected

### DIFF
--- a/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
@@ -268,6 +268,7 @@ public class SqlProducer extends DefaultProducer {
             } else {
                 // not a query so there is no result set to stream; close the statement
                 // and return the connection to the pool right away
+                exchange.getIn().setHeader(SqlConstants.SQL_UPDATE_COUNT, ps.getUpdateCount());
                 closeStatement(ps);
                 closeConnection(con);
             }
--- a/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/component/sql/SqlProducer.java
@@ -265,6 +265,11 @@ public class SqlProducer extends DefaultProducer {
                 // we do not know the row count so we cannot set a ROW_COUNT header
                 // defer closing the iterator when the exchange is complete
                 exchange.getExchangeExtension().addOnCompletion(new ResultSetIteratorCompletion(iterator));
+            } else {
+                // not a query so there is no result set to stream; close the statement
+                // and return the connection to the pool right away
+                closeStatement(ps);
+                closeConnection(con);
             }
             return iterator;
         } catch (Exception e) {
--- a/components/camel-sql/src/test/java/org/apache/camel/component/sql/SqlProducerStreamListNonSelectConnectionLeakTest.java
+++ b/components/camel-sql/src/test/java/org/apache/camel/component/sql/SqlProducerStreamListNonSelectConnectionLeakTest.java
@@ -68,6 +68,7 @@ public class SqlProducerStreamListNonSelectConnectionLeakTest extends CamelTestS
     public void testNonSelectStreamListDoesNotLeakConnection() throws Exception {
         MockEndpoint mock = getMockEndpoint("mock:result");
         mock.expectedMessageCount(1);
+        mock.expectedHeaderReceived(SqlConstants.SQL_UPDATE_COUNT, 1);
 
         template.sendBody("direct:update", "go");
 
--- a/components/camel-sql/src/test/java/org/apache/camel/component/sql/SqlProducerStreamListNonSelectConnectionLeakTest.java
+++ b/components/camel-sql/src/test/java/org/apache/camel/component/sql/SqlProducerStreamListNonSelectConnectionLeakTest.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Clean, minimal fix for a genuine connection leak in `SqlProducer.processStreamList()` when the executed statement is not a query (INSERT/UPDATE/DELETE/DDL). The fix correctly closes the statement and connection in the previously-missing `else` branch.

**Scanner coverage:** No static analysis tools (PMD, Checkstyle, semgrep, SpotBugs) available. Rely on CI.

Notes:
- The fix correctly relies on Spring's `JdbcUtils.closeStatement()` and `JdbcUtils.closeConnection()`, both of which handle nulls and swallow exceptions internally, so sequential close calls are safe.
- The test is well-crafted: the `ConnectionCountingDataSource` using JDK Proxy tracks connection lifecycle and verifies both the functional outcome and the resource invariant (zero open connections).

One minor observation: the new `else` branch does not set the `SQL_UPDATE_COUNT` header, unlike the equivalent non-query path in `processInternal()`. For completeness, consider adding `exchange.getIn().setHeader(SqlConstants.SQL_UPDATE_COUNT, ps.getUpdateCount())` before closing the statement. This is low severity since StreamList mode with non-queries is an edge case.

Croway:
Thanks @gnodet I've added the header

davsclaus (approved):

oscerd (commented):
Reviewed with focus on resource lifecycle across all execution paths — the fix is correct and I have no change requests. Summary of the verification:

**The bug is real and the fix closes exactly the leaked path.** On current main, `processStreamList` (`SqlProducer.java:260-269`) only handles `ps.execute() == true`: the ResultSet path hands `con`/`ps`/`rs` ownership to `ResultSetIterator` and defers cleanup to `ResultSetIteratorCompletion` at exchange completion. When the statement is *not* a query (`execute()` returns `false`), the method returns a `null` iterator and `con`/`ps` are never closed — the catch block (`SqlProducer.java:272-274`) only covers the exception path. The new `else { closeStatement(ps); closeConnection(con); }` branch closes them exactly once, cannot interfere with the streaming path (it only runs when there is no ResultSet to stream), and cannot double-close on exception (Spring's `JdbcUtils.close*` never throw, so the branch can't re-enter the catch).

**No double-close on the streaming side either**: `ResultSetIterator.close()` is idempotent via `AtomicBoolean.compareAndSet`, which covers the self-close-on-exhaustion and empty-result-set cases.

**Test quality is a highlight**: the regression test asserts the actual invariant (`openConnections() == 0` via a counting `DelegatingDataSource` proxy) rather than merely exercising the path, is fully synchronous (no `Thread.sleep`, native `assertIsSatisfied`), and fails on unfixed main by inspection.

**History check**: the missing else traces back to the original CAMEL-9849 StreamList implementation — there is no prior intentional decision to keep connections open on non-query statements, so this does not revert deliberate behavior. Behavior parity is preserved (body/headers untouched on the non-query path, as before; the absent `CamelSqlUpdateCount` on this path is pre-existing and would be a separate enhancement).

Two merge-time notes:
1. CI: the two build jobs (JDK 17/25) were still pending at review time — worth confirming green before merge.
2. The PR body declares "Generated with Claude Code" but the commit has no AI co-authorship trailer — per the repo's AI guidelines the squash-merge commit message should include the `Co-authored-by` trailer for the AI assistant.

gnodet (approved):

**Re-review focused on the new commit** (`2864ca45`): _CAMEL-24045: camel-sql - Set CamelSqlUpdateCount header for non-query StreamList statements_

This commit adds exactly two lines and directly implements the suggestion from the prior review:

1. **`SqlProducer.java`**: Sets `CamelSqlUpdateCount` via `ps.getUpdateCount()` in the non-query `else` branch of `processStreamList()`, before `closeStatement(ps)` — the ordering is correct since the update count is not accessible after the statement is closed.

2. **Test**: Adds `mock.expectedHeaderReceived(SqlConstants.SQL_UPDATE_COUNT, 1)` asserting the header value matches the expected single-row update.

The pattern is consistent with the equivalent non-query path in `processInternal()` (line 218). No issues found.

**Scanner coverage:** No standalone static analysis tools (PMD, Checkstyle, semgrep, SpotBugs) available. Relying on CI for validation.

gnodet (approved):

**Re-review focused on the new commit** (`2864ca45`): _CAMEL-24045: camel-sql - Set CamelSqlUpdateCount header for non-query StreamList statements_

This commit adds exactly two lines and directly implements the suggestion from the prior review:

1. **`SqlProducer.java`**: Sets `CamelSqlUpdateCount` via `ps.getUpdateCount()` in the non-query `else` branch of `processStreamList()`, before `closeStatement(ps)` — the ordering is correct since the update count is not accessible after the statement is closed.

2. **Test**: Adds `mock.expectedHeaderReceived(SqlConstants.SQL_UPDATE_COUNT, 1)` asserting the header value matches the expected single-row update.

The pattern is consistent with the equivalent non-query path in `processInternal()` (line 218). No issues found.

**Scanner coverage:** No standalone static analysis tools (PMD, Checkstyle, semgrep, SpotBugs) available. Relying on CI for validation.

PR #24657, opened 2026-07-13
CAMEL-24046: camel-sql - JdbcCachedMessageIdRepository must not cache a key whose insert failed, make cache thread-safe

Croway:
JIRA: https://issues.apache.org/jira/browse/CAMEL-24046

Two related defects in `JdbcCachedMessageIdRepository`:

1. `add(key)` put the key into the in-memory cache *before* calling `super.add(key)`. If the database insert failed (outage, PK race with another node), the exception propagated and the idempotent consumer never processed the message — but the key stayed cached, so every redelivery was rejected as a duplicate. The message was never processed and never stored until JVM restart.
2. The cache was a plain `HashMap` mutated from concurrent consumer threads with no synchronization and a non-volatile field (a runtime `reload()` might never become visible to other threads); concurrent puts during resize can lose entries, causing duplicate processing. The hit/miss counters were unsynchronized ints.

Changes:
- Populate the cache only after the database insert succeeded
- Use `ConcurrentHashMap` with a volatile field for safe publication on `reload()`, and `AtomicInteger` hit/miss counters (public getters unchanged)
- New test simulating a DB failure during `add()`: `contains()` must return false and a redelivered `add()` must succeed once the DB is back (fails on current `main`, passes with the fix); existing cache hit/miss-count test passes unchanged, confirming counting semantics are preserved

### DIFF
--- a/components/camel-sql/src/main/java/org/apache/camel/processor/idempotent/jdbc/JdbcCachedMessageIdRepository.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/processor/idempotent/jdbc/JdbcCachedMessageIdRepository.java
@@ -16,8 +16,9 @@
  */
 package org.apache.camel.processor.idempotent.jdbc;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.sql.DataSource;
 
@@ -29,9 +30,9 @@ import org.springframework.transaction.support.TransactionTemplate;
  * Caching version of {@link JdbcMessageIdRepository}
  */
 public class JdbcCachedMessageIdRepository extends JdbcMessageIdRepository {
-    private Map<String, Integer> cache = new HashMap<>();
-    private int hitCount;
-    private int missCount;
+    private volatile Map<String, Integer> cache = new ConcurrentHashMap<>();
+    private final AtomicInteger hitCount = new AtomicInteger();
+    private final AtomicInteger missCount = new AtomicInteger();
     private String queryAllString
             = "SELECT messageId, COUNT(*) FROM CAMEL_MESSAGEPROCESSED WHERE processorName = ? GROUP BY messageId";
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Both fixes are correct and well-tested. The cache-before-insert bug fix prevents permanent message loss on DB failure — moving `cache.merge()` to after `super.add()` ensures keys are only cached after a successful DB insert. The `ConcurrentHashMap`/`AtomicInteger` migration resolves thread-safety hazards in the shared cache.

**Scanner coverage:** No static analysis tools (PMD, Checkstyle, semgrep, SpotBugs) available. Rely on CI.

Notes:
- The `cache.merge(key, 1, Integer::sum)` after `super.add(key)` executes unconditionally regardless of the return value. When `super.add()` returns false (duplicate), the cache count is inflated — functionally harmless for boolean idempotent semantics and matches pre-existing behavior.
- `volatile` on the `cache` field correctly ensures safe publication when `reload()` replaces the map from a different thread.
- The test directly validates the regression scenario (drop table → simulate failure → recreate → verify key is not cached).

davsclaus (approved):

oscerd (commented):
Reviewed with focus on the cache/insert ordering and the concurrency changes — the fix is correct and I have no change requests. Summary of the verification:

**Both halves of the fix address confirmed defects on main.** The poisoned-cache bug is real: `add()` currently caches the key (`cache.put`, line 70) *before* `super.add(key)` (line 76), so an insert that throws leaves the key cached and every redelivery is rejected as a duplicate forever. The reordering (only `cache.merge` after `super.add` returns) fixes all paths, including the `super.add() == false` case (row already in DB), where caching still matches DB truth. The thread-safety half is equally justified: main uses a plain `HashMap` with non-volatile field and plain `int` counters, shared across concurrently invoked idempotent-consumer threads.

**The regression test is genuine**: `assertFalse(repository.contains("second"))` after a failed insert fails on current main (cache hit → true) and passes with the fix. Fully synchronous, no `Thread.sleep`.

**History check**: the cache-before-insert ordering dates to the original CAMEL-16770 contribution — no prior intentional decision is being reverted.

Two non-blocking observations, both pre-existing rather than introduced here:

1. Inside an outer transaction (`PROPAGATION_REQUIRED`), `super.add()` returning doesn't mean the insert is *committed* — if the enclosing transaction rolls back later, the DB row vanishes while the key stays cached. In practice the idempotent consumer's default `removeOnFailure=true` evicts the key on failure, and the new code is strictly better than the old (which cached before even attempting the insert), so no action needed — at most the comment could say "after the insert call returned" rather than "succeeded".
2. The hit-path `getOrDefault`/`merge` pair isn't atomic, so a concurrent `remove()` between them can resurrect a cache entry. Same window existed before the PR; cache values are only presence flags and the DB remains source of truth on a miss, so it's benign.

No file overlap with the sibling camel-sql PRs #24655/#24656 (disjoint packages), so merge order between the three doesn't matter. The two build jobs were still pending at review time — worth confirming green before merge.

PR #24661, opened 2026-07-13
CAMEL-23877: Fix camel-pinecone tls option documented default and add insecure:ssl annotation

davsclaus:
## Summary

- Fix the `tls` option's `@UriParam(defaultValue)` from `"true"` to `"false"` to match the actual Java field default
- Add `security = "insecure:ssl"` and `insecureValue = "false"` annotations to flag TLS-disabled as an insecure configuration
- Simplify TLS handling in `PineconeVectorDbEndpoint.createClient()` to always pass the configured value to the Pinecone SDK builder
- Add upgrade guide entry documenting the corrected default

No behavioral change — the runtime default was already `false`.

## Test plan

- [x] `mvn test` passes in `components/camel-ai/camel-pinecone` (2/2 tests)
- [x] Generated catalog (`pinecone.json`) shows `"defaultValue": false` with `"security": "insecure:ssl"`
- [x] Catalog, endpoint-dsl, and component-dsl regenerated


gnodet (approved):

Clean, well-scoped fix. I verified the following:

1. **Default value mismatch confirmed**: The Java field `private boolean tls;` (a primitive boolean) defaults to `false`, but `@UriParam(defaultValue = "true")` documented it as `true`. The PR correctly aligns the annotation with the runtime default. The existing test at `PineconeComponentConfigurationTest.java:65` already asserts `assertEquals(false, endpoint.getConfiguration().isTls())`, which confirms the runtime default was always `false`.

2. **Security annotation correct**: `security = "insecure:ssl"` with `insecureValue = "false"` is the correct pattern for a TLS-enable option where disabling it (setting to `false`) is insecure. This matches the convention used by `hostnameverification`, `validatecertificates`, `verifyssl`, etc. in `SecurityUtils`.

3. **Code simplification sound**: Changing from conditional `if (!tls) { builder.withTlsEnabled(tls); }` to unconditional `builder.withTlsEnabled(configuration.isTls())` is cleaner and explicitly communicates the TLS setting to the Pinecone SDK in both directions.

4. **SecurityUtils entry**: The `"tls"` entry is correctly placed alphabetically and uses `VALUE_FALSE`. Note that this will also apply to other components with a `tls` option (Qdrant, MongoDB) which have the same semantics -- this is a net positive for security coverage.

5. **Upgrade guide**: Properly documents this as a documentation-only correction with no behavioral change.

No issues found. LGTM.

**Scanner coverage:** No static analysis tools ran (Maven build not executed in this review). No SonarCloud, SpotBugs, or Error Prone output available. This review is based on manual code inspection of the diff, git history, and cross-referencing with project conventions.

PR #24662, opened 2026-07-13
CAMEL-24052: Fix model dumpers permanently destroying EndpointDSL builders on RouteDefinition

Croway:
## Summary

`resolveEndpointDslUris()` in all four model dumpers (`LwModelToXMLDumper`, `JaxbModelToXMLDumper` via `JaxbHelper`, `LwModelToYAMLDumper`, `LwModelToJavaDumper`) called `setUri()` on live `FromDefinition` and `SendDefinition` objects before serialization. `setUri()` internally calls `clear()`, which permanently nulls `endpointConsumerBuilder`/`endpointProducerBuilder`. For endpoints with non-primitive builder properties (e.g. a custom `S3Client`), this causes Camel to lose the pre-configured bean reference and fall back to a default unconfigured client, resulting in route startup failures.

The most common trigger is `AdviceWith.adviceWith()` with the default `logXml=true`, but any JMX `dumpRoutesAsXml()` call or startup dump on a context using Endpoint DSL hits the same bug.

## Changes

- `resolveEndpointDslUris()` now saves builder references before calling `setUri()` and returns a `Runnable` that restores them
- All four dumpers call the restorer in a `finally` block after serialization completes
- Added `AdviceWithEndpointDslMutationTest` in `camel-endpointdsl` covering the XML dumper directly and through `adviceWith()` for both `FromDefinition` and `SendDefinition` builders

## Related

- https://issues.apache.org/jira/browse/CAMEL-24052
- Reproducer: https://github.com/bridgerdier/camel-advicewith-endpointdsl/

### DIFF
--- a/core/camel-xml-jaxb/src/main/java/org/apache/camel/xml/jaxb/JaxbHelper.java
+++ b/core/camel-xml-jaxb/src/main/java/org/apache/camel/xml/jaxb/JaxbHelper.java
@@ -41,6 +41,8 @@ import org.apache.camel.CamelContext;
 import org.apache.camel.Expression;
 import org.apache.camel.NamedNode;
 import org.apache.camel.TypeConversionException;
+import org.apache.camel.builder.EndpointConsumerBuilder;
+import org.apache.camel.builder.EndpointProducerBuilder;
 import org.apache.camel.converter.jaxp.XmlConverter;
 import org.apache.camel.model.BasicExpressionNode;
 import org.apache.camel.model.ExpressionNode;
@@ -132,28 +134,35 @@ public final class JaxbHelper {
 
     /**
      * If the route has been built with endpoint-dsl, then the model will not have uri set which then cannot be included
-     * in the JAXB model dump
+     * in the JAXB model dump. Returns a {@link Runnable} that restores the original endpoint DSL builders after
+     * serialization, so that the live {@link RouteDefinition} is not permanently mutated.
      */
-    public static void resolveEndpointDslUris(RouteDefinition route) {
+    @SuppressWarnings("rawtypes")
+    public static Runnable resolveEndpointDslUris(RouteDefinition route) {
+        List<Runnable> restorers = new ArrayList<>();
         FromDefinition from = route.getInput();
         if (from != null && from.getEndpointConsumerBuilder() != null) {
-            String uri = from.getEndpointConsumerBuilder().getRawUri();
-            from.setUri(uri);
+            EndpointConsumerBuilder builder = from.getEndpointConsumerBuilder();
+            from.setUri(builder.getRawUri());
+            restorers.add(() -> from.setEndpointConsumerBuilder(builder));
         }
         Collection<SendDefinition> col = filterTypeInOutputs(route.getOutputs(), SendDefinition.class);
         for (SendDefinition<?> to : col) {
             if (to.getEndpointProducerBuilder() != null) {
-                String uri = to.getEndpointProducerBuilder().getRawUri();
-                to.setUri(uri);
+                EndpointProducerBuilder builder = to.getEndpointProducerBuilder();
+                to.setUri(builder.getRawUri());
+                restorers.add(() -> to.setEndpointProducerBuilder(builder));
             }
         }
         Collection<ToDynamicDefinition> col2 = filterTypeInOutputs(route.getOutputs(), ToDynamicDefinition.class);
         for (ToDynamicDefinition to : col2) {
             if (to.getEndpointProducerBuilder() != null) {
-                String uri = to.getEndpointProducerBuilder().getRawUri();
-                to.setUri(uri);
+                EndpointProducerBuilder builder = to.getEndpointProducerBuilder();
+                to.setUri(builder.getRawUri());
+                restorers.add(() -> to.setUri(null));
             }
         }
+        return () -> restorers.forEach(Runnable::run);
     }
 
     private static NamespaceAware getNamespaceAwareFromExpression(ExpressionNode expressionNode) {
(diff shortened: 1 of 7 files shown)


core/camel-xml-jaxb/src/main/java/org/apache/camel/xml/jaxb/JaxbHelper.java
```diff
@@ -44,6 +44,8 @@
 import org.apache.camel.converter.jaxp.XmlConverter;
 import org.apache.camel.model.BasicExpressionNode;
 import org.apache.camel.model.ExpressionNode;
+import org.apache.camel.builder.EndpointConsumerBuilder;
```
gnodet: Minor: these two imports should be placed before the `org.apache.camel.model.*` group to maintain alphabetical order within the `org.apache.camel` package. Running `mvn impsort:sort` in the `camel-xml-jaxb` module should fix this automatically.

gnodet (approved):

**Summary:** This PR fixes a real and impactful bug where model dumpers (XML, YAML, Java DSL) permanently destroyed `EndpointConsumerBuilder` / `EndpointProducerBuilder` references on live `RouteDefinition` objects. The most common trigger was `AdviceWith.adviceWith()` with the default `logXml=true`, but any JMX `dumpRoutesAsXml()` call would hit the same bug. Good work identifying the root cause and tracing it through all four dumper implementations.

**Fix analysis:**

I verified the `clear()` / `setUri()` behavior in the model classes:

- `FromDefinition.setUri()` calls `clear()`, which nulls `endpointConsumerBuilder`, `endpoint`, and `uri` -- confirmed destructive.
- `SendDefinition.setUri()` calls `clear()`, which nulls `endpointProducerBuilder`, `endpoint`, `uri`, and `endpointUriToString` -- confirmed destructive.
- `ToDynamicDefinition.setUri()` does NOT call `clear()`, just sets `uri` -- the builder is preserved. The PR correctly uses `restorers.add(() -> to.setUri(null))` instead of restoring the builder, which is the right approach here.

The restore logic is correct: `setEndpointConsumerBuilder(builder)` and `setEndpointProducerBuilder(builder)` both call `clear()` first (nulling the temporarily-set URI), then set the builder back to the original reference. The net effect is a clean restoration to the pre-dump state.

The `finally` blocks ensure restoration happens even if serialization throws an exception.

1. The import order in `JaxbHelper.java` places `org.apache.camel.builder.*` after `org.apache.camel.model.ExpressionNode`, which breaks alphabetical ordering. Please run `mvn impsort:sort` on the `camel-xml-jaxb` module if the build's sourcecheck flags it.

2. `JaxbHelper.resolveEndpointDslUris` is a `public static` method whose return type changed from `void` to `Runnable`. The internal callers in `modelToXml()` (lines 355, 363, 371, 378) now discard the returned `Runnable`. This is harmless -- those call sites process freshly deserialized routes that have no EndpointDSL builders set -- but worth noting for future awareness.

3. The `resolveEndpointDslUris` implementation is duplicated across all four dumper classes. This is a pre-existing code smell, not introduced by this PR.

4. There is still a narrow thread-safety window where the live `RouteDefinition` is temporarily mutated between URI resolution and serialization. This is a pre-existing concern that this PR cannot fully address without a larger refactoring (e.g., working on copies of the definitions). The PR does improve the situation by restoring state afterward.

**Tests:** Three well-structured test methods covering: (1) direct XML dumper invocation with `FromDefinition`, (2) `adviceWith()` integration with `FromDefinition`, and (3) `adviceWith()` integration with `SendDefinition` (producer builder). The route uses a non-primitive `ExceptionHandler` property to reproduce the `hash=XXXXXXXX` URI scenario. Good test design.

**Scanner coverage:** No static analysis tools available in this environment. Manual code review performed against the model source code (`FromDefinition`, `SendDefinition`, `ToDynamicDefinition`) to verify correctness of save/restore semantics. CI checks (build for JDK 17 and JDK 25) are still pending at review time.

This review does not replace specialized review tools such as CodeRabbit, Sourcery, or static analyzers such as SonarCloud.

core/camel-xml-jaxb/src/main/java/org/apache/camel/xml/jaxb/JaxbHelper.java
```diff
@@ -44,6 +44,8 @@
...
+import org.apache.camel.builder.EndpointProducerBuilder;
```
davsclaus: The `org.apache.camel.builder.*` imports are placed after `org.apache.camel.model.ExpressionNode`, but alphabetically they belong between `org.apache.camel.TypeConversionException` and `org.apache.camel.converter.jaxp.XmlConverter`. This will fail `mvn impsort:sort` in CI.

The other three dumper files have the imports correctly placed — only this file needs fixing.

Running `mvn impsort:sort` in the `core/camel-xml-jaxb` module should auto-fix it.

davsclaus (commented):
## Review: CAMEL-24052 — Fix model dumpers permanently destroying EndpointDSL builders

Nice fix — the analysis of the `setUri()` → `clear()` destruction chain is thorough, and the save/restore pattern correctly handles the different semantics across `FromDefinition`, `SendDefinition`, and `ToDynamicDefinition`.

One import-ordering issue to fix before merge (details in inline comment). Otherwise this looks good.

**Positive observations:**
- Restorer logic correctly distinguishes `FromDefinition`/`SendDefinition` (where `setUri()` calls `clear()`, destroying the builder) from `ToDynamicDefinition` (where `setUri()` does not call `clear()`, so only the URI needs resetting)
- `finally` blocks ensure restoration even if serialization throws
- `WireTapDefinition extends ToDynamicDefinition` is properly covered by `filterTypeInOutputs`
- Tests cover the three key scenarios: XML dump directly, `adviceWith` for consumer builder, `adviceWith` for producer builder
- Non-primitive `ExceptionHandler` property reproduces the `hash=XXXXXXXX` URI behavior well

PR #24663, opened 2026-07-13
CAMEL-24003: Add heap dump command to CLI, TUI, and MCP

davsclaus:
## Summary

- Add `camel cmd heap-dump` CLI command to write `.hprof` heap dump files from a running Camel integration for deep memory analysis with tools like Eclipse MAT or VisualVM
- Add `HeapDumpDevConsole` in camel-console as the server-side implementation using `HotSpotDiagnosticMXBean`
- Wire the heap-dump action in `LocalCliConnector` to bridge CLI/TUI requests to the dev console
- Add heap dump support to TUI Memory and MemoryLeak tabs via `h` key shortcut
- Register `write_heap_dump` AI tool and `camel_runtime_heap_dump` MCP tool
- Add documentation page for the new command

## Test plan

- [ ] Run `camel cmd heap-dump` against a running integration and verify `.hprof` file is created
- [ ] Run `camel cmd heap-dump --dump-name=test --live=false` and verify options work
- [ ] Open TUI Memory tab, press `h`, verify heap dump is written and notification appears
- [ ] Open TUI MemoryLeak tab, press `h`, verify heap dump is written and notification appears
- [ ] Verify MCP tool `camel_runtime_heap_dump` works via MCP client


core/camel-console/src/main/java/org/apache/camel/impl/console/HeapDumpDevConsole.java
```diff
@@ -0,0 +1,88 @@
...
+package org.apache.camel.impl.console;
+
+import java.io.File;
+import java.lang.management.ManagementFactory;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Map;
+
+import com.sun.management.HotSpotDiagnosticMXBean;
+import org.apache.camel.spi.Configurer;
```
gnodet, line 29: **[Medium]** Direct compile-time dependency on `com.sun.management.HotSpotDiagnosticMXBean`.

The existing `HeapHistogramDevConsole` avoids this by using string-based MBean lookup (`new ObjectName("com.sun.management:type=DiagnosticCommand")`), which fails gracefully at runtime on non-HotSpot JVMs. This direct import creates a hard compile-time dependency that could cause `NoClassDefFoundError` on alternative JVMs or in GraalVM native image builds.

Consider using `ManagementFactory.getPlatformMBeanServer()` with reflection or string-based invocation to match the pattern used by `HeapHistogramDevConsole`.

core/camel-console/src/main/java/org/apache/camel/impl/console/HeapDumpDevConsole.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        return "Heap dump written to: " + json.getString("file") + " (" + json.getLong("size") + " bytes)";
+    }
+
+    @Override
+    protected JsonObject doCallJson(Map<String, Object> options) {
+        JsonObject root = new JsonObject();
+
+        String name = optionString(options, NAME);
+        if (name == null || name.isBlank()) {
```
gnodet, line 69: **[Low]** The user-supplied `name` is used as-is in the path passed to `dumpHeap()`. A value containing path separators (e.g., `../../tmp/dump`) would write the heap dump to an arbitrary location. While Camel's security model treats CLI users as trusted, a defensive `Path.of(name).getFileName().toString()` would prevent accidental writes outside the working directory.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/MemoryTab.java
```diff
@@ -437,6 +441,47 @@ private static Span computeTrendSpan(LinkedList<Long> hist, long heapCeiling) {
...
+        }
+        notify("Writing heap dump...", false);
+        String pid = info.pid;
+        Thread t = new Thread(() -> {
+            Path outputFile = ctx.getOutputFile(pid);
+            PathUtils.deleteFile(outputFile);
+            JsonObject root = new JsonObject();
+            root.put("action", "heap-dump");
+            Path actionFile = ctx.getActionFile(pid);
+            PathUtils.writeTextSafely(root.toJson(), actionFile);
```
gnodet: **[Low]** This method is nearly identical to `MemoryLeakTab.triggerHeapDump()`. The only difference is that `MemoryLeakTab` uses its `startDaemonThread()` helper while this creates the thread inline. Consider extracting the shared logic to avoid the two copies drifting apart over time.

gnodet (commented):
## Review: CAMEL-24003 - Add heap dump command to CLI, TUI, and MCP

Nice feature addition. The implementation is well-structured and follows the established patterns for dev consoles, CLI commands, TUI integration, and MCP/AI tool registration. The layered approach (dev console -> CLI connector -> jbang command / TUI / MCP) is consistent with how heap-histogram, memory-leak, and other diagnostic features work.

A few items worth discussing:

### Findings

**1. [Medium] Direct compile-time dependency on `com.sun.management.HotSpotDiagnosticMXBean`** (`HeapDumpDevConsole.java`)

The existing `HeapHistogramDevConsole` avoids a compile-time dependency on `com.sun.management` by using string-based MBean lookup (`new ObjectName("com.sun.management:type=DiagnosticCommand")`). The new `HeapDumpDevConsole` directly imports `com.sun.management.HotSpotDiagnosticMXBean`, which creates a hard compile-time dependency on HotSpot-internal API. This could cause `NoClassDefFoundError` on alternative JVMs (e.g., OpenJ9, GraalVM native image). Consider using the MXBean via `ManagementFactory.getPlatformMBeanServer()` and string-based invocation instead, or at minimum wrap the call so it degrades gracefully.

**2. [Low] No input sanitization on dump file name** (`HeapDumpDevConsole.java`)

The user-supplied `name` parameter is passed directly to `bean.dumpHeap(name, live)` after only appending `.hprof`. A value like `../../tmp/evil` would write to an arbitrary path. Per Camel's security model, CLI/TUI users are trusted (they already have JVM-level access), so this is low severity. Still, stripping path separators or using `Path.getFileName()` would be a reasonable defensive measure.

**3. [Low] Code duplication of `triggerHeapDump()` and `notify()` between `MemoryTab` and `MemoryLeakTab`**

Both tabs contain nearly identical ~40-line `triggerHeapDump()` and `notify()` methods. The `MemoryLeakTab` version uses its existing `startDaemonThread()` helper, while `MemoryTab` creates the daemon thread inline. Consider extracting the shared logic to a utility (or a default method on a shared base) to avoid drift between the two copies.

**4. [Low] No test coverage** (project-wide)

No tests were added. While the existing `HeapHistogramDevConsole` also has no test, the project's `CLAUDE.md` states: "Every PR must include tests for new functionality or bug fixes." A basic unit test for the dev console (e.g., verifying JSON output shape, error propagation, default filename generation) would be straightforward and valuable.

**5. [Info] MCP tool uses `runtimeService.executeAction()` directly instead of `delegateToRegistry()`**

The adjacent `camel_runtime_heap_histogram` delegates via `delegateToRegistry("get_heap_histogram", ...)` while the new `camel_runtime_heap_dump` calls `runtimeService.executeAction()` directly. This is fine since heap dump needs custom parameter forwarding, but it is a pattern difference worth noting for consistency.

### Summary

The feature is well-designed and the code is clean. The main item to consider is the direct `com.sun.management` import which introduces a HotSpot-specific compile dependency, unlike the existing heap histogram console which avoids it. The other findings are minor suggestions.

### Scanner coverage

No static analysis tools (SonarCloud, SpotBugs, Error Prone) were executed as part of this review. This review does not replace those tools or specialized AI review tools like CodeRabbit.

core/camel-console/src/main/java/org/apache/camel/impl/console/HeapDumpDevConsole.java
davsclaus: Fixed — replaced the direct `com.sun.management.HotSpotDiagnosticMXBean` import with string-based `MBeanServer.invoke()` via `ObjectName("com.sun.management:type=HotSpotDiagnostic")`, matching the pattern used by `HeapHistogramDevConsole`.

core/camel-console/src/main/java/org/apache/camel/impl/console/HeapDumpDevConsole.java
davsclaus: Fixed — added `name = Path.of(name).getFileName().toString()` to strip path separators before the name is passed to `dumpHeap()`.

dsl/camel-jbang/camel-jbang-plugin-tui/src/main/java/org/apache/camel/dsl/jbang/core/commands/tui/MemoryTab.java
davsclaus: Fixed — extracted the shared logic into `TuiHelper.triggerHeapDump(MonitorContext, BiConsumer<String, Boolean>)` and both `MemoryTab` and `MemoryLeakTab` now delegate to it.

davsclaus:
Thanks for the review @gnodet! All three inline findings have been addressed in 17b46ab:

1. **MBeanServer** — switched to string-based `MBeanServer.invoke()` via `ObjectName("com.sun.management:type=HotSpotDiagnostic")`, matching the `HeapHistogramDevConsole` pattern.
2. **Path sanitization** — added `Path.of(name).getFileName().toString()` to strip path separators.
3. **Code duplication** — extracted shared `TuiHelper.triggerHeapDump()` used by both `MemoryTab` and `MemoryLeakTab`.

On the remaining items:

4. **Tests** — agreed, the dev console modules in `camel-console` currently have no test infrastructure set up. The heap dump action requires a running JVM with the HotSpot MBean, which makes unit testing non-trivial (the `dumpHeap` call creates a real file). We can add basic tests for the filename sanitization and default-name logic in a follow-up.

5. **MCP pattern** — correct, the heap dump tool uses `executeAction()` directly because it needs to pass the `name` and `live` parameters through the action file protocol, which doesn't fit the `delegateToRegistry()` pattern used for simple get-style calls like heap histogram.

PR #24668, opened 2026-07-13
CAMEL-23970: Mark sslEndpointAlgorithm as insecure SSL option

gbhavya07:
# Description

This PR addresses CAMEL-23970 by marking the `sslEndpointAlgorithm` endpoint option as an insecure-capable SSL configuration.

The `sslEndpointAlgorithm` option allows disabling TLS hostname verification by setting it to an empty value or `none`. However, it was not annotated with `security = "insecure:ssl"`, preventing Camel's security policy framework from identifying it as an option that weakens SSL verification.

This change adds the missing `security = "insecure:ssl"` attribute to the `@UriParam` annotation in `OpenAIConfiguration` and regenerates the component metadata so the security metadata is reflected in the generated `openai.json`.

# Target

# Tracking

- [x] I checked there is a JIRA issue filed for the change: CAMEL-23970

I ran ./mvnw clean install -DskipTests. The build fails in the unrelated camel-kserve module due to a protobuf-maven-plugin/protoc generation error. The camel-openai module builds successfully, and its tests pass.

# AI-assisted contributions


components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIConfiguration.java
```diff
@@ -240,7 +240,7 @@ public class OpenAIConfiguration implements Cloneable {
     @Metadata(description = "The algorithm used by the trust manager factory for SSL connections")
     private String sslTrustmanagerAlgorithm = "PKIX";
 
-    @UriParam(label = "security", defaultValue = "https")
+    @UriParam(label = "security", defaultValue = "https", security = "insecure:ssl")
```
gnodet, line 243: The annotation correctly adds `security = "insecure:ssl"`. However, consider also adding `insecureValue = ""` explicitly (or even `insecureValue = "none"`) to document the insecure value.

The current behavior without `insecureValue`: the build tooling defaults to `""` for String-type options (see `collectSecurityOption` in `UpdateSensitizeHelper.java`). This catches `sslEndpointAlgorithm=""` but misses `sslEndpointAlgorithm=none`, even though both disable hostname verification (see `OpenAIEndpoint.java` line 539).

Note: this same gap exists for `strictHostKeyChecking` in `BaseSftpConfiguration`, so this may be acceptable for now.

components/camel-ai/camel-openai/src/generated/resources/META-INF/org/apache/camel/component/openai/openai.json
```diff
@@ -113,7 +113,7 @@
...
+    "sslEndpointAlgorithm": { "index": 37, "kind": "parameter", "displayName": "Ssl Endpoint Algorithm", "group": "security", "label": "security", "required": false, "type": "string", "javaType": "java.lang.String", "deprecated": false, "deprecationNote": "", "autowired": false, "secret": false, "security": "insecure:ssl", "defaultValue": "https", "configurationClass": "org.apache.camel.component.openai.OpenAIConfiguration", "configurationField": "configuration", "description": "The endpoint identification algorithm to validate the server hostname using the server certificate. Set to an empty string or 'none' to disable hostname verification" },
```
gnodet, line 116: This component-level metadata is correctly updated. However, the **catalog copy** at `catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/openai.json` also needs the same update. It will be auto-generated when the catalog module builds successfully.

Additionally, `core/camel-util/src/main/java/org/apache/camel/util/SecurityUtils.java` needs a new entry in the `SECURITY_OPTIONS` map (also auto-generated during catalog build). Without it, the runtime security policy framework cannot enforce policies on this option.

gnodet (changes requested):
Thank you for working on this, @gbhavya07. The annotation change itself is correct — `sslEndpointAlgorithm` is indeed an insecure-capable SSL option that should be flagged by the security policy framework. However, the PR is incomplete because the full build did not finish, so required generated files are missing.

## Findings

### 1. Missing `SecurityUtils.java` update (blocking)

The auto-generated `SECURITY_OPTIONS` map in `core/camel-util/src/main/java/org/apache/camel/util/SecurityUtils.java` must include an entry for `sslendpointalgorithm`. This map is regenerated by the `update-sensitive-helper` mojo during the `camel-catalog` build phase. Since your build failed at `camel-kserve` (which is built before `catalog`), this file was never regenerated.

For reference, see how PR #24661 (camel-pinecone) includes the corresponding `SecurityUtils.java` change:
```java
map.put("tls", new SecurityOption(INSECURE_SSL, VALUE_FALSE));
```

Without this entry, the security policy framework **cannot enforce policies** on `sslEndpointAlgorithm` at runtime, which defeats the purpose of the annotation.

### 2. Missing catalog JSON update (blocking)

The catalog copy at `catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/openai.json` also needs to be regenerated to include the `"security": "insecure:ssl"` attribute. This is generated during the same catalog build phase.

### 3. Consider adding explicit `insecureValue` (suggestion)

The `sslEndpointAlgorithm` disables hostname verification for **both** empty string and `"none"` (see `OpenAIEndpoint.java` line 539). Without an explicit `insecureValue` on the annotation, the auto-generated SecurityUtils entry will default to `""` (empty string), which only catches the empty-string case. Setting `sslEndpointAlgorithm=none` would **not** be detected as insecure.

This is a pre-existing limitation also present in `strictHostKeyChecking`, so it may be acceptable for now, but worth noting.

### How to fix

To regenerate all required files, you need a successful full build. You can work around the `camel-kserve` failure by building with:
```bash
mvn clean install -DskipTests -pl :camel-openai -am && mvn clean install -DskipTests -pl :camel-catalog -am
```
Or exclude the failing module:
```bash
mvn clean install -DskipTests -pl '!:camel-kserve'
```
Then commit the regenerated `SecurityUtils.java` and `openai.json` (catalog copy).

## Scanner coverage

No static analysis tools were run in this review (PMD, Checkstyle, SpotBugs, semgrep, Error Prone are not available in this environment). Compilation-dependent checks require CI.

---

components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIConfiguration.java
gbhavya07: Thanks for the review. I've regenerated the missing catalog artifacts after completing the full catalog build. The PR now includes the generated SecurityUtils.java entry and the catalog openai.json update. I left insecureValue unchanged since this appears to be a broader existing behavior (e.g. strictHostKeyChecking) rather than something specific to this component.

components/camel-ai/camel-openai/src/generated/resources/META-INF/org/apache/camel/component/openai/openai.json
gbhavya07: Fixed. I reran the catalog generation and committed the regenerated catalog metadata, including catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/openai.json and the corresponding SecurityUtils.java update.

davsclaus (approved):
The annotation change is correct and all generated files are now included — the prior CHANGES_REQUESTED concerns are fully addressed.

**What looks good:**
- `@UriParam(security = "insecure:ssl")` on `sslEndpointAlgorithm` is the right annotation — setting this to `""` or `"none"` disables hostname verification.
- `SecurityUtils.java` entry is consistent with similar options (`stricthostkeychecking`, `x509hostnameverifier`).
- Both catalog JSON copies include the `"security": "insecure:ssl"` attribute.
- Commit message follows the `CAMEL-XXXX: description` convention.

- `camel-kafka` has the same `sslEndpointAlgorithm` option without the `security = "insecure:ssl"` annotation — a separate issue, not this PR's scope.
- The `"none"` insecure value gap is a pre-existing framework limitation already noted in the prior review.

PR #24669, opened 2026-07-13
CAMEL-24003: Pass F-keys through AI panel to global TUI handlers

davsclaus:
## Summary

- F-keys (F1-F7, F9-F12) now pass through the AI panel (F8) to their normal TUI handlers instead of being consumed by the AI input field
- Popups (F1 help, F2 actions, F3 filters) are prioritized over the AI panel in the key dispatch chain, so cursor keys and Esc work correctly in popups while the AI panel is open
- F8 still closes the AI panel as before

## Test plan

- [ ] Open TUI, press F8 to open AI panel
- [ ] Press F1 — help overlay should open and be navigable with cursor keys, Esc to close
- [ ] Press F2 — actions popup should open and be navigable with cursor keys, Esc to close
- [ ] Press F3 — filter popup should work correctly
- [ ] Press F8 — AI panel should close
- [ ] Verify typing still works normally in the AI input field


gnodet (approved):
Looks good -- clean and correct fix for F-key passthrough in the AI panel.

The two commits map cleanly to two logical changes:

1. **AiPanel.java** -- F-keys (F1-F7, F9-F12) are now explicitly rejected by `handleKeyEvent()` via a new `isFunctionKey()` guard, returning `false` to let the caller dispatch them elsewhere. F8 is correctly excluded since it closes the panel.

2. **CamelMonitor.java** -- The dispatch chain in `handleEvent()` is reordered so that `actionsPopup` and `popupManager` are checked **before** the AI panel. The AI panel dispatch changes from unconditional `return aiPanel.handleKeyEvent(ke)` to conditional `if (aiPanel.handleKeyEvent(ke)) { return true; }`, allowing unhandled keys to fall through to `handleGlobalKeys` and `handleTabKeys`.

The approach is sound: F-keys bubble from AiPanel up through CamelMonitor's dispatch chain to `handleGlobalKeys()`, which is exactly where F1 (help), F2 (actions), F3 (switch), F6 (shell), and F8 (AI) are handled. Popup windows opened via these keys get priority because their checks now precede the AI panel block.

One minor observation: the `isFunctionKey()` method would be a good candidate for a small unit test in `AiPanelTest.java` to guard against regressions if the key handling evolves further -- but that's not blocking.

---

**Scanner coverage**: ast-grep (`sg`) ran on both modified files with no findings. Other scanners (semgrep, gitleaks, PMD, SpotBugs) were not available in this environment. SpotBugs/ErrorProne require compilation -- see CI.

ammachado:
Thanks for this. I traced the full dispatch chain and the menu structures. The core fix is correct and nicely minimal, but I have a few asks before merge.

**What I verified holds:**
- Typing in the AI input is preserved: `PopupManager.handleKeyEvent` returns `false` when no popup is visible, so hoisting it above `aiPanel.isOpen()` is safe. Non-function keys still hit `AiPanel`'s greedy `return true` and never leak to global handlers.
- The fall-through surface is tightly bounded to function keys only, and F8 is correctly excluded from `isFunctionKey` and still closes the panel.
- The three parallel structures in `ActionsPopup` (`buildMcpSubmenuList`, `getActionLabels`, `renderMcpSubmenu`) stay in lockstep under the same `mcpEnabled` conditional, so the index->action mapping is preserved.

**Asks:**

1. **Add unit tests.** `AiPanelTest` already drives `handleKeyEvent` directly, so pinning this contract is cheap and would guard against a regression that reorders the F8 handling:
   ```java
   assertFalse(panel.handleKeyEvent(KeyEvent.ofKey(KeyCode.F1, KeyModifiers.NONE))); // passes through
   assertFalse(panel.handleKeyEvent(KeyEvent.ofKey(KeyCode.F2, KeyModifiers.NONE)));
   assertTrue(panel.handleKeyEvent(KeyEvent.ofKey(KeyCode.F8, KeyModifiers.NONE)));  // still consumed (close)
   ```
   The test plan in the description is currently all-manual.

2. **Comment the F8 exclusion in `isFunctionKey`.** Its omission from the list is load-bearing (F8 is the panel's own toggle) but invisible to a reader. Something like `// F8 is intentionally excluded - it closes the panel (handled above)` would prevent a maintainer from "completing" the sequence. The explicit enumeration itself is fine as-is.

**Two things to confirm (not blockers):**

3. The `isFunctionKey` check sits *before* the `thinking.get()` block, so F-keys now pass through even while a request is in flight (e.g. F6 shell / F2 actions mid-request). This seems reasonable, but it is a behavior change from the previous "swallow everything except Ctrl+C/Esc while thinking." Intended?

4. The `ActionsPopup` menu restructuring is logically separate from the F-key dispatch fix and isn't reflected in the title. Could you expand the description (or split it) so the changeset is self-describing?

davsclaus:
Addressed all review feedback:

1. **Unit tests added** — three new tests in `AiPanelTest`: `functionKeysPassThroughToGlobalHandlers`, `f8ClosesPanel`, and `functionKeysPassThroughWhileThinking`. These pin the F-key dispatch contract.

2. **F8 exclusion commented** — `isFunctionKey` now has a comment: `// F8 intentionally excluded — it closes the panel and is handled above`

3. **F-keys during thinking: intentional** — Yes, this behavior change is intended. F-keys (F2 actions, F3 switch, F6 shell, etc.) should remain accessible while the AI is processing a request. The user can open the actions menu or switch integrations without cancelling the running AI query. Ctrl+C/Esc still cancel the AI request.

4. **PR title and description updated** — Now covers both the F-key dispatch fix and the ActionsPopup menu restructuring with separate sections.

PR #24670, opened 2026-07-14
CAMEL-24055: QuartzEndpoint custom calendar name collision when multiple routes use customCalendar

prasanthrao:
# Description

\`QuartzEndpoint\` registered every custom calendar under the same hardcoded scheduler name
\`"CamelQuartzCustomCalendar"\`. When two or more routes in the same \`CamelContext\` each configured
a \`customCalendar\`, the second \`addCalendar(..., replace=true)\` call silently overwrote the first.
All triggers were wired to that same name via \`modifiedByCalendar(...)\`, so every trigger ended up
governed by whichever calendar registered last — not the one intended for it.

**Fix:** replace the hardcoded constant with a per-endpoint name derived from the trigger key:

\`\`\`
CamelQuartzCustomCalendar_<group>_<name>
\`\`\`

The trigger key is always set by \`QuartzComponent\` before the endpoint starts, is stable across
restarts (same URI → same trigger key), and is already guaranteed unique within the scheduler.

Custom calendars are now also deleted from the scheduler in \`doStop()\` to prevent stale entries
accumulating in JDBC-backed job stores.

A new regression test (\`QuartzCustomCalendarCollisionTest\`) verifies that two routes with different
custom calendars each register and retain their own independent calendar entry.

# Target

# Tracking

- [x] If this is a large change, bug fix, or code improvement, I checked there is a [JIRA issue](https://issues.apache.org/jira/browse/CAMEL-24055) filed for the change (usually before you start working on it).

# AI-assisted contributions

- [x] If this PR includes AI-generated code, commits have proper co-authorship attribution (e.g., \`Co-authored-by\` trailers) and the PR description identifies the AI tool used.

_Claude Code on behalf of Prasanth Rao_


components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
```diff
@@ -310,6 +310,10 @@ public void setCustomCalendar(Calendar customCalendar) {
         this.customCalendar = customCalendar;
     }
 
+    private String customCalendarName() {
```
gnodet, line 313: Good approach. Using the trigger key for the calendar name is the right choice -- it is set by `QuartzComponent.createEndpoint()` before `doStart()`, is unique within the scheduler, and is stable across restarts.

Minor pre-existing observation (not from this PR): there is no corresponding `scheduler.deleteCalendar(customCalendarName())` in `doStop()` / `removeJobInScheduler()`. This means custom calendars accumulate in the scheduler and are never cleaned up. For in-memory job stores this is harmless, but for JDBC-backed stores it could leave stale rows. Not blocking -- just noting for a potential follow-up.

components/camel-quartz/src/test/java/org/apache/camel/component/quartz/QuartzCustomCalendarCollisionTest.java
```diff
@@ -0,0 +1,78 @@
...
+        QuartzEndpoint endpointB = (QuartzEndpoint) context.getRoute("routeB").getConsumer().getEndpoint();
+
+        Calendar calA = scheduler.getCalendar(QuartzConstants.QUARTZ_CAMEL_CUSTOM_CALENDAR + "_" + endpointA.getGroupName()
+                                              + "_" + endpointA.getTriggerName());
+        Calendar calB = scheduler.getCalendar(QuartzConstants.QUARTZ_CAMEL_CUSTOM_CALENDAR + "_" + endpointB.getGroupName()
+                                              + "_" + endpointB.getTriggerName());
+
+        assertNotNull(calA, "Calendar for TimerA must be registered under its own unique name");
+        assertNotNull(calB, "Calendar for TimerB must be registered under its own unique name");
+        assertNotSame(calA, calB, "Each endpoint must register a distinct calendar object");
```
gnodet, line 76: Nice regression test. The `assertNotSame` check confirms the two calendar objects are distinct instances. You might also consider adding `assertSame(calA, context.getRegistry().lookupByName("calendarA"))` to verify each endpoint's registered calendar is the _exact_ instance from the registry (not just a different instance), but the current assertions are sufficient to prevent the original collision bug.

gnodet (approved):
## Review: CAMEL-24055 - Fix custom calendar name collision in QuartzEndpoint

**Verdict: APPROVE** | Confidence: HIGH

Good fix for a real bug. The problem is clear: when multiple routes each configure a `customCalendar`, they all registered under the hardcoded name `"CamelQuartzCustomCalendar"`. With `replace=true`, the second registration silently overwrote the first, causing all triggers to use whichever calendar was registered last.

The fix correctly derives a per-endpoint calendar name from the trigger key (`CamelQuartzCustomCalendar_<group>_<name>`), which is guaranteed unique within the scheduler and stable across restarts (same URI produces the same trigger key).

### What I checked

- **Correctness**: All 4 references to the old hardcoded name (1 in `addCalendar`, 3 in `modifiedByCalendar`) are updated to use `customCalendarName()`. The registration and the trigger configuration are consistent.
- **`triggerKey` availability**: `triggerKey` is set by `QuartzComponent.createEndpoint()` before `doStart()` is ever called, so there is no risk of NPE in `customCalendarName()`.
- **Backwards compatibility**: Existing single-calendar setups will now use a longer calendar name instead of the bare constant, but since the calendar name is an internal implementation detail (not user-facing), this is not a breaking change.
- **Thread safety**: No thread-safety concerns. The `triggerKey` is set once during endpoint creation and is effectively immutable after that.
- **Test coverage**: The new `QuartzCustomCalendarCollisionTest` directly verifies the collision scenario -- two routes with distinct calendars each get their own registered calendar. The existing tests (`QuartzCustomCalendarFireTest`, `QuartzCustomCalendarNoFireTest`) are updated to look up the calendar under the new naming scheme.
- **Git history**: Reviewed the history of `QuartzEndpoint.java` -- no prior changes conflict with this fix. The custom calendar feature was added in CAMEL-20115 and this is the first change to the calendar naming.

- **Pre-existing: no calendar cleanup in `doStop()`** -- when an endpoint is stopped or deleted, the calendar it registered with the scheduler is never removed (`scheduler.deleteCalendar(...)` is not called). This was true before this PR and is not a regression. For most use cases this is harmless since `addCalendar(replace=true)` handles re-registration on restart, but it could lead to stale calendar entries accumulating in clustered/JDBC job stores. Worth a follow-up issue if needed.

### Scanner coverage

- **ast-grep (`sg`)**: ran with Java rules from `~/.oss-helper/rules/java` -- no findings
- **Not available**: semgrep, gitleaks, PMD, ruff, bandit, eslint, golangci-lint, shellcheck, hadolint, cppcheck
- **Coverage gap**: SpotBugs/Checkstyle require compilation -- see CI results

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
prasanthrao: Thanks. I too felt this in my second review and have pushed the cleanup code

components/camel-quartz/src/test/java/org/apache/camel/component/quartz/QuartzCustomCalendarCollisionTest.java
prasanthrao: Thanks for the suggestion. Unfortunately this assertion cannot work in practice.`org.quartz.simpl.RAMJobStore.storeCalendar()` clones the calendar object before storing it, so `scheduler.getCalendar(name)` always returns a copy, never the original instance. assertSame would always fail regardless of the fix.

`assertNotSame(calA, calB)` combined with assertNotNull on each is the best assertion we can do today.

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
```diff
@@ -328,14 +332,16 @@ protected void doStart() throws Exception {
...
         if (ObjectHelper.isNotEmpty(customCalendar)) {
-            getComponent().getScheduler().addCalendar(QuartzConstants.QUARTZ_CAMEL_CUSTOM_CALENDAR, customCalendar, true,
-                    false);
+            getComponent().getScheduler().addCalendar(customCalendarName(), customCalendar, true, false);
         }
         addJobInScheduler();
     }
 
     @Override
     protected void doStop() throws Exception {
```
davsclaus, line 341: **Bug:** `deleteCalendar()` is called before `removeJobInScheduler()`. Quartz's `RAMJobStore.removeCalendar()` iterates all triggers and throws `JobPersistenceException("Calender cannot be removed if it referenced by a Trigger!")` if any trigger still references the calendar name.

Since `removeJobInScheduler()` (which calls `scheduler.unscheduleJob(triggerKey)`) hasn't run yet at this point, the trigger still exists and references the calendar — so this will throw on every endpoint stop with the default `deleteJob=true`.

Additionally, `deleteCalendar` should only be attempted when the trigger was actually removed. When `deleteJob=false` (trigger is paused or left running), or when the scheduler is clustered (the `!isClustered` guard in `removeJobInScheduler` skips the unschedule), the trigger persists and `deleteCalendar` would also throw.

```suggestion
    protected void doStop() throws Exception {
        removeJobInScheduler();
        if (deleteJob && ObjectHelper.isNotEmpty(customCalendar)) {
            Scheduler scheduler = getComponent().getScheduler();
            if (scheduler != null && !scheduler.isShutdown()) {
                try {
                    scheduler.deleteCalendar(customCalendarName());
                } catch (SchedulerException e) {
                    LOG.debug("Could not delete calendar {}: {}", customCalendarName(), e.getMessage());
                }
            }
        }
    }

**Review by davsclaus — CHANGES_REQUESTED** [review]
Thanks for the fix, @prasanthrao — the core idea of deriving per-endpoint calendar names from the trigger key is solid and correctly solves the collision bug.

However, the new calendar cleanup in `doStop()` has an ordering issue that will throw at runtime. Details below.

---

**Summary of findings:**

1. **`doStop()` ordering bug (blocking):** `deleteCalendar()` is called before `removeJobInScheduler()`. Quartz's `RAMJobStore.removeCalendar()` throws `JobPersistenceException("Calender cannot be removed if it referenced by a Trigger!")` if any trigger still references the calendar. Since the trigger is only removed in `removeJobInScheduler()`, which hasn't run yet, this will throw on every endpoint stop with the default `deleteJob=true`.

2. **Calendar deletion should be conditional (blocking):** Even with the order fixed, `deleteCalendar` should only run when the trigger was actually removed. When `deleteJob=false` (with `pauseJob=true`), the trigger persists and still references the calendar. Similarly, when `deleteJob=true` but the scheduler is clustered, the trigger is not removed. In both cases `deleteCalendar` would throw.

3. **Test doesn't cover stop path (minor):** The collision test verifies registration but not cleanup. Errors from `doStop()` are typically caught and logged by the Camel lifecycle during test teardown rather than failing the test, so the ordering bug above would not surface as a test failure.

4. **No upgrade guide note (minor):** The calendar naming change could leave orphaned entries in JDBC-backed job stores. Worth a brief mention in the upgrade guide.

This review does not replace specialized AI review tools or static analysis.

_This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

_Claude Code on behalf of davsclaus_

File: components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java (hunk 3 above)
Comment (prasanthrao) [reply components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java:341]: Good catch — fixed. Moved `deleteCalendar` after `removeJobInScheduler` so the trigger is unscheduled first, added the `deleteJob` guard and the `scheduler != null && !isShutdown()` check, and wrapped in `try/catch` to tolerate clustered stores where the unschedule may be skipped. Thanks!

File: components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java [hunk 4]
```diff
@@ -328,15 +332,24 @@ protected void doStart() throws Exception {
...
         removeJobInScheduler();
+        if (deleteJob && ObjectHelper.isNotEmpty(customCalendar)) {
+            Scheduler scheduler = getComponent().getScheduler();
+            if (scheduler != null && !scheduler.isShutdown()) {
+                try {
+                    scheduler.deleteCalendar(customCalendarName());
+                } catch (SchedulerException e) {
+                    LOG.debug("Could not delete calendar {}: {}", customCalendarName(), e.getMessage());
+                }
+            }
```
davsclaus: (Minor suggestion) The existing `removeJobInScheduler()` skips unscheduling in clustered mode (`!isClustered` guard). The new calendar cleanup doesn't have this guard — so in clustered mode, the trigger isn't removed but `deleteCalendar()` is attempted. Quartz refuses to delete a calendar with active trigger references, so this will always throw and get caught at DEBUG level. Not a functional bug, but attempting and failing on every stop in clustered deployments is unnecessary.

Consider adding the `isClustered` check for consistency:

```suggestion
        if (deleteJob && ObjectHelper.isNotEmpty(customCalendar)) {
            Scheduler scheduler = getComponent().getScheduler();
            if (scheduler != null && !scheduler.isShutdown()) {
                boolean isClustered = scheduler.getMetaData().isJobStoreClustered();
                if (!isClustered) {
                    try {
                        scheduler.deleteCalendar(customCalendarName());
                    } catch (SchedulerException e) {
                        LOG.debug("Could not delete calendar {}: {}", customCalendarName(), e.getMessage());
                    }
                }
            }
        }
```

davsclaus (commented):
Thanks for the fix @prasanthrao — the bug is real and the approach is sound. Using the trigger key for the calendar name is the right choice (unique, stable, already set before `doStart()` runs).

One minor suggestion about clustered-mode consistency, otherwise this looks good.

Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
prasanthrao: This commit should ideally fix the concerns with calender deletion https://github.com/apache/camel/pull/24670/changes/9e26f1502c245b4567c02a4e7b2fe9b54ed7c49a

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
prasanthrao: This commit should ideally fix the concerns with calender deletion https://github.com/apache/camel/commit/9e26f1502c245b4567c02a4e7b2fe9b54ed7c49a

components/camel-quartz/src/main/java/org/apache/camel/component/quartz/QuartzEndpoint.java
```diff
@@ -352,6 +355,13 @@ private void removeJobInScheduler() throws Exception {
                 scheduler.unscheduleJob(triggerKey);
 
                 jobAdded.set(false);
+                if (ObjectHelper.isNotEmpty(customCalendar)) {
+                    try {
+                        scheduler.deleteCalendar(customCalendarName());
+                    } catch (SchedulerException e) {
+                        LOG.error("Could not delete calendar {}: {}", customCalendarName(), e.getMessage());
```
davsclaus: Minor: consider `LOG.warn` here instead of `LOG.error` — this is a cleanup failure during shutdown, not a runtime error. Also, passing the exception as the last arg captures the stack trace:

```suggestion
                    } catch (SchedulerException e) {
                        LOG.warn("Could not delete calendar {}: {}", customCalendarName(), e.getMessage(), e);
```

davsclaus (approved):
Thank you for working on this, @prasanthrao. The fix is correct — replacing the hardcoded calendar name with a per-endpoint name derived from the trigger key is the right approach, and the regression test covers the collision scenario well.

**One minor suggestion** (non-blocking): the `LOG.error` in `removeJobInScheduler` for calendar deletion failure could be `LOG.warn` instead — this is consistent with the existing logging level in this file (line 392), and a cleanup failure during `doStop()` is a warning-level event rather than an error. Also, passing the exception as the last argument would capture the stack trace for debugging:

```java
LOG.warn("Could not delete calendar {}: {}", customCalendarName(), e.getMessage(), e);
```

**What looks good:**
- Calendar name `CamelQuartzCustomCalendar_<group>_<name>` is stable across restarts (same URI → same trigger key) and unique within the scheduler.
- Calendar cleanup in `doStop()` is correctly scoped inside the `deleteJob` block — paused jobs retain their calendar.
- Existing tests updated consistently.
- New regression test directly verifies the bug fix.

---

PR #24677, opened 2026-07-14
CAMEL-20141: camel-joor - Fix hot-reload of inline bean scripts with multiple Java files

davsclaus:
## Summary

Fixes [CAMEL-20141](https://issues.apache.org/jira/browse/CAMEL-20141): when running `camel run * --dev` with YAML routes that use inline Java bean scripts referencing multiple `.java` files (e.g. a `Customer` POJO and a `CustomerBuilder`), editing any `.java` file caused a reload failure with `NoSuchBeanException: Creating bean using script returned null`.

**Root cause:** During hot-reload, only the changed `.java` file was recompiled in a new `ByteArrayClassLoader`. Other Java files (e.g. a builder class) stayed loaded from the original classloader. When the builder's `build()` method created an object of the changed class, JVM class identity rules meant the returned object's class (from the old classloader) didn't match the `resultType` (resolved from the new classloader), so `convertTo()` returned null.

**Fix:**
- `JavaRoutesBuilderLoader` now tracks all previously compiled Java resources and includes them in any recompilation, ensuring all classes share the same `ByteArrayClassLoader` and class identity is consistent. Re-added resources go through the `CompilePreProcessor` chain.
- `ByteArrayClassLoader` now accepts a custom parent classloader so `MultiCompile` can set `CamelJoorClassLoader` as parent, enabling proper class resolution through the `JavaJoorClassLoader` chain
- `CamelJoorClassLoader` overrides `loadClass(String, boolean)` to delegate through `doLoadClass()` which checks `JavaJoorClassLoader` first, and honors the `resolve` parameter
- `JoorScriptingCompiler` cleans up previously compiled script classes to prevent classloader leaks during reloads, using thread-safe `ConcurrentHashMap.newKeySet()` for tracking

## Test plan

- [x] All existing tests pass (53 in camel-joor, 29 in camel-java-joor-dsl)
- [x] Manual test: `camel run * --dev` with the `bean-inlined-code` example — editing `Customer.java` now correctly reloads the route without errors

### DIFF
--- a/components/camel-joor/src/main/java/org/apache/camel/language/joor/ByteArrayClassLoader.java
+++ b/components/camel-joor/src/main/java/org/apache/camel/language/joor/ByteArrayClassLoader.java
@@ -30,6 +30,11 @@ final class ByteArrayClassLoader extends ClassLoader {
         this.classes = classes;
     }
 
+    public ByteArrayClassLoader(ClassLoader parent, Map<String, byte[]> classes) {
+        super(parent);
+        this.classes = classes;
+    }
+
     @Override
     protected Class<?> findClass(String name) throws ClassNotFoundException {
         byte[] bytes = classes.get(name);
--- a/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
+++ b/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
@@ -75,7 +75,11 @@ public class CamelJoorClassLoader extends URLClassLoader {
 
     @Override
     protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
-        return doLoadClass(name);
+        Class<?> c = doLoadClass(name);
+        if (resolve) {
+            resolveClass(c);
+        }
+        return c;
     }
 
     public Class<?> doLoadClass(String name) throws ClassNotFoundException {
--- a/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
+++ b/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
@@ -53,8 +53,7 @@ public class CamelJoorClassLoader extends URLClassLoader {
         CompileStrategy cs = camelContext.getCamelContextExtension().getContextPlugin(CompileStrategy.class);
         if (cs != null && cs.getWorkDir() != null) {
             try {
-                File dir = new File(cs.getWorkDir() + "/joor");
-                answer.add(dir.toURI().toURL());
+                answer.add(new File(cs.getWorkDir()).toURI().toURL());
             } catch (MalformedURLException e) {
                 // ignore
             }
--- a/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
+++ b/components/camel-joor/src/main/java/org/apache/camel/language/joor/CamelJoorClassLoader.java
@@ -73,6 +73,11 @@ public class CamelJoorClassLoader extends URLClassLoader {
         return doLoadClass(name);
     }
 
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+        return doLoadClass(name);
+    }
+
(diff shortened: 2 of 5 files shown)


gnodet (commented):
## Review: CAMEL-20141 — camel-joor - Fix hot-reload of inline bean scripts with multiple Java files

Nice work on diagnosing and fixing this subtle classloader identity bug! The architectural approach — recompiling all Java files together during hot-reload to maintain class identity across classloaders — is the correct solution. The supporting changes (parent classloader pass-through, compile directory alignment, defensive `ArrayList` copy) are coherent and well-reasoned.

A few items I'd like to understand better before approving:

### Questions / Suggestions

**1. `loadClass(String, boolean)` ignores the `resolve` parameter**
`CamelJoorClassLoader.java` — new `loadClass(name, resolve)` override

The `resolve` parameter controls whether the JVM performs class linking. Ignoring it could defer class resolution errors to a later point. The safest approach would be:

```java
@Override
protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
    Class<?> c = doLoadClass(name);
    if (resolve) {
        resolveClass(c);
    }
    return c;
}
```

Is ignoring `resolve` intentional here? In practice, callers in this chain may never pass `true`, but the JVM's internal verification calls do use `resolve = true`.

**2. Re-added resources skip `CompilePreProcessor` during hot-reload**
`JavaRoutesBuilderLoader.java` — lines 217-228

When previously compiled resources are re-added to the compilation unit during hot-reload, they bypass the `CompilePreProcessor` chain (compare with lines 195-213 where initial resources go through `preCompile()`). If any pre-processor performs content transformation, the re-added resources would be compiled without those transformations. Is this intentional?

**3. `workDir` path no longer appends `/joor`**
`JoorScriptingCompiler.java` line 96, `CamelJoorClassLoader.java` line 56

This fixes a pre-existing inconsistency between `JoorScriptingCompiler` (which wrote to `workDir + "/joor"`) and `JavaRoutesBuilderLoader` (which wrote to `workDir`). After this PR both are consistent. Worth noting that any cached `.class` files in the old `workDir/joor/` subdirectory will become orphaned on upgrade — minor since it's a transient cache.

**4. Test coverage for hot-reload scenario**

The PR mentions manual testing with `camel run * --dev`. Per the project guidelines, tests are expected for bug fixes. The classloader-split scenario is the exact kind of subtle regression that automated tests would catch. I understand testing classloader behavior in unit tests is non-trivial — would a follow-up ticket for integration test coverage be acceptable?

**5. (Minor) `compiledClassNames` thread safety**
`JoorScriptingCompiler.java` — `LinkedHashSet` is not thread-safe. The `compile()` method appears to be called during context startup (not at message-processing time), so the risk is low, but `ConcurrentHashMap.newKeySet()` would be safer if there's any concurrent access path.

---

davsclaus:

Thanks for the thorough review @gnodet! Addressed in 74de2ad:

**1. `loadClass(String, boolean)` ignores `resolve`** — Fixed. Now calls `resolveClass(c)` when `resolve=true`.

**2. Re-added resources skip `CompilePreProcessor`** — Fixed. The hot-reload loop now runs the `CompilePreProcessor` chain on re-added resources, consistent with the initial compilation path.

**3. `workDir` path no longer appends `/joor`** — Correct, this aligns both `JoorScriptingCompiler` and `JavaRoutesBuilderLoader` to use the same `workDir`. Old cached `.class` files under `workDir/joor/` will be orphaned but are harmless since it's a transient dev-mode cache.

**5. `compiledClassNames` thread safety** — Fixed. Switched from `LinkedHashSet` to `ConcurrentHashMap.newKeySet()`.

gnodet (approved):
Follow-up review after 74de2ad addressed all feedback from the initial review:

1. ✅ `loadClass(String, boolean)` now properly calls `resolveClass(c)` when `resolve=true`
2. ✅ Re-added resources during hot-reload now go through the `CompilePreProcessor` chain, consistent with the initial compilation path
3. ✅ `compiledClassNames` switched to `ConcurrentHashMap.newKeySet()` for thread safety

The classloader identity fix is sound — recompiling all Java files together during hot-reload ensures consistent class identity across classloaders. The `ByteArrayClassLoader` parent-classloader pass-through and `workDir` alignment are coherent supporting changes.

CI is green.

PR #24678, opened 2026-07-14
CAMEL-23844: Camel-PQC - use the mapped JCE algorithm name when restoring the secret key

oscerd:
# CAMEL-23844: Camel-PQC — use the mapped JCE algorithm name when restoring the secret key

## The bug

`PQCProducer.extractSecretKeyFromEncapsulation()` built the restored key as:

```java
new SecretKeySpec(payload.getEncoded(), getConfiguration().getSymmetricKeyAlgorithm())
```

That passes the **raw `PQCSymmetricAlgorithms` enum name** as the `SecretKeySpec` algorithm, whereas
`extractEncapsulation()` correctly maps it via `PQCSymmetricAlgorithms.valueOf(...).getAlgorithm()`.

Only two enum names differ from their JCE names, and the impact is very different for each:

| Enum name | JCE name | `Cipher.getInstance(<enum name>)` |
|---|---|---|
| `DESEDE` | `DESede` | **OK** — JCA lookup is case-insensitive |
| `GOST3412_2015` | `GOST3412-2015` | **`NoSuchAlgorithmException`** — underscore is not a valid JCE name |

So for **GOST3412_2015** the secret key handed back to the route carried an algorithm label that
cannot be resolved to a cipher at all, making the restored key unusable downstream (verified against
BouncyCastle 1.85).

The existing KEM round-trip tests do not catch this because they encrypt with an explicitly configured
`CryptoDataFormat("DESede")` and never read the key's own label.

## The fix

Use `PQCSymmetricAlgorithms.valueOf(getConfiguration().getSymmetricKeyAlgorithm()).getAlgorithm()` —
the same mapping `extractEncapsulation()` already uses — in **both** extract paths:
`extractSecretKeyFromEncapsulation` and `hybridExtractSecretKeyFromEncapsulation`.

## Testing

New `PQCExtractSecretKeyAlgorithmNameTest` (2 tests) runs a full
generate → extract → extract-secret-key route and asserts the restored key's algorithm is the **JCE**
name, and that `Cipher.getInstance(key.getAlgorithm())` resolves.

Verified it is a true regression test — **without** the fix it fails with:

```
expected: <GOST3412-2015> but was: <GOST3412_2015>
expected: <DESede> but was: <DESEDE>
```

All 17 existing KEM round-trip tests still pass. Full reactor build green.

## Note for reviewers (not changed here)

`generateEncapsulation()` passes the same **raw** enum name into `KEMGenerateSpec`, so the
`SecretKeyWithEncapsulation` it produces is labelled `GOST3412_2015` / `DESEDE` too. I deliberately left
that alone: the name there is only a label (it is not part of the KEM derivation — generate and extract
already disagree today and the round trip still works), and changing it would alter a **visible** output
label and require updating existing test assertions. Happy to align it in a follow-up if you'd prefer
full consistency.

## Backport

This is a bug → **4.18.x (4.18.4)** and **4.14.x (4.14.9)**. On those branches only the standard extract
path exists (the hybrid operations came later), so the backport is the single-line fix plus the test.

---

### DIFF
--- a/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
+++ b/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
@@ -603,7 +603,12 @@ public class PQCProducer extends DefaultProducer {
             throw new IllegalArgumentException("Symmetric Algorithm needs to be specified");
         }
 
-        SecretKey restoredKey = new SecretKeySpec(payload.getEncoded(), getConfiguration().getSymmetricKeyAlgorithm());
+        // Use the mapped JCE algorithm name (as extractEncapsulation does), not the raw enum name: for algorithms
+        // whose enum name differs from the JCE name (for example GOST3412_2015 -> GOST3412-2015) the raw name is not a
+        // resolvable cipher algorithm, so the restored key would carry an unusable algorithm label
+        SecretKey restoredKey = new SecretKeySpec(
+                payload.getEncoded(),
+                PQCSymmetricAlgorithms.valueOf(getConfiguration().getSymmetricKeyAlgorithm()).getAlgorithm());
 
         if (!getConfiguration().isStoreExtractedSecretKeyAsHeader()) {
             exchange.getMessage().setBody(restoredKey, SecretKey.class);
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

gnodet (approved):
Excellent bug fix! The root-cause analysis is thorough — the table mapping enum names to JCE names in the PR description is the gold standard for a bug-fix PR. The fix is minimal, targeted, and mirrors the correct pattern already used at line 535 (`extractEncapsulation`).

**What I like:**
- The regression test (`PQCExtractSecretKeyAlgorithmNameTest`) runs the full generate-then-extract-then-restore pipeline and asserts on the restored key's JCE algorithm name *and* verifies `Cipher.getInstance()` resolves. Both GOST and DESede cases are covered.
- The deliberate scoping decision to leave `generateEncapsulation`'s raw enum usage alone (since the name there is only a KEM label) shows good judgment.

**Observations (not blocking):**
- The same raw-enum-name inconsistency exists in `HybridKEM` internals (lines 99, 113, 170, 183) — the `SecretKey` stored in `HYBRID_SECRET_KEY` header still carries the raw name. The fix correctly re-wraps it at the `hybridExtractSecretKeyFromEncapsulation` step, so the consumer path produces correct keys. Worth noting as a follow-up if `HYBRID_SECRET_KEY` is ever consumed directly.
- The expression `PQCSymmetricAlgorithms.valueOf(getConfiguration().getSymmetricKeyAlgorithm()).getAlgorithm()` now appears in three places — a small `resolveJceAlgorithmName()` helper could reduce duplication (purely optional).

CI builds are still queued at the time of review — the change is structurally safe (same mapping pattern already proven at line 535). Clean PR — well done! 👍

---

PR #24680, opened 2026-07-14
[backport camel-4.14.x] CAMEL-23844: Camel-PQC - use the mapped JCE algorithm name when restoring the secret key

oscerd:
Backport of #24678 to `camel-4.14.x` (fixVersion **4.14.9**).

`extractSecretKeyFromEncapsulation()` built the restored key with the **raw** `PQCSymmetricAlgorithms` enum name instead of the mapped JCE name (unlike `extractEncapsulation()`, which maps it correctly).

| Enum name | JCE name | `Cipher.getInstance(<enum name>)` |
|---|---|---|
| `DESEDE` | `DESede` | OK (JCA lookup is case-insensitive) |
| `GOST3412_2015` | `GOST3412-2015` | **`NoSuchAlgorithmException`** |

So for GOST3412_2015 the restored key carried an algorithm label that cannot be resolved to a cipher, making it unusable downstream.

**Difference from main:** the hybrid extract path does not exist on this branch, so only the standard path is patched. The test is the same (adapted to `test.junit5`).

Validated locally: `PQCExtractSecretKeyAlgorithmNameTest` 2/2 green on this branch; module build clean, no generated drift.

### DIFF
--- a/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
+++ b/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
@@ -200,7 +200,12 @@ public class PQCProducer extends DefaultProducer {
             throw new IllegalArgumentException("Symmetric Algorithm needs to be specified");
         }
 
-        SecretKey restoredKey = new SecretKeySpec(payload.getEncoded(), getConfiguration().getSymmetricKeyAlgorithm());
+        // Use the mapped JCE algorithm name (as extractEncapsulation does), not the raw enum name: for algorithms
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean backport of #24678 to `camel-4.14.x`.

The fix correctly uses `PQCSymmetricAlgorithms.valueOf(...).getAlgorithm()` instead of the raw enum name — verified against the enum source on this branch. For `GOST3412_2015` the raw name is not a resolvable cipher algorithm, so the restored key was unusable downstream.

Tests cover both divergent enum entries (`GOST3412_2015` → `GOST3412-2015` and `DESEDE` → `DESede`) and prove the key is usable via `Cipher.getInstance()`.

PR #24683, opened 2026-07-14
CAMEL-24060: Use reflection for optional PQCMicrometerMetrics to avoid GraalVM native image failure

JiriOndrusek:
## Summary

- `PQCProducer.createMetrics()` directly references `PQCMicrometerMetrics`, causing GraalVM native image compilation to fail with `NoClassDefFoundError: io/micrometer/core/instrument/Meter` when micrometer is not on the classpath
- Replace the direct class reference with reflection via `CamelContext.getClassResolver()` so GraalVM's static analysis never follows the link to micrometer classes
- The existing runtime classpath guard and catch-all are preserved

## Test plan

- [x] `mvn install -f components/camel-pqc` passes
- [ ] Camel Quarkus PQC native integration test passes without adding micrometer-core dependency


gnodet (approved):
Clean fix. Replacing the direct `PQCMicrometerMetrics` class reference with reflection via `CamelContext.getClassResolver()` is the standard Camel pattern for optional dependencies and avoids GraalVM's static analysis following the link to micrometer classes.

The null check on `resolveClass()` and the existing `catch (Throwable)` fallback to `PQCMetrics.NOOP` are both preserved — runtime behavior is unchanged.

davsclaus (approved):

PR #24684, opened 2026-07-14
CAMEL-23994: Fix inverted key-format check in KafkaResumeAdapter

davsclaus:
## Summary

Fixes the first bug reported in [CAMEL-23994](https://issues.apache.org/jira/browse/CAMEL-23994): the condition in `KafkaResumeAdapter.deserialize()` was inverted, causing all valid `topic/partition` keys to be silently discarded and invalid keys to trigger `ArrayIndexOutOfBoundsException`.

**Changes:**
- Fixed the inverted condition from `keyParts == null || keyParts.length != 2` to `keyParts.length == 2` (the `null` check was also dead code since `String.split()` never returns `null`)
- Fixed typo in warning message ("in invalid" → "an invalid")
- Added `KafkaResumeAdapterTest` with tests for valid keys, multiple keys, invalid keys (no slash, too many parts), and invalid value types

**Impact:** Consumers configured with a `resumeStrategy` never restored persisted offsets on restart — `resumeCache` stayed empty, and the consumer silently restarted from the group/`auto.offset.reset` position.

**Bug present since:** CAMEL-18688 (commit `59046fe140ff`, Nov 2022).

## Test plan

- [x] Added unit tests for `KafkaResumeAdapter.deserialize()` covering valid keys, invalid keys, and invalid values
- [x] All tests pass (`mvn test -Dtest=KafkaResumeAdapterTest`)

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/resume/KafkaResumeAdapter.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/resume/KafkaResumeAdapter.java
@@ -57,7 +57,7 @@ public class KafkaResumeAdapter implements ResumeAdapter, Deserializable, Cachea
         if (keyObj instanceof String key) {
 
             final String[] keyParts = key.split("/");
-            if (keyParts == null || keyParts.length != 2) {
+            if (keyParts.length == 2) {
 
                 String topic = keyParts[0];
                 int partition = Integer.parseInt(keyParts[1]);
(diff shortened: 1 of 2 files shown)


Croway (approved):

gnodet (approved):
Correct fix — the `!=` guard was inverted since CAMEL-18688 (commit `59046fe140f`, Nov 2022), causing valid `topic/partition` keys to hit the warning branch and invalid keys to enter the processing block. The `== 2` condition is right, the dead `null` check removal is appropriate (`String.split()` never returns `null`), and the typo fix is a nice bonus.

Tests cover the key scenarios well: valid key, multiple keys, invalid key formats, and wrong value types.

PR #24685, opened 2026-07-14
CAMEL-23994: Fix off-by-one when writing to offset repository

davsclaus:
## Summary

Fixes the fourth bug reported in [CAMEL-23994](https://issues.apache.org/jira/browse/CAMEL-23994): `SyncCommitManager` and `AsyncCommitManager` stored `offset+1` in the offset repository, but `OffsetPartitionAssignmentAdapter.resumeFromOffset` adds `+1` when seeking (`state+1`). This double increment caused one message per partition to be lost on every restart when using an `offsetRepository`.

**The round-trip before the fix:**
- Record at offset 41 processed → repo stores 42 → on restart seeks to 42+1=43 → record 42 is lost

**After the fix:**
- Record at offset 41 processed → repo stores 41 → on restart seeks to 41+1=42 → correct

**Changes:**
- `SyncCommitManager`: save the raw record offset (`offset`) instead of `lastOffset` (offset+1) to the offset repository. The Kafka commit still correctly uses `offset+1` (Kafka convention).
- `AsyncCommitManager`: save `committed offset - 1` (the raw record offset) to the repository instead of the already-incremented `OffsetAndMetadata.offset()`.
- Updated `BaseManualCommitTestSupport` assertion from `"5"` to `"4"` to reflect the corrected stored value (last read offset, not next-to-read offset).

Both managers now match the contract established by `CommitToOffsetManager` (which already stored the raw offset correctly) and the comment in `resumeFromOffset`: _"The state contains the last read offset, so you need to seek from the next one"_.

**Bug present since:** CAMEL-18717 (Nov 2022).

## Test plan

- [x] All 141 unit tests pass
- [x] Updated `BaseManualCommitTestSupport` offset assertion to match new behavior
- [ ] CI integration tests (`KafkaConsumerSyncWithOffsetRepoCommitIT`, `KafkaConsumerAsyncWithOffsetRepoCommitIT`) will validate the full round-trip

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/AsyncCommitManager.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/AsyncCommitManager.java
@@ -94,7 +94,10 @@ public class AsyncCommitManager extends AbstractCommitManager {
         if (exception == null) {
             if (offsetRepository != null) {
                 for (var entry : committed.entrySet()) {
-                    saveStateToOffsetRepository(entry.getKey(), entry.getValue().offset(), offsetRepository);
+                    // Store the last read offset (Kafka committed offset - 1) to match
+                    // the contract in OffsetPartitionAssignmentAdapter.resumeFromOffset
+                    // which seeks to state + 1
+                    saveStateToOffsetRepository(entry.getKey(), entry.getValue().offset() - 1, offsetRepository);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Correct fix. Verified the full round-trip:

- `OffsetPartitionAssignmentAdapter.resumeFromOffset()` does `state + 1` and its comment says: _"The state contains the last read offset, so you need to seek from the next one"_
- `CommitToOffsetManager` was already storing the raw offset (correct)
- `SyncCommitManager` was storing `lastOffset` (offset+1) — double-incremented on resume → one message lost per partition per restart
- `AsyncCommitManager` was storing `OffsetAndMetadata.offset()` (already Kafka-incremented) — same double-increment issue

Both managers now store the raw last-read offset, matching `CommitToOffsetManager`'s contract. Test assertion update from `"5"` to `"4"` correctly reflects 5 records at offsets 0–4.

PR #24687, opened 2026-07-14
CAMEL-23994: Fix pollOnError seek skipping messages on all partitions

davsclaus:
## Summary

Fixes the second bug reported in [CAMEL-23994](https://issues.apache.org/jira/browse/CAMEL-23994): when a poll-loop exception was caught with `pollOnError=DISCARD` or `ERROR_HANDLER`, `SeekUtil.seekToNextOffset` was called with a hardcoded offset of -1 (lost during refactoring in CAMEL-16064), causing it to seek `position(tp) + 1` on ALL assigned partitions — silently dropping one unread message per partition.

**Before:** Any exception during polling → seek position+1 on every partition → silent message loss on all partitions
**After:**
- `RecordDeserializationException` (poison message) → seek only the affected partition to `offset+1`
- Other exceptions (transient Kafka errors, commit failures) → no position advancement since the error may not be record-related

**Changes:**
- Rewrote `SeekUtil.seekToNextOffset` to accept the `Exception` and use `RecordDeserializationException.topicPartition()` / `offset()` for targeted seeking
- Updated `DiscardErrorStrategy` and `BridgeErrorStrategy` to pass the exception
- The `PollExceptionStrategy.handle(long, Exception)` public API is unchanged

**Note:** The `partitionLastOffset` parameter in `PollExceptionStrategy.handle()` has been effectively dead since the CAMEL-16064 refactoring (always -1). The strategies now use the exception for seeking instead.

## Test plan

- [x] All 141 unit tests pass
- [ ] CI integration tests will validate poll error handling

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/BridgeErrorStrategy.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/errorhandler/BridgeErrorStrategy.java
@@ -48,7 +48,7 @@ public class BridgeErrorStrategy implements PollExceptionStrategy {
         // use bridge error handler to route with exception
         recordFetcher.getBridge().handleException(exception);
         // skip this poison message and seek to the next message
-        SeekUtil.seekToNextOffset(consumer, partitionLastOffset);
+        SeekUtil.seekToNextOffset(consumer, exception);
 
         if (exception instanceof AuthenticationException || exception instanceof AuthorizationException) {
             continueFlag = false;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Correct fix. Confirmed that `partitionLastOffset` has been hardcoded to `-1` in `KafkaFetchRecords` since the CAMEL-16064 refactoring, so the old `SeekUtil` always fell into the `else if` branch that advanced ALL assigned partitions by `position(tp) + 1` — silently dropping one message per partition on any poll exception, regardless of which partition had the error.

The new approach is significantly better:
- **`RecordDeserializationException`** (poison message): extracts `topicPartition()` and `offset()` from the exception itself → seeks only the affected partition to `offset + 1`
- **Other exceptions** (transient Kafka errors, commit failures): logs a warning and doesn't advance any partition — correct, since the error isn't record-related

Only `DiscardErrorStrategy` and `BridgeErrorStrategy` call `seekToNextOffset`; `ReconnectErrorStrategy` doesn't seek. `SeekUtil` is internal, so the signature change is safe.

oscerd (approved):

PR #24688, opened 2026-07-14
CAMEL-23994: Fix batching auto-commit committing unprocessed record offsets

davsclaus:

## Summary

Fixes **Bug 3** from [CAMEL-23994](https://issues.apache.org/jira/browse/CAMEL-23994): batching auto-commit over-commits unprocessed record offsets.

**Root cause:** In batching mode, `CommitSynchronization.onComplete()` called the no-arg `commitManager.commit()` which commits the Kafka consumer's **current position**. Because Kafka pre-fetches records ahead of what was actually delivered to the processor, the committed offset jumps past records that haven't been processed yet. On restart, those skipped records are silently lost.

**Fix:** Compute the maximum offset per `TopicPartition` from the actual batch exchanges (via their `KafkaConstants.TOPIC`, `PARTITION`, and `OFFSET` headers), then explicitly call `commitManager.recordOffset()` + `commitManager.commit(partition)` for each partition. This ensures only offsets of actually-processed records are committed — matching the behavior of the streaming facade.

## Changes

- Added `computeBatchOffsets(List<Exchange>)` method that extracts max offset per TopicPartition from batch exchange headers
- Modified `CommitSynchronization` inner class to accept and use explicit batch offsets instead of committing the consumer's current position
- Updated `autoCommitResultProcessing()` to compute and pass batch offsets to `CommitSynchronization`

## Test plan

- [x] All 141 existing unit tests pass (`mvn test` in camel-kafka module)
- [ ] CI integration tests (Kafka testcontainers)

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/batching/KafkaRecordBatchingProcessor.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/support/batching/KafkaRecordBatchingProcessor.java
@@ -16,7 +16,9 @@
  */
 package org.apache.camel.component.kafka.consumer.support.batching;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Correct fix. The old `commitManager.commit()` (no-arg) committed the Kafka consumer's **current position**, which can be ahead of what was actually delivered to the processor due to Kafka's internal pre-fetching. On restart, unprocessed records between the last batch and the committed position would be silently lost.

The fix correctly:
1. `computeBatchOffsets()` extracts max offset per `TopicPartition` from the actual exchange headers (`TOPIC`, `PARTITION`, `OFFSET`) using `offsets.merge(tp, offset, Math::max)` — only what was actually in the batch
2. `commitBatchOffsets()` does explicit `recordOffset()` + `commit(partition)` per partition — mirroring how the streaming facade commits

Both the `onComplete()` and `breakOnFirstError` paths in `CommitSynchronization` are updated. The `CommitSynchronization` constructor now captures the batch offsets at creation time, ensuring the offsets are fixed before the processor runs (correct snapshot semantics).

PR #24690, opened 2026-07-14
CAMEL-23995: Fix saslAuthType overriding explicit securityProtocol and generating non-working configs

davsclaus:
## Summary

Fix three bugs in the `saslAuthType` convenience layer (introduced in CAMEL-22864) that made 3 of 5 SASL auth types non-functional and silently overrode explicit user configuration:

- **Bug 1 -- SSL tautology**: `saslAuthType != KafkaAuthType.NONE` was always true inside the `isSasl()` branch, forcing `SASL_SSL` even when the user explicitly set `securityProtocol=SASL_PLAINTEXT`. Fixed by relying on the configurer's default (`useSsl=true`) and only downgrading when the user explicitly requests `SASL_PLAINTEXT`.
- **Bug 2 -- Kerberos always throws**: `buildJaasConfig()` called `validateKerberos()` (which throws on empty principal) *before* checking whether the principal was set, making `saslAuthType=KERBEROS` with external JAAS config impossible. Fixed by returning null early when principal is empty (with a debug log message), allowing external JAAS configuration.
- **Bug 3 -- Missing callback handlers**: OAuth and AWS MSK IAM set the login module in JAAS config but never set the required callback handler classes, causing authentication to fail at runtime. Fixed by adding `applySaslProperties()` (now backed by a shared `getSaslCallbackProperties()` helper) to set these as additional Kafka client properties with `putIfAbsent` semantics.

Also cleaned up the OAuth JAAS string: `oauth.token.endpoint.uri` is now set as a top-level Kafka property (where the Kafka client expects it) rather than embedded in the JAAS string, and the key name `oauth.scope` was corrected to `scope`.

## Review feedback addressed

- Fixed FQCN usage in `applySaslProperties()` parameter (now uses `Map` import)
- Extracted shared `getSaslCallbackProperties()` helper to eliminate code duplication between `applySaslProperties()` and `toProperties()`
- Added `LOG.debug` when Kerberos principal is not set (external JAAS assumption)
- Added upgrade guide entry for `saslAuthType` behavior changes

## Test plan

- [x] Unit tests for all auth types (PLAIN, SCRAM-SHA-256/512, OAuth, AWS MSK IAM, Kerberos, SSL, NONE)
- [x] Tests for JAAS config generation, security protocol selection, callback handler injection
- [x] Tests for putIfAbsent semantics (user-configured callback handler not overridden)
- [x] Tests for Kerberos without principal (external JAAS config)
- [x] Tests for toProperties() output
- [x] Full reactor build passes

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConfiguration.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConfiguration.java
@@ -449,15 +449,17 @@ public class KafkaConfiguration implements Cloneable, HeaderFilterStrategyAware 
                 // The saslAuthType just helps set the mechanism and protocol
             }
 
-            // Determine if we should use SSL (check if any SSL properties are set)
-            boolean hasSslConfig = ObjectHelper.isNotEmpty(sslTruststoreLocation)
-                    || ObjectHelper.isNotEmpty(sslKeystoreLocation)
-                    || sslContextParameters != null;
-
-            // For SASL types, default to SSL unless explicitly using SASL_PLAINTEXT
+            // For SASL types, the configurer defaults to useSsl=true (SASL_SSL).
+            // Only downgrade to SASL_PLAINTEXT if the user explicitly requested it.
             if (saslAuthType.isSasl()) {
-                boolean useSsl = !securityProtocol.equals("SASL_PLAINTEXT") && !securityProtocol.equals("PLAINTEXT");
-                configurer.withSsl(useSsl || hasSslConfig || saslAuthType != KafkaAuthType.NONE);
+                if (securityProtocol.equals("SASL_PLAINTEXT")) {
+                    configurer.withSsl(false);
+                }
+            } else if (saslAuthType == KafkaAuthType.NONE) {
+                boolean hasSslConfig = ObjectHelper.isNotEmpty(sslTruststoreLocation)
+                        || ObjectHelper.isNotEmpty(sslKeystoreLocation)
+                        || sslContextParameters != null;
+                configurer.withSsl(hasSslConfig || securityProtocol.equals("SSL"));
             }
 
             // Apply the configuration
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/security/KafkaSecurityConfigurer.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/security/KafkaSecurityConfigurer.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.kafka.security;
 
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 
@@ -23,6 +25,8 @@ import org.apache.camel.component.kafka.KafkaConfiguration;
 import org.apache.camel.util.ObjectHelper;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.common.config.SaslConfigs;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Builder for Kafka security configuration.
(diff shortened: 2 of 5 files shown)


gnodet (changes requested):
The PR correctly identifies and fixes three genuine bugs in the `saslAuthType` convenience layer (SSL tautology, Kerberos always throwing, missing callback handlers). The test coverage is thorough and CI passes. However, several issues need attention:

### 1. FQCN usage (medium)

The new method `applySaslProperties(java.util.Map<String, Object> additionalProperties)` at line 414 uses a fully-qualified class name instead of importing `java.util.Map`. This violates the project coding convention: "Do NOT use fully qualified class names (FQCNs) in Java code. Always add an import statement and use the simple class name." The CI build's OpenRewrite pass will likely flag this as well.

### 2. Code duplication (medium)

The callback handler configuration logic is duplicated between `applySaslProperties()` (lines 414-430, using `putIfAbsent`) and the inline switch in `toProperties()` (lines 451-465, using `setProperty`). These two code paths set the same constants for the same auth types but with different semantics. They must be kept in sync manually, and there's a subtle inconsistency: `putIfAbsent` allows user overrides while `setProperty` unconditionally overwrites. Consider refactoring `toProperties()` to reuse `applySaslProperties()` via a shared helper.

### 3. Missing upgrade guide entry (medium)

The PR changes user-visible behavior of a feature shipped since camel-4.18.0:
- OAuth JAAS string no longer includes `oauth.token.endpoint.uri` (moved to top-level Kafka config)
- OAuth JAAS key changed from `oauth.scope` to `scope`
- Kerberos without principal no longer throws `IllegalArgumentException`

Per project standards, any user-visible behavior change must be documented in `camel-4x-upgrade-guide-4_22.adoc`.

### 4. Silent Kerberos behavior change (low)

The early `return null` for Kerberos without principal silently changes behavior from a clear `IllegalArgumentException` to quiet success (no JAAS config set). While this is the intended design (support external JAAS configuration), a `LOG.debug()` or `LOG.info()` message like _"Kerberos principal not set; assuming external JAAS configuration"_ would improve observability.

davsclaus:
Thanks @gnodet for the thorough review! All four items have been addressed in the latest commit:

1. **FQCN usage** — Fixed. Added `import java.util.Map` and changed the parameter type to `Map<String, Object>`.

2. **Code duplication** — Fixed. Extracted a shared `getSaslCallbackProperties()` helper method that returns the callback handler properties as a `Map<String, String>`. Both `applySaslProperties()` (using `putIfAbsent` to preserve user overrides) and `toProperties()` (using `setProperty` on a fresh Properties) now delegate to this helper, eliminating the duplicated switch blocks.

3. **Missing upgrade guide entry** — Added. Documented the OAuth JAAS string format changes, the `putIfAbsent` semantics for explicit overrides, and the Kerberos external JAAS configuration support in `camel-4x-upgrade-guide-4_22.adoc`.

4. **Silent Kerberos behavior change** — Fixed. Added `LOG.debug("Kerberos principal not set; assuming external JAAS configuration (e.g. java.security.auth.login.config)")` before the early return.

gnodet (approved):
The new commit addresses all issues from the previous review:

1. **FQCN violations** — Fixed. `java.util.Map` replaced with simple `Map` import, along with proper imports for `LinkedHashMap`, `Logger`, and `LoggerFactory`.

2. **Missing upgrade guide** — Fixed. Clear entry added to `camel-4x-upgrade-guide-4_22.adoc` documenting all three behavioral changes: `putIfAbsent` semantics, OAuth JAAS string changes, and Kerberos without principal.

3. **Code duplication** (bonus) — Eliminated via shared `getSaslCallbackProperties()` method, correctly preserving the semantic difference (`putIfAbsent` in `applySaslProperties` vs `setProperty` in `toProperties`).

4. **Kerberos observability** — Added `LOG.debug` for the silent no-principal code path.

Test coverage expanded with seven new tests covering the key behavioral changes.

PR #24692, opened 2026-07-14
CAMEL-23996: Fix SingleNodeKafkaResumeStrategy correctness bugs

davsclaus:
## Summary

Fixes three correctness bugs in `SingleNodeKafkaResumeStrategy`:

- **Double subscribe replaces rebalance listener**: `refresh()` called `subscribe(consumer)` (which registers a `ConsumerRebalanceListener` via `checkAndSubscribe`), then immediately called `consumer.subscribe()` again without the listener, replacing the subscription. Removed the redundant second subscribe call.
- **`getAdapter()` never waits for initialization**: the condition `adapter == null` was never true because `setAdapter()` is always called before `loadCache()`. Changed to check `initLatch != null` instead, so callers block until cache loading completes.
- **`stop()` throws `IllegalMonitorStateException`**: `writeLock.tryLock()` can return false, but the `finally` block always called `writeLock.unlock()`. Added `isHeldByCurrentThread()` guard before unlocking.

Also adds `SingleNodeKafkaResumeStrategyTest` with unit tests for the fixed behaviors.

## Test plan

- [x] New `SingleNodeKafkaResumeStrategyTest` (3 tests) — all pass
- [x] All 149 existing `camel-kafka` tests pass with no regressions

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/SingleNodeKafkaResumeStrategy.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/processor/resume/kafka/SingleNodeKafkaResumeStrategy.java
@@ -209,8 +209,6 @@ public class SingleNodeKafkaResumeStrategy implements KafkaResumeStrategy, Camel
             subscribe(consumer);
 
             LOG.debug("Loading records from topic {}", resumeStrategyConfiguration.getTopic());
-            consumer.subscribe(Collections.singletonList(resumeStrategyConfiguration.getTopic()));
-
             poll(consumer, latch);
         } catch (WakeupException e) {
             LOG.info("Kafka consumer was interrupted during a blocking call");
@@ -358,7 +356,7 @@ public class SingleNodeKafkaResumeStrategy implements KafkaResumeStrategy, Camel
 
     @Override
     public ResumeAdapter getAdapter() {
-        if (adapter == null) {
+        if (initLatch != null) {
             waitForInitialization();
         }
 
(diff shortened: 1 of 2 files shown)


oscerd (approved):

gnodet (commented):
The three correctness fixes are valid and address real bugs:

1. **Double-subscribe removal** in `refresh()`: Correct — the second `consumer.subscribe()` without a `ConsumerRebalanceListener` was replacing the subscription registered by `subscribe(consumer)`. This was a bug introduced in CAMEL-18675.

2. **`initLatch` fix** in `getAdapter()`: Correct — the original `if (adapter == null)` was never true because `setAdapter()` is always called before `loadCache()`, meaning `waitForInitialization()` was never being called. The new `if (initLatch != null)` check correctly ensures initialization completes before returning the adapter.

3. **`isHeldByCurrentThread()` guard** in `stop()`: Correct — prevents `IllegalMonitorStateException` when `tryLock` returns false.

### Minor test cleanup issue (low)

In `testStopDoesNotThrowWhenLockNotAcquired`, the cleanup code (lines 49-51) spawns a "releaser" thread that calls `lock.unlock()`, but the lock is owned by the terminated "holder" thread. `ReentrantLock.unlock()` from a non-owning thread throws `IllegalMonitorStateException`, which is silently swallowed. The lock is never actually released. This doesn't affect test correctness (the assertion already passed), but the cleanup is misleading — it appears to work but fails silently. Simplest fix: remove the cleanup entirely since the lock is local and will be garbage collected.

PR #24695, opened 2026-07-14
CAMEL-24061: Wire OAuth credentials onto gRPC stub channel in camel-zeebe

davsclaus:
## Summary

- Fix authentication gap where 7 out of 9 `ZeebeService` operations bypassed OAuth because they used a raw `ManagedChannel` (always plaintext) instead of the authenticated `ZeebeClient`
- Wrap the `managedChannel` with a gRPC `ClientInterceptor` that applies `OAuthCredentialsProvider` credentials to every outgoing call when OAuth is configured
- When no credentials are configured, behavior is unchanged (plain channel, no interceptor)

**Affected operations (now authenticated):** `cancelProcessInstance`, `publishMessage`, `completeJob`, `failJob`, `updateJobRetries`, `throwError`, `deployResource`

**Already working:** `startProcess` and `registerJobHandler` (use `ZeebeClient` which was fixed in CAMEL-23974)

Note: `camel-zeebe` is deprecated since 4.19.0 in favor of `camel-camunda`, which does not have this issue (single `CamundaClient` for all operations). Users should migrate to `camel-camunda`.

## Test plan

- [x] Unit tests pass (`mvn test`)
- [ ] Verify with an OAuth-enabled Zeebe gateway that gRPC-stub operations now authenticate

### DIFF
--- a/components/camel-zeebe/src/main/java/org/apache/camel/component/zeebe/internal/ZeebeService.java
+++ b/components/camel-zeebe/src/main/java/org/apache/camel/component/zeebe/internal/ZeebeService.java
@@ -17,6 +17,7 @@
 
 package org.apache.camel.component.zeebe.internal;
 
+import java.io.IOException;
 import java.time.Duration;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
The PR correctly addresses the authentication gap for 7 gRPC-stub operations in the deprecated camel-zeebe component by adding an `OAuthInterceptor`. The approach is the standard gRPC pattern for adding per-call credentials. CI passes.

Two non-blocking observations for a deprecated component:

### 1. Token refresh gap (medium)

The `OAuthInterceptor` only calls `credentialsProvider.applyCredentials()` but does not implement `shouldRetryRequest()` logic. The native `ZeebeClient` (used by `startProcess` and `registerJobHandler`) handles `UNAUTHENTICATED` responses by calling `shouldRetryRequest()` which triggers token refresh and retries the call. The 7 gRPC-stub operations patched by this PR will **not** benefit from that retry mechanism — if the token expires mid-session, these operations will fail with `UNAUTHENTICATED` rather than transparently refreshing.

This is still a significant improvement over the prior state (no auth at all) and acceptable for a deprecated component, but worth documenting as a known limitation.

### 2. Pre-existing plaintext channel (medium, pre-existing)

The `managedChannel` always uses `usePlaintext()`, meaning OAuth bearer tokens for gRPC-stub operations are transmitted over an unencrypted channel. By contrast, when OAuth is configured, the `ZeebeClient` does **not** use `usePlaintext()`, getting TLS by default. This is a pre-existing issue not introduced by this PR, but now that credentials are being attached, the security implication is more concrete. Per Camunda's docs: "the Java client will not prevent you from adding credentials to gRPC calls while using an insecure connection, but you should be aware that doing so will expose your access token by transmitting it in plaintext."

Given the component is deprecated since 4.19.0, these are observations rather than blockers.

davsclaus:
Thanks for the review and the thorough analysis.

Both observations are valid and acknowledged as known limitations. Since camel-zeebe has been deprecated since 4.19.0, we won't be investing further in addressing these — the PR is a targeted fix to wire OAuth onto the gRPC stub channel, which is a significant improvement over having no auth at all on those 7 operations.

PR #24696, opened 2026-07-14
CAMEL-23997: Fix manual commit mode auto-committing offsets

davsclaus:
## Summary

When `allowManualCommit=true` was used together with a `DefaultKafkaManualCommitFactory` or
`DefaultKafkaManualAsyncCommitFactory`, the streaming processor recorded every processed offset via
`commitManager.recordOffset()` — even when the route did not call `KafkaManualCommit.commit()`. The
commit manager then flushed these offsets, defeating the purpose of manual commit and causing silent
message loss (at-most-once instead of at-least-once).

The fix guards the framework's auto-commit call (`commitManager.commit(partition)`) in
`KafkaRecordStreamingProcessorFacade.processPolledRecords()` with `!isAllowManualCommit()`.
The `recordOffset()` call (which only caches offsets internally without committing) is left
unconditional because `breakOnFirstError` and offset repository updates depend on the cached offsets.

When manual commit is enabled, only the route's explicit `manual.commit()` triggers a Kafka commit.

Part of [CAMEL-23997](https://issues.apache.org/jira/browse/CAMEL-23997) (finding #7).

## Changes

- `KafkaRecordStreamingProcessorFacade.processPolledRecords()`: guarded `commitManager.commit(partition)` with `!isAllowManualCommit()` — this is the auto-commit after all records in a partition are processed
- `KafkaRecordStreamingProcessor.processExchange()`: reverted the `recordOffset()` guard — offset caching must remain unconditional for `breakOnFirstError` and offset repository support
- Updated `KafkaRecordStreamingProcessorTest` to verify `recordOffset()` is always called
- `CommitManagersTest` with 6 tests covering all commit manager selection paths
- Upgrade guide entry documenting the behavior change

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/AbstractCommitManager.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/AbstractCommitManager.java
@@ -92,6 +92,11 @@ public abstract class AbstractCommitManager implements CommitManager {
         consumer.commitSync(
                 Collections.singletonMap(partition, new OffsetAndMetadata(partitionLastOffset + 1)),
                 Duration.ofMillis(timeout));
+
+        StateRepository<String, String> offsetRepository = configuration.getOffsetRepository();
+        if (offsetRepository != null) {
+            saveStateToOffsetRepository(partition, partitionLastOffset, offsetRepository);
+        }
     }
 
     protected void saveStateToOffsetRepository(
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/CommitManagers.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/consumer/CommitManagers.java
@@ -34,21 +34,12 @@ public final class CommitManagers {
         KafkaConfiguration configuration = kafkaConsumer.getEndpoint().getConfiguration();
 
         if (configuration.isAllowManualCommit()) {
-            LOG.debug("Allowing manual commit management");
-            KafkaManualCommitFactory manualCommitFactory = kafkaConsumer.getEndpoint().getKafkaManualCommitFactory();
-            if (manualCommitFactory instanceof DefaultKafkaManualAsyncCommitFactory) {
-                LOG.debug("Using an async commit manager for manual commit management");
-                return new AsyncCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-            } else {
-                if (manualCommitFactory instanceof DefaultKafkaManualCommitFactory) {
-                    LOG.debug("Using a sync commit manager for manual commit management");
-                    return new SyncCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-                } else {
-                    // This has been the default behavior for Camel
-                    LOG.debug("Using an NO-OP commit manager for manual commit management");
-                    return new NoopCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
-                }
-            }
+            // Manual commit mode: the framework must never auto-commit offsets.
+            // The configured KafkaManualCommitFactory (sync or async) controls how
+            // the route's explicit KafkaManualCommit.commit() call executes — that
+            // is independent of the commit manager selection.
+            LOG.debug("Using a NO-OP commit manager for manual commit management");
+            return new NoopCommitManager(consumer, kafkaConsumer, threadId, printableTopic);
         } else {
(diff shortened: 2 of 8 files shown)


oscerd (approved):

gnodet (commented):
The one-line production fix is correct and minimal — guarding `commitManager.recordOffset()` with `!configuration.isAllowManualCommit()` correctly prevents the streaming processor from silently recording offsets that would be auto-committed, defeating manual commit semantics.

The fix correctly handles the `breakOnFirstError` interaction: `commitManager.commit(topicPartition)` at line 128 becomes a harmless no-op since no offsets were recorded (both `SyncCommitManager` and `AsyncCommitManager` check the offset cache and return early if empty). The batching processor is unaffected as it already has separate manual/auto paths.

The upgrade guide entry is well-written.

### Test coverage gap (medium)

The new `CommitManagersTest` covers `CommitManagers.createCommitManager()` selection logic, which was **not modified** by this PR. No test directly validates the actual fix: that `KafkaRecordStreamingProcessor.processExchange()` does not call `commitManager.recordOffset()` when `allowManualCommit=true`.

A unit test mocking `CommitManager` and verifying `verify(commitManager, never()).recordOffset(any(), anyLong())` when `allowManualCommit=true` would directly prove the fix works and prevent regressions.

davsclaus:
@gnodet Good point on the test coverage gap. Added `KafkaRecordStreamingProcessorTest` in the streaming package (since the class is package-private) with two tests that directly validate the fix:

1. `manualCommitDoesNotRecordOffset` — verifies `commitManager.recordOffset()` is never called when `allowManualCommit=true`
2. `autoCommitRecordsOffset` — counter-test verifying `recordOffset()` is called with the correct partition and offset when `allowManualCommit=false`

gnodet (approved):
Thanks for addressing the feedback! The new `KafkaRecordStreamingProcessorTest` is exactly what was needed:

- **`manualCommitDoesNotRecordOffset()`** — verifies `verify(commitManager, never()).recordOffset(any(), anyLong())` when `allowManualCommit=true`, directly proving the fix prevents offset recording in manual commit mode
- **`autoCommitRecordsOffset()`** — verifies the positive path with exact offset value assertion `verify(commitManager).recordOffset(topicPartition, 42L)`

Combined with the existing `CommitManagersTest` for selection logic and the well-written upgrade guide entry, the test coverage is now comprehensive. The production fix, test coverage, and documentation are all solid.

gnodet (approved):
Good design refinement. Moving the manual-commit guard from `recordOffset()` to `commit(partition)` in the facade is more correct:

- **Previous approach**: guarded `recordOffset()` → prevented offset caching entirely, which would break `breakOnFirstError` combined with `allowManualCommit=true` (the error-recovery path at `processException()` line 128 needs cached offsets to commit the failing record's position)
- **New approach**: allows offset caching but guards the end-of-partition auto-flush → preserves error-handler recovery semantics while still preventing the unwanted auto-commit during normal processing

The unguarded `commitManager.commit()` calls in the error-handling paths (consumer listener `afterProcess` failure at line 113, and `processException`/`breakOnFirstError` at line 128) appear intentionally unguarded — they handle error recovery where committing progress before pausing/seeking is correct regardless of manual commit mode.

The test update (`manualCommitDoesNotRecordOffset` → `manualCommitStillRecordsOffset` with reversed assertion) correctly validates the new behavior.

gnodet (approved):
**Re-review after new commit** (`32b2018f` — "Fix state repository not updated in manual commit mode")

The 5th commit correctly fills two remaining gaps in the manual commit fix:

1. **`forceCommit()` state persistence** — `AbstractCommitManager.forceCommit()` now calls `saveStateToOffsetRepository()` after `commitSync()` succeeds. Correct ordering: we only persist state after Kafka has accepted the commit.

2. **`DefaultKafkaManualAsyncCommit.commit()` completion** — adding `commitManager.commit(getPartition())` after `recordOffset()` ensures that when a route explicitly calls `manual.commit()`, the offset is actually committed through the `AsyncCommitManager` (which calls `commitAsync()` on the Kafka consumer). Previously, the async manual commit only recorded the offset locally without ever triggering a Kafka commit — defeating the purpose of manual commit mode.

3. **Facade guard** — the `isAllowManualCommit()` check on the main auto-commit path (line 119) correctly prevents the framework from auto-committing after processing each partition when manual commit is enabled.

All 5 commits together tell a coherent story: fix commit manager selection → guard auto-commit paths → persist state on explicit commit. The upgrade guide entry clearly documents the behavior change. Well done. 👍

---

PR #24697, opened 2026-07-14
CAMEL-24062: Fix Multicast EIP to honor UseOriginalAggregationStrategy

davsclaus:
## Summary

- Move the `UseOriginalAggregationStrategy` binding from `Splitter` and `RecipientListProcessor` into `MulticastProcessor` (their shared base class), so all three EIPs consistently bind the original exchange on the strategy.
- Previously Multicast never called `newInstance(exchange)`, making the strategy silently ineffective — especially in error scenarios where the aggregated result could overwrite the original exchange body instead of preserving it.
- Remove the now-redundant duplicate binding blocks from `Splitter` and `RecipientListProcessor` (Splitter's null-strategy fallback for its default behavior is preserved).

## Test plan

- [x] Added body-preservation assertion to `MulticastUseOriginalPropagateExceptionCaughtTest`
- [x] All 18 `UseOriginal*` tests pass across Multicast, Splitter, and RecipientList
- [x] Full `camel-core` test suite passes (7264 tests, 0 failures)

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -52,6 +52,8 @@ import org.apache.camel.Route;
 import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.StreamCache;
 import org.apache.camel.Traceable;
+import org.apache.camel.processor.aggregate.ShareUnitOfWorkAggregationStrategy;
+import org.apache.camel.processor.aggregate.UseOriginalAggregationStrategy;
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean, well-scoped refactoring that consolidates `UseOriginalAggregationStrategy` binding from `Splitter` and `RecipientListProcessor` into their shared base class `MulticastProcessor`, fixing the bug where Multicast silently ignored the strategy.

The code correctly handles:
- Nested multicast/splitter scenarios (CAMEL-3333) via per-processor-instance exchange property keys
- Concurrency (CAMEL-12709) via `newInstance()` creating per-exchange strategy clones
- The existing null-strategy fallback in Splitter (mutually exclusive with UseOriginal)

Tests and upgrade guide entry are included. Nice work.

PR #24705, opened 2026-07-15
CAMEL-24064: Fix module registration in jackson3 type converter and wrap JacksonException in data type transformers

Croway:
Fixes two regressions in `camel-jackson3` from the Jackson 2 → Jackson 3 port (CAMEL-22857):

1. **Type converter drops all but the last configured Jackson module** — `JacksonTypeConverters.resolveObjectMapper()` replaced the mapper with a fresh `JsonMapper.builder().addModule(module).build()` on every loop iteration over `CamelJacksonTypeConverterModuleClassNames`, so only the last module was registered. Now all modules are accumulated on a single builder (matching the Jackson 2 behavior of `registerModule`).

2. **Jackson exceptions escape data type transformers unwrapped** — `JsonStructDataTypeTransformer` and `JsonPojoDataTypeTransformer` still caught the checked `IOException` that Jackson 3 read/write calls no longer throw, so malformed JSON escaped as a raw unchecked `JacksonException` instead of being wrapped in `CamelExecutionException` with the exchange attached (as `JsonDataTypeTransformer` already does). Added `JacksonException` to the catch clauses.

Each new test was verified to fail against the unfixed code and pass with the fix; the full `camel-jackson3` suite passes (103 tests).

Note: `camel-jackson3` was introduced in 4.19.0, so no backport to `camel-4.18.x` is applicable.

JIRA: https://issues.apache.org/jira/browse/CAMEL-24064

### DIFF
--- a/components/camel-jackson3/src/main/java/org/apache/camel/component/jackson3/converter/JacksonTypeConverters.java
+++ b/components/camel-jackson3/src/main/java/org/apache/camel/component/jackson3/converter/JacksonTypeConverters.java
@@ -305,7 +305,7 @@ public final class JacksonTypeConverters {
             lock.lock();
             try {
                 if (defaultMapper == null) {
-                    ObjectMapper mapper = new ObjectMapper();
+                    JsonMapper.Builder builder = JsonMapper.builder();
                     if (moduleClassNames != null) {
                         for (Object o : ObjectHelper.createIterable(moduleClassNames)) {
                             Class<JacksonModule> type
@@ -313,11 +313,11 @@ public final class JacksonTypeConverters {
                             JacksonModule module = camelContext.getInjector().newInstance(type);
 
                             LOG.debug("Registering module: {} -> {}", o, module);
-                            mapper = JsonMapper.builder().addModule(module).build();
+                            builder.addModule(module);
                         }
                     }
 
-                    defaultMapper = mapper;
+                    defaultMapper = builder.build();
                 }
             } finally {
                 lock.unlock();
--- a/components/camel-jackson3/src/main/java/org/apache/camel/component/jackson3/transform/JsonPojoDataTypeTransformer.java
+++ b/components/camel-jackson3/src/main/java/org/apache/camel/component/jackson3/transform/JsonPojoDataTypeTransformer.java
@@ -35,6 +35,7 @@ import org.apache.camel.spi.MimeType;
 import org.apache.camel.spi.Transformer;
 import org.apache.camel.util.ObjectHelper;
 import tools.jackson.core.FormatSchema;
+import tools.jackson.core.JacksonException;
 
 /**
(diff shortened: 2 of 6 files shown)


gnodet (approved):
Clean, well-targeted fix for two genuine regressions in the Jackson 2 → Jackson 3 port.

**Module registration fix** (`JacksonTypeConverters.java`): The broken code called `JsonMapper.builder().addModule(module).build()` inside the loop, creating a fresh mapper per iteration and discarding previously registered modules. The fix correctly accumulates modules on a single `JsonMapper.Builder` instance before calling `build()` once — matches the Jackson 3 idiomatic pattern.

**JacksonException wrapping** (`JsonPojoDataTypeTransformer` / `JsonStructDataTypeTransformer`): In Jackson 3, `JacksonException` extends `RuntimeException` (not `IOException` as in Jackson 2), so the existing `IOException` catch clauses missed Jackson parse errors. The added `JacksonException` catches are correct and consistent with how `JsonDataTypeTransformer` (the third transformer in the same package) already handles it.

**Test coverage** is solid — the multi-module test verifies both custom modules are active, and the malformed-JSON tests confirm `CamelExecutionException` wrapping.

Minor observation (non-blocking): the `IOException` catch in `JsonPojoDataTypeTransformer` is now only reachable via `getBodyAsStream()` I/O errors, not Jackson parse errors. This is technically correct but could be clarified with a comment in a future cleanup.

davsclaus (approved):

oscerd (commented):
Thanks Federico — both fixes look correct, and I like that each is backed by a test that fails before / passes after.

**Fix 1 — module registration (`JacksonTypeConverters`):** good catch. The old loop rebuilt the mapper on every iteration (`mapper = JsonMapper.builder().addModule(module).build();`), so only the *last* discovered module actually survived. Hoisting a single `JsonMapper.Builder` and calling `build()` once restores the Jackson-2 `registerModule` semantics, and `JacksonConversionsMultipleModulesTest` (Foo + Bar) is exactly the test that would have caught the regression.

**Fix 2 — `JacksonException` in the transformers:** correct. In Jackson 3, `tools.jackson.core.JacksonException extends RuntimeException` (vs Jackson 2's `JsonProcessingException extends IOException`), so malformed JSON was slipping past the checked-only catch. Adding it to the multi-catch and wrapping in `CamelExecutionException(..., exchange, e)` preserves the original cause and matches the sibling `JsonDataTypeTransformer`.

One trivial, non-blocking nit: `JsonPojoDataTypeTransformer` keeps `IOException` in its catch list while `JsonStructDataTypeTransformer` doesn't — for the Jackson read path it's now effectively dead (only kept legal by `getJavaObject`'s `throws IOException`). Harmless; could drop it for symmetry, but not worth a revision on its own.

A couple of CI checks are still running as I write this — worth letting it go fully green before merge, but the code itself looks good.

---

PR #24710, opened 2026-07-15
CAMEL-24048: camel-sql - Fix stale remove, idempotent race, and lexer error handling

davsclaus:
## Summary

Fixes three correctness defects in camel-sql reported in [CAMEL-24048](https://issues.apache.org/jira/browse/CAMEL-24048):

- **Stale `remove()` in `JdbcAggregationRepository` / `ClusteredJdbcAggregationRepository`**: `remove()` deleted with `WHERE id = ? AND version = ?` but discarded the update count. A stale version (0 rows deleted) would still insert into the completed table and report success, leading to duplicate downstream delivery. Now checks the delete count and throws `OptimisticLockingException` on mismatch, symmetric with `updateHelper()`.

- **`DuplicateKeyException` race in `AbstractJdbcMessageIdRepository.add()`**: The check-then-act pattern (`queryForInt` then `insert`) allows a cross-node race where the `INSERT` hits the primary-key constraint. The `DuplicateKeyException` now returns `false` (key already exists) instead of propagating as a raw exception that fails the exchange. Calls `status.setRollbackOnly()` to ensure a clean rollback on databases like PostgreSQL where a constraint violation marks the transaction as failed. Includes debug-level logging for diagnostics.

- **`TokenMgrError` escaping `TemplateParser`**: `parseTemplate()` only caught `ParseException` but the JavaCC lexer throws `TokenMgrError` (a `java.lang.Error`) for characters outside the token alphabet. Now caught and wrapped in `ParseRuntimeException`, consistent with `ParseException` handling.

## Test plan

- [x] `JdbcAggregationRepositoryStaleRemoveTest` — verifies stale remove throws `OptimisticLockingException`
- [x] `JdbcMessageIdRepositoryDuplicateInsertRaceTest` — simulates concurrent insert race deterministically via `queryForInt()` override
- [x] `TemplateParserLexicalErrorTest` — verifies semicolon and backtick produce `ParseRuntimeException` not `Error`
- [x] Full camel-sql test suite passes (239 tests, 0 failures)

### DIFF
--- a/components/camel-sql/src/main/java/org/apache/camel/component/sql/stored/template/TemplateParser.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/component/sql/stored/template/TemplateParser.java
@@ -22,6 +22,7 @@ import org.apache.camel.component.sql.stored.template.ast.ParseRuntimeException;
 import org.apache.camel.component.sql.stored.template.ast.Template;
 import org.apache.camel.component.sql.stored.template.generated.ParseException;
 import org.apache.camel.component.sql.stored.template.generated.SSPTParser;
+import org.apache.camel.component.sql.stored.template.generated.TokenMgrError;
 import org.apache.camel.spi.ClassResolver;
 import org.apache.camel.util.ObjectHelper;
 
@@ -41,6 +42,8 @@ public class TemplateParser {
 
         } catch (ParseException parseException) {
             throw new ParseRuntimeException(parseException);
+        } catch (TokenMgrError tokenMgrError) {
+            throw new ParseRuntimeException(tokenMgrError);
         }
     }
 
--- a/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/ClusteredJdbcAggregationRepository.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/ClusteredJdbcAggregationRepository.java
@@ -77,11 +77,17 @@ public class ClusteredJdbcAggregationRepository extends JdbcAggregationRepositor
                     LOG.debug("Removing key {}", correlationId);
                     String table = getRepositoryName();
                     verifyTableName(table);
-                    jdbcTemplate.update("DELETE FROM " + table + " WHERE " + ID + " = ? AND " + VERSION + " = ?", // NOSONAR
+                    int deleteCount = jdbcTemplate.update(
+                            "DELETE FROM " + table + " WHERE " + ID + " = ? AND " + VERSION + " = ?", // NOSONAR
                             correlationId, version);
+                    if (deleteCount != 1) {
+                        throw new OptimisticLockingException();
+                    }
 
                     insert(camelContext, confirmKey, exchange, getRepositoryNameCompleted(), version, true);
 
+                } catch (OptimisticLockingException e) {
+                    throw e;
                 } catch (Exception e) {
                     throw new RuntimeException(
                             "Error removing key " + correlationId + " from repository " + getRepositoryName(), e);
--- a/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/JdbcAggregationRepository.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/processor/aggregate/jdbc/JdbcAggregationRepository.java
@@ -438,12 +438,18 @@ public class JdbcAggregationRepository extends ServiceSupport
                     LOG.debug("Removing key {}", correlationId);
                     String table = getRepositoryName();
                     verifyTableName(table);
-                    jdbcTemplate.update("DELETE FROM " + table + " WHERE " + ID + " = ? AND " + VERSION + " = ?", // NOSONAR
+                    int deleteCount = jdbcTemplate.update(
+                            "DELETE FROM " + table + " WHERE " + ID + " = ? AND " + VERSION + " = ?", // NOSONAR
                             correlationId, version);
+                    if (deleteCount != 1) {
+                        throw new OptimisticLockingException();
+                    }
 
                     insert(camelContext, confirmKey, exchange, getRepositoryNameCompleted(), version);
                     LOG.debug("Removed key {}", correlationId);
 
+                } catch (OptimisticLockingException e) {
+                    throw e;
                 } catch (Exception e) {
                     throw new RuntimeException("Error removing key " + correlationId + " from repository " + repositoryName, e);
                 }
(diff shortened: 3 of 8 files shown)


gnodet (approved):
Three well-targeted correctness fixes for `camel-sql`, each addressing a real defect:

**1. Stale remove** (`JdbcAggregationRepository` / `ClusteredJdbcAggregationRepository`): Adds a version check to the `DELETE` statement and throws `OptimisticLockingException` on mismatch. This is symmetric with the existing `updateHelper()` pattern (which already throws OLE on stale version). The `catch (OptimisticLockingException e) { throw e; }` block correctly prevents the generic `catch (Exception e)` from wrapping the OLE in a plain `RuntimeException`, which would make it invisible to the aggregation strategy's retry logic.

**2. Idempotent race** (`AbstractJdbcMessageIdRepository`): Correctly handles the TOCTOU window between `queryForInt()` returning 0 and `insert()` hitting the PK constraint. Catching `DuplicateKeyException` and returning `false` (key already exists = duplicate) is the right semantic. All operations happen inside `TransactionTemplate.execute()` callbacks, so thread safety is maintained.

**3. Lexer error** (`TemplateParser`): `TokenMgrError` is a `java.lang.Error` subclass that would propagate uncaught and crash the route/thread. Wrapping it in `ParseRuntimeException` ensures it's handled the same way as `ParseException`.

No SQL injection concerns: `verifyTableName()` enforces `[a-zA-Z_][a-zA-Z0-9_]*`, and all values use parameterized `?` placeholders. Test coverage is solid for all three fixes.

Minor non-blocking note: the `DuplicateKeyException` catch is silent (no logging). A `log.debug()` would help operators diagnose concurrent idempotent-check behavior in multi-node deployments.

oscerd (commented):
Nice set of fixes, Claus — all three are real, and I like that each ships a deterministic test that simulates the window directly rather than leaning on timing.

What I checked:
- **Lexer** (`TemplateParser`): catching `TokenMgrError` (a JavaCC `java.lang.Error`, so it slipped past the existing `catch (ParseException)`) and rewrapping as `ParseRuntimeException` is correct; the template is route-author-supplied, so no trust-boundary angle.
- **Stale remove** (`JdbcAggregationRepository` + clustered): the versioned `DELETE ... WHERE ID=? AND VERSION=?` with `deleteCount != 1 → OptimisticLockingException`, rethrown ahead of the generic catch, closes the race at the DB level and lets the aggregator's optimistic-locking retry kick in. Good.
- **Idempotent add** (`AbstractJdbcMessageIdRepository`): catching `DuplicateKeyException` and returning `false` gives the correct "already present" semantics for the thread that loses the race.

One question on the idempotent fix — not blocking: the `catch (DuplicateKeyException e)` sits **inside** the `doInTransaction(...)` callback, and returning `Boolean.FALSE` lets the `TransactionTemplate` go on to commit. On H2 (the test DB) that's fine, but on PostgreSQL a constraint violation aborts the whole transaction, so the subsequent commit surfaces as `UnexpectedRollbackException` rather than a clean `false` — and a few other vendors behave similarly. Have you been able to confirm the behavior against Postgres/Oracle? If it's a concern there, a savepoint around the `insert` (or an `INSERT ... ON CONFLICT`-style path) would keep the losing thread returning a clean `false` on every database camel-sql targets, not just H2.

Minor: since a stale aggregation `remove` now throws `OptimisticLockingException` and malformed stored-procedure templates now throw `ParseRuntimeException`, a one-line upgrade-guide note might help anyone who caught or subclassed the old types — optional.

Otherwise this looks good.

---

davsclaus:
@gnodet @oscerd Thanks for the reviews! Pushed a follow-up commit addressing both points:

1. **PostgreSQL transaction safety** (@oscerd): Added `status.setRollbackOnly()` before returning `Boolean.FALSE` in the `DuplicateKeyException` catch. This ensures a clean rollback on databases like PostgreSQL where a constraint violation marks the transaction as failed — without it, the `TransactionTemplate` would attempt to commit the failed transaction and throw `UnexpectedRollbackException`.

2. **Debug logging** (@gnodet): Added `log.debug()` for the duplicate-key race so operators can see the concurrent idempotent-check behavior in multi-node deployments.

3. **Upgrade guide** (@oscerd): Added an entry documenting the new `OptimisticLockingException` from stale remove and `ParseRuntimeException` from malformed templates.

gnodet (approved):
Re-reviewed after new commit `aef8cdf9` addressing review feedback. APPROVE maintained.

**New changes are correct:**

- **`status.setRollbackOnly()`** in the `DuplicateKeyException` catch block — correctly handles PostgreSQL's transaction-abort behavior. When PostgreSQL detects a constraint violation, it marks the transaction as aborted and rejects all subsequent statements until ROLLBACK. Calling `setRollbackOnly()` tells Spring's `TransactionTemplate` to roll back cleanly instead of attempting a commit (which would otherwise surface as `UnexpectedRollbackException`). The `return Boolean.FALSE` value is captured before the rollback executes, so the return path is correct.

- **`log.debug()`** addition — clear diagnostic message with the key value, helpful for operators diagnosing concurrent idempotent-check behavior in multi-node deployments.

- **Upgrade guide entry** — well-written, documents both the new `OptimisticLockingException` from stale removes and the `ParseRuntimeException` wrapping of `TokenMgrError`.

All three original fixes remain correct and well-tested after the rebase.

oscerd (approved):
LGTM — I went through all three fixes and each restores the intended contract:

- **Stale remove:** `remove()` now checks the `DELETE ... WHERE id=? AND version=?` update count and throws `OptimisticLockingException` when it isn't exactly 1, with the `catch (OptimisticLockingException) { throw; }` correctly ordered before the generic `catch` (it's a `RuntimeException`, so it would otherwise be wrapped and hidden). No ABA risk since `version` is monotonic, and it matches the existing `updateHelper()` symmetry. Applied to both the plain and clustered repositories.
- **Idempotent race:** atomicity was already guaranteed by the `PRIMARY KEY (processorName, messageId)` constraint — the check-then-act was never what enforced it. The fix catches the loser's `DuplicateKeyException` and returns `FALSE` ("already exists"), converting a spurious propagated `DataAccessException` into the correct duplicate semantic. No dedup bypass in either direction.
- **Lexer:** `TokenMgrError` (a JavaCC `Error`, so it slipped past the old `catch (ParseException)`) is now caught and wrapped in `ParseRuntimeException` — surfaced, not swallowed.

Tests cover all three deterministically (no `Thread.sleep`), and the behavior changes are documented in the 4.22 upgrade guide. Nice.

One non-blocking nit: `ClusteredJdbcAggregationRepository.remove()` gets the same stale-remove fix but no dedicated test — it's line-identical to the covered `JdbcAggregationRepository` path, so low risk, just noting it.

PR #24711, opened 2026-07-15
CAMEL-24065: camel-spring - Fix ConcurrentModificationException in EventComponent

davsclaus:
## Summary

Fixes [CAMEL-24065](https://issues.apache.org/jira/browse/CAMEL-24065).

`EventComponent.endpoints` was a plain `LinkedHashSet` iterated by `onApplicationEvent()` and mutated by `consumerStarted()`/`consumerStopped()` without synchronization, causing `ConcurrentModificationException` when a `spring-event` route is stopped while events are dispatched.

- Replace `LinkedHashSet` with `CopyOnWriteArraySet` — ideal for this write-rare, read-frequent pattern
- Fix premature endpoint removal in `EventEndpoint.consumerStopped()` — only remove the endpoint from the component's set when the load balancer has no processors left, preventing a surviving consumer from silently losing events

## Test plan

- [x] New `EventComponentConcurrencyTest.stoppingRouteDuringEventDispatchMustNotThrow` — verifies no exception when stopping a route during event dispatch (the reported bug)
- [x] New `EventComponentConcurrencyTest.survivingEndpointStillReceivesEventsAfterAnotherIsStopped` — verifies that remaining endpoints still receive events after another is stopped
- [x] Existing `EventRouteTest` passes

### DIFF
--- a/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/component/event/EventComponentConcurrencyTest.java
+++ b/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/component/event/EventComponentConcurrencyTest.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.event;
+
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-targeted fix for a real concurrency bug in `EventComponent`.

**CopyOnWriteArraySet** is the ideal choice for this access pattern — event dispatch iterates the endpoints set frequently (read path), while route lifecycle operations mutate it rarely (write path). `CopyOnWriteArraySet` gives iterators a consistent snapshot without synchronization overhead on reads.

**consumerStopped reordering** correctly fixes a secondary bug: the original code unconditionally removed the endpoint from the component when any consumer stopped, even if other consumers on the same endpoint (via the load balancer) were still active. The fix first removes the processor from the load balancer, then only removes the endpoint from the component when the load balancer is empty.

**Tests** are well-designed:
- `stoppingRouteDuringEventDispatchMustNotThrow` directly reproduces the CME (one route's processor stops another route during event dispatch)
- `survivingEndpointStillReceivesEventsAfterAnotherIsStopped` verifies the secondary fix

Minor non-blocking observation: `consumerStarted` unconditionally calls `getComponent().consumerStarted(this)` on every consumer start, which is a no-op for `CopyOnWriteArraySet` duplicates but could be guarded with an emptiness check for symmetry with `consumerStopped`.

oscerd (commented):
The `CopyOnWriteArraySet` swap is the right call here. The CME came from `EventComponent.onApplicationEvent` iterating `endpoints` while a concurrently-stopping route mutated the same set (`consumerStopped` → `endpoints.remove`); COW gives snapshot iteration, so the race is closed rather than just narrowed, and the write-rare / iterate-often profile of that registry is exactly what it suits. `stoppingRouteDuringEventDispatchMustNotThrow` reproduces it deterministically (no `Thread.sleep`). Looks good.

---

PR #24712, opened 2026-07-15
CAMEL-24066: Fix SpringScheduledPollConsumerScheduler ignoring startScheduler=false

davsclaus:
## Summary

- Fix `SpringScheduledPollConsumerScheduler` violating the `ScheduledPollConsumerScheduler` SPI contract by scheduling the cron task in `doStart()` instead of `startScheduler()`
- Move `taskScheduler.schedule(runnable, trigger)` from `doStart()` to `startScheduler()` with an idempotent guard, aligning with `DefaultScheduledPollConsumerScheduler`
- Fix `isSchedulerStarted()` to check actual scheduling state (`future != null && !future.isCancelled()`) instead of executor liveness

## What was wrong

The Spring scheduler implementation scheduled the cron trigger unconditionally inside `doStart()`, which is called via `ServiceHelper.startService(scheduler)` **before** the `isStartScheduler()` gate in `ScheduledPollConsumer.doStart()`. This meant `startScheduler=false` was silently ignored — the cron started firing as soon as the route started.

This affected file, FTP, SFTP, SMB consumers and `GenericFilePollingConsumer` (`pollEnrich()`) when combined with `scheduler=spring`, as they set `startScheduler=false` internally.

## Test plan

- [x] New test verifies `startScheduler=false` prevents polling with `scheduler=spring`
- [x] Existing `FileConsumerSpringSchedulerTest` still passes (normal scheduling works)

Follow-up: [CAMEL-24079](https://issues.apache.org/jira/browse/CAMEL-24079) — Quartz scheduler has the same structural pattern.

### DIFF
--- a/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/spring/pollingconsumer/SpringScheduledPollConsumerSchedulerStartSchedulerTest.java
+++ b/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/spring/pollingconsumer/SpringScheduledPollConsumerSchedulerStartSchedulerTest.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.spring.pollingconsumer;
+
+import java.util.concurrent.TimeUnit;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.ContextTestSupport;
+import org.apache.camel.Exchange;
+import org.junit.jupiter.api.Test;
+
+import static org.apache.camel.spring.processor.SpringTestHelper.createSpringCamelContext;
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-scoped fix that correctly aligns `SpringScheduledPollConsumerScheduler` with the `ScheduledPollConsumerScheduler` SPI contract. 

**Root cause**: `doStart()` is called via `ServiceHelper.startService(scheduler)` at line 680 of `ScheduledPollConsumer`, which runs *before* the `isStartScheduler()` gate at line 682. By moving the `taskScheduler.schedule(runnable, trigger)` call from `doStart()` to `startScheduler()`, the scheduling now properly respects the `startScheduler=false` flag.

**Key observations:**

- The separation of "prepare infrastructure" (`doStart`) from "begin scheduling" (`startScheduler`) now mirrors `DefaultScheduledPollConsumerScheduler` cleanly — `doStart()` still initializes `trigger` and `taskScheduler`, so dependencies are in place when `startScheduler()` is later called.

- The `isSchedulerStarted()` change from checking `taskScheduler.getScheduledExecutor().isShutdown()` to `future != null && !future.isCancelled()` is semantically correct. The old check answered "is the executor pool alive?" (true as soon as `doStart()` creates it, even with no task scheduled). The new check answers "is a cron task actually scheduled?" — matching `DefaultScheduledPollConsumerScheduler`'s `return !futures.isEmpty()` pattern.

- The `future` field is `volatile`, providing proper visibility for `isSchedulerStarted()` and the idempotency guard in `startScheduler()`.

- Good test — the `during(3, SECONDS)` Awaitility assertion correctly proves the negative case (scheduler doesn't start when `startScheduler=false`).

👍

---

oscerd (approved):
LGTM — the Spring sibling of #24715, and consistent with it. Scheduling now happens in `startScheduler()` behind `if (future == null)`, and `isSchedulerStarted()` returns `future != null && !future.isCancelled()` instead of the old executor-liveness check (which was true immediately after `doStart()` even with nothing scheduled). So `startScheduler=false` no longer polls, while the default path is unchanged. `doStop()` cancels/nulls the `volatile future` for a clean restart, and the test proves the flag is now honored.

Same unrelated `detect-changes` infra failure as the rest of the batch (no merge base on a shallow fetch) — not this change; #24718 addresses it.

PR #24714, opened 2026-07-15
CAMEL-24067: Fix SpringInjector not injecting CamelContext into CamelContextAware beans via factory method

davsclaus:
## Summary

- `SpringInjector.newInstance(type, factoryClass, factoryMethod)` was missing the `CamelContextAware.trySetCamelContext()` call that `DefaultInjector` has, so `CamelContextAware` beans created through `#class:...#factoryMethod` syntax did not get the `CamelContext` injected when running in Spring/Spring Boot
- Added a `CamelContext` field to `SpringInjector` and set it from `SpringCamelContext.createInjector()`

## Test plan

- [x] New `SpringInjectorFactoryMethodTest` with two tests verifying CamelContext injection via both `newInstance(type, factoryMethod)` and `newInstance(type, factoryClass, factoryMethod)`
- [ ] Verify no regressions in existing camel-spring-xml tests

### DIFF
--- a/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/spring/spi/SpringInjectorFactoryMethodTest.java
+++ b/components/camel-spring-parent/camel-spring-xml/src/test/java/org/apache/camel/spring/spi/SpringInjectorFactoryMethodTest.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.spring.spi;
+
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-targeted fix that aligns `SpringInjector`'s factory method path with `DefaultInjector` by adding `CamelContextAware` injection.

**Analysis:**

The bug existed because the factory method code path in `SpringInjector` bypasses Spring's `createBean()` lifecycle (and thus `CamelBeanPostProcessor`), so `CamelContextAware` beans created via factory methods never received their `CamelContext`. The fix correctly places `CamelContextAware.trySetCamelContext()` in the same position as `DefaultInjector` does — after the factory method invoke, inside the try block.

**Key observations:**

- The `newInstance(type, postProcessBean)` path does **not** need this fix because it delegates to Spring's `createBean()`, which runs the full `BeanPostProcessor` chain including `CamelBeanPostProcessor` — that post-processor already handles `CamelContextAware` injection.
- `trySetCamelContext()` is null-safe for both arguments (guards on `camelContext != null` and `object instanceof CamelContextAware`), so passing a null answer is harmless.
- The `instanceof` pattern matching modernization in `SpringCamelContext.createInjector()` is a welcome cleanup.
- Both test methods exercise the two factory method overloads (with and without explicit `factoryClass`), providing good coverage.

oscerd (approved):
LGTM. The fix places `CamelContextAware.trySetCamelContext(answer, camelContext)` on the factory-method path exactly where `DefaultInjector` already does it, so `SpringInjector` now behaves consistently for factory-method-created beans. It's null-safe (`trySetCamelContext` guards on both `camelContext != null` and `instanceof CamelContextAware`), and there's no double-injection risk since the direct path still goes through Spring's `CamelBeanPostProcessor`. Both factory-method overloads are covered by `SpringInjectorFactoryMethodTest` (no `Thread.sleep`).

The red CI is unrelated: the `detect-changes` job fails with `fatal: origin/main...HEAD: no merge base` — a shallow-fetch (`fetch-depth: 50`) infra limitation, not this change. #24718 fixes exactly that.

PR #24644, opened 2026-07-13
CAMEL-23847: Camel-PQC - stream sign/verify for large payloads

oscerd:
# CAMEL-23847: Camel-PQC — stream sign/verify for large payloads

## Motivation

`PQCProducer` sign/verify materialised the entire body as a String
(`getMandatoryBody(String.class)`) before feeding it to `Signature.update()`. For large payloads
(files, large messages) this loaded everything into memory, and for non-UTF-8 binary payloads the
String round-trip corrupted the data, so verification failed.

## What this does

`signature()` and `verification()` now feed the signature directly from the message body via an
`updateSignatureFromBody(Signature, Message)` helper:

- a **`byte[]`** body is used as-is;
- an **`InputStream`** body is read in **chunks** (8 KB buffer) — never fully materialised — and is
  **reset** afterwards when it is a re-readable `StreamCache`, so downstream processors still see
  the payload;
- any other body type falls back to its String representation (String **charset** handling is the
  sibling bug **CAMEL-23843**).

The **hybrid** sign/verify variants obtain the payload as a `byte[]` (via `bodyToByteArray(Message)`)
without a String round-trip (`HybridSignature` needs the full array in memory).

## Testing

`PQCStreamingSignatureTest` (5 tests, all green):
- `byte[]` sign/verify round trip;
- sign an `InputStream`, verify the same bytes;
- sign bytes, verify an `InputStream`;
- **binary** (non-UTF-8) payloads sign and verify correctly — this fails with the old String
  materialisation;
- a **1 MB** stream signs and verifies within one route (stream caching → `StreamCache` reset).

## Documentation

`pqc-component.adoc` gains a note on payload streaming under *Signature and Verification*.

## Backport

None — improvement, `main` only (4.22.0). The related String-charset correctness fix is tracked
separately as **CAMEL-23843** (a bug, which is the backport candidate).

---
_Claude Code on behalf of Andrea Cosentino._

PR #24716, opened 2026-07-15
CAMEL-23843: Camel-PQC - sign and verify String payloads as UTF-8

oscerd:
# CAMEL-23843: Camel-PQC — sign and verify String payloads as UTF-8

## The bug

`PQCProducer` sign/verify (and the hybrid variants) read the body and, when it is not already binary,
encode it with `String.getBytes()` — i.e. the **JVM default charset**. The default charset is platform
dependent, so signing on one JVM and verifying on another with a different default disagrees on the
bytes of a **non-ASCII** payload and verification fails.

The issue also called out that binary payloads were forced through a String conversion. **That half is
already fixed on `main`** by CAMEL-23847 (#24644, merged): `byte[]` and `InputStream` bodies are now fed
to the signature byte-for-byte. What remained is the charset.

## The fix

Pin `StandardCharsets.UTF_8` in both body-to-bytes fallbacks (`updateSignatureFromBody`, used by
sign/verify, and `bodyToByteArray`, used by the hybrid variants), so sign and verify always agree on
the encoding. Binary bodies are untouched.

## Scope / honesty about impact

On **Java 18+** the default charset is already UTF-8 (JEP 400), so this is a **no-op** there — the value
of the change is that the encoding becomes *explicit and deterministic* rather than
environment-dependent. On **Java 17** with a non-UTF-8 default (which Camel still supports) the bug is
live, and this fixes it.

Because it can change the signature of a non-ASCII `String` payload on such a JVM, it is documented in
the **4.22 upgrade guide**.

## Testing

New `PQCSignatureCharsetTest` (2 tests):
- a `String` body and its UTF-8 bytes are **interchangeable** — signing the String verifies against the
  UTF-8 bytes and vice versa;
- a **negative** check: a signature over the String must *not* verify against the same text encoded as
  ISO-8859-1.

The existing `PQCStreamingSignatureTest` (5) still passes. Module build green, no generated drift.

## Backport

**4.18.x (4.18.4)** and **4.14.x (4.14.9)** — **charset only**. Those branches never received the
streaming/binary-payload work (CAMEL-23847, an improvement kept to `main`), so the backport pins UTF-8
for `String` bodies but does **not** add native `byte[]`/`InputStream` handling: that remains the
main-only improvement. The determinism bug for `String` payloads is fixed on all three lines.

---


davsclaus (approved):
Minimal, well-scoped fix for the last remaining piece after CAMEL-23847 handled binary bodies. Both String-to-bytes conversion points (`updateSignatureFromBody` and `bodyToByteArray`) are pinned to UTF-8, so sign and verify always agree regardless of the JVM default charset.

The test covers both directions (String ↔ UTF-8 bytes interchangeable, String ↛ ISO-8859-1 bytes) with non-ASCII text where the encodings genuinely differ. Upgrade guide entry is honest about impact (no-op on Java 18+, live fix on Java 17 with non-UTF-8 default). Component docs updated.

LGTM.

gnodet (approved):
Clean fix that pins UTF-8 encoding in both String-to-bytes fallback paths for PQC sign/verify operations. Correct, well-tested, and thoroughly documented.

**Key observations:**

- The fix is applied consistently to both conversion points: `updateSignatureFromBody` (used by `sign`/`verify`) and `bodyToByteArray` (used by `hybridSign`/`hybridVerify`). This symmetry prevents encoding mismatches between standard and hybrid code paths.

- Good use of `StandardCharsets.UTF_8` (compile-time constant) rather than `Charset.forName("UTF-8")`. The inline comments explaining the platform-dependent charset rationale are helpful for future maintainers.

- The test uses non-ASCII characters (accented Latin, em-dash) which is the right choice for detecting encoding mismatches — these are multi-byte in UTF-8 but single-byte in ISO-8859-1. The explicit cross-check at line 87 that ISO-8859-1 produces different bytes confirms the test's discriminating power.

- The upgrade guide entry correctly notes this is a no-op on Java 18+ (where `UTF-8` is the default charset) and only a live fix on Java 17 with a non-UTF-8 default. This is exactly the kind of precision users need.

- The PR description's honesty about scope and impact ("no-op on Java 18+, live fix on Java 17 with non-UTF-8 default") is a model for security-sensitive changes.

apupier:
conflict to resolve

PR #24717, opened 2026-07-15
CAMEL-24080: camel-aws2-kinesis - make consumer shard state thread-safe for multi-shard streams

oscerd:
## Description

Fixes [CAMEL-24080](https://issues.apache.org/jira/browse/CAMEL-24080).

When no `shardId` is configured, `Kinesis2Consumer.poll()` iterates the shard list with a `parallelStream()` and each worker runs `fetchAndPrepareRecordsForCamel(...)` concurrently. The shared per-consumer state written from those workers was not thread-safe:

- **`currentShardIterators`** was a plain `HashMap` (and **`warnLogged`** a plain `HashSet`), mutated concurrently via `updateShardIterator(...)` / `getShardIterator(...)`. Concurrent `put`/`add` can corrupt the map or lose iterator updates, leading to **lost or duplicated records** when consuming a stream with more than one shard.
- **`processedExchangeCount`** was updated with `getAndSet(processBatch(...))` (and reset with `set(0)` on the closed-shard path), so parallel shards overwrote each other and `poll()` returned only the **last-finishing shard's count** instead of the total. That count feeds `ScheduledBatchPollingConsumer` idle/backoff accounting (e.g. `backoffIdleThreshold`).

## Fix

- Use `ConcurrentHashMap` for `currentShardIterators` and `ConcurrentHashMap.newKeySet()` for `warnLogged`.
- Accumulate the poll count with `addAndGet(...)` and drop the `set(0)` reset on the closed-shard branch.
- `ConcurrentHashMap` forbids `null` values, and the closed-shard detection (`isShardClosed`, `updateShardIterator`) relied on storing a `null` iterator to mark a shard closed. A sentinel (empty string — already treated as "no iterator" by the existing `ObjectHelper.isEmpty(...)` checks) is stored instead, so closed-shard behavior is unchanged.

## Tests

- New `KinesisConsumerMultiShardTest` — two open shards, two records each; asserts `poll()` returns **4** (the sum), which fails against the old `getAndSet` logic (returned 2).
- Existing `KinesisConsumerClosedShardWith{Silent,Fail}Test` continue to pass, covering the null→sentinel closed-shard path.

## Backport

The same code is present on `camel-4.18.x` and `camel-4.14.x`; this will be backported to both after merge (fixVersions 4.22.0 / 4.18.4 / 4.14.9).

---

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Consumer.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/Kinesis2Consumer.java
@@ -19,12 +19,11 @@ package org.apache.camel.component.aws2.kinesis;
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.util.ArrayDeque;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


davsclaus (approved):
Clean, well-targeted fix for a real concurrency bug introduced when `synchronized` blocks were removed (CAMEL-20199). The analysis is thorough and the changes are minimal:

- `HashMap`/`HashSet` → `ConcurrentHashMap`/`ConcurrentHashMap.newKeySet()` — correct for the `parallelStream()` mutation in `poll()`
- `getAndSet` → `addAndGet` — properly accumulates across shards instead of last-writer-wins
- Removing `set(0)` on the closed-shard path — prevents resetting counts accumulated by other parallel workers
- Empty-string sentinel for null — clean solution for ConcurrentHashMap's null prohibition, logically equivalent to the old null-based detection

The new `KinesisConsumerMultiShardTest` exercises the exact failure mode (2 shards × 2 records, asserts total is 4). Existing closed-shard tests continue to validate that path.

LGTM.

gnodet (approved):
Well-analyzed concurrency fix that correctly replaces `HashMap`/`HashSet` with `ConcurrentHashMap` equivalents and fixes the `processedExchangeCount` race from `getAndSet` to `addAndGet`.

**Key observations:**

- The two bugs are real: (1) `HashMap` structural corruption under concurrent `put` from `parallelStream` workers, and (2) `getAndSet` overwriting the accumulated count — with 2 shards × 2 records, the old code would report only the last shard's count instead of the sum.

- The sentinel pattern (`CLOSED_SHARD_ITERATOR = ""`) for null values in `ConcurrentHashMap` is clean. `ObjectHelper.isEmpty("")` returns `true`, so existing null checks work correctly with the sentinel.

- Removing `processedExchangeCount.set(0)` on the closed-shard early-return path is correct — with `addAndGet`, a closed shard contributes zero by returning early. The old `set(0)` would have been destructive under `parallelStream` by zeroing counts from other shards.

- The pre-existing check-then-act in `getShardIterator` is not atomic, but is safe in practice because each shard operates on its own distinct key — no two `parallelStream` workers contend on the same `shardId`.

- The test validates the `addAndGet` accumulation fix (2 shards × 2 records = 4 total). True concurrent contention isn't exercised since Mockito stubs are synchronous, but this is acceptable for a unit test.

PR #24722, opened 2026-07-15
CAMEL-24083: camel-sjms - Use consumer exception handler for async failures

davsclaus:

## Summary

Same bug as CAMEL-24069 (camel-spring-rabbitmq), but in camel-sjms.

When `asyncConsumer=true` and an exchange fails after the listener returns (async completion), the failure was silently dropped because `EndpointMessageListener` called `endpoint.getExceptionHandler()` which is `null` by default, with no fallback.

The fix uses `consumer.getExceptionHandler()` instead, which always has a default `LoggingExceptionHandler` (set by `DefaultConsumer` constructor). This matches the canonical pattern used by `DefaultConsumer.DefaultConsumerCallback` and other components like camel-pulsar.

## Changes

- **`EndpointMessageListener.java`** (camel-sjms): Changed the async failure path from `endpoint.getExceptionHandler()` (nullable, no fallback) to `consumer.getExceptionHandler()` (always non-null)

## Test plan

- [x] All existing tests pass (`mvn verify` — 4 tests, 0 failures)

### DIFF
--- a/components/camel-sjms/src/main/java/org/apache/camel/component/sjms/consumer/EndpointMessageListener.java
+++ b/components/camel-sjms/src/main/java/org/apache/camel/component/sjms/consumer/EndpointMessageListener.java
@@ -456,10 +456,8 @@ public class EndpointMessageListener implements SessionMessageListener {
                     // method and rethrow it
                     exchange.setException(rce);
                 } else {
-                    // we were done async, so use the exception handler
-                    if (endpoint.getExceptionHandler() != null) {
-                        endpoint.getExceptionHandler().handleException(rce);
-                    }
+                    // we were done async, so use the consumer exception handler
+                    consumer.getExceptionHandler().handleException("Error processing exchange", exchange, rce);
                 }
             }
 


gnodet (approved):
Same fix as #24721 (spring-rabbitmq) applied to camel-sjms — both `EndpointMessageListener` implementations had the identical silent-error-swallowing pattern.

`DefaultEndpoint.exceptionHandler` is `null` by default, so the old `if (endpoint.getExceptionHandler() != null)` guard was effectively a no-op for async errors. Switching to `consumer.getExceptionHandler()` (always initialized to `LoggingExceptionHandler` in `DefaultConsumer`) ensures errors are never silently dropped, and the exchange context is passed through for diagnostics.

---

oscerd (approved):
LGTM. Switching from the endpoint's exception handler to `consumer.getExceptionHandler().handleException("Error processing exchange", exchange, rce)` for the async-completion failure path is the right call — it routes through the consumer's configured exception handler and carries the `exchange` context (and message) into the handler, which the old endpoint-handler overload dropped. Consistent with the sibling fix in camel-spring-rabbitmq (#24721). CI green.

PR #24723, opened 2026-07-15
[backport camel-4.18.x] CAMEL-23843: Camel-PQC - sign and verify String payloads as UTF-8

oscerd:
Backport of #24716 to `camel-4.18.x` (fixVersion **4.18.4**) — **charset only**.

`sign`/`verify` encoded a `String` body with `String.getBytes()` (the JVM default charset). The default charset is platform dependent, so signing on one JVM and verifying on another with a different default disagrees on the bytes of a non-ASCII payload and verification fails. Pin `StandardCharsets.UTF_8`.

**Scope vs main:** the native `byte[]`/`InputStream` handling delivered on main by CAMEL-23847 is an *improvement* kept to `main`, so it is intentionally **not** in this backport — only the charset determinism fix for `String` payloads. This branch has no hybrid sign/verify, so only `signature()`/`verification()` are touched.

**Note on the test:** the bug is JVM-default-charset dependent (a no-op on Java 18+ where the default is already UTF-8, per JEP 400), so the test pins the UTF-8 *contract* — it independently verifies the produced signature over the payload's UTF-8 bytes and confirms it does not match the ISO-8859-1 bytes. Validated green on this branch; no generated drift.

### DIFF
--- a/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
+++ b/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.component.pqc;
 
+import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.cert.Certificate;
 
@@ -129,7 +130,7 @@ public class PQCProducer extends DefaultProducer {
         String payload = exchange.getMessage().getMandatoryBody(String.class);
 
         signer.initSign(keyPair.getPrivate());
-        signer.update(payload.getBytes());
+        signer.update(payload.getBytes(StandardCharsets.UTF_8));
 
         byte[] signature = signer.sign();
         exchange.getMessage().setHeader(PQCConstants.SIGNATURE, signature);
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Well-scoped, minimal backport of the charset determinism fix — looks good.

**What works well:**

- The fix is correct and complete: both `signature()` and `verification()` now use `StandardCharsets.UTF_8` instead of the platform-dependent `String.getBytes()`.
- No other `.getBytes()` calls exist in `PQCProducer.java` on this branch, so nothing is missed.
- The test design is solid — independently verifying the signature over UTF-8 bytes *and* confirming it does *not* match ISO-8859-1 bytes pins the UTF-8 contract regardless of the JVM's default charset.
- Backport scope is appropriate: only the charset fix, no hybrid crypto code (which correctly doesn't exist on `camel-4.18.x`).

**Checklist:**
- [x] Tests included (PQCSignatureCharsetTest)
- [x] No `@author` tags
- [x] Backport scope — minimal, targeted fix only
- [x] Backward compat — improves cross-platform correctness
- [x] Security — deterministic encoding prevents subtle signature verification failures
- [x] Formatting — import placement correct

PR #24727, opened 2026-07-15
CAMEL-24070: camel-spring-rabbitmq - Fix thread-unsafe lazy template creation in SpringRabbitMQProducer

davsclaus:
## Summary

`SpringRabbitMQProducer` lazily creates its `RabbitTemplate` and `AsyncRabbitTemplate` without synchronization. Since a producer is a singleton invoked concurrently by route threads, concurrent first messages can each create their own template; all but the last assignment are dropped without being stopped. In the InOut case each racing thread also starts its template, so the losers are leaked in started state.

## Changes

- Added a `ReentrantLock` to guard `getInOnlyTemplate()`, `getInOutTemplate()`, and `doStop()` — same pattern already used by `EndpointMessageListener.getTemplate()` in this component
- Moved `inOutTemplate.start()` inside the null-check so it is only called once on creation, not on every invocation
- Fixed `Boolean.FALSE == sent` identity comparison on a primitive boolean to `!sent`
- Added `SpringRabbitMQProducerThreadSafetyTest` that verifies concurrent `getInOnlyTemplate()` calls from 8 threads always return the same instance

Fixes: [CAMEL-24070](https://issues.apache.org/jira/browse/CAMEL-24070)

## Test plan

- [x] New `SpringRabbitMQProducerThreadSafetyTest` passes — 8 threads behind a barrier all get the same template instance
- [x] All existing tests pass (35 run, 4 skipped integration tests)
- [x] Full reactor build passes

### DIFF
--- a/components/camel-spring-parent/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQProducer.java
+++ b/components/camel-spring-parent/camel-spring-rabbitmq/src/main/java/org/apache/camel/component/springrabbit/SpringRabbitMQProducer.java
@@ -19,6 +19,8 @@ package org.apache.camel.component.springrabbit;
 import java.util.Map;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.camel.AsyncCallback;
 import org.apache.camel.Endpoint;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean thread-safety fix — the race in lazy template creation is real and the `ReentrantLock` approach is correct.

**What works well:**

- Lock pattern matches the existing `EndpointMessageListener.getTemplate()` in this same component — consistent.
- `inOutTemplate.start()` moved inside the null-check is an important secondary fix: previously every `getInOutTemplate()` call re-started the template, which is wasteful and potentially confusing.
- `doStop()` also guarded by the lock — prevents a race between a late `process()` call and shutdown.
- `Boolean.FALSE == sent` → `!sent` is a good cleanup (identity comparison on autoboxed primitive).
- The test design with a `CountDownLatch` barrier maximises contention to expose the race.

**Checklist:**
- [x] Tests included (8-thread concurrent access test)
- [x] No public API changes
- [x] Thread-safety — lock scope is correct (covers check-then-act on both getters + stop)
- [x] No `@author` tags
- [ ] CI — builds pending

PR #24728, opened 2026-07-15
CAMEL-24086: camel-aws2-sqs - give each FIFO batch entry a distinct MessageDeduplicationId

oscerd:
## Description

Fixes [CAMEL-24086](https://issues.apache.org/jira/browse/CAMEL-24086).

When sending a batch (a `List` body, or a `String` split by `batchSeparator`) to a **FIFO** SQS queue, every entry is built from the same `Exchange`, so `configureFifoAttributes(SendMessageBatchRequestEntry.Builder, Exchange)` derived the `MessageDeduplicationId` from that one `Exchange`. The default `ExchangeIdMessageDeduplicationIdStrategy` returns `exchange.getExchangeId()` — identical for all entries — so SQS accepted the batch but **silently dropped all but the first message** within the 5-minute deduplication window.

## Fix

- The batch overload now appends the entry position to the strategy-provided id (`<id>-<index>`), giving each message a **distinct, deterministic** deduplication id — a redelivered batch still deduplicates correctly.
- A `null` id (content-based deduplication via `NullMessageDeduplicationIdStrategy`) is left unset.
- The group-id / dedup-id strategies are now **null-guarded** consistently with the single-message overload (the batch overload previously could NPE on a null strategy).
- The single-message send path is unchanged.

## Tests

Adds `SqsProducerFifoBatchDeduplicationTest` — sends a 4-element batch to a `.fifo` queue and asserts the four entries receive four **distinct** deduplication ids (and each sets a group id). Against the old code all four shared one id, so the test fails without the fix. The existing `SqsProducerBatchTest` still passes.

## Scope note

This PR addresses only the FIFO deduplication-id collision. The separate "batch not chunked to the 10-entry API limit" observation is not included here.

## Backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Producer.java
+++ b/components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Producer.java
@@ -121,6 +121,7 @@ public class Sqs2Producer extends DefaultProducer {
         Collection<SendMessageBatchRequestEntry> entries = new ArrayList<>();
         if (exchange.getIn().getBody() instanceof Iterable) {
             Iterable c = exchange.getIn().getBody(Iterable.class);
+            int index = 0;
             for (Object o : c) {
                 String object = (String) o;
                 SendMessageBatchRequestEntry.Builder entry = SendMessageBatchRequestEntry.builder();
@@ -128,7 +129,7 @@ public class Sqs2Producer extends DefaultProducer {
                 entry.messageAttributes(translateAttributes(exchange.getIn().getHeaders(), exchange));
                 entry.messageBody(object);
                 addDelay(entry, exchange);
-                configureFifoAttributes(entry, exchange);
+                configureFifoAttributes(entry, exchange, index++);
                 entries.add(entry.build());
             }
             request.entries(entries);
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-scoped fix for FIFO batch deduplication. A few observations:

**What the PR does well:**
- Correctly identifies that `MessageDeduplicationId` must be unique per batch entry — SQS silently drops entries with duplicate dedup IDs in FIFO queues
- Appending `-<index>` is deterministic and preserves the strategy-derived prefix, so the same batch re-sent produces the same per-entry IDs (idempotency preserved)
- Adds `ObjectHelper.isNotEmpty()` null guards on the batch path, aligning with the single-message `configureFifoAttributes` overload (line 227–238) which already had them
- Handles the `null` dedup ID case (content-based deduplication) by leaving it unset — correct behavior
- Covers both entry paths (`Iterable` and `String`-split)
- Good test with distinct ID assertion via `HashSet` size check

**Checklist:**
- [x] Tests: new `SqsProducerFifoBatchDeduplicationTest` covers the fix
- [x] Commit convention: JIRA-linked (`CAMEL-24086`)
- [x] Backward compatibility: maintained — only batch FIFO behavior changes (and was broken before)
- [x] No public API changes
- [x] Mock class already supports `getSendMessageBatchRequests()` — no test infra changes needed

davsclaus (approved):
Clean, well-scoped fix for the FIFO batch deduplication-id collision. The three changes are correct and minimal:

- Index-suffixed dedup ID (`<strategy-id>-<index>`) gives each batch entry a distinct, deterministic id — redelivered batches still deduplicate correctly
- Null id (content-based dedup via `NullMessageDeduplicationIdStrategy`) is correctly left unset
- Null guards on the batch overload now match the single-message overload, fixing a potential NPE

Test covers the key scenario (4-element batch → 4 distinct dedup ids). Single-message send path is unchanged.

PR #24729, opened 2026-07-15
CAMEL-24072: camel-jms - Fix transacted=true breaking ConnectionFactory autowiring

davsclaus:
## Summary

Fixes component-level `transacted=true` breaking `ConnectionFactory` autowiring during startup with `IllegalArgumentException: connectionFactory must be specified`.

The guard condition in `JmsComponent.doInit()` used `getOrCreateTransactionManager()` which eagerly creates a `JmsTransactionManager` — and that requires a `ConnectionFactory` that hasn't been autowired yet (since the autowiring is the body of the very if-statement being evaluated). Changed to `getTransactionManager()` (plain getter, no side effects) so the guard only checks whether a TM was explicitly configured, without triggering lazy creation.

- **Root cause**: CAMEL-14220 changed `getTransactionManager()` → `getOrCreateTransactionManager()` to fix camel-activemq (which overrides `createConnectionFactory()` and is unaffected). Plain camel-jms has no such override and throws.
- **Fix**: Revert the guard to `getTransactionManager()`. The lazy TM creation still happens later when the endpoint/consumer is actually created.
- **Tests**: Added `JmsTransactedAutowiredConnectionFactoryTest` — reproducer that fails without the fix. All camel-jms (437 tests), camel-activemq, and camel-activemq6 tests pass.

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsComponent.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsComponent.java
@@ -1126,7 +1126,7 @@ public class JmsComponent extends HeaderFilterStrategyComponent {
     @Override
     protected void doInit() throws Exception {
         // only attempt to set connection factory if there is no transaction manager
-        if (configuration.getConnectionFactory() == null && configuration.getOrCreateTransactionManager() == null
+        if (configuration.getConnectionFactory() == null && configuration.getTransactionManager() == null
                 && isAllowAutoWiredConnectionFactory()) {
             Set<ConnectionFactory> beans = getCamelContext().getRegistry().findByType(ConnectionFactory.class);
             if (beans.size() == 1) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean one-line fix for the autowiring regression. The root cause is well-identified:

**Analysis:**
- `doInit()` guards CF autowiring with `configuration.getOrCreateTransactionManager() == null` — but `getOrCreateTransactionManager()` has a side-effect: when `transacted=true` (and `lazyCreateTransactionManager=true`, the default), it eagerly creates a `JmsTransactionManager`, so the null check always fails and autowiring is skipped
- The fix uses `getTransactionManager()` (pure getter, no side-effect) which correctly returns `null` when no TM was explicitly configured by the user, allowing the CF to be autowired as intended
- The TM still gets created later via `getOrCreateTransactionManager()` at line 1808 during template configuration — so transactional behavior is preserved

**Checklist:**
- [x] Tests: new `JmsTransactedAutowiredConnectionFactoryTest` reproduces the bug and verifies the fix — uses ArtemisService test infra properly
- [x] Commit convention: JIRA-linked (`CAMEL-24072`)
- [x] Minimal and focused change — only the affected line is touched
- [x] No public API changes, backward compatible
- [x] No static analysis findings

PR #24730, opened 2026-07-15
CAMEL-24073: camel-jms - Cancel reply correlation on send failure to prevent double callback

davsclaus:

## Summary

Fixes a bug where `AsyncCallback.done()` is invoked twice when a JMS send fails after the reply correlation has been registered in `JmsProducer.processInOut`.

**Root cause:** `replyManager.registerReply()` runs inside `MessageCreator.createMessage()` (before the actual JMS send). If the send fails (broker quota, max message size, connection drop), the exception path completes the exchange via `callback.done(true)`, but the reply handler stays registered in the `CorrelationTimeoutMap`. At `requestTimeout` (default 20s), the eviction fires `onTimeout` → `processReply`, which sets `ExchangeTimedOutException` on the already-completed exchange and invokes `callback.done(false)` a second time.

**Fix:**
- Add `cancelCorrelationId(String)` to `ReplyManager` interface — removes a pending reply handler from the correlation map without invoking the callback
- Implement in `ReplyManagerSupport`
- Wrap `doSend()` in `JmsProducer.processInOut` with try/catch to cancel the registered correlation on send failure before re-throwing

## Test plan

- [x] New test `JmsInOutSendFailureCallbackTest` — uses a proxied `ConnectionFactory` that makes `MessageProducer.send()` throw, verifies the exchange exception remains the original JMS send failure and is not overwritten by `ExchangeTimedOutException` after the request timeout elapses
- [ ] CI passes

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsProducer.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsProducer.java
@@ -221,6 +221,8 @@ public class JmsProducer extends DefaultAsyncProducer {
             in.setHeader(correlationPropertyToUse, GENERATED_CORRELATION_ID_PREFIX + getUuidGenerator().generateUuid());
         }
 
+        final String[] registeredCorrelationId = new String[1];
+
         MessageCreator messageCreator = new MessageCreator() {
             public Message createMessage(Session session) throws JMSException {
                 Message answer = endpoint.getBinding().makeJmsMessage(exchange, in, session, null);
@@ -241,6 +243,7 @@ public class JmsProducer extends DefaultAsyncProducer {
 
                 String correlationId = determineCorrelationId(answer, provisionalCorrelationId);
                 replyManager.registerReply(replyManager, exchange, callback, originalCorrelationId, correlationId, timeout);
+                registeredCorrelationId[0] = correlationId;
 
                 if (correlationProperty != null) {
                     replyManager.setCorrelationProperty(correlationProperty);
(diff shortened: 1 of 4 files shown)


gnodet (approved):
Excellent fix for a subtle double-callback race in JMS InOut. The analysis:

**The bug:**
1. `processInOut` registers a reply correlation (with timeout handler) via `replyManager.registerReply()` inside `MessageCreator.createMessage()`
2. Then `doSend()` is called to send the message
3. If `doSend()` fails, the exception propagates and the exchange is completed with the send error
4. But the correlation entry is still registered — when the timeout fires, it invokes the callback a second time with `ExchangeTimedOutException`, overwriting the original error

**The fix:**
- Captures the correlation ID via `String[1]` array (standard pattern for anonymous class value capture)
- Wraps `doSend()` in try/catch — on failure, calls `replyManager.cancelCorrelationId()` to remove the orphaned correlation entry, then re-throws
- `cancelCorrelationId()` is null-safe: handles the case where `createMessage()` throws before `registerReply()` is reached (correlation ID still null)

**Design observations:**
- New `cancelCorrelationId()` on `ReplyManager` interface is clean — component-internal interface (in the `reply` sub-package), not public API
- `ReplyManagerSupport` provides the implementation; both `QueueReplyManager` and `TemporaryQueueReplyManager` inherit it
- The `correlation.remove()` is atomic on `ConcurrentMap`, so no race with the timeout thread

**Test design:**
- Clever JDK proxy chain (CF → Connection → Session → MessageProducer) to inject send failure without mocking framework
- Uses Awaitility `during(3s)` to wait past the `requestTimeout=2000` and verify the exception remains stable (not overwritten by timeout handler)
- Asserts the exception is NOT `ExchangeTimedOutException` — the exact symptom of the double-callback bug

**Checklist:**
- [x] Tests: thorough reproducer with timeout-stability verification
- [x] Commit convention: JIRA-linked (`CAMEL-24073`)
- [x] Static analysis: only pre-existing broad-exception-catch warnings (consistent with file style)
- [x] Backward compatible — only fixes buggy behavior
- [x] Javadoc on new interface method explains the purpose

PR #24731, opened 2026-07-15
CAMEL-24074: camel-jms - Synchronize TemporaryQueueReplyManager destination resolution

davsclaus:
## Summary

- Fix race condition in `TemporaryReplyQueueDestinationResolver.resolveDestinationName` where concurrent listener threads could each create separate temporary queues, causing replies to be black-holed
- Apply the same `ReentrantLock` + `volatile` double-checked locking pattern already used by `QueueReplyManager.DestinationResolverDelegate`
- Use a dedicated lock (not `BaseService.lock`) to preserve the CAMEL-20769 shutdown-deadlock fix

## Test plan

- [x] Existing JMS request-reply tests pass (15 tests)
- [x] Module builds clean
- [ ] CI green

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/TemporaryQueueReplyManager.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/TemporaryQueueReplyManager.java
@@ -17,6 +17,8 @@
 package org.apache.camel.component.jms.reply;
 
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import jakarta.jms.Destination;
 import jakarta.jms.ExceptionListener;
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Well-analyzed deadlock fix with correct concurrent resolution pattern. The analysis:

**The deadlock scenario (documented in the code comment):**
1. `BaseService.stop()` holds its own lock while calling `doStop()`
2. `doStop()` calls `listenerContainer.destroy()` → `doShutdown()`, which waits for listener threads to finish
3. If a listener thread concurrently calls `resolveDestinationName()` and that method synchronized on `BaseService.lock`, it deadlocks — the stop thread holds the lock and waits for the listener, while the listener waits for the lock

**The fix:**
- Introduces a dedicated `ReentrantLock destinationLock` separate from `BaseService.lock`, breaking the deadlock cycle
- Makes `queue` field `volatile` for safe unsynchronized reads on the fast path
- Implements correct double-check locking:
  - Fast path: volatile read of `queue` + atomic read of `refreshWanted` — returns immediately if both satisfied, no lock contention
  - Slow path: acquires `destinationLock`, rechecks, creates/refreshes the queue
- `doStop()` acquires the same `destinationLock` to ensure no resolution is in-flight during cleanup

**Correctness observations:**
- The local variable `answer` in the fast path avoids reading `volatile` twice (correct pattern)
- `refreshWanted` is `AtomicBoolean` — safe to read without the lock in the fast path
- `doStop()` sets `queue = null` inside the lock, so concurrent resolution threads won't see a deleted queue
- Visibility change `private` → package-private on the inner class is minor (likely for testing)

**No test included** — understandable for a deadlock fix; these are race-dependent and hard to reproduce deterministically in unit tests. The fix is verifiable by code inspection and the deadlock scenario is clearly documented.

**Checklist:**
- [x] Commit convention: JIRA-linked (`CAMEL-24074`)
- [x] Backward compatible
- [x] Prior concurrency work in this file (`630b74d1` "jms inout concurrency issue") confirms this area needs care

PR #24733, opened 2026-07-15
CAMEL-24075: camel-jms/camel-sjms - transferException should not swallow failures when no JMSReplyTo

davsclaus:
## Summary

- Fix `transferException=true` silently swallowing processing failures (and committing the transaction) when the incoming JMS message has no `JMSReplyTo` header
- The `transferException` feature was designed for request-reply (InOut) only — it suppresses the exception to transfer it as a reply body. When no reply destination exists, the exception must flow through normally so the transaction rolls back and the message is redelivered
- Applied the same fix to both `camel-jms` and `camel-sjms`

## What changed

**camel-jms** `EndpointMessageListener.evaluateExchangeResult()`: changed condition from `!endpoint.isTransferException()` to `!(endpoint.isTransferException() && sendReply)` — only suppress the exception when a reply will actually be sent.

**camel-sjms** `EndpointMessageListener.done()`: added `&& sendReply` to the `endpoint.isTransferException()` guard — same logical fix in the pre-refactoring code style.

## Test plan

- [x] New `JmsTransferExceptionNoReplyToIT` — sends a fire-and-forget message (InOnly, no JMSReplyTo) to a transacted queue with `transferException=true`; verifies the processor is invoked more than once (redelivery) and the message reaches the dead letter channel
- [x] Existing `JmsTransferExceptionIT` — verifies InOut request-reply with `transferException=true` still works (exception transferred as reply body)
- [x] Full camel-jms test suite (81 tests pass)
- [x] camel-sjms compiles successfully

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/EndpointMessageListener.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/EndpointMessageListener.java
@@ -213,7 +213,8 @@ public class EndpointMessageListener implements SessionAwareMessageListener {
             if (exchange.isRollbackOnly()) {
                 return wrapRuntimeCamelException(new RollbackExchangeException(exchange));
             }
-            if (exchange.isFailed() && exchange.getException() != null && !endpoint.isTransferException()) {
+            if (exchange.isFailed() && exchange.getException() != null
+                    && !(endpoint.isTransferException() && sendReply)) {
                 return wrapRuntimeCamelException(exchange.getException());
             }
             return null;
--- a/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/JmsTransferExceptionNoReplyToIT.java
+++ b/components/camel-jms/src/test/java/org/apache/camel/component/jms/integration/JmsTransferExceptionNoReplyToIT.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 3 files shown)


gnodet (approved):
Important fix that prevents silent message loss. Well-analyzed root cause.

**The bug:**
When `transferException=true` is set on a JMS consumer and a fire-and-forget message arrives (no `JMSReplyTo`), the consumer suppresses the exception unconditionally. Since there's no reply destination to serialize the exception to, it simply vanishes — the transaction commits and the message is lost without any error handling, redelivery, or dead-letter routing.

**The fix:**
- `camel-jms`: Changes `!endpoint.isTransferException()` to `!(endpoint.isTransferException() && sendReply)` — exception is only suppressed when there's an actual reply destination to transfer it to
- `camel-sjms`: Same logic — adds `&& sendReply` to the `isTransferException()` check

The truth table confirms correctness:

| `transferException` | `sendReply` | Suppress? | Why |
|---|---|---|---|
| true | true | Yes | Exception will be serialized in the reply message |
| true | false | No | No destination to transfer to; must re-throw for error handling |
| false | * | No | Transfer not enabled; re-throw as always |

`sendReply` is derived from `replyDestination != null && !disableReplyTo` (line 81 in camel-jms), so it correctly reflects whether an actual reply channel exists.

**Test:**
- `JmsTransferExceptionNoReplyToIT` sends a fire-and-forget message to a transacted queue with `transferException=true`
- Route processor always throws `IllegalArgumentException`
- Verifies redelivery happens (counter > 1) and the message reaches the dead letter channel
- Without the fix, the message is silently committed and lost

**Checklist:**
- [x] Tests: integration test reproduces the bug with proper assertions
- [x] Commit convention: JIRA-linked (`CAMEL-24075`)
- [x] Both `camel-jms` and `camel-sjms` fixed consistently
- [x] Backward compatible — only changes behavior for the buggy case (no reply destination)

PR #24736, opened 2026-07-15
CAMEL-24077: camel-amqp - Only append SSL transport options when configured

davsclaus:
## Summary

- Only append `transport.*` options to the Qpid JMS connection URI when the corresponding store location is actually configured (non-null). Previously, all six options were always appended with empty strings when `useSsl=true`, causing Qpid JMS to throw `FileNotFoundException` instead of using the JVM default trust store.
- URL-encode transport option values to prevent URI corruption from special characters in store paths or passwords.
- Add tests for trust-store-only TLS (server-only TLS, no client key store) and URL encoding of special characters.

## Test plan

- [x] `testEnabledSslComponent` — verifies `useSsl=true` with no stores produces `amqps://localhost:5672` (no empty query params)
- [x] `testConfiguredSslComponent` — verifies fully configured SSL still works
- [x] `testTrustStoreOnlySslComponent` — verifies trust-store-only TLS (common server-only case)
- [x] `testSslUrlEncodedPassword` — verifies special chars (`@`, `=`, `&`, `/`) are URL-encoded
- [x] Full `camel-amqp` test suite (25 tests pass)

### DIFF
--- a/components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPComponent.java
+++ b/components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPComponent.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.amqp;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean fix addressing two distinct issues in AMQP SSL URI construction.

**Bug 1 — Empty parameters:**
When `useSsl=true` but no trust/key store was configured, the URI contained empty parameters like `transport.trustStoreLocation=&transport.trustStorePassword=`, which could confuse the Qpid JMS connection factory.

**Bug 2 — Missing URL encoding:**
File paths (e.g., `/path/to/store.jks`) and passwords with special characters (e.g., `p@ss=w&rd`) were appended raw — `&` and `=` in passwords would break URI parsing.

**The fix:**
- Only appends trust store block when `trustStoreLocation != null`; only appends key store block when `keyStoreLocation != null`
- Passwords are conditional (appended only when non-null)
- All values are URL-encoded via `URLEncoder.encode(value, StandardCharsets.UTF_8)`
- `first` flag correctly manages `?` vs `&` separator

**Tests:**
- Updated `testEnabledSslComponent`: `useSsl=true` with no stores → clean `amqps://localhost:5672` (no trailing empty params)
- New `testTrustStoreOnlySslComponent`: trust store only, no key store → only trust store params appear
- New `testSslUrlEncodedPassword`: `p@ss=w&rd` → `p%40ss%3Dw%26rd` — verifies encoding of `@`, `=`, `&`

**Checklist:**
- [x] Tests: three test cases cover empty config, partial config, and special characters
- [x] Commit convention: JIRA-linked (`CAMEL-24077`)
- [x] Static analysis: no findings
- [x] Backward compatible — only changes behavior for previously broken edge cases
- [x] No new dependencies (`URLEncoder` and `StandardCharsets` are JDK)

PR #24737, opened 2026-07-15
CAMEL-24078: camel-jms/camel-amqp - Fix medium findings from code review

davsclaus:
## Summary

Fixes four medium-severity findings from the camel-jms/camel-amqp code review ([CAMEL-24078](https://issues.apache.org/jira/browse/CAMEL-24078)):

- **M3**: Endpoint-level `includeAllJMSXProperties=true` was silently ignored because the header filter strategy was constructed before URI options were applied. Now resets the cached `JmsHeaderFilterStrategy` when `includeAllJMSXProperties` changes.
- **M4**: Component-level `messageListenerContainerFactory` was silently ignored because `consumerType` was not switched to `Custom`. Aligned with the endpoint-level setter that already does this.
- **M6**: `asyncStartListener=true` defeated `autoStartup=false` due to a race — `initialized` was set before the async task ran, bypassing the auto-startup check. Moved the flag assignment into `prepareAndStartListenerContainer()`.
- **M9**: Setting `username`/`password` on the AMQP component clobbered an explicitly configured `connectionFactory` with a new `localhost:5672` factory. Guarded factory creation with a null check.

## Test plan

- [x] `mvn test` in `components/camel-jms` — all tests pass
- [x] `mvn test` in `components/camel-amqp` — all tests pass
- [ ] CI green

### DIFF
--- a/components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPComponent.java
+++ b/components/camel-amqp/src/main/java/org/apache/camel/component/amqp/AMQPComponent.java
@@ -110,8 +110,9 @@ public class AMQPComponent extends JmsComponent {
                 connectionFactory.setTopicPrefix("topic://");
             }
             getConfiguration().setConnectionFactory(connectionFactory);
-        } else if (host != null || port != null || getUsername() != null || getPassword() != null || useTopicPrefix != null
-                || useSsl != null) {
+        } else if (getConfiguration().getConnectionFactory() == null
+                && (host != null || port != null || getUsername() != null || getPassword() != null || useTopicPrefix != null
+                        || useSsl != null)) {
             StringBuilder sb = new StringBuilder();
             sb.append(useSsl == Boolean.TRUE ? "amqps://" : "amqp://");
             sb.append(host == null ? AMQP_DEFAULT_HOST : host).append(":").append(port == null ? AMQP_DEFAULT_PORT : port);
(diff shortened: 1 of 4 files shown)


gnodet (approved):
Good cleanup of four medium-severity findings across camel-jms and camel-amqp.

**Fix 1: AMQPComponent — Don't overwrite explicit ConnectionFactory**
The `else if` branch that auto-creates a `JmsConnectionFactory` now checks `getConfiguration().getConnectionFactory() == null` first. Previously, setting any of `host`/`port`/`username`/`password`/`useSsl`/`useTopicPrefix` would silently overwrite a user-provided CF. Correct guard.

**Fix 2: JmsComponent — Auto-set `ConsumerType.Custom` with custom factory**
`setMessageListenerContainerFactory()` now also sets `consumerType = Custom`. Without this, a custom factory could be provided but never used because the consumer type didn't match. Prevents a misconfiguration trap.

**Fix 3: JmsConsumer — Move `initialized = true` after auto-startup check**
Moved from `doStart()` (unconditionally set, even on failure) into `prepareAndStartListenerContainer()` (after `startListenerContainer()`). This means `initialized` stays `false` if container startup fails — more correct for the lazy-init path that checks this flag.

**Fix 4: JmsEndpoint — Reset header filter strategy on `includeAllJMSXProperties` change**
When `setIncludeAllJMSXProperties()` is called after the `JmsHeaderFilterStrategy` was already lazily created (line 394), the strategy wouldn't reflect the new value. Now it's nulled so `getHeaderFilterStrategy()` re-creates it with the updated flag. The `instanceof JmsHeaderFilterStrategy` guard correctly preserves user-provided custom strategies.

**Checklist:**
- [x] Commit convention: JIRA-linked (`CAMEL-24078`)
- [x] All fixes are defensive/correctness improvements — no behavioral regression for correct configurations
- [x] Header filter strategy reset is security-relevant and correctly handled

PR #24740, opened 2026-07-15
CAMEL-24126 - Ensure no message sent before the VertxWebsocket is fully configured

apupier:
it was visible with flaky test
VertxWebsocketHandshakeHeadersTest.testHandshakeHeadersAsConsumer see rtx-websocket:surefire:test@default-test/details/org.apache.camel.component.vertx.websocket.VertxWebsocketHandshakeHeadersTest/testHandshakeHeadersAsConsumer?top-execution=1

Co-authored-by: IBM Bob IDE 2.0.1

On main branch, i reproduce the flakiness something like 1 out of 3 times. With this PR, it has never failed in 10 runs

# Description

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketEndpoint.java
+++ b/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketEndpoint.java
@@ -152,6 +152,16 @@ public class VertxWebsocketEndpoint extends DefaultEndpoint implements EndpointS
     }
 
     protected WebSocket getWebSocket() throws Exception {
+        return getWebSocket((java.util.function.Consumer<WebSocket>) null);
+    }
+
+    /**
+     * Connects to the WebSocket server, optionally invoking a setup callback on the socket <em>before</em> the blocking
+     * {@code future.get()} returns. This guarantees that any message/close/exception handlers supplied by the callback
+     * are registered atomically with the connection, so no messages sent by the server immediately after the handshake
+     * can be missed.
+     */
+    protected WebSocket getWebSocket(java.util.function.Consumer<WebSocket> setupCallback) throws Exception {
         if (client == null) {
             resolvedClientOptions = configuration.getClientOptions();
             if (resolvedClientOptions == null) {
@@ -171,8 +181,17 @@ public class VertxWebsocketEndpoint extends DefaultEndpoint implements EndpointS
             CompletableFuture<WebSocket> future = new CompletableFuture<>();
             client.connect(connectOptions).onComplete(result -> {
                 if (result.succeeded()) {
-                    LOG.info("Connected to WebSocket on {}", result.result().remoteAddress());
-                    future.complete(result.result());
+                    WebSocket ws = result.result();
+                    LOG.info("Connected to WebSocket on {}", ws.remoteAddress());
+                    if (setupCallback != null) {
+                        try {
+                            setupCallback.accept(ws);
+                        } catch (Exception e) {
+                            future.completeExceptionally(e);
+                            return;
+                        }
+                    }
+                    future.complete(ws);
                 } else {
                     webSocket = null;
                     future.completeExceptionally(result.cause());
--- a/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketClientConsumer.java
+++ b/components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketClientConsumer.java
@@ -43,7 +43,7 @@ public class VertxWebsocketClientConsumer extends DefaultConsumer {
 
     @Override
     protected void doStart() throws Exception {
-        configureWebSocketHandlers(getEndpoint().getWebSocket());
+        getEndpoint().getWebSocket(this::configureWebSocketHandlers);
     }
 
     protected void configureWebSocketHandlers(WebSocket webSocket) {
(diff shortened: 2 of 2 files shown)


gnodet (approved):
**Well-designed race condition fix — LGTM.**

### What the PR does

Fixes a race where the WebSocket server could send messages immediately after the handshake completes, but before the consumer's message/close/exception handlers were registered — causing lost messages.

**Before:** `doStart()` called `getWebSocket()` (which blocks until connection completes), then separately called `configureWebSocketHandlers(ws)`. A window existed between connection establishment and handler registration.

**After:** `doStart()` calls `getWebSocket(this::configureWebSocketHandlers)`, passing handler setup as a callback. The callback is invoked inside the Vert.x `onComplete` block *before* `future.complete(ws)`, so handlers are registered atomically with the connection — no window for lost messages.

The same fix is correctly applied to the reconnection path as well.

### Design notes
- The FQCN `java.util.function.Consumer<WebSocket>` in the method signature is the right call — `org.apache.camel.Consumer` is already imported (line 34), so the simple name `Consumer` would be ambiguous.
- The Javadoc on the new `getWebSocket(Consumer<WebSocket>)` overload clearly explains the atomicity guarantee.
- No test is included, which is expected — this is a timing-sensitive race condition that is not reliably reproducible in a unit test. The fix is verifiable by code inspection.

### Checklist
- [x] Commit convention — descriptive title
- [x] No public API break — `getWebSocket()` is `protected`, new overload is additive
- [x] No security concerns
- [x] CI — all checks passing (JDK 17, JDK 25)

### Static analysis
ast-grep (only tool available) — no findings on the modified files.

---

components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketEndpoint.java
```diff
@@ -171,8 +181,12 @@ protected WebSocket getWebSocket() throws Exception {
...
             client.connect(connectOptions).onComplete(result -> {
                 if (result.succeeded()) {
-                    LOG.info("Connected to WebSocket on {}", result.result().remoteAddress());
-                    future.complete(result.result());
+                    WebSocket ws = result.result();
+                    LOG.info("Connected to WebSocket on {}", ws.remoteAddress());
+                    if (setupCallback != null) {
+                        setupCallback.accept(ws);
+                    }
+                    future.complete(ws);
```
davsclaus, line 194: Minor robustness suggestion: if `setupCallback.accept(ws)` throws, `future.complete(ws)` is never reached and the caller gets a `TimeoutException` rather than the real cause. Consider:

```suggestion
                    if (setupCallback != null) {
                        try {
                            setupCallback.accept(ws);
                        } catch (Exception e) {
                            future.completeExceptionally(e);
                            return;
                        }
                    }
```

Low-risk in practice since `configureWebSocketHandlers` only calls handler setters, but would help debugging if this pattern is reused elsewhere.

davsclaus (commented):
Thanks for the contribution @apupier — nice catch on the race condition at the framework level.

The fix looks correct: registering handlers inside the `onComplete` callback before `future.complete(ws)` ensures no messages are lost between connection and handler setup. The producer path and reconnect path are both unaffected/correctly updated.

A couple of observations:

1. **JIRA ticket** — Since this modifies production code (`VertxWebsocketEndpoint` and `VertxWebsocketClientConsumer`), it would be good to have a JIRA issue for tracking. The commit message would then follow the `CAMEL-XXXX: Brief description` convention.

2. **Minor error-handling gap** — If `setupCallback.accept(ws)` were to throw, `future.complete(ws)` would never be called and the caller would get a `TimeoutException` instead of the real cause. In practice `configureWebSocketHandlers` only calls handler setters so this is low-risk, but wrapping the callback in a try/catch that calls `future.completeExceptionally(e)` would be more robust.

components/camel-vertx/camel-vertx-websocket/src/main/java/org/apache/camel/component/vertx/websocket/VertxWebsocketEndpoint.java
```diff
@@ -171,8 +181,17 @@ protected WebSocket getWebSocket() throws Exception {
...
-                    LOG.info("Connected to WebSocket on {}", result.result().remoteAddress());
-                    future.complete(result.result());
+                    WebSocket ws = result.result();
+                    LOG.info("Connected to WebSocket on {}", ws.remoteAddress());
+                    if (setupCallback != null) {
+                        try {
+                            setupCallback.accept(ws);
+                        } catch(Exception e) {
+                            future.completeExceptionally(e);
+                            return;
```
davsclaus, line 191: Minor formatting nit — missing space before the parenthesis:

```suggestion
                        } catch (Exception e) {
```

davsclaus (commented):
Thanks for fixing this race condition, @apupier — the approach of registering handlers inside the `onComplete` callback before `future.complete()` is clean and correct.

One minor formatting nit inline. Otherwise this looks good.

davsclaus (approved):

oscerd (approved):
LGTM. The fix registers the message/close/exception handlers inside the connect `onComplete` block (on the Vert.x event-loop thread) *before* `future.complete(ws)`, closing the window where an inbound frame delivered between completion and the old route-thread handler-registration could be dropped. No deadlock (the blocking `future.get()` runs on a different thread from the completing callback), and the producer path is unchanged. Your earlier concern about a throwing callback surfacing as `TimeoutException` is already handled in HEAD via the `try/catch` → `completeExceptionally`. (Coverage rests on the existing — historically flaky — handshake test rather than a new deterministic one, but the race isn't easily reproducible.)

PR #24742, opened 2026-07-15
CAMEL-24124: Drain pending reply handlers during shutdown

davsclaus:
## Summary

- Fix `DefaultTimeoutMap.doStop()` to emit `Evict` events for remaining entries instead of silently calling `map.clear()` — ensures listeners are notified during shutdown
- Relax `processReply()` `isRunAllowed()` guard in camel-jms, camel-sjms, and camel-rocketmq reply managers to allow reply processing during the STOPPING phase
- Replies arriving during shutdown are processed normally; pending entries that never received a reply are drained with `RejectedExecutionException` and their callbacks completed

This fixes a 45-second graceful-shutdown stall caused by InOut exchanges whose `AsyncCallback` was never completed when the reply manager stopped.

Fixes: [CAMEL-24124](https://issues.apache.org/jira/browse/CAMEL-24124)

## Test plan

- [x] `mvn verify` passes for camel-jms
- [x] `mvn verify` passes for camel-sjms
- [x] `mvn install -DskipTests` passes for camel-rocketmq (requires broker for tests)
- [x] `mvn install -DskipTests` passes for core/camel-support

### DIFF
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/CorrelationTimeoutMap.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/CorrelationTimeoutMap.java
@@ -58,6 +58,12 @@ class CorrelationTimeoutMap extends DefaultTimeoutMap<String, ReplyHandler> {
     }
 
     @Override
+    protected void doStop() throws Exception {
+        drainAndEvictAll();
+        super.doStop();
+    }
+
+    @Override
     public ReplyHandler put(String key, ReplyHandler value, long timeoutMillis) {
         return super.put(key, value, encode(timeoutMillis));
     }
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/ReplyManagerSupport.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/reply/ReplyManagerSupport.java
@@ -19,6 +19,7 @@ package org.apache.camel.component.jms.reply;
 import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 
 import jakarta.jms.Destination;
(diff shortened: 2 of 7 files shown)


gnodet (approved):
**Solid shutdown-drain fix that prevents hanging exchanges — LGTM.**

### What the PR does

Fixes a shutdown hang where pending InOut reply handlers were silently discarded, leaving callers blocked forever.

**Root cause:** `DefaultTimeoutMap.doStop()` previously called `map.clear()` without emitting `Evict` events. Even when regular timeout eviction fired during shutdown, the `processReply()` methods in all three reply managers had an `!isRunAllowed()` guard that silently returned — never invoking the callback, leaving the waiting exchange stuck.

**Fix (two parts):**

1. **`DefaultTimeoutMap.doStop()`**: Drains remaining entries under the lock, then emits `Evict` events for each *outside* the lock. This ensures `CorrelationTimeoutMap`'s listener fires `handler.onTimeout(key)` for every pending reply, which creates a `ReplyHolder(timeout=true)` and calls `processReply()`.

2. **`processReply()` in all three reply managers** (camel-jms, camel-sjms, camel-rocketmq):
   - Removes the `!isRunAllowed()` early return that was silently discarding entries during shutdown.
   - Replaces with `null` check on `holder` (the actual guard needed).
   - During shutdown (`isStopping() || isStopped()`), timeout entries get a `RejectedExecutionException` instead of the normal timeout log/exception — providing a meaningful error while still invoking the callback to unblock the caller.
   - Non-timeout entries (real replies that arrived) are still processed normally during shutdown.

### Design observations

- The drain-then-emit pattern (collect under lock, emit outside) correctly avoids holding the lock during listener callbacks.
- The `RejectedExecutionException` is semantically appropriate — it signals "work was rejected because the service is shutting down," which is exactly what's happening.
- The camel-sjms and camel-rocketmq refactoring into `handleTimeout()` / `handleSuccessfulReply()` methods improves readability without changing behavior.
- The RocketMQ `processReply()` also gets aligned with the JMS pattern — it previously had inline timeout logic, now consistently uses `handleTimeout()`.

### Checklist
- [x] Consistent across all three messaging components (JMS, SJMS, RocketMQ)
- [x] Core change (`DefaultTimeoutMap`) is backward-compatible — existing listeners just receive additional Evict events during shutdown
- [x] No public API break
- [x] No security concerns

### Static analysis
ast-grep (only tool available) flagged pre-existing broad `catch (Exception)` blocks in `DefaultTimeoutMap` and `ReplyManagerSupport` — none introduced by this PR.

---

PR #24743, opened 2026-07-15
CAMEL-24089: camel-file - Loop transferTo to handle files larger than 2 GB

davsclaus:
## Summary

- Fix `FileChannel.transferTo()` in `FileOperations.writeFileByFile` to loop until all bytes are transferred, preventing silent file truncation for files larger than ~2 GB when using `readLock=fileLock` on Linux.

## Details

`FileChannel.transferTo()` is documented to possibly transfer fewer bytes than requested. On Linux, the JDK caps each call at `Integer.MAX_VALUE` bytes (~2 GB) via `sendfile(2)`. The existing code made a single call and discarded the return value, so any file larger than ~2 GB was silently truncated while the producer reported success.

The fix replaces the single call with the standard `transferTo` loop pattern — the same pattern the pre-2017 code in this area used before it was removed in d8c0a53b6787.

Only the `readLock=fileLock` code path is affected. The non-locked path uses `Files.copy()` which handles large files correctly.

## Test plan

- [x] `mvn verify` passes in `components/camel-file`

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/FileOperations.java
@@ -522,7 +522,11 @@ public class FileOperations implements GenericFileOperations<File> {
             try (FileOutputStream fos = new FileOutputStream(target);
                  FileChannel out = fos.getChannel()) {
                 LOG.trace("writeFileByFile using FileChannel: {} -> {}", source, target);
-                channel.transferTo(0, channel.size(), out);
+                long size = channel.size();
+                long position = 0;
+                while (position < size) {
+                    position += channel.transferTo(position, size - position, out);
+                }
             }
         } else {
             // use regular file copy


gnodet (approved):
**Textbook fix for a well-known `FileChannel.transferTo` limitation — LGTM.**

### What the PR does

Fixes `writeFileByFile()` to loop `FileChannel.transferTo()` until all bytes are transferred, instead of assuming a single call transfers everything.

**Before:** `channel.transferTo(0, channel.size(), out)` — a single call that may silently transfer fewer bytes than requested, truncating files larger than ~2 GB (the exact threshold is OS/filesystem-dependent).

**After:** Standard loop pattern that tracks position and keeps calling `transferTo()` until all bytes are transferred:
```java
long position = 0;
while (position < size) {
    position += channel.transferTo(position, size - position, out);
}
```

This is explicitly documented in the [Java API](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/FileChannel.html#transferTo(long,long,java.nio.channels.WritableByteChannel)): *"An invocation of this method may or may not transfer all of the requested bytes"*.

### Checklist
- [x] No public API break — internal implementation detail
- [x] No security concerns

### Static analysis
ast-grep (only tool available) — no findings on the modified file.

---

PR #24745, opened 2026-07-15
CAMEL-24090: camel-file - Drain eagerly-added idempotent keys on poll exception

davsclaus:
## Summary

Fixes [CAMEL-24090](https://issues.apache.org/jira/browse/CAMEL-24090).

With `idempotent=true` (implied by `noop=true`) and default `idempotentEager=true`, keys are added to the idempotent repository during directory scanning. If `pollDirectory` then throws (FTP disconnect mid-scan, I/O error, filter exception), the catch block in `GenericFileConsumer.poll()` drains via the **List overload** of `removeExcessiveInProgressFiles`, which only removes in-progress keys — the eagerly-added idempotent keys stay behind. Files are then permanently skipped on subsequent polls.

CAMEL-21830 (commit `130bde60462b`) fixed the **Deque overload** used for batch draining but missed the List overload on the poll-exception path.

### Changes

- **`GenericFileConsumer.java`**: Mirror the Deque overload — in the List drain, when `isIdempotentEager()` and an idempotent repository is configured, call `removeExcessiveIdempotentFile(file, null)`.
- **`FileIdempotentEagerPollExceptionTest.java`**: New test that creates two files with `preSort=name` for deterministic ordering. A filter throws on the second file during the first poll only. Without the fix, the first file's idempotent key leaks and it is never consumed (test times out). With the fix, both files are consumed on subsequent polls.

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileConsumer.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileConsumer.java
@@ -311,6 +311,9 @@ public abstract class GenericFileConsumer<T> extends ScheduledBatchPollingConsum
         for (GenericFile file : files) {
             String key = file.getAbsoluteFilePath();
             endpoint.getInProgressRepository().remove(key);
+            if (endpoint.isIdempotentEager() && endpoint.getIdempotentRepository() != null) {
+                removeExcessiveIdempotentFile(file, null);
+            }
         }
     }
 
--- a/core/camel-core/src/test/java/org/apache/camel/component/file/FileIdempotentEagerPollExceptionTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/file/FileIdempotentEagerPollExceptionTest.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


gnodet (approved):
**Clean fix for idempotent key leak on mid-scan failure — LGTM.**

### What the PR does

Fixes a bug where eagerly-added idempotent keys leak when `pollDirectory()` throws mid-scan, causing files to be permanently skipped on subsequent polls.

**Scenario (with `idempotentEager=true`, which is the default):**
1. Two files exist: `a.txt`, `b.txt`. `preSort=name` guarantees `a.txt` is scanned first.
2. `a.txt` is scanned — its idempotent key is eagerly added to the repository.
3. While scanning `b.txt`, an exception occurs (e.g., a filter throws).
4. The poll fails; `removeExcessiveInProgressFiles()` is called to clean up.
5. **Before fix:** Only in-progress repository entries are cleaned — `a.txt`'s idempotent key remains, so it's permanently skipped.
6. **After fix:** Eagerly-added idempotent keys are also removed, so `a.txt` is correctly retried on the next poll.

**The fix:** 3 lines added to `removeExcessiveInProgressFiles()` — when `idempotentEager` is enabled, also call `removeExcessiveIdempotentFile(file, null)` for each file. The `null` dynamic exchange parameter is correct — `removeExcessiveIdempotentFile` handles it via `GenericFileHelper.createDummy()`.

### Test
`FileIdempotentEagerPollExceptionTest` — `ThrowOnceFilter` simulates a mid-scan failure on the first poll, then verifies both files are consumed on subsequent polls. Uses `MockEndpoint.assertIsSatisfied(context, 10, TimeUnit.SECONDS)` for the timed assertion — correct per project conventions.

### Checklist
- [x] Tests — new test reproduces the exact scenario
- [x] No public API break
- [x] No security concerns

### Static analysis
ast-grep (only tool available) — all findings are pre-existing broad-exception catches, none introduced by this PR.

---

PR #24746, opened 2026-07-15
CAMEL-24091: camel-file - Snapshot idempotent key at poll time to prevent key mismatch

davsclaus:
## Summary

- Fixes a regression from CAMEL-21733 (Camel 4.11) where `GenericFileOnCompletion` re-evaluates the idempotent key expression at completion time using the routed exchange's mutated headers, causing key mismatch
- Snapshots the idempotent key as an exchange property (`CamelFileIdempotentKey`) in `processExchange()` before routing, and reads it back at commit/rollback time instead of re-evaluating
- Fixes three failure modes: failed files never retried (eager+rollback), endless duplicates (non-eager+noop), and unconfirmed entries after restart (JDBC/Kafka repos)

## Test plan

- [x] New test `FileConsumerIdempotentKeyHeaderMutationTest` — route mutates `CamelFileName`, first attempt fails, verifies file is retried on next poll
- [x] All 29 existing idempotent file tests pass (`FileConsumerIdempotentKey*`, `FileConsumeNoopIdempotent*`, `FileIdempotent*`)

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileConsumer.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileConsumer.java
@@ -410,6 +410,13 @@ public abstract class GenericFileConsumer<T> extends ScheduledBatchPollingConsum
         // and then the file name would be changed
         String absoluteFileName = file.getAbsoluteFilePath();
 
+        // snapshot the idempotent key now while the exchange still has pristine file headers;
+        // after routing, headers like CamelFileName may be mutated, causing key mismatch at completion time
+        if (Boolean.TRUE.equals(endpoint.isIdempotent()) && endpoint.getIdempotentKey() != null) {
+            String key = endpoint.getIdempotentKey().evaluate(exchange, String.class);
+            exchange.setProperty(Exchange.FILE_IDEMPOTENT_KEY, key);
+        }
+
         // check if we can begin processing the file
         Exception beginCause = null;
         boolean begin = false;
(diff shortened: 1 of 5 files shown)


gnodet (approved):
**Well-designed fix for idempotent key mismatch caused by header mutation — LGTM.**

### What the PR does

Fixes a bug where mutating `CamelFileName` (or other file headers) during routing causes the idempotent key to be re-evaluated to a different value at completion time, breaking both commit and rollback.

**Scenario (with `idempotentKey=${file:onlyname}-${file:size}` and `idempotentEager=true`):**
1. At poll time, the idempotent key is computed from pristine headers (e.g., `report.txt-1024`) and eagerly added to the repository.
2. During routing, the route mutates `CamelFileName` (e.g., `.setHeader(Exchange.FILE_NAME, "output.txt")`).
3. At completion time, the key is re-evaluated from mutated headers → `output.txt-1024` (wrong key).
4. On **rollback**: The wrong key is removed, the original key is orphaned → file is never retried.
5. On **commit**: The wrong key is confirmed, original key leaks → file may be re-processed.

**Fix — snapshot-at-poll-time pattern:**
1. **`GenericFileConsumer.processExchange()`**: Snapshots the idempotent key into `exchange.setProperty(Exchange.FILE_IDEMPOTENT_KEY, key)` while headers are still pristine.
2. **`GenericFileOnCompletion.processStrategyCommit()` / `processStrategyRollback()`**: Reads the snapshot from the exchange property instead of re-evaluating the expression. Falls back to `absoluteFileName` if the property is absent (backward-compatible with older exchanges).
3. **New constant** `Exchange.FILE_IDEMPOTENT_KEY = "CamelFileIdempotentKey"` — consistent with existing `FILE_*` constants. Generated `ExchangeConstantProvider` updated accordingly.

### Test
`FileConsumerIdempotentKeyHeaderMutationTest` — sets up `idempotentKey=${file:onlyname}-${file:size}`, mutates `CamelFileName` in the route, throws on first attempt (triggering rollback), and verifies the file is successfully retried and consumed on the second attempt.

### Checklist
- [x] Tests — new test reproduces the exact scenario
- [x] Generated files updated (`ExchangeConstantProvider.java`)
- [x] New public API constant is consistent with existing `FILE_*` naming convention
- [x] No security concerns

### Static analysis
ast-grep (only tool available) — all findings are pre-existing, none introduced by this PR.

---

PR #24747, opened 2026-07-15
CAMEL-24092: camel-file - readLock=fileLock must return false on IOException

davsclaus:
## Summary

Fix two bugs in `FileLockExclusiveReadLockStrategy`:

- **IOException fall-through returns `true` with no lock held**: When `IOException` occurs during lock acquisition (file deleted between poll and lock, permission error, Windows AV lock) with `timeout > 0`, `handleIOException` returned `false`, causing the catch block to NOT return `false`. Control fell through: the `finally` block deleted the `.camelLock` marker file (via `releaseExclusiveReadLockOnAbort`), then the method returned `true` with `lock = null` — the consumer would process the file with no OS lock and no marker protection, enabling duplicate consumption by competing consumers. Now always returns `false` on `IOException`.

- **Wrong property key in `doReleaseExclusiveReadLock`**: Used `FILE_LOCK_EXCLUSIVE_LOCK` instead of `FILE_LOCK_RANDOM_ACCESS_FILE` to retrieve the `RandomAccessFile`, so `rac` was always `null`. No fd leak today only because closing the `FileLock`'s channel closes the shared fd, but the code was silently wrong.

The IOException fall-through dates to CAMEL-1195 (2008, deliberate "process anyway on Windows-AV" degradation). CAMEL-5324 (2012) later layered marker files into this strategy, and CAMEL-7988 (2014) added the `finally` cleanup — the combination silently dropped the cross-JVM protection that CAMEL-5324 added.

## Test plan

- [x] New `FileLockExclusiveReadLockStrategyTest` — makes file read-only so `RandomAccessFile("rw")` throws, asserts `acquireExclusiveReadLock` returns `false` and marker file is cleaned up

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/strategy/FileLockExclusiveReadLockStrategy.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/strategy/FileLockExclusiveReadLockStrategy.java
@@ -116,9 +116,8 @@ public class FileLockExclusiveReadLockStrategy extends MarkerFileExclusiveReadLo
             // somehow hold a lock to a file
             // such as AntiVirus or MS Office that has special locks for it's
             // supported files
-            if (handleIOException(e)) {
-                return false;
-            }
+            handleIOException(e);
+            return false;
         } finally {
             // close channels if we did not grab the lock
             if (!exclusive) {
@@ -139,19 +138,10 @@ public class FileLockExclusiveReadLockStrategy extends MarkerFileExclusiveReadLo
         return true;
     }
 
-    private boolean handleIOException(IOException e) {
-        if (timeout == 0) {
-            // if not using timeout, then we cant retry, so return false
-            return true;
+    private void handleIOException(IOException e) {
+        if (timeout > 0) {
+            LOG.debug("Cannot acquire read lock due to an IOException, will skip the file.", e);
         }
-        LOG.debug("Cannot acquire read lock. Will try again.", e);
-        boolean interrupted = sleep();
-        if (interrupted) {
-            // we were interrupted while sleeping, we are likely being
-            // shutdown so return false
-            return true;
-        }
-        return false;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Two good fixes in one — IOException handling simplified and a copy-paste property key bug fixed. LGTM.**

### What the PR does

**Fix 1 — Always return `false` on `IOException`:**

Before, `handleIOException()` would sleep and then return to the retry loop when `timeout > 0`, effectively retrying the `RandomAccessFile("rw")` open until the timeout expired. But an `IOException` here (permission denied, antivirus lock, read-only file) is not a transient lock contention — it's a fundamental problem that won't resolve by sleeping. The old behavior caused the consumer to hang for the full timeout duration on files that could never be locked.

Now, `IOException` immediately returns `false` (skip the file). The polling consumer's natural retry cycle will try again on the next poll, which is the right granularity.

**Fix 2 — Wrong property key in `doReleaseExclusiveReadLock()`:**

```java
// Bug — used FILE_LOCK_EXCLUSIVE_LOCK for the RandomAccessFile lookup:
exchange.getProperty(asExclusiveReadLockKey(file, Exchange.FILE_LOCK_EXCLUSIVE_LOCK), RandomAccessFile.class);

// Fix — uses the correct FILE_LOCK_RANDOM_ACCESS_FILE key:
exchange.getProperty(asExclusiveReadLockKey(file, Exchange.FILE_LOCK_RANDOM_ACCESS_FILE), RandomAccessFile.class);
```

This copy-paste bug meant the `RandomAccessFile` was never properly retrieved during release, potentially leaking file handles.

### Test
`FileLockExclusiveReadLockStrategyTest` — makes a file read-only to trigger `IOException` on `new RandomAccessFile("rw")`, verifies `acquireExclusiveReadLock` returns `false`, and checks the marker file is cleaned up. `@DisabledOnOs(OS.WINDOWS)` is appropriate since Windows handles file permissions differently.

### Checklist
- [x] Tests — new test covers the IOException path
- [x] No public API break
- [x] No security concerns

### Static analysis
ast-grep (only tool available) — no findings on the modified files.

---

PR #24750, opened 2026-07-15
CAMEL-24095: camel-file - Add missing Move branch to handleExistingTarget for tempFileName with eagerDeleteTargetFile=false

davsclaus:
## Summary

`GenericFileProducer.handleExistingTarget` (the non-eager path used when writing via `tempFileName`/`tempPrefix` with `eagerDeleteTargetFile=false`) handled `Ignore`, `Fail`, and `Override` but had **no `Move` branch**. This meant `fileExist=Move` with `eagerDeleteTargetFile=false` and `tempFileName` silently overwrote the existing target instead of moving it to the `moveExisting` destination.

The eager twin `handleExistingTargetEager` has had the `Move` branch since CAMEL-15822 (2020), but the non-eager method was never patched.

### Changes

- **`GenericFileProducer.java`**: Added the missing `Move` branch to `handleExistingTarget`, mirroring `handleExistingTargetEager` — calls `endpoint.getMoveExistingFileStrategy().moveExistingFile()`.
- **`FileProducerMoveExistingTest.java`**: Updated `testFailOnMoveExistingFileExistsEagerDeleteFalseTempFileName` to assert correct behavior: when the move destination already exists and `eagerDeleteTargetFile=false`, the move strategy should throw `GenericFileOperationFailedException` (consistent with the non-tempFileName variant `testFailOnMoveExistingFileExistsEagerDeleteFalse`).

## Test plan

- [x] `FileProducerMoveExistingTest` — all 14 tests pass (including the corrected test)
- [x] Full reactor build passes (`mvn clean install -DskipTests`)

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileProducer.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileProducer.java
@@ -254,6 +254,9 @@ public class GenericFileProducer<T> extends DefaultAsyncProducer {
             return doIgnore(target);
         } else if (endpoint.getFileExist() == GenericFileExist.Fail) {
             throwFileAlreadyExistException(target);
+        } else if (endpoint.getFileExist() == GenericFileExist.Move) {
+            // move any existing file first
+            this.endpoint.getMoveExistingFileStrategy().moveExistingFile(endpoint, operations, target);
         } else if (endpoint.getFileExist() == GenericFileExist.Override) {
             // we override the target so we do this by deleting
             // it so the temp file can be renamed later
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Clean, surgical bug fix.** The `Move` case was genuinely missing from `handleExistingTarget()` — confirmed by comparing with `handleExistingTargetEager()` (line 235-237) which has had this branch since CAMEL-15822. The non-eager path never got patched, meaning `fileExist=Move` + `eagerDeleteTargetFile=false` + `tempFileName` silently fell through to the rename step, overwriting the target instead of moving it first.

**What looks good:**

- **3-line production fix** exactly mirrors the eager-path code — same method call, same arguments, same position in the if-else chain
- **Test correction** is the strongest part: the old test was asserting *buggy behavior* (silent overwrite) as correct. The updated test now properly expects `GenericFileOperationFailedException("Cannot move existing file...")` when the move destination already exists — matching the existing `testFailOnMoveExistingFileExistsEagerDeleteFalse` test (the analogous non-tempFileName variant)
- The fix also benefits `RemoteFileProducer` (FTP/SFTP) and `SmbProducer` through inheritance

No issues found.

---

PR #24754, opened 2026-07-15
CAMEL-23703: camel-launcher - cross-compile native camel.exe for x64 and arm64 via llvm-mingw

ammachado:

# Description

This PR splits the native Windows executable cross-compile work out of the larger CAMEL-23703 package distribution branch.

It replaces the Windows/MSVC-only `camel.exe` build with a host-independent clang/llvm-mingw flow and publishes separate launcher executables for Windows x64 and Windows ARM64:

- `bin/camel-x64.exe`
- `bin/camel-arm64.exe`

The release profile now activates `camel.exe.build`, and the launcher module copies both classified `camel-exe` artifacts into the distribution when that property is enabled. The native executable workflow is renamed from the old Windows-only workflow to `camel-launcher-native-exe.yml`.

The native bootstrap uses a wide-character `wmain` entry point to preserve Unicode command lines, so both llvm-mingw compile invocations pass `-municode`.

## Toolchain provisioning

The llvm-mingw download/verify logic (pinned version + SHA256) lives in a single shared script, `.github/actions/setup-llvm-mingw/install-llvm-mingw.sh`, used by both the GitHub Actions composite action and `Jenkinsfile.deploy`. The CI snapshot-deploy pipeline reuses an agent-provided toolchain when the cross-compilers are already on `PATH`, and otherwise downloads the pinned, SHA256-verified copy into a persistent cache. This removes the need to pre-provision llvm-mingw on the ASF Jenkins agents. A local `-Prelease` build still requires the toolchain on `PATH` (documented in the release guide).

The native `camel.exe` build is scoped to the branches that actually carry the launcher: the `Setup llvm-mingw` stage and `-Dcamel.exe.build=true` only run on `main` and the upcoming `camel-4.22.x` LTS. The older LTS branches (`camel-4.18.x`, `camel-4.14.x`) continue to deploy their snapshots without the native build.

# Target

# Tracking

Formatter verification was run locally with `mvn formatter:format impsort:sort` and passed. Full build/test verification is intentionally deferred to CI/PR for this split.

# AI-assisted contributions

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -55,7 +55,7 @@ pipeline {
            }
         }
 
-        stage('Verify llvm-mingw') {
+        stage('Setup llvm-mingw') {
             when {
                 anyOf {
                     branch 'main'
@@ -64,11 +64,29 @@ pipeline {
                 }
             }
             steps {
+                script {
+                    // Prefer an agent-provided llvm-mingw when the cross-compilers are already
+                    // on PATH; otherwise download the pinned, SHA256-verified toolchain into a
+                    // persistent cache so snapshot deploys don't require it to be pre-provisioned.
+                    // The install script is the single source of truth shared with
+                    // .github/actions/setup-llvm-mingw/action.yml.
+                    env.LLVM_MINGW_BIN = sh(
+                        returnStdout: true,
+                        script: '''
+                            set -eo pipefail
+                            if command -v x86_64-w64-mingw32-clang >/dev/null 2>&1 \
+                                    && command -v aarch64-w64-mingw32-clang >/dev/null 2>&1; then
+                                dirname "$(command -v x86_64-w64-mingw32-clang)"
+                            else
+                                ./.github/actions/setup-llvm-mingw/install-llvm-mingw.sh "$HOME/.cache/llvm-mingw"
+                            fi
+                        '''
+                    ).trim()
+                }
                 sh '''
                     for compiler in x86_64-w64-mingw32-clang aarch64-w64-mingw32-clang; do
-                        if ! command -v "$compiler" >/dev/null 2>&1; then
-                            echo "ERROR: $compiler not found on PATH."
-                            echo "Install llvm-mingw per docs/user-manual/modules/ROOT/pages/release-guide.adoc"
+                        if ! [ -x "${LLVM_MINGW_BIN}/${compiler}" ]; then
+                            echo "ERROR: $compiler not found under ${LLVM_MINGW_BIN}."
                             exit 1
                         fi
                     done
@@ -85,7 +103,7 @@ pipeline {
                 }
             }
             steps {
-                sh "./mvnw $MAVEN_PARAMS -Pdeploy,apache-snapshots -Dquickly clean deploy"
+                sh "PATH=\"${LLVM_MINGW_BIN}:\$PATH\" ./mvnw $MAVEN_PARAMS -Pdeploy,apache-snapshots -Dquickly clean deploy"
             }
         }
 
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -17,7 +17,7 @@
 def AGENT_LABEL = env.AGENT_LABEL ?: 'ubuntu'
 def JDK_NAME = env.JDK_NAME ?: 'jdk_25_latest'
 
-def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly -Dcamel.exe.build=true "
+def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly "
 
 pipeline {
 
@@ -59,8 +59,7 @@ pipeline {
             when {
                 anyOf {
                     branch 'main'
-                    branch 'camel-4.18.x'
-                    branch 'camel-4.14.x'
+                    branch 'camel-4.22.x'
                 }
             }
             steps {
@@ -103,7 +102,15 @@ pipeline {
                 }
             }
             steps {
-                sh "PATH=\"${LLVM_MINGW_BIN}:\$PATH\" ./mvnw $MAVEN_PARAMS -Pdeploy,apache-snapshots -Dquickly clean deploy"
+                script {
+                    // The native camel.exe cross-compile only exists on main and the upcoming
+                    // camel-4.22.x LTS. Older LTS branches still deploy their snapshots, but
+                    // without the native build (no -Dcamel.exe.build and no llvm-mingw on PATH).
+                    def nativeExe = env.BRANCH_NAME in ['main', 'camel-4.22.x']
+                    def exeParams = nativeExe ? '-Dcamel.exe.build=true ' : ''
+                    def pathPrefix = nativeExe ? "PATH=\"${env.LLVM_MINGW_BIN}:\$PATH\" " : ''
+                    sh "${pathPrefix}./mvnw $MAVEN_PARAMS ${exeParams}-Pdeploy,apache-snapshots -Dquickly clean deploy"
+                }
             }
         }
 
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -17,7 +17,7 @@
 def AGENT_LABEL = env.AGENT_LABEL ?: 'ubuntu'
 def JDK_NAME = env.JDK_NAME ?: 'jdk_25_latest'
 
-def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly "
+def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly -Dcamel.exe.build=true "
 
 pipeline {
 
@@ -55,6 +55,27 @@ pipeline {
            }
         }
 
+        stage('Verify llvm-mingw') {
+            when {
+                anyOf {
+                    branch 'main'
+                    branch 'camel-4.18.x'
+                    branch 'camel-4.14.x'
+                }
+            }
+            steps {
+                sh '''
+                    for compiler in x86_64-w64-mingw32-clang aarch64-w64-mingw32-clang; do
+                        if ! command -v "$compiler" >/dev/null 2>&1; then
+                            echo "ERROR: $compiler not found on PATH."
+                            echo "Install llvm-mingw per docs/user-manual/modules/ROOT/pages/release-guide.adoc"
+                            exit 1
+                        fi
+                    done
+                '''
+            }
+        }
+
         stage('Build & Deploy') {
             when {
                 anyOf {
@@ -89,4 +110,3 @@ pipeline {
         }
     }
 }
-
--- a/.github/actions/install-mvnd/action.yml
+++ b/.github/actions/install-mvnd/action.yml
@@ -15,6 +15,10 @@
 # limitations under the License.
 #
 
+# Use mvnd only for jobs that build a large, parallelizable reactor, where the
+# daemon's concurrent module builds outweigh its cold-start cost. For small
+# reactors (a handful of modules) prefer plain `mvn`: mvnd adds no parallelism
+# benefit there and introduces a daemon startup timeout as a flaky failure mode.
 name: 'install-mvnd'
 description: 'Install the maven daemon'
 inputs:
--- a/.github/actions/setup-llvm-mingw/action.yml
+++ b/.github/actions/setup-llvm-mingw/action.yml
@@ -1,0 +1,38 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: 'setup-llvm-mingw'
+description: >-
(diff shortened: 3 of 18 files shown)


gnodet (approved):
**Clean infrastructure migration.** Moves the native `camel.exe` build from MSVC (Windows-only, x64-only) to clang/llvm-mingw (any host OS, x64 + arm64), and all CI is green.

**What looks good:**

- **SHA256-pinned toolchain setup** — the `setup-llvm-mingw` composite action pins version `20260616` with SHA256 verification before extracting, preventing supply-chain attacks on the toolchain itself
- **Profile activation migration** — from OS-based (`<family>windows</family>`) to property-based (`camel.exe.build=true`), correctly wired through `-Prelease` in root `pom.xml`
- **CI restructured well** — Linux-based cross-compilation for the build, with a separate `camel-exe-windows-smoke` job that downloads the artifact and smoke-tests on actual Windows
- **Release gate preserved** — `maven-enforcer-plugin` checks for both `camel-x64.exe` and `camel-arm64.exe` before packaging
- **Jenkinsfile.deploy** adds a `Verify llvm-mingw` stage that fails loudly if cross-compilers aren't on PATH
- **Comprehensive documentation** — upgrade guide explains the naming change (`camel.exe` → `camel-x64.exe`/`camel-arm64.exe`), release guide adds llvm-mingw as a prerequisite, all READMEs updated
- **`CamelExeBootstrapTest`** correctly auto-selects the architecture-appropriate exe (`os.arch` → `x64`/`arm64`)
- **`RepackageMojo.includeArtifact()`** comment and test updated to reference both exe files

**CI status:** All green — `camel-exe` (1m), `camel-exe-windows-smoke` (1m), `camel-launcher-native` (10m), `build (17)` (2h25m), `build (25)` (2h11m) all passed.

No issues found.

---

davsclaus (commented):
## Review: CAMEL-23703 cross-compile native camel.exe for x64 and arm64

Well-structured PR with comprehensive documentation updates. All CI checks pass. One infrastructure-coordination concern.

### Confirmed Issues

**1. Jenkinsfile.deploy: snapshot deploys will require llvm-mingw on Jenkins agents (Medium)**

Adding `-Dcamel.exe.build=true` to `MAVEN_PARAMS` means every snapshot deploy on `main`, `camel-4.18.x`, and `camel-4.14.x` will activate the `build-native-exe` profile in `tooling/camel-exe/pom.xml`. The exec-maven-plugin invocations run at `generate-test-resources` phase — this phase executes even with `-Dquickly` and `-DskipTests`. If the ASF Jenkins agents don't have `x86_64-w64-mingw32-clang` and `aarch64-w64-mingw32-clang` on PATH, all snapshot deploys break.

The added `Verify llvm-mingw` stage provides a clear early error, which is good. But this is a new infrastructure prerequisite that didn't exist before.

### Questions

**1. Are the ASF Jenkins agents already provisioned with llvm-mingw?**

The GitHub Actions workflow self-installs via `.github/actions/setup-llvm-mingw/`, but the Jenkinsfile assumes the toolchain is pre-installed. The `release-guide.adoc` update documents this as a prerequisite. If the agents aren't ready, should the Jenkinsfile changes be deferred to a follow-up PR merged after infrastructure setup?

**2. Dual exe compilation in CI workflow (minor)**

The exe is compiled twice in the GitHub Actions workflow: once in the `camel-exe` job and again in the `camel-launcher-native` job. Not incorrect, but the `camel-launcher-native` job could potentially reuse the artifact from `camel-exe` to save CI minutes.

### No Issues Found

- **Git history:** Builds on `d603bdd` (original camel.exe) — no prior intentional work reverted; this evolves the same feature.
- **Tests:** Properly updated. Bootstrap test remains `@EnabledOnOs(WINDOWS)`. Launcher IT drops the OS restriction (correct for cross-compiled file-existence checks).
- **Documentation:** Upgrade guide documents the `camel.exe` → `camel-x64.exe`/`camel-arm64.exe` rename. Release guide documents llvm-mingw prerequisite. All READMEs updated.
- **Security:** SHA256 pinning for the llvm-mingw download.
- **Attribution:** Follows project rules.
- **Release profile:** `camel.exe.build=true` added to root `release` profile — correct.

This review covers project rules and conventions. It does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud.

Jenkinsfile.deploy
```diff
@@ -17,7 +17,7 @@
 def AGENT_LABEL = env.AGENT_LABEL ?: 'ubuntu'
 def JDK_NAME = env.JDK_NAME ?: 'jdk_25_latest'
 
-def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly "
+def MAVEN_PARAMS = "-U -B -e -fae -V -Dnoassembly -Dcamel.exe.build=true "
```
davsclaus: Adding `-Dcamel.exe.build=true` here means every snapshot deploy will activate the `build-native-exe` profile in `tooling/camel-exe/pom.xml`. The exec-maven-plugin runs at `generate-test-resources` phase, which executes even with `-Dquickly` and `-DskipTests`.

If the ASF Jenkins agents don't already have `x86_64-w64-mingw32-clang` and `aarch64-w64-mingw32-clang` on PATH, snapshot deploys for `main`, `camel-4.18.x`, and `camel-4.14.x` will break. The `Verify llvm-mingw` stage (line 58) provides a clear early error, which is good — but the infrastructure needs to be ready before this PR merges.

Is llvm-mingw already installed on the Jenkins agents, or should the Jenkinsfile changes be deferred until after provisioning?

oscerd (commented):
The migration itself looks solid — genuine dual-target build (`x86_64-w64-mingw32-clang` → `camel-x64.exe` and `aarch64-w64-mingw32-clang` → `camel-arm64.exe`, both attached with separate classifiers and required by the enforcer), the llvm-mingw toolchain is version-pinned and SHA-256-verified over HTTPS with `set -eo pipefail`, all `actions/*` are SHA-pinned with `persist-credentials: false`, no hardcoded secrets, and `-municode`/`wmain` is correct. CI (camel-exe, windows-smoke, launcher-native) is green.

Two things worth resolving before merge, both non-code:
- **Release/deploy prerequisite** — `Jenkinsfile.deploy` sets `-Dcamel.exe.build=true` and the `build-native-exe` exec runs at `generate-test-resources` (even with `-DskipTests`), so if the ASF Jenkins agents don't have `x86_64-w64-mingw32-clang`/`aarch64-w64-mingw32-clang` on PATH, snapshot deploys on main (and the 4.18.x/4.14.x branches the Jenkinsfile gating anticipates) would break. The new loud-fail "Verify llvm-mingw" stage mitigates it, but the "are the agents provisioned?" question you raised looks still-open.
- arm64 is compiled but not runtime-smoke-tested (no Windows-arm64 CI runner) — acceptable given the shared ~100-line source, just noting the residual risk.

ammachado:
On the "are the ASF Jenkins agents provisioned with llvm-mingw?" question (thanks @davsclaus, @oscerd): `Jenkinsfile.deploy` no longer depends on that. The `Setup llvm-mingw` stage reuses the toolchain if the cross-compilers are already on the agent's PATH, and otherwise downloads the pinned, SHA256-verified llvm-mingw into a persistent cache (`$HOME/.cache/llvm-mingw`) before deploying. So snapshot deploys work whether or not the agents are pre-provisioned, and if infra installs it system-wide later, the pipeline just stops downloading with no code change.

The download/verify logic and the pinned version + hash now live in one shared script, `.github/actions/setup-llvm-mingw/install-llvm-mingw.sh`, used by both the GitHub Actions composite action and the Jenkinsfile, so there's a single source of truth. The release guide is updated to reflect this.

I left the arm64 runtime-smoke gap as-is, since there's no Windows-arm64 runner; the source is shared with x64, which is smoke-tested.

davsclaus (approved):

ammachado:
On the dual exe compilation (thanks @davsclaus): I looked at reusing the `camel-exe` artifact in `camel-launcher-native`, and it turned out the exe compile wasn't where the time was going. The exe is a ~100-line static C file, so each cross-compile is a sub-second clang call. Reusing that specific artifact across jobs would mean uploading both classifiers, downloading and `install:install-file`-ing them, and splitting the single `camel.exe.build` property that currently drives the compile, the copy-into-launcher, and the release gate together. That is a fair bit of coupling to shave off a couple of seconds.

The real cost in that job was `-am` rebuilding the whole launcher reactor (camel-core, the jbang tree, and the rest) from source. So I dropped `-am`: `camel-launcher-native` now builds only `buildingtools`, `camel-exe`, and `camel-launcher`, and resolves the launcher's other in-repo modules as the current version's snapshot from Apache snapshots. That takes the job from ~10 min down to a couple. A preflight step checks that snapshot tree is actually published (using `camel-launcher` itself as the sentinel, since it is deployed after its dependencies) and fails early with a clear message if it is not, so a version-bump window reads as "snapshot not deployed yet" rather than a deep resolution error mid-build.

While in there I also moved both exe jobs off mvnd to plain `mvn`. With the smaller reactors the daemon bought no parallelism, and its cold-start was intermittently timing out, which was the earlier `camel-launcher-native` failure.

One tradeoff worth calling out: the launcher packaging check now trusts Apache snapshots for its unchanged upstream modules instead of building them from source. Since this workflow only triggers on `camel-exe`/`camel-launcher` changes, those upstream modules are identical to main, so it validates the same bytecode either way. Happy to reconsider if you would rather keep it hermetic with `-am`.

Jenkinsfile.deploy
```diff
@@ -55,6 +55,45 @@ pipeline {
            }
         }
 
+        stage('Setup llvm-mingw') {
+            when {
+                anyOf {
+                    branch 'main'
```
davsclaus, line 61: this should only be main branch and in the future camel-4.22.x LTS (older LTS should not be included)

davsclaus (changes requested):

Jenkinsfile.deploy
ammachado: Done in 1cbf8eb. The snapshot-deploy pipeline still runs on `main`, `camel-4.18.x` and `camel-4.14.x`, but the native `camel.exe` bits are now gated to `main` and the future `camel-4.22.x` only:

- the `Setup llvm-mingw` stage's `when` block now lists just `main` and `camel-4.22.x`;
- `-Dcamel.exe.build=true` moved out of the shared `MAVEN_PARAMS` and, together with the `llvm-mingw` `PATH` prefix, is only added in `Build & Deploy` when the branch is `main` or `camel-4.22.x`.

The older LTS branches keep deploying their snapshots as before, just without the native build.

ammachado: Good call, scoped it down to `main` and the future `camel-4.22.x` in 1cbf8eb.

Jenkinsfile.deploy
ammachado: No pre-provisioning needed anymore, and this is now moot for the older LTS branches. Two things changed since you wrote this:

1. **Self-provisioning** (cb815e0): the `Setup llvm-mingw` stage runs `.github/actions/setup-llvm-mingw/install-llvm-mingw.sh`, which downloads a pinned, SHA256-verified llvm-mingw into `$HOME/.cache/llvm-mingw` when the cross-compilers aren't already on `PATH`. The agents don't need it pre-installed; the persistent cache keeps repeat runs cheap, and an agent that already has the compilers on `PATH` short-circuits the download. The compiler-existence check at the end of the stage still fails fast if provisioning ever goes wrong.

2. **Branch scoping** (1cbf8eb, from your other comment): `-Dcamel.exe.build=true` moved out of the shared `MAVEN_PARAMS`, and both the `Setup llvm-mingw` stage and the native flags now only run on `main` and the future `camel-4.22.x`. `camel-4.18.x` and `camel-4.14.x` deploy their snapshots without touching the native build, so those pipelines can't break.

So no deferral needed: old LTS is untouched, and `main`'s worst case is a one-time toolchain download after a cache wipe.

davsclaus (approved):

Jenkinsfile.deploy
davsclaus: Thanks

ammachado: One honest caveat on timing, to correct my note above: I can't exercise the native stage on demand. The snapshot-deploy pipeline runs on ASF Jenkins, and the `camel-4.22.x` branch only comes into existence at release time, so the first real run of the toolchain self-provisioning happens in ASF CI as part of the release cycle, not something that can be triggered beforehand.

That said, it shouldn't need to block or be deferred: the native stage is inert until that branch exists (older LTS never touches it), and when it does run, the self-provisioning downloads the pinned toolchain if it's missing. The compiler-existence check in the `Setup llvm-mingw` stage is there so that first run fails fast and loud if anything is off, rather than silently producing a broken deploy.

gnodet (approved):

## Re-review — rebased with CI refinements

Good refinements since my last review. The new commits add:

### 1. Self-provisioning llvm-mingw in Jenkinsfile.deploy ✅
- `Setup llvm-mingw` stage gated to `main` and `camel-4.22.x` — older LTS branches deploy without native exe
- Reuses the shared `install-llvm-mingw.sh` script (single source of truth for pinned version + SHA256)
- Falls back to agent-provided toolchain if already on PATH, otherwise downloads into `$HOME/.cache/llvm-mingw`
- Verification step confirms both cross-compilers are present before proceeding

### 2. Plain mvn for small reactors ✅
- Explains rationale in `install-mvnd/action.yml` comment: mvnd's daemon cold-start timeout adds a flaky failure mode for small reactors with no parallelism benefit
- `camel-exe` and `camel-launcher-native` jobs use plain `mvn` appropriately

### 3. Snapshot dependency resolution for launcher-native ✅
- Uses `-P apache-snapshots` without `-am` to resolve upstream modules from Apache snapshot repo
- Smart sentinel check: verifies `camel-launcher` snapshot metadata exists before starting, failing early with a clear message instead of a deep Maven resolution error
- Significantly reduces build scope (handful of modules vs full launcher reactor)

### 4. Branch gating ✅
- Native exe build gated to `main` and `camel-4.22.x` in both Jenkinsfile.deploy and the workflow comments
- Older LTS branches continue deploying snapshots normally without the native build

All the infrastructure decisions are well-reasoned and documented with inline comments. CI is running. LGTM.

davsclaus (approved):

PR #24756, opened 2026-07-16
CAMEL-24097: camel-ftp - SFTP knownHostsUri/knownHosts clobbered by ~/.ssh/known_hosts fallback

davsclaus:
## Summary

- Fix `SftpOperations.createSession` unconditionally applying `~/.ssh/known_hosts` fallback that overwrites any previously configured `knownHostsUri` or `knownHosts` byte array (JSch's `setKnownHosts()` wipes the prior repository)
- The fallback guard now also checks that `knownHostsUri` and `knownHosts` are not set, so it only fires when none of the three explicit known-hosts sources are configured

## Test plan

- [x] `camel-ftp` module compiles
- [ ] Existing SFTP integration tests pass
- [ ] Verify that `knownHostsUri` without `useUserKnownHostsFile=false` no longer gets clobbered

### DIFF
--- a/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/SftpOperations.java
+++ b/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/SftpOperations.java
@@ -333,7 +333,8 @@ public class SftpOperations implements RemoteFileOperations<SftpRemoteFile> {
         }
 
         String knownHostsFile = sftpConfig.getKnownHostsFile();
-        if (knownHostsFile == null && sftpConfig.isUseUserKnownHostsFile()) {
+        if (knownHostsFile == null && !isNotEmpty(sftpConfig.getKnownHostsUri())
+                && sftpConfig.getKnownHosts() == null && sftpConfig.isUseUserKnownHostsFile()) {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean, surgical fix for a real configuration-clobbering bug.

**The problem:** When a user configures `knownHostsUri` or `knownHosts` (byte array) but not `knownHostsFile`, the fallback at line 336 only checked `knownHostsFile == null` — so it triggered and overwrote the user's explicit configuration by calling `jsch.setKnownHosts("~/.ssh/known_hosts")` *after* the URI/byte-array known hosts had already been set (lines 319-333).

**The fix:** Guards the fallback with two additional checks — `!isNotEmpty(knownHostsUri)` and `knownHosts == null` — so `~/.ssh/known_hosts` is only applied when *no* known hosts source has been configured. This correctly preserves the priority: explicit `knownHostsFile` > `knownHostsUri` > `knownHosts` (byte[]) > user default `~/.ssh/known_hosts`.

**Checklist:**
- ✅ Logic is correct — the three known-hosts sources (file, URI, byte array) are now properly guarded
- ✅ Consistent with the existing configuration priority in `createSession()`
- ℹ️ No test, but this is difficult to unit test without an SFTP server — the fix is straightforward enough that the code speaks for itself

PR #24757, opened 2026-07-16
CAMEL-24098: camel-ftp - Use per-session config for SFTP ciphers and keyExchangeProtocols

davsclaus:
## Summary

`SftpOperations.createSession()` applied the `ciphers` and `keyExchangeProtocols` endpoint options via `JSch.setConfig(...)` — which is **static and JVM-global** — on every (re)connect. This caused cross-endpoint contamination: one endpoint's cipher/kex settings permanently altered the JSch defaults for every other SFTP endpoint in the JVM.

This PR moves both options to use `session.setConfig(...)` (per-session), consistent with every other option in the same method (`StrictHostKeyChecking`, `compression.*`, `PreferredAuthentications`, `server_host_key`, `PubkeyAcceptedAlgorithms`, `kex`, `ca_signature_algorithms`).

**Changes:**
- Replace `JSch.setConfig(ciphers)` (global) with `session.setConfig("cipher.s2c"/cipher.c2s", ...)` (per-session)
- Replace `JSch.setConfig("kex", ...)` (global) with `session.setConfig("kex", ...)` (per-session)
- Remove unused `Hashtable` import

## Test plan

- [x] New `SftpCipherGlobalConfigIT` — records JSch global defaults, sends a message through an endpoint with custom ciphers and kex, asserts the globals are unchanged
- [x] Existing `SftpSetCipherIT` passes (functional cipher test)
- [x] Existing `SftpKeyExchangeProtocolsIT` passes (functional kex test)
- [x] Full `mvn verify` on `camel-ftp` passes

### DIFF
--- a/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/SftpOperations.java
+++ b/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/SftpOperations.java
@@ -33,7 +33,6 @@ import java.nio.file.Paths;
 import java.security.KeyPair;
 import java.time.Duration;
 import java.util.Base64;
-import java.util.Hashtable;
 import java.util.List;
 import java.util.Vector;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Great catch — this fixes a real thread-safety and isolation issue with global JSch state mutation.

**The problem:** `JSch.setConfig()` is a **static/global** method — calling it with per-endpoint cipher and KEX settings mutates the configuration for *every* SFTP session created by *any* `JSch` instance in the JVM. When multiple Camel SFTP endpoints run concurrently with different cipher configurations, they silently interfere with each other: the last endpoint to call `JSch.setConfig()` wins for all connections.

**The fix:** Moves cipher and KEX configuration from `JSch.setConfig()` (global, static) to `session.setConfig()` (per-session), which is the correct JSch API for per-connection overrides. The config is applied after `jsch.getSession()` creates the session but before `session.connect()`.

**What this PR does well:**

- Properly scopes security-sensitive configuration (ciphers, key exchange protocols) to individual sessions
- Removes the `Hashtable` import — the old `new Hashtable<>()` was only needed for `JSch.setConfig(Hashtable)`
- Includes an integration test (`SftpCipherGlobalConfigIT`) that verifies JSch global config is **not** mutated after an endpoint connection — tests exactly the bug that was fixed
- The test snapshots defaults before, connects with custom ciphers, then asserts globals are unchanged

**Checklist:**
- ✅ Tests: integration test verifies global state is not mutated
- ✅ Security: cipher/KEX isolation prevents cross-endpoint contamination
- ✅ No public API change

orpiske (approved):

PR #24758, opened 2026-07-16
CAMEL-24099: camel-bean - Use ServiceHelper for BeanProcessor lifecycle delegation

davsclaus:
## Summary

- Fix `BeanProcessor` lifecycle delegation to use `ServiceHelper` instead of direct method calls, ensuring the `BaseService` state machine is properly traversed. The key bug was `doStop()` calling `delegate.doStop()` (raw method), which bypassed state tracking and prevented bean restart.
- All four lifecycle methods now use `ServiceHelper`: `initService`, `startService`, `stopService`, `stopAndShutdownService`.

## Test plan

- [x] `mvn test` passes in `components/camel-bean`
- [ ] CI green

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanProcessor.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/component/bean/BeanProcessor.java
@@ -27,6 +27,7 @@ import org.apache.camel.Exchange;
 import org.apache.camel.Processor;
 import org.apache.camel.spi.ErrorHandlerAware;
 import org.apache.camel.spi.IdAware;
+import org.apache.camel.support.service.ServiceHelper;
 import org.apache.camel.support.service.ServiceSupport;
 
 public class BeanProcessor extends ServiceSupport implements AsyncProcessor, ErrorHandlerAware, IdAware {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean refactoring — uses the standard Camel lifecycle delegation pattern via `ServiceHelper`.

**Key fix:** The original `delegate.doStop()` call was bypassing `ServiceSupport`'s lifecycle state management — `doStop()` is a protected template method meant to be called *by* the framework, not directly. The other methods (`init()`, `start()`, `shutdown()`) were correctly calling the public API, but `doStop()` was the odd one out. `ServiceHelper.stopService()` properly dispatches to `Service.stop()`, which handles state transitions (e.g., marking the service as stopped).

Using `ServiceHelper` also adds null-safety (all methods accept `@Nullable`) and consistent error handling across all lifecycle methods. This is the idiomatic Camel pattern for delegating lifecycle management.

**Checklist:**
- ✅ Consistent use of `ServiceHelper` for all lifecycle methods (init, start, stop, shutdown)
- ✅ Fixes direct `doStop()` call that bypassed lifecycle state management
- ✅ `doShutdown()` now uses `stopAndShutdownService()` which ensures both stop + shutdown
- ✅ No public API change

PR #24762, opened 2026-07-16
CAMEL-24101: camel-bean - Fix OGNL list index off-by-one at boundary

davsclaus:
## Summary

- Fix off-by-one in `BeanExpression.lookupResult()` list bounds check: `list.size() > num - 1` → `list.size() > num`
- When the OGNL index equals the list size, the guard incorrectly passed and `list.get(num)` threw a raw JDK `IndexOutOfBoundsException` instead of the friendly Camel error with bean/OGNL context
- Null-safe OGNL (`?` operator) was also broken at this boundary — threw instead of returning `null`

Fixes: https://issues.apache.org/jira/browse/CAMEL-24101

## Test plan

- [x] New `testBodyOGNLOrderListOutOfBoundsAtBoundary` — verifies friendly error at index == size
- [x] New `testBodyOGNLOrderListOutOfBoundsAtBoundaryWithNullSafe` — verifies null-safe returns null at boundary
- [x] All existing out-of-bounds tests (index > size) continue to pass

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/language/bean/BeanExpression.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/language/bean/BeanExpression.java
@@ -552,7 +552,7 @@ public class BeanExpression implements Expression, Predicate {
                         }
                     }
                 }
-                if (num != null && num >= 0 && !list.isEmpty() && list.size() > num - 1) {
+                if (num != null && num >= 0 && !list.isEmpty() && list.size() > num) {
                     return list.get(num);
                 }
                 if (!nullSafe) {
--- a/core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/language/simple/SimpleTest.java
@@ -1372,6 +1372,35 @@ public class SimpleTest extends LanguageTestSupport {
     }
 
     @Test
+    public void testBodyOGNLOrderListOutOfBoundsAtBoundary() {
+        List<OrderLine> lines = new ArrayList<>();
+        lines.add(new OrderLine(123, "Camel in Action"));
+        lines.add(new OrderLine(456, "ActiveMQ in Action"));
+        Order order = new Order(lines);
+
+        exchange.getIn().setBody(order);
+
+        RuntimeBeanExpressionException e = assertThrows(RuntimeBeanExpressionException.class,
+                () -> assertExpression("${in.body.getLines[2].getId}", 123),
+                "Should have thrown an exception");
+
+        IndexOutOfBoundsException cause = assertIsInstanceOf(IndexOutOfBoundsException.class, e.getCause());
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Clean fix for a classic off-by-one boundary error in OGNL list index access.

**The bug:** The guard `list.size() > num - 1` allowed access to `list.get(num)` when `num == list.size()`, which is out of bounds. For a 2-element list (indices 0, 1), accessing index 2:
- Old: `2 > 2 - 1` → `2 > 1` → `true` → `list.get(2)` → 💥 `IndexOutOfBoundsException`
- New: `2 > 2` → `false` → properly falls through to null-safe/exception path

**The fix:** `list.size() > num` is the correct 0-based index boundary check — equivalent to `num < list.size()`, the standard idiom for valid list access.

**Tests are well-designed:**
- `testBodyOGNLOrderListOutOfBoundsAtBoundary` — verifies accessing index `[2]` on a 2-element list throws `IndexOutOfBoundsException` with a descriptive message
- `testBodyOGNLOrderListOutOfBoundsAtBoundaryWithNullSafe` — verifies the null-safe variant (`?.`) returns `null` instead of throwing

**Checklist:**
- ✅ One-line fix, correct boundary check
- ✅ Two boundary tests covering both null-safe and non-null-safe paths

oscerd (changes requested):
Thanks for fixing the boundary off-by-one — the change to `list.size() > num` (from `> num - 1`) is correct: it stops `num == size` slipping into the direct-index block, so the boundary case now returns the friendly Camel `IndexOutOfBoundsException` instead of the raw JDK one, and no previously-valid index regresses.

The reason `build (25)` is red (in camel-core, not an infra flake) is that this message change breaks four pre-existing `SimpleTest` assertions that still expect the old raw-JDK text:

- `SimpleTest.java:446` (testOGNLPropertyList)
- `SimpleTest.java:466` (testOGNLPropertyLinesList)
- `SimpleTest.java:964` (testOGNLHeaderList)
- `SimpleTest.java:983` (testOGNLHeaderLinesList)

Each does `assertEquals(INDEX_OUT_OF_BOUNDS_ERROR_MSG, cause.getMessage())` where `INDEX_OUT_OF_BOUNDS_ERROR_MSG = "Index 2 out of bounds for length 2"` (SimpleTest.java:74) — the exact index-2-on-length-2 boundary you're re-routing. CI shows `expected: <Index 2 out of bounds for length 2> but was: <Index: 2, Size: 2 out of bounds with List from bean: ...>`. Your two new tests correctly use `startsWith(...)`; the four legacy ones need the same treatment (the new message embeds the bean toString, so `assertEquals` on the old constant can't hold). Updating those four should turn CI green — the fix itself is good.

davsclaus:
@oscerd Good catch — updated the four `SimpleTest` assertions to use `startsWith("Index: 2, Size: 2 out of bounds with List from bean")` instead of `assertEquals` with the old JDK message, and removed the unused `INDEX_OUT_OF_BOUNDS_ERROR_MSG` constant. All 202 SimpleTest cases pass.

PR #24764, opened 2026-07-16
CAMEL-24102: camel-bean - AmbiguousMethodCallException lists actual candidates

davsclaus:
## Summary

- Fix `AmbiguousMethodCallException` in body-conversion matching to list the actual distinct method candidates instead of the same method listed twice
- Collect convertible candidates in a list during the matching loop instead of only keeping the last match

Fixes: https://issues.apache.org/jira/browse/CAMEL-24102

## Test plan

- [x] `BeanAmbiguousBodyConversionTest` — verifies the exception lists 2 distinct candidates
- [x] All existing ambiguous method tests pass (6 tests)

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.bean;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+import org.apache.camel.CamelExecutionException;
+import org.apache.camel.ContextTestSupport;
+import org.apache.camel.builder.RouteBuilder;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class BeanAmbiguousBodyConversionTest extends ContextTestSupport {
+
+    @Test
+    public void testAmbiguousExceptionListsDistinctCandidates() {
+        CamelExecutionException e = assertThrows(CamelExecutionException.class,
+                () -> template.sendBody("direct:start", "123"));
+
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Great catch — the old code's error message was actively misleading.

**The bug:** When body type conversion matched multiple methods (ambiguous call), the code incremented a counter but only kept the *last* matched `MethodInfo`. Then it threw `AmbiguousMethodCallException(exchange, Arrays.asList(matched, matched))` — listing the **same method reference twice**. A developer debugging an ambiguous method call would see the same method listed twice in the exception, making it impossible to identify the actual conflicting candidates.

**The fix:** Replaces the counter with `List<MethodInfo> candidates` that collects each matching `MethodInfo`. The exception now receives the actual distinct candidates, producing a useful error message like "Ambiguous method call: foo(Integer) vs bar(Double)".

**Test is well-designed:**
- Uses a `MyAmbiguousBean` with `foo(Integer)` and `bar(Double)` — sending a `String` body triggers type conversion ambiguity
- Asserts the exception contains exactly 2 candidates
- Verifies the candidates are **different instances** (`list.get(0) != list.get(1)`) — directly testing the fix

**Checklist:**
- ✅ Fix is correct — collects actual candidates instead of duplicating the last match
- ✅ Test verifies the exception contains distinct method candidates
- ✅ No public API change

core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @Test
+    public void testAmbiguousExceptionListsDistinctCandidates() {
+        CamelExecutionException e = assertThrows(CamelExecutionException.class,
+                () -> template.sendBody("direct:start", "123"));
+
+        AmbiguousMethodCallException cause = assertInstanceOf(AmbiguousMethodCallException.class, e.getCause());
+        Collection<MethodInfo> methods = cause.getMethods();
+
+        assertEquals(2, methods.size(), "Should list exactly two distinct candidates");
+        ArrayList<MethodInfo> list = new ArrayList<>(methods);
```
apupier:
```suggestion
        List<MethodInfo> list = new ArrayList<>(methods);
```

core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertTrue(list.get(0) != list.get(1),
+                "The two candidates should be different method instances");
```
apupier: maybe can use AssertJ.assertThat().doesNotHaveDuplicates()

it will give a better error message of what is inside the list in case of failure

apupier (approved):

oscerd (approved):
LGTM — nice catch. The exception was built with `Arrays.asList(matched, matched)` (the same reference twice), so `AmbiguousMethodCallException` never actually listed the *distinct* conflicting methods; building a real `List<MethodInfo> candidates` means it now names both. The test drives an ambiguous String→Integer/Double conversion and asserts two distinct candidates.

core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
davsclaus: Applied — went a step further and removed the `list` variable entirely since `doesNotHaveDuplicates()` works directly on the `methods` collection. Thanks!

core/camel-core/src/test/java/org/apache/camel/component/bean/BeanAmbiguousBodyConversionTest.java
davsclaus: Good suggestion — switched to AssertJ `assertThat(methods).doesNotHaveDuplicates()` which produces a much clearer failure message showing the actual list contents. Thanks!

PR #24765, opened 2026-07-16
CAMEL-24103: camel-bean - Descriptive error for null CompletionStage return

davsclaus:
## Summary

- Add null-check before calling `.whenComplete()` on a bean method's `CompletionStage` return value
- When null, throws a `RuntimeCamelException` naming the bean class and method instead of an opaque NPE

Fixes: https://issues.apache.org/jira/browse/CAMEL-24103

## Test plan

- [x] `BeanNullCompletionStageTest` — verifies the descriptive error message includes method name and "null CompletionStage"

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/component/bean/MethodInfo.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/component/bean/MethodInfo.java
@@ -332,6 +332,11 @@ public class MethodInfo {
 
                 //If it's Java 8 async result
                 if (CompletionStage.class.isAssignableFrom(method.getReturnType())) {
+                    if (result == null) {
+                        throw new RuntimeCamelException(
+                                "Bean method " + method.getDeclaringClass().getName() + "." + method.getName()
+                                                        + " returned null CompletionStage");
+                    }
                     CompletionStage<?> completionStage = (CompletionStage<?>) result;
 
                     completionStage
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, focused fix. When a bean method declares `CompletionStage` as its return type but returns `null` at runtime, the current code enters the `CompletionStage` handling block (line 334, based on `method.getReturnType()` — not the actual result), the cast `(CompletionStage<?>) null` succeeds silently, and then `completionStage.whenComplete(...)` throws an unhelpful `NullPointerException`.

The null guard catches this early and throws a descriptive `RuntimeCamelException` that includes both the class name and method name — much easier to diagnose.

**Review:**

- ✅ Null check correctly placed between the `isAssignableFrom` guard and the cast
- ✅ Error message includes both declaring class and method name for easy debugging
- ✅ Test covers the exact scenario: bean method declared as returning `CompletionStage<String>` but returning `null`
- ✅ Test assertions verify both the exception type chain (`CamelExecutionException` → `RuntimeCamelException`) and the message content
- ✅ No new static analysis findings introduced

oscerd (approved):
LGTM. The null-guard is placed exactly right — between the `CompletionStage`-assignability type check and the unchecked `(CompletionStage<?>) result` cast in `MethodInfo` — so a bean method declared to return a `CompletionStage` but actually returning `null` now yields a descriptive error ("...returned null CompletionStage") instead of an opaque NPE from `whenComplete()`. Test asserts both the message and the offending method name.

PR #24766, opened 2026-07-16
CAMEL-24104: camel-bean - Fix racy Processor-adapter lookup in AbstractBeanProcessor

davsclaus:
## Summary

- Make `processor` and `lookupProcessorDone` fields `volatile` so racing threads see consistent values
- Add double-check of `lookupProcessorDone` after acquiring the lock to prevent duplicate conversions
- Move `lookupProcessorDone = true` after the conversion succeeds so a conversion failure doesn't permanently disable the lookup

Fixes: https://issues.apache.org/jira/browse/CAMEL-24104

## Test plan

- [x] All 346 existing bean component tests pass — no regressions
- No new test: the visibility race is non-deterministic and not reliably reproducible in a unit test

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/component/bean/AbstractBeanProcessor.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/component/bean/AbstractBeanProcessor.java
@@ -37,9 +37,9 @@ public abstract class AbstractBeanProcessor extends AsyncProcessorSupport {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractBeanProcessor.class);
 
     private final BeanHolder beanHolder;
-    private transient Processor processor;
+    private transient volatile Processor processor;
     private transient Object bean;
-    private transient boolean lookupProcessorDone;
+    private transient volatile boolean lookupProcessorDone;
     private final Lock lock = new ReentrantLock();
     private BeanScope scope;
     private String method;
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Textbook double-checked locking fix — three issues addressed:

1. **`volatile` added** to both `processor` and `lookupProcessorDone` — ensures reads outside the lock see the latest values (Java Memory Model guarantees happens-before for volatile writes → volatile reads).

2. **Inner double-check** added inside the lock — without it, thread B that lost the race to `lock.lock()` would re-execute the type converter lookup. The new `else { target = processor; }` branch correctly reads the already-computed value for the losing thread.

3. **Flag order corrected** — `lookupProcessorDone = true` now executes AFTER `processor = target`. Combined with volatile, this guarantees any thread that reads `lookupProcessorDone == true` (either at the outer check or after `getProcessor()` returns non-null) will also see the correct `processor` value.

**Note:** The race predates this PR — it was present in the original `synchronized` version too (the outer `if (!lookupProcessorDone)` check without synchronization could read a stale non-volatile field). The CAMEL-20199 migration to `ReentrantLock` preserved the bug. This PR properly fixes it.

**Review:**

- ✅ Standard double-checked locking with volatile — correct and well-understood pattern
- ✅ `getProcessor()` at the top of `getCustomAdapter()` reads the now-volatile `processor` field, so subsequent callers after the lookup is done take the fast path without entering the lock at all
- ✅ Static analysis: only pre-existing broad-exception-catch findings, no new issues

PR #24767, opened 2026-07-16
CAMEL-24105: camel-bean - Do not register annotation-driven EIP processors as CamelContext services

davsclaus:
## Summary

The `MethodInfo` constructor registered `@RoutingSlip`, `@RecipientList`, and `@DynamicRouter` annotation-driven EIP processors as CamelContext services via `camelContext.addService()`. These services were never removed when the owning `BeanInfo` was evicted from the LRU cache in `BeanComponent`, causing unbounded growth of the context's internal services list over long uptimes.

The fix removes the three `addService` calls. The processors are already started lazily on first use (checked via `ServiceHelper.isStarted()` before each invocation), so the `addService` registration was redundant — it only served to pin stale processor instances in the services list.

- Removed `camelContext.addService(routingSlip/dynamicRouter/recipientList)` from the `MethodInfo` constructor
- Added `BeanAnnotationServiceLeakTest` verifying that `@RoutingSlip` processors are not registered as CamelContext services

## Test plan

- [x] New `BeanAnnotationServiceLeakTest` verifies no service leak
- [x] Existing `RoutingSlipPOJOTest` passes (annotation-based routing slip still works)
- [x] Existing `AsyncEndpointRoutingSlipBean*Test` tests pass
- [x] Existing `BeanRecipientListTest`, `BeanRecipientListInterfaceAnnotationTest` pass
- [x] Existing `DynamicRouterAnnotationTest` passes

### DIFF
--- a/components/camel-bean/src/main/java/org/apache/camel/component/bean/MethodInfo.java
+++ b/components/camel-bean/src/main/java/org/apache/camel/component/bean/MethodInfo.java
@@ -139,12 +139,6 @@ public class MethodInfo {
         if (routingSlipAnnotation != null) {
             routingSlip = PluginHelper.getAnnotationBasedProcessorFactory(camelContext)
                     .createRoutingSlip(camelContext, routingSlipAnnotation);
-            // add created routingSlip as a service so we have its lifecycle managed
-            try {
-                camelContext.addService(routingSlip);
-            } catch (Exception e) {
-                throw RuntimeCamelException.wrapRuntimeCamelException(e);
-            }
         }
 
         DynamicRouter dynamicRouterAnnotation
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean fix for a service leak. The three annotation-driven EIP processors (`@RoutingSlip`, `@DynamicRouter`, `@RecipientList`) were registered as CamelContext services in the `MethodInfo` constructor. Since `MethodInfo` instances are created per cache entry in `BeanInfo`, evicting and re-creating cache entries would register new services without removing the old ones — an unbounded leak.

The `addService()` calls were introduced during the CAMEL-13449 migration (moving camel-bean out of camel-core) but were redundant from the start: all three processors already have lazy-start guards in their invocation methods:

```java
if (!ServiceHelper.isStarted(routingSlip)) {
    ServiceHelper.startService(routingSlip);
}
```

So removing the context-level registration doesn't break startup lifecycle — processors are started on first use.

**Review:**

- ✅ Consistent removal across all three annotations (`@RoutingSlip`, `@DynamicRouter`, `@RecipientList`)
- ✅ Test verifies no `RoutingSlip` processor is registered as a CamelContext service after routing
- ✅ Lazy-start pattern in `routingSlipInvocation()`, `recipientListInvocation()`, and `dynamicRouterInvocation()` ensures correct startup lifecycle without context registration
- ✅ No new static analysis findings — only pre-existing broad-exception-catch and mutable-collection-return findings

PR #24769, opened 2026-07-16
CAMEL-24127: Fix flaky tests (batch 15)

gnodet:
## Summary

Fix 5 flaky tests identified by [Develocity](https://develocity.apache.org):

### 1. ElasticSearchLocalContainerInfraService — disk watermark 429 errors
- **Root cause**: CI machines with >95% disk usage trigger Elasticsearch flood-stage watermark, blocking all index writes with HTTP 429.
- **Fix**: Disable disk watermarks via `cluster.routing.allocation.disk.threshold_enabled=false` (same fix as OpenSearch in CAMEL-24018).

### 2. LoopNoBreakOnShutdownTest — SEDA shutdown race condition
- **Root cause**: Race between SEDA `queue.poll()` (which drains the queue, making it appear empty) and the inflight repository registration. The shutdown strategy could see 0 inflight exchanges and proceed with immediate shutdown before the LoopProcessor had a chance to register.
- **Fix**: Wait for at least 1 loop iteration to complete at `mock:result` before calling `context.stop()`. This guarantees the exchange is registered in the inflight repository and the LoopProcessor's `taskCount > 0`.

### 3. HazelcastReplicatedmapConsumerTest.testRemove — leaked REMOVED events
- **Root cause**: `ReplicatedMap.clear()` in `resetState()` fires individual `entryRemoved()` events (unlike `IMap.clear()` which fires `mapCleared`). These events are dispatched asynchronously from Hazelcast's event thread pool and can arrive after `resetState()` returns, bleeding into the next test's mock expectations. Only `testRemove` is affected because only `mock:removed` can receive stale events from `clear()`.
- **Fix**: Wait for the ADDED event on key "4711" (leveraging Hazelcast's per-key event ordering guarantee) then reset `mock:removed` before performing the test's own `remove()`.

### 4. LRAFailuresIT.testCompletionAfterFailures — periodic recovery timing (78% flaky)
- **Root cause**: The Narayana LRA coordinator retries failed completion callbacks via its periodic recovery manager (default period: 120s). The `JAVA_TOOL_OPTIONS` env var attempts to shorten this to 2s via Arjuna system properties, but these may not be honored by the Quarkus Narayana extension's config mapping. With the default 120s period and the old 60s test timeout, the test fails whenever the retry doesn't happen in time.
- **Fix**: Instead of waiting passively, explicitly trigger the coordinator's `/lra-coordinator/recovery` REST endpoint (documented in [Narayana docs](https://www.narayana.io/docs/project/index.html)) to force an immediate recovery scan. This makes the test deterministic regardless of the periodic timer configuration.
- **Bonus**: Fix `WaitAllStrategy` bug in `MicroprofileLRALocalContainerInfraService` where the second `waitingFor()` call silently replaced the first.

### 5. TimerRouteAutoConfigIT — timing-dependent metric collection
- **Root cause**: The test used a chain of `PeriodicMetricReader` → `LoggingMetricExporter` → JUL `Logger` → `MemoryLogHandler`, which is inherently timing-dependent. The periodic reader might not have flushed metrics to the logger by the time the test checks.
- **Fix**: Replace with `InMemoryMetricReader` for synchronous, deterministic metric collection. The test calls `metricReader.collectAllMetrics()` directly after `sendBody` — no timing dependency.

## Test plan
- [x] `LoopNoBreakOnShutdownTest` passes locally
- [x] `HazelcastReplicatedmapConsumerTest` passes locally (all 3 methods)
- [x] `TimerRouteAutoConfigIT` passes locally
- [ ] `ElasticSearchLocalContainerInfraService` — infrastructure-level fix, validated by CI
- [ ] `LRAFailuresIT` — requires LRA coordinator container, validated by CI
- [ ] CI passes on this PR

### DIFF
--- a/components/camel-hazelcast/src/test/java/org/apache/camel/component/hazelcast/HazelcastReplicatedmapConsumerTest.java
+++ b/components/camel-hazelcast/src/test/java/org/apache/camel/component/hazelcast/HazelcastReplicatedmapConsumerTest.java
@@ -102,16 +102,21 @@ public class HazelcastReplicatedmapConsumerTest extends CamelTestSupport {
 
     @Test
     public void testRemove() throws InterruptedException {
-        MockEndpoint out = getMockEndpoint("mock:removed");
-        out.expectedMessageCount(1);
-
         map.put("4711", "my-foo");
 
-        // Wait for the ADDED event to be fully processed before removing,
-        // otherwise the remove may execute before the async ADDED event is delivered.
+        // Wait for the ADDED event to be fully processed. Since events for
+        // key "4711" are dispatched in partition order within Hazelcast's
+        // event system, any stale REMOVED events from resetState()'s clear()
+        // are guaranteed to have been delivered by the time we see the ADDED.
         MockEndpoint added = getMockEndpoint("mock:added");
         Awaitility.await().atMost(10, TimeUnit.SECONDS)
                 .until(() -> added.getReceivedCounter() >= 1);
+
+        // Reset mock:removed to discard any stale REMOVED events that leaked
+        // from clear(), then set the expectation for the test's own remove.
+        MockEndpoint out = getMockEndpoint("mock:removed");
+        out.reset();
+        out.expectedMessageCount(1);
 
         map.remove("4711");
         MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS);
--- a/components/camel-lra/src/test/java/org/apache/camel/service/lra/AbstractLRATestSupport.java
+++ b/components/camel-lra/src/test/java/org/apache/camel/service/lra/AbstractLRATestSupport.java
@@ -36,7 +36,6 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.awaitility.Awaitility.await;
-import static org.hamcrest.Matchers.equalTo;
 
 /**
  * Base class for LRA based tests.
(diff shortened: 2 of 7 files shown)


davsclaus (approved):

oscerd (approved):
LGTM — these are genuine determinism fixes, not just longer waits (no `Thread.sleep` added, and the assertions are preserved):

- **Hazelcast**: resets `mock:removed` only after the ADDED event on key 4711 is delivered, discarding stale REMOVED events leaked from `resetState()`'s `clear()`.
- **LRA** (both files): replaces the passive 20s/60s wait with an active `triggerRecovery()` against `/lra-coordinator/recovery` on a 2s poll — deterministic regardless of the 120s periodic timer.
- **TimerRouteAutoConfigIT**: swaps the PeriodicMetricReader + log-capture for a synchronous `InMemoryMetricReader.collectAllMetrics()`, keeping the same HistogramPointData / count / min assertions.
- **LoopNoBreakOnShutdownTest**: waits for `mock:result >= 1` (guaranteeing inflight registration + taskCount>0) instead of `future.isDone()`.
- **Infra**: ES disk-watermark disable, plus a real `WaitAllStrategy` bug fix where a second `waitingFor()` had silently replaced the first.

Nice batch.

davsclaus (approved):

gnodet (commented):
LGTM — five well-analyzed flaky test fixes. Notes on each:

**1. ElasticSearchLocalContainerInfraService** — `cluster.routing.allocation.disk.threshold_enabled=false`
mirrors the OpenSearch fix (CAMEL-24018). Clean.

**2. LoopNoBreakOnShutdownTest** — The old code waited for the SEDA enqueue future, but that
doesn't guarantee the exchange is registered in the inflight repository when `context.stop()`
runs. Waiting for `mock.getReceivedCounter() >= 1` ensures the LoopProcessor has started
iterating, so the shutdown strategy properly waits for all 100 iterations.

**3. HazelcastReplicatedmapConsumerTest.testRemove** — Smart use of Hazelcast's partition-ordered
event delivery: waiting for the ADDED event guarantees any stale REMOVED events from
`resetState()`'s `clear()` have also been delivered. Resetting `mock:removed` after that
clears the slate for the test's own `map.remove()`.

**4. LRA — triggerRecovery() via REST** — Much better than waiting for the 120s periodic scan.
Also a good catch on the `WaitAllStrategy` bug — the second `.waitingFor()` was silently
replacing the first, so `Wait.forListeningPort()` was never checked. Combining them with
`WaitAllStrategy` is the correct Testcontainers pattern.

**5. TimerRouteAutoConfigIT** — Replacing the PeriodicMetricReader → LoggingMetricExporter → JUL
log capture chain with `InMemoryMetricReader.collectAllMetrics()` eliminates all timing
dependency. Much cleaner — deterministic synchronous collection instead of polling log output.

PR #24772, opened 2026-07-16
CAMEL-24106: camel-xslt - source option silently ignored when body is InputStream

davsclaus:
## Summary

`XmlSourceHandlerFactoryImpl.getSource(Exchange, Expression)` checked `isInputStreamNeeded(exchange)` against the **message body** before evaluating the configured `source` expression. When the body was an `InputStream`, `StreamCache`, or any type without a direct `Source` converter — which is the common case with `file`, `http`, `ftp`, `netty` consumers or stream caching — `source=header:x` / `source=variable:x` / `source=property:x` silently transformed the **body** instead of the configured source. No error, wrong output.

### Fix

Evaluate the `source` expression **first** when it is configured, and pass the result directly to the type-based conversion method. The `isInputStreamNeeded` optimization is only applied when no source expression is configured (i.e. the body is the intended input).

This follows the same pattern already used by `JsonSourceHandlerFactoryImpl` in `camel-xj`, which correctly evaluates the `source` expression before handling by type.

### Affected components

- `xslt` — fix in `XmlSourceHandlerFactoryImpl`
- `xslt-saxon` — inherits the fix (does not override `getSource(Exchange, Expression)`)
- `xj` (XML→JSON) — **not affected** (already correct)

### Tests

Added `XsltSaxonSourceInputStreamBodyTest` with 5 tests covering:
- `source=header` with `InputStream` body
- `source=variable` with `InputStream` body
- `source=property` with `InputStream` body
- `source=header` with stream caching enabled (body wrapped as `StreamCache`)
- `source=variable` with stream caching enabled

### DIFF
--- a/components/camel-xslt-saxon/src/test/java/org/apache/camel/component/xslt/saxon/XsltSaxonSourceInputStreamBodyTest.java
+++ b/components/camel-xslt-saxon/src/test/java/org/apache/camel/component/xslt/saxon/XsltSaxonSourceInputStreamBodyTest.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.xslt.saxon;
+
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Nice catch and clean fix.**

The bug is clear: `isInputStreamNeeded(exchange)` short-circuits to transform the **body** before the `source` expression is ever evaluated. When the body is an `InputStream` (common with `file:`, `http:`, `ftp:`, or stream caching), `source=header:x` / `source=variable:x` / `source=property:x` is silently ignored — no error, wrong output.

The fix correctly evaluates the `source` expression **first** when configured, and only falls through to the `isInputStreamNeeded` optimization when no source expression is present. This aligns `XmlSourceHandlerFactoryImpl` with `JsonSourceHandlerFactoryImpl` in `camel-xj`, which already handles this correctly.

**Checklist:**
- ✅ Tests: 5 new tests covering header/variable/property sources with both raw InputStream and StreamCache bodies
- ✅ Test resource (`xslt/transform.xsl`) already exists
- ✅ No public API changes
- ✅ Backward compatible — no-source-expression code path unchanged
- ✅ No static analysis findings (ast-grep)
- ⏳ CI: builds in progress

oscerd (approved):
LGTM. On main `getSource(Exchange, Expression)` calls `isInputStreamNeeded(exchange)` (which inspects the *body*) before ever looking at the `source` expression, so an InputStream/StreamCache body short-circuited and the `source` option was silently dropped. Evaluating `source` first fixes exactly that, and it's well-scoped: the `source == null` path is byte-identical to before, and when `source` is set the body stream is never read (no double-consume), with the result flowing through `getSource(exchange, Object)`'s SAX→Stream→DOM conversion — same pattern as camel-xj. Nice touch that camel-xslt-saxon only overrides the `(Exchange, Object)` overload, so it inherits the fix. Good test coverage (header/variable/property + two stream-caching cases).

PR #24774, opened 2026-07-16
camel-lumberjack: fix flaky tests by awaiting Netty shutdown

davsclaus:
## Summary

Fix flaky lumberjack tests caused by incomplete Netty resource cleanup between test runs:

- **Await `shutdownGracefully()` in `LumberjackServer.stop()`**: the three Netty thread groups (`bossGroup`, `workerGroup`, `executorService`) were shut down fire-and-forget, so the server port could still be in use when the next `@Isolated` test tried to bind. Now awaits completion with `syncUninterruptibly()`.
- **Await client channel close and event loop in `LumberjackUtil`**: `channel.close()` was fire-and-forget and `shutdownGracefully()` on the client event loop was not awaited, which could leave half-open connections that cause spurious ack count mismatches on the server side.
- **Thread-safe `ErrorProcessor.count`**: the `int count` field in `LumberjackDisconnectionTest.ErrorProcessor` was read/written from Netty executor threads without synchronization. Replaced with `AtomicInteger`.
- **Explicit Awaitility timeout**: the `Awaitility.await()` call in `LumberjackUtil.sendMessages()` relied on the 10-second default, which can be tight on slow CI nodes (especially for SSL tests). Now uses an explicit 30-second timeout.

## Test plan

- [x] `mvn test` in `components/camel-lumberjack` — all 5 tests pass
- [ ] CI validates no regressions across platforms

### DIFF
--- a/components/camel-lumberjack/src/test/java/org/apache/camel/component/lumberjack/LumberjackUtil.java
+++ b/components/camel-lumberjack/src/test/java/org/apache/camel/component/lumberjack/LumberjackUtil.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLEngine;
 
@@ -94,16 +95,16 @@ final class LumberjackUtil {
             // send 5 frame windows, without pausing
             windows.stream().forEach(window -> channel.writeAndFlush(readSample(String.format("io/window%s.bin", window))));
             if (waitForResult) {
-                Awaitility.await().until(() -> windows.size() == responses.size());
+                Awaitility.await().atMost(30, TimeUnit.SECONDS).until(() -> windows.size() == responses.size());
             }
 
-            channel.close();
+            channel.close().sync();
 
             synchronized (responses) {
                 return responses;
             }
         } finally {
-            eventLoopGroup.shutdownGracefully();
+            eventLoopGroup.shutdownGracefully().syncUninterruptibly();
         }
     }
 
--- a/components/camel-lumberjack/src/main/java/org/apache/camel/component/lumberjack/io/LumberjackServer.java
+++ b/components/camel-lumberjack/src/main/java/org/apache/camel/component/lumberjack/io/LumberjackServer.java
@@ -107,9 +107,11 @@ public final class LumberjackServer {
             // Wait for the channel to be indeed closed before shutting the groups & service
             channel.close().sync();
         } finally {
-            bossGroup.shutdownGracefully();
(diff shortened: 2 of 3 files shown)


gnodet (approved):
**Solid flaky-test fix — all four changes are well-motivated.**

1. **`syncUninterruptibly()` on Netty shutdown** (`LumberjackServer.stop()`) — fire-and-forget `shutdownGracefully()` is the classic cause of port-still-in-use between `@Isolated` tests. Awaiting completion is the standard Netty pattern.

2. **`channel.close().sync()`** (`LumberjackUtil`) — ensures the client channel is fully closed before proceeding, preventing half-open connections that cause ack count mismatches.

3. **`AtomicInteger` for `ErrorProcessor.count`** — the processor runs on Netty executor threads, so the bare `int` was a legitimate race condition. Clean fix.

4. **Explicit 30s Awaitility timeout** — follows the project convention of always setting an explicit `atMost` rather than relying on the 10s default. 30s is reasonable for SSL tests on slower CI nodes.

**Static analysis note:** ast-grep flagged a pre-existing `ByteArrayOutputStream` without try-with-resources in `LumberjackUtil.readSample()` (line 112) — not introduced by this PR.

**Checklist:**
- ✅ Tests: existing tests cover the scenarios; the fix is in resource cleanup, not logic
- ✅ No public API changes
- ✅ Backward compatible — shutdown is just properly awaited now
- ✅ No security concerns
- ⏳ CI: builds in progress

components/camel-lumberjack/src/test/java/org/apache/camel/component/lumberjack/LumberjackUtil.java
```diff
@@ -94,16 +95,16 @@ protected void channelRead0(ChannelHandlerContext ctx, ByteBuf msg) {
             // send 5 frame windows, without pausing
             windows.stream().forEach(window -> channel.writeAndFlush(readSample(String.format("io/window%s.bin", window))));
             if (waitForResult) {
-                Awaitility.await().until(() -> windows.size() == responses.size());
+                Awaitility.await().atMost(30, TimeUnit.SECONDS).until(() -> windows.size() == responses.size());
```
apupier, line 98: I'm not sure that it is useful given that when things are working the test is taking less than 300 ms

https://develocity.apache.org/scans/tests?search.relativeStartTime=P90D&search.rootProjectNames=camel&search.timeZoneId=Europe%2FParis&tests.container=org.apache.camel.component.lumberjack.LumberjackComponentTest

https://develocity.apache.org/scans/tests?search.relativeStartTime=P90D&search.rootProjectNames=camel&search.timeZoneId=Europe%2FParis&tests.container=org.apache.camel.component.lumberjack.LumberjackDisconnectionTest

apupier (approved):

PR #24776, opened 2026-07-16
CAMEL-24107: camel-xslt - stylesheet reload is not thread-safe, stale pooled transformers keep serving the old stylesheet

davsclaus:
## Summary

Fixes thread-safety issues in camel-xslt stylesheet reload (`contentCache=false` / `clearCachedStylesheet` JMX operation) where stale pooled transformers keep serving the old stylesheet after a reload.

### Problems fixed

1. **Stale pooled transformers survive reload** — with `transformerCacheSize > 0`, a `Transformer` borrowed before `setTemplate()` is re-offered to the pool after the transform completes, causing the old stylesheet to be served indefinitely
2. **`XsltBuilder.template` not volatile** — plain field with no visibility guarantee across threads
3. **Shared `TransformerFactory` used concurrently** — `setTransformerSource()` mutates error listener/URI resolver and calls `newTemplates()` without synchronization; JAXP does not guarantee `TransformerFactory` thread safety
4. **Redundant parallel reloads** — `cacheCleared` flag reset only after full load; concurrent messages trigger redundant stylesheet compilations

### Changes

- **`XsltBuilder`**: make `template` volatile; add `AtomicLong templateGeneration` counter incremented on every `setTemplate()` call; in `releaseTransformer()`, discard the transformer if its generation doesn't match current (prevents stale re-entry); wrap `setTransformerSource()` with a `ReentrantLock` to serialize `TransformerFactory` access
- **`XsltEndpoint`**: add `ReentrantLock reloadLock` with double-checked locking around `loadResource()` in `onExchange()` to prevent redundant parallel reloads

## Test plan

- [x] New `XsltBuilderReloadTest` with 3 tests: basic pool invalidation on template change, multiple reloads, concurrent processing after reload
- [x] All existing XSLT tests pass (`XsltBuilderTest`, `XsltTest`, `XsltRouteTest`)
- [x] Full `camel-core` test suite passes

### DIFF
--- a/components/camel-xslt/src/main/java/org/apache/camel/component/xslt/XsltBuilder.java
+++ b/components/camel-xslt/src/main/java/org/apache/camel/component/xslt/XsltBuilder.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -66,7 +67,8 @@ public class XsltBuilder implements Processor {
 
     protected static final Logger LOG = LoggerFactory.getLogger(XsltBuilder.class);
     private Map<String, Object> parameters = new HashMap<>();
-    private Templates template;
+    private volatile Templates template;
+    private final AtomicLong templateGeneration = new AtomicLong();
     private volatile BlockingQueue<Transformer> transformers;
     private volatile SourceHandlerFactory sourceHandlerFactory;
     private ResultHandlerFactory resultHandlerFactory = new StringResultHandlerFactory();
@@ -80,6 +82,7 @@ public class XsltBuilder implements Processor {
 
     private final XMLConverterHelper converter = new XMLConverterHelper();
     private final Lock sourceHandlerFactoryLock = new ReentrantLock();
+    private final Lock transformerSourceLock = new ReentrantLock();
 
     public XsltBuilder() {
     }
@@ -103,6 +106,7 @@ public class XsltBuilder implements Processor {
             exchange.getExchangeExtension().addOnCompletion(new XsltBuilderOnCompletion(fileName));
         }
 
+        long gen = templateGeneration.get();
         Transformer transformer = getTransformer();
         configureTransformer(transformer, exchange);
 
@@ -128,7 +132,7 @@ public class XsltBuilder implements Processor {
             LOG.trace("Transform complete with result {}", result);
             resultHandler.setBody(out);
         } finally {
-            releaseTransformer(transformer);
+            releaseTransformer(transformer, gen);
             // IOHelper can handle if null
             IOHelper.close(is);
         }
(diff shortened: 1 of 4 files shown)


gnodet (approved):
**Thorough concurrency fix — well-analyzed and well-tested.**

The four problems identified are all real:

1. **`volatile template`** — correct. This field is read by `process()` threads and written by `setTemplate()`. Without `volatile`, JMM allows stale reads.

2. **Generation-based pool invalidation** (`AtomicLong templateGeneration`) — elegant approach. Capturing the generation before borrowing and checking on release ensures stale transformers are discarded rather than re-pooled. The `setTemplate()` → `clear()` combination provides belt-and-suspenders: `clear()` evicts idle transformers immediately, the generation check catches in-flight ones.

3. **`transformerSourceLock`** around `setTransformerSource()` — correct. JAXP `TransformerFactory` does not guarantee thread safety, and this method mutates error listener + URI resolver + calls `newTemplates()`. Serializing access is the right call.

4. **DCL `reloadLock`** in `XsltEndpoint.onExchange()` — prevents redundant parallel reloads when `cacheCleared` is set (JMX operation). Verified that `loadResource()` resets `cacheCleared = false` (line 397) inside the lock, so the inner check correctly prevents duplicate reloads.

**Tests are solid** — three scenarios covering basic pool invalidation, multiple reloads (back-and-forth), and concurrent processing with 8 threads after reload. The `example2.xsl` (`<farewell>`) vs `example.xsl` (`<goodbye>`) makes it easy to distinguish which template was used.

**Static analysis:** ast-grep flagged two pre-existing `getParameters()` mutable collection return issues — not introduced by this PR.

**Checklist:**
- ✅ Tests: 3 new tests covering pool invalidation, multi-reload, and concurrency
- ✅ No public API changes (internal `releaseTransformer` signature change is private)
- ✅ Backward compatible — behavior unchanged for single-threaded / non-reload paths
- ✅ No security concerns
- ⏳ CI: builds in progress

oscerd (commented):
Good fix — this is a real improvement over the status quo, where `releaseTransformer()` unconditionally re-pooled a borrowed `Transformer` even if it was borrowed before a `setTemplate()`, so the pool kept serving the *old* compiled stylesheet indefinitely after a reload. Stamping a borrow-time `templateGeneration` and discarding on mismatch, making `template` volatile, and DCL-guarding the reload in `XsltEndpoint` all look correct, and the latch-based tests avoid `Thread.sleep`.

One residual window worth a look, since this is squarely a thread-safety fix: `setTemplate()` does `templateGeneration.incrementAndGet()` and `transformers.clear()` as two separate steps, while `process()` captures `gen = templateGeneration.get()` *before* polling. Interleaving: a reload bumps gen 0→1, a concurrent `process()` reads gen=1, then `poll()` hands back an old (gen-0) transformer that `clear()` hasn't removed yet; on release, `generation != templateGeneration.get()` is `1 != 1` → false, so that stale transformer is re-offered and keeps serving the old stylesheet until the next reload. Narrow (two adjacent statements) and only under `contentCache=false` high concurrency, but sticky once hit. Stamping each transformer with *its own creation generation* (and comparing that on release, rather than the borrow-time gen) would close it fully. The current `testConcurrentProcessingDuringReload` reloads first and then spawns threads, so it exercises reload-then-read rather than reload-concurrent-with-transform — a test on that ordering would guard the window.

Non-blocking — net still much better than main.

PR #24777, opened 2026-07-16
CAMEL-24108: camel-xslt-saxon - apply secureProcessing unconditionally

davsclaus:
## Summary

- **Fix `secureProcessing` no-op:** `FEATURE_SECURE_PROCESSING` is now applied to the Saxon `TransformerFactory` unconditionally, independent of whether `saxonExtensionFunctions` is configured. Previously the documented default (`true`) was silently ignored unless extension functions were present.
- **Restrict external access:** The Saxon factory now attempts to set `ACCESS_EXTERNAL_DTD` and `ACCESS_EXTERNAL_STYLESHEET` to empty strings (matching the plain `xslt` component behavior). Saxon HE does not support these JAXP attributes, so the attempt is caught gracefully.
- **Honor `transformerFactoryConfigurationStrategy`:** The strategy is now invoked on all factory creation paths for both `xslt-saxon` (previously never invoked) and plain `xslt` (previously only invoked when `transformerFactoryClass` was explicitly set).

## Test plan

- [x] New `XsltSaxonSecureProcessingTest` — verifies secure processing is on by default, can be disabled, and transformations work correctly
- [x] New `XsltSaxonConfigurationStrategyTest` — verifies the configuration strategy is actually invoked on the default factory path
- [x] All 41 existing `camel-xslt-saxon` tests pass (including `SaxonXsltDTDTest`)
- [x] Upgrade guide entry added for 4.22

### DIFF
--- a/components/camel-xslt-saxon/src/main/java/org/apache/camel/component/xslt/saxon/XsltSaxonEndpoint.java
+++ b/components/camel-xslt-saxon/src/main/java/org/apache/camel/component/xslt/saxon/XsltSaxonEndpoint.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
@@ -45,6 +46,7 @@ import org.apache.camel.Component;
 import org.apache.camel.Exchange;
 import org.apache.camel.api.management.ManagedAttribute;
 import org.apache.camel.api.management.ManagedResource;
+import org.apache.camel.component.xslt.TransformerFactoryConfigurationStrategy;
 import org.apache.camel.component.xslt.XsltBuilder;
 import org.apache.camel.component.xslt.XsltEndpoint;
 import org.apache.camel.spi.ClassResolver;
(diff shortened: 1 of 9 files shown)


gnodet (approved):
**Security hardening and configuration strategy fix — well-structured and well-tested.**

This PR combines the thread-safety fixes from #24776 with three additional improvements to camel-xslt-saxon:

### 1. `secureProcessing` applied unconditionally
Previously, `FEATURE_SECURE_PROCESSING` was only set inside `registerSaxonExtensionFunctions()` — so with no extension functions configured (the common case), the documented default `secureProcessing=true` was silently ignored. Now `configureSecureProcessing()` is called as a dedicated step during factory initialization, independent of extension functions.

### 2. External access restrictions
`ACCESS_EXTERNAL_DTD` and `ACCESS_EXTERNAL_STYLESHEET` are now set to empty strings on the Saxon factory, wrapped in try-catch since Saxon HE doesn't support these JAXP attributes. Graceful degradation is the right approach.

### 3. `transformerFactoryConfigurationStrategy` honored on all paths
- **`xslt-saxon`**: previously never invoked — now called after factory creation and security hardening.
- **Plain `xslt`**: previously only invoked when `transformerFactoryClass` was explicitly set — now moved outside that conditional to cover all factory creation paths (including the default JDK factory).

### Code structure observations
- Saxon-specific configuration (secure processing, extension functions, config properties) correctly moved inside the `factory == null` block — only applied when a NEW factory is created, not when the user provides a pre-configured one via `getTransformerFactory()`.
- The `configureSecureProcessing()` extraction keeps `registerSaxonExtensionFunctions()` focused on extension registration only.

### Upgrade guide
Good documentation of the behavior change — users of Saxon EE relying on Java extension functions from XSLT will need to set `secureProcessing=false` explicitly.

**Static analysis:** ast-grep flagged three pre-existing mutable collection return issues — not introduced by this PR.

**Checklist:**
- ✅ Tests: 2 new Saxon-specific tests (secure processing, configuration strategy) + 3 reload tests from the thread-safety fix
- ✅ Upgrade guide entry added for 4.22
- ✅ No public API changes (internal method signature change in `XsltSaxonHelper`)
- ✅ Security: strengthens defaults — `secureProcessing` and external access now enforced
- ⏳ CI: builds in progress

oscerd (approved):
LGTM, and a genuinely useful security fix. The important part: `FEATURE_SECURE_PROCESSING` was previously only set inside `registerSaxonExtensionFunctions()`, guarded by a non-empty extension-function list — so the *documented* `secureProcessing=true` default was a silent no-op in the common case (no extension functions). Moving it into an unconditional `configureSecureProcessing()` step in `createXsltBuilder()` means Camel-created Saxon factories are now actually hardened, with `secureProcessing=false` as a documented escape hatch. Also sets `ACCESS_EXTERNAL_DTD`/`ACCESS_EXTERNAL_STYLESHEET=""` best-effort (Saxon HE throws `IllegalArgumentException`, caught/logged — graceful degradation matching plain xslt). Good that the breaking impact for Saxon PE/EE extension-function users is called out in the 4.22 upgrade guide, and the new tests cover default-on / can-disable.

Two non-blocking notes:
- The PR bundles the CAMEL-24107 `XsltBuilder` thread-safety change (volatile `template` + `AtomicLong` generation + generation-based pool invalidation) alongside the CAMEL-24108 work — worth making sure that doesn't collide with #24776 (also CAMEL-24107) at merge.
- Saxon-specific config + secureProcessing were moved inside `if (factory == null)`, so a user-supplied `transformerFactory` no longer receives them (it did before). Reasonable ("respect the pre-configured factory"), but it's a behaviour change that isn't mentioned in the upgrade-guide note.

PR #24778, opened 2026-07-16
CAMEL-24109: camel-xslt-saxon - remove duplicated doInit/doStart lifecycle

davsclaus:
## Summary

- Remove duplicated `doInit()` and `doStart()` overrides from `XsltSaxonEndpoint` — the parent `XsltEndpoint` already performs the full lifecycle through virtual dispatch to the overridden `createXsltBuilder()` and `loadResource()`, so the stylesheet was being created, loaded, and compiled twice on every startup.
- Copy-paste remnant from the CAMEL-14176 module split.

## Test plan

- [x] All 37 existing camel-xslt-saxon tests pass

🤖 Generated with [Claude Code](https://claude.ai/code) on behalf of @davsclaus

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #24779, opened 2026-07-16
CAMEL-24110: camel-xslt-saxon - fix saxonReaderProperties null-guard, DOMSource URIResolver, unhardened SAXParser

davsclaus:
## Summary

Fixes four interrelated bugs in `XsltSaxonEndpoint` originating from CAMEL-14501 (2020):

- **Dead null-guard**: `saxonReaderProperties` is initialized to an empty `HashMap`, so `if (saxonReaderProperties != null)` is always true — every stylesheet is unconditionally re-wrapped through `createReaderForSource()`. Fixed by guarding with `!saxonReaderProperties.isEmpty()`.

- **DOMSource broken**: A custom `URIResolver` returning a `DOMSource` gets `SAXSource.sourceToInputSource()` → `null`, producing a `SAXSource` with null `InputSource`. Fixed by checking if `sourceToInputSource()` returns null and passing the original source through unchanged.

- **Unhardened SAXParserFactory**: `createReaderForSource()` creates a bare `SAXParserFactory.newInstance()` with zero security features. Fixed by enabling `FEATURE_SECURE_PROCESSING`, `disallow-doctype-decl`, and disabling external entities — matching the hardening used elsewhere in Camel.

- **Unreachable null-source error**: In `loadResource()`, the wrap happened before the `source == null` check, so the intended `IOException("Cannot load schema resource ...")` could never fire. Fixed by checking null first, matching the parent `XsltEndpoint.loadResource()`.

Additionally: `Boolean.valueOf()` → `Boolean.parseBoolean()` (avoids unnecessary boxing), and error fallback returns the original source instead of null.

## Test plan

- [x] New `XsltSaxonUriResolverDomSourceTest` — custom `URIResolver` returning `DOMSource` with xslt-saxon route (was broken, now passes)
- [x] All 38 existing camel-xslt-saxon tests pass

### DIFF
--- a/components/camel-xslt-saxon/src/main/java/org/apache/camel/component/xslt/saxon/XsltSaxonEndpoint.java
+++ b/components/camel-xslt-saxon/src/main/java/org/apache/camel/component/xslt/saxon/XsltSaxonEndpoint.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
@@ -282,9 +283,7 @@ public class XsltSaxonEndpoint extends XsltEndpoint {
         InputStream is = getCamelContext().getTypeConverter().mandatoryConvertTo(InputStream.class, exchange, template);
         XsltBuilder builder = createXsltBuilder();
         Source source = new StreamSource(is);
-        if (this.saxonReaderProperties != null) {
-            //for Saxon we need to create XMLReader for the coming source
-            //so that the features configuration can take effect
+        if (!this.saxonReaderProperties.isEmpty()) {
             source = createReaderForSource(source);
         }
         builder.setTransformerSource(source);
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**LGTM — four well-identified bugs fixed with proper security hardening.**

**Verification of each fix:**

1. **Dead null-guard** ✅ — Confirmed `saxonReaderProperties` is initialized as `new HashMap<>()` (line 76), so the `!= null` check was always true. The `!isEmpty()` fix correctly skips the unnecessary `createReaderForSource()` wrapping when no reader properties are configured.

2. **DOMSource broken** ✅ — `SAXSource.sourceToInputSource()` returns null for `DOMSource` (by spec — DOMSource has no InputSource representation). Moving the null check to the top of `createReaderForSource()` and returning the original source unchanged is the correct fix. The new `XsltSaxonUriResolverDomSourceTest` directly exercises this path.

3. **Unhardened SAXParser** ✅ — The `SAXParserFactory.newInstance()` without security features was exposed to XXE. The added features (`FEATURE_SECURE_PROCESSING`, `disallow-doctype-decl`, disable external entities) match the standard Camel XML hardening pattern. The Apache Xerces-specific `disallow-doctype-decl` feature is safely wrapped in the try-catch — if a non-Xerces parser is used, it falls through to the catch and returns the original source.

4. **Unreachable null-source error** ✅ — Confirmed the parent `XsltEndpoint.loadResource()` checks `source == null` first (before any wrapping). The Saxon override had the wrapping before the null check, making the `IOException` unreachable. Fixed by matching the parent's order.

**Additional fixes** ✅ — `Boolean.valueOf()` → `Boolean.parseBoolean()` avoids unnecessary boxing. Error fallback `return source` instead of `return null` prevents NPE downstream.

**Static analysis:** ast-grep found 3 pre-existing `mutable-collection-return` warnings on getter methods — these are unrelated to this PR (standard Camel endpoint configuration pattern).

**Note:** This PR and #24778 both modify `XsltSaxonEndpoint.java` from the same base. The changes are in different regions (doInit/doStart removal vs. loadResource/createReaderForSource), so they should merge cleanly.

**CI:** Builds still in progress.

oscerd (approved):
LGTM — real XXE hardening on the stylesheet-parse path. `createReaderForSource` now sets `FEATURE_SECURE_PROCESSING=true`, `disallow-doctype-decl=true`, and external general/parameter entities `=false` on the `SAXParserFactory` *before* `newSAXParser()`, so the hardening lands on the actual parser that produces the returned `SAXSource` (not a throwaway). `disallow-doctype-decl` also makes `load-external-dtd`/XInclude moot. The DOMSource fix is sound too — a `DOMSource` from a custom `URIResolver` is already parsed, so returning it unwrapped introduces no re-parse/SSRF (and it fixes the previously-broken `SAXSource(reader, null)`).

Two non-blocking notes:
- Changing the guard from `saxonReaderProperties != null` to `!saxonReaderProperties.isEmpty()` means `setSaxonReaderProperties(null)` would NPE — low risk since the field defaults to an empty map, but a `!= null` belt is cheap.
- The newly-hardened path has only a happy-path DOMSource test; a negative test feeding a malicious DOCTYPE/external-entity *stylesheet* and asserting it's now rejected would lock the hardening in against regressions.

PR #24782, opened 2026-07-16
CAMEL-24116: Fix RestOpenApiProcessor lifecycle - services never stopped

davsclaus:
## Summary

_Claude Code on behalf of davsclaus_

- Fix `RestOpenApiProcessor` lifecycle management that became dead code after the advice conversion in CAMEL-22742
- Add `doStop()` to `RestOpenApiEndpoint` to stop the per-consumer `RestOpenApiProcessor` (stops per-operation `RestBindingAdvice` bindings and clears paths)
- Add `doStop()` to `RestOpenApiComponent` to stop the shared `RestOpenapiProcessorStrategy` (stops the `ProducerCache` and removes stale HTTP endpoint registrations from `PlatformHttpComponent`)

### Problem

Commit `bd63a7eb8213` (CAMEL-22742) moved `RestOpenApiProcessor` out of the consumer's processor chain into a `RestOpenApiProcessorAdvice` added via `ip.addAdvice()`. Previously `DefaultConsumer` drove the processor's lifecycle; advices get no lifecycle management, so `doInit()`/`doStop()` on `RestOpenApiProcessor` became dead code.

Consequences:
- `DefaultRestOpenapiProcessorStrategy.doStop()` never runs: producer cache is never stopped and `PlatformHttpComponent.removeHttpEndpoint()` is never called
- Stale HTTP endpoint registrations accumulate on route reload (e.g. `camel-jbang --dev`)
- Per-operation `RestBindingAdvice` services are never stopped

### Fix

- `RestOpenApiEndpoint` now tracks the `RestOpenApiProcessor` it creates and stops it in `doStop()`, which in turn stops the per-operation bindings
- `RestOpenApiComponent` now stops the shared `restOpenapiProcessorStrategy` in `doStop()`, which stops the producer cache and removes HTTP endpoint registrations

## Test plan

- [x] All 135 existing `camel-rest-openapi` tests pass
- [ ] Manual verification with `camel-jbang --dev` route reload to confirm stale registrations are cleaned up

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #24783, opened 2026-07-16
CAMEL-24117: camel-rest-openapi - Create per-endpoint strategy instance

davsclaus:
## Summary

- Fix shared `DefaultRestOpenapiProcessorStrategy` instance causing cross-talk between multiple `rest().openApi()` consumers
- Each endpoint now gets its own strategy instance (unless the user explicitly supplies a custom one at the component level)
- Fixes `missingOperation`/`mockIncludePattern` overwrite, silent fail-fast loss, and `uris` list mixing across consumers

## Root cause

`RestOpenApiComponent.doInit()` created ONE `DefaultRestOpenapiProcessorStrategy` and `createEndpoint()` shared it across all endpoints. Each consumer's `afterPropertiesConfigured()` then called `setMissingOperation()` / `setMockIncludePattern()` on the shared object — last writer wins. Combined with `ServiceHelper.initService` idempotency, the second consumer's null `missingOperation` was never resolved, causing `validateOpenApi()` to silently skip all fail/ignore/mock branches.

## Changes

- **`RestOpenApiComponent.createEndpoint()`**: Create a fresh `DefaultRestOpenapiProcessorStrategy` per endpoint when no user-supplied custom strategy exists. Share only an explicitly user-supplied strategy.
- **`RestOpenApiComponent.doInit()`**: Remove auto-creation of shared default strategy (no longer needed).

### DIFF
--- a/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiComponent.java
+++ b/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiComponent.java
@@ -178,7 +178,13 @@ public final class RestOpenApiComponent extends DefaultComponent implements SSLC
         }
         endpoint.setMissingOperation(getMissingOperation());
         endpoint.setMockIncludePattern(getMockIncludePattern());
-        endpoint.setRestOpenapiProcessorStrategy(getRestOpenapiProcessorStrategy());
+        if (restOpenapiProcessorStrategy != null) {
+            endpoint.setRestOpenapiProcessorStrategy(restOpenapiProcessorStrategy);
+        } else {
+            DefaultRestOpenapiProcessorStrategy perEndpoint = new DefaultRestOpenapiProcessorStrategy();
+            CamelContextAware.trySetCamelContext(perEndpoint, getCamelContext());
+            endpoint.setRestOpenapiProcessorStrategy(perEndpoint);
+        }
         setProperties(endpoint, parameters);
         return endpoint;
     }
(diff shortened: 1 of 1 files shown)


gnodet (approved):
**LGTM — correct fix for shared strategy cross-talk.**

**Verification:**
- `DefaultRestOpenapiProcessorStrategy` has per-consumer mutable state: `uris` list (line 77, accumulates HTTP endpoint registrations), `missingOperation` (line 414), and `mockIncludePattern` (line 424). Sharing one instance across multiple `rest().openApi()` consumers causes last-writer-wins on configuration and URI mixing.
- The fix correctly distinguishes user-supplied (`restOpenapiProcessorStrategy != null` — field access, not getter) from auto-created strategies. User-supplied strategies are shared as intended; default strategies are created per-endpoint.
- Removing the auto-creation from `doInit()` is safe — the field stays null unless explicitly set by the user, and each endpoint gets its own fresh instance in `createEndpoint()`.
- Compatible with #24782 (lifecycle fix): the component's `doStop()` handles null via `ServiceHelper.stopService()`, and per-endpoint strategies are stopped by each endpoint's own `doStop()`.

oscerd (commented):
The fix is correct — `DefaultRestOpenapiProcessorStrategy` holds per-consumer mutable state (`uris`, `missingOperation`, `producerCache`), and on main a single auto-created instance was shared across every endpoint; since `ServiceHelper.initService` is idempotent, a second consumer's re-init was skipped and `uris` accumulated (last-writer-wins cross-talk). Creating a fresh instance per endpoint (sharing only an explicitly user-supplied one) is the right fix, and each per-endpoint strategy is init/started/stopped by its owning processor.

One thing before merge: the multi-consumer cross-talk this fixes ships with **no regression test** — two `rest().openApi()` consumers asserting their URIs/config don't bleed into each other is exactly the scenario that silently regressed, and it's straightforward to unit-test. Also note the per-endpoint strategies only actually get *stopped* once #24782 lands (this PR adds no stop path on its own), so the two are best merged together.

PR #24784, opened 2026-07-16
CAMEL-24118: Fix RestOpenApiReader to emit valid OpenAPI 3.x

davsclaus:
## Summary

Fixes a family of 7+ related defects in `RestOpenApiReader` where the generated OpenAPI 3.x documents are invalid or semantically wrong. All defects predate the swagger-core migration and were faithfully ported from the apicurio-era code.

### Fixes included

1. **formData params** — no longer emit illegal `"in":"formData"`. Instead, formData params are collected and emitted as a `requestBody` with `application/x-www-form-urlencoded` (or `multipart/form-data`) object schema, per OAS 3.x spec.

2. **collectionFormat -> style mapping** — `multi` now maps to `style: form, explode: true` (was incorrectly `deepObject`). `csv` now maps to `style: form, explode: false` (was missing `explode: false`, silently getting multi semantics).

3. **boolean outType** — `outType("boolean")` now emits `BooleanSchema` (`type: boolean`) instead of `NumberSchema.format("boolean")` (`type: number, format: boolean`).

4. **Primitive-array outTypes** — `outType("int[]")`, `outType("string[]")` etc. now use the computed typed schema as array items instead of discarding it. Previously emitted `{"type":"array","items":{}}`.

5. **Response header types** — `long`, `float`, `double` now map to valid JSON Schema types: `long` -> `integer/int64`, `float` -> `number/float`, `double` -> `number/double`. Previously emitted raw type strings which are not valid OAS 3.x types.

6. **Array response headers** — `setHeaderSchemaOas30` now wraps the inner type in an `ArraySchema`. Previously set the inner type directly as the header schema, losing the array wrapper.

7. **OAS 3.1 type loss** — response header schemas and byte[] now use typed schema classes (`IntegerSchema`, `NumberSchema`, `BooleanSchema`, `StringSchema`) which correctly populate both the `type` field (3.0) and `types` set (3.1). Previously used raw `new Schema().type(...)` which only sets the 3.0 field.

8. **byte[] outType** — now uses `ByteArraySchema` (`type: string, format: byte`) instead of `new Schema().type("number").format("byte")`.

## Test plan

- [x] All 97 existing `camel-openapi-java` tests pass
- [ ] Manual validation of generated OpenAPI documents for each fixed scenario

### DIFF
--- a/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
+++ b/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestOpenApiReader.java
@@ -78,6 +78,7 @@ import org.apache.camel.model.rest.ParamDefinition;
 import org.apache.camel.model.rest.ResponseHeaderDefinition;
 import org.apache.camel.model.rest.ResponseMessageDefinition;
 import org.apache.camel.model.rest.RestDefinition;
+import org.apache.camel.model.rest.RestParamType;
 import org.apache.camel.model.rest.RestPropertyDefinition;
 import org.apache.camel.model.rest.RestSecuritiesDefinition;
 import org.apache.camel.model.rest.RestSecurityDefinition;
@@ -545,8 +546,19 @@ public class RestOpenApiReader {
             op.addSecurityItem(securityRequirement);
         }
 
+        // Collect formData params to build a single requestBody with object schema
+        List<ParamDefinition> formDataParams = new ArrayList<>();
+
         for (ParamDefinition param : verb.getParams()) {
-            Parameter parameter = new Parameter().in(param.getType().name());
+            RestParamType paramType = param.getType();
+
+            // In OAS 3.x, formData is not a valid "in" value - collect for requestBody handling
+            if (RestParamType.formData == paramType) {
+                formDataParams.add(param);
+                continue;
+            }
+
+            Parameter parameter = new Parameter().in(paramType.name());
 
             if (parameter != null) {
                 parameter.setName(getValue(camelContext, param.getName()));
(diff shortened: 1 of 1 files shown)


gnodet (approved):
**LGTM — comprehensive OAS 3.x compliance fixes.**

**Verification of key fixes:**

1. **formData → requestBody** ✅ — `"in":"formData"` is indeed illegal in OAS 3.x (only `query`/`header`/`path`/`cookie` are valid). The new code correctly collects formData params and emits them as a `requestBody` with an object schema under the appropriate media type. The `multipart/form-data` detection from the `consumes` list is a nice touch.

2. **collectionFormat mapping** ✅ — `multi` → `DEEPOBJECT` was wrong; `deepObject` is for bracket notation (`filter[name]=value`), not multi-value params. The fix (`multi` → `FORM` + `explode: true`, `csv` → `FORM` + `explode: false`) matches the [OAS 3.x Parameter Serialization spec](https://swagger.io/docs/specification/v3_0/serialization/).

3. **boolean outType** ✅ — `NumberSchema().format("boolean")` → `BooleanSchema()` is clearly correct. `"type": "number", "format": "boolean"` is not a valid JSON Schema type.

4. **Primitive-array outTypes** ✅ — The old code discarded the computed typed schema and used empty items (`new Schema<>()`). Now correctly wraps `prop` (the typed schema) as array items when there's no `$ref`.

5. **byte[] outType** ✅ — `ByteArraySchema` from swagger-core correctly emits `type: string, format: byte`, not the old `type: number, format: byte`.

6. **Response header typed schemas** ✅ — Using `IntegerSchema`/`NumberSchema`/`BooleanSchema` classes which populate both `type` (3.0) and `types` set (3.1), instead of `new Schema().type(...)` which only sets the 3.0 field.

7. **Array response headers** ✅ — `setHeaderSchemaOas30` now wraps items in `ArraySchema` instead of setting the inner type directly as the header schema.

- The `openApi` parameter in `createHeaderTypedSchema()` and `createTypedSchema()` is unused in those methods. The typed schema classes handle versioning internally. Not a bug — just a minor style note.
- No new tests are added for these 8 fixes. The existing 97 tests pass, and the changes are structural (correct schema classes vs. raw schemas), but dedicated tests for formData-to-requestBody conversion and the collectionFormat mapping would strengthen confidence for future refactors.

**Static analysis:** Only pre-existing `broad-exception-catch` warnings. Nothing introduced by this PR.

PR #24789, opened 2026-07-16
CAMEL-24123: camel-smb - Fix readLock=changed with recursive=true

davsclaus:

## Summary

- Fix `readLock=changed` with `recursive=true` never consuming files in sub-directories
- `SmbExclusiveReadLockCheck.getSmbFiles()` was passing `file.getFileName()` (relative path, e.g. `sub/hello.txt`) as the SMB search pattern instead of `file.getFileNameOnly()` (plain name, e.g. `hello.txt`)
- SMB QueryDirectory search patterns must be plain names/wildcards; a pattern containing a path separator matches nothing, so the lock check always saw `newLength=0 < minLength` and the read lock was never acquired

## Test plan

- [ ] Verify build passes: `cd components/camel-smb && mvn install -DskipTests`
- [ ] Run existing camel-smb tests: `cd components/camel-smb && mvn test`
- [ ] Verify with `recursive=true` and `readLock=changed` that files in sub-directories are consumed

### DIFF
--- a/components/camel-smb/src/main/java/org/apache/camel/component/smb/strategy/SmbExclusiveReadLockCheck.java
+++ b/components/camel-smb/src/main/java/org/apache/camel/component/smb/strategy/SmbExclusiveReadLockCheck.java
@@ -84,7 +84,7 @@ public class SmbExclusiveReadLockCheck {
 
         String path = file.getParent();
         if (operations instanceof SmbOperations smbOperations) {
-            return smbOperations.listFiles(path, file.getFileName());
+            return smbOperations.listFiles(path, file.getFileNameOnly());
         }
         return operations.listFiles(path);
     }


gnodet (approved):
Clean 1-line fix — the bug is obvious once spotted, and the code itself already proves the correct method.

**The bug:** `getSmbFiles()` passes `file.getFileName()` (relative path, e.g. `sub/hello.txt` for recursive consumers) as the SMB QueryDirectory search pattern. SMB search patterns are plain names/wildcards within a directory — a pattern with path separators matches nothing, so `share.list(path, "sub/hello.txt")` returns an empty array, `newLength=0`, and `isReadLockAcquired()` never succeeds.

**The proof is in the same method:** At line 59, the match logic already uses `file.getFileNameOnly()`:
```java
boolean match = f.getFileName().equals(file.getFileNameOnly());
```
This confirms the search pattern should also be `getFileNameOnly()` — the listing and the matching must agree on the name format.

**Fix:** Change `file.getFileName()` → `file.getFileNameOnly()` in the `listFiles()` call. The parent directory is already correctly extracted via `file.getParent()`.

**Checklist:**
- ✅ 1-line fix, obviously correct
- ✅ Consistent with the match at line 59
- ✅ No new test, but the fix is self-evidently correct from the surrounding code
- ✅ CI pending (builds running)

PR #24790, opened 2026-07-16
CAMEL-24129: camel-jpa - JpaConsumer should use getPreDeleteHandler() instead of createPreDeleteHandler()

davsclaus:
## Summary

`JpaConsumer.processBatch()` calls `createPreDeleteHandler()` directly instead of the cached accessor `getPreDeleteHandler()`, causing two bugs:

1. A user-supplied `preDeleteHandler` (configured via the `@UriParam` option on `JpaEndpoint`) is **never invoked** — `createPreDeleteHandler()` always re-derives the handler from the `@PreConsumed` annotation scan, completely bypassing anything set via `setPreDeleteHandler()`.
2. `createPreDeleteHandler()` runs `AnnotationHelper.findMethodsWithAnnotation()` — a reflective scan of the entity's methods — for **every entity of every poll batch**, instead of once per consumer lifetime.

The `@Consumed`/`getDeleteHandler()` counterpart a few lines below already uses the correct cached accessor pattern.

## Fix

Changed line 199 from `createPreDeleteHandler()` to `getPreDeleteHandler()`, mirroring the existing `getDeleteHandler()` pattern.

## Test

Added `JpaPreDeleteHandlerTest` that sets a custom `preDeleteHandler` programmatically on the endpoint and verifies it is actually invoked when entities are consumed.

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaConsumer.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaConsumer.java
@@ -196,7 +196,7 @@ public class JpaConsumer extends ScheduledBatchPollingConsumer {
             pendingExchanges = total - index - 1;
             if (lockEntity(result, batchEntityManager)) {
                 // Run the @PreConsumed callback
-                createPreDeleteHandler().deleteObject(batchEntityManager, result, exchange);
+                getPreDeleteHandler().deleteObject(batchEntityManager, result, exchange);
 
                 // process the current exchange
                 LOG.debug("Processing exchange: {}", exchange);
--- a/components/camel-jpa/src/test/java/org/apache/camel/processor/jpa/JpaPreDeleteHandlerTest.java
+++ b/components/camel-jpa/src/test/java/org/apache/camel/processor/jpa/JpaPreDeleteHandlerTest.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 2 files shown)


gnodet (approved):
Classic accessor bug — `createPreDeleteHandler()` instead of `getPreDeleteHandler()`. The fix is a 1-method-name change, and the code speaks for itself.

**The bug (two consequences):**
1. A user-supplied `preDeleteHandler` (set via `setPreDeleteHandler()` / the `@UriParam` option on `JpaEndpoint`) is **never invoked** — `createPreDeleteHandler()` always creates a fresh handler from `@PreConsumed` annotation scanning, completely bypassing the `preDeleteHandler` field.
2. `createPreDeleteHandler()` runs `AnnotationHelper.findMethodsWithAnnotation()` — a reflective scan — **for every entity of every poll batch** instead of once per consumer lifetime.

**The proof is 16 lines below:**
```java
// Line 215: the @Consumed counterpart correctly uses the cached accessor
getDeleteHandler().deleteObject(batchEntityManager, result, exchange);
```
The `getPreDeleteHandler()` accessor (lines 259–264) does the right thing: returns the user-set handler if present, otherwise creates and caches one via `createPreDeleteHandler()`.

**Test is well-designed:**
- Sets a custom `preDeleteHandler` via `jpa.setPreDeleteHandler(...)` lambda that flips an `AtomicBoolean`
- Sends a `SendEmail` entity, consumes it via the JPA consumer, and verifies the flag was flipped
- Uses `MockEndpoint.assertIsSatisfied(context)` correctly (no Awaitility wrapping)

**Checklist:**
- ✅ 1-line fix, mirrors the existing `getDeleteHandler()` pattern
- ✅ Test proves the user-supplied handler is now invoked
- ✅ CI pending (builds running)

PR #24792, opened 2026-07-16
CAMEL-24130: camel-jpa - JpaMessageIdRepository should not reuse EntityManager from exchange header

davsclaus:
## Summary

`JpaMessageIdRepository` hardcodes `usePassedInEntityManager=true` in all `getTargetEntityManager()` calls, which causes it to grab the `EntityManager` from the exchange's `CamelEntityManager` header. When used downstream of a JPA consumer (which puts its own poll-scoped EM in that header), the repository then **closes** the consumer's EM in its `finally` block, causing `IllegalStateException: EntityManager is closed` during the consumer's subsequent flush/clear.

## Fix

Changed `usePassedInEntityManager` from `true` to `false` in all four `getTargetEntityManager()` calls (`add`, `contains`, `remove`, `clear`). This makes the repository always create its own EM via its own `entityManagerFactory`, matching `JpaEndpoint`'s default behavior. The existing `finally` close logic is then correct — the repository only closes EMs it created itself.

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/processor/idempotent/jpa/JpaMessageIdRepository.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/processor/idempotent/jpa/JpaMessageIdRepository.java
@@ -90,7 +90,7 @@ public class JpaMessageIdRepository extends ServiceSupport implements Idempotent
     @Override
     public boolean add(final Exchange exchange, final String messageId) {
         final EntityManager entityManager
-                = getTargetEntityManager(exchange, entityManagerFactory, true, sharedEntityManager, true);
+                = getTargetEntityManager(exchange, entityManagerFactory, false, sharedEntityManager, true);
         // Run this in single transaction.
         final Boolean[] rc = new Boolean[1];
         transactionStrategy.executeInTransaction(() -> {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
### Review: CAMEL-24130 — JpaMessageIdRepository should not reuse EntityManager from exchange header

**Verdict: Approve** ✅

The fix correctly changes `usePassedInEntityManager` from `true` to `false` in all four `getTargetEntityManager()` calls (`add`, `contains`, `remove`, `clear`).

**Why this is correct:**

`JpaMessageIdRepository` manages its own EntityManager lifecycle — each method creates an EM, uses it within `transactionStrategy.executeInTransaction()`, and closes it in the `finally` block. When `usePassedInEntityManager=true`, `getTargetEntityManager()` first checks the exchange header for `CamelEntityManager`, which the JPA consumer populates with its poll-scoped EM. The repository then closes *that* EM in its `finally` block, leaving the consumer with a closed EM for its subsequent flush/clear — causing `IllegalStateException: EntityManager is closed`.

The `usePassedInEntityManager` parameter was set to `true` when CAMEL-18450 refactored the repository to use `getTargetEntityManager()` (commit `bbbb47efd8f3`), but this was likely unintentional — the repository previously created its own EM directly via the factory and never reused one from the exchange. Setting it to `false` restores that self-contained lifecycle.

**Notes:**
- The remaining EM resolution path (exchange property → transaction context → shared EM → create new) still works correctly. The `allowRecreate=true` parameter ensures a closed EM from the exchange property is replaced with a fresh one.
- Change is minimal and focused — 4 identical boolean flips across the 4 public methods.

PR #24793, opened 2026-07-16
CAMEL-24131: camel-jpa - Guard consumeDelete handler against Object[] from native queries

davsclaus:
## Summary

When a JPA consumer uses `nativeQuery` without `resultClass`, results are returned as `Object[]`. The default `consumeDelete=true` handler calls `em.remove(entityBean)` unconditionally, which throws `IllegalArgumentException` on `Object[]` since it's not a managed entity. This causes the row to never be deleted and reprocessed on every poll — an infinite reprocessing loop.

## Fix

Added an array type guard to the `consumeDelete` handler, matching the pattern already used by `lockEntity()` (line 376) which anticipates `Object[]` from native queries. When the entity bean is an array, the delete is skipped and a WARN log is emitted advising the user to set `consumeDelete=false` or configure a `resultClass`.

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaConsumer.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaConsumer.java
@@ -471,7 +471,15 @@ public class JpaConsumer extends ScheduledBatchPollingConsumer {
             }
         }
         if (getEndpoint().isConsumeDelete()) {
-            return (EntityManager em, Object entityBean, Exchange exchange) -> em.remove(entityBean);
+            return (EntityManager em, Object entityBean, Exchange exchange) -> {
+                if (entityBean != null && !entityBean.getClass().isArray()) {
+                    em.remove(entityBean);
+                } else {
+                    LOG.warn("Cannot auto-delete entity of type {} (e.g. native query result without resultClass)."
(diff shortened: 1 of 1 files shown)


gnodet (approved):
### Review: CAMEL-24131 — Guard consumeDelete handler against Object[] from native queries

**Verdict: Approve** ✅

The fix adds a type guard to the default `consumeDelete` handler in `createDeleteHandler()`, preventing `em.remove()` from being called on `Object[]` results (which aren't managed JPA entities).

**Why this is correct:**

When using `nativeQuery` without `resultClass`, JPA returns `Object[]` rows. Calling `em.remove()` on an `Object[]` throws `IllegalArgumentException` because it's not a managed entity. This causes the row to be reprocessed on every poll — an infinite reprocessing loop.

The guard follows the established precedent in the same class:
- `lockEntity()` (line 377) already uses `entity.getClass().isArray()` to handle native query results, with a `TODO` comment acknowledging the native query limitation
- The new guard mirrors this pattern exactly

**The warning message is well-crafted** — it tells the user both how the problem arose (native query without resultClass) and the two concrete remediation options (`consumeDelete=false` or configure `resultClass`).

**Edge cases handled:**
- `null` entityBean → guarded with `entityBean != null` check, logged as type "null"
- `Object[]` → `isArray()` catches it, log shows the actual class name

PR #24794, opened 2026-07-16
CAMEL-24132: camel-jpa - Fix medium-severity findings from component review

davsclaus:
## Summary

Fixes 5 of 6 medium-severity findings from the July 2026 camel-jpa component review ([CAMEL-24132](https://issues.apache.org/jira/browse/CAMEL-24132)):

- **M1**: `receiveNoWait()` no longer delegates to `receive()` with pessimistic locking — refactored into `doReceive(boolean useLocking)` so `receiveNoWait()` runs without locks
- **M2**: `receive(timeout)` now calls `future.cancel(true)` on `TimeoutException` so abandoned tasks don't accumulate holding DB locks
- **M3**: `setLockMode()` is skipped for native queries, which throw `IllegalStateException` per JPA spec
- **M4**: Concurrent insert race in `JpaMessageIdRepository.add()` now catches constraint violations and maps them to `return false` (duplicate detected) instead of failing the exchange. Also removed premature `entityManager.close()` inside the transaction lambda
- **M6**: `isUseExecuteUpdate()` now trims query strings and detects update keywords (`INSERT`/`UPDATE`/`DELETE`) instead of select, so queries with leading whitespace or CTEs are correctly classified

M5 (parallel EntityManager fix missing from Multicast/RecipientList/WireTap) is tracked separately in [CAMEL-24174](https://issues.apache.org/jira/browse/CAMEL-24174).

## Test plan

- [x] All 96 existing camel-jpa tests pass (0 failures)
- [ ] Manual verification of M3 with `nativeQuery` + default `consumeLockEntity=true`
- [ ] Manual verification of M6 with leading-whitespace and CTE queries

### DIFF
--- a/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
+++ b/components/camel-jpa/src/main/java/org/apache/camel/component/jpa/JpaPollingConsumer.java
@@ -124,11 +124,21 @@ public class JpaPollingConsumer extends PollingConsumerSupport {
 
     @Override
     public Exchange receive() {
-        Exchange exchange = getEndpoint().createExchange();
+        return doReceive(true);
+    }
 
+    @Override
+    public Exchange receiveNoWait() {
+        // do not use locking so the call returns immediately without blocking on row locks
+        return doReceive(false);
+    }
+
+    private Exchange doReceive(boolean useLocking) {
         // resolve the entity manager before evaluating the expression
-        final EntityManager entityManager = getTargetEntityManager(exchange, entityManagerFactory,
+        final EntityManager entityManager = getTargetEntityManager(null, entityManagerFactory,
                 getEndpoint().isUsePassedInEntityManager(), getEndpoint().isSharedEntityManager(), true);
+
+        Exchange exchange = getEndpoint().createExchange();
         exchange.getIn().setHeader(JpaConstants.ENTITY_MANAGER, entityManager);
         transactionStrategy.executeInTransaction(new Runnable() {
             @Override
@@ -140,7 +150,9 @@ public class JpaPollingConsumer extends PollingConsumerSupport {
                 Query innerQuery = getQueryFactory().createQuery(entityManager);
                 configureParameters(innerQuery);
 
-                if (getEndpoint().isConsumeLockEntity()) {
+                // only apply lock mode when locking is enabled and not using native queries
+                // as JPA spec does not support setLockMode on native queries
+                if (useLocking && getEndpoint().isConsumeLockEntity() && nativeQuery == null) {
                     innerQuery.setLockMode(getLockModeType());
                 }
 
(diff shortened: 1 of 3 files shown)


gnodet (approved):
### Review: CAMEL-24132 — Fix medium-severity findings from component review

**Verdict: Approve** ✅

This PR addresses 5 of 6 medium-severity findings from the JPA component review. Each fix is well-targeted:

**M1 — `receiveNoWait()` refactored (`JpaPollingConsumer`)**
Clean refactoring into `doReceive(boolean useLocking)`. `receiveNoWait()` now correctly skips pessimistic locking so it returns immediately rather than blocking on row locks.

**M2 — `future.cancel(true)` on timeout (`JpaPollingConsumer`)**
Previously, `receive(timeout)` caught `TimeoutException` but left the future running — meaning the submitted `Callable` could continue holding DB locks indefinitely. The `future.cancel(true)` fix correctly interrupts the abandoned task.

**M3 — Skip `setLockMode()` for native queries (`JpaPollingConsumer`)**
The JPA spec throws `IllegalStateException` when `setLockMode()` is called on a native query. The guard `nativeQuery == null` correctly prevents this, and the `nativeQuery` field is already defined in the class (line 53). This is consistent with the native-query guards added in PR #24793 (`createDeleteHandler`) and the existing `lockEntity()` method.

**M4 — Constraint violation race handling (`JpaMessageIdRepository`)**
Two good changes:
- Catching `SQLIntegrityConstraintViolationException` / `EntityExistsException` and returning `false` (duplicate) instead of throwing is correct for concurrent insert races. Walking the cause chain handles JPA provider wrapping.
- Removing the premature `entityManager.close()` inside the `try` block (after `flush()`) is correct — the `finally` block already handles cleanup.

**M6 — `isUpdateQuery()` whitespace & keyword detection (`JpaProducer`)**
Fixes leading-whitespace queries that were misclassified. The logic change from "not-SELECT → update" to "INSERT/UPDATE/DELETE → update" is more explicit. This does change behavior for CTE-based DML (`WITH ... UPDATE ...`) and `MERGE`/`CALL` statements, which would now default to non-update, but users can set `useExecuteUpdate` explicitly for those edge cases, and CTE selects (`WITH ... SELECT ...`) — the more common case — are now correctly handled.

1. **FQCNs in `isConstraintViolation()`**: `java.sql.SQLIntegrityConstraintViolationException` and `jakarta.persistence.EntityExistsException` are used as fully qualified names rather than imported. Per project conventions, these should be imported. The build's OpenRewrite rule should auto-fix this, but the formatter/impsort might not add the imports automatically.

2. **Premature `entityManager.close()` in `remove()` and `clear()`**: The premature close-after-flush was correctly removed from `add()`, but the same pattern still exists in `remove()` (line 206) and `clear()` (line 256). These are benign (the `finally` guard prevents double-close) but inconsistent.

PR #24795, opened 2026-07-16
CAMEL-24174: Strip JPA EntityManager property in parallel multicast, recipientList and wireTap

davsclaus:
## Summary

CAMEL-22534 fixed a concurrency issue where the Splitter shared the non-thread-safe JPA `EntityManager` map across parallel sub-exchanges. The same fix was missing from three other EIPs:

- **MulticastProcessor** — strip `CamelEntityManager` when `parallelProcessing` is enabled
- **RecipientListProcessor** — strip `CamelEntityManager` when `parallelProcessing` is enabled
- **WireTapProcessor** — always strip `CamelEntityManager` since wire tap runs asynchronously

When the property is removed, the JPA component creates a fresh dedicated EntityManager for each branch, avoiding concurrent access to a non-thread-safe EM and HashMap corruption.

No JPA dependency is introduced in `camel-core-processor` — the property key is just a string constant from `Exchange`.

## Test plan

- [x] `camel-core-processor` builds successfully
- [x] 78 parallel multicast / recipient list / wire tap tests pass
- [x] camel-jpa `JpaWireTapTest` passes

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -1020,6 +1020,10 @@ public class MulticastProcessor extends BaseProcessorSupport
             // copy exchange, and do not share the unit of work
             Exchange copy = processorExchangeFactory.createCorrelatedCopy(exchange, false);
             copy.getExchangeExtension().setTransacted(exchange.isTransacted());
+            if (isParallelProcessing()) {
+                // do not share JPA EntityManager in parallel mode as it is not thread-safe (CAMEL-22534)
+                copy.removeProperty(Exchange.JPA_ENTITY_MANAGER);
+            }
             // If we are in a transaction, set TRANSACTION_CONTEXT_DATA property for new exchanges to share txData
             // during the transaction.
             if (exchange.isTransacted() && copy.getProperty(Exchange.TRANSACTION_CONTEXT_DATA) == null) {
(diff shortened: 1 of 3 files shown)


gnodet (approved):
### Review: CAMEL-24174 — Strip JPA EntityManager property in parallel multicast, recipientList and wireTap

**Verdict: Approve** ✅

This is a straightforward extension of the CAMEL-22534 fix (commit `20764a2811c1`) that was applied to `Splitter` but missed three other EIPs that also copy exchanges for concurrent processing.

**Consistency with the Splitter fix:**

The pattern is identical across all three files — `copy.removeProperty(Exchange.JPA_ENTITY_MANAGER)` to prevent sharing a non-thread-safe `EntityManager` map across parallel branches:

- **`MulticastProcessor`** — guarded by `isParallelProcessing()`, placed right after `setTransacted()` ✓
- **`RecipientListProcessor`** — guarded by `isParallelProcessing()`, same position ✓
- **`WireTapProcessor`** — unconditional (correct: wire tap always runs asynchronously on a separate thread) ✓

**Key design points:**
- No JPA dependency in `camel-core-processor` — `Exchange.JPA_ENTITY_MANAGER` is a string constant (`"CamelEntityManager"`) defined in `camel-api`, so the core processor module stays lightweight.
- When the property is removed, `JpaHelper.getTargetEntityManager()` falls through to creating a dedicated `EntityManager` per branch, which is the correct thread-safe behavior.
- The placement is right after `createCorrelatedCopy()` + `setTransacted()`, matching the Splitter's pattern.

Minimal, focused, and correct.

PR #24797, opened 2026-07-16
CAMEL-24134: Circuit Breaker EIP - prevent timed-out worker from writing back to exchange

davsclaus:
## Summary

Both `ResilienceProcessor` and `FaultToleranceProcessor` have a race condition where the worker thread writes results back to the original exchange after a timeout has triggered fallback processing on the caller thread. This can:

- Overwrite the fallback result with the late worker result
- Reinstate an exception after the fallback cleared it
- Corrupt the exchange through non-atomic interleaving of `ExchangeHelper.copyResults`

**Fix:** Add a shared `AtomicBoolean` guard between the worker task and the fallback task. The worker uses `compareAndSet(false, true)` before writing back — if the fallback already claimed the exchange, the write-back is skipped. The fallback sets the guard to `true` before touching the exchange.

Both the normal result path and the exception catch path in the worker are guarded.

**Note:** This PR depends on CAMEL-24133 (also included in this branch) which fixes a related pooled exchange factory release issue in the same processors.

- [CAMEL-24134](https://issues.apache.org/jira/browse/CAMEL-24134)

## Test plan

- [x] New `ResilienceTimeoutWriteBackRaceTest` — slow processor (3s) with short timeout (500ms), verifies fallback result survives after worker completes
- [x] New `FaultToleranceTimeoutWriteBackRaceTest` — same pattern for MicroProfile Fault Tolerance
- [x] All existing resilience4j tests pass
- [x] All existing microprofile-fault-tolerance tests pass

### DIFF
--- a/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
+++ b/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.smallrye.faulttolerance.api.CircuitBreakerMaintenance;
 import io.smallrye.faulttolerance.api.CircuitBreakerState;
@@ -231,6 +232,10 @@ public class FaultToleranceProcessor extends BaseProcessorSupport
         // run this as if we run inside try / catch so there is no regular Camel error handler
         exchange.setProperty(ExchangePropertyKey.TRY_ROUTE_BLOCK, true);
         CircuitBreakerTask task = (CircuitBreakerTask) taskFactory.acquire(exchange, callback);
+        // guard to prevent the worker thread from writing results back to the original exchange
+        // after a timeout has triggered fallback processing on the caller thread
+        AtomicBoolean exchangeWriteGuard = new AtomicBoolean(false);
+        task.exchangeWriteGuard = exchangeWriteGuard;
         CircuitBreakerFallbackTask fallbackTask = null;
 
         try {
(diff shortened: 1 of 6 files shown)


gnodet (approved):
### Review: CAMEL-24134 — Circuit Breaker EIP - prevent timed-out worker from writing back to exchange

**Verdict: Approve** ✅

Well-designed concurrency fix. The race condition is real: when a circuit breaker timeout fires, the fallback runs on the caller thread while the worker may still be executing on its thread. Without a guard, the worker can overwrite the fallback result via `ExchangeHelper.copyResults()` or `exchange.setException()`.

**Guard design is correct:**

The shared `AtomicBoolean exchangeWriteGuard` provides mutual exclusion on write-back:
- **Worker** uses `compareAndSet(false, true)` — atomic claim-or-skip before both the success path (`copyResults`) and the exception path (`setException`)
- **Fallback** uses `set(true)` — unconditionally claims the guard before processing

This handles all orderings correctly:
1. **Fallback first** → `set(true)` → worker's CAS fails → worker skips write-back ✓
2. **Worker first** → CAS succeeds → worker writes, then fallback `set(true)` and overwrites → final state is the fallback result, which is the correct outcome ✓
3. **No timeout** → worker CAS succeeds → writes normally, no fallback invoked ✓

**Both implementations are consistent:**
- `ResilienceProcessor`: guard shared between `CircuitBreakerTask` and `CircuitBreakerFallbackTask`, passed to `processTask()` as a parameter. The `call()` and `get()` methods both forward it correctly.
- `FaultToleranceProcessor`: guard set on task field, fallback handled in `process()` catch block with `exchangeWriteGuard.set(true)` before fallback processing.
- Both include null guards for defensive safety, and both include the CAMEL-24133 fix (`release(copy)` instead of `release(exchange)`).

**Tests are well-designed:**
Both `*TimeoutWriteBackRaceTest` classes create a deterministic race (3s sleep vs 500ms timeout), verify the fallback result wins, then use a `CountDownLatch` to wait for the worker to fully complete and re-verify no late write-back occurred.

The test's `process()` lambda has misleading indentation — `exchange.getIn().setBody("Worker result")` appears to be outside the `try` block visually but is syntactically inside it. The formatter should fix this.

PR #24800, opened 2026-07-16
CAMEL-24136: Fix FT circuit breaker exchange properties, bulkhead, dead option, docs

davsclaus:
## Summary

Fixes medium-severity findings from a review of the Circuit Breaker EIP in `camel-microprofile-fault-tolerance` ([CAMEL-24136](https://issues.apache.org/jira/browse/CAMEL-24136)).

- **M1+M2 — Exchange property contract**: `RESPONSE_TIMED_OUT` is now set in all timeout paths (with and without fallback). Guard exceptions (timeout, CB open) are now stored on the exchange before fallback invocation so the fallback can see them via `EXCEPTION_CAUGHT`. Removed ~20 lines of unreachable dead code from `CircuitBreakerFallbackTask`.
- **Bulkhead enablement bug**: `FaultToleranceReifier.configureBulkhead()` never called `setBulkheadEnabled(true)` on the target config, so the bulkhead was never actually wired onto the `TypedGuard`. Fixed.
- **M4 — Dead `timeoutPoolSize` option**: Deprecated across model, configuration, DSL, and JMX with `@Deprecated(since = "4.22.0")`. The option has been inert since the CAMEL-21857 TypedGuard migration.
- **M3 (partial) — Misleading JMX description**: `getFailureRate()` description now says "configured failure ratio threshold" instead of "current failure rate in percentage" (it returns the static config value, not live metrics). Documented the FT dev console metrics limitation vs resilience4j (SmallRye API does not expose per-breaker counters).
- **M6 — Documentation errors**: Fixed incorrect timeout default claim in `circuitBreaker-eip.adoc`, wrong option names in `resilience4j-eip.adoc` and `fault-tolerance-eip.adoc`, incorrect no-fallback exception behavior description, and a misordered YAML example.
- **Tests**: 2 new tests covering `RESPONSE_TIMED_OUT` and `EXCEPTION_CAUGHT` visibility in both no-fallback and with-fallback timeout scenarios. All 22 module tests pass.

## Test plan

- [x] `mvn verify` passes in `camel-microprofile-fault-tolerance` (22 tests, 0 failures)
- [ ] CI green on all platforms
- [ ] Verify existing circuit breaker tests in `camel-resilience4j` are unaffected


gnodet (approved):
**APPROVE** — Thorough cleanup of the Fault Tolerance circuit breaker implementation. Each fix is well-targeted and addresses a real issue.

**Fixes reviewed:**

1. **Bulkhead enablement bug** (`FaultToleranceReifier.java`): `configureBulkhead()` was setting `maxConcurrentCalls` and `waitingTaskQueue` but never calling `setBulkheadEnabled(true)` on the target config. The bulkhead was therefore never actually wired onto the `TypedGuard`. Bug introduced in CAMEL-21857 (TypedGuard migration). One-line fix, high impact.

2. **Timeout exchange properties** (`FaultToleranceProcessor.java`): The old code tried to set `RESPONSE_TIMED_OUT` and other properties inside `CircuitBreakerFallbackTask.call()`, but that code was guarded by `if (fallbackProcessor == null)` — unreachable since the task is only instantiated when a fallback exists. The new code correctly handles timeout-without-fallback in a dedicated `catch (TimeoutException e)` block at the `process()` level, and stores the guard exception on the exchange before invoking the fallback so it's visible via `EXCEPTION_CAUGHT`. Dead code cleanup is clean — ~20 lines removed with equivalent behavior preserved in the right location.

3. **`timeoutPoolSize` deprecation**: Properly deprecated across model (`FaultToleranceConfigurationCommon`, `FaultToleranceConfigurationDefinition`), runtime config (`FaultToleranceConfiguration`), JMX (`FaultToleranceProcessor`), and generated metadata. All use `@Deprecated(since = "4.22.0")` with clear deprecation notes referencing CAMEL-21857.

4. **JMX `getNumberOfSlowSuccessfulCalls()` bug** (`ResilienceProcessor.java`): Was calling `getNumberOfSlowCalls()` instead of `getNumberOfSlowSuccessfulCalls()` — obvious copy-paste error. Good catch.

5. **`ResilienceConsole` display fix**: `fr > 0` → `fr >= 0` is correct (0% failure rate should display the full stats line); the `else` branch now properly uses the `fc` variable instead of hardcoding `failure: 0`.

6. **JMX description fix**: `getFailureRate()` description updated from "current failure rate in percentage" to "configured failure ratio threshold (0.0-1.0)" — it returns the static config value, not live metrics.

7. **Documentation**: Fixed incorrect default timeout claim, corrected option names (`resilience4jConfiguration`, `configuration`), fixed YAML example ordering (steps must come before `onFallback`), and clarified no-fallback/open-circuit behavior including the `throwExceptionWhenHalfOpenOrOpenState` option.

8. **Tests**: Two new focused tests covering the exact scenarios fixed — timeout without fallback (verifying all 4 exchange properties + exception type) and timeout with fallback (verifying `EXCEPTION_CAUGHT` visibility to the fallback processor and `RESPONSE_TIMED_OUT` on the result exchange). Both use the standard `CamelTestSupport` pattern.

**Checklist:**
- [x] Tests included (2 new targeted tests)
- [x] Generated metadata regenerated
- [x] No public API breakage (deprecation, not removal)
- [x] Documentation updated (3 EIP docs + generated metadata)
- [x] Commit convention followed
- [x] CI: pending

---

PR #24802, opened 2026-07-16
CAMEL-24112: camel-rest-openapi - Fix path parameters leaking into query string with multi-param operations

davsclaus:
## Summary

Fixes a logic bug in `RestOpenApiEndpoint.determineEndpointParameters()` where path parameters leak into the query string when an operation has more than one parameter.

The inner loop was logically inverted: an endpoint parameter entry was added to `nestedParameters` inside the inner per-operation-parameter loop whenever *any* operation parameter did not clash, instead of only when *no* parameter clashed. This meant the path-param filtering only worked for single-parameter operations.

**Example:** `GET /users/{userId}/orders/{orderId}` with `userId=1&orderId=2` would produce a request to `GET /users/1/orders/2?userId=1&orderId=2` — the path params leaked as query params.

**Fix:** Replace the nested loop with a stream `anyMatch` check so the clash is evaluated across all operation parameters before deciding whether to include the entry.

Introduced in commit 94847d8c4a84 (CAMEL-17049), unfixed since 2021 due to lack of test coverage for the multi-parameter case.

## Test plan

- [x] Added `shouldNotLeakPathParametersAsQueryParameters` — two path params, verifies neither leaks and a non-path param is preserved
- [x] Added `shouldKeepEndpointParametersWhenNoPathParamClash` — only query params in operation, verifies all endpoint params are preserved
- [x] All 137 existing tests pass

### DIFF
--- a/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
+++ b/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
@@ -683,13 +683,12 @@ public final class RestOpenApiEndpoint extends DefaultEndpoint {
         if (this.parameters != null) {
             if (operation.getParameters() != null) {
                 for (Map.Entry<String, Object> entry : this.parameters.entrySet()) {
-                    for (Parameter param : operation.getParameters()) {
-                        // skip parameters that are part of the operation as path as otherwise
-                        // it will be duplicated as query parameter as well
-                        boolean clash = "path".equals(param.getIn()) && entry.getKey().equals(param.getName());
-                        if (!clash) {
-                            nestedParameters.put(entry.getKey(), entry.getValue());
-                        }
+                    // skip parameters that are part of the operation as path as otherwise
+                    // it will be duplicated as query parameter as well
+                    boolean clash = operation.getParameters().stream()
+                            .anyMatch(p -> "path".equals(p.getIn()) && entry.getKey().equals(p.getName()));
+                    if (!clash) {
+                        nestedParameters.put(entry.getKey(), entry.getValue());
                     }
                 }
             } else {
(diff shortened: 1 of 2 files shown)


gnodet (approved):

## Review — CAMEL-24112: camel-rest-openapi — Fix path parameters leaking into query string

Clear, correct bug fix. The logic error in the original nested loop is well-analyzed in the PR description, and the fix is clean.

### The bug

The inner loop was logically inverted — it added an endpoint parameter entry to `nestedParameters` inside the per-operation-parameter loop whenever *any* single operation parameter did not clash, instead of only when *no* operation parameter clashed:

```java
// Old — adds entry for EACH non-clashing param (wrong):
for (Parameter param : operation.getParameters()) {
    boolean clash = "path".equals(param.getIn()) && entry.getKey().equals(param.getName());
    if (!clash) {
        nestedParameters.put(entry.getKey(), entry.getValue());  // fires N-1 times
    }
}
```

With 2 path params (`userId`, `orderId`), `userId` clashes with itself but NOT with `orderId`, so it gets added anyway. Only worked correctly for single-parameter operations.

### The fix

```java
// New — checks ALL params before deciding (correct):
boolean clash = operation.getParameters().stream()
        .anyMatch(p -> "path".equals(p.getIn()) && entry.getKey().equals(p.getName()));
if (!clash) {
    nestedParameters.put(entry.getKey(), entry.getValue());
}
```

Semantically equivalent to the original intent, now correctly implemented. ✅

### Tests

Both new tests are well-constructed:
- `shouldNotLeakPathParametersAsQueryParameters` — exactly the bug scenario (2 path params + 1 non-path param, verifies path params excluded and non-path param preserved)
- `shouldKeepEndpointParametersWhenNoPathParamClash` — regression test for the non-bug case (query-only operation params, verifies all endpoint params preserved)

Both use AssertJ assertions per project conventions. ✅

LGTM — nice find and clean fix.

Croway (approved):

PR #24803, opened 2026-07-16
CAMEL-24113: Fix rest-openapi producer cross-contamination for same operation

davsclaus:
## Summary

- Fix `RestOpenApiEndpoint.createProducerFor()` to include `produces`, `consumes`, `queryParameters`, and `producerComponentName` in the delegate `rest:` endpoint URI, so each unique combination gets its own cached endpoint
- Previously, only `host` was included in the URI cache key, causing two rest-openapi endpoints targeting the same operation (same HTTP method + base path + URI template + host) to share one cached `RestEndpoint` singleton — the last `configureProperties()` call silently overwrote the first's configuration
- Add two tests verifying that different `produces` values and different `queryParameters` values produce distinct delegate endpoints

## Test plan

- [x] Existing 21 unit tests in `RestOpenApiEndpointV3Test` pass
- [x] New test `shouldNotCrossContaminateProducersForSameOperation` verifies two endpoints with different `produces` get distinct delegate endpoints
- [x] New test `shouldNotCrossContaminateQueryParameters` verifies two endpoints with different literal query parameter values get distinct delegate endpoints
- [x] Full module test suite (139 tests) passes

### DIFF
--- a/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
+++ b/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
@@ -559,8 +559,40 @@ public final class RestOpenApiEndpoint extends DefaultEndpoint {
         boolean hasHost = params.containsKey("host");
         String basePath = determineBasePath(openapi);
         String componentEndpointUri = "rest:" + method + ":" + basePath + ":" + uriTemplate;
+
+        // include all distinguishing options in the URI so each unique combination
+        // gets its own cached endpoint and avoids cross-contamination (CAMEL-24113)
+        StringBuilder query = new StringBuilder();
         if (hasHost) {
-            componentEndpointUri += "?host=" + params.get("host");
+            query.append("host=").append(params.get("host"));
+        }
+        if (params.containsKey("producerComponentName")) {
+            if (!query.isEmpty()) {
+                query.append('&');
+            }
+            query.append("producerComponentName=").append(params.get("producerComponentName"));
+        }
+        if (params.containsKey("consumes")) {
+            if (!query.isEmpty()) {
+                query.append('&');
+            }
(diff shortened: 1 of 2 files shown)


gnodet (approved):

## Review — CAMEL-24113: Fix rest-openapi producer cross-contamination for same operation

Well-analyzed bug fix. The endpoint cache key (`componentEndpointUri`) previously only included `host`, so two rest-openapi endpoints targeting the same operation with different `produces`, `consumes`, `queryParameters`, or `producerComponentName` shared one cached `RestEndpoint` — the last `configureProperties()` call silently overwrote the first's settings.

### The fix

Including all distinguishing options in the delegate URI so each unique combination gets its own cached endpoint. Verified the coverage:

| Parameter | In URI key? | Notes |
|-----------|-------------|-------|
| `host` | ✅ | Was already there (only one before) |
| `producerComponentName` | ✅ | New |
| `consumes` | ✅ | New |
| `produces` | ✅ | New |
| `queryParameters` | ✅ | New, URL-encoded. Includes resolved literal values (via `literalQueryParameterValue`) |

I initially wondered whether `queryParameters` would just be a template from the spec (same for both endpoints), but traced through `queryParameter()` → `literalQueryParameterValue()`: when `this.parameters` contains the param name, it resolves to `name=encodedLiteralValue` rather than `name={name}`. So the URI key correctly differentiates literal-value variants. ✅

The `UnsafeUriCharactersEncoder.encode()` on `queryParameters` is correct since it may contain `=` and `&` characters. ✅

### Tests

Both tests use real `DefaultCamelContext` (started/stopped in try-finally) and verify:
1. `shouldNotCrossContaminateProducersForSameOperation` — different `produces` → 2 distinct delegate endpoints with correct media types ✅
2. `shouldNotCrossContaminateQueryParameters` — different literal query param values → 2 distinct delegate endpoints with correct resolved values ✅

Both use AssertJ assertions per project conventions. ✅

LGTM — nice companion fix to #24802.

PR #24804, opened 2026-07-16
Fix flaky tests: Disruptor, Seda, Scheduler (batch 16)

gnodet:

## Summary

Fix 3 flaky tests identified by Develocity, each with a distinct root cause:

### 1. `DisruptorBlockWhenFullTest.testDisruptorExceptionWhenFull` (11 flaky / 715 runs — 1.5%)

**Root cause:** The `delay(100)` EIP defaults to async mode. The consumer's `onEvent()` returns in microseconds because the delay is scheduled on a background executor — it does not block the ring buffer consumption. Under CI load, the buffer never fills up and `tryNext()` succeeds instead of throwing `InsufficientCapacityException`.

**Fix:** Add `.syncDelayed()` so the consumer blocks for the full 100ms inside `onEvent()`, keeping the ring buffer full long enough for the producer to reliably hit the exception.

### 2. `SedaBlockWhenFullTest.testAsyncSedaBlockingWhenFull` (65 flaky / 716 runs — 9.1%)

**Root cause:** With `blockWhenFull=true` and async sends, all messages eventually arrive (they block and wait for queue space). The exact-count assertion (`setExpectedMessageCount(2)`) opens its `CountDownLatch` after 2 messages, then races against additional messages arriving before `assertEquals` runs — causing "expected 2 but was 3+".

**Fix:** Switch to `expectedMinimumMessageCount(2)` which uses a `>=` check. The test's purpose is verifying that blocked messages eventually succeed, not that exactly N arrive.

### 3. `SchedulerNoPolledMessagesTest` (75 flaky / 716 runs — 10.5%)

**Root cause:** The `arrives().afterPrevious()` MockEndpoint API has a subtle race: when computing "after previous", it actually compares with the **next** message's timestamp (if already received), not the previous one. A 4th message arriving ~100ms after the 3rd violates the 200ms lower bound intended for the backoff gap between messages 1→2.

Prior widening attempts (4× wider windows) didn't help because the race is structural, not a matter of window width.

**Fix:** Replace the fragile timing API with direct timestamp comparison on already-received (immutable) exchanges after `assertIsSatisfied()` completes. This is immune to additional messages arriving concurrently.

## Test plan

- [x] `DisruptorBlockWhenFullTest` — 2 tests pass (both blocking and exception variants)
- [x] `SedaBlockWhenFullTest` — 4 tests pass (all variants including the fixed async one)
- [x] `SchedulerNoPolledMessagesTest` — passes with direct timestamp assertions
- [ ] CI green on the full build

### DIFF
--- a/components/camel-disruptor/src/test/java/org/apache/camel/component/disruptor/DisruptorBlockWhenFullTest.java
+++ b/components/camel-disruptor/src/test/java/org/apache/camel/component/disruptor/DisruptorBlockWhenFullTest.java
@@ -44,7 +44,11 @@ public class DisruptorBlockWhenFullTest extends CamelTestSupport {
         return new RouteBuilder() {
             @Override
             public void configure() {
-                from(DEFAULT_URI).delay(DELAY).to(MOCK_URI);
+                // syncDelayed() is required so the consumer blocks for
+                // the full delay inside onEvent(), keeping the ring buffer
+                // full long enough for the "exception when full" test to
+                // hit InsufficientCapacityException reliably.
+                from(DEFAULT_URI).delay(DELAY).syncDelayed().to(MOCK_URI);
             }
         };
     }
--- a/core/camel-core/src/test/java/org/apache/camel/component/scheduler/SchedulerNoPolledMessagesTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/scheduler/SchedulerNoPolledMessagesTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.scheduler;
 
+import java.util.Date;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.ContextTestSupport;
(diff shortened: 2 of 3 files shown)


gnodet (commented):
LGTM — three well-analyzed flaky test fixes, each targeting the actual root cause.

**DisruptorBlockWhenFullTest** — `.syncDelayed()` makes the consumer block for the full 100ms
inside `onEvent()`, keeping the ring buffer full long enough for `tryNext()` to reliably hit
`InsufficientCapacityException`. Without it, the async delay returns immediately and the buffer
drains too fast.

**SedaBlockWhenFullTest** — `expectedMinimumMessageCount(QUEUE_SIZE + 1)` is the correct
semantic: with `blockWhenFull=true`, all messages *eventually* arrive. The exact-count assertion
raced against additional messages flowing through the pipeline after the latch opened.

**SchedulerNoPolledMessagesTest** — Direct timestamp comparison on already-received exchanges
eliminates the `arrives().afterPrevious()` race entirely. Verified that `MockEndpoint` stores
`RECEIVED_TIMESTAMP` as `new Date()` (camel-mock `MockEndpoint.java:1848`), so the
`getProperty(Exchange.RECEIVED_TIMESTAMP, Date.class)` access is safe.

gnodet (commented):
Nice batch — all three fixes target the actual mechanism of the flakiness rather than widening timeout windows. Clean work.

### What's good

- **Root cause depth is outstanding.** Each fix targets the specific mechanism: async delay semantics in Disruptor, exact-vs-minimum count race in SEDA, and the `arrives()` API race in Scheduler. The Develocity data provides empirical backing.
- **The SchedulerNoPolledMessagesTest rewrite is a genuine upgrade.** After the prior widening attempt (commit `3009c0da4c7e`), this PR takes the fundamentally better approach: wait for messages, then examine timestamps on a stable list. Eliminates the entire class of race.
- **Comment quality is excellent.** Each change explains *why*, not just *what*.

### Minor suggestions

**SchedulerNoPolledMessagesTest — dropped upper bound on backoff gap**
`core/camel-core/src/test/java/org/apache/camel/component/scheduler/SchedulerNoPolledMessagesTest.java`

The old code checked `between(200, 5000)` for the backoff gap. The new code only checks `gap12 >= 200` (lower bound only). Dropping the upper bound is the right call for CI stability, but if you want to catch a hypothetical regression where the scheduler ignores the backoff entirely and waits far too long, a generous upper bound could help:

```java
assertTrue(gap12 <= 10000,
        "Backoff gap should not exceed 10s (expected ~1000 ms), was " + gap12 + " ms");
```

Truly optional — the test's purpose is verifying backoff kicks in, not capping it.

### Verified claims

- ✅ `delay()` defaults to async: `DelayProcessorSupport.java:44` — `private boolean asyncDelayed = true`
- ✅ `Exchange.RECEIVED_TIMESTAMP` exists and is set by MockEndpoint
- ✅ No `Thread.sleep()` introduced
- ✅ `MockEndpoint.assertIsSatisfied(context, 30, TimeUnit.SECONDS)` used correctly per project guidelines
- ✅ All new imports are used; no unused imports

### CI status

Two jobs failed (`build (17, false)` and `build (25, false)`) — both on `camel-smb` integration tests, unrelated to this PR.

### Checklist

| Check | Status |
|-------|--------|
| Tests | ✅ This PR IS the test fix |
| MockEndpoint usage | ✅ Correct per guidelines |
| Documentation | N/A (test-only) |
| Commit convention | ✅ |
| Public API / backward compat | N/A (test-only) |

---

oscerd (approved):
LGTM — all three attack the actual race rather than widening a window:
- `DisruptorBlockWhenFullTest`: `.syncDelayed()` makes the delay block on the event-loop thread so the ring buffer stays full and `InsufficientCapacityException` fires reliably (`delay()` is async by default).
- `SedaBlockWhenFullTest`: `expectedMessageCount` → `expectedMinimumMessageCount` — the correct semantic for `blockWhenFull=true` (all messages eventually arrive), removing the exact-count-vs-latch race.
- `SchedulerNoPolledMessagesTest`: replaces the fragile `arrives().afterPrevious()` with a direct `RECEIVED_TIMESTAMP` comparison on already-received exchanges — the cleanest of the batch.

No `Thread.sleep` added. The red `build (25)` is an unrelated camel-smb Testcontainers IT flake — camel-smb is only pulled into the reactor as a transitive dependent of the modified `camel-core` (which itself built and tested fine), and `build (17)` was cancelled by matrix fail-fast, not an independent failure. A re-run should clear it. (Minor: the `SchedulerNoPolledMessagesTest` change drops the old upper bound on the backoff gap, which you already flagged as optional in your own note.)

davsclaus (approved):

gnodet (commented):

**Re-review after new commits** — all three fixes look solid.

### DisruptorBlockWhenFullTest
`.syncDelayed()` is the right fix. Without it, the default async delay causes messages to pass through the ring buffer too quickly, so the buffer may not be full when the test expects `InsufficientCapacityException`. With `.syncDelayed()`, the consumer blocks for the full delay period, keeping the slot occupied. The comment explaining the rationale is a nice touch.

### SchedulerNoPolledMessagesTest
Excellent analysis of the `arrives().afterPrevious()` race condition. The direct timestamp comparison using `Exchange.RECEIVED_TIMESTAMP` is more reliable and avoids the subtle issue where a 4th message arriving ~100ms after the 3rd violates the lower bound intended for the backoff gap (msg 1→2). The wide timing windows (gap01 ≤ 2000ms, gap12 ≥ 200ms) are appropriately generous for CI environments.

### SedaBlockWhenFullTest
`expectedMinimumMessageCount(QUEUE_SIZE + 1)` is correct. With `blockWhenFull=true` and async sends, all messages eventually arrive — they block and wait for queue space rather than failing. An exact-count assertion races against the remaining messages still flowing through the 130ms delay pipeline. Good explanatory comment.

All changes are well-documented with comments explaining the reasoning. LGTM ✅

PR #24805, opened 2026-07-16
CAMEL-24187: camel-core - CamelContext.getEndpoint() never caches a uri containing a % character

Croway:
## Summary

`AbstractCamelContext.doGetEndpoint()` normalizes the uri once for its cache lookup (`NormalizedUri.newNormalizedUri(uri, true)` - "already normalized"), but on a cache miss it handed that *already normalized* uri to `addEndpointToRegistry()`, which called `getEndpointKey(uri, endpoint)` -> `NormalizedUri.newNormalizedUri(uri, false)` - "not normalized yet, please normalize it" - re-running `EndpointHelper.normalizeEndpointUri()` a **second** time before storing the registry key.

`EndpointHelper.normalizeEndpointUri()` (via `URISupport.normalizeUri` / `UnsafeUriCharactersEncoder`) is not idempotent for any uri containing a literal `%`: it encodes it to `%25` on every pass, so a second pass turns `%41` into `%2541`. Because the *stored* key ends up double-normalized while the *lookup* key computed fresh on a later call is only single-normalized, they never match again - `CamelContext#getEndpoint(String)` creates and starts a **brand-new** `Endpoint` on every single call for any uri containing a `%`, instead of reusing the cached singleton.

- Fixes CAMEL-24187
- Discovered while investigating CAMEL-24171: camel-plc4x S7 tag addresses conventionally contain a literal `%` (e.g. `RAW(%DB1.DBX0.0:BOOL)`), unlike Modbus' (`RAW(coil:1)`), which is why only the S7 route reloaded its `DefaultPlcDriverManager` on every poll while the structurally identical Modbus route did not. camel-plc4x itself is not at fault - the same symptom applies to any component whose endpoint uri legitimately contains a `%` and is re-resolved by string on every use, as `pollEnrich`/`toD`/`recipientList` do.

## Fix

Added a `normalized`-aware overload of `getEndpointKey()`/`addEndpointToRegistry()` so only the `doGetEndpoint()` call site (whose uri is provably already normalized) skips the redundant second normalization pass. The public `hasEndpoint(String)`, `addEndpoint(String, Endpoint)` and `removeEndpoints(String)` APIs still receive raw, caller-supplied uris directly and keep normalizing exactly as before - flipping the existing method wholesale would have silently broken those three call sites.

## Test plan

- [x] New camel-core test `DefaultEndpointRegistryTest#testGetEndpointIsCachedForUriContainingPercentCharacter` isolates the bug with no plc4x involvement (`controlbus:route?routeId=RAW(%41)&action=status` resolved twice returns two different instances before the fix, the same instance after).
- [x] New camel-plc4x test `Plc4XPollEnrichDoStartReproducerTest` reproduces the real-world consequence: with the fix, both a S7-shaped and a Modbus-shaped `pollEnrich` route only start their `Plc4XEndpoint` once, no matter how many poll cycles run.
- [x] Full `camel-core` test suite run locally with the fix: 7300 tests, 18 failures / 3 errors - all individually verified to fail identically with the fix reverted (pre-existing on `origin/main`, unrelated to this change). Zero regressions introduced.

---

_Claude Sonnet 5 on behalf of Croway_

### DIFF
--- a/components/camel-plc4x/src/test/java/org/apache/camel/component/plc4x/Plc4XPollEnrichDoStartReproducerTest.java
+++ b/components/camel-plc4x/src/test/java/org/apache/camel/component/plc4x/Plc4XPollEnrichDoStartReproducerTest.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.plc4x;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.Endpoint;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
+import org.apache.camel.test.junit6.CamelTestSupport;
+import org.apache.camel.util.PropertiesHelper;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.sameInstance;
+
+/**
+ * Reproducer for CAMEL-24171: the reporter observed that, on a route shaped like
+ *
+ * <pre>
+ * from(timer("s7").period(5000))
+ *         .pollEnrich(plc4x("s7://localhost:102").autoReconnect(true).tags(Map.of("var1", "RAW(%DB1.DBX0.0:BOOL)")), 5000)
+ * </pre>
+ *
+ * {@code Plc4XEndpoint#doStart()} runs again on every poll cycle (reloading {@code DefaultPlcDriverManager}), while a
+ * structurally identical Modbus route only runs it once.
+ * <p>
+ * {@link #directGetEndpointCachingProbe()} isolates the actual root cause: it is not specific to S7, pollEnrich, or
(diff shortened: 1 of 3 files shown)


gnodet (approved):

## Review — CAMEL-24187: CamelContext.getEndpoint() never caches URIs containing %

Excellent bug analysis and clean, minimal fix. The root cause is clearly traced from the real-world plc4x S7 symptom down to the non-idempotent double-normalization in `AbstractCamelContext`.

### The bug

Verified the flow on `main`:

1. `doGetEndpoint()` line ~810: `uri = EndpointHelper.normalizeEndpointUri(uri)` — first normalization
2. Line ~819: `NormalizedUri.newNormalizedUri(uri, true)` — lookup key, `true` = "already normalized"
3. On cache miss, line ~876: `addEndpointToRegistry(uri, answer)` → `getEndpointKey(uri, endpoint)` → `NormalizedUri.newNormalizedUri(uri, false)` — second normalization

For URIs with `%`, the second pass turns `%41` into `%2541`, so the stored key never matches the lookup key. Every subsequent `getEndpoint()` call creates a new endpoint. ✅

### The fix

Adding a `normalized` boolean parameter to `addEndpointToRegistry()` and `getEndpointKey()` is the right approach:

- Only the `doGetEndpoint()` call site passes `normalized=true` (its URI is provably already normalized at line ~810)
- The existing no-arg overloads delegate with `normalized=false`, preserving backward compatibility for `hasEndpoint()`, `addEndpoint()`, `removeEndpoints()` which receive raw, caller-supplied URIs
- Javadoc on the new parameter clearly explains the non-idempotency issue

This is minimal, targeted, and doesn't change any existing public API behavior. ✅

### Tests

**Core test** (`DefaultEndpointRegistryTest#testGetEndpointIsCachedForUriContainingPercentCharacter`):
- Uses `controlbus:route?routeId=RAW(%41)&action=status` — clean isolation of the bug with no component dependencies
- Verifies both `assertSame` (same instance) and endpoint count (no registry accumulation)
- ✅

**Plc4x reproducer** (`Plc4XPollEnrichDoStartReproducerTest`):
- `directGetEndpointCachingProbe` — verifies `getEndpoint()` caching for S7-style URI
- `doStartRunsOnlyOnceEvenThoughBothRoutesArePolledManyTimes` — reproduces the real-world consequence: S7 route (`%` in tag) and Modbus route (no `%`) both verify `doStart()` runs only once across 10+ poll cycles
- `CountingPlc4XComponent` subclass counts `doStart()` invocations without modifying production code — good pattern
- Uses `MockEndpoint.assertIsSatisfied(context, 5, TimeUnit.SECONDS)` correctly (not Awaitility)
- ✅

LGTM — well-analyzed root cause, minimal fix, excellent test coverage at both the core and component level.

davsclaus:
The PR has a plc4j test as well

davsclaus (approved):

Croway:
> The PR has a plc4j test as well

yeah, that is expected, the issue was reported in https://issues.apache.org/jira/browse/CAMEL-24171

Croway:
the `SmbStreamDownloadIT.testStreamDownload` test failure is unrelated, merging

PR #24806, opened 2026-07-16
CAMEL-24114: camel-rest-openapi - Request validator also checks exchange variables and endpoint literal parameters

davsclaus:
## Summary

`DefaultRequestValidator` rejected requests whose required query parameters were supplied via endpoint URI literals or exchange variables, because it only checked `message.getHeader()`.

The fix extends the query parameter validation in `DefaultRequestValidator.validate()` to mirror the resolution order used by `RestProducer.createQueryParameters()`:
1. Message header
2. Exchange variable
3. Literal endpoint parameter (from the endpoint URI)

Changes:
- `DefaultRequestValidator` — added `endpointParameters` field and extended the required query parameter check to also consult exchange variables and endpoint literal parameters
- `RestOpenApiEndpoint.configureRequestValidator()` — passes the endpoint's remaining parameters to the validator
- `RestOpenApiRequestValidationTest` — added two new parameterized tests covering literal endpoint URI query params and exchange variables with validation enabled

Fixes: [CAMEL-24114](https://issues.apache.org/jira/browse/CAMEL-24114)

### DIFF
--- a/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
+++ b/components/camel-rest-openapi/src/main/java/org/apache/camel/component/rest/openapi/RestOpenApiEndpoint.java
@@ -864,6 +864,7 @@ public final class RestOpenApiEndpoint extends DefaultEndpoint {
             OpenAPI openAPI, Operation operation, String method, String uriTemplate) {
         DefaultRequestValidator answer = new DefaultRequestValidator();
         answer.setOperation(new RestOpenApiOperation(operation, method, uriTemplate));
+        answer.setEndpointParameters(parameters);
         return answer;
     }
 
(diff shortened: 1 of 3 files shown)


gnodet (approved):
LGTM — clean, well-scoped fix.

The fallback chain (header → exchange variable → endpoint literal parameter) correctly mirrors
the resolution order in `RestProducer.createQueryParameters()`, so the validator now accepts
required query parameters regardless of how they are supplied.

Verified:
- `parameters` field in `RestOpenApiEndpoint` holds the literal URI parameters (e.g. `status=available`)
- Null-safe: `setEndpointParameters` defaults to `Collections.emptyMap()`
- Non-query params like `requestValidationEnabled` won't false-positive since the validator only
  checks names that match the OpenAPI operation's required query parameters
- Both new test scenarios (endpoint URI literal + exchange variable) cover the two new resolution paths

PR #24807, opened 2026-07-16
CAMEL-24088: camel-ftp - streamDownload treats refused RETR as success

davsclaus:
## Summary

Fixes two related bugs in `FtpOperations` where the `streamDownload=true` path never checks the FTP client's success indicators, causing silent data loss:

- **`retrieveFileToStreamInBody`**: `FTPClient.retrieveFileStream()` returns `null` when the server refuses RETR (550/450/425), but the code unconditionally set `result = true`. The consumer then routes the exchange with a null body and the process strategy commits — moving or deleting the remote file that was never downloaded. Fixed by setting `result = is != null` and only populating the body/header when the stream is available.

- **`releaseRetrievedFileResources`**: `FTPClient.completePendingCommand()` returns `false` when the transfer did not complete (server sent 426/451), but the return value was discarded. A truncated stream would be silently accepted and the commit would still run. Fixed by checking the return value and throwing `GenericFileOperationFailedException` on failure.

These fixes also re-enable `ignoreFileNotFoundOrPermissionError` for streamDownload mode — it was previously dead code because `retrieved` was always `true`.

## Test plan

- [x] New unit test `FtpOperationsStreamDownloadTest` with 5 test cases:
  - `testRetrieveFileStreamReturnsNullOnServerError` — verifies `retrieveFile` returns `false` when server refuses RETR
  - `testRetrieveFileStreamSuccessReturnsTrue` — verifies normal success path still works
  - `testReleaseResourcesThrowsOnFailedCompletion` — verifies exception on failed `completePendingCommand`
  - `testReleaseResourcesSucceedsOnCompletedTransfer` — verifies normal release path
  - `testReleaseResourcesSkipsWhenNoStream` — verifies no-op when no stream header present

### DIFF
--- a/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/FtpOperations.java
+++ b/components/camel-ftp/src/main/java/org/apache/camel/component/file/remote/FtpOperations.java
@@ -442,7 +442,11 @@ public class FtpOperations implements RemoteFileOperations<FTPFile> {
         if (is != null) {
             try {
                 IOHelper.close(is);
-                client.completePendingCommand();
+                boolean ok = client.completePendingCommand();
+                if (!ok) {
+                    throw new GenericFileOperationFailedException(
+                            client.getReplyCode(), client.getReplyString());
+                }
             } catch (IOException e) {
                 throw new GenericFileOperationFailedException(e.getMessage(), e);
             }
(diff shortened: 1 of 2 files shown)


gnodet (approved):
LGTM — important fix for silent data loss in FTP streamDownload mode.

**`retrieveFileToStreamInBody`**: `FTPClient.retrieveFileStream()` returns `null` on server refusal
(550/450/425), but the old code unconditionally set `result = true` (line 487 on main). This
caused the consumer to route an exchange with a null body, and the process strategy would then
commit — moving or deleting the remote file that was never actually downloaded. The fix
(`result = is != null` + guarded body/header) is clean and correct.

**`releaseRetrievedFileResources`**: The discarded `completePendingCommand()` return value meant a
truncated transfer (426/451) was silently accepted. The new check-and-throw is correct —
`GenericFileOperationFailedException` extends `RuntimeCamelException` (a `RuntimeException`),
so it won't be caught by the existing `IOException` catch clause in the same try block.

Side benefit: `ignoreFileNotFoundOrPermissionError` was effectively dead code for streamDownload
mode since `retrieved` was always `true` — this fix re-enables it.

Tests cover all five paths well (null stream, success stream, failed completion, success completion,
no-stream release).

PR #24809, opened 2026-07-16
[backport camel-4.18.x] CAMEL-24046: camel-sql - JdbcCachedMessageIdRepository must not cache a key whose insert failed, make cache thread-safe

Croway:
## Backport of #24657

Cherry-pick of #24657 onto `camel-4.18.x`.

**Original PR:** #24657 - CAMEL-24046: camel-sql - JdbcCachedMessageIdRepository must not cache a key whose insert failed, make cache thread-safe
**Original author:** @Croway

### Original description

JIRA: https://issues.apache.org/jira/browse/CAMEL-24046

Two related defects in `JdbcCachedMessageIdRepository`:

1. `add(key)` put the key into the in-memory cache *before* calling `super.add(key)`. If the database insert failed (outage, PK race with another node), the exception propagated and the idempotent consumer never processed the message — but the key stayed cached, so every redelivery was rejected as a duplicate. The message was never processed and never stored until JVM restart.
2. The cache was a plain `HashMap` mutated from concurrent consumer threads with no synchronization and a non-volatile field (a runtime `reload()` might never become visible to other threads); concurrent puts during resize can lose entries, causing duplicate processing. The hit/miss counters were unsynchronized ints.

Changes:
- Populate the cache only after the database insert succeeded
- Use `ConcurrentHashMap` with a volatile field for safe publication on `reload()`, and `AtomicInteger` hit/miss counters (public getters unchanged)
- New test simulating a DB failure during `add()`: `contains()` must return false and a redelivered `add()` must succeed once the DB is back (fails on current `main`, passes with the fix); existing cache hit/miss-count test passes unchanged, confirming counting semantics are preserved

### DIFF
--- a/components/camel-sql/src/main/java/org/apache/camel/processor/idempotent/jdbc/JdbcCachedMessageIdRepository.java
+++ b/components/camel-sql/src/main/java/org/apache/camel/processor/idempotent/jdbc/JdbcCachedMessageIdRepository.java
@@ -16,8 +16,9 @@
  */
 package org.apache.camel.processor.idempotent.jdbc;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
LGTM — clean backport of #24657 to 4.18.x.

Reviewed the fix logic:
- **Cache-after-insert**: `cache.merge(key, 1, Integer::sum)` is correctly placed *after*
  `super.add(key)`, so a failed DB insert (exception) skips the cache population. If
  `super.add()` returns `false` (key already in DB but not in cache), the key is still
  cached correctly — it's a valid known-duplicate.
- **Thread safety**: `ConcurrentHashMap` + `volatile` field for `reload()` visibility +
  `AtomicInteger` counters. The `merge()` call is atomic within ConcurrentHashMap, and the
  DB uniqueness constraint handles cross-thread insert races.
- **Test**: Drops the table to simulate outage, verifies the key isn't cached, recreates
  the table, and confirms redelivery works. Good coverage.

PR #24811, opened 2026-07-16
CAMEL-24162: Fix moduleClassNames typo in JacksonXMLDataFormatReifier

Croway:
## Summary

- Fix typo in `JacksonXMLDataFormatReifier.prepareDataFormatConfig()`: the property key `"modulesClassNames"` (extra `s`) is corrected to `"moduleClassNames"`, matching the actual field name in `JacksonXMLDataFormat` and its generated configurer.
- Add regression test (`JacksonXMLModuleClassNamesReifierTest`) that exercises the reifier path by configuring `moduleClassNames` through the model class, confirming the route starts and the custom module is applied.

The equivalent bug in camel-jackson (JSON) was fixed in 2016 (commit `eb2f171ff39f`), but the XML reifier was overlooked. The same reifier also serves camel-jackson3xml, which is equally fixed.

## Test plan

- [x] New test `JacksonXMLModuleClassNamesReifierTest` confirmed RED before fix, GREEN after
- [x] Full `mvn verify` on `components/camel-jacksonxml` passes (42 tests)

### DIFF
--- a/components/camel-jacksonxml/src/test/java/org/apache/camel/component/jacksonxml/JacksonXMLModuleClassNamesReifierTest.java
+++ b/components/camel-jacksonxml/src/test/java/org/apache/camel/component/jacksonxml/JacksonXMLModuleClassNamesReifierTest.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — clean and well-scoped fix.

**The bug**: `JacksonXMLDataFormatReifier.prepareDataFormatConfig()` stored the value under key `"modulesClassNames"` (extra `s`), while the runtime `JacksonXMLDataFormat` field and its generated configurer expect `"moduleClassNames"`. The typo caused `moduleClassNames` set through the DSL model to be silently ignored — the configurer never found a matching property key. The bug has been present since `e3460f86f43e` (Oct 2020) when the reifier was first extracted.

**The fix**: Single-character typo correction. Matches the JSON reifier's 2016 fix (`eb2f171ff39f`).

**The test**: Good regression test — `MyModule` applies a `"my-"` prefix naming strategy, so the expected XML output (`<my-name>`, `<my-country>`) directly proves the module was loaded through the reifier path. Without the fix, the output would have plain `<name>`/`<country>` elements.

Nice catch on the XML/jackson3xml side of this long-standing bug.

oscerd (approved):
LGTM. The reifier key `"modulesClassNames"` matched nothing in the generated configurer (which only knows `moduleClassNames`/`ModuleClassNames` → `setModuleClassNames`), so the option was silently dropped from XML/YAML DSL; correcting the key fixes it. The added `JacksonXMLModuleClassNamesReifierTest` is a genuine regression test — `MyModule.translate` prefixes property names with `my-`, so the asserted `<my-name>`/`<my-country>` output only appears when the module is actually applied — and it ran green in CI. The red `build (17)` is an unrelated flaky `camel-vertx-websocket` handshake test; camel-jacksonxml and the new test both passed.

PR #24812, opened 2026-07-16
CAMEL-24165: camel-soap - Fix ignoreUnmarshalledHeaders builder recursion and reifier omission

Croway:
## Summary

Fixes [CAMEL-24165](https://issues.apache.org/jira/browse/CAMEL-24165) — the `ignoreUnmarshalledHeaders` option on `SoapDataFormat` is completely unusable due to two bugs introduced with CAMEL-22354:

- **Problem A — StackOverflowError**: The builder's `boolean` overload at `SoapDataFormat.Builder.ignoreUnmarshalledHeaders(boolean)` calls `Boolean.valueOf(boolean)` which returns a `Boolean` that auto-unboxes back to `boolean`, hitting the same overload recursively. Fixed by using `Boolean.toString(boolean)` to route to the `String` overload, matching the pattern used by every other builder in the package.

- **Problem B — Reifier silently drops the option**: `SoapDataFormatReifier.prepareDataFormatConfig()` never maps `ignoreUnmarshalledHeaders` into the properties map, so the setting never reaches the runtime data format regardless of DSL. Fixed by adding the missing `properties.put(...)` call, following the same pattern as `JaxbDataFormatReifier` and `JsonDataFormatReifier`.

## Test plan

- [x] `mvn verify` on `components/camel-soap` passes — existing tests `SoapToSoapIgnoreTest` and `SoapToSoapDontIgnoreTest` exercise the runtime behavior of `ignoreUnmarshalledHeaders`
- [x] `mvn install` on `core/camel-core-model` and `core/camel-core-reifier` succeeds

### DIFF
--- a/core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/SoapDataFormat.java
+++ b/core/camel-core-model/src/main/java/org/apache/camel/model/dataformat/SoapDataFormat.java
@@ -295,7 +295,7 @@ public class SoapDataFormat extends DataFormatDefinition {
          * problematic header.
          */
         public Builder ignoreUnmarshalledHeaders(boolean ignoreUnmarshalledHeaders) {
-            return ignoreUnmarshalledHeaders(Boolean.valueOf(ignoreUnmarshalledHeaders));
+            return ignoreUnmarshalledHeaders(Boolean.toString(ignoreUnmarshalledHeaders));
         }
 
         /**
--- a/core/camel-core-reifier/src/main/java/org/apache/camel/reifier/dataformat/SoapDataFormatReifier.java
+++ b/core/camel-core-reifier/src/main/java/org/apache/camel/reifier/dataformat/SoapDataFormatReifier.java
@@ -37,6 +37,7 @@ public class SoapDataFormatReifier extends DataFormatReifier<SoapDataFormat> {
         properties.put("namespacePrefix", asRef(definition.getNamespacePrefix()));
         properties.put("schema", definition.getSchema());
         properties.put("contextPath", definition.getContextPath());
+        properties.put("ignoreUnmarshalledHeaders", definition.getIgnoreUnmarshalledHeaders());
     }
 
 }


gnodet (approved):

**LGTM** — two clear bugs, two minimal fixes.

**Problem A — StackOverflowError in builder (line 298):**
`Boolean.valueOf(boolean)` returns `Boolean`. Java overload resolution then unboxes it back to `boolean` (since `Boolean` ≠ `String`, the `(String)` overload doesn't match), calling the same `(boolean)` method recursively → StackOverflow. The fix `Boolean.toString(boolean)` produces a `String` that correctly routes to the `(String)` overload on line 306.

**Problem B — Reifier omission:**
`SoapDataFormatReifier.prepareDataFormatConfig()` maps 6 properties but omits `ignoreUnmarshalledHeaders`, so the setting never reaches the runtime `SoapDataFormat` via the configurer. Adding the `properties.put(...)` call matches the pattern used by every other property in the reifier and by analogous reifiers (JaxbDataFormatReifier, JsonDataFormatReifier).

Both bugs introduced in `5dfc5c2857e9` (CAMEL-22354, Aug 2025). The existing tests `SoapToSoapIgnoreTest` / `SoapToSoapDontIgnoreTest` exercise the runtime behavior and confirm correctness.

davsclaus (approved):

oscerd (commented):
The two fixes are correct: `ignoreUnmarshalledHeaders(boolean)` and `(String)` are the only overloads, so the old `ignoreUnmarshalledHeaders(Boolean.valueOf(b))` unboxed and re-invoked the boolean overload → StackOverflowError; routing through `Boolean.toString(...)` to the String overload fixes it. And the reifier was missing the `"ignoreUnmarshalledHeaders"` mapping (which matches the runtime `setIgnoreUnmarshalledHeaders` via the configurer), so the option never took effect from XML/YAML DSL — now wired. The red CI is unrelated flakiness (camel-smb `SmbStreamDownloadIT`, camel-undertow `BindException`, camel-sjms timeout); camel-soap itself built clean.

One gap worth closing before merge: neither fixed path actually has a regression test. `SoapToSoapIgnoreTest`/`SoapToSoapDontIgnoreTest` construct the runtime `SoapDataFormat` and call `setIgnoreUnmarshalledHeaders(...)` directly, then marshal the object — which bypasses **both** the model Builder (the recursion) **and** the reifier (the omission), so they pass with or without this change. A test that configures the option through the DSL/Builder (and one that exercises the builder setter) would guard against re-regression.

PR #24813, opened 2026-07-16
CAMEL-24164: camel-jaxb - Fix CamelJaxbPartClass header state pollution

Croway:
## Summary

The `JaxbDataFormat.unmarshal()` and `toElement()` methods write the header-derived `CamelJaxbPartClass` value directly into the shared **instance field** `this.partClass`. Since a `DataFormat` is shared across all exchanges in a route step, this causes:

- **Stickiness**: once any exchange carries the header, `partClass` remains non-null permanently — all subsequent exchanges without the header silently unmarshal/marshal using the leaked class.
- **Data race**: concurrent exchanges with different headers cross-contaminate.

The fix uses a **local variable** (`effectivePartClass`) instead of mutating the instance field, matching the pattern already used by Jackson, CBOR, Gson, and Jsonb data formats.

The bug was introduced in 2017 (CAMEL-5723, commit `03519184b36a`) and carried through the 2024 refactoring (commit `f0b4bb237331`).

## Changes

- **`JaxbDataFormat.java`**: Replace `this.partClass = ...` with local `effectivePartClass` in both `unmarshal()` and `toElement()`.
- **New test `JaxbPartClassHeaderStatePollutionTest.java`**: Two test cases (unmarshal + marshal) that send Exchange 1 with header, then Exchange 2 without, asserting no state leaks.

## Test plan

- [x] New test `JaxbPartClassHeaderStatePollutionTest` covers both unmarshal and marshal paths
- [x] All 113 existing camel-jaxb tests pass (`mvn verify` in `components/camel-jaxb`)
- [x] Verified that existing `JaxbDataFormatPartClassHeaderTest` and `JaxbDataFormatPartClassTest` still pass (header-driven and configured partClass paths)

### DIFF
--- a/components/camel-jaxb/src/main/java/org/apache/camel/converter/jaxb/JaxbDataFormat.java
+++ b/components/camel-jaxb/src/main/java/org/apache/camel/converter/jaxb/JaxbDataFormat.java
@@ -290,9 +290,11 @@ public class JaxbDataFormat extends ServiceSupport
     private Object toElement(
             Object graph, String partClassFromHeader, String partNamespaceFromHeader, QName partNamespaceOnDataFormat)
             throws JAXBException {
+        Class<?> effectivePartClass = partClass;
         if (partClassFromHeader != null) {
             try {
-                partClass = camelContext.getClassResolver().resolveMandatoryClass(partClassFromHeader, Object.class);
+                effectivePartClass
+                        = camelContext.getClassResolver().resolveMandatoryClass(partClassFromHeader, Object.class);
             } catch (ClassNotFoundException e) {
                 throw new JAXBException(e);
             }
@@ -300,7 +302,7 @@ public class JaxbDataFormat extends ServiceSupport
         if (partNamespaceFromHeader != null) {
             partNamespaceOnDataFormat = QName.valueOf(partNamespaceFromHeader);
         }
-        return new JAXBElement<>(partNamespaceOnDataFormat, (Class<Object>) partClass, graph);
+        return new JAXBElement<>(partNamespaceOnDataFormat, (Class<Object>) effectivePartClass, graph);
     }
 
     private boolean asXmlStreamWriter(Exchange exchange) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — important concurrency/state-pollution fix.

**The bug**: `JaxbDataFormat` is shared across all exchanges in a route step. Both `unmarshal()` and `toElement()` write the header-derived `CamelJaxbPartClass` value directly into the instance field `this.partClass`. This causes:
1. **Stickiness** — once any exchange carries the header, `partClass` stays non-null forever, silently using the leaked class for all subsequent exchanges without the header.
2. **Data race** — concurrent exchanges with different headers cross-contaminate through the shared field.

**The fix**: Introduces a local `effectivePartClass` variable initialized from `this.partClass`, overridden from the header when present. The instance field is never mutated. This matches the established pattern used by Jackson, CBOR, Gson, and Jsonb data formats.

**Note**: `partNamespace` is already safe — it's read into a local via `getPartNamespace()` and the header override mutates the method parameter `partNamespaceOnDataFormat`, not the instance field.

**The test**: Excellent coverage — sends Exchange 1 with `CamelJaxbPartClass` header, then Exchange 2 without, asserting no state leaks in both unmarshal and marshal paths. The assertions are specific (`assertInstanceOf(PurchaseOrder.class, ...)` / `assertFalse(payload2.contains("<address:address"))`) — directly proving the absence of pollution.

Bug has been present since 2017 (CAMEL-5723). Nice catch.

davsclaus (approved):

oscerd (approved):
LGTM — real cross-exchange contamination fix. Mutating the shared `this.partClass` field per request meant a `CamelJaxbPartClass` header set for one exchange stuck around for the next (and raced under concurrency). Replacing it with a local `effectivePartClass` threaded through both `toElement`/marshal and the unmarshal path is logically equivalent across all configured×header combinations (header still takes precedence, full-vs-partial selection preserved) and eliminates the stickiness. The two-exchange isolation test (header on exchange 1, none on exchange 2, asserting exchange 2's payload isn't partial) targets exactly the deterministic manifestation. Clean.

PR #24814, opened 2026-07-16
CAMEL-24152: Fix concurrent marshal/unmarshal data corruption in camel-snakeyaml

Croway:
## Summary

Fixes [CAMEL-24152](https://issues.apache.org/jira/browse/CAMEL-24152): concurrent `marshal`/`unmarshal` operations corrupt data because all per-thread `Yaml` instances share single `BaseConstructor` and `Representer` singletons created once in `doStart()`.

This was a regression introduced by CAMEL-22354 (commit `5dfc5c2857e9`, released in 4.15.0), which replaced `Function<CamelContext, BaseConstructor>` factories with shared singletons while retaining the `ThreadLocal<Yaml>` cache. Since SnakeYAML's `BaseConstructor` and `Representer` are heavily stateful, sharing them across threads causes:

1. Internal SnakeYAML crashes (`IndexOutOfBoundsException`, `ParserException`, `NPE`)
2. Marshal crashes (`NullPointerException` on nodes)
3. Silent cross-thread data bleed (one thread's unmarshal returns another thread's payload)

### Changes

- **`SnakeYAMLDataFormat.doStart()`** — Removed default singleton creation for `constructor`, `representer`, `dumperOptions`, `resolver`.
- **`SnakeYAMLDataFormat.getYaml()`** — When creating a new `Yaml`, calls `default*()` factory methods to produce fresh per-thread instances instead of reusing shared singletons. User-supplied custom instances (via `setConstructor()` etc.) are still used as-is.
- **`SnakeYAMLConcurrentTest`** — New test: 8 threads × 500 marshal/unmarshal round-trips with distinct payloads, asserting no exceptions and no cross-thread data corruption.

## Test plan

- [x] `mvn verify` in `components/camel-snakeyaml` — all 19 tests pass (including new concurrency test)
- [ ] Verify existing SnakeYAML tests pass in CI (SnakeYAMLTest, SnakeYAMLDoSTest, SnakeYAMLTypeFilterTest, etc.)
- [ ] Verify the new concurrent test passes in CI

### DIFF
--- a/components/camel-snakeyaml/src/main/java/org/apache/camel/component/snakeyaml/SnakeYAMLDataFormat.java
+++ b/components/camel-snakeyaml/src/main/java/org/apache/camel/component/snakeyaml/SnakeYAMLDataFormat.java
@@ -134,18 +134,6 @@ public final class SnakeYAMLDataFormat extends ServiceSupport implements DataFor
         if (allowAnyType) {
             typeFilters = "*";
         }
-        if (this.constructor == null) {
-            this.constructor = defaultConstructor(camelContext);
-        }
-        if (this.representer == null) {
-            this.representer = defaultRepresenter();
-        }
-        if (this.dumperOptions == null) {
-            this.dumperOptions = defaultDumperOptions();
-        }
-        if (this.resolver == null) {
-            this.resolver = defaultResolver();
-        }
     }
 
     @Override
@@ -167,7 +155,11 @@ public final class SnakeYAMLDataFormat extends ServiceSupport implements DataFor
             options.setTagInspector(new TrustedTagInspector());
             options.setAllowRecursiveKeys(allowRecursiveKeys);
             options.setMaxAliasesForCollections(maxAliasesForCollections);
-            yaml = new Yaml(constructor, representer, dumperOptions, options, resolver);
+            BaseConstructor c = constructor != null ? constructor : defaultConstructor(camelContext);
+            Representer r = representer != null ? representer : defaultRepresenter();
+            DumperOptions d = dumperOptions != null ? dumperOptions : defaultDumperOptions();
+            Resolver res = resolver != null ? resolver : defaultResolver();
+            yaml = new Yaml(c, r, d, options, res);
             yamlCache.set(new WeakReference<>(yaml));
         }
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — important concurrency fix for a real data-corruption bug.

**Root cause**: CAMEL-22354 (`5dfc5c2857e9`, Aug 2025) replaced per-thread `Function<CamelContext, BaseConstructor>` factories with shared singletons. The `ThreadLocal<Yaml>` cache remained, but all per-thread `Yaml` instances now shared the same `BaseConstructor`, `Representer`, `DumperOptions`, and `Resolver`. Since SnakeYAML's `BaseConstructor` and `Representer` mutate internal state on every `load()`/`dump()`, concurrent exchanges produce crashes (`IndexOutOfBoundsException`, NPEs) and silent cross-thread data corruption.

**The fix**: Correct and minimal — moves construction of default `BaseConstructor`, `Representer`, `DumperOptions`, and `Resolver` into `getYaml()` so each thread-cached `Yaml` gets fresh instances via the `default*()` factory methods. User-supplied custom instances (via setters) are still used as-is, which is a reasonable trade-off — restoring the factory pattern for custom instances would require an API change.

**The test**: Good stress test — 8 threads × 500 marshal/unmarshal round-trips with distinct per-thread payloads, `CountDownLatch` for maximum contention, and explicit per-iteration data integrity assertion.

**Minor note**: The dual `LoaderOptions` pattern (one in `defaultConstructor()`, another in `getYaml()`) is pre-existing — the PR preserves the original behavior.

This should be considered for backport to 4.18.x if 4.15.0+ releases are affected.

davsclaus (approved):

oscerd (approved):
LGTM — correct thread-safety fix for a real data-corruption bug. The shared `SnakeYAMLDataFormat` reused stateful `BaseConstructor`/`Representer`/`DumperOptions`/`Resolver` singletons across threads (snakeyaml's `Yaml` isn't thread-safe), so concurrent marshal/unmarshal corrupted output. Building fresh `default*()` instances per thread-cached `Yaml` isolates them; since they're cached in the existing per-thread `ThreadLocal<WeakReference<Yaml>>` (created once on cache-miss, not per call) there's no perf cliff and no new ThreadLocal leak, and `doStop()` still clears the cache. `SnakeYAMLConcurrentTest` (8 threads × 500 round-trips) is a good regression guard.

Two non-blocking notes: a user-supplied custom `constructor`/`representer` (via the setters) is still shared across threads — opt-in and out of scope here, but worth a follow-up if you want full isolation. And `getConstructor()`/`getRepresenter()`/etc. now return null after start when not user-set (was internal defaults) — no signature change and nothing internal depends on it, just flagging the observable getter change.

PR #24815, opened 2026-07-16
[backport camel-4.18.x] CAMEL-24044: camel-sql - Fix Postgres aggregation repositories binding version without a placeholder

Croway:
## Backport of #24655

Cherry-pick of #24655 onto `camel-4.18.x`.

**Original PR:** #24655 - CAMEL-24044: camel-sql - Fix Postgres aggregation repositories binding version without a placeholder
**Original author:** @Croway

### Conflict resolution

The upgrade guide entry (`camel-4x-upgrade-guide-4_22.adoc`) was removed from the backport since it does not exist on the 4.18.x branch — per project guidelines, upgrade guide entries for backported changes belong on `main`.

The `camel-test-infra-postgres` dependency was updated to use `<type>test-jar</type>` because on this branch the service classes (`PostgresService`, `PostgresServiceFactory`) are still in `src/test/java` (they were moved to `src/main/java` on `main`).

### Original description

Since the CAMEL-7810 optimistic-locking refactor (Camel 3.4), `PostgresAggregationRepository.insert()` built `INSERT INTO t(exchange, id, ...) ON CONFLICT DO NOTHING` without the `version` column while the base `insertHelper` binds blob, key **and version** — one more bound parameter than placeholders, so every `add()` failed with `The column index is out of range` (verified against real PostgreSQL). With `storeBodyAsText=true` the version was silently bound into the `body` placeholder before failing. `ClusteredPostgresAggregationRepository` had the same defect in its completed-table insert (every `remove()` failed at completion) and never wrote the `instance_id` column, so instance-scoped recovery could never match any row.

Changes:
- Include the `version` column in both Postgres `INSERT` statements and pass the actual version instead of the hardcoded `1L`
- Clustered variant additionally writes `instance_id` when `completed && recoveryByInstance`
- New integration tests against real PostgreSQL (`camel-test-infra-postgres`) covering the full add/get/update/remove/recover/confirm lifecycle, `storeBodyAsText`, and instance-scoped recovery with two nodes — these fail on current `main` and pass with the fix
- Fixed the stale `INSERT` example in the component docs and added an upgrade-guide entry (the aggregation tables need the `version BIGINT` column, as `JdbcAggregationRepository` already required for reads/updates)

### DIFF
--- a/components/camel-sql/pom.xml
+++ b/components/camel-sql/pom.xml
@@ -95,6 +95,20 @@
             <version>2.4.240</version>
             <scope>test</scope>
         </dependency>
+        <!-- PostgreSQL only used for the aggregation repository integration tests -->
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>${pgjdbc-driver-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-infra-postgres</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/components/camel-sql/src/main/docs/sql-component.adoc
+++ b/components/camel-sql/src/main/docs/sql-component.adoc
@@ -757,7 +757,7 @@ uses special `INSERT .. ON CONFLICT ..` statement to provide optimistic locking 
 This statement is (with default aggregation table definition):
 [source,sql]
 ----
-INSERT INTO aggregation (id, exchange) values (?, ?) ON CONFLICT DO NOTHING
+INSERT INTO aggregation (exchange, id, version) VALUES (?, ?, ?) ON CONFLICT DO NOTHING
 ----
 
 Details can be found https://www.postgresql.org/docs/9.5/sql-insert.html[in PostgreSQL documentation].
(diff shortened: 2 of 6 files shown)


gnodet (approved):

**LGTM** — clean backport of #24655 (merged to main July 13).

The fix is well-understood: both `PostgresAggregationRepository` and `ClusteredPostgresAggregationRepository` were building `INSERT` statements without the `version` column while `insertHelper()` binds blob, key, **and** version — the parameter count mismatch caused every `add()` to fail with `The column index is out of range` against real PostgreSQL. The clustered variant additionally now writes `instance_id` for instance-scoped recovery.

**Backport adaptations are correct:**
- Upgrade guide entry removed (doesn't exist on 4.18.x — per project guidelines, it belongs on `main`)
- `camel-test-infra-postgres` uses `<type>test-jar</type>` since service classes are still in `src/test/java` on this branch

**Integration tests** are thorough — full add/get/update/remove/recover/confirm lifecycle against real PostgreSQL, plus `storeBodyAsText` and instance-scoped recovery with two nodes. Good coverage for a critical data-path fix.

davsclaus (approved):

oscerd (approved):
LGTM — faithful backport and the SQL is correct against the 4.18.x baseline. Both changed source files are byte-identical to the main fix and both ITs match main. The key correctness point checks out: `insertHelper` always binds blob(EXCHANGE)→key(ID)→version(VERSION)[→body][→headers], but the Postgres override's column list stopped at ID (2 placeholders) while 3 values were bound → "column index out of range"; the fix adds the `VERSION` column + placeholder so count and order match, and switches `1L`→`version` to align with the base class. The clustered variant appends `instance_id` last with a matching placeholder (previously never written, so instance-scoped recovery could never match). Backport adaptations are right too — the `<type>test-jar</type>` on the postgres test-infra dep matches the established 4.18.x pattern, and the upgrade-guide entry is correctly left on main. Two ITs added.

PR #24816, opened 2026-07-16
CAMEL-24167: camel-csv - Fix format option silently ignored

Croway:
## Summary

Fixes two stacked bugs in `CsvDataFormat` that cause the `format` option (e.g., `<csv format="MYSQL"/>`) to be silently ignored:

- **Dead guard:** The `csvFormat` field is eagerly initialized to `CSVFormat.DEFAULT`, making the `doInit()` condition `if (csvFormat == null && format != null)` unreachable. Removed the `csvFormat == null` guard so the `format` string is resolved when set via DSL/XML/YAML.
- **Name mismatch:** `CSVFormat.valueOf(format)` delegates to `CSVFormat.Predefined.valueOf()` which uses CamelCase names (`Excel`, `InformixUnload`) while Camel's model advertises uppercase names (`EXCEL`, `INFORMIX_UNLOAD`). Replaced with a case-insensitive, underscore-stripped lookup.

**Regression history:** Introduced in CAMEL-22354 (getter/setter alignment), partially addressed in CAMEL-23031 (init ordering) but the dead guard and name mismatch remained.

**Failure scenario:** `setFormat("MYSQL")` (which should produce TAB-delimited output) silently produces comma-delimited DEFAULT output instead.

## Test plan

- [x] Added `shouldResolveFormatByName` — verifies `setFormat("MYSQL")` produces TAB delimiter
- [x] Added `shouldResolveFormatByNameCaseInsensitive` — verifies `setFormat("EXCEL")` resolves despite CamelCase enum mismatch
- [x] Existing Spring XML tests (`CsvMarshalAutogenColumnsSpringTest`, `CsvMarshalAutogenColumnsSpringQuoteModeTest`) that use `format="EXCEL"` now exercise the fix path and pass
- [x] Full `mvn verify` on `camel-csv` module passes (92 tests)

### DIFF
--- a/components/camel-csv/src/main/java/org/apache/camel/dataformat/csv/CsvDataFormat.java
+++ b/components/camel-csv/src/main/java/org/apache/camel/dataformat/csv/CsvDataFormat.java
@@ -105,8 +105,8 @@ public class CsvDataFormat extends ServiceSupport implements DataFormat, DataFor
     protected void doInit() throws Exception {
         super.doInit();
 
-        if (csvFormat == null && format != null) {
-            csvFormat = CSVFormat.valueOf(format);
+        if (format != null) {
+            csvFormat = resolveFormat(format);
         }
         if (csvFormat == null) {
             csvFormat = CSVFormat.DEFAULT;
@@ -119,6 +119,19 @@ public class CsvDataFormat extends ServiceSupport implements DataFormat, DataFor
     @Override
     protected void doStop() throws Exception {
         // noop
+    }
+
+    private static CSVFormat resolveFormat(String name) {
+        // The model advertises uppercase names (EXCEL, INFORMIX_UNLOAD) but
+        // commons-csv Predefined uses CamelCase (Excel, InformixUnload).
+        // Match by stripping underscores and comparing case-insensitively.
+        String normalized = name.replace("_", "");
+        for (CSVFormat.Predefined p : CSVFormat.Predefined.values()) {
+            if (p.name().equalsIgnoreCase(normalized)) {
+                return p.getFormat();
+            }
+        }
+        throw new IllegalArgumentException("Unknown CSV format: " + name);
     }
 
     CSVFormat getActiveFormat() {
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — two stacked bugs, clean minimal fix.

**Bug 1 — Dead guard (line 108):** `csvFormat` is eagerly initialized to `CSVFormat.DEFAULT` on line 41, so `if (csvFormat == null && format != null)` is always false — the `format` string is never resolved. Fix correctly removes the `csvFormat == null` guard so `format` overrides the default when set via DSL/XML/YAML.

**Bug 2 — Name mismatch:** `CSVFormat.valueOf(format)` delegates to `CSVFormat.Predefined.valueOf()` which uses CamelCase names (`Excel`, `MySQL`, `InformixUnload`), while Camel's model advertises uppercase names (`EXCEL`, `MYSQL`, `INFORMIX_UNLOAD`). The `resolveFormat()` method correctly handles this with case-insensitive + underscore-stripped lookup:
- `MYSQL` → normalized `MYSQL` matches `MySQL` ✓
- `EXCEL` → normalized `EXCEL` matches `Excel` ✓
- `INFORMIX_UNLOAD` → normalized `INFORMIXUNLOAD` matches `InformixUnload` ✓

Throws `IllegalArgumentException` on unknown names (fail-fast, good).

The tests verify both the TAB-delimiter behavior (`MYSQL`) and the case-insensitive resolution (`EXCEL`). The existing Spring XML tests (`CsvMarshalAutogenColumnsSpringTest`) that use `format="EXCEL"` now also exercise the fixed path.

davsclaus (approved):

oscerd (approved):
LGTM — two real bugs fixed. The `if (csvFormat == null && format != null)` guard was dead because `csvFormat` is eagerly initialised to `CSVFormat.DEFAULT`, so a configured `format` preset was silently ignored; changing it to `if (format != null)` fixes that, and `resolveFormat()` maps names case-insensitively (`MYSQL`→`MySQL`, `EXCEL`→`Excel`, fail-fast on unknown). Composition is correct too — `getActiveFormat()` uses the resolved preset as the base and layers delimiter/header/quote overrides on top. Two tests added.

Minor: since output now changes for anyone who previously set `format=` and silently got DEFAULT, a short upgrade-guide note would be worth adding.

PR #24817, opened 2026-07-16
CAMEL-24166: Enforce maxDecompressedSize in iterator/splitter mode for zip and tar

Croway:
## Summary

The `maxDecompressedSize` setting (default 1 GB) protects against decompression bombs, but was only enforced when `usingIterator=false`. When `usingIterator=true` — the common mode for multi-entry archives and streaming splits — neither `ZipIterator` nor `TarIterator` received or enforced the limit, completely defeating decompression-bomb protection in iterator/splitter mode.

**Changes:**
- **`ZipIterator`**: added `maxDecompressedSize` field and changed `IOHelper.copy` call in `getNextElement()` to use the 5-argument overload that enforces the limit
- **`ZipFileDataFormat`**: propagates `maxDecompressedSize` to `ZipIterator` in the `usingIterator` unmarshal branch
- **`TarIterator`**: added `maxDecompressedSize` field and wraps entry streams in `MaxDecompressedSizeInputStream` when the limit is active
- **`TarFileDataFormat`**: propagates `maxDecompressedSize` to `TarIterator` in the `usingIterator` unmarshal branch
- **`MaxDecompressedSizeInputStream`** (new): a `FilterInputStream` that counts bytes read and throws `IOException` when the decompressed size limit is exceeded — used by `TarIterator` to preserve lazy/streaming reads required by `TarArchiveInputStream`
- **Tests**: added `testUnzipMaxDecompressedSizeWithIterator` and `testUntarMaxDecompressedSizeWithIterator` to confirm the limit is enforced in iterator mode

## Test plan

- [x] `cd components/camel-zipfile && mvn verify` — 44 tests pass (0 failures)
- [x] `cd components/camel-tarfile && mvn verify` — 43 tests pass (0 failures)
- [x] New tests confirm `CamelExecutionException` is thrown when `maxDecompressedSize=10` with `usingIterator=true`
- [x] Existing tests confirm no regression in non-iterator mode or default behavior


gnodet (approved):

**LGTM** — important security fix closing a gap in decompression-bomb protection.

**The gap**: `maxDecompressedSize` (default 1 GB) was only enforced when `usingIterator=false`. In iterator/splitter mode — the common mode for multi-entry archives and streaming splits — neither `ZipIterator` nor `TarIterator` received or enforced the limit, completely defeating decompression-bomb protection.

**Zip fix**: Propagates `maxDecompressedSize` to `ZipIterator` and uses the existing `IOHelper.copy(input, output, bufferSize, flush, maxSize)` 5-argument overload (line 259 checks `maxSize > 0 && total > maxSize`). Clean — reuses existing infrastructure.

**Tar fix**: Propagates `maxDecompressedSize` to `TarIterator` and wraps entry streams in `MaxDecompressedSizeInputStream`. Different approach because tar iterator sets the raw `TarElementInputStreamWrapper` as the exchange body (lazy streaming) — there's no single `IOHelper.copy()` call to intercept. The `FilterInputStream` wrapper correctly enforces the limit lazily during consumer reads by counting bytes in both `read()` and `read(byte[], int, int)`.

**`MaxDecompressedSizeInputStream`**: Clean implementation. Overrides both fundamental read methods. Minor note: `skip()` is inherited from `FilterInputStream` (delegates to `in.skip()`, bypassing the counter), but this is acceptable — the threat model is decompression bombs where consumers read full content, not adversarial code calling `skip()`.

Tests correctly verify the limit is enforced in iterator mode with `maxDecompressedSize=10` and streaming split.

davsclaus (changes requested):
the new option should be added in the model for the data formats

oscerd (commented):
Reviewed this from the decompression-bomb angle — the security fix itself is **correct**. The limit is now enforced in iterator/splitter mode for both formats, and crucially *before* over-materialization: zip goes through `IOHelper.copy(zipInputStream, cos, DEFAULT_BUFFER_SIZE, false, maxDecompressedSize)` (aborts within one buffer of the limit into a disk-spilling `CachedOutputStream`), and tar wraps each entry in commons-io `BoundedInputStream` with `setMaxCount`/throw-on-max. Default is 1 GB, on by default, `-1` to disable. Good tests (zip + tar, iterator and splitter, `maxDecompressedSize=10` → `CamelExecutionException`). The per-entry (rather than cumulative) bound is fine under the streaming threat model.

A couple of process notes so this can get unstuck rather than code issues:

- On @davsclaus's "add the option to the model" request: `maxDecompressedSize` is **already** modeled on the data formats — `ZipFileDataFormat.java:55` and `TarFileDataFormat.java:53` (both `@Metadata(label="advanced", defaultValue="1073741824")`), and it's in the generated catalog. The only genuinely *new* option here is on `ZipSplitter`/`TarSplitter`, which are `Expression` classes with no DSL model. Might be worth confirming with him whether that split-level option is what he wants surfaced — the data-format option itself is already there.
- Heads-up that @gnodet's approval predates the final commit: it reviewed a custom `MaxDecompressedSizeInputStream` that `3fe487c7` replaced with commons-io `BoundedInputStream`, so the current tar approach hasn't actually been signed off — a fresh look at the final version would be good.
- Minor: since iterator/splitter mode previously did **not** enforce `maxDecompressedSize`, a multi-entry archive that streamed fine before can now throw at >1 GB — that broadened enforcement is worth a short `camel-4x-upgrade-guide-4_22.adoc` note even though the default value is unchanged.

Nice hardening overall.

davsclaus (approved):
okay a nit pick would be to add javadoc to the iterator classes for these options and mention the default value

PR #24820, opened 2026-07-17
CAMEL-24140: Fix UseOriginalAggregationStrategy not honored when shareUnitOfWork enabled

davsclaus:
## Summary

- Fix `UseOriginalAggregationStrategy` being silently ignored when `shareUnitOfWork=true` on Multicast, Split, and RecipientList EIPs
- Remove `ShareUnitOfWorkAggregationStrategy` wrapping from the three reifiers (MulticastReifier, SplitReifier, RecipientListReifier), centralizing all wrapping logic in `MulticastProcessor.process()`
- Add tests covering `UseOriginalAggregationStrategy` with and without `shareUnitOfWork` for all three EIPs

The root cause: the reifiers wrapped the strategy in `ShareUnitOfWorkAggregationStrategy` at route-build time, so the `instanceof UseOriginalAggregationStrategy` check in `MulticastProcessor.process()` (added by CAMEL-24062) never matched, and `newInstance(exchange)` was never called — the original body was lost.

## Test plan

- [x] `MulticastUseOriginalShareUnitOfWorkTest` — multicast with and without `shareUnitOfWork`
- [x] `SplitUseOriginalShareUnitOfWorkTest` — split with and without `shareUnitOfWork`
- [x] `RecipientListUseOriginalShareUnitOfWorkTest` — recipient list with and without `shareUnitOfWork`
- [x] `MulticastUseOriginalPropagateExceptionCaughtTest` — existing regression test
- [x] `MulticastShareUnitOfWorkOnExceptionHandledFalseIssueTest` — existing regression test

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -332,6 +332,9 @@ public class MulticastProcessor extends BaseProcessorSupport
                 clone = new ShareUnitOfWorkAggregationStrategy(clone);
             }
             setAggregationStrategyOnExchange(exchange, clone);
+        } else if (isShareUnitOfWork() && strategy != null
+                && !(strategy instanceof ShareUnitOfWorkAggregationStrategy)) {
+            setAggregationStrategyOnExchange(exchange, new ShareUnitOfWorkAggregationStrategy(strategy));
         }
 
         if (synchronous) {
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -333,6 +333,7 @@ public class MulticastProcessor extends BaseProcessorSupport
             }
             setAggregationStrategyOnExchange(exchange, clone);
         } else if (isShareUnitOfWork() && strategy != null
+                // wrap here (not in the reifier) so the UseOriginal instanceof check above sees the unwrapped strategy
                 && !(strategy instanceof ShareUnitOfWorkAggregationStrategy)) {
             setAggregationStrategyOnExchange(exchange, new ShareUnitOfWorkAggregationStrategy(strategy));
         }
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -333,7 +333,7 @@ public class MulticastProcessor extends BaseProcessorSupport
             }
             setAggregationStrategyOnExchange(exchange, clone);
         } else if (isShareUnitOfWork() && strategy != null
-                // wrap here (not in the reifier) so the UseOriginal instanceof check above sees the unwrapped strategy
+        // wrap here (not in the reifier) so the UseOriginal instanceof check above sees the unwrapped strategy
                 && !(strategy instanceof ShareUnitOfWorkAggregationStrategy)) {
             setAggregationStrategyOnExchange(exchange, new ShareUnitOfWorkAggregationStrategy(strategy));
         }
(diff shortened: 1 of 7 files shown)


gnodet (approved):
## Review: CAMEL-24140 — Fix UseOriginalAggregationStrategy not honored when shareUnitOfWork enabled

### What works well

**Precise root cause identification.** The reifiers wrapped the strategy in `ShareUnitOfWorkAggregationStrategy` at route-build time, hiding the inner `UseOriginalAggregationStrategy` from the `instanceof` check in `MulticastProcessor.process()` (added by CAMEL-24062). Centralizing the wrapping into `process()` — after the `UseOriginalAggregationStrategy` check — is the clean fix.

**Good defensive coding.** The `!(strategy instanceof ShareUnitOfWorkAggregationStrategy)` guard prevents double-wrapping, and the `strategy != null` check is necessary since `Splitter.process()` handles the null-strategy case independently (lines 192–197 of `Splitter.java`).

**Consistent with existing pattern.** The `if` branch (CAMEL-24062) already creates per-exchange `ShareUnitOfWorkAggregationStrategy` instances; the new `else if` follows the same approach.

### Suggestions

**[Suggestion] Consider extending test coverage to Split and RecipientList.**
The PR removes `ShareUnitOfWorkAggregationStrategy` wrapping from all three reifiers but only tests multicast. While `Splitter` and `RecipientListProcessor` delegate to `MulticastProcessor.process()` via `super.process()`, they have their own `process()` overrides with distinct logic (e.g. `Splitter` has a separate null-strategy fallback path at line 192). Adding equivalent tests for `split()` and `recipientList()` with a custom `AggregationStrategy` + `.shareUnitOfWork()` would confirm the wrapping works through all entry paths.

**[Nit] A brief comment on the `else if` branch would help future readers understand why wrapping was moved from the reifiers to `process()`.**
Something like:
```java
// Wrap non-UseOriginal strategies here (not in the reifier) so the
// UseOriginal instanceof check above runs against the unwrapped strategy first
```

### Checklist

| Item | Status |
|------|--------|
| Tests | ✅ New test covers multicast with/without shareUnitOfWork |
| Docs / Upgrade guide | N/A (bugfix for behavior broken by recent unreleased CAMEL-24062) |
| Commit convention | ✅ `CAMEL-24140: <description>` |
| Public API / backward compat | ✅ Internal change only |
| Security | N/A |
| CI | ⏳ Pending |

**Static analysis:** semgrep (0 findings), gitleaks (false positives only), ast-grep (0 findings).

Overall the fix is correct and well-structured. The suggestions above are non-blocking.

---
_Reviewed with Claude Code (oss-code-reviewer) on behalf of gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

davsclaus:
Thanks for the review @gnodet!

Addressed both suggestions:

1. **Added test coverage for Split and RecipientList** — `SplitUseOriginalShareUnitOfWorkTest` and `RecipientListUseOriginalShareUnitOfWorkTest` each test with and without `shareUnitOfWork`, confirming the wrapping works through all entry paths.

2. **Added comment on the `else if` branch** — explains why wrapping was moved from the reifiers to `process()`.

oscerd (commented):
The fix looks correct and regression-free: centralizing the `UseOriginalAggregationStrategy` wrapping in `process()` (after the `instanceof` check) instead of at build time means the check is no longer masked when `shareUnitOfWork=true`. I verified the per-exchange strategy is consumed via `getAggregationStrategy(exchange)`, the no-arg `timeout()` path delegates identically through `ShareUnitOfWorkAggregationStrategy.timeout()`, `parseBoolean` is still used in all three reifiers, and `isShareUnitOfWork` is still passed to the constructor. Good coverage (multicast + split + recipientList, with and without share).

The red `build (17)`/`build (25)` is just a formatting violation, not a test failure: the "Fail if there are uncommitted changes" gate trips because the review-feedback comment in the new `else if` is indented as an expression continuation (16 spaces) where the Eclipse formatter wants statement level (8), so `formatter:format` reformats it. A `mvn formatter:format` + recommit should turn it green.

PR #24821, opened 2026-07-17
CAMEL-24141: Aggregate EIP - guard background tasks against exceptions

davsclaus:
## Summary

- Add `try/catch(Exception)` guard to `AggregationIntervalTask.run()` and `RecoverTask.run()` so that a transient exception (e.g. repository hiccup) does not permanently cancel the scheduled background task
- Make `aggregateCompletionCounter()` null-safe to avoid NPE when `CamelAggregatedCompletedBy` property is absent on recovered exchanges
- Follows the same pattern already used by `DefaultTimeoutMap.purgeTask()` which catches and logs exceptions to keep the scheduled executor alive

## Test plan

- [x] All 177 existing aggregator tests pass (`org.apache.camel.processor.aggregator.*Test`)
- [ ] Verify with the reproducer tests attached to CAMEL-24141

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/aggregate/AggregateProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/aggregate/AggregateProcessor.java
@@ -923,6 +923,9 @@ public class AggregateProcessor extends BaseProcessorSupport
 
     private void aggregateCompletionCounter(Exchange exchange) {
         String completedBy = exchange.getProperty(ExchangePropertyKey.AGGREGATED_COMPLETED_BY, String.class);
+        if (completedBy == null) {
+            return;
+        }
         switch (completedBy) {
             case COMPLETED_BY_INTERVAL:
                 completedByInterval.incrementAndGet();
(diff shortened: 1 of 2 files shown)


gnodet (approved):
## Review: CAMEL-24141 — Aggregate EIP: guard background tasks against exceptions

### What works well

**Correct fix for a well-known Java scheduling pitfall.** When a `Runnable` submitted to `ScheduledExecutorService` throws an unchecked exception, the executor silently stops scheduling future runs. Wrapping both `AggregationIntervalTask.run()` and `RecoverTask.run()` in `try/catch(Exception)` with a `LOG.warn` is the standard defensive pattern — and is already used by `DefaultTimeoutMap.purgeTask()` in this codebase.

**Null-safe `aggregateCompletionCounter()`.** Adding the early return when `completedBy` is null prevents an NPE in the `switch` statement for recovered exchanges that may not have the `CamelAggregatedCompletedBy` property set. Clean and minimal.

**Correct `catch` placement in `RecoverTask`.** The `catch(Exception)` sits between the recovery logic and the `finally` block that clears `recoveryInProgress` and `inProgressCompleteExchangesForRecoveryTask`, so cleanup still runs on exception.

### Checklist

| Item | Status |
|------|--------|
| Tests | ✅ 177 existing aggregator tests cover normal paths |
| Docs / Upgrade guide | N/A (internal defensive hardening) |
| Commit convention | ✅ `CAMEL-24141: <description>` |
| Security | N/A |
| CI | ⏳ Pending |

**Note:** This PR includes the CAMEL-24140 commit on the same branch — the AggregateProcessor changes are the only new code for this PR.

**Static analysis:** semgrep (0 findings).

Clean, minimal, and correct. No concerns.

---

oscerd (commented):
The guards themselves look correct — wrapping `AggregationIntervalTask.run()` and `RecoverTask.run()` in `try/catch(Exception) + LOG.warn` keeps a single background-task exception from killing the scheduler, the inner `lock/finally unlock` is preserved (no lock leak), and `RecoverTask`'s catch sits before the `finally` that clears `recoveryInProgress`, matching the existing `DefaultTimeoutMap.purgeTask()` convention. The null-safe `aggregateCompletionCounter` early-return is a reasonable NPE guard too.

One thing before merge: the CAMEL-24141 change (scheduler survives after a background task throws) has no dedicated test — the only test added here (`MulticastUseOriginalShareUnitOfWorkTest`) belongs to the stacked CAMEL-24140. A small test that makes an interval/recover task throw and asserts the scheduler keeps running would lock in the actual fix. (Also note this PR is stacked with #24820's reifier change — worth confirming the merge order.)

PR #24822, opened 2026-07-17
CAMEL-24142: Fix Multicast/Split timeout silently lost when fired during aggregation

davsclaus:
## Summary

- Fix race condition in `MulticastTask.timeout()` where `tryLock()` silently drops the timeout when the aggregation lock is held, causing the exchange to hang forever
- Change `timeout()` to use `lock.lock()` instead of `tryLock()` — the timeout is a one-shot scheduled task that runs on a dedicated executor, so blocking until the lock is available is safe and correct
- Add `done.get()` guard after acquiring the lock to skip timeout handling if the multicast already completed normally
- Add reproducer test that makes the race deterministic by using a slow aggregation strategy that holds the lock while the timeout fires

The bug was introduced in CAMEL-16103 (2021) when `tryLock()` was correctly applied to `aggregate()` (called repeatedly, safe to skip) but also incorrectly applied to `timeout()` (called once, never retried). The fix affects Multicast, Split, and RecipientList EIPs since they all share `MulticastProcessor`.

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/MulticastProcessor.java
@@ -505,33 +505,38 @@ public class MulticastProcessor extends BaseProcessorSupport
         }
 
         protected void timeout() {
+            // use lock() instead of tryLock() because timeout is a one-shot scheduled task
+            // if tryLock fails (lock held by aggregate), the timeout would be silently lost
+            // and the exchange would hang forever (CAMEL-24142)
             Lock lock = this.lock;
-            if (lock.tryLock()) {
-                try {
-                    while (nbAggregated.get() < nbExchangeSent.get()) {
-                        Exchange exchange = completion.pollUnordered();
-                        int index = exchange != null ? getExchangeIndex(exchange) : nbExchangeSent.get();
-                        while (nbAggregated.get() < index) {
-                            int idx = nbAggregated.getAndIncrement();
-                            AggregationStrategy strategy = getAggregationStrategy(null);
-                            if (strategy != null) {
-                                strategy.timeout(result.get() != null ? result.get() : original,
-                                        idx, nbExchangeSent.get(), timeout);
-                            }
-                        }
-                        if (exchange != null) {
-                            doAggregate(result, exchange, original);
-                            nbAggregated.incrementAndGet();
+            lock.lock();
+            try {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
## Review: CAMEL-24142 — Fix Multicast/Split timeout silently lost when fired during aggregation

### What works well

**Precise diagnosis of the race condition.** The `timeout()` method used `tryLock()`, which was correct for `aggregate()` (called repeatedly from the completion queue — safe to skip since the next call retries) but incorrect for `timeout()` (a one-shot scheduled task — if it fails to acquire the lock, it never retries, and the exchange hangs forever).

**Correct fix.** Changing to `lock.lock()` is safe because the timeout task runs on a dedicated scheduled executor thread, not a routing thread. Blocking until the aggregation releases the lock is the right behavior — the timeout will fire slightly late but will fire.

**Essential `done.get()` guard.** After acquiring the lock, checking `done.get()` prevents the timeout from processing if the multicast completed normally while the timeout was waiting. This avoids double-completion.

**Well-designed reproducer test.** `MulticastParallelTimeoutDuringAggregationTest` makes the race deterministic: the `SlowAggregationStrategy` holds the lock for 2 seconds while the 500ms timeout fires mid-aggregation. Before the fix, `tryLock()` would fail and the exchange would hang. The `@Isolated` annotation and `CountDownLatch` coordination are appropriate.

### Checklist

| Item | Status |
|------|--------|
| Tests | ✅ New reproducer test + 177 existing aggregator tests |
| Docs / Upgrade guide | N/A (bug fix for internal race condition, no API change) |
| Commit convention | ✅ `CAMEL-24142: <description>` |
| Security | N/A |
| CI | ⏳ Pending |

**Static analysis:** semgrep (0 findings).

Clean, minimal, and correct fix for a race condition that has existed since CAMEL-16103 (2021). No concerns.

---

oscerd (approved):
LGTM — correct race fix. Switching the one-shot `timeout()` from `tryLock()` to `lock.lock()` with a post-acquire `done.get()` guard means a timeout that races with aggregation no longer silently drops: it now blocks for the per-task lock and, once acquired, either surfaces the timeout or (if completion won) returns via the guard. No deadlock — a single per-task `ReentrantLock` shared by `aggregate()`/`timeout()`/completion, no lock ordering; `done` is CAS-set under the lock in `doDone`/`doTimeoutDone`, and the cancelled timeout thread just acquires after release and returns. Good deterministic reproducer.

PR #24823, opened 2026-07-17
CAMEL-24138: Splitter - Fix failure tracker leaking into nested splits

davsclaus:
## Summary

The Splitter's `errorThreshold`/`maxFailedRecords` feature (CAMEL-23264) stores its `SplitFailureTracker` as a plain exchange property (`CamelSplitFailureTracker`). Since `exchange.copy()` shallow-copies the properties map, this tracker leaks into any nested splitter running inside the outer split.

When the inner splitter has no thresholds configured, `shouldContinueOnFailure` still finds the leaked outer tracker, records failures against it with inner indexes, clears the exception, and returns `true` — silently swallowing inner split failures.

**Fix:** Guard `shouldContinueOnFailure` and the `totalItems` increment so they only activate when *this* splitter instance has thresholds configured (`errorThreshold > 0 || maxFailedRecords > 0`). A splitter without thresholds now falls through to `super.shouldContinueOnFailure()`, preserving normal failure propagation.

- Inner split failures no longer silently succeed when an outer split has `maxFailedRecords`/`errorThreshold`
- Inner splits without thresholds no longer corrupt the outer tracker's `totalItems` count

## Test plan

- [x] New `SplitterNestedFailureTrackerLeakTest` with 3 scenarios:
  - Inner split failure propagates when outer has `maxFailedRecords` but inner does not
  - Inner split success still works with outer `maxFailedRecords`
  - Both outer and inner with `maxFailedRecords` — inner absorbs within threshold, outer unaffected
- [x] All 29 existing splitter threshold tests pass (no regressions)

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
@@ -417,10 +417,13 @@ public class Splitter extends MulticastProcessor {
                             in.setBody(part);
                         }
                         // track total items for SplitResult (works in both streaming and non-streaming mode)
-                        SplitFailureTracker tracker
-                                = original.getProperty(SPLIT_FAILURE_TRACKER, SplitFailureTracker.class);
-                        if (tracker != null) {
-                            tracker.incrementTotalItems();
+                        // only when THIS splitter has thresholds configured to avoid corrupting an outer tracker
+                        if (errorThreshold > 0 || maxFailedRecords > 0) {
+                            SplitFailureTracker tracker
+                                    = original.getProperty(SPLIT_FAILURE_TRACKER, SplitFailureTracker.class);
+                            if (tracker != null) {
+                                tracker.incrementTotalItems();
+                            }
                         }
                         return createProcessorExchangePair(index++, processor, newExchange, route);
                     } else {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
## Review: CAMEL-24138 — Splitter: Fix failure tracker leaking into nested splits

### What works well

**Clear root cause.** `SplitFailureTracker` is stored as a plain exchange property (`CamelSplitFailureTracker`). Since `exchange.copy()` shallow-copies properties, a nested splitter inherits the outer's tracker. When the inner splitter has no thresholds, `shouldContinueOnFailure()` still finds the leaked tracker, records failures against it, clears the exception, and returns `true` — silently swallowing inner split failures and corrupting the outer tracker's `totalItems` count.

**Minimal, targeted guards.** Two changes, both checking `errorThreshold > 0 || maxFailedRecords > 0`:
1. `shouldContinueOnFailure()` — falls through to `super` when this splitter has no thresholds, preventing a threshold-less inner splitter from using the outer's tracker
2. `incrementTotalItems()` — only accesses the tracker when this splitter has thresholds, preventing count corruption

**Good test coverage.** Three scenarios that exercise the key interactions:
- Inner failure propagates when outer has threshold but inner doesn't (the bug)
- Inner success still works with outer threshold (no regression)
- Both with thresholds — inner absorbs within its threshold, outer unaffected (both features compose)

### Minor observation

A more thorough isolation (e.g., per-splitter-instance property keys) would fully prevent any tracker cross-talk, but this guard is sufficient, less invasive, and follows the principle of least surprise — a splitter without thresholds simply doesn't interact with the tracker at all. If tracker isolation becomes needed for other reasons, it can be revisited.

### Checklist

| Item | Status |
|------|--------|
| Tests | ✅ New 3-scenario test + 29 existing splitter threshold tests |
| Docs / Upgrade guide | N/A (bug fix for feature introduced in CAMEL-23264, same release cycle) |
| Commit convention | ✅ `CAMEL-24138: <description>` |
| Security | N/A |
| CI | ⏳ Pending |

**Static analysis:** semgrep (0 findings on Splitter.java from prior scan).

Clean and correct. No concerns.

---

oscerd (approved):
LGTM. The failure-tracker scoping is correct: `errorThreshold`/`maxFailedRecords` are per-instance and each `split()` builds a distinct Splitter, so the tracker only exists when *that* instance has thresholds. The added guard in `shouldContinueOnFailure` and the `incrementTotalItems` block means a leaked outer tracker (shallow-copied via `exchange.copy()`) is ignored by an inner threshold-less splitter, which falls through to `super.shouldContinueOnFailure`; the both-thresholds case is a no-op for the guard since the inner `process()` overwrites the property with its own tracker. Three scenarios covered.

PR #24824, opened 2026-07-17
CAMEL-24139: Fix Splitter watermark advancing past unrouted items on streaming abort

davsclaus:
## Summary

- Fix `SplitterIterable` watermark count being inflated when `MulticastProcessor.doDone()` drains remaining stream items after an early abort
- Track whether the primary (processing) iterator has been created; only the primary iterator updates `SPLIT_WATERMARK_COUNT` eagerly on each `next()` call
- Drain iterators (created by `doDone()` to release pooled exchanges) no longer touch the watermark count
- Add test for `streaming()` + `stopOnException()` + `handled(true)` abort path that verifies the watermark covers only items actually routed

**Root cause:** `SplitterIterable.iterator()` creates new wrapper iterators sharing the same counting iterator. When `doDone()` drains remaining items, the shared `rawItemCount` is driven to the full stream size. The old code only set `SPLIT_WATERMARK_COUNT` when the iterator was exhausted (`hasNext()` returning false), so the drain's exhaustion stored the inflated count. With no exception on the exchange (handled or timeout), `updateWatermark()` persisted it — causing silent data loss on resume.

**Fix:** The primary iterator now eagerly sets `SPLIT_WATERMARK_COUNT` on each `next()` call. Subsequent drain iterators are marked non-primary and skip the watermark update.

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/Splitter.java
@@ -290,6 +290,10 @@ public class Splitter extends MulticastProcessor {
         private final Exchange original;
         // tracks individual (raw) item count, independent of grouping
         private final AtomicInteger rawItemCount = new AtomicInteger();
+        // tracks whether the primary (processing) iterator has been created;
+        // subsequent iterators (e.g. the drain in MulticastProcessor.doDone)
+        // must not update the watermark count (CAMEL-24139)
+        private boolean primaryIteratorCreated;
 
         private SplitterIterable(Exchange exchange, Object value) {
             this.original = exchange;
@@ -349,6 +353,11 @@ public class Splitter extends MulticastProcessor {
 
         @Override
         public Iterator<ProcessorExchangePair> iterator() {
+            // only the first (primary) iterator tracks watermark count;
+            // subsequent iterators (drain in doDone) must not inflate it (CAMEL-24139)
+            boolean isPrimary = !primaryIteratorCreated;
+            primaryIteratorCreated = true;
+
             return new Iterator<>() {
                 private final Processor processor = getProcessors().iterator().next();
                 private int index;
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM — clean fix for a subtle watermark inflation bug.**

The root cause is well-identified: when `handled(true)` clears the exception on the exchange, `updateWatermark()` proceeds (line 664's `getException() != null` guard doesn't trigger), but the old code only set `SPLIT_WATERMARK_COUNT` on iterator exhaustion in `hasNext()`. On an abort with handled exception, the primary iterator is abandoned mid-stream (never exhausts), then `doDone()` creates a drain iterator that exhausts the remaining items and inflates the count to the full stream size — causing silent data loss on resume.

**Why the fix works:**

1. `primaryIteratorCreated` flag correctly distinguishes the processing iterator from the drain iterator — sequential access (process → doDone) means no synchronization needed.

2. Moving watermark update from lazy (exhaustion in `hasNext()`) to eager (each `next()` call) ensures the count reflects only items actually routed, even when the primary is abandoned mid-stream.

3. The `isPrimary` boolean is captured as a local variable in the closure, so each iterator instance has its own immutable copy — clean and thread-safe.

4. The math checks out: `lastAbsoluteIndex = offset + count - 1`. Two items consumed (a, FAIL) → count=2 → watermark="1". Without fix, drain inflates count to 5 → watermark="4" → 3 items silently skipped on resume.

5. Test accurately reproduces the bug scenario: `streaming()` + `stopOnException()` + `handled(true)` with a 5-item list where item 2 fails.

Verified that `rawItemCount` is shared across iterators (via the counting decorator wrapping the raw iterator), but only the primary iterator writes to `SPLIT_WATERMARK_COUNT` — drain increments `rawItemCount` but the property is already set to the correct value.

oscerd (approved):
LGTM. Root cause confirmed: `rawItemCount` and the underlying iterator are shared across the wrapper iterators, and the drain iterator from `MulticastProcessor.doDone()` drove the count to the full stream size, which the old lazy set in `hasNext()` then persisted. Moving the watermark set to an eager, primary-only assignment in `next()` (and dropping the `hasNext()` block) makes the count reflect exactly the items the primary iterator dispatched. Test asserts the expected watermark for a routed-then-aborted stream. Nice.

(Heads-up: #24823 also edits this same `Splitter.next()` tracker region, so whichever of the two lands second will want a quick rebase.)

PR #24827, opened 2026-07-17
CAMEL-24144: InMemorySagaService - Remove coordinators from map after completion or compensation

davsclaus:
## Summary

`InMemorySagaService.newSaga()` puts every new coordinator into the `coordinators` ConcurrentHashMap, but nothing ever removes entries — not on `complete()`, not on `compensate()`, not on timeout. This causes unbounded memory growth in long-running applications.

**Fix:**
- Added package-private `removeSaga(String)` method to `InMemorySagaService`
- `InMemorySagaCoordinator` now calls `removeSaga()` after reaching terminal state (COMPENSATED or COMPLETED) in `doCompensate()` and `doComplete()`
- Removal happens in the `.thenApply()` block after the finalization chain completes, so all step callbacks have finished before the coordinator is removed

## Test plan

- [x] New `SagaCoordinatorCleanupTest` verifying coordinators are removed from the map after both completion and compensation
- [x] Updated `SagaTimeoutTest` to accept the additional valid outcome when a coordinator is removed before a late joiner tries to access it
- [x] All existing saga tests pass (21 tests across 7 test classes)

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/SagaCoordinatorCleanupTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/SagaCoordinatorCleanupTest.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.processor;
+
+import java.util.concurrent.TimeUnit;
+
+import org.apache.camel.ContextTestSupport;
+import org.apache.camel.Exchange;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.mock.MockEndpoint;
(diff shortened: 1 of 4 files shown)


gnodet (approved):

**LGTM — straightforward memory leak fix.**

The `coordinators` ConcurrentHashMap grows unboundedly because entries are never removed after reaching terminal state. The fix is minimal and correct:

1. **Removal placement** — `removeSaga()` is called in `.thenApply()` after `currentStatus.set(COMPENSATED/COMPLETED)`, ensuring all step callbacks in `doFinalize()` have completed before the coordinator is evicted from the map. This ordering is important — any code that references the coordinator during finalization still has a valid reference.

2. **Package-private visibility** — `removeSaga()` is appropriately scoped since only `InMemorySagaCoordinator` (same package) needs to call it.

3. **Timeout path covered** — timeout-triggered compensation flows through `doCompensate()`, so the removal happens there too.

4. **`SagaTimeoutTest` update** — correctly accounts for the race where a late-joining participant calls `begin()` after the coordinator has already been removed from the map, resulting in "Exchange is not part of a saga" (null coordinator) instead of "Cannot begin: status is COMPENSATING/COMPENSATED".

5. **Test coverage** — both terminal paths (completion and compensation) are tested with Awaitility to handle the async nature of the `.thenApply()` chain.

Clean fix for a real production concern in long-running applications.

oscerd (approved):
LGTM — real memory-leak fix. `InMemorySagaService` kept every coordinator in its `coordinators` map for the CamelContext lifetime, so a long-running app accumulated one entry per saga forever. Adding `removeSaga(sagaId)` (`coordinators.remove(...)`) and calling it after both the completion and compensation terminal transitions bounds the map to in-flight sagas. `remove` is idempotent so the two call sites are harmless, and since it only fires after a terminal state there's no risk of dropping a still-active coordinator.

Good test coverage: `SagaCoordinatorCleanupTest` asserts the coordinator is gone after both completion and compensation using `await().atMost(5s).untilAsserted(...)` (no `Thread.sleep`), and the `SagaTimeoutTest` assertion is correctly widened to also accept "Exchange is not part of a saga" — the expected outcome now that a completed saga's coordinator is removed rather than lingering in a COMPENSATED state. CI is green.

PR #24724, opened 2026-07-15
CAMEL-24084: Apply header filter strategy to structured-mode CloudEvent extension headers

oscerd:
## Summary

The Knative consumer filters inbound HTTP headers through `KnativeHttpHeaderFilterStrategy`, so Camel-internal headers (`Camel*`) are not mapped onto the message. When a CloudEvent is received in **structured content mode** (`application/cloudevents+json`), `CloudEventProcessors.decodeStructuredContent` mapped the remaining event fields (extensions) directly onto the message headers **without** applying any `HeaderFilterStrategy` — inconsistent with the binary/HTTP-header path.

This routes structured-mode extension fields through a `HeaderFilterStrategy` before setting them as headers, so `Camel*` headers (matched case-insensitively) are filtered consistently on both content modes. Ordinary CloudEvent extension attributes are unaffected.

## Changes
- `AbstractCloudEventProcessor`: add a `HeaderFilterStrategy` and a shared `mapExtensionAsHeader(...)` helper.
- `CloudEventProcessors`: route the extension mapping (all spec versions) through the helper.
- `KnativeHttpTest`: new test asserting a regular extension is propagated and a `Camel*`-named field is filtered.
- Upgrade guide (4.22): documents the consistency change.

## Notes
- On `main`/`4.21.x`, `DefaultHeaderFilterStrategy` filters `Camel*` inbound by default; the `4.18.x`/`4.14.x` backports will set `setInFilterStartsWith(CAMEL_FILTER_STARTS_WITH)` explicitly.
- JIRA: [CAMEL-24084](https://issues.apache.org/jira/browse/CAMEL-24084)

_Claude Code on behalf of Andrea Cosentino (@oscerd)_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24828, opened 2026-07-17
CAMEL-24085: Apply header filter strategy to IronMQ message envelope headers

oscerd:
### Motivation

When receiving a message with `preserveHeaders=true`, the IronMQ consumer deserializes a JSON envelope (a headers map and a body) and copies the embedded header entries onto the Camel message (`GsonUtil.copyFrom`, called from `IronMQConsumer`) without applying a `HeaderFilterStrategy`. The envelope is fully sender-controlled, so a sender can inject Camel-internal control headers (the `Camel*` namespace) onto the exchange.

### Changes

- `GsonUtil.copyFrom` now routes the embedded envelope headers through a `DefaultHeaderFilterStrategy` (`applyFilterToExternalHeaders`), so `Camel*` headers (matched case-insensitively) are no longer mapped onto the Camel message — consistent with the inbound header filtering performed by other consumers, and with the same approach as the structured-mode CloudEvent fix in CAMEL-24084 (#24724).
- New `IronMQConsumerHeaderFilterTest` pushing a crafted envelope through the existing `MockQueue`: ordinary headers still propagate, `Camel*` entries are filtered.
- Upgrade-guide entry in `camel-4x-upgrade-guide-4_22.adoc`.

Note: camel-ironmq is deprecated since 4.21, but still ships; the same envelope mapping exists on the 4.18.x/4.14.x maintenance branches, so this is a backport candidate (those branches need the explicit `setInFilterStartsWith` variant, pre-CAMEL-23543).

### Testing

- `mvn clean install` in `components/camel-ironmq`: all tests pass (new test + existing preserve-headers round-trip).

### DIFF
--- a/components/camel-ironmq/src/main/java/org/apache/camel/component/ironmq/GsonUtil.java
+++ b/components/camel-ironmq/src/main/java/org/apache/camel/component/ironmq/GsonUtil.java
@@ -20,10 +20,16 @@ import java.util.Map;
 
 import com.google.gson.Gson;
 import org.apache.camel.Message;
+import org.apache.camel.spi.HeaderFilterStrategy;
+import org.apache.camel.support.DefaultHeaderFilterStrategy;
 
 @Deprecated(since = "4.21")
 public final class GsonUtil {
     private static final Gson GSON = new Gson();
+
+    // Filters internal Camel headers (Camel*/camel*) embedded in the IronMQ message envelope, consistent with how
+    // other consumers apply a HeaderFilterStrategy to externally supplied headers.
+    private static final HeaderFilterStrategy HEADER_FILTER_STRATEGY = new DefaultHeaderFilterStrategy();
 
     private GsonUtil() {
     }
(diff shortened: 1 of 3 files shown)


gnodet (approved):

**LGTM — security fix for header injection via IronMQ message envelope.**

This is exactly the kind of consumer-side header filtering gap the project's security model flags (untrusted input mapped into the Exchange header map without a `HeaderFilterStrategy`). The fix is consistent with the CloudEvent structured-mode fix in CAMEL-24084 (#24724).

**Review notes:**

1. **Correct pattern** — `DefaultHeaderFilterStrategy` filters `Camel*` headers case-insensitively by default, matching the standard inbound filtering behavior of other consumers. The `!applyFilterToExternalHeaders()` negation correctly blocks filtered headers.

2. **Thread-safe** — `DefaultHeaderFilterStrategy` is stateless, so the `private static final` instance is safe for concurrent use across consumer threads.

3. **Existing-header precedence preserved** — the `target.getHeader(key) == null` check remains, so component-set headers (like `MESSAGE_ID`) aren't overwritten by envelope data.

4. **Test** — crafts a realistic attack payload with both `CamelFileName` (Pascal-case) and `camelExecCommandExecutable` (camel-case) injected headers. Verifies legitimate headers propagate while `Camel*` entries are filtered. Good case-insensitivity coverage.

5. **Upgrade guide** — clear migration note following the same structure as the CloudEvent entry. Appropriate given the behavioral change for `preserveHeaders=true` consumers.

6. **Backport note** — good call flagging the 4.18.x/4.14.x applicability with the `setInFilterStartsWith` variant needed for pre-CAMEL-23543 branches.

Minor: the component is deprecated since 4.21 but this fix is still worthwhile since it ships in 4.22 and the maintenance branches.

davsclaus (approved):

PR #24829, opened 2026-07-17
CAMEL-24145: camel-saga - InMemorySagaCoordinator compensate/complete now propagate finalization outcome

davsclaus:
## Summary

`InMemorySagaCoordinator.compensate()` and `complete()` returned `CompletableFuture.completedFuture(null)` unconditionally, meaning the exchange finished before any compensation/completion endpoint was invoked and finalization failures never propagated to the caller.

- `compensate()` and `complete()` now return the actual future from `doCompensate()`/`doComplete()` so the exchange waits for finalization (including retries) to finish
- When all retry attempts fail, the failure is propagated as a `RuntimeCamelException` on the exchange, consistent with the `camel-lra` coordinator behavior
- Added 2 new tests verifying failure propagation
- Added upgrade guide note for the behavior change

Fixes: [CAMEL-24145](https://issues.apache.org/jira/browse/CAMEL-24145)

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/SagaFailuresTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/SagaFailuresTest.java
@@ -19,10 +19,13 @@ package org.apache.camel.processor;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.camel.ContextTestSupport;
+import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.apache.camel.saga.InMemorySagaService;
 import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class SagaFailuresTest extends ContextTestSupport {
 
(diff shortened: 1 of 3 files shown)


gnodet (approved):

**LGTM — fixes a significant behavioral bug where saga finalization was fire-and-forget.**

The old code called `doCompensate(exchange)` / `doComplete(exchange)` inside the `if (doAction)` block but then fell through to `return CompletableFuture.completedFuture(null)` — the actual finalization future was discarded, so the exchange completed immediately and finalization failures were silently swallowed.

**Review notes:**

1. **Core fix** — returning the actual future from `doCompensate()`/`doComplete()` with a `.thenApply()` that converts `false` (all retries exhausted) into a `RuntimeCamelException` is the right approach. The exchange now waits for finalization to complete and sees failures.

2. **Consistent with `camel-lra`** — aligns the in-memory coordinator with the LRA coordinator's behavior of propagating finalization failures, which is the correct contract for callers that need to know whether compensation/completion succeeded.

3. **Clean refactoring** — flattening the `else` into a fall-through after the early `return` is structurally cleaner while being functionally identical for the non-action path (already COMPENSATING/COMPENSATED or COMPLETING/COMPLETED).

4. **Tests** — both paths (completion and compensation failure propagation) are tested using the existing `SagaFailuresTest` infrastructure with `maxFailures=3` to exhaust retries.

5. **Upgrade guide** — clearly documents the behavior change and provides a migration path (`MANUAL` completion mode) for routes that intentionally relied on fire-and-forget finalization. Well-phrased.

6. **Composability with #24827** — no conflict; this PR modifies `compensate()`/`complete()` while #24827 modifies `doCompensate()`/`doComplete()`. Both changes compose cleanly in the future chain.

oscerd (approved):
LGTM — real fix. `compensate()`/`complete()` previously discarded the `doFinalize` future, so a compensation/completion that exhausted its retries reported success. Returning `doCompensate(...).thenApply(res -> { if (!res) throw new RuntimeCamelException(...); return null; })` makes a `false` outcome surface as an exception that `SagaProcessor.handleSagaCompletion` sets on the exchange, so failures are no longer swallowed. The new `SagaFailuresTest` cases exercise real retry-exhaustion (maxRetryAttempts/maxFailures) and assert `result.getException()`, and the behavior change + MANUAL migration is documented in the 4.22 upgrade guide. Composes cleanly with the terminal-status handling.

PR #24830, opened 2026-07-17
CAMEL-24146: Fix FluentProducerTemplate leak in InMemorySagaCoordinator

davsclaus:
## Summary

`InMemorySagaCoordinator.doFinalize` created a new `FluentProducerTemplate` per compensation/completion call (including retries) and never stopped it. Each leaked template holds a started `ProducerTemplate` with a producer cache (default max 1000). With the default 5 retry attempts, a single saga finalization could leak up to N*5 templates.

**Fix:** Create a shared `ProducerTemplate` in `InMemorySagaService` (which already has `doStart`/`doStop` lifecycle) and reuse it across all coordinator finalization calls. The `FluentProducerTemplate` was overkill here — the only call was `.to(endpoint).withExchange(target).send()`, which delegates to `ProducerTemplate.send(Endpoint, Exchange)`.

Changes:
- `InMemorySagaService`: add a `ProducerTemplate` field, create in `doStart()`, stop in `doStop()`
- `InMemorySagaCoordinator`: replace per-call `createFluentProducerTemplate()` with the shared `sagaService.getProducerTemplate().send()`

## Test plan

- [x] All 24 existing saga tests pass (`SagaTest`, `SagaFailuresTest`, `SagaOptionsTest`, `SagaPropagationTest`, `SagaRemoveHeadersTest`, `SagaTimeoutTest`, `SagaComponentTest`)
- [x] `SagaFailuresTest` exercises the retry code path where the leak was most severe

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
+++ b/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaCoordinator.java
@@ -196,7 +196,7 @@ public class InMemorySagaCoordinator implements CamelSagaCoordinator {
         Exchange target = createExchange(exchange, endpoint, step);
 
         return CompletableFuture.supplyAsync(() -> {
-            Exchange res = camelContext.createFluentProducerTemplate().to(endpoint).withExchange(target).send();
+            Exchange res = sagaService.getProducerTemplate().send(endpoint, target);
             Exception ex = res.getException();
             if (ex != null) {
                 throw new RuntimeCamelException(res.getException());
--- a/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaService.java
+++ b/core/camel-support/src/main/java/org/apache/camel/saga/InMemorySagaService.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ScheduledExecutorService;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.Exchange;
+import org.apache.camel.ProducerTemplate;
 import org.apache.camel.support.service.ServiceSupport;
 import org.apache.camel.util.ObjectHelper;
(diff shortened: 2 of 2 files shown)


gnodet (approved):

**LGTM — straightforward resource leak fix.**

Every `doFinalize()` call created a new `FluentProducerTemplate` that was never stopped, leaking a `ProducerTemplate` with its producer cache on each invocation (including retries). The fix is clean:

1. **Proper lifecycle** — shared `ProducerTemplate` created in `doStart()`, stopped in `doStop()`, following the same pattern as `executorService`. Null guard handles restart scenarios.

2. **Thread-safe** — `ProducerTemplate` is designed to be shared across threads, so reusing it across concurrent coordinator finalization calls is correct.

3. **Simplification** — replacing `createFluentProducerTemplate().to(endpoint).withExchange(target).send()` with `producerTemplate.send(endpoint, target)` is a clean simplification — the fluent API added no value for this single-endpoint, exchange-based send.

4. **Consistent API** — `getProducerTemplate()` mirrors the existing `getExecutorService()` pattern on the same class.

Minimal change, clear impact.

oscerd (approved):
LGTM — clean leak fix. Creating a `FluentProducerTemplate` per finalization call leaked it (never closed); replacing it with a single lifecycle-managed `ProducerTemplate` created in `doStart()` and stopped+nulled in `doStop()` (mirroring the existing `executorService` pattern, null-guarded for restart) fixes that, and `getProducerTemplate().send(endpoint, target)` is semantically identical to the old `.to(endpoint).withExchange(target).send()`. `ProducerTemplate` is thread-safe for concurrent finalization. (No direct no-leak assertion, which is fair — hard to unit-test; covered indirectly by the existing retry-path test.)

PR #24831, opened 2026-07-17
CAMEL-24147: camel-lra - URL-encode callback query parameters and fix parseQuery split

davsclaus:
## Summary

Two related defects in camel-lra cause saga compensation/completion to silently fail when endpoint URIs contain query parameters or special characters:

- **`LRAUrlBuilder.query()`** concatenated keys and values raw with no `URLEncoder`. An endpoint URI like `seda:cancelOrder?concurrentConsumers=2&size=1000` corrupts the callback URL registered with the LRA coordinator.
- **`LRASagaRoutes.parseQuery()`** used `split("=")` without a limit, dropping everything after the second `=` in a value.

### Changes

- `LRAUrlBuilder.query()`: URL-encode both key and value with `URLEncoder.encode()`
- `LRASagaRoutes.parseQuery()`: use `split("=", 2)` to preserve values containing `=`
- Added `LRAUrlBuilderTest` with encoding and round-trip tests
- Added new test cases to `LRASagaRoutesTest` for encoded special characters and equals-in-value

## Test plan

- [x] New `LRAUrlBuilderTest` verifies encoding of special characters, round-trip fidelity, and realistic compensation URIs with query parameters
- [x] New `LRASagaRoutesTest` cases verify `parseQuery()` handles encoded `&`, `=`, `+` and preserves values with `=`
- [x] All 9 tests pass in `mvn verify`

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
@@ -69,8 +69,8 @@ public class LRASagaRoutes extends RouteBuilder {
             // first, split by parameter separator '&'
             // then collect the map with the variable name '[0]' and value '[1]', both url decoded
             result = Arrays.stream(queryStr.split("&")).collect(
-                    Collectors.toMap(element -> decode(saveArrayAccess(element.split("="), 0)),
-                            element -> decode(saveArrayAccess(element.split("="), 1))));
+                    Collectors.toMap(element -> decode(saveArrayAccess(element.split("=", 2), 0)),
+                            element -> decode(saveArrayAccess(element.split("=", 2), 1))));
 
         } else {
             LOG.debug("query param is empty, nothing to parse.");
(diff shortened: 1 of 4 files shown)


gnodet (approved):

**LGTM — fixes two related URL-handling bugs that silently corrupt LRA callback URLs.**

**Bug 1 — missing URL-encoding in `LRAUrlBuilder.query()`:**
An endpoint URI like `seda:cancelOrder?concurrentConsumers=2&size=1000` registered as a compensation callback would have its `?`, `&`, and `=` characters interpreted as part of the callback URL's own query structure, corrupting the registered callback. The fix correctly applies `URLEncoder.encode(value, StandardCharsets.UTF_8)` to both key and value.

**Bug 2 — `split("=")` without limit in `parseQuery()`:**
Values containing `=` (e.g., URL-encoded endpoint URIs after decoding, or base64 values) would be silently truncated at the second `=`. `split("=", 2)` correctly limits the split to `[key, rest-of-value]`.

**Review notes:**

1. Both fixes work in concert — the builder now encodes special characters, and the parser correctly handles the full value after decoding.

2. The round-trip test is excellent — verifies that `LRAUrlBuilder` encoding + `parseQuery` decoding is lossless for realistic compensation URIs with query parameters and special characters.

3. `URLEncoder.encode()` with `StandardCharsets.UTF_8` is the correct choice — it handles all RFC 3986 reserved characters and the `decode()` method already used in `parseQuery()` uses `URLDecoder.decode()` with UTF-8.

4. No upgrade guide needed — this is a pure bugfix for behavior that was silently broken (data corruption), not a behavioral change users might rely on.

PR #24832, opened 2026-07-17
CAMEL-24148: Fix saga option values overwritten on repeated step enlistment

davsclaus:
## Summary

`InMemorySagaCoordinator` keyed option values by `CamelSagaStep` identity (`Map<CamelSagaStep, Map<String, Object>>`). Since `SagaReifier` creates one step singleton per `saga()` DSL block, enlisting the same step multiple times in one saga (e.g. a loop/split calling a sub-route with `MANDATORY` propagation) caused each enlistment to overwrite the previous values. All compensation calls then received the **last** exchange's option values — silent data corruption.

**Fix:** Replace the dual `steps` list + `optionValues` map with a single `List<StepEnlistment>`, where each enlistment pairs a step with its own independent option values map. This ensures each enlistment preserves its own evaluated options.

Also includes the CAMEL-24146 fix (shared `ProducerTemplate` in `InMemorySagaService`) since both issues touch the same code.

Changes:
- `InMemorySagaCoordinator`: introduce `StepEnlistment` inner class, replace step-keyed storage with per-enlistment list
- `InMemorySagaService`: add shared `ProducerTemplate` lifecycle (CAMEL-24146)
- `SagaOptionsTest`: add `testOptionsPreservedOnRepeatedEnlistment` — enlists the same step twice with different `itemId` values, verifies each compensation receives its own value

## Test plan

- [x] All 25 saga tests pass (24 existing + 1 new)
- [x] New test `testOptionsPreservedOnRepeatedEnlistment` verifies both `item-A` and `item-B` are received in compensation

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/processor/SagaOptionsTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/SagaOptionsTest.java
@@ -16,13 +16,18 @@
  */
 package org.apache.camel.processor;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
+import org.apache.camel.model.SagaPropagation;
 import org.apache.camel.saga.InMemorySagaService;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class SagaOptionsTest extends ContextTestSupport {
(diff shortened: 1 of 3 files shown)


gnodet (approved):

**LGTM — fixes silent data corruption when the same saga step is enlisted multiple times.**

**Root cause:** `SagaReifier` creates one `CamelSagaStep` singleton per `saga()` DSL block, so using the step as a map key (`Map<CamelSagaStep, Map<String, Object>>`) meant that repeated enlistments (e.g., a loop or split calling a sub-route with `MANDATORY` propagation) would overwrite previous option values. All compensations then received the last exchange's values — silent data loss.

**Review notes:**

1. **Clean structural fix** — replacing the dual `steps` list + `optionValues` map with a single `List<StepEnlistment>` is the right model. Each enlistment is now an independent pairing of step + evaluated option values. No more identity-keyed map ambiguity.

2. **`StepEnlistment`** — minimal private inner class with package-private fields. `optionValues` is always non-null (empty HashMap when no options), simplifying `createExchange()` by removing the null check.

3. **Thread safety preserved** — `CopyOnWriteArrayList<StepEnlistment>` maintains the same concurrency guarantees as the previous `CopyOnWriteArrayList<CamelSagaStep>`.

4. **Option evaluation cleaned up** — removed the redundant `putIfAbsent` + `computeIfAbsent` dance, replaced with a straightforward local `HashMap` populated via enhanced for-each over `step.getOptions().entrySet()`.

5. **Test** — excellent reproduction: enlists the same step twice via `MANDATORY` propagation sub-route with different `itemId` values, forces failure, and verifies both "item-A" and "item-B" appear in compensation (sorted comparison). Directly exercises the bug scenario.

6. **Note:** includes the CAMEL-24146 `ProducerTemplate` changes (also in #24830) — trivial merge conflict if both land, since the changes are identical.

oscerd (approved):
LGTM — correct model fix. The bug is real: `optionValues` was keyed by the singleton `CamelSagaStep`, so enlisting the same step twice overwrote the first enlistment's values and both compensations read the last one. Pairing each enlistment's step with its own option values (`StepEnlistment` in a `CopyOnWriteArrayList`) and iterating those in `createExchange` gives each enlistment independent values while preserving single-enlistment behavior and the prior concurrency guarantees. The new test reproduces the exact scenario (same MANDATORY sub-route enlisted twice with itemId A/B, asserting both show up in compensation).

Heads-up: this overlaps the other in-flight saga PRs — #24830's `InMemorySagaService` `ProducerTemplate` change is a byte-identical subset of this one, and #24834 rewrites the same `beginStep`/`createExchange`/fields region incompatibly, so whichever of #24832/#24834 lands second will need a real rebase (they can't both apply cleanly).

PR #24834, opened 2026-07-17
CAMEL-24150, CAMEL-24151: Saga EIP fixes — shared service lifecycle, code review findings

davsclaus:
## Summary

- **CAMEL-24150**: Do not stop the shared saga service when an individual route's SagaProcessor is stopped. The service is context-scoped, not route-scoped.
  - Also fixed `SagaDownloader` (JBang/kamelet-main path) to use `context.addService()` instead of `main.bind()`, ensuring the `InMemorySagaService` is properly lifecycle-managed. Without this, removing `SagaProcessor.doStart()` would leave the registry-bound service unstarted. (Addresses @oscerd's review feedback.)
- **CAMEL-24151**: Fix medium and low findings from a July 2026 Saga EIP code review across `camel-core`, `camel-saga`, and `camel-lra`:
  - **M1**: LRAClient now accepts 2xx range (not just 200) and includes HTTP status, body, and LRA URL in error messages
  - **M2**: Timeout `ScheduledFuture`s stored and cancelled on compensate/complete
  - **M3**: Status machine uses `whenComplete` so terminal status is always set, even on exceptional completion
  - **M4**: Warn at reifier time when MANUAL completion is used without a timeout
  - **M5**: `ifNotException` wraps `code.run()` in try/catch to prevent exception swallowing
  - **M6**: Route setup moved from `LRASagaService.setCamelContext()` to `doStart()`
  - **L1**: `SagaProcessor.toString()` returns actual id, not literal `"id"`
  - **L2**: Removed dead `putIfAbsent` code; consistent `ConcurrentHashMap` usage
  - **L3**: `CompletableFuture.failedFuture()` replaces `supplyAsync(() -> throw)`
  - **L4**: Fixed double-`?` URL risk in LRAClient join
  - **L5**: Validate `coordinatorUrl` and `localParticipantUrl` at `doStart`
  - **L6**: `Optional.ifPresent()` instead of `.map()` for side effects
  - **L7**: Removed unused `CamelContext` parameter from SagaProcessor hierarchy
  - **L8**: EIP doc corrected — no default timeout exists
  - **L9**: SagaProducer error message says "cannot compensate" on the compensate path
  - **M7** filed as follow-up: [CAMEL-24192](https://issues.apache.org/jira/browse/CAMEL-24192)

## Test plan

- [x] All 26 saga tests pass (`SagaTest`, `SagaFailuresTest`, `SagaTimeoutTest`, `SagaPropagationTest`, `SagaOptionsTest`, `SagaRemoveHeadersTest`, `SagaSharedServiceRouteStopTest`, `SagaComponentTest`)
- [x] `camel-lra` tests pass
- [x] `camel-saga` tests pass
- [x] `camel-kamelet-main` builds cleanly with SagaDownloader change
- [x] All changed files compile cleanly

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAClient.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRAClient.java
@@ -125,7 +125,8 @@ public class LRAClient implements Closeable {
 
             String lraEndpoint = lra.toString();
             if (step.getTimeoutInMilliseconds().isPresent()) {
-                lraEndpoint = lraEndpoint + "?" + HEADER_TIME_LIMIT + "=" + step.getTimeoutInMilliseconds().get();
+                String separator = lraEndpoint.contains("?") ? "&" : "?";
+                lraEndpoint = lraEndpoint + separator + HEADER_TIME_LIMIT + "=" + step.getTimeoutInMilliseconds().get();
             }
             HttpRequest request = prepareRequest(URI.create(lraEndpoint), exchange)
                     .setHeader(HEADER_LINK, link.toString())
@@ -138,8 +139,10 @@ public class LRAClient implements Closeable {
         }, sagaService.getExecutorService())
                 .thenCompose(Function.identity())
                 .thenApply(response -> {
-                    if (response.statusCode() != HttpURLConnection.HTTP_OK) {
-                        throw new RuntimeCamelException("Cannot join LRA");
+                    int status = response.statusCode();
+                    if (status >= HttpURLConnection.HTTP_BAD_REQUEST) {
+                        throw new RuntimeCamelException(
+                                "Cannot join LRA " + lra + " (HTTP " + status + "): " + response.body());
                     }
 
                     return null;
@@ -155,8 +158,10 @@ public class LRAClient implements Closeable {
         CompletableFuture<HttpResponse<String>> future = client.sendAsync(request, HttpResponse.BodyHandlers.ofString());
 
         return future.thenApply(response -> {
-            if (response.statusCode() != HttpURLConnection.HTTP_OK) {
-                throw new RuntimeCamelException("Cannot complete LRA");
+            int status = response.statusCode();
+            if (status >= HttpURLConnection.HTTP_BAD_REQUEST) {
+                throw new RuntimeCamelException(
+                        "Cannot complete LRA " + lra + " (HTTP " + status + "): " + response.body());
             }
 
             return null;
@@ -172,8 +177,10 @@ public class LRAClient implements Closeable {
         CompletableFuture<HttpResponse<String>> future = client.sendAsync(request, HttpResponse.BodyHandlers.ofString());
 
         return future.thenApply(response -> {
-            if (response.statusCode() != HttpURLConnection.HTTP_OK) {
-                throw new RuntimeCamelException("Cannot compensate LRA");
+            int status = response.statusCode();
+            if (status >= HttpURLConnection.HTTP_BAD_REQUEST) {
+                throw new RuntimeCamelException(
+                        "Cannot compensate LRA " + lra + " (HTTP " + status + "): " + response.body());
             }
 
             return null;
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaCoordinator.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaCoordinator.java
@@ -41,9 +41,7 @@ public class LRASagaCoordinator implements CamelSagaCoordinator {
         try {
             sagaStep = LRASagaStep.fromCamelSagaStep(step, exchange);
         } catch (RuntimeException ex) {
-            return CompletableFuture.supplyAsync(() -> {
-                throw ex;
-            });
+            return CompletableFuture.failedFuture(ex);
         }
         return sagaService.getClient().join(this.lraURL, sagaStep, exchange);
     }
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaService.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaService.java
@@ -79,13 +79,23 @@ public class LRASagaService extends ServiceSupport implements StaticService, Cam
 
     @Override
     public void registerStep(CamelSagaStep step) {
-        // Register which uris should be exposed
-        step.getCompensation().map(Endpoint::getEndpointUri).map(this.sagaURIs::add);
-        step.getCompletion().map(Endpoint::getEndpointUri).map(this.sagaURIs::add);
+        step.getCompensation().map(Endpoint::getEndpointUri).ifPresent(this.sagaURIs::add);
+        step.getCompletion().map(Endpoint::getEndpointUri).ifPresent(this.sagaURIs::add);
     }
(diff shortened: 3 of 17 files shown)


gnodet (approved):

**LGTM — removes incorrect shared-service lifecycle management from individual route processors.**

The saga service is a context-level singleton added via `context.addService()`, which makes the CamelContext its lifecycle owner. `SagaProcessor.doStart()`/`doStop()` calling `ServiceHelper.startService()`/`stopService()` on the shared service was a lifecycle ownership violation — stopping one saga route would shut down the executor, producer template, and LRA client for every other saga route in the context.

**Review notes:**

1. **Pure deletion** — 12 lines removed, 0 production lines added. The safest kind of fix.

2. **`doStart()` removal is safe** — `context.addService(sagaService)` already starts the service before any route processor's `doStart()` runs.

3. **`doStop()` removal is the fix** — prevents one route shutdown from cascading to all saga routes. Particularly important with the shared `ProducerTemplate` added in CAMEL-24146 (#24830) — that would also be stopped prematurely.

4. **Tests** — structural check (service stays started after route stop) plus functional check (compensation on route-b works after route-a stops). Both scenarios directly exercise the bug.

Nice capstone to the saga fix series (CAMEL-24144 through CAMEL-24150).

oscerd (commented):
The lifecycle change is correct for the common `context.addService(sagaService)` path (covered by the new `SagaSharedServiceRouteStopTest` and by camel-lra), and the M1–L9 cleanups look sound — timeout `ScheduledFuture` cancellation, terminal status on exceptional completion, the `failedFuture`/`ifPresent`/message tidy-ups, and the LRAClient `<400` acceptance.

One thing worth verifying before merge, though: this deletes `SagaProcessor.doStart()`/`doStop()` (which did `ServiceHelper.startService(sagaService)` — `SagaProcessor.java:175-181` on main). That looks like the only place a saga service resolved as a **registry bean** (rather than via `context.addService`) got started. The kamelet-main/JBang path binds it exactly that way — `SagaDownloader` does `registry bind`, not `addService`, and `SagaReifier.resolveSagaService()` falls through to `mandatoryFindSingleByType(CamelSagaService.class)` for registry beans with no `addService`. If registry `Service` beans aren't auto-started elsewhere, `getExecutorService()`/`getProducerTemplate()` would be null at the first saga step → NPE. The new test only exercises the `addService` path, so it wouldn't catch this. Could you confirm the registry-bound saga path still starts the service (e.g. run a JBang saga route)? If it doesn't, having `SagaDownloader` use `context.addService(...)` would cover it.

Minor: removing the `CamelContext` constructor param from `SagaProcessor` + its six subclasses (and `SagaProcessorBuilder.camelContext()`) is a breaking signature change on public-ish processor classes — worth an upgrade-guide line. And as with #24832, this collides with the other three in-flight saga PRs on `InMemorySagaCoordinator`, so it'll need rebasing whenever it lands relative to them.

gnodet (approved):

Re-reviewed after rebase (3 new commits replacing the previous changeset). All 15 files reviewed.

## CAMEL-24150: Shared service lifecycle

The core fix — removing `ServiceHelper.startService/stopService(sagaService)` from `SagaProcessor.doStart()/doStop()` — is correct. The saga service is context-scoped, not route-scoped. Stopping one route was stopping the shared service, breaking all other saga routes.

The new `SagaSharedServiceRouteStopTest` properly validates both that the service stays started and that compensation on a remaining route still works after stopping another route. Good use of `MockEndpoint.assertIsSatisfied(context, 10, TimeUnit.SECONDS)` per project guidelines.

## CAMEL-24151: Code review findings (M1–M7, L1–L9)

All well-implemented:

| Finding | Change | Verdict |
|---------|--------|---------|
| **M1** | LRAClient accepts 2xx/3xx (>=400 = error), includes status+body+URL in error msgs | ✅ Correct — 201/204 are valid HTTP success |
| **M2** | Timeout `ScheduledFuture`s stored and cancelled on compensate/complete | ✅ Correct — prevents post-completion timeout triggering compensation |
| **M3** | `whenComplete` replaces `thenApply` for terminal status + saga cleanup | ✅ Correct — ensures status is always set, even on exceptional path; prevents saga leaks |
| **M4** | WARN at reifier time for MANUAL completion without timeout | ✅ Correct — catches a common misconfiguration |
| **M5** | `ifNotException` wraps `code.run()` in try/catch | ✅ Correct — prevents uncaught exceptions from swallowing the error |
| **M6** | Route setup moved from `setCamelContext()` to `doStart()` | ✅ Correct — aligns with Camel lifecycle |
| **L1** | `toString()` returns `id` not literal `"id"` | ✅ |
| **L3** | `CompletableFuture.failedFuture()` replaces `supplyAsync(() -> throw)` | ✅ Cleaner, avoids ForkJoinPool |
| **L4** | `?` vs `&` separator in LRAClient join URL | ✅ Prevents double-`?` |
| **L5** | Validate coordinatorUrl/localParticipantUrl at doStart | ✅ Fail-fast |
| **L6** | `ifPresent()` instead of `map()` for side effects | ✅ Correct Optional usage |
| **L7** | Removed unused CamelContext from SagaProcessor hierarchy | ✅ Consistent across all 6 subclasses + builder + reifier |
| **L8** | Doc: "no default timeout exists" (was misleading) | ✅ |
| **L9** | Error message says "cannot compensate" on compensate path | ✅ |

## One note

The constructor signature change for `SagaProcessor` and its subclasses is technically a public API change (the subclass constructors are `public`). Since these are created through `SagaProcessorBuilder`, real-world impact should be zero, and the justification (removing an unused parameter) is sound. Worth noting in the upgrade guide if not already there.

LGTM 👍

davsclaus:
Thanks for the thorough review @oscerd! Both points are now addressed in the latest commit:

1. **Registry-bound saga service**: Good catch — `SagaDownloader` was using `main.bind()` which only puts the `InMemorySagaService` in the registry without lifecycle management. With `SagaProcessor.doStart()` removed, the service would never be started in the JBang/kamelet-main path. Fixed by switching to `context.addService(new InMemorySagaService())` with a guard against duplicate registration when multiple saga definitions exist.

2. **Constructor signature change**: The `SagaProcessor` subclass constructors are `public` but are only created through `SagaProcessorBuilder` — real-world impact is zero. Will add an upgrade guide note if needed.

gnodet (approved):

**Re-review after new commit (fabcd706): SagaDownloader uses addService instead of registry bind** ✅

Good improvement — this directly addresses the shared service lifecycle theme of CAMEL-24150:

1. **`addService()` instead of `bind()`** — properly integrates `InMemorySagaService` with the CamelContext lifecycle (start/stop/shutdown), whereas the old `main.bind("inMemorySagaService", ...)` only put it in the registry with no lifecycle management. This was exactly the class of bug this PR set out to fix.

2. **Idempotency guard** — `ctx.hasService(CamelSagaService.class) == null` prevents adding duplicate saga services if the reifier runs multiple times (e.g., multiple saga routes in the same context).

3. **Removed `KameletMain` parameter** — `registerDownloadReifiers()` no longer needs a reference to `KameletMain`, which is a cleaner API. The reifier lambda captures the route's `CamelContext` directly, which is the right scope.

4. **try-catch on `addService()`** — correct since `addService` can throw; wrapping in `RuntimeException` is appropriate for a reifier context where checked exceptions can't propagate.

Still LGTM with this addition.

PR #24835, opened 2026-07-17
CAMEL-24191: camel-aws2-kinesis - fix KCL consumer session credentials and checkpoint/error handling

oscerd:
## Description

Fixes [CAMEL-24191](https://issues.apache.org/jira/browse/CAMEL-24191) (split out of the AWS review umbrella CAMEL-24156). Two defects in the KCL-based Kinesis consumer (`KclKinesis2Consumer`):

1. **Session credentials silently ignored.** In both the DynamoDB and CloudWatch async-client builders the branch order tested `accessKey && secretKey` before `accessKey && secretKey && sessionToken`, so the session-token branch was unreachable and those clients never used STS temporary credentials. The session-credentials branch is now checked first in both builders.

2. **Checkpoint/error handling lost or reprocessed data.** `processRecords` wrapped processing in `catch (Throwable t)` that logged a message **without** the exception and did not rethrow, so the scheduler later checkpointed past records that failed to process (data loss); it also never checkpointed after a successful batch, so a crash reprocessed the whole lease. It now:
   - checkpoints **only after** the batch has been processed successfully;
   - on failure, routes the exception to the consumer `ExceptionHandler` (with stack trace) and does **not** checkpoint, so KCL redelivers the batch from the last checkpoint.

## Tests

Adds `KclKinesis2ConsumerRecordProcessingTest` — asserts the checkpoint is taken after a successful batch and is **never** taken when processing throws. Existing kinesis tests continue to pass.

## Backport

Related to CAMEL-24156.

---

### DIFF
--- a/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/KclKinesis2Consumer.java
+++ b/components/camel-aws/camel-aws2-kinesis/src/main/java/org/apache/camel/component/aws2/kinesis/KclKinesis2Consumer.java
@@ -92,6 +92,12 @@ public class KclKinesis2Consumer extends DefaultConsumer {
         if (ObjectHelper.isEmpty(getEndpoint().getConfiguration().getDynamoDbAsyncClient())) {
             DynamoDbAsyncClientBuilder clientBuilder = DynamoDbAsyncClient.builder();
             if (ObjectHelper.isNotEmpty(configuration.getAccessKey())
+                    && ObjectHelper.isNotEmpty(configuration.getSecretKey())
+                    && ObjectHelper.isNotEmpty(configuration.getSessionToken())) {
+                clientBuilder = clientBuilder.credentialsProvider(StaticCredentialsProvider
+                        .create(AwsSessionCredentials.create(configuration.getAccessKey(), configuration.getSecretKey(),
+                                configuration.getSessionToken())));
+            } else if (ObjectHelper.isNotEmpty(configuration.getAccessKey())
                     && ObjectHelper.isNotEmpty(configuration.getSecretKey())) {
                 clientBuilder = clientBuilder.credentialsProvider(StaticCredentialsProvider.create(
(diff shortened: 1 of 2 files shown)


gnodet (approved):

Clean, well-scoped fix for two real bugs in the KCL-based Kinesis consumer. Both issues have been present since the original session-credential support was added (CAMEL-20912/20913, commit 8a2116f).

### Fix 1 — Session credentials branch ordering ✅

The `if/else-if` chain checked `accessKey && secretKey` before the more specific `accessKey && secretKey && sessionToken`. Since the 2-param condition is a strict superset of the 3-param one, the session-token branch was dead code — STS temporary credentials silently fell back to basic credentials, dropping the session token. The fix correctly reorders to most-specific-first in both the DynamoDB and CloudWatch builders.

### Fix 2 — Checkpoint/error handling ✅

The old `processRecords()` had `catch (Throwable t) { LOG.error("..."); }` — three problems in one line:
1. Exception swallowed (no stack trace logged)
2. No checkpoint on success → crash = reprocess entire lease (duplication)
3. No rethrow/routing on failure → KCL could advance past failed records (data loss)

The new code:
- Checkpoints **only after** a successful batch — correct at-least-once semantics
- Routes processing failures to `ExceptionHandler` **without** checkpointing → KCL redelivers from last checkpoint
- Catches `ShutdownException | InvalidStateException` for checkpoint failures — consistent with `shardEnded()` and `shutdownRequested()` handlers in the same class

The `forEach` → plain `for` loop change is a nice simplification — the lambda wrapping `Exception` → `RuntimeException` is no longer needed since the for-loop naturally propagates to the surrounding try-catch.

### Test ✅

`KclKinesis2ConsumerRecordProcessingTest` directly constructs the inner `CamelKinesisRecordProcessor` (package-private, so accessible from same-package test) and verifies both the success-checkpoint and failure-no-checkpoint behaviors with mocked `RecordProcessorCheckpointer`.

No concerns. Good fix for a genuine data-safety issue.

davsclaus (approved):

PR #24847, opened 2026-07-17
CAMEL-24193: camel-aws-secrets-manager - fix SQS-notification secret refresh NPE and queue draining

oscerd:
## Description

Fixes [CAMEL-24193](https://issues.apache.org/jira/browse/CAMEL-24193). In the SQS-notification branch of `CloudTrailReloadTriggerTask` (the `aws-secret-refresh` task with `useSqsNotification=true`):

1. **NPE on a missing queue.** `QueueDoesNotExistException` from `receiveMessage` was caught and only logged, leaving `messageResult` null, which was then dereferenced by `for (Message m : messageResult.messages())` — an NPE on every poll (swallowed by the outer catch and logged as an error). The loop is now guarded against a null receive result.
2. **Queue never drained.** `deleteMessage` ran only inside `if (matchSecret) { if (isReloadEnabled()) { … } }` with a `break` after the first match, so non-matching messages — and everything when reload was disabled — were never deleted and were redelivered on every poll (unbounded queue growth). Every examined message is now deleted (matched → reload recorded; unmatched or malformed → discarded), so the queue is drained. A malformed body is logged and discarded rather than aborting the whole batch.
3. **`ObjectMapper` per message** replaced with a single reused instance.

## Tests

Adds `CloudTrailReloadTriggerTaskSqsTest` — verifies every message is deleted even when only some match (the drain fix; the old code deleted at most one), and that a missing queue does not throw / does not delete.

## Scope note

Two behavior-sensitive observations in the same class are intentionally **not** included here: the CloudTrail-polling branch advancing its cursor by newest-event-time + 1000ms, and `matchSecret` using substring `contains()`. They can be addressed separately.

## Backport

---

### DIFF
--- a/components/camel-aws/camel-aws-secrets-manager/pom.xml
+++ b/components/camel-aws/camel-aws-secrets-manager/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>camel-test-spring-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-version}</version>
+            <scope>test</scope>
+        </dependency>
 
         <!-- test infra -->
         <dependency>
--- a/components/camel-aws/camel-aws-secrets-manager/src/main/java/org/apache/camel/component/aws/secretsmanager/vault/CloudTrailReloadTriggerTask.java
+++ b/components/camel-aws/camel-aws-secrets-manager/src/main/java/org/apache/camel/component/aws/secretsmanager/vault/CloudTrailReloadTriggerTask.java
@@ -92,6 +92,7 @@ public class CloudTrailReloadTriggerTask extends ServiceSupport implements Camel
     private static final String CAMEL_AWS_VAULT_URI_ENDPOINT_OVERRIDE = "CAMEL_AWS_VAULT_URI_ENDPOINT_OVERRIDE";
 
     private static final Logger LOG = LoggerFactory.getLogger(CloudTrailReloadTriggerTask.class);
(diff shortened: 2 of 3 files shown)


gnodet (approved):

Solid fix for three real defects in the SQS-notification branch of `CloudTrailReloadTriggerTask`.

### Fix 1 — NPE on missing queue ✅

`QueueDoesNotExistException` was caught and logged, but `messageResult` stayed null → `messageResult.messages()` threw NPE on every poll (swallowed by the outer catch). Fixed with `if (messageResult != null)` guard.

### Fix 2 — Queue never drained ✅

`deleteMessage` was buried inside `if (matchSecret) { if (isReloadEnabled()) { … break; } }` — so:
- Non-matching messages → never deleted → redelivered every poll
- Matching but reload-disabled → never deleted → redelivered every poll
- After first match → `break` → remaining messages never examined

Fix correctly moves `deleteMessage` outside all conditionals (every examined message is deleted) and removes the `break` so ALL messages in the batch are processed. This prevents unbounded queue growth.

### Fix 3 — ObjectMapper per message ✅

Replaced `new ObjectMapper()` per message with a static `MAPPER` instance — small but sensible.

### Bonus — Malformed message handling ✅

`JsonProcessingException` is now caught per-message (logged as warn, message still deleted) rather than wrapped in `RuntimeException` which would abort the entire batch. Good resilience improvement.

### Test ✅

`CloudTrailReloadTriggerTaskSqsTest` uses reflection to inject the mocked `SqsClient` and verifies:
1. Both matching and non-matching messages are deleted (drain fix)
2. Missing queue doesn't throw/NPE and doesn't attempt deletes

The scope note explicitly calling out `contains()`-based matching and CloudTrail cursor advancement as separate concerns is appreciated — clean separation.

No concerns.

PR #24852, opened 2026-07-17
CAMEL-24155: camel-salesforce - Fix streamQueryResult iterator hang when queryMore fails

davsclaus:
## Summary

Fixes a bug where `QueryResultIterator.next()` hangs the route thread forever when `queryMore` fails during paginated result iteration with `streamQueryResult=true`.

Three interrelated bugs in the callback:
- The `exception` parameter was completely ignored — on any failure, `objectMapper.readValue(null, ...)` would throw
- `latch.countDown()` was inside the `try` block, not in `finally` — any exception skipped it, deadlocking the route thread
- `latch.await()` had no timeout — the thread would block indefinitely

### Fix
- Check `exception != null` at callback entry and store it in an error holder
- Catch all exceptions (not just `IOException`) and store them in the error holder
- Move `latch.countDown()` to the `finally` block so it is always reached
- Add a 30-second timeout to `latch.await()`
- After the latch, check the error holder and rethrow on the route thread as `RuntimeCamelException`

## Test plan

- [x] Added `QueryResultIteratorTest` with 5 tests:
  - Single-page iteration (happy path)
  - Multi-page iteration (happy path with `queryMore`)
  - `queryMore` exception does not hang (synchronous callback)
  - `queryMore` exception from separate thread does not hang (async callback, mimics production)
  - Malformed response does not hang (parse error)
- [x] `mvn verify` passes on `camel-salesforce-component`

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/processor/QueryResultIterator.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/processor/QueryResultIterator.java
@@ -23,6 +23,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.camel.RuntimeCamelException;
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — fixes a real hang bug in the Salesforce `streamQueryResult` iterator.

**The bug**: `QueryResultIterator.next()` blocks forever when `queryMore` fails, because `latch.countDown()` was only on the happy path inside the `try` block. Three failure modes all led to infinite hang:
1. Callback receives a non-null `exception` → code tries `objectMapper.readValue(null, ...)` → NPE → latch never counted down
2. Malformed JSON response → `IOException` caught, rethrown as `RuntimeCamelException` → latch never counted down  
3. Any other unexpected exception → same result

**The fix** (all correct):
- Move `latch.countDown()` into `finally` — guarantees countdown regardless of outcome
- Check `exception != null` first, store in `AtomicReference<Exception> errorHolder`
- Widen catch from `IOException` to `Exception` — store in errorHolder instead of rethrowing
- Add 30s timeout on `latch.await()` — safety net against edge cases
- Check `errorHolder` after await and surface as `RuntimeCamelException`

**Tests** (190 lines, 5 tests) — thorough coverage:
- Single-page and multi-page happy paths
- Callback exception (401 session expired) → no hang, error surfaced
- Async callback from separate thread → no hang
- Malformed JSON response → no hang

Clean, minimal fix. The `finally` pattern for latch countdown is textbook correct for this kind of async callback coordination.

oscerd (approved):
LGTM — good fix for a real hang. When `queryMore` failed, the async callback never signaled completion, so the iterator blocked forever. Capturing the callback's exception in an `AtomicReference` and (a) bounding the wait with a timeout (`RuntimeCamelException("Timeout waiting for Salesforce queryMore response")`) and (b) rethrowing the captured error (`"Failed to fetch next page of Salesforce query results"`) means the iterator now always terminates — either surfacing the real failure or timing out — instead of deadlocking the consuming route. Broadening the `catch (IOException)` to `catch (Exception)` closes the gap for non-IO failures too. Nice test coverage: the success path (A1/A2 → `NoSuchElementException`) and a `queryMore`-throws path asserting the exception propagates.

PR #24854, opened 2026-07-17
CAMEL-24194: camel-aws-config - removeConformancePack reads the correct header

oscerd:
## Description

Fixes [CAMEL-24194](https://issues.apache.org/jira/browse/CAMEL-24194). `AWSConfigProducer.removeConformancePack` guarded on the `CamelAwsConformancePackName` header but read the value from `CamelAwsConfigRuleName`:

```java
if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(AWSConfigConstants.CONFORMACE_PACK_NAME))) {
    String conformancePackName = exchange.getIn().getHeader(AWSConfigConstants.RULE_NAME, String.class);
```

So the documented `CamelAwsConformancePackName` header was ignored and the delete request was built with the (usually null) rule-name value. It now reads the value from `CONFORMACE_PACK_NAME`. The sibling `putConformancePack` already reads it correctly.

## Tests

Adds `AWSConfigRemoveConformancePackTest` — captures the `DeleteConformancePackRequest` and asserts it uses the conformance-pack-name header even when a different `CamelAwsConfigRuleName` header is present (the old code would have used the rule name).

## Backport

---

### DIFF
--- a/components/camel-aws/camel-aws-config/pom.xml
+++ b/components/camel-aws/camel-aws-config/pom.xml
@@ -70,6 +70,12 @@
             <artifactId>camel-test-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-version}</version>
+            <scope>test</scope>
+        </dependency>
 
         <!-- test infra -->
         <dependency>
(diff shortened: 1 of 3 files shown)


gnodet (approved):

**LGTM** — textbook copy-paste bug fix.

**The bug**: `removeConformancePack()` checked whether the `CONFORMACE_PACK_NAME` header was set, but then read the value from the wrong header (`RULE_NAME`). This meant the conformance pack name was always silently wrong (it got whatever rule name was set, or `null`).

**The fix**: One-line change — reads from `CONFORMACE_PACK_NAME` instead of `RULE_NAME`.

**Test**: Well-designed — sets both headers to different values (`"my-pack"` vs `"some-rule"`) and uses `ArgumentCaptor` to verify the `DeleteConformancePackRequest` received `"my-pack"`, proving the correct header is now read.

**Dependency**: `mockito-junit-jupiter` added as test-scoped dependency — appropriate for the new mock-based test.

davsclaus (approved):

PR #24848, opened 2026-07-17
[backport camel-4.18.x] CAMEL-24191: camel-aws2-kinesis - fix KCL consumer session credentials and checkpoint/error handling

oscerd:
Backport of #24835 to `camel-4.18.x`. Fixes [CAMEL-24191](https://issues.apache.org/jira/browse/CAMEL-24191) on the 4.18.x line: reorder the DynamoDB/CloudWatch credential branches so session tokens are used, and make the KCL `processRecords` checkpoint only after a successful batch (routing failures to the ExceptionHandler without checkpointing, so KCL redelivers). Includes `KclKinesis2ConsumerRecordProcessingTest`. Related to CAMEL-24156.

---
_Claude Code on behalf of oscerd_

PR #24857, opened 2026-07-17
CAMEL-24161: camel-salesforce - fix medium-severity bugs from code review

davsclaus:
## Summary

Fixes 11 medium-severity bugs found during the July 2026 deep code review of camel-salesforce ([CAMEL-24161](https://issues.apache.org/jira/browse/CAMEL-24161)). Companion to #24848 which fixed the high-severity findings.

**Auth / session / HTTP client:**
- Mask `refreshToken` in `SalesforceLoginConfig.toString()` — was printed in clear text while all other secrets were masked (since CAMEL-15425)
- Check `body.rewind()` return value in `SalesforceSecurityHandler` — multipart uploads silently sent empty body on 401 retry
- Detect Jetty failure chunks in `AbstractClientBase` replay-buffering loop — mid-read I/O error produced silently truncated requests
- Use `put()` instead of `add()` in `DefaultRawClient.setAccessToken()` — auth retry produced duplicate `X-SFDC-Session`/`Authorization` headers
- Coordinate re-logins via `session.login()` on worker pool in `SalesforceSecurityHandler` — N concurrent 401s now share one login instead of each firing its own

**Streaming / Pub/Sub:**
- Use `getOrDefault()` in `SubscriptionHelper` connect listener and clean up `channelsToSubscribe` on unsubscribe — prevents NPE and silent channel loss on concurrent unsubscribe
- Invoke `consumer.getExceptionHandler()` on per-event decode failures in `PubSubApiClient` — events were silently lost (log only, replay ID advanced past them)
- Stop infinite retry on corrupted replay ID when `fallbackToLatestReplayId=false` in `PubSubApiClient`, and make replay preset state per-observer instead of shared — prevents infinite loop and cross-consumer preset overwrite

**REST / Bulk / Composite:**
- Clean up `RESPONSE_CLASS_DEFERRED`/`RESPONSE_CLASS_PREFIX` exchange properties in `JsonRestProcessor.processResponse()` — leaked into subsequent producers on the same exchange
- Preserve exception when Salesforce returns both error body and exception in `CompositeApiProcessor` — HTTP failures were silently reported as success in raw/batch/composite modes
- Fix `deleteSObjectWithId` in `AbstractRestProcessor` to restore DTO fields *before* completing the exchange (matching all sibling operations), and null-check external-id field value to prevent NPE

## Test plan

- [x] All existing camel-salesforce-component tests pass (190 run, 0 failures)
- [x] Updated `PubSubApiTest.testInvokesExceptionHandlerWhenReplayIdIsCorruptedAndFallbackToLatestReplayIdIsDisabled` to verify fixed behavior (stops after one attempt instead of infinite retry)
- [ ] CI green

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceLoginConfig.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceLoginConfig.java
@@ -270,7 +270,7 @@ public class SalesforceLoginConfig {
     public String toString() {
         return "SalesforceLoginConfig[" + "instanceUrl= '" + instanceUrl + "', loginUrl='" + loginUrl + '\'' + ","
                + "clientId='" + clientId + '\'' + ", clientSecret='********'"
-               + ", refreshToken='" + refreshToken + '\'' + ", userName='" + userName + '\'' + ", password=********'"
+               + ", refreshToken='********'" + ", userName='" + userName + '\'' + ", password=********'"
                + ", keystore=********', audience='" + jwtAudience + '\'' + ","
                + ", lazyLogin=" + lazyLogin + ']';
     }
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/AbstractClientBase.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/AbstractClientBase.java
@@ -190,6 +190,11 @@ public abstract class AbstractClientBase extends ServiceSupport
             final List<ByteBuffer> buffers = new ArrayList<>();
             while (true) {
                 Content.Chunk chunk = inputStreamRequestContent.read();
+                if (Content.Chunk.isFailure(chunk)) {
+                    chunk.release();
+                    throw new RuntimeException(
+                            new IOException("Failed to buffer request content", chunk.getFailure()));
+                }
                 if (chunk.isLast()) {
                     break;
                 } else {
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/DefaultRawClient.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/client/DefaultRawClient.java
@@ -48,8 +48,8 @@ public class DefaultRawClient extends AbstractClientBase implements RawClient {
     @Override
     protected void setAccessToken(Request request) {
         // replace old token
-        request.headers(h -> h.add(BULK_TOKEN_HEADER, accessToken));
-        request.headers(h -> h.add(REST_TOKEN_HEADER, TOKEN_PREFIX + accessToken));
+        request.headers(h -> h.put(BULK_TOKEN_HEADER, accessToken));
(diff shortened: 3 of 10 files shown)


gnodet (approved):

**Excellent work** — thorough, well-organized fix for 11 distinct medium-severity bugs in camel-salesforce. Each fix is narrow, clearly motivated, and well-documented in the PR description.

### Changes verified:

**Auth / session / HTTP client:**

1. **`SalesforceLoginConfig.toString()` — refreshToken masking** ✅
   `refreshToken` was the only secret printed in cleartext while `clientSecret`, `password`, and `keystore` were already masked since CAMEL-15425. Trivially correct.

2. **`AbstractClientBase` — failure chunk detection** ✅
   Follows the Jetty `Content.Chunk` API contract: `isFailure()` must be checked before `isLast()`. Without this, a mid-read I/O error silently truncated the request body. Properly releases the chunk before throwing.

3. **`DefaultRawClient.setAccessToken()` — `put()` vs `add()`** ✅
   Confirmed `add()` in the current `main` branch — would produce duplicate `X-SFDC-Session`/`Authorization` headers on auth retry. `put()` correctly replaces.

4. **`SalesforceSecurityHandler` — coordinated re-login via worker pool** ✅
   Moves from per-401 inline login to `session.login()` on the worker pool for single-flight coordination. Also checks `body.rewind()` return value — appropriate warning-only approach since bodyless requests (GET/DELETE) would still succeed.

**Streaming / Pub/Sub:**

5. **`PubSubApiClient` — per-observer replay state** ✅
   Moving `initialReplayPreset`, `initialReplayId`, `fallbackToLatestReplayId` from class-level to `FetchResponseObserver` instance fields prevents cross-consumer preset overwrite when multiple consumers share one `PubSubApiClient`. Clean solution.

6. **`PubSubApiClient` — stop infinite retry on corrupt replay ID** ✅
   When `fallbackToLatestReplayId=false`, the `return` after setting `replayId = null` correctly stops the retry loop. The exception handler is already invoked to report the error.

7. **`PubSubApiClient` — per-event decode failure handling** ✅
   Previously only logged errors — events were silently lost (replay ID advanced past them). Now routes to `consumer.getExceptionHandler()`.

8. **`SubscriptionHelper` — `getOrDefault()` and `channelsToSubscribe` cleanup** ✅
   Prevents NPE from concurrent unsubscribe race, and removes stale entries from `channelsToSubscribe` to prevent silent channel loss on reconnect.

**REST / Bulk / Composite:**

9. **`JsonRestProcessor` — `RESPONSE_CLASS_DEFERRED`/`RESPONSE_CLASS_PREFIX` property cleanup** ✅
   `RESPONSE_CLASS` was already cleaned up but these two were leaked into subsequent producers on the same exchange.

10. **`CompositeApiProcessor` — exception preservation** ✅
    Critical fix — three response methods all had the same bug: exception was silently swallowed. All three now set `exchange.setException(exception)` when non-null.

11. **`AbstractRestProcessor` — `deleteSobjectWithId` fixes** ✅
    - **Ordering**: confirmed all sibling methods (`getSobjectWithId`, `upsertSobject`) call `restoreFields` *before* `processResponse` — this was the only one with reversed order.
    - **Null-check**: `oldValue.toString()` NPE risk confirmed in `main` branch. Correct fallback to `getParameter()`.

### Test update:

`PubSubApiTest` correctly reflects the behavioral change — expects 1 subscribe call and 1 exception handler call (was 3 of each). `Thread.sleep(1000)` removal is a bonus.

No issues found.

PR #24859, opened 2026-07-17
CAMEL-24153: camel-aws2-s3 - Fix streaming upload truncation and orphaned multipart uploads

davsclaus:
## Summary

Fixes three bugs in `AWS2S3StreamUploadProducer` that cause data loss and orphaned multipart uploads when `multiPartUpload=true` and the body is larger than `partSize`:

- **`maxRead` not reset after batch-flush**: after a batch completion, `maxRead` stayed at 0, so `toByteArray(is, 0)` read nothing and the loop exited with remaining data unread. Added `maxRead = partSize` reset (guarded by `isMultiPartUpload()`) in both `processWithoutTimestampGrouping` and `processWithTimestampGrouping`.
- **`createMultipartUpload` called on every iteration**: CAMEL-20728 changed `state.index++` to `state.index = 1`, breaking the `if (state.index == 1)` guard so `createMultipartUpload` fired on every loop iteration. Each call returned a new `uploadId`, orphaning parts from earlier iterations and accruing storage charges. Changed guard to `if (state.initResponse == null)`.
- **`maxRead` going negative**: when `uploadAggregate.buffer.size() > partSize`, subtracting it from `maxRead` drove `maxRead` negative, silently dropping the next exchange body. Added `Math.max(1, maxRead)` clamp.

## Test plan

- [x] Added `AWS2S3StreamUploadMultipartTest` with 3 unit tests (mocked S3Client):
  - Body larger than `partSize` is not truncated
  - `createMultipartUpload` called exactly once per upload
  - Multiple parts use sequential part numbers and the same upload ID
- [x] All existing unit tests pass (18/18)
- [x] All existing integration tests pass (including `S3StreamUploadMultipartIT` and `S3StreamUploadMultipartAsyncIT`)

### DIFF
--- a/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/stream/AWS2S3StreamUploadProducer.java
+++ b/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/stream/AWS2S3StreamUploadProducer.java
@@ -181,6 +181,7 @@ public class AWS2S3StreamUploadProducer extends DefaultProducer {
         if (uploadAggregate != null) {
             uploadAggregate.index++;
             maxRead -= uploadAggregate.buffer.size();
+            maxRead = Math.max(1, maxRead);
         }
 
         while ((b = AWS2S3Utils.toByteArray(is, maxRead)) != null && b.length > 0) {
@@ -206,6 +207,9 @@ public class AWS2S3StreamUploadProducer extends DefaultProducer {
                             uploadPart(uploadAggregate);
                         CompleteMultipartUploadResponse uploadResult = completeUpload(uploadAggregate);
                         this.uploadAggregate = null;
+                        if (getConfiguration().isMultiPartUpload()) {
+                            maxRead = Math.toIntExact(getConfiguration().getPartSize());
+                        }
                         Message message = getMessageForResponse(exchange);
                         message.setHeader(AWS2S3Constants.E_TAG, uploadResult.eTag());
                         if (uploadResult.versionId() != null) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**Solid fix for three distinct data-loss bugs in `AWS2S3StreamUploadProducer`** — each well-analyzed with clear root cause.

### Changes verified:

1. **`maxRead` not reset after batch-flush** ✅
   After `completeUpload()`, `maxRead` stayed at 0 (or whatever it was post-subtraction), so `toByteArray(is, 0)` read nothing on the next exchange. Reset to `partSize` is correctly applied in all 4 completion sites (2 in `processWithoutTimestampGrouping`, 2 in `processWithTimestampGrouping`), guarded by `isMultiPartUpload()`.

2. **`createMultipartUpload` called on every iteration** ✅
   CAMEL-20728's `state.index = 1` broke the `if (state.index == 1)` guard — it was always true. `if (state.initResponse == null)` is a better invariant since it's impossible to regress. Consistently applied to all 3 guarded blocks (random naming, timestamp naming, and the actual `createMultipartUpload` call).

3. **`maxRead` going negative** ✅
   `Math.max(1, maxRead)` is the right clamp — using 1 instead of 0 ensures `toByteArray` still attempts a read, keeping the loop alive. Applied in both `processWithoutTimestampGrouping` and `processWithTimestampGrouping`.

### Tests:

All 3 tests are well-designed with mocked S3Client:
- `testBodyLargerThanPartSizeIsNotTruncated` — reassembles all uploaded parts and verifies byte-for-byte match
- `testCreateMultipartUploadCalledOncePerUpload` — `verify(times(1))` catches the orphan-upload regression
- `testMultiplePartsUploadedCorrectly` — verifies sequential part numbers and consistent `uploadId`

No issues found.

PR #24853, opened 2026-07-17
(chores): fix SonarCloud S1845 naming case issues

gnodet:
## Summary

_Claude Code on behalf of Guillaume Nodet_

Fix two genuine SonarCloud S1845 (identifier name differs only by case) issues:

- **PubSubApiProcessor**: Make `LOG` field `static final` instead of non-static instance field. This eliminates the clash with the inherited `log` field from the superclass and follows the standard Camel logging convention.
- **PGPDataFormatUtil**: Fix method name typo `findPrivateKeyWithkeyId` → `findPrivateKeyWithKeyId` (uppercase 'K'). The old method is deprecated for backwards compatibility and delegates to the correctly named one. Internal caller `DefaultPGPSecretKeyAccessor` updated to use the new name.

### SonarCloud context

This PR is part of the BLOCKER remediation effort. Of the 46 S1845 issues, 41 were triaged as won't-fix (intentional patterns: inherited logger shadowing, field vs constant naming, deprecated API migration, method overloads with different signatures). The remaining 5 included these 2 genuine fixes and 3 test-infra `username`/`userName` deprecation transitions (marked as won't-fix since the naming is intentional during API migration).

## Test plan
- [x] `mvn test -pl components/camel-crypto-pgp` — all 69 tests pass (22 skipped)
- [ ] CI full build

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24860, opened 2026-07-17
(chores): fix SonarCloud S1845 naming convention issues in salesforce and crypto-pgp

gnodet:
## Summary

Fix 2 remaining SonarCloud S1845 (BLOCKER) issues where class members differ only by capitalization.

### Changes

**1. `PubSubApiProcessor` — remove shadowing `LOG` field**
- Removed redundant `private static final Logger LOG` field that shadowed the inherited `log` field from `AbstractSalesforceProcessor`
- The class already uses `log.debug(...)` (inherited), so the `LOG` field was unused
- Removed now-unused `Logger`/`LoggerFactory` imports

**2. `PGPDataFormatUtil` — remove deprecated typo-named method wrapper**
- PR #24853 fixed the `findPrivateKeyWithkeyId` typo by adding a correctly-named `findPrivateKeyWithKeyId` method and keeping the old name as a `@Deprecated` wrapper
- However, having both `findPrivateKeyWithkeyId` and `findPrivateKeyWithKeyId` in the same class still triggers S1845
- Since 4.22 has not been released yet, this PR removes the deprecated wrapper entirely (no backward compatibility concern)
- All callers already use the correctly-named `findPrivateKeyWithKeyId`

### SonarCloud Rule
- **S1845**: Members and types of a class or interface should not have the same name differing only by capitalization

### DIFF
--- a/components/camel-crypto-pgp/src/main/java/org/apache/camel/converter/crypto/PGPDataFormatUtil.java
+++ b/components/camel-crypto-pgp/src/main/java/org/apache/camel/converter/crypto/PGPDataFormatUtil.java
@@ -102,19 +102,6 @@ public final class PGPDataFormatUtil {
         return findPrivateKeyWithKeyId(keyid, passphrase, passphraseAccessor, provider, pgpSec);
     }
 
-    /**
-     * @deprecated Use
-     *             {@link #findPrivateKeyWithKeyId(long, String, PGPPassphraseAccessor, String, PGPSecretKeyRingCollection)}
-     *             instead (note the uppercase 'K' in 'KeyId').
-     */
-    @Deprecated(since = "4.22")
-    public static PGPPrivateKey findPrivateKeyWithkeyId(
-            long keyid, String passphrase, PGPPassphraseAccessor passphraseAccessor,
-            String provider, PGPSecretKeyRingCollection pgpSec)
-            throws PGPException {
-        return findPrivateKeyWithKeyId(keyid, passphrase, passphraseAccessor, provider, pgpSec);
-    }
-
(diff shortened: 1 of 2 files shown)


gnodet (commented):

**Clean S1845 fixes** — two changes reviewed:

### 1. PubSubApiProcessor — remove shadowing `LOG` field ✅

Removing the `private Logger LOG` and using the inherited `log` from `AbstractSalesforceProcessor` is the correct fix. The `LOG` vs `log` name collision was the S1845 trigger.

### 2. PGP `findPrivateKeyWithkeyId` → `findPrivateKeyWithKeyId` — heads up on overlap ⚠️

This PR does a **direct rename** (no deprecated wrapper). Note that **PR #24853** touches the same code but takes a more conservative approach: it keeps the old method as a `@Deprecated(since = "4.22")` wrapper for backward compatibility.

Since `findPrivateKeyWithkeyId` is a **`public static`** method, the direct rename in this PR is a public API break. If downstream code calls this method by name, it would fail at compile time after upgrading.

**Both PRs also touch overlapping code** in PubSubApiProcessor (different approaches) and PGP — they'll conflict when one merges. Worth coordinating which approach to take:
- **#24853**: deprecated wrapper (backward-compatible) + `LOG` made `static final`
- **#24860**: direct rename (breaking) + `LOG` removed entirely (cleaner)

For PubSubApiProcessor, this PR's approach (removing LOG, using inherited `log`) is the better fix. For PGP, the deprecated wrapper from #24853 is safer unless the project is OK with the public API break.

davsclaus (approved):

PR #24865, opened 2026-07-17
CAMEL-24159: camel-azure-servicebus - fix medium-severity bugs from code review

davsclaus:
## Summary

Fixes 5 medium-severity findings in camel-azure-servicebus from the July 2026 code review ([CAMEL-24159](https://issues.apache.org/jira/browse/CAMEL-24159)). This is PR 1 of 5 (one per azure component).

- **Credential override guard** — autowired `TokenCredential` no longer silently overrides an explicitly configured `credentialType` (regression from CAMEL-21871)
- **Client leak on route restart** — internally-created `ServiceBusProcessorClient` is now properly closed in `doStop()`, preventing AMQP connection leaks on stop/start cycles
- **User-provided client lifecycle** — user-provided (autowired) sender/processor clients are no longer closed by the component; only internally-created clients are managed
- **CloudEvent content type** — fixed `containsKey()` (returns Boolean) to `getHeader()` (returns the actual content type string) in the CloudEvent transformer (bug since CAMEL-20450)
- **APPLICATION_PROPERTIES mutation** — producer now defensive-copies the user-supplied `APPLICATION_PROPERTIES` map before adding propagated headers, preventing `UnsupportedOperationException` on immutable maps

The 6th finding (maxConcurrentCalls backpressure for async routes) is a design limitation and will be addressed in a separate follow-up issue.

## Test plan

- [x] Updated existing lifecycle tests to match new close/stop behavior
- [x] Added `doStopStopsUserProvidedClientWithoutClosing` test
- [x] Added `doShutdownDoesNotCloseUserProvidedClient` test
- [x] `mvn clean install` passes (72 tests, 0 failures)

### DIFF
--- a/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusComponent.java
+++ b/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusComponent.java
@@ -62,9 +62,11 @@ public class ServiceBusComponent extends DefaultComponent {
                 endpoint.getConfiguration().setCredentialType(CredentialType.CONNECTION_STRING);
             }
         } else {
-            boolean azure = endpoint.getConfiguration().getTokenCredential() instanceof DefaultAzureCredential;
-            endpoint.getConfiguration()
-                    .setCredentialType(azure ? CredentialType.AZURE_IDENTITY : CredentialType.TOKEN_CREDENTIAL);
+            if (endpoint.getConfiguration().getCredentialType() == null) {
+                boolean azure = endpoint.getConfiguration().getTokenCredential() instanceof DefaultAzureCredential;
+                endpoint.getConfiguration()
+                        .setCredentialType(azure ? CredentialType.AZURE_IDENTITY : CredentialType.TOKEN_CREDENTIAL);
+            }
         }
 
         return endpoint;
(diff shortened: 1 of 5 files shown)


gnodet (approved):

Five clean, well-scoped fixes. Each one addresses a real bug with proper tests.

### 1. Credential override guard (ServiceBusComponent.java)
Correct fix. The `if (getCredentialType() == null)` guard ensures an autowired `TokenCredential` only sets `credentialType` when the user hasn't already configured it explicitly. Without this, an autowired credential silently overrides the user's chosen type — a regression from CAMEL-21871.

### 2. Client leak on route restart (ServiceBusConsumer.java — doStart)
Good catch. On route restart (stop→start), the old `ServiceBusProcessorClient` was never closed, leaking its AMQP connection. The new `doStart()` now closes any previously created client before creating a new one. The `clientCloseable && client != null` guard correctly avoids double-close since `doStop()` already nulls the field.

### 3. User-provided client lifecycle (Consumer + Producer — clientCloseable flag)
The `clientCloseable` flag cleanly separates ownership semantics:
- Internally created → component owns it → close on stop
- User-provided → user owns it → just stop (consumer) / don't close (producer)

This follows the standard Camel pattern for autowired resources. Tests are comprehensive — both `doStopStopsUserProvidedClientWithoutClosing` and `doShutdownDoesNotCloseUserProvidedClient` verify the user-provided path, while the renamed tests verify the internally-created path.

### 4. CloudEvent content type (ServicebusCloudEventDataTypeTransformer.java)
Classic bug — `containsKey()` returns `boolean`, not the value. The `if` above already checks `containsKey()`, so the nested call was always going to put `true` as the content type header. `getHeader()` is the correct fix. Bug since CAMEL-20450.

### 5. APPLICATION_PROPERTIES defensive copy (ServiceBusProducer.java)
Correct. Both `sendMessages()` and `scheduleMessages()` now defensive-copy the user's map before `propagateHeaders()` mutates it. Prevents `UnsupportedOperationException` on `Map.of()` / `Collections.unmodifiableMap()` and avoids surprising side effects on the caller's map.

LGTM ✅

PR #24866, opened 2026-07-17
[backport camel-4.18.x] CAMEL-24193: camel-aws-secrets-manager - fix SQS-notification secret refresh NPE and queue draining

oscerd:
Backport of #24847 to `camel-4.18.x`. Fixes [CAMEL-24193](https://issues.apache.org/jira/browse/CAMEL-24193) on the 4.18.x line: guard against a null receive result (missing queue no longer NPEs), delete every examined SQS message so the queue is drained, and reuse a single ObjectMapper. Adds `CloudTrailReloadTriggerTaskSqsTest` + test-scoped mockito.

---

### DIFF
--- a/components/camel-aws/camel-aws-secrets-manager/pom.xml
+++ b/components/camel-aws/camel-aws-secrets-manager/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>camel-test-spring-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-version}</version>
+            <scope>test</scope>
+        </dependency>
 
         <!-- test infra -->
        <dependency>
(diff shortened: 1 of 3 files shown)


gnodet (approved):

Clean backport of #24847 to camel-4.18.x. All three fixes are correct and well-tested.

### NPE guard on null receiveMessage result
When `QueueDoesNotExistException` is caught, `messageResult` stays null. The original code then fell through to `messageResult.messages()` → NPE. The `if (messageResult != null)` guard is the right fix.

### Queue draining — delete every examined message
This was the most impactful bug. Previously:
- Only the **first matching** message was deleted, then `break` exited the loop
- Non-matching messages were left in the queue, redelivered on every poll indefinitely
- Messages after the first match were never examined

The fix correctly:
- Removes the `break` so all messages in the batch are processed
- Moves the delete **outside** the match condition so every examined message is deleted (matched → reload recorded; unmatched/malformed → discarded)
- Logs a warning for malformed JSON instead of `throw new RuntimeException(e)`, allowing the loop to continue processing remaining messages

### Static ObjectMapper
`ObjectMapper` is thread-safe for read operations. Reusing a single static instance instead of creating one per message is correct and avoids unnecessary allocation.

### Test coverage
Both tests cover the key scenarios:
- `drainsEveryExaminedMessageIncludingNonMatching` — verifies 2 messages (1 match + 1 non-match) both get deleted
- `doesNotThrowOrDeleteWhenQueueDoesNotExist` — verifies the null-guard prevents NPE

LGTM ✅

davsclaus (approved):

PR #24867, opened 2026-07-17
CAMEL-24159: camel-azure-eventhubs - fix medium-severity bugs from code review

davsclaus:
## Summary

Fixes 4 medium-severity findings in camel-azure-eventhubs from the July 2026 code review ([CAMEL-24159](https://issues.apache.org/jira/browse/CAMEL-24159)). This is PR 2 of 5 (one per azure component).

- **Credential override guard** — autowired `TokenCredential` no longer silently overrides an explicitly configured `credentialType` (regression from CAMEL-21871)
- **Checkpoint task race condition** — `processCommit()` is now `synchronized` to prevent concurrent partition threads from racing on `lastTask`/`lastScheduledTask`; `eventContext` field in `EventHubsCheckpointUpdaterTask` is now `volatile` for cross-thread visibility
- **Silent checkpoint failures** — checkpoint update errors are now logged at WARN (with stack trace) instead of DEBUG, so operators get visibility when the checkpoint store breaks and unbounded replay would otherwise occur silently
- **User-provided producer client lifecycle** — user-provided (autowired) `EventHubProducerAsyncClient` is no longer closed by the component on stop; only internally-created clients are managed

## Test plan

- [x] `mvn clean install` passes in camel-azure-eventhubs module
- [x] No existing tests broken

### DIFF
--- a/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsCheckpointUpdaterTask.java
+++ b/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsCheckpointUpdaterTask.java
@@ -29,7 +29,7 @@ public class EventHubsCheckpointUpdaterTask implements Runnable {
 
     private static final Logger LOG = LoggerFactory.getLogger(EventHubsCheckpointUpdaterTask.class);
 
-    private EventContext eventContext;
+    private volatile EventContext eventContext;
     private final AtomicInteger processedEvents;
     private volatile long scheduledTime;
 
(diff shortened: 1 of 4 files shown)


gnodet (approved):

Four clean fixes, consistent with the patterns from PR #24865 (servicebus). All correct.

### 1. Credential override guard (EventHubsComponent.java)
Same fix as servicebus — `if (getCredentialType() == null)` guard prevents autowired `TokenCredential` from overriding explicit `credentialType` configuration. Correct.

### 2. Volatile eventContext + synchronized processCommit
Two related concurrency fixes:

**`volatile EventContext eventContext`** — This field is written by partition consumer threads and read by the timer thread running `EventHubsCheckpointUpdaterTask.run()`. Without `volatile`, the timer thread could see a stale reference due to JMM. Good fix.

**`synchronized processCommit()`** — Multiple partition threads can call `processCommit()` concurrently, racing on `lastTask`/`lastScheduledTask`. Without synchronization, two threads could both see `lastTask == null`, both create new tasks, and one would silently overwrite the other — losing the checkpoint schedule. The method-level `synchronized` is the simplest correct fix for this.

### 3. Checkpoint error logging (WARN + stack trace)
Changed `LOG.debug(error.getMessage())` → `LOG.warn(error.getMessage(), error)` in both the timer task and the consumer. Two improvements:
- **WARN** instead of **DEBUG** — checkpoint failures can cause unbounded event replay on restart. This is operational severity, not debug noise.
- **Stack trace** — the `error` parameter as the third SLF4J argument logs the full cause, which is essential for diagnosing checkpoint store failures (connectivity, permissions, schema mismatch).

### 4. User-provided producer client lifecycle (EventHubsProducer.java)
Same `clientCloseable` pattern as servicebus — only close the producer client if it was internally created. User-provided clients are left for the user to manage.

LGTM ✅

PR #24869, opened 2026-07-17
CAMEL-24159: camel-azure-storage-blob - Fix medium-severity bugs from code review

davsclaus:
## Summary

Fixes 5 medium-severity findings in `camel-azure-storage-blob` from CAMEL-24159 code review:

- **Credential override guard** (`BlobComponent`): wrap the autowiring branch with `if (credentialType == null)` so user-explicit credential type is not overridden when a `TokenCredential` bean is found in the registry.
- **changeFeed time bounds** (`BlobChangeFeedOperations`): change `||` to `&&` so that setting only `startTime` or only `endTime` correctly calls the time-bounded overload instead of discarding both.
- **copyBlob validation** (`BlobOperations`): fix the second `isEmpty` check — it was re-checking `sourceAccountName` instead of `sourceContainerName`.
- **ClassCastException with BlobRequestConditions** (`BlobCommonRequestOptions` + `BlobOperations`): remove the unsafe generic `(T)` cast from `getBlobRequestConditions()` that would throw ClassCastException at runtime when append/page blob operations need `AppendBlobRequestConditions` or `PageBlobRequestConditions`. Add proper type-safe conversion methods.
- **Consumer fileDir download failures** (`BlobOperations`): create parent directories before downloading (fixes `NoSuchFileException` for nested blob names like `dir/file.txt`) and delete existing file first (fixes `FileAlreadyExistsException` on re-poll).

## Test plan

- [x] Module builds successfully: `cd components/camel-azure/camel-azure-storage-blob && mvn clean install`
- [ ] Verify credential override guard preserves user-explicit `credentialType=SHARED_KEY_CREDENTIAL`
- [ ] Verify changeFeed with only startTime set returns bounded results
- [ ] Verify copyBlob validates sourceContainerName is provided
- [ ] Verify append/page blob operations with request conditions don't throw ClassCastException
- [ ] Verify consumer re-poll downloads successfully without FileAlreadyExistsException

### DIFF
--- a/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobCommonRequestOptions.java
+++ b/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobCommonRequestOptions.java
@@ -54,9 +54,8 @@ public class BlobCommonRequestOptions {
         return accessTier;
     }
 
-    @SuppressWarnings("unchecked")
-    public <T extends BlobRequestConditions> T getBlobRequestConditions() {
-        return blobRequestConditions == null ? null : (T) blobRequestConditions;
+    public BlobRequestConditions getBlobRequestConditions() {
+        return blobRequestConditions;
     }
 
     public byte[] getContentMD5() {
--- a/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobComponent.java
+++ b/components/camel-azure/camel-azure-storage-blob/src/main/java/org/apache/camel/component/azure/storage/blob/BlobComponent.java
@@ -91,7 +91,9 @@ public class BlobComponent extends HealthCheckComponent {
                     configuration.setCredentialType(AZURE_IDENTITY);
                 }
             } else {
-                configuration.setCredentialType(CredentialType.SHARED_KEY_CREDENTIAL);
+                if (configuration.getCredentialType() == null) {
+                    configuration.setCredentialType(CredentialType.SHARED_KEY_CREDENTIAL);
+                }
             }
         }
     }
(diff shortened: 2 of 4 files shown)


gnodet (approved):

Five solid fixes. Each addresses a real bug — two of them are particularly good catches.

### 1. Credential override guard (BlobComponent.java)
Consistent with the servicebus/eventhubs PRs. The `if (credentialType == null)` guard prevents the autowiring fallback from overriding a user's explicit `credentialType` setting.

### 2. changeFeed time bounds — `||` → `&&` (BlobChangeFeedOperations.java)
Nice catch. With `||`, setting **only** `startTime` (without `endTime`) would hit the unbounded branch, discarding the time bound entirely. With `&&`, the time-bounded overload is correctly used whenever at least one bound is set — the null bound becomes open-ended.

### 3. copyBlob validation — wrong variable (BlobOperations.java)
Classic copy-paste bug. The second `isEmpty` check validated `sourceAccountName` again instead of the freshly-retrieved `sourceContainerName`. This meant `copyBlob` would accept a null/empty container name and fail with an obscure Azure SDK error instead of a clear validation message.

### 4. ClassCastException — type-safe BlobRequestConditions conversion (BlobCommonRequestOptions + BlobOperations)
Well-implemented. The old generic `<T>` cast was a latent `ClassCastException` — passing a plain `BlobRequestConditions` to an append/page blob operation would fail at runtime. The new conversion methods are clean:
- Null-safe
- Pass-through for correct subtypes (`instanceof` pattern match)
- Copy all base-class properties when converting from the parent type
- Subclass-specific properties (`maxSize`, `appendPosition`, sequence number conditions) default to unset, which is correct since the source didn't have them

### 5. Consumer fileDir download — parent directories + existing file (BlobOperations.java)
Two practical fixes:
- `parentDir.mkdirs()` — nested blob names like `dir/file.txt` need parent directories created locally
- `fileToDownload.delete()` — prevents `FileAlreadyExistsException` on re-poll

Both are defensive and safe: `mkdirs()` is a no-op if directories exist, and `delete()` returns `false` silently if the file doesn't exist.

LGTM ✅

PR #24870, opened 2026-07-17
CAMEL-24159: camel-azure-storage-queue - Fix medium-severity bugs from code review

davsclaus:
## Summary

Fixes 2 medium-severity findings in `camel-azure-storage-queue` from CAMEL-24159 code review:

- **Credential override guard** (`QueueComponent`): wrap the else branch with `if (credentialType == null)` so user-explicit credential type is not overridden when a `StorageSharedKeyCredential` bean is found in the registry.
- **Idempotent queue creation** (`QueueOperations` + `QueueClientWrapper`): when `createQueue=true` on `sendMessage`, use `createIfNotExistsWithResponse()` instead of `create()` so repeated sends don't throw HTTP 409 Conflict when the queue already exists with different metadata.

## Review feedback addressed

- Fixed `createIfNotExists` wrapper to forward `metadata` and `timeout` via `createIfNotExistsWithResponse(metadata, timeout, Context.NONE)` instead of the no-arg `createIfNotExists()` that silently dropped both parameters.

## Test plan

- [x] Module builds successfully: `cd components/camel-azure/camel-azure-storage-queue && mvn clean install`
- [ ] Verify credential override guard preserves user-explicit `credentialType`
- [ ] Verify sendMessage with createQueue=true succeeds on second call when queue already exists

### DIFF
--- a/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/QueueComponent.java
+++ b/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/QueueComponent.java
@@ -73,7 +73,9 @@ public class QueueComponent extends HealthCheckComponent {
                     configuration.setCredentialType(CredentialType.SHARED_ACCOUNT_KEY);
                 }
             } else {
-                configuration.setCredentialType(CredentialType.SHARED_KEY_CREDENTIAL);
+                if (configuration.getCredentialType() == null) {
+                    configuration.setCredentialType(CredentialType.SHARED_KEY_CREDENTIAL);
+                }
             }
         }
 
--- a/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/client/QueueClientWrapper.java
+++ b/components/camel-azure/camel-azure-storage-queue/src/main/java/org/apache/camel/component/azure/storage/queue/client/QueueClientWrapper.java
@@ -43,6 +43,10 @@ public class QueueClientWrapper {
         return client.createWithResponse(metadata, timeout, Context.NONE);
     }
 
+    public void createIfNotExists(Map<String, String> metadata, Duration timeout) {
+        client.createIfNotExists();
+    }
+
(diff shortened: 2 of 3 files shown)


gnodet (commented):

The credential override guard is correct (same pattern as the other azure PRs). The `createIfNotExists` fix addresses a real usability issue.

### Credential override guard (QueueComponent.java)
Consistent with servicebus/eventhubs/blob PRs. `if (credentialType == null)` prevents overriding explicit user configuration. ✅

### Idempotent queue creation (QueueOperations + QueueClientWrapper)
The switch from `create()` (throws 409 on existing queue) to `createIfNotExists()` is the right fix for the `createQueue=true` use case — producers shouldn't fail on repeated sends.

**However, the wrapper method silently drops its parameters:**

```java
public void createIfNotExists(Map<String, String> metadata, Duration timeout) {
    client.createIfNotExists();  // ← metadata and timeout are ignored!
}
```

The calling code in `QueueOperations` extracts metadata and timeout from the exchange and passes them in, but they're never forwarded. This is a **regression** compared to the old `create(metadata, timeout)` call — users who set queue metadata via exchange headers would silently lose that metadata.

The Azure SDK's `QueueClient` has `createIfNotExistsWithResponse(Map<String, String> metadata, Duration timeout, Context context)` — the implementation should be:

```java
public void createIfNotExists(Map<String, String> metadata, Duration timeout) {
    client.createIfNotExistsWithResponse(metadata, timeout, Context.NONE);
}
```

This should be a small fix. Everything else LGTM.

davsclaus:

@gnodet Good catch — fixed in `1aacea0`. The wrapper now calls `client.createIfNotExistsWithResponse(metadata, timeout, Context.NONE)`, matching the pattern used by the existing `create()` method directly above it.

gnodet (approved):

## Re-review — commit 1aacea0

The \`createIfNotExists\` parameter-dropping bug I flagged is now fixed:

```java
// Before: ignored metadata and timeout
client.createIfNotExists();

// After: forwards both parameters
client.createIfNotExistsWithResponse(metadata, timeout, Context.NONE);
```

This was the last outstanding item. LGTM.

PR #24873, opened 2026-07-17
CAMEL-24196: camel-aws2-sqs - Fix QueueDoesNotExist swallowed, thread pool leak, and batch chunking

davsclaus:
## Summary

Fixes three bugs in camel-aws2-sqs:

- **QueueDoesNotExistException silently swallowed** (`Sqs2Consumer`): When `maxMessagesPerPoll > 10`, parallel polling sends N requests. Only the first thread to catch `QueueDoesNotExistException` fires an error (errorCount=1), but the error propagation check requires `errorCount == N`, which never matches. Fix: added explicit check for `isQueueMissing() && hasErrors()` before the all-errors check.

- **Thread pool leak on stop/start** (`Sqs2Consumer`): `doStart()` unconditionally creates a new `PollingTask` with a new thread pool, but cleanup only happened in `doShutdown()`, not `doStop()`. Each route stop/start cycle leaked the previous thread pool. Fix: added `doStop()` override that cleans up the PollingTask, TimeoutExtender, and scheduled executor.

- **sendBatchMessage missing 10-entry chunking** (`Sqs2Producer`): All entries were collected into a single `sendMessageBatch()` call. SQS hard-limits batches to 10 entries; 11+ throws `TooManyEntriesInBatchRequestException`. Fix: extracted `sendBatchEntries()` helper that chunks entries into groups of 10.

## Test plan

- [ ] Existing integration tests still pass
- [ ] Bug 1 verified by code inspection — queue-missing error now propagated regardless of parallel request count
- [ ] Bug 2 verified by code inspection — `doStop()` cleans up all resources that `doStart()` creates
- [ ] Bug 3 verified by code inspection — entries chunked into groups of 10 before SQS API call

### DIFF
--- a/components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
+++ b/components/camel-aws/camel-aws2-sqs/src/main/java/org/apache/camel/component/aws2/sqs/Sqs2Consumer.java
@@ -339,6 +339,28 @@ public class Sqs2Consumer extends ScheduledBatchPollingConsumer {
     }
 
     @Override
+    protected void doStop() throws Exception {
+        super.doStop();
+
+        if (ObjectHelper.isNotEmpty(timeoutExtender)) {
+            timeoutExtender.cancel();
+            timeoutExtender = null;
+        }
+        if (ObjectHelper.isNotEmpty(scheduledFuture)) {
+            scheduledFuture.cancel(true);
+            scheduledFuture = null;
+        }
+        if (ObjectHelper.isNotEmpty(scheduledExecutor)) {
+            getEndpoint().getCamelContext().getExecutorServiceManager().shutdown(scheduledExecutor);
+            scheduledExecutor = null;
+        }
+        if (ObjectHelper.isNotEmpty(pollingTask)) {
+            pollingTask.close();
+            pollingTask = null;
+        }
+    }
+
+    @Override
     protected void doShutdown() throws Exception {
         if (ObjectHelper.isNotEmpty(timeoutExtender)) {
             timeoutExtender.cancel();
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**CAMEL-24196: camel-aws2-sqs — three well-scoped fixes — LGTM** ✅

### 1. QueueDoesNotExistException silently swallowed ✅
Good catch. When `maxMessagesPerPoll > 10`, the parallel polling sends N requests but the existing `errorCount == numberOfRequestsPerPoll` check fails when only a subset of threads register the error (e.g., `errorCount == 1` but `N > 1`). The new `isQueueMissing() && hasErrors()` check short-circuits before that, correctly surfacing the queue-missing error regardless of how many parallel requests were sent.

### 2. Thread pool leak on stop/start ✅
Proper lifecycle fix. `doStop()` mirrors `doShutdown()` with two correct differences:
- `doStop()` calls `super.doStop()` first (top-down), `doShutdown()` calls `super.doShutdown()` last (bottom-up) — standard Camel lifecycle pattern
- `doStop()` uses graceful `shutdown()`, `doShutdown()` uses `shutdownNow()` — correct escalation semantics

Since `doStop()` nulls all references, `doShutdown()` safely skips (idempotent).

### 3. sendBatchMessage 10-entry chunking ✅
Fixes the hard SQS limit violation (`TooManyEntriesInBatchRequestException` on >10 entries). Clean extraction into `sendBatchEntries()`. The `totalFailed`/`totalSuccessful` aggregation across chunks is correct.

**Note:** `message.setBody(lastResult)` only captures the final chunk's `SendMessageBatchResponse`. For ≤10 entries (the only case that worked before), behavior is identical. For >10 entries (previously broken), callers who need per-entry results would need to rely on the aggregate headers rather than the body. This is a reasonable tradeoff for a fix that makes >10 entries work at all — just worth documenting if users ask.

Also nice cleanup: `Iterable` → `Iterable<?>` removes the raw type warning.

PR #24878, opened 2026-07-17
CAMEL-24197: camel-aws2-kinesis - Fix fixed-shardId DescribeStream on every poll and batch partition key

davsclaus:

## Summary

Fixes two medium-severity findings in the Kinesis component:

**1. Fixed-shardId consumer called DescribeStream on every poll**
- Replaced per-poll `DescribeStream` API call with a lookup against the cached shard list from `ShardMonitor`
- `ShardMonitor` uses `ListShards` (paginated, 100 TPS limit) vs `DescribeStream` (unpaginated, 10 TPS limit)
- Added pagination to `ShardMonitor.getShardList()` to support streams with >1000 shards
- Fixed misleading test name `itDoesNotMakeADescribeStreamRequestIfShardIdIsSet` — the test now actually verifies no `DescribeStream` call is made

**2. Batch producer applied single partition key to all records**
- Added new `CamelAwsKinesisPartitionKeys` header (`List<String>`) for per-record partition keys in batch `PutRecords`
- When set, each record gets its own partition key by index, allowing proper shard distribution
- Falls back to single `CamelAwsKinesisPartitionKey` header for records beyond the list size
- Backwards compatible — existing behavior unchanged when new header is absent

## Test plan

- [x] Existing consumer tests pass (fixed-shardId path, closed shard, multi-shard)
- [x] New `Kinesis2ProducerTest` verifies: single partition key behavior, per-record partition keys, fallback when list is shorter than records
- [x] Test now verifies `DescribeStream` is never called when `shardId` is configured
- [x] Module build passes: `mvn verify` in `camel-aws2-kinesis`
- [x] Upgrade guide entry added for 4.22


gnodet (commented):

## Review — PR #24878

### Fix 1: Fixed-shardId consumer uses ShardMonitor cache ✅

Well-designed optimization:

- **DescribeStream → ShardMonitor cache**: eliminates a per-poll API call (10 TPS limit) and reuses the `ShardMonitor` already running in the multi-shard path. The `ListShards` API it uses has a 100 TPS limit and is now properly paginated.
- **Pagination in `getShardList()`**: correctly follows the AWS `nextToken` convention — on subsequent pages, only `nextToken` is set (no `streamName`), which is the AWS API requirement.
- **Graceful degradation**: changed from `IllegalStateException` (crash) to `LOG.warn` + `return 0` when the shard isn't found in the cached list. This handles transient staleness during shard resharding without crashing the consumer.
- **Test improvement**: the test was renamed to `itUsesShardMonitorCacheNotDescribeStreamForFixedShardId` and now actually verifies `verify(kinesisClient, never()).describeStream(...)` — much more useful assertion.

### Fix 2: Per-record partition keys in batch producer ⚠️ One issue

The new `CamelAwsKinesisPartitionKeys` header and the per-record key assignment logic are well-implemented. The upgrade guide entry and constant metadata are good.

**Potential NPE in fallback path:**

In `sendBatchRecords()`, when `partitionKeys != null`, the `ensurePartitionKeyNotNull(partitionKey)` check is skipped:

```java
if (partitionKeys == null) {
    ensurePartitionKeyNotNull(partitionKey);
}
```

But in `createRequestBatchList()`, if the list is shorter than the records, it falls back to:

```java
key = partitionKey.toString();  // NPE if partitionKey is null
```

If a user sets `CamelAwsKinesisPartitionKeys` with 2 entries but sends 5 records, and does NOT set `CamelAwsKinesisPartitionKey`, records 3–5 will NPE on `partitionKey.toString()`.

**Suggested fix — validate the fallback key when the list is shorter:**

```java
if (partitionKeys == null) {
    ensurePartitionKeyNotNull(partitionKey);
} else if (partitionKeys.isEmpty()) {
    throw new IllegalArgumentException("CamelAwsKinesisPartitionKeys header is empty");
}
// Note: if partitionKeys.size() < record count, partitionKey must also be set.
// This is validated lazily in createRequestBatchList, but could be documented
// or caught eagerly.
```

Or add a null guard in the fallback path:

```java
} else {
    if (partitionKey == null) {
            "CamelAwsKinesisPartitionKey header is required as fallback when "
            + "CamelAwsKinesisPartitionKeys has fewer entries than records");
    }
    key = partitionKey.toString();
}
```

### Other observations

- The upgrade guide entries are clear and well-written
- Generated catalog/endpoint DSL files are consistent with the new constant
- All 3 producer tests cover the right scenarios (single key, per-record, fallback)
- A 4th test for the NPE edge case (list shorter + no single key) would be useful

Overall excellent work. The consumer fix is ready; the producer just needs the NPE guard.

davsclaus (commented):

Thanks for the thorough review @gnodet! The NPE finding was spot on.

**Fixed in 49f797d:**

1. **NPE guard in fallback path** — added `ensurePartitionKeyNotNull(partitionKey)` in the `else` branch of `createRequestBatchList()` so it throws a clear `IllegalArgumentException` instead of NPE when the partition keys list is shorter than the records and no single key is set.

2. **Empty list validation** — added `if (partitionKeys.isEmpty())` check in `sendBatchRecords()` to reject an empty `CamelAwsKinesisPartitionKeys` list early.

3. **Two new tests** — `batchWithFewerPartitionKeysAndNoFallbackThrows` and `batchWithEmptyPartitionKeysListThrows` cover both edge cases.

gnodet (approved):

## Re-review — commit 49f797d

Both NPE edge cases addressed:

1. **Empty list guard**: `partitionKeys.isEmpty()` → clear `IllegalArgumentException` ✅
2. **Fallback NPE guard**: `ensurePartitionKeyNotNull(partitionKey)` added in the fallback path before `partitionKey.toString()` ✅
3. **Two new tests** covering both cases with `assertThrows(IllegalArgumentException.class, ...)` ✅

LGTM — both fixes (DescribeStream elimination + per-record partition keys) are solid and well-tested.

PR #24883, opened 2026-07-18
CAMEL-24143: Fix medium findings in Split & Aggregate EIP processors

davsclaus:
## Summary

Fixes [CAMEL-24143](https://issues.apache.org/jira/browse/CAMEL-24143) — medium-severity findings from code review of the Split and Aggregate EIP processors.

### Splitter

- **Processed vs total items**: `SplitResult.getSuccessCount()` now uses actual processed count instead of total items, giving correct results after an abort
- **Stop/rollback semantics**: `shouldContinueOnFailure` no longer counts deliberate `.stop()` or rollback as failures
- **Error threshold accuracy**: ratio calculation now uses processed items (not index+1), avoiding off-by-one in parallel mode
- **Exception-only failures**: only unhandled exceptions are recorded as failures — error-handler-handled exceptions don't inflate the count
- **Watermark validation**: `NumberFormatException` on invalid watermark values is now caught and wrapped with a clear error message

### AggregateProcessor

- **Missing reset**: `completedByInterval` counter was not reset in `resetCounters()`
- **Null-safe stats**: `aggregateCompletionCounter` guards against null `completedBy` property
- **Optimistic lock retry safety**: scheduled retry lambda now catches exceptions and propagates them to the exchange instead of losing them

### SplitResult API

- Added `getProcessedItems()` method (with `@since 4.22` tag) to distinguish processed from total items

## Test plan

- [x] `SplitterErrorThresholdTest` — updated to reliably trigger the threshold regardless of parallel completion order
- [x] `SplitterSplitResultTest` — updated for new `SplitResult` constructor with `processedItems`

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/SplitResult.java
+++ b/core/camel-api/src/main/java/org/apache/camel/SplitResult.java
@@ -40,12 +40,14 @@ public final class SplitResult {
     }
 
     private final int totalItems;
+    private final int processedItems;
     private final int failureCount;
     private final List<Failure> failures;
     private final boolean aborted;
 
-    public SplitResult(int totalItems, int failureCount, List<Failure> failures, boolean aborted) {
+    public SplitResult(int totalItems, int processedItems, int failureCount, List<Failure> failures, boolean aborted) {
         this.totalItems = totalItems;
+        this.processedItems = processedItems;
         this.failureCount = failureCount;
         this.failures = failures != null ? Collections.unmodifiableList(failures) : Collections.emptyList();
         this.aborted = aborted;
@@ -60,10 +62,19 @@ public final class SplitResult {
     }
 
     /**
+     * The number of items that were actually processed. After an abort, this may be less than {@code getTotalItems()}.
+     *
+     * @since 4.22
+     */
+    public int getProcessedItems() {
+        return processedItems;
+    }
+
+    /**
      * The number of items that completed successfully.
      */
     public int getSuccessCount() {
-        return totalItems - failureCount;
+        return processedItems - failureCount;
     }
 
     /**
@@ -90,6 +101,7 @@ public final class SplitResult {
     @Override
     public String toString() {
         return "SplitResult[total=" + totalItems
+               + ", processed=" + processedItems
                + ", success=" + getSuccessCount()
                + ", failures=" + failureCount
                + ", aborted=" + aborted + "]";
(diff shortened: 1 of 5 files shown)


gnodet (approved):
## Review — CAMEL-24143: Fix medium findings in Split & Aggregate EIP processors

Solid set of correctness fixes for the Splitter error-threshold and `AggregateProcessor` internals. Since `SplitResult` and the error-threshold feature were both introduced in 4.22 (unreleased), the constructor signature change and semantic corrections are safe to land without backward-compat shims or upgrade-guide entries.

### Splitter — all changes verified ✅

| Fix | Verdict |
|-----|---------|
| **`processedItems` tracking via `afterSend`** | Clean approach — `afterSend` is called for every pair regardless of outcome, so the counter accurately reflects items that were actually dispatched to the processor. |
| **`shouldContinueOnFailure` refactoring** | Logically correct — `.stop()` / rollback signals are intentional control flow, not failures. Only unhandled exceptions should inflate the failure count. The guard `subExchange.getException() != null` before recording is the right predicate. |
| **Error threshold ratio using `processedItems`** | Fixes the off-by-one in parallel mode where `index + 1` doesn't correspond to how many items have completed. The `processed > 0` guard also prevents division-by-zero. |
| **Watermark validation** | Good defensive fix — a corrupted or non-numeric watermark now gives a clear error message instead of a bare `NumberFormatException`. |

### AggregateProcessor — three clear bug fixes ✅

1. **Missing `completedByInterval.set(0)` in `resetCounters()`** — confirmed by reading the current code: all other `completedBy*` counters are reset except this one. Clear omission.
2. **Null-safe `completedBy` in `aggregateCompletionCounter`** — prevents NPE when the property is absent.
3. **Exception propagation in optimistic lock retry** — the scheduled lambda previously let exceptions escape silently (the `ScheduledExecutorService` swallows them). Wrapping with try-catch and routing to `exchange.setException()` + `callback.done(false)` ensures failures surface properly.

### SplitResult API ✅

- `getProcessedItems()` has proper `@since 4.22` Javadoc tag
- `getSuccessCount()` correctly returns `processedItems - failureCount` (not `totalItems - failureCount`)
- Constructor change is safe — `SplitResult` is `@since 4.22` (unreleased)

### Tests ✅

- `SplitterErrorThresholdTest` — switching to 3 failures out of 5 items (60%) reliably exceeds the 50% threshold regardless of parallel completion order. The previous 2/5 (40%) was below threshold with the corrected ratio calculation.
- `SplitterSplitResultTest` — updated for the new constructor signature.

### CI
Builds still running — no results yet.

### Checklist
- [x] Tests updated for changed behavior
- [x] `@since 4.22` on new public API method
- [x] No upgrade guide needed (all changes are to unreleased 4.22 features)
- [x] Thread-safe: `AtomicInteger` for `processedItems`, `CopyOnWriteArrayList` for failures
- [x] No new dependencies

---

PR #24885, opened 2026-07-18
CAMEL-24160: camel-salesforce - Fix high-severity bugs from code review

davsclaus:
## Summary

Fixes 9 bugs found during a deep code review of camel-salesforce:

- **JsonRestProcessor**: fix infinite loop on non-object JSON in deferred SObject type detection (`while` loop never terminates on arrays/primitives), and fix `RuntimeException` escaping `IOException` catch block causing silent data corruption
- **DefaultBulkApiV2Client**: add missing `return` after error callback in `changeJobState`/`changeQueryJobState` to prevent double callback execution
- **SalesforceEndpointConfig**: add `allOrNone` to `toValueMap()` so the URI option actually takes effect, and fix `REPLAY_PRESET` mapped to wrong field (`initialReplayIdMap` instead of `replayPreset`)
- **PubSubApiClient**: skip binary trailer keys in `onError` to prevent `IllegalArgumentException` killing reconnection, add stop/channel checks in `resubscribeOnError` to prevent infinite loop on shutdown, cap reconnect delay at `maxBackoff`, and restore thread interrupt flag
- **SubscriptionHelper**: add backoff delay to handshake retry to prevent tight retry loop hammering the Salesforce API
- **SalesforceSecurityHandler**: guard against null `client` in 401 handling to prevent NPE for CometD streaming requests
- **SalesforceSession**: make latch `volatile` and remove dead spin loop to fix race condition in `attemptLoginUntilSuccessful`

## Test plan

- [ ] Verify `mvn install` passes for `camel-salesforce-component`
- [ ] Review each fix individually for correctness
- [ ] Verify no API/behavioral regressions

### DIFF
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceEndpointConfig.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/SalesforceEndpointConfig.java
@@ -834,6 +834,7 @@ public class SalesforceEndpointConfig implements Cloneable {
         valueMap.put(APEX_URL, apexUrl);
         // apexQueryParams are handled explicitly in AbstractRestProcessor
         valueMap.put(COMPOSITE_METHOD, compositeMethod);
+        valueMap.put(ALL_OR_NONE, allOrNone);
         valueMap.put(LIMIT, limit);
         valueMap.put(APPROVAL, approval);
         valueMap.put(EVENT_NAME, eventName);
@@ -869,7 +870,7 @@ public class SalesforceEndpointConfig implements Cloneable {
         valueMap.put(INITIAL_REPLAY_ID_MAP, initialReplayIdMap);
 
         // add Pub/Sub API properties
-        valueMap.put(REPLAY_PRESET, initialReplayIdMap);
+        valueMap.put(REPLAY_PRESET, replayPreset);
         valueMap.put(PUB_SUB_DESERIALIZE_TYPE, pubSubDeserializeType);
         valueMap.put(PUB_SUB_POJO_CLASS, pubSubPojoClass);
 
--- a/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/SalesforceSession.java
+++ b/components/camel-salesforce/camel-salesforce-component/src/main/java/org/apache/camel/component/salesforce/internal/SalesforceSession.java
@@ -91,7 +91,7 @@ public class SalesforceSession extends ServiceSupport {
 
     private final CamelContext camelContext;
     private final AtomicBoolean loggingIn = new AtomicBoolean();
-    private CountDownLatch latch = new CountDownLatch(1);
+    private volatile CountDownLatch latch = new CountDownLatch(1);
 
     public SalesforceSession(CamelContext camelContext, SalesforceHttpClient httpClient, long timeout,
                              SalesforceLoginConfig config) {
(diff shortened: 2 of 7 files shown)


gnodet (approved):
## Review — CAMEL-24160: camel-salesforce - Fix high-severity bugs from code review

Strong set of 9 correctness fixes spanning the Salesforce component's REST, Bulk API, PubSub, Streaming, and session layers. Each fix addresses a genuine bug. Verified git history — none conflict with prior intentional design decisions.

### Fix-by-fix analysis

#### JsonRestProcessor: Infinite loop + exception fix ✅ (HIGH)

The `while (parser.nextToken() != JsonToken.END_OBJECT)` loop is an infinite-loop bug on non-object JSON — `nextToken()` returns `null` at EOF, and `null != JsonToken.END_OBJECT` is always `true`. The fix checks both `null` and `END_OBJECT`. The null-check for `type` after the loop prevents NPE on JSON without a `"type"` field. Wrapping `SalesforceException` in `IOException` instead of `RuntimeException` preserves the checked exception contract.

#### DefaultBulkApiV2Client: Missing `return` after error callback ✅ (HIGH)

Classic double-callback bug in both `changeJobState` and `changeQueryJobState`. Without the `return`, the error callback fires, then the code falls through to deserialize a null/error response and calls the callback a second time with potentially corrupt data.

#### PubSubApiClient: Binary trailer + reconnect fixes ✅ (HIGH/MEDIUM)

- **Binary trailer keys**: Using `Metadata.ASCII_STRING_MARSHALLER` on a binary-suffixed key throws `IllegalArgumentException`, which would kill the `onError` handler and prevent reconnection. Skipping `-bin` keys is correct.
- **Reconnect delay cap**: `Math.min(reconnectDelay + backoffIncrement, maxBackoff)` prevents unbounded delay growth.
- **Stop/channel guards**: Checking `isStoppingOrStopped()` and `channel.isShutdown()` both before and after the sleep prevents infinite reconnection loops during shutdown.
- **Interrupt flag**: `Thread.currentThread().interrupt()` instead of `throw new RuntimeException(e)` — correct pattern.

#### SalesforceEndpointConfig: `allOrNone` + `REPLAY_PRESET` ✅ (MEDIUM)

- `allOrNone` was silently ignored because it was missing from `toValueMap()` — clear omission
- `REPLAY_PRESET` was mapped to `initialReplayIdMap` instead of `replayPreset` — copy-paste error that would make PubSub API replay preset configuration non-functional

#### SubscriptionHelper: Handshake retry backoff ✅ (MEDIUM)

The previous `client.handshake()` with no delay would tight-loop against the Salesforce API on repeated failures. Using `Tasks.foregroundTask()` with `Budgets.iterationBudget()` is the idiomatic Camel pattern for delayed execution (matches prior migration in CAMEL-22949). The `maxBackoff` cap and abort logic prevent infinite retry.

#### SalesforceSecurityHandler: Null-safe `client` ✅ (MEDIUM)

CometD streaming requests go through this handler but don't have a REST `client`. The `if (client != null)` guard in both the 401 handling and `createSalesforceException` prevents NPE on the streaming path.

#### SalesforceSession: `volatile` latch + dead spin loop ✅ (LOW)

The `latch` field is written by the login thread and read by waiting threads — `volatile` ensures visibility. The `while (latch == null)` spin loop was dead code since `latch` is initialized in the field declaration and never set to null. The `TODO: This is janky` comment confirms this was a known issue.

### Checklist

- [x] All 9 fixes address genuine bugs
- [x] No public API changes
- [x] No new dependencies
- [x] Git history reviewed — no conflicts with prior intentional changes
- [x] No upgrade guide needed (all fixes correct existing broken behavior)
- [ ] Tests: No new tests (understandable — these internal classes interact with Salesforce APIs, requiring extensive mocking)

### CI
Builds still pending.

---

PR #24889, opened 2026-07-18
CAMEL-23960: camel-openai - Fail fast when outputClass cannot be resolved

davsclaus:
## Summary

- **Fail-fast on unresolvable `outputClass`**: Previously, if the `outputClass` option or `CamelOpenAIOutputClass` header contained an unresolvable class name, it was silently ignored and no structured output format was applied. Now uses `resolveMandatoryClass` which throws `ClassNotFoundException`.
- **Resolve endpoint-level `outputClass` once at startup** via `doStart()` for efficiency; header-level overrides are still resolved dynamically per exchange.
- **Fix header metadata**: Changed `javaType` from `Class` to `String` for `CamelOpenAIOutputClass` since the header value is a FQCN string, not a `Class` object.

## Test plan

- [x] New `OpenAIOutputClassTest` with two tests:
  - Unknown class via header fails the exchange with `ClassNotFoundException`
  - Unknown class on endpoint fails at startup with `FailedToStartRouteException` wrapping `ClassNotFoundException`
- [x] All 92 existing tests pass (no regressions)


gnodet (approved):

## Review: APPROVE ✅

Well-structured change that replaces silent failure with fail-fast behavior for unresolvable \`outputClass\` in the OpenAI component.

### Changes reviewed

1. **\`OpenAIProducer.doStart()\`** — Resolves endpoint-level \`outputClass\` once at startup via \`resolveMandatoryClass()\`. This is the right approach: endpoint config is immutable per producer instance, so resolving it eagerly is both efficient and provides early feedback (startup fails immediately with \`ClassNotFoundException\` instead of silently producing unstructured output at runtime).

2. **\`OpenAIProducer.processInternal()\`** — Header-level override is correctly resolved per-exchange via \`resolveMandatoryClass()\`, falling back to the pre-resolved \`outputClassResolved\` when no header override is present. The two-tier resolution (header → cached config) is clean.

3. **\`OpenAIConstants.OUTPUT_CLASS\` metadata** — \`javaType\` corrected from \`Class\` to \`String\` since the header value is a FQCN string, not a \`Class\` object. Description updated accordingly.

4. **\`OpenAIOutputClassTest\`** — Good test coverage with two paths:
   - Header with unknown class → exchange fails with \`ClassNotFoundException\`
   - Endpoint config with unknown class → startup fails with \`FailedToStartRouteException\` wrapping \`ClassNotFoundException\`
   - Uses AssertJ assertions ✓, mock infrastructure ✓

5. **Generated files** — Catalog JSON, component JSON, and endpoint DSL builder properly regenerated.

### Previous behavior vs new

| Scenario | Before | After |
|---|---|---|
| Endpoint \`outputClass\` unresolvable | Silently ignored, no structured output | Fails at startup with \`ClassNotFoundException\` |
| Header \`OUTPUT_CLASS\` unresolvable | Silently ignored | Exchange fails with \`ClassNotFoundException\` |

The fail-fast behavior is strictly better — silently dropping structured output formatting is a confusing silent failure mode.

---
🤖 This review was generated by an AI agent (Claude Code) on behalf of gnodet. AI-generated reviews are advisory — human judgment should guide final decisions.

PR #24891, opened 2026-07-18
CAMEL-23958: camel-openai - Skip duplicate MCP tool names from advertised tool list

davsclaus:
## Summary

- When multiple MCP servers expose tools with the same name, only the first-registered tool is now added to the advertised function list sent to the OpenAI API. Previously, duplicates were excluded from the dispatch map (via `putIfAbsent`) but still included in the tool list, causing API rejections or silent cross-server misdispatch.
- Same fix applied to the `doReconnectMcpServer` path which had the same defect.

## Test plan

- [x] All 90 existing tests pass (no regressions)
- [x] Existing `OpenAIEndpointMcpReconnectConcurrencyTest` covers the reconnect path

### DIFF
--- a/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
+++ b/components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIEndpoint.java
@@ -193,18 +193,17 @@ public class OpenAIEndpoint extends DefaultEndpoint {
             McpSchema.ListToolsResult toolsResult = mcpClient.listTools();
             List<McpSchema.Tool> serverTools = toolsResult.tools();
 
-            tools.addAll(McpToolConverter.convert(serverTools));
-
             for (McpSchema.Tool tool : serverTools) {
                 if (toolClientMap.putIfAbsent(tool.name(), mcpClient) != null) {
                     LOG.warn("Duplicate MCP tool name '{}' from server '{}', using first registered", tool.name(),
                             serverName);
                 } else {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
**Review: CAMEL-23958 — Skip duplicate MCP tool names from advertised tool list**

Well-targeted fix for a consistency bug in how duplicate MCP tool names were handled.

**The bug:** When multiple MCP servers advertised tools with the same name, `toolClientMap.putIfAbsent()` correctly prevented duplicate dispatch entries, but the advertised function list (`tools`) and `returnDirectTools` were populated unconditionally — duplicates leaked into the OpenAI API request, causing rejections or silent misdispatch.

**What the fix does:**

1. **`initializeMcpServers()`**: Moves `tools.addAll(McpToolConverter.convert(...))` and `returnDirectTools.add()` inside the `else` block of the `putIfAbsent` check, so only successfully-registered (non-duplicate) tools are advertised and marked return-direct.

2. **`doReconnectMcpServer()`**: Same treatment — adds an explicit `containsKey` guard before registering the reconnected server's tools. Since the old server's entries are properly cleaned from all maps (lines 377–380) before re-registration, `containsKey` correctly detects cross-server duplicates only.

Both paths now consistently: register → advertise → mark return-direct as a single atomic action per tool.

The per-tool `McpToolConverter.convert(List.of(tool))` is slightly less batch-efficient than the prior bulk conversion, but MCP tool counts are small enough that this is negligible.

Clean, minimal, correct. LGTM ✅

---
_AI-generated review — Claude Code on behalf of gnodet._
_Review based on code analysis only; tests were not executed._

PR #24899, opened 2026-07-18
CAMEL-24185: Circuit Breaker EIP - suspend/resume should preserve breaker state

davsclaus:

## Summary

Neither `FaultToleranceProcessor` nor `ResilienceProcessor` override `doSuspend()`/`doResume()`, so the circuit breaker state can be lost during route suspend/resume. This PR adds explicit no-op overrides to preserve breaker state across the suspend/resume lifecycle:

- **FaultToleranceProcessor**: `doStop()` calls `CircuitBreakerMaintenance.get().reset(id)` which resets the breaker to CLOSED. The new `doSuspend()`/`doResume()` overrides skip this cleanup.
- **ResilienceProcessor**: `doStop()` calls `circuitBreakerRegistry.remove(id)` which removes the breaker entirely. The new `doSuspend()`/`doResume()` overrides skip this cleanup.

Both processors also implement the `Suspendable` marker interface so that `ServiceHelper.suspendService()` correctly routes to the suspend path instead of falling through to `stopService()`.

Stop/start resetting state remains unchanged (full lifecycle reset). Suspend/resume is lightweight and temporary — breaker state survives it.

Related: CAMEL-24136 (umbrella for circuit breaker review findings, finding M5).

## Test plan

- [x] New `ResilienceSuspendResumeTest` - verifies circuit breaker state (OPEN) is preserved across suspend/resume
- [x] New `FaultToleranceSuspendResumeTest` - verifies circuit breaker state is preserved across suspend/resume
- [x] All existing resilience4j tests pass (47 tests)
- [x] All existing fault-tolerance tests pass (25 tests)

### DIFF
--- a/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
+++ b/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
@@ -35,6 +35,7 @@ import org.apache.camel.Navigate;
 import org.apache.camel.Processor;
 import org.apache.camel.Route;
 import org.apache.camel.RuntimeExchangeException;
+import org.apache.camel.Suspendable;
 import org.apache.camel.Traceable;
 import org.apache.camel.api.management.ManagedAttribute;
 import org.apache.camel.api.management.ManagedResource;
@@ -62,7 +63,7 @@ import org.slf4j.LoggerFactory;
  */
 @ManagedResource(description = "Managed FaultTolerance Processor")
 public class FaultToleranceProcessor extends BaseProcessorSupport
-        implements CamelContextAware, Navigate<Processor>, Traceable, IdAware, RouteIdAware {
+        implements CamelContextAware, Navigate<Processor>, Traceable, IdAware, RouteIdAware, Suspendable {
 
     private static final Logger LOG = LoggerFactory.getLogger(FaultToleranceProcessor.class);
 
--- a/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
+++ b/components/camel-microprofile/camel-microprofile-fault-tolerance/src/main/java/org/apache/camel/component/microprofile/faulttolerance/FaultToleranceProcessor.java
@@ -382,6 +382,16 @@ public class FaultToleranceProcessor extends BaseProcessorSupport
     }
 
     @Override
+    protected void doSuspend() throws Exception {
+        // noop - preserve circuit breaker state across suspend/resume
+    }
+
+    @Override
+    protected void doResume() throws Exception {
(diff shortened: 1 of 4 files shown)


gnodet (approved):
## Review: CAMEL-24185 — Circuit Breaker EIP suspend/resume preserves breaker state

Clean, focused change with good tests.

### What the PR does

Adds explicit no-op `doSuspend()`/`doResume()` overrides to both `FaultToleranceProcessor` and `ResilienceProcessor` to document and protect against circuit breaker state loss during route suspend/resume:

- **FaultToleranceProcessor.doStop()** calls `CircuitBreakerMaintenance.get().reset(id)` → resets breaker to CLOSED
- **ResilienceProcessor.doStop()** calls `circuitBreakerRegistry.remove(id)` → removes the breaker entirely

The explicit no-op overrides ensure `doStop()` cleanup logic is not invoked during suspend.

### Tests are well-structured

Both tests follow the same pattern: send a message → verify breaker state → suspend → resume → assert state preserved → send another message to verify the route still works. The resilience4j test goes further by transitioning to OPEN before suspend, which is a stronger assertion (CLOSED → CLOSED could pass vacuously; OPEN → OPEN cannot).

### One observation (non-blocking)

I traced the class hierarchy: `FaultToleranceProcessor` → `BaseProcessorSupport` → `AsyncProcessorSupport` → `ServiceSupport` → `BaseService`. None of these implement `Suspendable`/`SuspendableService`. In `ServiceHelper.suspendService()`, if a service doesn't implement those interfaces, it falls through to `stopService()`:

```java
if (service instanceof Suspendable && service instanceof SuspendableService) {
    ss.suspend();
} else {
    stopService(service);  // falls through here for non-Suspendable services
}
```

The tests call `processor.suspend()` directly (which goes through `BaseService.suspend()` → `doSuspend()` — the no-op path), but if route suspension ever goes through `ServiceHelper.suspendService()`, it would still call `stopService()` → `doStop()` regardless of the `doSuspend()` override. If that's a concern, the processors might also need to implement `Suspendable` + `SuspendableService`. But this may not be how routes suspend their child processors in practice — just flagging for awareness.

LGTM ✅

---
_🤖 This review was generated by an AI agent (Claude Code) on behalf of gnodet. Opinions and technical assessments are AI-generated and should be validated by human reviewers._

davsclaus:
Good catch on the `Suspendable` marker interface — you're right that without it, `ServiceHelper.suspendService()` falls through to `stopService()`, which would reset the breaker state.

The processors already inherit `SuspendableService` via `ServiceSupport` → `StatefulService`, so they just needed the `Suspendable` marker. Added it to both `FaultToleranceProcessor` and `ResilienceProcessor` in the latest push.

gnodet (commented):
**Re-review of commit 2 (`c28e4b0`)** — _adds Suspendable marker interface_

Both `FaultToleranceProcessor` and `ResilienceProcessor` now implement the `Suspendable` marker interface. This is the correct follow-up to commit 1 — the marker tells Camel's lifecycle framework that these processors have custom suspend/resume logic (preserving breaker state), rather than falling back to the default stop/start behavior from `ServiceSupport`.

Clean, minimal change. Approval stands.

---

PR #24902, opened 2026-07-19
CAMEL-23393: Enforce agentic loop token budget and expose cumulative …

atiaomar1978-hub:
## CAMEL-23393 — Agentic loop token budget

**Issue:** https://issues.apache.org/jira/browse/CAMEL-23393  
**Branch:** `CAMEL-23393-agentic-token-budget`

### Problem
The MCP agentic loop in `OpenAIProducer.processNonStreamingAgentic()` was only capped by `maxToolIterations` (default 50). A loop could still burn through hundreds of thousands of tokens as the full conversation grew on each iteration.

### Solution

| Addition | Default | Behavior |
|----------|---------|----------|
| `maxAgenticTokens` | `0` (unlimited) | Stops the loop when cumulative `prompt + completion` tokens exceed the budget **before further tool execution** |
| `CamelOpenAIAgenticPromptTokens` | — | Cumulative prompt tokens across all agentic iterations |
| `CamelOpenAIAgenticCompletionTokens` | — | Cumulative completion tokens across all agentic iterations |
| `CamelOpenAIAgenticTotalTokens` | — | Cumulative total across all agentic iterations |

Headers are set whenever usage data is available, even if no budget is configured.

When the budget is exceeded after an API call that requests further tool execution, an `IllegalStateException` is thrown (similar to `maxToolIterations`) with the iteration number and cumulative token counts. A final text response is still returned when cumulative usage exceeds the budget, because no further spend occurs after that point. Enforcement is post-call, so actual spend may exceed the configured budget by up to one API call.

### Review follow-up (oscerd)
- Enforce budget only before continuing the tool loop, not on final answers
- Document post-call overshoot and cumulative vs per-call headers in `openai-mcp.adoc`
- Correct per-call token header types to `Long`
- Expanded tests: final answer over budget, exact boundary, multi-iteration exceed

### Tests
- `OpenAIAgenticTokenTrackerTest` — unit test for token accumulation
- `OpenAIAgenticTokenBudgetTest` — cumulative headers, budget enforcement, final-answer retention, boundary, multi-iteration stop (6 tests)


gnodet (approved):

## CAMEL-23393: Enforce agentic loop token budget and expose cumulative token usage

**Verdict: APPROVE** ✅

Addresses a real operational concern — the MCP agentic loop was only capped by `maxToolIterations` (default 50), but each iteration could consume significant tokens, leading to unexpectedly large costs without any token-level guardrail.

### What's good

- **Clean tracker design**: `OpenAIAgenticTokenTracker` is a focused accumulator — `final class`, null-safe `addUsage(ChatCompletion)` that handles optional `usage()`, and computed `getTotalTokens()`. Using `long` for the counters is appropriate given that cumulative token counts across many iterations could exceed int range.

- **Correct enforcement timing**: The budget check runs _after_ each API call (`tokenTracker.addUsage` → `setAgenticTokenHeaders` → `enforceAgenticTokenBudget`). This is the right order — you can't know the token count until after the call completes. The exception prevents further iterations while allowing the response from the budget-exceeding call to be available on the exchange (the headers are set before the exception).

- **Observability regardless of budget**: The three cumulative headers (`CamelOpenAIAgenticPromptTokens`, `CamelOpenAIAgenticCompletionTokens`, `CamelOpenAIAgenticTotalTokens`) are set on every agentic loop iteration whether or not a budget is configured. This enables monitoring and dashboarding without requiring a budget.

- **Consistent error pattern**: `IllegalStateException` with a descriptive message matches the existing `maxToolIterations` enforcement pattern. The message includes the budget, iteration number, and per-category token breakdown — very useful for debugging.

- **Backward-compatible mock changes**: The `ResponseBuilder` updates maintain all existing method signatures and delegate to new overloads with defaults (`DEFAULT_PROMPT_TOKENS = 10`, `DEFAULT_COMPLETION_TOKENS = 5`). Existing tests are unaffected.

- **Good test coverage**: The unit test verifies accumulation, and the two integration tests cover both the observability path (headers set on success) and the enforcement path (exception with correct message when budget exceeded).

Well-implemented feature — clean, focused, and consistent with existing patterns. 👏

davsclaus:
@atiaomar1978-hub would you mind leaving a Co-Authored-By in the commit message when you use AI coding assitance

atiaomar1978-hub:
@davsclaus  sure.  
I’ve been doing that, but Cursor removed the Co-authored-by line. I’ll make sure it’s included next time.

gnodet (commented):
**Re-review (3rd commit `4dc63fea`):** Regenerated OpenAI metadata for agentic token budget.

Standard generated metadata update — catalog JSON (`openai.json`), component JSON, and Endpoint DSL builder factory. Changes add the new token budget headers and options from the prior functional commit.

Approval stands. ✅

gnodet (commented):
**Re-review (4th commit `02e2532d`):** Regenerate endpoint DSL headers for agentic token usage.

Standard generated code — adds `openAIAgenticPromptTokens()`, `openAIAgenticCompletionTokens()`, and `openAIAgenticTotalTokens()` header constants to the Endpoint DSL builder. Consistent with the token budget feature from the prior commits.

Approval stands. ✅

atiaomar1978-hub:
@davsclaus would you please trigger the CI build ?

atiaomar1978-hub:
I will resolve conflicts after the merge of CAMEL-23394: Limit OpenAI conversation history size 
https://github.com/apache/camel/pull/24901

oscerd (commented):
Good addition — a cost ceiling on the agentic loop is a real operational need, and this is implemented cleanly. A couple of things I checked specifically that came out well:

- **The tracker is correctly scoped.** `OpenAIAgenticTokenTracker` is a local variable inside `processNonStreamingAgentic` (`OpenAIProducer.java:484`), not a field on the Producer or Endpoint, so a fresh instance exists per invocation and it is thread-confined to the calling thread. No token-count leakage between concurrent exchanges — which is the failure mode I was most worried about. It is also package-private, so no new public API surface.
- **It genuinely enforces**, rather than only observing: `enforceAgenticTokenBudget` (`OpenAIProducer.java:620-630`) throws `IllegalStateException` with a descriptive message, mirroring the existing `maxToolIterations` behaviour.
- **Headers follow the convention.** The three new constants carry `@Metadata` in `OpenAIConstants`, use the `Camel` + component + feature PascalCase form per `design/headers.adoc`, and are therefore covered by `DefaultHeaderFilterStrategy`.
- Off by default (`0`), so no upgrade-guide entry is required. Generated files are all present and correct.

Two things I would like you to consider before this merges.

### A successful final answer that crosses the budget is thrown away

`enforceAgenticTokenBudget` is called at `OpenAIProducer.java:491`, which is *before* `requireFirstChoice` at `:493` and before the final-response branch at `:495`. So when the model returns a normal `stop` response with no tool calls, and that response happens to push cumulative usage over the budget, the producer throws instead of returning the answer.

The concrete case: `maxAgenticTokens=1000`, MCP tools configured, and the model answers directly on the first call using 1200 tokens. No agentic looping happened at all, the tokens are already spent and unrecoverable, a complete valid answer is sitting in `choice` — and the route fails with no body. The budget exists to prevent *further* spend, and after a final response there is no further spend to prevent, so throwing there costs the user their answer without saving anything.

Moving the check to after the final-response early-return, or gating it on `isToolCallsFinishReason(choice)`, would keep the protection while not discarding work already paid for.

### Enforcement is post-call, and the option description reads like a hard cap

The check runs after `create()` at `:488`, which is the pragmatic choice — you cannot know a call's token cost without a tokenizer. But it means real spend can exceed the configured budget by up to one full call, and since the prompt grows each iteration, that last call is typically the most expensive one. The description ("Maximum cumulative prompt plus completion tokens allowed across the MCP agentic loop") does not convey that overshoot. Worth a sentence.

Related, and the one project-rule gap I found: **the PR touches no `.adoc` file**. `components/camel-ai/camel-openai/src/main/docs/openai-mcp.adoc:188` documents the sibling option in exactly the place this one belongs:

> The `maxToolIterations` option (default: 50) prevents infinite loops. If exceeded, an `IllegalStateException` is thrown.

`maxAgenticTokens` is the direct analogue and should sit on the adjacent line, covering the default, the post-call overshoot, and how the cumulative headers differ from the existing per-call ones. The generated option/header tables will populate automatically via the included partials, so this is narrative documentation only — but CLAUDE.md does ask for doc updates where applicable, and this is applicable.

### Smaller notes

- **Confusable header pairs, with inconsistent declared types.** In the agentic path both sets land on the same message: `setResponseHeaders` sets `CamelOpenAIPromptTokens` / `CompletionTokens` / `TotalTokens` (last call only) while `setAgenticTokenHeaders` sets the `...Agentic...` variants (cumulative). The pre-existing three are declared `javaType = "Integer"` at `OpenAIConstants.java:71-76`, the new three `"Long"`, though both are populated from the same `long`-returning SDK accessor. Your new declarations are the correct ones — the old ones are wrong. Not caused by this PR, but it puts the inconsistency in adjacent catalog rows, so it may be worth fixing opportunistically or filing a follow-up.
- **Test-infra mock now emits `usage` unconditionally.** `ResponseBuilder.createBaseChatCompletion` always adds a usage block (10/5). Signatures are preserved, but the emitted JSON changed for every consumer, so existing mock-backed tests now get `CamelOpenAIPromptTokens` etc. populated where they previously were absent. Nothing in-repo asserts their absence and CI is green, but `camel-test-infra-openai-mock` is a published artifact used by `camel-langchain4j-agent` and `camel-langchain4j-tools`, so downstream repos could see the change. An opt-in emission, or a brief note, would be safer.
- **Test coverage gaps.** The budget test only covers "first response already over budget". Missing: accumulation crossing the budget across *multiple* iterations (the actual scenario in the JIRA), the boundary case where `total == maxAgenticTokens` must **not** throw, and an assertion that the loop actually stopped issuing requests.
- **Negative values silently disable** the budget (`:623` guards `<= 0` while the docs say "When 0"). Cosmetic, and consistent with `maxToolIterations`.
- **Scope**: only `processNonStreamingAgentic` honours the budget; the simple and streaming paths ignore it. The description does say "across the MCP agentic loop", so it is stated — just worth being deliberate about.

None of this is a correctness bug, so I am leaving it as a comment rather than blocking. The final-answer case is the one I would most like to see addressed.

---

gnodet (commented):
## Re-review: Follow-up commit (e3193e93) — Address agentic token budget review feedback

Good improvements across the board. The core logic change, documentation, and new tests are all well done.

### Key changes reviewed

**1. Enforcement logic placement (OpenAIProducer.java) — Correct ✅**

Moving `enforceAgenticTokenBudget()` from before the stop-condition check to after it is the right call. If the model returns a final text answer (no more tool calls), there's no value in rejecting it — the user already paid for those tokens and no further spend will occur. Budget enforcement only makes sense as a gate before spending more tokens on additional tool calls.

**2. Integer → Long for per-call token headers — Correct ✅**

The OpenAI Java SDK returns `long` from `CompletionUsage.promptTokens()`, so the previous `Integer` annotation in `@Metadata` was actually wrong. Since camel-openai is unreleased (new in 4.22 SNAPSHOT), this carries no backward-compat risk.

**3. New tests — Thorough ✅**

The three new tests cover important scenarios:
- `finalAnswerOverBudgetShouldStillReturnResponse` — validates the enforcement-position fix
- `maxAgenticTokensAtExactBoundaryShouldAllowLoopToContinue` — boundary case (total == budget → continues)
- `maxAgenticTokensShouldStopAfterMultipleIterations` — multi-step accumulation with `verify(weatherClient, times(1))` confirming only one tool call before enforcement

**4. Documentation (openai-mcp.adoc) — Good ✅**

Clear explanation of enforcement timing, the distinction between per-call and cumulative headers, and the behavior when budget is exceeded.

### One item to address

**Catalog-level generated JSON may be stale**: The component-level `openai.json` was correctly updated (Integer → Long, improved descriptions), but the catalog-level copy at `catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/openai.json` may not have been regenerated after this commit's metadata changes. The CI build checks for uncommitted generated file changes, so this should be caught — but worth running the build/regeneration step to sync.

atiaomar1978-hub:
@gnodet 

Good catch — this was valid at the time of `e3193e93a7ab` ("Address agentic token budget review feedback"), which updated the component-level `openai.json` but not the catalog copy.

That gap is closed in `c40eb0559289` (merge of `main` into `CAMEL-23393-agentic-token-budget`), which regenerated the catalog metadata. I verified that:

- `components/camel-ai/camel-openai/src/generated/resources/META-INF/org/apache/camel/component/openai/openai.json`
- `catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/openai.json`

are now identical, including:
- `Integer` → `Long` for per-call token headers
- the updated `maxAgenticTokens` description
- the agentic cumulative token headers

CI sourcecheck should pass on this — no further regeneration needed on the branch.

davsclaus:
same here with merge conflict - rebase

components/camel-ai/camel-openai/src/test/java/org/apache/camel/component/openai/OpenAIAgenticTokenBudgetTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Test
+    void finalAnswerOverBudgetShouldStillReturnResponse() {
+        String endpointUri = "openai:chat-completion?model=gpt-5&apiKey=dummy&autoToolExecution=true"
+                             + "&maxAgenticTokens=100&maxToolIterations=5&baseUrl="
+                             + tokenMock.getBaseUrl() + "/v1";
+        injectMcpTools(endpointUri, createWeatherToolClients());
+
+        Exchange result = template.request("direct:agentic-token-budget",
+                e -> e.getIn().setBody("expensive direct answer"));
```
davsclaus, line 140: Project convention: do not mix JUnit and AssertJ assertions in the same test method (`CLAUDE.md` § Assertions). `assertInstanceOf` is a JUnit assertion, while `assertThat` on the next line is AssertJ. The same pattern appears in `maxAgenticTokensShouldStopAfterMultipleIterations()`.

```suggestion
        assertThat(result.getException()).isInstanceOf(IllegalStateException.class);
```

davsclaus (changes requested):
_Claude Code review on behalf of davsclaus — AI-generated review_

## CAMEL-23393: Agentic loop token budget

Nice work on this feature — the cumulative token tracking and budget enforcement logic is clean, the test coverage is solid, and the mock infra enhancements are well-scoped. Two items need attention before merge:

1. **Missing upgrade guide entry** — The per-call token headers (`CamelOpenAIPromptTokens`, `CamelOpenAICompletionTokens`, `CamelOpenAITotalTokens`) changed their declared type from `Integer` to `Long`. While the SDK returns `long` values (so this is a correctness fix), it's a user-visible API change that should be documented in the `=== camel-openai` section of `camel-4x-upgrade-guide-4_22.adoc`.

2. **Mixed JUnit/AssertJ assertions** — Two test methods mix `assertInstanceOf` (JUnit) with `assertThat` (AssertJ). Per project conventions, assertions should not be mixed within the same method. See inline comment.

This review covers project rules and conventions only. It is not a replacement for specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

davsclaus:
another rebase is needed

components/camel-ai/camel-openai/src/test/java/org/apache/camel/component/openai/OpenAIAgenticTokenBudgetTest.java
atiaomar1978-hub: @davsclaus I will fix.

atiaomar1978-hub: Done

atiaomar1978-hub:
Merged main into this branch and fixed review feedback. 
@davsclaus could you please trigger the CI build and tests ?

davsclaus (approved):
_Claude Code review on behalf of davsclaus — AI-generated review_

All prior review feedback has been addressed:

- Upgrade guide entry for the Integer→Long header type change — added
- Mixed JUnit/AssertJ assertions — fixed (AssertJ-only now)
- Final-answer enforcement placement — correctly moved after the stop-condition early-return
- Post-call overshoot documented in `openai-mcp.adoc`
- Expanded test coverage (6 tests including boundary, multi-iteration, and final-answer-over-budget)
- Generated metadata regenerated and in sync

CI is green. Nice work on this feature — the token budget is a real operational need for agentic loops, and the implementation is clean and well-tested.

This review covers project rules and conventions only. It is not a replacement for specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

PR #24903, opened 2026-07-19
CAMEL-24198: Track Event Hubs checkpoint batching per partition

atiaomar1978-hub:
**CAMEL-24198** is implemented, tested, committed, and pushed.

## Issue
**[CAMEL-24198](https://issues.apache.org/jira/browse/CAMEL-24198)** — *camel-azure-eventhubs - Redesign checkpoint store to use per-partition batching*

The consumer used a single shared `processedEvents`, `lastTask`, and `lastScheduledTask` across all partitions, so checkpoint commits from different partitions could interfere with each other.

## Fix
In `EventHubsConsumer.java`, replaced shared fields with per-partition maps:
- `processedEventsByPartition`
- `checkpointTasksByPartition`
- `scheduledTasksByPartition`

Each partition now tracks its own batch count and schedules its own timeout checkpoint independently.

## Tests
Added `EventHubsConsumerPerPartitionCheckpointTest.java` with 4 tests:
- Per-partition event counters stay independent
- Batch-size checkpoint on one partition does not reset another
- Each partition gets its own scheduled checkpoint task
- Same partition reuses its checkpoint task within a batch window

**12/12 tests passed** (including existing shutdown and updater task tests).

## Git
| Item | Value |
|------|-------|
| **Branch** | `CAMEL-24198-eventhubs-per-partition-checkpoint-batching` |
| **Commit** | `8896d7755582` |
| **Remote** | Pushed to `origin` |

### DIFF
--- a/components/camel-azure/camel-azure-eventhubs/pom.xml
+++ b/components/camel-azure/camel-azure-eventhubs/pom.xml
@@ -66,6 +66,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <version>${mockito-version}</version>
--- a/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
+++ b/components/camel-azure/camel-azure-eventhubs/src/main/java/org/apache/camel/component/azure/eventhubs/EventHubsConsumer.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.component.azure.eventhubs;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -46,16 +48,14 @@ public class EventHubsConsumer extends DefaultConsumer implements ShutdownAware 
     // we use the EventProcessorClient as recommended by Azure docs to consume from all partitions
     private EventProcessorClient processorClient;
 
-    private final AtomicInteger processedEvents;
     private final AtomicInteger pendingExchanges = new AtomicInteger();
+    private final Map<String, AtomicInteger> processedEventsByPartition = new ConcurrentHashMap<>();
+    private final Map<String, ScheduledFuture<?>> scheduledTasksByPartition = new ConcurrentHashMap<>();
+    private final Map<String, EventHubsCheckpointUpdaterTask> checkpointTasksByPartition = new ConcurrentHashMap<>();
     private ScheduledExecutorService scheduledExecutorService;
-    private ScheduledFuture<?> lastScheduledTask;
-    private EventHubsCheckpointUpdaterTask lastTask;
 
     public EventHubsConsumer(final EventHubsEndpoint endpoint, final Processor processor) {
         super(endpoint, processor);
-
-        this.processedEvents = new AtomicInteger();
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


gnodet (approved):

## CAMEL-24198: Track Event Hubs checkpoint batching per partition

**Verdict: APPROVE** ✅

Fixes a real concurrency bug — the consumer used shared `processedEvents`, `lastTask`, and `lastScheduledTask` across all partitions, so checkpoint batching from different partitions interfered with each other (e.g., partition 0 reaching batch size could reset the counter that partition 1 was tracking).

### What's good

- **Correct data structure choice**: `ConcurrentHashMap` for all three per-partition maps (`processedEventsByPartition`, `checkpointTasksByPartition`, `scheduledTasksByPartition`). Even though `processCommit()` is `synchronized`, the maps may be read from other paths, and `ConcurrentHashMap` is the right default for concurrent access.

- **Minimal, focused change**: Only the `processCommit()` method changes — the data structures shift from shared fields to per-partition maps, and the logic is straightforward: look up by `partitionId`, create if absent, operate on partition-local state.

- **Null-check simplification is correct**: `lastTask != null && lastTask.isExpired()` → `checkpointTask.isExpired()`. The null check was needed before because `lastTask` could be null from a prior invocation. Now `checkpointTask` is guaranteed non-null at the `else if` point — either just created (first branch) or fetched from the map (second branch).

- **Improved debug logging**: All log messages now include the `partitionId`, which is essential for diagnosing multi-partition issues in production.

- **Good test coverage**: 4 tests verify the key partition isolation properties — independent counters, batch-size checkpoint isolation, separate scheduled tasks, and task reuse within the same partition. The `ArgumentCaptor` verification that scheduled tasks are distinct instances is a nice touch.

### Non-blocking observations

- **Memory cleanup**: The per-partition maps grow as partitions are seen but aren't explicitly cleared on `doStop()`. This is fine in practice since Event Hub partition counts are bounded (typically 2–32), but a `clear()` in `doStop()` would be tidy for completeness.

- **Test assertions style**: The tests use JUnit 5 `assertEquals`/`assertNotSame` rather than AssertJ. Not a blocker — both are valid in Camel, and the assertions are clear.

Clean fix for a real partition-interference bug. 👏

atiaomar1978-hub:
**Last change summary** — commit `0905a931b031` on `CAMEL-24198-eventhubs-per-partition-checkpoint-batching`

### What changed
Review feedback on the per-partition checkpoint batching work:

**1. `EventHubsConsumer.java`**
- In `doStop()`, clear all three per-partition maps:
  - `processedEventsByPartition`
  - `scheduledTasksByPartition`
  - `checkpointTasksByPartition`
- Avoids leaving partition state around after the consumer stops (partition count is small, but this keeps cleanup explicit).

**2. `EventHubsConsumerPerPartitionCheckpointTest.java`**
- Switched assertions from JUnit to **AssertJ** (`assertThat(...).isEqualTo()`, `.isZero()`, `.hasSize()`, `.isEmpty()`, `.isNotSameAs()`).
- Added **`doStopClearsPerPartitionCheckpointState`** to verify all maps are empty after `doStop()`.

**3. `pom.xml`**
- Added **`assertj-core`** (test scope), aligned with other Azure components like `camel-azure-servicebus`.

### Verification
- **20/20** EventHubs tests passed
- **Pushed** to `origin/CAMEL-24198-eventhubs-per-partition-checkpoint-batching`

### Branch history
| Commit | Description |
|--------|-------------|
| `8896d7755582` | Initial per-partition checkpoint batching fix |
| `0905a931b031` | Review follow-up (cleanup + AssertJ + new test) |

gnodet (commented):
**Re-review of commit 2 (`0905a93`)** — _addresses review feedback_

This commit cleanly addresses both non-blocking observations from my initial review:

1. **Memory cleanup in `doStop()`** ✅ — All three per-partition maps (`processedEventsByPartition`, `scheduledTasksByPartition`, `checkpointTasksByPartition`) are now cleared in `doStop()`, placed correctly after stopping the `processorClient` but before `super.doStop()`. This prevents stale state from accumulating across restart cycles.

2. **AssertJ assertion style** ✅ — All assertions migrated from JUnit `assertEquals`/`assertNotSame` to idiomatic AssertJ (`assertThat(...).isEqualTo(...)`, `.isZero()`, `.hasSize()`, `.isNotSameAs()`, `.isEmpty()`). The `assertj-core` test dependency was added to `pom.xml`.

3. **New test `doStopClearsPerPartitionCheckpointState()`** ✅ — Good addition that verifies all three maps are empty after `doStop()` and confirms the processor client is stopped. The test properly sets up the `processorClient` mock via reflection and populates two partitions before stopping.

Approval from the previous review stands. Well done addressing the feedback!

---

PR #24904, opened 2026-07-19
CAMEL-24192: camel-lra - separate compensation/completion URI allowlists and prune on route removal

davsclaus:
## Summary

Splits the single shared `sagaURIs` allowlist in `LRASagaService` into separate compensation and completion URI sets, and adds automatic cleanup when routes are removed.

**Problem:** The LRA Saga service maintained one flat `Set<String>` for both compensation and completion callback URIs. This meant (1) a completion callback could invoke a compensation endpoint and vice versa, and (2) URIs from dynamically removed routes remained authorized for the entire context lifetime (memory leak + stale authorization).

**Changes:**
- **`CamelSagaStep`** — added `routeId` field so steps can be tracked by the route that registered them
- **`SagaReifier`** — sets `routeId` on the step before calling `registerStep()`
- **`LRASagaService`** — replaced `sagaURIs` with separate `compensationURIs` and `completionURIs` sets; added `stepsByRouteId` tracking map; registered a `LifecycleStrategySupport` that prunes URIs when routes are removed; shared URIs are only pruned when all referencing routes are removed
- **`LRASagaRoutes`** — split `verifyRequest()` into `verifyCompensationRequest()` and `verifyCompletionRequest()` so each callback type validates only against its own allowlist
- **New test** `LRASagaServiceAllowlistTest` — covers URI separation, pruning, shared URI retention, and no-op for unknown routes

No interface changes — `CamelSagaService.registerStep(CamelSagaStep)` signature is preserved. `InMemorySagaService` is unaffected.

## Test plan

- [x] `LRASagaServiceAllowlistTest` — 6 tests covering separation, pruning, shared URIs, union, and no-op
- [x] All existing LRA tests pass (`mvn verify` in `components/camel-lra`)
- [x] `camel-core-reifier` compiles with the new `step.setRouteId()` call

### DIFF
--- a/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
+++ b/components/camel-lra/src/main/java/org/apache/camel/service/lra/LRASagaRoutes.java
@@ -18,7 +18,10 @@ package org.apache.camel.service.lra;
 
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.camel.Exchange;
@@ -44,7 +47,7 @@ public class LRASagaRoutes extends RouteBuilder {
         rest(sagaService.getLocalParticipantContextPath())
                 .put(PARTICIPANT_PATH_COMPENSATE).to("direct:lra-compensation");
         from("direct:lra-compensation").routeId("lra-compensation")
-                .process(this::verifyRequest)
+                .process(this::verifyCompensationRequest)
                 .choice()
                 .when(header(URL_COMPENSATION_KEY).isNotNull())
                 .toD("${header." + URL_COMPENSATION_KEY + "}")
@@ -53,7 +56,7 @@ public class LRASagaRoutes extends RouteBuilder {
         rest(sagaService.getLocalParticipantContextPath())
                 .put(PARTICIPANT_PATH_COMPLETE).to("direct:lra-completion");
         from("direct:lra-completion").routeId("lra-completion")
-                .process(this::verifyRequest)
+                .process(this::verifyCompletionRequest)
                 .choice()
                 .when(header(URL_COMPLETION_KEY).isNotNull())
                 .toD("${header." + URL_COMPLETION_KEY + "}")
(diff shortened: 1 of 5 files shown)


gnodet (approved):
## Review: CAMEL-24192 — Separate LRA compensation/completion URI allowlists

**Overall: Clean security improvement** — splits the flat `sagaURIs` allowlist into type-specific sets so compensation callbacks can only invoke compensation endpoints and vice versa, and adds automatic cleanup when routes are removed.

### What this PR does well

1. **Security hardening** — The original single `Set<String>` allowed cross-type callback invocation (a completion callback could dispatch to a compensation URI). Splitting into `compensationURIs` and `completionURIs` with type-specific validation in `verifyRequest(Exchange, boolean)` closes this gap.

2. **Correct cleanup design** — `unregisterSteps()` properly handles shared URIs: it collects URIs still referenced by other routes before removing, preventing premature pruning. The `LifecycleStrategy` hook fires on `onRoutesRemove()` to trigger cleanup automatically.

3. **Backward-compatible `getRegisteredURIs()`** — Returns the union of both sets, so existing callers are unaffected.

4. **API fix** — Changed `Optional.map()` to `Optional.ifPresent()` for side-effect operations in `registerStep()` (the old `.map(this.sagaURIs::add)` misused `map` for side effects).

5. **Good test coverage** — 6 tests covering separation, union, pruning, shared-URI retention, and unknown-route no-op.

6. **Import cleanup** — Wildcard `java.util.*` replaced with specific imports in `LRASagaRoutes`.

### Non-blocking observations

1. **Thread safety in `unregisterSteps()`** — There's a narrow TOCTOU window between `stepsByRouteId.remove(routeId)` and iterating `stepsByRouteId.values()`: a concurrent `registerStep()` could add a step referencing the same URI between those two operations. In practice this isn't a problem since route registration and removal are sequential during lifecycle transitions, but worth noting.

2. **Lifecycle strategy cleanup** — Consider removing the lifecycle strategy in `doStop()` (via `camelContext.removeLifecycleStrategy(this.lifecycleStrategy)`) alongside clearing the maps, and resetting `this.lifecycleStrategy = null` so it gets re-registered on restart. Currently the null-guard prevents double-registration, but explicit removal would be cleaner for restart scenarios.

3. **Assertion style** — Tests use JUnit `assertTrue`/`assertFalse`/`assertEquals`. The project preference is AssertJ (`assertThat(...).contains(...)`, `.doesNotContain(...)`, `.isEmpty()`), which would also give better failure messages for set membership checks.

---

davsclaus:

Thanks for the review!

Addressed observation #2 — lifecycle strategy is now explicitly removed in `doStop()` and the field is reset to `null`, so it gets properly re-registered on restart. See d84f42edc76c.

On the other observations:
- **Thread safety**: Agreed it's fine in practice since route lifecycle transitions are sequential.
- **Assertion style**: The existing LRA tests all use JUnit `Assertions` (zero `assertThat` usage across the module), so kept consistent with the existing pattern.

gnodet (commented):
Re-review after 2nd commit (d84f42e): **Lifecycle strategy cleanup in doStop() looks good** 👍

The new commit properly deregisters the `LifecycleStrategy` from the CamelContext during `doStop()`:

```java
if (this.lifecycleStrategy != null) {
    camelContext.getLifecycleStrategies().remove(this.lifecycleStrategy);
    this.lifecycleStrategy = null;
}
```

This prevents the strategy from lingering if the service is stopped while the CamelContext continues running. Clean follow-up. My earlier approval stands.

PR #24906, opened 2026-07-19
CAMEL-23973: camel export - Add --parent-pom option

davsclaus:
## Summary

- Adds `--parent-pom=groupId:artifactId:version` option to `camel export` for setting a custom Maven parent POM in the exported project
- Supports all three runtimes: Spring Boot, Quarkus, and Camel Main
- For Spring Boot, replaces the default `spring-boot-starter-parent`; the `spring-boot-dependencies` BOM in `<dependencyManagement>` keeps version management intact
- Also configurable via `camel.jbang.parentPom` in `application.properties`
- Adds `MavenGav.parseStrictGav()` for shared strict GAV validation — used by both `--gav` and `--parent-pom` validation across all export commands
- Validates `--parent-pom` early in `export()` with a friendly error message (consistent with `--gav` validation)
- Includes documentation in the export guide

## Test plan

- [x] Unit tests: `shouldExportWithParentPom` — verifies custom parent is set for all runtimes, and Spring Boot BOM is preserved
- [x] Unit tests: `shouldExportWithoutParentPom` — verifies default behavior is unchanged
- [x] All 81 existing ExportTest tests pass
- [x] Full reactor build passes with regenerated metadata


gnodet (approved):
## Review: CAMEL-23973 — camel export --parent-pom option

**Overall: Well-implemented feature** that addresses a real organizational need for custom parent POMs in exported Camel projects.

### What works well

1. **Thoughtful Spring Boot BOM handling** — When `--parent-pom` replaces the default `spring-boot-starter-parent`, the `spring-boot-dependencies` BOM remains in `<dependencyManagement>` so version management continues to work. The FreeMarker templates use `[#if ParentGroupId??]` conditionals cleanly.

2. **Clean layering with `enrichParentPom()`** — Centralizing the model population in `ExportBaseCommand` and calling it uniformly from all three runtime exporters avoids duplication. Each exporter adds one line (`enrichParentPom(model)`) and the templates handle the rest.

3. **Comprehensive coverage** — All six POM templates updated, tests for both with/without parent POM across all runtimes, documentation in the export guide with examples, and `application.properties` support via `camel.jbang.parentPom`.

4. **`<relativePath/>` in generated parent block** — Prevents Maven from searching parent directories, which is the right default for a remote parent.

### Non-blocking observation

**Input validation for malformed `--parent-pom`** — The `enrichParentPom()` method splits on `:` and silently ignores input with != 3 parts. If a user specifies `--parent-pom=com.ourorg:camel-parent` (forgetting the version), the export succeeds with exit 0 but the parent POM is never applied. For Spring Boot, this silently falls back to the default `spring-boot-starter-parent`.

The existing `--gav` option validates and fails fast:
```java
if (ids.length != 3) {
    printer().printErr("--gav must be in syntax: groupId:artifactId:version");
    return 1;
}
```

Consider adding similar validation for `--parent-pom` for consistency:
```java
protected void enrichParentPom(Map<String, Object> model) {
    if (parentPom != null) {
        String[] parts = parentPom.split(":");
        if (parts.length != 3) {
                "--parent-pom must be in syntax: groupId:artifactId:version, got: " + parentPom);
        }
        model.put("ParentGroupId", parts[0]);
        model.put("ParentArtifactId", parts[1]);
        model.put("ParentVersion", parts[2]);
    }
}
```

This can be addressed as a follow-up — the feature works correctly for valid input.

---

davsclaus:
Thanks for the review @gnodet! Good catch on the silent ignore for malformed `--parent-pom`.

Addressed by adding `MavenGav.parseStrictGav(String)` — a shared strict GAV validation method that requires exactly `groupId:artifactId:version` and throws `IllegalArgumentException` otherwise. Updated all four call sites:

- `ExportSpringBoot`, `ExportCamelMain`, `ExportQuarkus` — `--gav` validation now uses `parseStrictGav` instead of inline split+check
- `ExportBaseCommand.enrichParentPom()` — `--parent-pom` now uses `parseStrictGav` and fails fast on malformed input instead of silently ignoring

Also regenerated `camel-jbang-configuration-metadata.json`.

gnodet (commented):
Re-review after 2nd commit (36f357c): **Strict GAV validation via `MavenGav.parseStrictGav()` — nice** 👍

The new commit addresses the observation from my earlier review about silently ignoring malformed `--parent-pom` values. Changes:

1. **New `MavenGav.parseStrictGav()`** — validates exactly 3 colon-separated parts, throws `IllegalArgumentException` with a descriptive message otherwise. Good Javadoc.
2. **`enrichParentPom()`** — now uses `parseStrictGav()` instead of manual `split(":")`, so malformed input throws rather than being silently ignored.
3. **Bonus: unified `--gav` validation** — `ExportCamelMain`, `ExportQuarkus`, and `ExportSpringBoot` all upgraded to use `parseStrictGav()` for `--gav` as well, replacing the manual `ids.length != 3` checks. Consistent.

Minor note: `enrichParentPom()` lets the `IllegalArgumentException` propagate uncaught, whereas the `--gav` validation wraps it in a try/catch with a clean error message. If a user passes `--parent-pom foo:bar` they'll get a stack trace rather than a friendly message. Not blocking — it's strictly better than silently ignoring, and easily refined later if needed.

My earlier approval stands.

gnodet (commented):
**Re-review (3rd commit `816a4a11`):** Early `--parent-pom` validation with friendly error message.

Good addition — validates `--parent-pom` early (right after `--gav` validation) using `parseStrictGav()` with a user-friendly error message, consistent across all three export commands (CamelMain, Quarkus, SpringBoot). This catches malformed parent POM coordinates before any export work begins, matching the existing `--gav` validation pattern directly above.

Prior approval stands. ✅

PR #24909, opened 2026-07-19
CAMEL-24200: camel-aws2-timestream - createScheduledQuery sets the time column correctly

oscerd:
## Description

Fixes [CAMEL-24200](https://issues.apache.org/jira/browse/CAMEL-24200). In `Timestream2QueryProducer.createScheduledQuery` the time-column header was written into the target configuration's **table name**:

```java
if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(Timestream2Constants.TIME_COLUMN))) {
    String timeColumn = exchange.getIn().getHeader(Timestream2Constants.TIME_COLUMN, String.class);
    timestreamConfigBuilder.tableName(timeColumn);   // should be .timeColumn(...)
}
```

So the `timeColumn` (required on a scheduled-query target) was never set, and when `CamelAwsTimestreamTableName` was also supplied the time-column value silently **overwrote the table name** — creating the scheduled query against the wrong table, or having AWS reject it.

Also removes a duplicated block that set `scheduledQueryExecutionRoleArn` twice from the same header.

## Tests

Extends the existing `Timestream2QueryProducerTest.timestreamCreateScheduledQueryTest` (which already sent all three headers but only asserted the response ARN) to capture the `CreateScheduledQueryRequest` via the client mock and assert `databaseName`, `tableName` and `timeColumn` each come from their own header. Against the old code the table name would be `"time"` and the time column `null`, so the test fails without the fix.

## Backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-timestream/src/main/java/org/apache/camel/component/aws2/timestream/query/Timestream2QueryProducer.java
+++ b/components/camel-aws/camel-aws2-timestream/src/main/java/org/apache/camel/component/aws2/timestream/query/Timestream2QueryProducer.java
@@ -199,11 +199,6 @@ public class Timestream2QueryProducer extends DefaultProducer {
                         = ErrorReportConfiguration.builder().s3Configuration(s3Configuration.build()).build();
                 builder.errorReportConfiguration(errorReportConfiguration);
             }
-            if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(Timestream2Constants.SCHEDULED_QUERY_EXECUTION_ROLE_ARN))) {
-                String roleArn
-                        = exchange.getIn().getHeader(Timestream2Constants.SCHEDULED_QUERY_EXECUTION_ROLE_ARN, String.class);
-                builder.scheduledQueryExecutionRoleArn(roleArn);
-            }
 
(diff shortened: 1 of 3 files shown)


gnodet (approved):
**LGTM** — clear, well-tested bug fix.

**The bug**: `timestreamConfigBuilder.tableName(timeColumn)` should be `.timeColumn(timeColumn)`. This meant the time column was never set, and when both `CamelAwsTimestreamTableName` and `CamelAwsTimestreamTimeColumn` headers were supplied, the time-column value silently overwrote the table name. Straightforward copy-paste error.

**The fix**:
1. `tableName(timeColumn)` → `timeColumn(timeColumn)` — correct builder method ✅
2. Removed duplicate `scheduledQueryExecutionRoleArn` block (confirmed: lines 178 and 205 on `main` are identical; the PR removes the second) ✅

**The test**: Capturing `CreateScheduledQueryRequest` in the mock client and asserting that `databaseName`, `tableName`, and `timeColumn` each map to their respective headers is the right pattern. The test would fail without the fix (`tableName` → `"time"`, `timeColumn` → `null`).

No concerns.

davsclaus (approved):

PR #24919, opened 2026-07-20
CAMEL-24168: Fix BigDecimal precision handling when no explicit precision set

davsclaus:
## Summary

- Fix `@DataField.precision()` default from `0` to `-1` so that BigDecimal fields without explicit precision preserve the original scale from input, instead of crashing with `ArithmeticException` on unmarshal or silently truncating decimals on marshal
- Fix `AbstractNumberFormat` to allow `-1` (unset) to pass through instead of clamping to `0`, and skip `DecimalFormat` fraction digit configuration when precision is unset
- Wire `@DataField.rounding()` into `BigDecimalFormatFactory` — previously it was only honored when a `pattern` was set (via `BigDecimalPatternFormatFactory`)
- Add upgrade guide entry documenting the default change

## Test plan

- [x] New `BindyBigDecimalNoPrecisionTest` — 4 tests covering unmarshal/marshal round-trip of fractional BigDecimal without precision
- [x] Existing `BindyNumberTest` (2 tests) — no regressions for fixed-length BigDecimal/Double with and without explicit precision
- [x] Existing `BindyBigDecimalRoundingUnmarshallTest` — no regression for precision+rounding+pattern path
- [x] Full camel-bindy test suite — 226 tests pass, 0 failures

### DIFF
--- a/components/camel-bindy/src/main/docs/bindy-dataformat.adoc
+++ b/components/camel-bindy/src/main/docs/bindy-dataformat.adoc
@@ -483,7 +483,8 @@ use "default" to use platform default locale.
 the field in the CSV generated (output message) must be different compare to input position (pos). See the pos
 parameter.
 
-| precision | int |  | 0 | precision of the {@link java.math.BigDecimal} number to be created
+| precision | int |  | -1 a| precision of the {@link java.math.BigDecimal} number to be created. Use -1 (default) to preserve the original
+scale from the input.
 
 | required | boolean |  | false | Indicates if the field is mandatory
 
@@ -1346,7 +1347,7 @@ to @Message.continueParseOnFailure.
 | position | int |  | 0 a| Position of the field in the message generated - must be used when the position of the key/tag in the FIX message
 must be different
 
-| precision | int |  | 0 | precision of the BigDecimal number to be created
+| precision | int |  | -1 | precision of the BigDecimal number to be created. Use -1 (default) to preserve the original scale from the input.
 
 | required | boolean |  | false | Indicates if the field is mandatory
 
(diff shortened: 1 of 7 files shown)


gnodet (approved):
**Thorough and well-structured fix.** The change correctly addresses the `@DataField.precision()` default-value problem that caused both `ArithmeticException` on unmarshal and silent truncation on marshal.

**Code review observations:**

1. **Sentinel value approach** (`-1` = "unset") — clean and consistent across both `DataField` and `KeyValuePairField` annotations.

2. **`AbstractNumberFormat` changes** — Two key fixes:
   - Removing the `precision > 0 ? precision : 0` clamp allows `-1` to flow through correctly as "unset"
   - `Math.max(precision, 0)` in the `impliedDecimalPosition` branch safely guards against negative precision (multiplier becomes `10^0 = 1`)
   - `else if (this.precision >= 0)` correctly skips fraction-digit configuration when precision is unset — this is what prevents the format from forcing a scale

3. **`BigDecimalFormatFactory` rounding wiring** — Good catch. Previously `rounding` was only honored through `BigDecimalPatternFormatFactory`. Now `result.setScale(precision, RoundingMode.valueOf(rounding))` properly uses the configured rounding mode instead of bare `setScale()` which would throw `ArithmeticException` on inexact values. This is safe because `getRounding()` defaults to `"HALF_EVEN"` from the annotation.

4. **Test coverage** — 4 test cases covering the important scenarios: unmarshal with decimals, marshal round-trip, multi-decimal precision preservation, and integer value handling. All use `CamelTestSupport` with `MockEndpoint` assertions.

5. **Upgrade guide** — Clear migration guidance explaining the default change and what to do if relying on the old `precision=0` behavior.

**One minor note:** The existing check `if (super.getPrecision() != -1)` in `BigDecimalFormatFactory.parse()` was already in place (presumably anticipating this sentinel), which makes the fix even cleaner — no additional conditional logic needed on the parse path.

LGTM ✅

PR #24921, opened 2026-07-20
CAMEL-24169: Schema resolver latches first resolved schema forever

davsclaus:
## Summary
- Fix `AvroSchemaResolver` and `ProtobufSchemaResolver` (both jackson and jackson3 variants) unconditionally promoting dynamically resolved schemas into the `this.schema` instance field
- After the first exchange resolves a schema, the per-type `schemes` ConcurrentMap becomes dead code — every subsequent exchange short-circuits to the latched schema regardless of body type or per-exchange properties
- The fix removes the `this.schema = answer` latch from `computeIfAbsent()` in all four resolvers, so `this.schema` is only authoritative when explicitly set via `setSchema()`

## Test plan
- [x] `mvn verify` passes in all four modules: `camel-jackson-avro`, `camel-jackson-protobuf`, `camel-jackson3-avro`, `camel-jackson3-protobuf`
- [ ] Verify with attached reproducer from JIRA (multi-type route with distinct inline schemas)

### DIFF
--- a/components/camel-jackson-avro/src/main/java/org/apache/camel/component/jackson/avro/transform/AvroSchemaResolver.java
+++ b/components/camel-jackson-avro/src/main/java/org/apache/camel/component/jackson/avro/transform/AvroSchemaResolver.java
@@ -165,10 +165,6 @@ public class AvroSchemaResolver implements SchemaResolver, Processor {
             }
         }
 
-        if (answer != null) {
-            this.schema = answer;
-        }
-
         return answer;
     }
 }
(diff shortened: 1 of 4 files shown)


gnodet (approved):
**Clean and correct fix.** The bug is well-diagnosed — `this.schema = answer` in `computeIfAbsent()` was promoting the first dynamically-resolved schema to the instance field, making all subsequent exchanges short-circuit to that latched schema regardless of body type or per-exchange properties.

The fix is minimal and safe: removing the `this.schema = answer` assignment (4 lines × 4 files) ensures `this.schema` is only authoritative when explicitly set via `setSchema()`, while the per-type `schemes` ConcurrentMap correctly handles dynamic resolution.

No risk of regression — the `schemes` map already handles caching by type, so removing the instance-level latch doesn't lose any performance benefit. The `setSchema()` path for explicit configuration remains untouched.

LGTM ✅

PR #24924, opened 2026-07-20
CAMEL-24176: camel-cxfrs - Fix async producer never mapping HTTP errors to CxfOperationException

davsclaus:

## Summary

Fix two bugs in the async CXF-RS producer's `fail()` callback in both `CxfInvocationCallback` and `CxfProxyInvocationCallback`:

1. **Reversed `isInstance` check** — `throwable.getClass().isInstance(WebApplicationException.class)` always evaluates to `false` because it tests whether the `Class` object is an instance of the throwable's class (which it never is). Both `if` branches were dead code, so every async failure fell through to `exchange.setException(throwable)` with the raw JAX-RS exception instead of wrapping it in `CxfOperationException`. Fixed by replacing with `throwable instanceof WebApplicationException` (same for `ResponseProcessingException`).

2. **Silent exception swallowing** — When `shouldHandleError()` returned `false` (e.g., `throwExceptionOnFailure=false` or status ≤ 207), neither `handleError()` nor `exchange.setException()` was called — the exception was silently dropped. Added `else { exchange.setException(throwable); }` fallback in each branch.

## Impact

With the default `synchronous=false`, a server error (404/500) now correctly yields a `CxfOperationException` (carrying status code, response headers and body) — matching the synchronous path behavior. `onException(CxfOperationException.class)` handlers now work consistently regardless of sync/async mode.

## Test plan

- [x] Added `testAsyncProxyProducerServerErrorMappedToCxfOperationException` — calls via the proxy client API with an invalid ID that triggers a 500 on the server; verifies the exchange exception is `CxfOperationException` with status 500 (not a raw `InternalServerErrorException`)
- [x] All 18 existing `CxfRsAsyncProducerTest` tests pass

### DIFF
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsProducer.java
@@ -746,17 +746,19 @@ public class CxfRsProducer extends DefaultAsyncProducer {
         }
 
         private void fail(Throwable throwable) {
-            if (throwable.getClass().isInstance(WebApplicationException.class)) {
-                final WebApplicationException cast = WebApplicationException.class.cast(throwable);
-                final Response response = cast.getResponse();
+            if (throwable instanceof WebApplicationException wae) {
+                Response response = wae.getResponse();
                 if (shouldHandleError(response)) {
                     handleError(response);
+                } else {
+                    exchange.setException(throwable);
                 }
-            } else if (throwable.getClass().isInstance(ResponseProcessingException.class)) {
-                final ResponseProcessingException cast = ResponseProcessingException.class.cast(throwable);
-                final Response response = cast.getResponse();
+            } else if (throwable instanceof ResponseProcessingException rpe) {
+                Response response = rpe.getResponse();
                 if (shouldHandleError(response)) {
                     handleError(response);
+                } else {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Great catch — two real bugs fixed.**

**Bug 1: Reversed `isInstance`** — `throwable.getClass().isInstance(WebApplicationException.class)` always returns `false` because it tests whether the `Class<WebApplicationException>` object is an instance of the throwable's runtime class (which it never is — `Class` is not an exception). This made both `if`/`else if` branches dead code, so every async `WebApplicationException` and `ResponseProcessingException` fell through to the raw `exchange.setException(throwable)` path, bypassing `CxfOperationException` wrapping entirely. The fix correctly uses `instanceof` with pattern matching.

**Bug 2: Silent exception swallowing** — When `shouldHandleError()` returned `false` (e.g., `throwExceptionOnFailure=false` or status ≤ 207), neither `handleError()` nor `exchange.setException()` was called — the failure was silently dropped. The added `else { exchange.setException(throwable); }` ensures the exception always reaches the exchange.

**Code quality:** Modern `instanceof` pattern matching replaces verbose `.getClass().isInstance()` + `.class.cast()` — cleaner and correct. Fix correctly applied to both `CxfInvocationCallback` and `CxfProxyInvocationCallback`.

**Test:** Well-structured — sends an invalid ID that triggers a server-side 500, then asserts the async proxy producer properly maps the `WebApplicationException` to `CxfOperationException` with the correct status code.

LGTM ✅

PR #24925, opened 2026-07-20
CAMEL-24177: camel-cxfrs - Fix contentLanguage latching first request value

davsclaus:

## Summary

Fix `DefaultCxfRsBinding.contentLanguage` instance field that latched the first request's `Content-Language` value and reused it for all subsequent requests, ignoring per-request headers.

- **Root cause:** The "polish up" commit for CAMEL-16460 (`e96225adbabe`) turned a local variable into an instance field. Since `DefaultCxfRsBinding` is shared across all exchanges, the `contentLanguage == null` guard meant only the first request ever read the header.
- **Fix:** Made `contentLanguage` a local variable (per-request) and replaced the `AbstractClient.getState()` reflection hack with the public `webClient.getHeaders()` API, which returns the same data without `setAccessible(true)`.
- Removed unused imports: `AbstractClient`, `ClientState`

## Test plan

- [x] Added `testContentLanguagePerRequest` — two sequential calls on the same binding with `fr` then `de`; verifies each request uses its own Content-Language (previously the second call returned `en`)
- [x] Added `testContentLanguageDefaultsToUS` — verifies fallback to `Locale.US` when no WebClient or no Content-Language header
- [x] All existing `DefaultCxfRsBindingTest` tests pass

### DIFF
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/DefaultCxfRsBinding.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/DefaultCxfRsBinding.java
@@ -45,8 +45,6 @@ import org.apache.camel.support.ExchangeHelper;
 import org.apache.camel.util.ObjectHelper;
 import org.apache.cxf.helpers.CastUtils;
 import org.apache.cxf.helpers.HttpHeaderHelper;
-import org.apache.cxf.jaxrs.client.AbstractClient;
-import org.apache.cxf.jaxrs.client.ClientState;
 import org.apache.cxf.jaxrs.client.WebClient;
 import org.apache.cxf.jaxrs.impl.MetadataMap;
 import org.apache.cxf.jaxrs.model.OperationResourceInfoStack;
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**Clean fix with a nice bonus improvement.**

**Bug:** `contentLanguage` as an instance field on the shared `DefaultCxfRsBinding` meant the first request's Content-Language was latched forever — all subsequent requests reused that value regardless of their own headers.

**Fix is correct on both fronts:**

1. **Local variable** — `contentLanguage` is now per-request, which is the right scope since it's per-exchange state on a shared binding instance.

2. **Reflection elimination** — Replacing `AbstractClient.getState()` via `getDeclaredMethod` + `setAccessible(true)` with the public `webClient.getHeaders()` API is a significant improvement:
   - No more reflection fragility (module system, security manager)
   - Eliminates the try/catch block that silently fell back to US locale on reflection failures
   - Same data, public API — `getHeaders()` returns the client's configured headers

**Code reduction:** The deeply nested 15-line try/catch + reflection block collapses to a clean 4-line check. Net -19 lines of production code.

**Tests:** Both cases are well-chosen:
- `testContentLanguagePerRequest` — Proves the latch is gone: two calls on the same binding with different languages return different locales
- `testContentLanguageDefaultsToUS` — Proves the default path still works when no webClient is provided

LGTM ✅

PR #24928, opened 2026-07-20
CAMEL-24201: camel-aws2-step-functions - listExecutions sets the state machine ARN

oscerd:
## Description

Fixes [CAMEL-24201](https://issues.apache.org/jira/browse/CAMEL-24201). `StepFunctions2Producer.listExecutions` built the request with only `maxResults`:

```java
ListExecutionsRequest.Builder builder = ListExecutionsRequest.builder();
if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(StepFunctions2Constants.EXECUTIONS_MAX_RESULTS))) {
    builder.maxResults(maxRes);
}
```

The AWS `ListExecutions` API **requires** a `stateMachineArn` (or a `mapRunArn`), so in header (non-POJO) mode the call was always rejected by AWS — the operation was effectively unusable. The `CamelAwsStepFunctionsStateMachineArn` header already exists and is read by the sibling operations (`deleteStateMachine`, `updateStateMachine`, `describeStateMachine`, …), but `listExecutions` never read it.

The ARN is now read from that header using the same optional-header pattern as its siblings.

## Tests

Extends the existing `sfnListExecutionsTest` to set the state-machine-ARN header and assert (via the client mock, which now records the request) that it reaches the `ListExecutionsRequest`. All 17 tests in `StepFunctions2ProducerTest` pass.

## Backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-step-functions/src/main/java/org/apache/camel/component/aws2/stepfunctions/StepFunctions2Producer.java
+++ b/components/camel-aws/camel-aws2-step-functions/src/main/java/org/apache/camel/component/aws2/stepfunctions/StepFunctions2Producer.java
@@ -633,6 +633,10 @@ public class StepFunctions2Producer extends DefaultProducer {
             }
         } else {
             ListExecutionsRequest.Builder builder = ListExecutionsRequest.builder();
+            if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(StepFunctions2Constants.STATE_MACHINE_ARN))) {
+                String stateMachineArn = exchange.getIn().getHeader(StepFunctions2Constants.STATE_MACHINE_ARN, String.class);
+                builder.stateMachineArn(stateMachineArn);
+            }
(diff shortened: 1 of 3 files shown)


gnodet (approved):
**Clear and correct fix.** The `listExecutions` non-POJO path was missing the required `stateMachineArn` parameter, making the operation always fail against AWS. The fix reads it from the existing `CamelAwsStepFunctionsStateMachineArn` header, consistent with how all sibling operations (`deleteStateMachine`, `updateStateMachine`, `describeStateMachine`, etc.) read the same header.

**Code:** The 4-line addition follows the exact same pattern used for `maxResults` directly below it, and for `STATE_MACHINE_ARN` in the other operations — consistent and idiomatic.

**Test:** Good approach — capturing the request in the mock with `getLastListExecutionsRequest()` lets the test assert the ARN was actually wired into the request builder, not just that the call didn't throw.

LGTM ✅

davsclaus (approved):

PR #24929, opened 2026-07-20
CAMEL-24179: camel-cxf - Fix continuation timeout race with async processing

davsclaus:

## Summary

Fix race condition in `camel-cxf-soap` (`CxfConsumer`) and `camel-cxf-rest` (`CxfRsInvoker`) where a CXF continuation timeout fires while the Camel async processor is still running, leading to:

1. **Double `continuation.resume()`** — the `done()` callback unconditionally called `resume()` even after timeout already produced a response
2. **Double UoW close** — `UnitOfWorkCloserInterceptor` closes UoW on the timeout response path, then the late callback's resume triggers a second close. `DefaultUnitOfWork.done()` is not idempotent — a double-call fires synchronization callbacks twice, double-removes from inflight registry, and with pooled exchanges recycles the Exchange while the worker still holds it
3. **Concurrent Exchange mutation** — timeout thread sets `ExchangeTimedOutException` on the same Exchange the async worker is still processing

### Fix

Introduce an `AtomicBoolean completed` per-request so exactly one of {timeout, async completion} wins the response and UoW lifecycle:

- **`done()` callback**: guard `continuation.resume()` with `completed.compareAndSet(false, true)`. If timeout already won, the callback closes the UoW (safely, since the worker has finished)
- **Timeout branch**: guard with the same `compareAndSet`. If won, write the timeout response and detach the Camel Exchange from the CXF exchange so `UnitOfWorkCloserInterceptor` won't close UoW prematurely. The late `done()` callback closes it after the worker finishes

### Files changed

- `CxfRsInvoker.java` — REST async invoke race guard
- `CxfConsumer.java` — SOAP async invoke race guard (symmetric change)
- `CxfConsumerContinuationTimeoutTest.java` — add test verifying timeout followed by normal request (no state corruption)

## Test plan

- [x] `CxfConsumerContinuationTimeoutTest` — all 3 tests pass (no-timeout, timeout, timeout-then-normal)
- [x] Full `camel-cxf-soap` module test suite passes (161 tests)
- [x] Full `camel-cxf-rest` module test suite passes
- [x] Full root build (`mvn clean install -DskipTests`) passes with no regenerated file changes

### DIFF
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsInvoker.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsInvoker.java
@@ -17,6 +17,7 @@
 package org.apache.camel.component.cxf.jaxrs;
 
 import java.lang.reflect.Method;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.HttpHeaders;
@@ -43,6 +44,7 @@ import org.slf4j.LoggerFactory;
 public class CxfRsInvoker extends JAXRSInvoker {
     private static final Logger LOG = LoggerFactory.getLogger(CxfRsInvoker.class);
     private static final String SUSPENDED = "org.apache.camel.component.cxf.jaxrs.suspend";
+    private static final String COMPLETED = "org.apache.camel.component.cxf.jaxrs.completed";
     private final CxfRsConsumer cxfRsConsumer;
     private final CxfRsEndpoint endpoint;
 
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Excellent concurrency fix — the `AtomicBoolean completed` with `compareAndSet(false, true)` is the textbook solution for this two-party race.

**What I verified:**

1. **UoW lifecycle is correct in all paths:**
   - **Async completion wins**: `done()` callback resumes continuation → `UnitOfWorkCloserInterceptor` closes UoW via the Exchange attached to cxfExchange _(normal path, unchanged)_
   - **Timeout wins**: timeout sets response, detaches Exchange from cxfExchange via `cxfExchange.put(org.apache.camel.Exchange.class, null)`, so `UnitOfWorkCloserInterceptor.handleMessage()` → `CxfUtils.closeCamelUnitOfWork()` safely skips (already null-guards the Exchange lookup). The late `done()` callback then calls `doneUoW(camelExchange)` directly

2. **Null-safety confirmed**: `CxfUtils.closeCamelUnitOfWork` checks `exchange != null` before calling `UnitOfWorkHelper.doneUow`, so the detachment is safe

3. **Error path coverage**: Both REST and SOAP timeout branches properly detach the Exchange in the catch block before calling `doneUoW` and re-throwing

4. **Symmetric implementation**: REST (`CxfRsInvoker`) and SOAP (`CxfConsumer.CxfConsumerInvoker`) changes are structurally identical, with appropriate namespace differences in the COMPLETED constant

5. **Defensive null check**: `completed != null && completed.compareAndSet(...)` in the timeout branch handles edge cases gracefully

6. **Test validates state corruption prevention**: `testTimeoutThenNormalRequest` confirms a timed-out request doesn't corrupt state for subsequent requests — this was the user-visible symptom of the double-resume/double-UoW-close race

One minor observation (non-blocking): the concurrent Exchange mutation after timeout wins (async worker still processing while timeout sets `ExchangeTimedOutException`) is inherent to the continuation model and can't be fully eliminated without cancellation support. The fix correctly addresses the actionable symptoms (double resume, double UoW close).

PR #24930, opened 2026-07-20
CAMEL-24180: Fix CxfMessageHelper fallback setting null instead of actual body

davsclaus:

## Summary

- Fix copy-paste bug in `CxfMessageHelper.getCxfInMessage()` where the fallback branch (line 63) passed the null `body` variable instead of `message.getBody()`, causing non-InputStream message bodies (e.g. POJOs) to arrive as null content in downstream CXF interceptors
- Add test coverage for the fallback branch with POJO and null bodies

## Details

The bug has been dormant since 2011 because most Camel bodies have an `InputStream` type converter. The `else if` branch is only reachable when `body` (the `InputStream` variable) is null, so `answer.setContent(Object.class, body)` always set null. The fix changes it to `answer.setContent(Object.class, message.getBody())`.

## Test plan

- [x] `testGetCxfInMessageWithPojoBody` — verifies a `List` body (no InputStream converter) is preserved via the fallback
- [x] `testGetCxfInMessageWithNullBody` — verifies null body produces no content (no NPE)
- [x] Existing `testGetCxfInMessage` — verifies String/DOMSource/File bodies still work via InputStream path

### DIFF
--- a/components/camel-cxf/camel-cxf-transport/src/main/java/org/apache/camel/component/cxf/transport/message/CxfMessageHelper.java
+++ b/components/camel-cxf/camel-cxf-transport/src/main/java/org/apache/camel/component/cxf/transport/message/CxfMessageHelper.java
@@ -60,7 +60,7 @@ public final class CxfMessageHelper {
             answer.setContent(InputStream.class, body);
         } else if (message.getBody() != null) {
             // fallback and set the body as what it is
-            answer.setContent(Object.class, body);
+            answer.setContent(Object.class, message.getBody());
         }
 
         answer.putAll(message.getHeaders());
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-diagnosed fix for a long-dormant copy-paste bug.

**The bug**: `body` is the `InputStream` variable from `message.getBody(InputStream.class)`, which is guaranteed null in the `else if` branch (that's why we're in the `else if`). So `answer.setContent(Object.class, body)` always set null.

**The fix**: `message.getBody()` correctly retrieves the original body. One-line change, exactly right.

**Tests**: Good coverage of both the POJO fallback path (`assertSame` confirms identity, not just equality) and the null-body edge case (no NPE, no spurious content).

As the PR description notes, this has been dormant since 2011 because most Camel bodies have an `InputStream` type converter — the fallback is only reached for types without one (POJOs, custom objects). Low risk, high correctness improvement.

PR #24931, opened 2026-07-20
CAMEL-24181: Do not shut down shared CXF bus when stopping CxfSpringEndpoint

davsclaus:

## Summary

- Remove `createBus = true` from `CxfSpringEndpoint.setApplicationContext()` — the bus retrieved via `BusWiringBeanFactoryPostProcessor.addDefaultBus(ctx)` is the shared application-wide CXF bus, not an endpoint-owned bus. Its lifecycle belongs to Spring (`SpringBus` shuts down on `ContextClosedEvent`).
- Previously, every `<cxf:cxfEndpoint>` bean marked itself as the bus owner, so stopping any single endpoint would call `getBus().shutdown(false)` on the shared bus, killing all other CXF endpoints/clients in the application.
- Added by CAMEL-12829 to fix a bus leak, but didn't distinguish between a genuinely auto-created bus (via `getBus()`) and a shared context bus (via `setApplicationContext()`).

## Test plan

- [x] `testStopEndpointDoesNotShutdownSharedBus` — stops an endpoint and verifies the shared bus is still RUNNING
- [x] `testTwoEndpointsShareSameBus` — verifies two endpoints share the same bus and stopping one doesn't affect the other
- [x] All 160 existing tests in camel-cxf-spring-soap pass

### DIFF
--- a/components/camel-cxf/camel-cxf-spring-soap/src/main/java/org/apache/camel/component/cxf/spring/jaxws/CxfSpringEndpoint.java
+++ b/components/camel-cxf/camel-cxf-spring-soap/src/main/java/org/apache/camel/component/cxf/spring/jaxws/CxfSpringEndpoint.java
@@ -273,7 +273,6 @@ public class CxfSpringEndpoint extends CxfEndpoint implements ApplicationContext
         applicationContext = ctx;
 
         if (bus == null) {
-            createBus = true;
             bus = BusWiringBeanFactoryPostProcessor.addDefaultBus(ctx);
         }
     }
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean fix with clear root-cause analysis. Verified the mechanics:

**How `createBus` works**: `CxfEndpoint.doStop()` (line 1175) checks `if (createBus && bus != null)` and calls `getBus().shutdown(false)`. The flag means "this endpoint created and owns this bus."

**The bug**: `setApplicationContext()` set `createBus = true`, but the bus obtained from `BusWiringBeanFactoryPostProcessor.addDefaultBus(ctx)` is the shared application-wide Spring bus — its lifecycle is managed by `SpringBus` on `ContextClosedEvent`. Every `<cxf:cxfEndpoint>` bean incorrectly claimed ownership, so stopping any single endpoint shut down the bus for all others.

**The fix**: Remove `createBus = true` — one line, exactly correct. The `getBus()` path (lines 1023-1038) still sets `createBus = true` for genuinely auto-created buses, so that case is unaffected.

**Tests**: Both test cases directly verify the fix scenario — stopping an endpoint leaves the shared bus in RUNNING/INITIAL state. Using `assertSame` to confirm both endpoints share the same bus instance is a nice touch.

Good historical context in the PR description about CAMEL-12829 introducing the regression while fixing a bus leak.

PR #24932, opened 2026-07-20
CAMEL-24182: Fix SpringBusFactoryBean to shut down bus on context close

davsclaus:

## Summary

- Implement `DisposableBean` on `SpringBusFactoryBean` so Spring calls `destroy()` on context close, which calls `bus.shutdown(true)` to release the `BusApplicationContext`, workqueue threads, and classloader references
- Store the created `Bus` as a field (was previously returned as a local and lost), cache it for singleton semantics
- Add null-guard in `setCfgFiles` to prevent NPE on null input

Without this fix, the bus, its internal Spring context, and thread pools leak on every context close/refresh (servlet redeploy, Spring Boot devtools restart, repeated integration tests).

## Test plan

- [x] `testBusShutdownOnContextClose` — verifies bus state transitions to SHUTDOWN after context close
- [x] Existing `getTheBusInstance` test still passes
- [x] All 160 tests in camel-cxf-spring-soap pass

### DIFF
--- a/components/camel-cxf/camel-cxf-spring-common/src/main/java/org/apache/camel/component/cxf/spring/SpringBusFactoryBean.java
+++ b/components/camel-cxf/camel-cxf-spring-common/src/main/java/org/apache/camel/component/cxf/spring/SpringBusFactoryBean.java
@@ -18,6 +18,7 @@ package org.apache.camel.component.cxf.spring;
 
 import org.apache.cxf.Bus;
 import org.apache.cxf.bus.spring.SpringBusFactory;
+import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.SmartFactoryBean;
 
 /**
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Excellent companion to #24931 — together they complete the correct CXF bus lifecycle:
- #24931: endpoints don't kill the shared bus on stop
- #24932: the factory bean properly shuts down the bus on context close

**Changes verified:**

1. **`DisposableBean` implementation**: `destroy()` calls `bus.shutdown(true)` (wait for in-progress work) and nulls the reference — correct for context close cleanup. Prevents double-shutdown.

2. **Singleton caching**: Storing the bus as a field aligns with `isSingleton()` returning `true`. Previously `getObject()` created a new bus on every call (Spring caches the singleton, but the factory lost its own reference, so it couldn't shut it down). The `if (bus == null)` lazy-init is belt-and-suspenders since Spring calls `getObject()` once for singletons, but it's harmless and defensive.

3. **Null-guard in `setCfgFiles`**: Prevents NPE if called with null — minor but sensible defensive improvement.

4. **Test**: `testBusShutdownOnContextClose` directly verifies RUNNING → context.close() → SHUTDOWN. Setting `ctx = null` to prevent double-close in tearDown is good test hygiene.

The leak scenario described (servlet redeploy, devtools restart, repeated integration tests) is real and insidious — thread pools and classloader references from the leaked `BusApplicationContext` can cause classloader leaks in web container environments.

PR #24934, opened 2026-07-20
CAMEL-24183: Fix Spring rsClient producer to apply endpoint URI options

davsclaus:

## Summary

- Fix `CxfRsSpringEndpoint.setupJAXRSClientFactoryBean` to call `setupCommonFactoryProperties(cfb)`, matching the parent `CxfRsEndpoint` implementation
- Previously, the Spring override completely replaced the parent method without applying providers, features, interceptors, properties, logging feature, or skip-fault-logging — all silently ignored for Spring-based cxfrs producers
- Also adds the missing `modelRef` and `resourceClasses` setup from the parent
- Spring bean config is applied first (`configurer.configureBean`) so URI options correctly override XML config

## Details

The ordering now matches the parent: Spring configurer -> modelRef/resourceClasses -> common properties -> threadSafe -> endpoint configurer -> address. The server side was already correct (uses inherited `setupJAXRSServerFactoryBean`).

## Test plan

- [x] All existing tests in camel-cxf-spring-rest pass

### DIFF
--- a/components/camel-cxf/camel-cxf-spring-rest/src/main/java/org/apache/camel/component/cxf/spring/jaxrs/CxfRsSpringEndpoint.java
+++ b/components/camel-cxf/camel-cxf-spring-rest/src/main/java/org/apache/camel/component/cxf/spring/jaxrs/CxfRsSpringEndpoint.java
@@ -66,11 +66,21 @@ public class CxfRsSpringEndpoint extends CxfRsEndpoint implements BeanIdAware {
 
     @Override
     protected void setupJAXRSClientFactoryBean(JAXRSClientFactoryBean cfb, String address) {
+        // apply Spring bean config first so URI options can override
         configurer.configureBean(beanId, cfb);
-        // support to call the configurer here
-        getNullSafeCxfRsEndpointConfigurer().configure(cfb);
-        cfb.setAddress(address);
+        if (getModelRef() != null) {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Verified against the parent `CxfRsEndpoint.setupJAXRSClientFactoryBean` — the Spring override now matches the parent's structure exactly:

**Parent order**: modelRef → resourceClasses → `setupCommonFactoryProperties(cfb)` → threadSafe → endpointConfigurer → address

**Spring override (after fix)**: Spring configurer → modelRef → resourceClasses → `setupCommonFactoryProperties(cfb)` → threadSafe → endpointConfigurer → address

The Spring configurer correctly runs first so URI options (applied via `setupCommonFactoryProperties` which sets providers, features, interceptors, properties, logging, etc.) can override XML bean config. Address is set last since it can be overridden by message headers.

**What was missing before**: The old implementation only called `configurer.configureBean`, `getNullSafeCxfRsEndpointConfigurer().configure(cfb)`, and `cfb.setAddress(address)` — completely skipping `setupCommonFactoryProperties`, `modelRef`, and `resourceClasses`. This meant providers, features, interceptors, properties, loggingFeature, skipFaultLogging, and publishedEndpointUrl were all silently ignored for Spring-based CXF-RS producers.

The null-guard on `address` also matches the parent's pattern (address can be overridden by message header, so null is valid).

PR #24940, opened 2026-07-20
CAMEL-24203: camel-aws-security-hub - implement the declared getFindingAggregator operation

oscerd:
## Description

Fixes [CAMEL-24203](https://issues.apache.org/jira/browse/CAMEL-24203). `SecurityHubOperations` declares `getFindingAggregator`, but `SecurityHubProducer.process` had no `case` for it, so the switch fell through to:

```java
default:
    throw new IllegalArgumentException("Unsupported operation");
```

Selecting `operation=getFindingAggregator` therefore always failed, even though the enum and catalog advertise it.

## Fix

- Implements the operation using the module's existing `executeOperation` helper — POJO `GetFindingAggregatorRequest` support plus a header-driven path.
- Adds the `CamelAwsSecurityHubFindingAggregatorArn` header constant the AWS `GetFindingAggregator` API requires (it is mandatory, so the header path fails fast with a clear message when absent).

Chose to implement rather than drop the enum value, since removing a published operation would be a breaking API change.

## Tests

Adds `SecurityHubGetFindingAggregatorTest` — drives the operation through a route with a mocked `SecurityHubClient` and asserts the ARN header reaches the request. Against the old code the exchange fails with "Unsupported operation".

Regenerated component/catalog/endpoint-DSL metadata for the new header is included.

## Backport

---


gnodet (approved):
Clean fix — the `getFindingAggregator` enum value was declared in `SecurityHubOperations` but had no `case` in the producer's switch, so the operation always threw `IllegalArgumentException`. This PR implements it properly.

**What I checked:**

- **Pattern consistency**: the new `getFindingAggregator` method follows the exact same `executeOperation` helper pattern as `getFindingHistory` and other operations — POJO body support plus a header-driven fallback path.
- **Header constant**: `FINDING_AGGREGATOR_ARN` follows the established naming convention (`CamelAwsSecurityHub*`) and is correctly annotated with `@Metadata`.
- **Validation**: the ARN is validated as non-empty before use, consistent with how other required headers are validated (e.g., `getFindingHistory` validates `findingId` and `productArn`).
- **Test**: `SecurityHubGetFindingAggregatorTest` uses a Mockito mock for `SecurityHubClient`, verifies the ARN header is correctly propagated to the `GetFindingAggregatorRequest` via `ArgumentCaptor`. Good coverage for the fix.
- **Generated files**: catalog JSON and endpoint DSL factory correctly include the new header at index 21.
- **API compatibility**: implementing the existing enum value (rather than removing it) is the right call — removing a published operation would be a breaking change.

LGTM ✅

davsclaus (approved):
Clean fix for a declared-but-unimplemented operation. The `getFindingAggregator` enum value has existed since the component was introduced (CAMEL-22919), but the producer switch had no case for it — this PR fills that gap.

**What looks good:**
- Implementation follows the exact same `executeOperation` helper pattern used by all other operations in `SecurityHubProducer`
- New `FINDING_AGGREGATOR_ARN` header constant properly annotated with `@Metadata`
- Test uses Mockito to mock the `SecurityHubClient`, drives the operation through a Camel route, and asserts the ARN header reaches the AWS request — effectively proving the fix
- Catalog, component JSON, and endpoint-DSL metadata regenerated correctly
- `mockito-junit-jupiter` test dependency already used by several other `camel-aws` modules

LGTM.

PR #24941, opened 2026-07-20
CAMEL-24184: Fix 19 medium-severity findings from camel-cxf code review

davsclaus:
## Summary

Fixes all 19 medium-severity findings from the July 2026 AI-assisted code review of the `components/camel-cxf/` component family ([CAMEL-24184](https://issues.apache.org/jira/browse/CAMEL-24184)).

The high-severity findings (CAMEL-24176 through CAMEL-24183) were already fixed in separate PRs. This PR addresses the remaining medium-severity items across 6 CXF sub-modules (15 files changed).

### camel-cxf-soap / camel-cxf-common
- **#1** `replaceMultiPartContentType` double-strips `type=` prefix, corrupting Content-Type for unquoted MTOM/multipart
- **#2** `CxfClientCallback` NPE when out message is absent during failover
- **#3** `populateCxfHeaderFromCamelExchangeBeforeCheckError` protocol headers never reach fault response (dead code)
- **#4** `CxfEndpoint.bus` field made `volatile` to prevent visibility issues in lazy init

### camel-cxf-rest
- **#5** `CxfRsInvoker` checks `Void.class` instead of `Void.TYPE` for `void` methods — InOnly MEP never selected
- **#6** `SubResourceClassInvocationHandler` same `Void.class` vs `Void.TYPE` confusion — NPE on void methods
- **#7** Async `completed()` callbacks leak JAX-RS `Response` for InOnly exchanges (connection pinned until GC)
- **#8** Multi-valued Camel headers flattened to literal `"[a, b]"` via `toString()` in `CxfHeaderHelper`
- **#9** Inbound protocol headers truncated to first value (`get(0)`) in 3 locations
- **#10** `setProtocolHeaders` NPE when CXF out message is absent (oneway/fault paths)
- **#11** `CxfConverter` fallback NPE when `Response.getEntity()` returns null (204/empty responses)

### camel-cxf-transport
- **#12** `CamelConduit` producer leak on abandoned conduit + use-after-stop race
- **#13** `asyncInvokeFromWorkQueue` NPE when no fault observer is registered; unsafe `PhaseInterceptorChain` cast
- **#14** `getBasePath` regex: second `replaceFirst` unreachable; combined into single pattern

### camel-cxf-spring-*
- **#15-16** `enableSpringBusShutdownGracefully` was dead code (never called in standard path), spawned a thread for every `ApplicationEvent`, and matched listeners unreliably. Fixed: call from `setApplicationContext`, match by identity, only spawn thread for `ContextClosedEvent`
- **#17** `performInvocation` not propagated to `SpringJAXRSServerFactoryBean` — interface-model resources not filtered in Spring variant
- **#18** `SpringJAXRSServerFactoryBean` missing `setProperties` merge override — `<cxf:properties>` wipes `NullFaultListener`
- **#19** `CxfRsSpringEndpoint` hard-casts `CamelContext` to `SpringCamelContext` — `ClassCastException` in camel-main

## Test plan
- [x] `mvn test` in camel-cxf-common — passed
- [x] `mvn test` in camel-cxf-soap — passed
- [x] `mvn test` in camel-cxf-rest — passed
- [x] `mvn test` in camel-cxf-transport — passed
- [x] `mvn test` in camel-cxf-spring-soap — passed
- [x] `mvn test` in camel-cxf-spring-rest — passed

### DIFF
--- a/components/camel-cxf/camel-cxf-common/src/main/java/org/apache/camel/component/cxf/common/header/CxfHeaderHelper.java
+++ b/components/camel-cxf/camel-cxf-common/src/main/java/org/apache/camel/component/cxf/common/header/CxfHeaderHelper.java
@@ -104,7 +104,12 @@ public final class CxfHeaderHelper {
 
             LOG.trace("Propagate Camel header: {}={} as {}", entry.getKey(), entry.getValue(), cxfHeaderName);
 
-            requestHeaders.put(cxfHeaderName, Arrays.asList(entry.getValue().toString()));
+            Object values = entry.getValue();
+            if (values instanceof List<?>) {
+                requestHeaders.put(cxfHeaderName, CastUtils.cast((List<?>) values, String.class));
+            } else {
+                requestHeaders.put(cxfHeaderName, Arrays.asList(values.toString()));
+            }
         });
     }
 
@@ -191,8 +196,8 @@ public final class CxfHeaderHelper {
 
             LOG.trace("Populate external header: {}={} as {}", entry.getKey(), entry.getValue(), camelHeaderName);
             if (!camelHeaderName.startsWith(":")) {
-                ///* Ignore HTTP/2 pseudo headers such as :status */
-                camelHeaders.put(camelHeaderName, entry.getValue().get(0));
+                List<Object> values = entry.getValue();
+                camelHeaders.put(camelHeaderName, values.size() == 1 ? values.get(0) : values);
             }
         });
     }
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfConverter.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfConverter.java
@@ -223,6 +223,10 @@ public final class CxfConverter {
             Response response = (Response) value;
             Object entity = response.getEntity();
 
+            if (entity == null) {
+                return (T) MISS_VALUE;
+            }
+
             TypeConverter tc = registry.lookup(type, entity.getClass());
             if (tc != null) {
                 return tc.convertTo(type, exchange, entity);
--- a/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsInvoker.java
+++ b/components/camel-cxf/camel-cxf-rest/src/main/java/org/apache/camel/component/cxf/jaxrs/CxfRsInvoker.java
@@ -182,7 +182,7 @@ public class CxfRsInvoker extends JAXRSInvoker {
             Exchange cxfExchange, Method method,
             Object[] paramArray, Object response) {
         ExchangePattern ep = ExchangePattern.InOut;
-        if (method.getReturnType() == Void.class) {
+        if (method.getReturnType() == Void.TYPE || method.getReturnType() == Void.class) {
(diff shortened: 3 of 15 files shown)


gnodet (approved):
Thorough analysis of 19 medium-severity CXF fixes — each one targeted and well-reasoned. This follows the high-severity fixes already merged in PRs #24929–#24934. Verified each fix against the existing codebase patterns:

**Correctness bugs (critical fixes):**

- **#1 `replaceMultiPartContentType`**: After `part.substring(5)` strips `type=`, the else branch did `part.substring(5)` *again* — double-stripping turned `text/xml` → `xml`. Fix to `result = part` is correct.
- **#5 & #6 Void.class vs Void.TYPE**: Classic Java reflection bug. `Method.getReturnType()` returns `void.class` (i.e., `Void.TYPE`), never `Void.class`. The checks never matched, so InOnly MEP was never selected for `void` methods. Fix correctly checks both.
- **#11 CxfConverter entity null**: `Response.getEntity()` returns null for 204/empty responses → NPE in the fallback converter. Returning `MISS_VALUE` is correct.

**NPE guards (defensive fixes):**

- **#2 CxfClientCallback**: `getOutMessage()` can be null during failover — guard added.
- **#10 setProtocolHeaders**: Rewrites the guard to early-return instead of falling through to use `getOutMessage()` unconditionally later.
- **#13 asyncInvokeFromWorkQueue**: `instanceof PhaseInterceptorChain` instead of hard cast, plus null guard on fault observer. The LOG.error fallback is appropriate — silent swallowing would be worse.

**Resource leaks:**

- **#7 InOnly Response leak**: When `isOutCapable()` is false, the Response was never closed — connection stays pinned until GC. Fix correctly closes it.
- **#12 CamelConduit use-after-stop**: `volatile boolean closed` flag + check in `prepare()` prevents use after `close()`. Simple and effective.

**Header handling (data-loss fixes):**

- **#8 Multi-valued Camel→CXF**: `toString()` on a List produced `"[a, b]"` literal. Fix preserves List values via `CastUtils.cast`.
- **#9 CXF→Camel truncation**: `get(0)` drops all values after the first. Fix preserves multi-valued headers as List when `size() > 1` — good backward-compatible approach.
- **#3 Protocol headers on fault path**: Headers stored only on the out-message (which doesn't exist yet in the fault path). Fix also stores on the CXF exchange so headers survive into fault handling.

**Thread safety and lifecycle:**

- **#4 `bus` volatile**: Prevents visibility issues in lazy init (the `getBus()` method creates a bus if null).
- **#15-16 enableSpringBusShutdownGracefully**: Three improvements: (a) now called from `setApplicationContext()` for shared bus case (was only called from `getBus()` for created bus), (b) identity-based listener matching instead of class-name string search, (c) thread only spawned for `ContextClosedEvent` instead of every `ApplicationEvent`.

**Spring integration:**

- **#17 performInvocation**: Propagated to `SpringJAXRSServerFactoryBean` via override.
- **#18 setProperties merge**: `putAll` instead of replace preserves `NullFaultListener` from constructor.
- **#19 SpringCamelContext cast**: `instanceof` pattern matching prevents ClassCastException in camel-main, with null-safe `configurer` usage downstream.

**Code cleanup:**

- **#14 getBasePath regex**: Combined two `replaceFirst` calls (second was unreachable) into a single regex with capture group. Clean.

All fixes follow existing patterns and are well-scoped. This complements the high-severity CXF fixes in #24929–#24934 nicely.

LGTM ✅

PR #24946, opened 2026-07-20
CAMEL-24170: Fix remaining HIGH findings from data format review

davsclaus:
## Summary

Fixes 6 remaining HIGH findings from the July 2026 data format review ([CAMEL-24170](https://issues.apache.org/jira/browse/CAMEL-24170)), excluding H9/H10 (univocity lazy-load architectural issues, to be addressed separately).

### Fixes

- **H2** — `JacksonXMLDataFormatReifier`: `xmlMapper` property missing `asRef()`, causing `NoTypeConversionAvailableException` when referencing a registry bean by name without `#` prefix
- **H11** — `BindyKeyValuePairFactory`: `split(separator)` without limit truncates values containing the key-value separator (e.g. FIX tag `58=a=b=c` → `a` instead of `a=b=c`). Fixed with `split(separator, 2)`
- **H13** — `UnicodeHelper.indexOf`: off-by-one in loop condition (`<` → `<=`) misses delimiter at end-of-record, causing `IndexOutOfBoundsException` from `substring(0, -1)`
- **H15** — `ProtobufDataFormat`: JSON unmarshal uses `new InputStreamReader(inputStream)` (platform default charset) while marshal writes UTF-8. Fixed to explicitly use `StandardCharsets.UTF_8`
- **H16** — `AvroDataFormat` model: missing `implements ContentTypeHeaderAware`, so `contentTypeHeader=false` option is silently ignored by the generic reifier
- **H17** — `YAMLDataFormat.Builder.typeFilter(Class...)`: `StringJoiner(".")` joins class names with `.` instead of `,`, producing a bogus single pattern that matches nothing (CAMEL-22354 regression)

## Test plan

- [x] New test: `UnicodeHelperTest.testIndexOfAtEndOfString` — verifies delimiter found at end-of-string (H13)
- [x] New test: `BindyKeyValuePairSeparatorInValueTest` — verifies KVP unmarshal preserves values containing `=` (H11)
- [x] New test: `YAMLDataFormatTest` — verifies `typeFilter(Class...)` joins with `,` (H17)
- [x] Full `camel-bindy` test suite passes
- [x] Full `camel-protobuf` test suite passes

### DIFF
--- a/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/BindyKeyValuePairFactory.java
+++ b/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/BindyKeyValuePairFactory.java
@@ -141,7 +141,7 @@ public class BindyKeyValuePairFactory extends BindyAbstractFactory implements Bi
         for (String s : data) {
 
             // Get KeyValuePair
-            String[] keyValuePair = s.split(getKeyValuePairSeparator());
+            String[] keyValuePair = s.split(getKeyValuePairSeparator(), 2);
 
             // Extract only if value is populated in key:value pair in incoming message.
             if (keyValuePair.length > 1) {
--- a/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/UnicodeHelper.java
+++ b/components/camel-bindy/src/main/java/org/apache/camel/dataformat/bindy/UnicodeHelper.java
@@ -124,7 +124,7 @@ public class UnicodeHelper implements Serializable {
 
         final int len = new UnicodeHelper(str, method).length();
 
-        for (int index = fromIndex; index + len < length(); index++) {
+        for (int index = fromIndex; index + len <= length(); index++) {
             if (str.equals(input.substring(splitted.get(index), splitted.get(index + len)))) {
                 return index;
             }
--- a/components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/UnicodeHelperTest.java
+++ b/components/camel-bindy/src/test/java/org/apache/camel/dataformat/bindy/UnicodeHelperTest.java
@@ -206,6 +206,21 @@ public class UnicodeHelperTest {
         assertEquals(3, lh3.indexOf("m̂", 2));
     }
 
+    @Test
+    public void testIndexOfAtEndOfString() {
+        UnicodeHelper lh = new UnicodeHelper("ab|", Method.CODEPOINTS);
+        assertEquals(2, lh.indexOf("|"));
+
+        UnicodeHelper lh2 = new UnicodeHelper("abc|", Method.CODEPOINTS);
+        assertEquals(3, lh2.indexOf("|"));
+
+        UnicodeHelper lh3 = new UnicodeHelper("|", Method.CODEPOINTS);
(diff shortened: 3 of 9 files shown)


gnodet (approved):
Six well-targeted HIGH fixes across data format components. Each is clean, well-scoped, and appropriately tested.

**Fix analysis:**

- **H2 — JacksonXMLDataFormatReifier `asRef()`**: Without `asRef()`, a bean name like `myMapper` is passed as a raw string instead of being resolved from the registry. Adding `asRef(definition.getXmlMapper())` is consistent with how other reifiers handle bean references (e.g., the Jackson JSON reifier already does this).

- **H11 — BindyKeyValuePairFactory `split(sep, 2)`**: Classic split-without-limit bug. FIX tag `58=a=b=c` splits into `["58","a","b","c"]` → `keyValuePair[1]` gives `"a"` instead of `"a=b=c"`. The `split(separator, 2)` fix is the standard solution. Good test with a real FIX message scenario.

- **H13 — UnicodeHelper.indexOf off-by-one**: `index + len < length()` → `<= length()`. The strict `<` skips the last valid position when the delimiter sits at end-of-string, causing `indexOf` to return -1, then `substring(0, -1)` throws IOOBE. Test covers both CODEPOINTS and GRAPHEME methods.

- **H15 — ProtobufDataFormat charset**: `new InputStreamReader(inputStream)` uses platform default; marshal writes UTF-8. On non-UTF-8 locales, multi-byte JSON content corrupts silently. Explicit `StandardCharsets.UTF_8` is the standard fix.

- **H16 — AvroDataFormat `ContentTypeHeaderAware`**: Without implementing this marker interface, the generic reifier ignores the `contentTypeHeader` option. Other data formats (Jackson, Protobuf) that support this option already implement it — Avro was just missing.

- **H17 — YAML typeFilter `StringJoiner(".")`**: `"."` separator produces `java.lang.String.java.lang.Integer` which looks like one (nonexistent) class name. Fix to `","` correctly produces `java.lang.String,java.lang.Integer`. Clean regression from CAMEL-22354. Test verifies single-class, multi-class, and string input paths.

LGTM ✅

gnodet (approved):
Re-reviewed after force-push — the PR content changed completely from the BigDecimal precision fixes to 6 remaining HIGH findings from the data format review. All changes reviewed:

**H2** (`JacksonXMLDataFormatReifier`): `asRef()` wrapper for `xmlMapper` — correct, follows the pattern of other reifiers.

**H11** (`BindyKeyValuePairFactory`): `split(separator, 2)` — correctly preserves values containing the key-value separator. Good test covering the FIX protocol scenario.

**H13** (`UnicodeHelper.indexOf`): off-by-one fix `<` → `<=` — allows finding delimiter at end-of-string. Test confirms the fix.

**H15** (`ProtobufDataFormat`): explicit `StandardCharsets.UTF_8` — correct for consistency with the marshal path.

**H16** (`AvroDataFormat`): `implements ContentTypeHeaderAware` — straightforward interface addition.

**H17** (`YAMLDataFormat.Builder.typeFilter`): `StringJoiner(".")` → `StringJoiner(",")` — clear bug fix with comprehensive tests.

All fixes are minimal, targeted, and well-covered by new tests. LGTM.

PR #24954, opened 2026-07-20
Fix CxfTimeoutTest flakiness: check cause chain for HttpTimeoutException

gnodet:
## Summary

Fix flaky `CxfTimeoutTest` in `camel-cxf-spring-soap` by checking the full exception cause chain for `HttpTimeoutException` instead of only the top-level exception.

## Root Cause Analysis

The test configures a 100ms `ReceiveTimeout` via `*.http-conduit` XML configuration and expects a `java.net.http.HttpTimeoutException` when calling a server that sleeps 2000ms. The timeout **is** applied correctly — the flakiness is in how CXF propagates the exception.

CXF's interceptor chain may either:
1. **Propagate `HttpTimeoutException` directly** — test passes
2. **Wrap it in `org.apache.cxf.interceptor.Fault: Could not send Message.`** with `HttpTimeoutException` as the cause — test fails

Which path occurs depends on timing within CXF's interceptor chain. The original assertion only checked `e instanceof HttpTimeoutException` on the top-level exception, so case 2 caused assertion failures.

**Develocity CI data (269 runs, 10-day window):**
- `testInvokingFromCamelRoute`: 7 failures (2.6%)
- `testInvokingJaxWsServerWithCxfEndpoint`: 1 failure (0.37%)
- `testInvokingJaxWsServerWithBusUriParams`: 1 failure (0.37%)
- Methods without SSL params or with explicit 60s timeout: 0 failures

## Changes

- **`CxfTimeoutTest.java`**: Modified `sendTimeOutMessage()` to accept `HttpTimeoutException` at any level in the cause chain using AssertJ's `satisfiesAnyOf`/`hasRootCauseInstanceOf`. Also migrated other assertions to AssertJ style (aligned with project OpenRewrite rules).
- **`pom.xml`**: Added `assertj-core` test dependency (version-managed in parent POM).

## Test plan

- [x] Ran `CxfTimeoutTest` 10 consecutive times locally — all passed (0 failures)
- [ ] CI build passes

### DIFF
--- a/components/camel-cxf/camel-cxf-spring-soap/src/test/java/org/apache/camel/component/cxf/CxfTimeoutTest.java
+++ b/components/camel-cxf/camel-cxf-spring-soap/src/test/java/org/apache/camel/component/cxf/CxfTimeoutTest.java
@@ -40,10 +40,7 @@ import org.junit.jupiter.api.parallel.Isolated;
 import org.springframework.context.support.AbstractXmlApplicationContext;
 import org.springframework.context.support.ClassPathXmlApplicationContext;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
 
 @Isolated
 public class CxfTimeoutTest extends CamelSpringTestSupport {
@@ -78,8 +75,8 @@ public class CxfTimeoutTest extends CamelSpringTestSupport {
     public void testInvokingJaxWsServerWithCxfEndpointWithConfigurer() throws Exception {
         Exchange reply = sendJaxWsMessage("cxf://bean:springEndpoint?cxfConfigurer=#myConfigurer");
         // we don't expect the exception here
-        assertFalse(reply.isFailed(), "We don't expect the exception here");
-        assertEquals("Greet Hello World!", reply.getMessage().getBody(String.class), "Get a wrong response");
+        assertThat(reply.isFailed()).as("We don't expect the exception here").isFalse();
+        assertThat(reply.getMessage().getBody(String.class)).as("response body").isEqualTo("Greet Hello World!");
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Well-investigated flaky test fix — the root cause analysis is thorough and the fix is clean.

**What works well:**
- The Develocity CI data quantifying the flake rate (2.6% on \`testInvokingFromCamelRoute\`) gives confidence this is a real intermittent, not a persistent failure.
- The root cause is clearly identified: CXF's interceptor chain non-deterministically wraps \`HttpTimeoutException\` in a \`Fault\` depending on timing — both paths correctly indicate the timeout was enforced.
- The \`hasHttpTimeoutException()\` helper is clean and correct — simple cause-chain walk with no risk of infinite loops (Java exception chains are acyclic by convention, and the \`getCause()\` null-terminates).
- Improved assertion message now includes the actual exception class and message, which makes any future failure immediately diagnosable.
- Good inline comment explaining why the cause chain check is needed.

LGTM ✅

components/camel-cxf/camel-cxf-spring-soap/src/test/java/org/apache/camel/component/cxf/CxfTimeoutTest.java
```diff
@@ -96,7 +96,23 @@ protected void sendTimeOutMessage(String endpointUri) throws Exception {
...
+    }
+
+    private static boolean hasHttpTimeoutException(Throwable t) {
+        while (t != null) {
+            if (t instanceof HttpTimeoutException) {
+                return true;
+            }
+            t = t.getCause();
+        }
+        return false;
```
apupier: Can use assertj assertThat().hasCause() directly

apupier (changes requested):

gnodet (commented):
New commit `104a8d5dc943` reviewed — clean improvements on both fronts.

**CxfTimeoutTest AssertJ migration:**
The `satisfiesAnyOf` pattern is a great replacement for the manual `hasHttpTimeoutException()` cause-chain walker:
```java
assertThat(e).satisfiesAnyOf(
    ex -> assertThat(ex).isInstanceOf(HttpTimeoutException.class),
    ex -> assertThat(ex).hasRootCauseInstanceOf(HttpTimeoutException.class));
```
This is more readable, provides better failure diagnostics, and `hasRootCauseInstanceOf` already handles the full cause-chain traversal that the manual loop was doing. Net reduction of ~13 lines while gaining expressiveness. 

**AGENTS.md AssertJ guidelines:**
Good addition — makes an implicit project convention explicit for AI agents. The "don't mix styles in the same method" rule and the carve-out for Camel-specific assertions (`MockEndpoint.assertIsSatisfied()`) are practical guidelines that will prevent awkward hybrid test methods.

LGTM.

gnodet:

Related: #24956 adds the AssertJ assertion rule to AGENTS.md that this PR follows.

gnodet (commented):
Commit `9bb8ce1bdf25` reviewed — clean split: AGENTS.md guideline moved to #24956, this PR is now focused solely on the CxfTimeoutTest fix. The AssertJ migration and `satisfiesAnyOf` cause-chain check are unchanged from prior review. LGTM.

apupier (approved):

davsclaus (approved):

PR #24959, opened 2026-07-20
CAMEL-24119: Fix thread-safety in camel-openapi-java global state

davsclaus:
## Summary

- **RestModelConverters**: use `new ModelConverters()` instead of `ModelConverters.getInstance()` to create Camel-private instances, preventing Camel's custom converters from polluting the JVM-global swagger-core singleton shared by other libraries in the same process.
- **RestOpenApiSupport**: use `ObjectWriter.with(DateFormat)` instead of temporarily mutating the global `ObjectMapper`'s date format (save → set → serialize → restore), eliminating a race condition when concurrent HTTP requests serialize OpenAPI specs via `getFromOpenAPI3`.

Part of the [CAMEL-24119](https://issues.apache.org/jira/browse/CAMEL-24119) umbrella (thread-safety category).

## Test plan

- [x] All 97 existing tests in `camel-openapi-java` pass (model conversion, reader, support, complex types, security schemes)
- [ ] Verify no regressions in downstream consumers (`camel-rest-openapi`, Spring Boot / Quarkus OpenAPI integration tests)

### DIFF
--- a/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestModelConverters.java
+++ b/components/camel-openapi-java/src/main/java/org/apache/camel/openapi/RestModelConverters.java
@@ -44,14 +44,14 @@ public class RestModelConverters {
     private static final ModelConverters MODEL31_CONVERTERS;
 
     static {
-        MODEL31_CONVERTERS = ModelConverters.getInstance(true);
+        MODEL31_CONVERTERS = new ModelConverters(true);
         MODEL31_CONVERTERS.addConverter(new ClassNameExtensionModelResolver(new FqnModelResolver(true)));
     }
 
     private static final ModelConverters MODEL30_CONVERTERS;
 
     static {
-        MODEL30_CONVERTERS = ModelConverters.getInstance();
+        MODEL30_CONVERTERS = new ModelConverters();
         MODEL30_CONVERTERS.addConverter(new ClassNameExtensionModelResolver(new FqnModelResolver()));
     }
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
## Review: CAMEL-24119 — Fix thread-safety in camel-openapi-java global state

Clean, well-scoped thread-safety fix. Both changes are correct and minimal.

### RestModelConverters — `new ModelConverters()` vs `getInstance()`

`ModelConverters.getInstance()` returns a JVM-global singleton. Calling it and then adding Camel's custom `ClassNameExtensionModelResolver` permanently mutates that singleton, polluting any other library in the same JVM that uses swagger-core's `ModelConverters.getInstance()`. Using `new ModelConverters()` correctly isolates Camel's converter chain.

The `init()` skip is fine — the skipped-packages optimization and SPI-loaded third-party converters are exactly what isolation intends to avoid. Camel explicitly registers its own converters right after construction.

### RestOpenApiSupport — `ObjectWriter.with(DateFormat)`

The old save-set-serialize-restore pattern on `mapper.setDateFormat()` is a textbook race condition when the `mapper` is a shared singleton (from `Json.mapper()` / `Yaml.mapper()`). The new `mapper.writer(...).with(DEFAULT_DATE_FORMAT)` creates an immutable `ObjectWriter` without mutating the shared mapper. Thread-safety of the shared `SimpleDateFormat` is also fine — Jackson's `SerializerProvider.getDateFormat()` calls `clone()` before each serialization.

### Checklist

- ✅ **Tests**: 97 existing tests exercise both code paths; no behavioral change in single-threaded mode
- ✅ **Docs**: no user-visible change, no upgrade guide entry needed
- ✅ **Commit convention**: `CAMEL-24119: Fix thread-safety in ...` ✓
- ✅ **Public API**: no changes — `RestModelConverters` and `getFromOpenAPI3` are internal
- ✅ **Backward compat**: observable behavior identical
- ✅ **Security**: reduces attack surface by isolating model converters

PR #24972, opened 2026-07-21
chore: add rule to drop public qualifier from test classes and methods

gnodet:
## Summary

- Adds a new `### Test Visibility: Drop public From Test Classes and Methods` section to `AGENTS.md`
- JUnit 5 does not require `public` on test classes or methods — package-private is sufficient and preferred
- The rule is **incremental**: new tests must omit `public`, and modifications to existing tests should remove it from touched declarations — no bulk sweeps

## Test plan

- [x] Verify the new section is correctly placed after the Awaitility section and before Issue Investigation
- [x] Verify the examples are syntactically correct
- [x] Verify the rules are clear about incremental adoption (no standalone bulk PRs)


apupier (approved):
this is crazy that i is not the default of all models already

PR #24982, opened 2026-07-21
CAMEL-24224: camel-aws-bedrock - capture streaming metadata for Claude/Nova and accept inference profile ids

oscerd:
## Description

Fixes [CAMEL-24224](https://issues.apache.org/jira/browse/CAMEL-24224). Two defects in the Bedrock runtime producer.

### 1. Streaming metadata never captured for Claude/Nova

`BedrockStreamHandler` extracted metadata only inside `if (parser.isFinalChunk(chunkJson))`. For Claude — and Nova, which reuses `ClaudeStreamParser` — `isFinalChunk` matches `message_stop`, but `stop_reason` and `usage` arrive on the earlier **`message_delta`** chunk. A chunk is never both, so the completion reason and token count were **always absent** for those families.

The parsers were already correct (`ClaudeStreamParserTest` asserts extraction from a `message_delta` chunk) — only the handler's gating was wrong. Every chunk is now inspected and whatever metadata it carries is kept; parsers return `null` for chunks without it, so Titan/Llama/Mistral/Cohere (which report on their final chunk) are unaffected. The identical extraction in both handler variants is factored into one `applyChunkMetadata` method.

### 2. Inference profile ids and ARNs rejected

`setResponseText` switched on the exact configured `modelId` with a throwing `default`. Bedrock also accepts **cross-region inference profile ids** (`us.` / `eu.` / `apac.` / `us-gov.` prefixed, e.g. `us.anthropic.claude-3-5-sonnet-20241022-v2:0`) and **inference-profile ARNs**, which address the same foundation model — these failed with `Unexpected model: …`, as did any model released after the hard-coded list.

Now the underlying foundation-model id is resolved first (stripping the region prefix or ARN path), and ids that are still unlisted fall back to **family-prefix dispatch** (the response format is defined by the family, not the individual model). Non-text families — image, embedding, rerank, video, speech — still fail with a clear message rather than being parsed as text.

## Tests

- `BedrockStreamHandlerMetadataTest` — asserts Claude metadata is captured from the **non-final** `message_delta` chunk (explicitly asserting `isFinalChunk` is false for it), that a realistic chunk sequence preserves it, and that Titan's final-chunk metadata still works.
- `BedrockProducerModelIdTest` — plain ids, all four region prefixes, an inference-profile ARN, and a null id.

All 13 pre-existing stream tests still pass. Note: this module has no `assertj-core` test dependency, so these follow the module's existing JUnit assertion style rather than adding a dependency for style alone.

## Backport

Targeting 4.22.0 and 4.18.4 (the streaming/Converse code does not exist on 4.14.x — will confirm before backporting).

---

### DIFF
--- a/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/runtime/BedrockProducer.java
+++ b/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/runtime/BedrockProducer.java
@@ -63,6 +63,11 @@ import software.amazon.awssdk.services.bedrockruntime.model.ToolConfiguration;
 public class BedrockProducer extends DefaultProducer {
 
     private static final Logger LOG = LoggerFactory.getLogger(BedrockProducer.class);
+
+    /**
+     * Region prefixes used by Bedrock cross-region inference profile ids, e.g. {@code us.anthropic.claude-...}.
+     */
+    private static final List<String> INFERENCE_PROFILE_REGION_PREFIXES = List.of("us.", "eu.", "apac.", "us-gov.");
     private transient String bedrockProducerToString;
 
     public BedrockProducer(Endpoint endpoint) {
@@ -282,7 +287,7 @@ public class BedrockProducer extends DefaultProducer {
     }
 
     protected void setResponseText(InvokeModelResponse result, Message message) {
-        String modelId = getConfiguration().getModelId();
+        String modelId = resolveFoundationModelId(getConfiguration().getModelId());
         switch (modelId) {
             // Amazon Titan Models
             case "amazon.titan-text-express-v1":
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

gnodet (approved):
## Review: CAMEL-24224 — Bedrock streaming metadata + inference profile ids

Both fixes are correct and well-tested:

1. **Streaming metadata capture**: Refactoring the metadata extraction into `applyChunkMetadata` and calling it on every chunk (rather than only on `isFinalChunk`) correctly captures `stop_reason` and `usage` from Claude/Nova's `message_delta` chunk. Parsers return `null` for chunks without metadata, so Titan/Llama/Mistral/Cohere are unaffected. Good tests in `BedrockStreamHandlerMetadataTest` explicitly assert that `isFinalChunk` is false for the metadata-carrying chunk.

2. **Model id resolution**: `resolveFoundationModelId` correctly strips region prefixes (`us.`, `eu.`, `apac.`, `us-gov.`) and extracts foundation-model ids from ARNs. The family-prefix fallback dispatch handles unlisted future models gracefully. `BedrockProducerModelIdTest` covers all four prefixes, ARNs, and null.

### Observation

`StreamParserFactory.getParser()` in `BedrockStreamHandler` still receives the raw model id without `resolveFoundationModelId`, so the streaming path would still reject inference profile ids. This is pre-existing behavior (not introduced by this PR) and would be a natural follow-up.

LGTM ✅

---

PR #24984, opened 2026-07-21
CAMEL-24210: redact api-key and Authorization from sanitized URIs

oscerd:
### Motivation

[CAMEL-24210](https://issues.apache.org/jira/browse/CAMEL-24210): a credential passed through a hyphenated or `Authorization`-style endpoint parameter/header name is written in **clear text** wherever a sanitized endpoint URI is emitted (logs, JMX attributes, exception messages, developer console), even though the camelCase `apiKey` option is already redacted.

Root cause: `URISupport.sanitizeUri` is regex/keyword-driven and, unlike `SensitiveUtils.containsSensitive` (which strips dashes before matching), matches the raw parameter name. So:

| URI parameter | Masked before | Masked after |
|---|---|---|
| `apiKey=...` | yes | yes |
| `additionalHeader.api-key=...` (Azure OpenAI auth header) | **no** | yes |
| `additionalHeader.Authorization=...` | **no** | yes |

`api-key` is precisely the shape `camel-openai`'s `additionalHeader` option documentation recommends, so the recommended form was the one not redacted.

This is **hardening / defence in depth**, not a framework vulnerability — the route author placing the secret in the URI is fully trusted under the threat model; no untrusted input crosses a boundary. Filed and fixed as a normal improvement. The concern is that Camel already redacts `apiKey`, so users reasonably expect credentials in endpoint URIs to be masked.

### Changes

- Add `api-key` and `authorization` to the sensitive-keyword generator's `EXTRA_KEYS` (`UpdateSensitizeHelper`), and regenerate `SensitiveUtils` + `sensitive-keys.json`. `api-key` is intentionally dashed: the `sanitizeUri` regex path (wrapped in `\Q...\E`) needs the literal hyphen, whereas the dash-stripping `containsSensitive` path already handled it. The generator comment is updated to explain this exception to the "no dash" convention.
- Unit tests in `URISupportTest` for both the hyphenated `api-key` and the `Authorization` header (AssertJ, package-private per the current test conventions).
- Upgrade-guide entry noting the added redaction **and** the small, fail-safe over-match: because matching is a case-insensitive substring, non-secret names containing `authorization` (`adjustAuthorization`, `jwtAuthorizationType`, `proxyAuthorizationPolicy`) now also show `xxxxxx` in sanitized URIs — a cosmetic change to sanitized output only.

No `camel-openai` change is needed: with `api-key` now redacted, the option's documented example is no longer steering users toward an unredacted shape.

### Testing

- `mvn test -Dtest=URISupportTest` in `core/camel-util`: 54 tests pass (incl. 2 new).
- Full reactor `mvn clean install -DskipTests` from root: success; `SensitiveUtils`/`sensitive-keys.json` regenerated with only the two additions (no unrelated drift).

Main-only (4.22.0): defence-in-depth improvement, no backport.


davsclaus (approved):

gnodet (approved):
## Review: CAMEL-24210 — Redact api-key and Authorization from sanitized URIs

Thorough security hardening fix. The implementation is correct and complete:

- **Generator** (`UpdateSensitizeHelper`): `api-key` and `authorization` added to `EXTRA_KEYS` with a clear comment explaining why `api-key` intentionally breaks the "no dashes" convention (the `sanitizeUri` regex path needs the literal hyphen, unlike `containsSensitive` which strips dashes).
- **Generated files** (`SensitiveUtils.java`, `sensitive-keys.json`): Both keys appear in the sorted set and the `\Q...\E`-quoted regex pattern.
- **Tests**: Two new tests in `URISupportTest` covering the exact real-world patterns (`additionalHeader.api-key=...` and `additionalHeader.Authorization=...`).
- **Upgrade guide**: Documents the change and honestly discloses the known over-match for non-secret parameter names containing `authorization` — cosmetic-only impact.

The known over-match tradeoff (redacting `adjustAuthorization` etc.) is the right call — masking too much in sanitized output is far safer than masking too little.

LGTM ✅

---

PR #24985, opened 2026-07-21
CAMEL-24227: Add volatile to JMX-writable fields read on routing threads

gnodet:

## Summary

An audit of the Camel management layer revealed a systemic JMM-unsafe pattern across **15 engine classes (~50+ fields)**. Fields writable via JMX `@ManagedAttribute` setters are stored as plain (non-volatile) fields in engine classes, but read on routing threads without any memory visibility guarantee.

### Fix approach

- **Simple boolean/int/long fields**: added `volatile`. On x86 this compiles to the same instruction as a plain load, so there is **zero performance cost**. On ARM/POWER architectures with weaker memory ordering, this provides the required happens-before guarantee.

- **Compound writes** (three cases): replaced coupled field pairs with **immutable holder records** swapped atomically via a single volatile reference, ensuring routing threads always see consistent values:
  - `BacklogTracer`: `tracePattern`+`patterns` → `TracePatternHolder`, `traceFilter`+`predicate` → `TraceFilterHolder`
  - `DefaultTracer`: `tracePattern`+`patterns` → `TracePatternHolder`
  - `ThrottlingInflightRoutePolicy`: `maxInflightExchanges`+`resumeInflightExchanges` → `ThrottlingLimits`

### Affected classes (15 files)

| Engine class | Fields made volatile | Hot path |
|---|---|---|
| `BacklogTracer` | `enabled`, `standby`, `bodyMaxChars`, `bodyIncludeStreams`, `bodyIncludeFiles`, `includeExchangeProperties`, `includeExchangeVariables`, `includeException`, `backlogSize`, `activitySize`, `removeOnDump`, `activityEnabled`, `traceRests`, `traceTemplates` + holder records | `shouldTrace()`, `traceNode()` |
| `DefaultTracer` | `enabled`, `standby`, `traceRests`, `traceTemplates`, `traceBeforeAndAfterRoute` + holder record | `shouldTrace()` |
| `BaseProcessorSupport` | `disabled` | `CamelInternalProcessor.process()` |
| `BaseDelegateProcessorSupport` | `disabled` | `CamelInternalProcessor.process()` |
| `WrapProcessor` | `disabled` | `CamelInternalProcessor.process()` |
| `DefaultBacklogDebugger` | `fallbackTimeout`, `bodyMaxChars`, `bodyIncludeStreams`, `bodyIncludeFiles`, `includeExchangeProperties`, `includeExchangeVariables`, `includeException`, `singleStepIncludeStartEnd` | `NodeBreakpoint.beforeProcess()` |
| `TotalRequestsThrottler` | `timePeriodMillis` | `process()` |
| `AbstractThrottler` | `maxRequestsExpression` | `process()` |
| `DefaultStreamCachingStrategy` | `spoolThreshold`, `spoolUsedHeapMemoryThreshold`, `anySpoolRules`, `bufferSize` + `UtilizationStatistics.statisticsEnabled` | `shouldSpoolCache()` |
| `ThrottlingInflightRoutePolicy` | holder record for `maxInflightExchanges`+`resumeInflightExchanges` + `scope` | `throttle()` |
| `ThrottlingExceptionRoutePolicy` | `halfOpenAfter`, `failureWindow`, `failureThreshold` | `calculateState()` |
| `ManagedPerformanceCounter` | `statisticsEnabled` | `DefaultInstrumentationProcessor.before()` |
| `ScheduledPollConsumer` | `greedy`, `sendEmptyMessageWhenIdle`, `delay`, `initialDelay`, `timeUnit`, `useFixedDelay`, `runLoggingLevel` | `doRun()` |
| `Delayer` | `delay` | `calculateDelay()` |

### Bonus fix

`BacklogTracer.setTraceFilter(null)` previously nulled the filter string but never cleared the compiled `predicate`, leaving `shouldTrace()` evaluating a stale filter after the user cleared it via JMX. The `TraceFilterHolder` pattern atomically clears both. Added `BacklogTracerFilterClearTest` to cover this.

## Test plan

- [x] All 15 affected modules compile successfully
- [x] Core tests pass (throttler, backlog tracer, stream caching, scheduled poll, delayer)
- [x] Management tests pass (BacklogTracer, BacklogDebugger, Tracer, PerformanceCounter)
- [x] New `BacklogTracerFilterClearTest` verifies setTraceFilter(null) fix
- [x] ThrottlingInflightRoutePolicyTest passes with holder record changes
- [x] Code formatting verified clean
- [ ] Full CI


davsclaus (approved):

gnodet (commented):

## Review Summary

This PR addresses a systemic JMM-unsafe pattern across 12 engine classes where JMX-writable fields lacked memory visibility guarantees for routing threads. The fix is methodical and well-scoped: simple fields get `volatile`, and compound writes (three cases) are correctly replaced with immutable holder records swapped via a single volatile reference. After thorough analysis of all 12 files and their corresponding MBeans, the changes are correct and complete.

## What I Love

1. **The holder record pattern is textbook-perfect.** In all three compound-write cases (`BacklogTracer.TracePatternHolder`/`TraceFilterHolder`, `DefaultTracer.TracePatternHolder`, `ThrottlingInflightRoutePolicy.ThrottlingLimits`), the snapshot-then-use idiom reads the volatile reference exactly once into a local, guaranteeing routing threads never see a torn pair. The method signatures were also updated to pass the snapshot values as parameters (e.g., `shouldTracePattern(definition, ph.patterns())`, `startConsumer(size, consumer, resumeInflight)`) rather than re-reading the holder -- this is the right way to propagate a consistent snapshot through a call chain.

2. **The PR silently fixes a pre-existing bug in `BacklogTracer.setTraceFilter(null)`.** The old code set `traceFilter = null` but never cleared `predicate`, leaving `shouldTraceFilter()` evaluating a stale predicate after the filter was removed. The new code correctly sets `traceFilterHolder = null`, and the `if (fh != null)` guard in `shouldTrace()` skips the filter check entirely. The same fix applies to `setTracePattern(null)`.

3. **Discipline in scoping.** Fields only set during construction or lifecycle startup (e.g., `AbstractThrottler.rejectExecution`, `ScheduledPollConsumer.initialDelay`, `ScheduledPollConsumer.timeUnit`, `DefaultBacklogDebugger.suspendMode`) are correctly left non-volatile. The PR doesn't over-volatilize, which shows careful analysis of which fields are truly written by JMX at runtime vs. configured once before routing threads start.

## Findings

### Critical

None -- nice work!

### Important

None.

### Suggestions & Nits

**[Question]** `DefaultTracer.traceCounter` (line 62): This is a plain `long` that's incremented with `traceCounter++` on routing threads and read/reset via JMX (`ManagedTracer.getTraceCounter()`/`resetTraceCounter()`). Strictly speaking, this has two JMM issues: (a) visibility -- JMX reads may see stale values, and (b) atomicity -- the `++` is a non-atomic read-modify-write. `BacklogTracer` already uses `AtomicLong` for the same purpose. I understand this is the reverse direction (routing writes -> JMX reads, vs. this PR's focus on JMX writes -> routing reads), so it's out of scope. Would it be worth a follow-up ticket to align `DefaultTracer.traceCounter` with `BacklogTracer`'s `AtomicLong` pattern?

## Overall

Clean, well-reasoned concurrency fix. The approach of volatile for simple fields + immutable holders for compound state is the standard JMM-safe pattern, and it's applied consistently across all 12 classes. On x86 the volatile reads compile to plain loads (zero overhead); on ARM/POWER the memory barriers are exactly what's needed. The bonus fix for the stale-predicate bug in `setTraceFilter(null)` is a nice win. LGTM.

oscerd (commented):
Went through each field's read and write sites rather than just checking the modifier. The core of this is right: for every field the PR marks `volatile`, the reader side does a plain read with no compound action, so visibility is the correct and sufficient fix. The three genuine read-modify-write cases were correctly spotted and moved to immutable holders instead — that's the right distinction to draw, and it's the part these PRs usually get wrong.

Two things I'd like to see before this lands, plus one unrelated bug I tripped over.

### The sweep is incomplete for the `disabled` flag

`BaseProcessorSupport.disabled` is now volatile, but two siblings with the *identical* writer/reader pair were left plain:

- `core/camel-core-processor/src/main/java/org/apache/camel/processor/BaseDelegateProcessorSupport.java:31` — `private boolean disabled;`
- `core/camel-core-processor/src/main/java/org/apache/camel/processor/WrapProcessor.java:32` — `private boolean disabled;`

Both are written from JMX via `ManagedProcessor.enable()` / `disable()` (`core/camel-management/src/main/java/org/apache/camel/management/mbean/ManagedProcessor.java:232`, `:239` → `da.setDisabled(...)`) and read on the routing hot path at `core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java:305` (`... instanceof DisabledAware da && da.isDisabled()`).

That matters because `BaseDelegateProcessorSupport` is the base of `FilterProcessor`, `LoopProcessor`, `CatchProcessor`, `FinallyProcessor`, `DelayProcessorSupport` (hence `Delayer`, which this PR *does* touch), `SagaProcessor` and others, while `WrapProcessor` is installed by the policy/transacted reifiers. So as it stands, `enable()`/`disable()` over JMX remains unreliable for the majority of processors it applies to — which is the exact bug this PR sets out to fix.

`DefaultStreamCachingStrategy.UtilizationStatistics.statisticsEnabled` looks like the same category and is in a file you're already editing: written from `ManagedStreamCachingStrategy.setStatisticsEnabled`, read on the caching path in `doCache`. Same pattern as `ManagedPerformanceCounter`, which you did make volatile.

### The `setTraceFilter(null)` fix needs a test

`BacklogTracer.setTraceFilter(null)` previously nulled `traceFilter` but left `predicate` at its old value, so `shouldTrace()` kept applying a filter the user had just cleared. Nulling the holder fixes it, and it's reachable over JMX through `ManagedBacklogTracer`. That's a genuine, deterministic bug fix rather than a visibility change — the project rule is that bug fixes come with a test, and this one is a plain unit test. Worth an upgrade-guide line too, since the observable behaviour changes.

### Correcting something I initially thought was a blocker

I want to flag this explicitly because it nearly went in as an API-break finding: `shouldTracePattern` in `DefaultTracer` gains a `String[] patterns` parameter, and `DefaultTracer` is public and non-final — but the method is `private`, not `protected`. So there is **no** source or binary break and no upgrade-guide obligation. Disregard if anyone raises it.

### Smaller points

- **`ScheduledPollConsumer.delay`**: `volatile` here has no practical effect. It's snapshot into the scheduler in `afterConfigured`-style setup (`scheduler.setDelay(delay)` and friends), not read from `run()` as the PR body's table suggests. More to the point, `initialDelay`, `timeUnit` and `useFixedDelay` are equally JMX-writable and read in the same place but were left plain — so the treatment is inconsistent either way. Suggest doing all four or none.
- **`ThrottlingInflightRoutePolicy.setResumePercentOfMax`** still does a read-modify-write across the pair (it reads `maxInflightExchanges` to compute the resume value), so two concurrent JMX writers can still publish a holder built from a mixed snapshot. The holder fixes reader-side tearing, not writer-side atomicity.
- **`resumeInflightExchanges`** is now write-only — `startConsumer` takes it as a parameter, there's no getter, and `toString()` doesn't use it. The default is also duplicated between the field initialisers and the literal `new ThrottlingLimits(...)`.
- **`ThrottlingInflightRoutePolicy` has no `@ManagedAttribute` at all** — it's a `@Metadata(label = "bean")` RoutePolicy. The "JMX-writable" framing doesn't apply to that class, though the visibility argument still holds for programmatic reconfiguration.
- **Reverse-direction gaps** (you flagged the first yourself): `DefaultTracer.traceCounter` is a plain `long` incremented non-atomically on routing threads and read/reset over JMX, and `Delayer.delayValue` likewise. Both are non-atomic on 32-bit JVMs. Worth a follow-up ticket rather than growing this PR.

### Unrelated bug in a file this PR touches

While tracing `statisticsEnabled` I found a live defect in `DefaultStreamCachingStrategy` on `main` — not introduced here, and not something I'd ask you to fix in this PR:

```java
void updateSpool(long size) {
    lock.lock();
    try {
        spoolAverageSize.set(spoolSize.addAndGet(size) / spoolCounter.incrementAndGet());
    } finally {
        lock.lock();     // should be unlock()
    }
}
```

`updateMemory` directly above it correctly calls `unlock()`. Because `lock` is a `ReentrantLock`, the calling thread doesn't self-deadlock — it just leaves the hold count at 2, permanently — but any *other* thread entering `updateSpool` blocks forever. It's reachable whenever stream-caching statistics are enabled and spooling to disk occurs. I've raised it separately so it doesn't get lost in this review.

CI is green. Nothing here is blocking from my side; the incomplete sweep is the one I'd most like addressed, since a partial fix here is easy to mistake for a complete one later.

---

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultTracer.java
```diff
@@ -347,7 +352,7 @@ protected void dumpTrace(String out, Object node) {
         }
     }
 
-    protected boolean shouldTracePattern(NamedNode definition) {
+    protected boolean shouldTracePattern(NamedNode definition, String[] patterns) {
```
davsclaus, line 355: This is a `protected` method in a public non-final class — oscerd's review dismissed this as `private`, but it is `protected` on `main` (line 350). The signature change from `shouldTracePattern(NamedNode)` to `shouldTracePattern(NamedNode, String[])` is a source/binary compatibility break for any downstream code extending `DefaultTracer` and overriding this method.

No known subclasses exist in the Camel codebase, and the class is in `camel-base-engine` (internal module), so practical risk is very low. Just flagging that the concern was dismissed on incorrect grounds.

core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogTracer.java
```diff
@@ -478,37 +484,41 @@ public void setTraceTemplates(boolean traceTemplates) {
...
-                predicate = camelContext.resolveLanguage(name).createPredicate(filter);
+                p = camelContext.resolveLanguage(name).createPredicate(filter);
             } else {
                 // use simple language by default
-                predicate = simple.createPredicate(filter);
+                p = simple.createPredicate(filter);
             }
+            this.traceFilterHolder = new TraceFilterHolder(filter, p);
+        } else {
+            this.traceFilterHolder = null;
```
davsclaus, line 521: This `else` branch fixes a pre-existing bug: the old code set `traceFilter = null` but never cleared `predicate`, so `shouldTrace()` would keep evaluating a stale predicate after the user cleared the filter via JMX. The fix is correct.

Since this is a deterministic behavior change (not just a visibility fix), it should have a test — e.g. set a filter, verify it's applied, clear it with `setTraceFilter(null)`, verify filtering is disabled.

davsclaus (commented):

## Review Summary

The core change is correct and valuable — volatile for simple fields, immutable holders for compound writes is textbook JMM-safe. CI is green and the scope is well-defined.

I want to flag one factual correction on oscerd's review and echo two of their actionable findings.

### Factual correction: `DefaultTracer.shouldTracePattern` IS `protected`, not `private`

oscerd's review dismissed the signature change saying "the method is `private`, not `protected`." This is incorrect — on `main` at line 350 the method is `protected boolean shouldTracePattern(NamedNode definition)`. The PR changes it to `protected boolean shouldTracePattern(NamedNode definition, String[] patterns)`. Since `DefaultTracer` is a public non-final class, this is technically a source/binary compatibility break for anyone extending it and overriding the method. There are no known subclasses in the codebase, and the class lives in `camel-base-engine` (internal), so the practical risk is very low — but the concern was incorrectly dismissed.

### Confirming: incomplete `disabled` sweep

oscerd correctly flagged that `BaseDelegateProcessorSupport.disabled` and `WrapProcessor.disabled` have the identical JMX-write / routing-read pattern as `BaseProcessorSupport.disabled` but were left plain. Since `BaseDelegateProcessorSupport` is the base of `FilterProcessor`, `LoopProcessor`, `CatchProcessor`, `FinallyProcessor`, `DelayProcessorSupport` and others, the `enable()`/`disable()` JMX operation remains unreliable for the majority of processors. Should be addressed in this PR or an explicit follow-up.

### Confirming: `setTraceFilter(null)` bugfix deserves a test

The old code set `traceFilter = null` but never cleared `predicate`, leaving `shouldTrace()` evaluating a stale predicate. The new holder pattern correctly fixes this. This is a deterministic behavior change that warrants a test per project conventions.

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultTracer.java
gnodet: _Claude Code on behalf of gnodet_

Acknowledged — you're right that `shouldTracePattern` is `protected`, not `private`. The signature change is technically a source/binary break for downstream subclasses. Since no known subclasses exist and the class is in `camel-base-engine` (internal), the practical risk is very low, but it's correctly flagged. Thank you for the correction.

core/camel-base-engine/src/main/java/org/apache/camel/impl/debugger/BacklogTracer.java
gnodet: _Claude Code on behalf of gnodet_

Good catch — added `BacklogTracerFilterClearTest` in d999aab0c95. The test verifies the full lifecycle: set filter → trace with filter → clear filter → trace without filter. It confirms the stale-predicate bug is fixed by the holder pattern.

gnodet:

## Review feedback addressed in d999aab0c95

All actionable findings from @oscerd and @davsclaus have been addressed:

### 1. Incomplete `disabled` sweep (oscerd, confirmed by davsclaus)
Added `volatile` to:
- `BaseDelegateProcessorSupport.disabled` — base of `FilterProcessor`, `LoopProcessor`, `CatchProcessor`, `FinallyProcessor`, `DelayProcessorSupport`, etc.
- `WrapProcessor.disabled` — installed by policy/transacted reifiers

Both are written from JMX via `ManagedProcessor.enable()`/`disable()` and read on the routing hot path at `CamelInternalProcessor:305`.

### 2. `statisticsEnabled` in `DefaultStreamCachingStrategy` (oscerd)
Added `volatile` to `UtilizationStatistics.statisticsEnabled` — written from `ManagedStreamCachingStrategy.setStatisticsEnabled`, read on the caching path.

### 3. `ScheduledPollConsumer` inconsistency (oscerd)
Added `volatile` to `initialDelay`, `timeUnit`, and `useFixedDelay` for consistency with `delay` which was already volatile. All four are equally JMX-writable via `ManagedScheduledPollConsumer`.

### 4. `setTraceFilter(null)` test (oscerd, confirmed by davsclaus)
Added `BacklogTracerFilterClearTest` that verifies the full lifecycle: set filter → trace with filter active → clear filter via `setTraceFilter(null)` → verify messages are traced without filter. Confirms the stale-predicate bug is fixed by the `TraceFilterHolder` pattern.

### 5. `shouldTracePattern` is protected, not private (davsclaus)
Acknowledged — the signature change from `shouldTracePattern(NamedNode)` to `shouldTracePattern(NamedNode, String[])` is technically a source/binary break. No known subclasses exist and the class is in `camel-base-engine` (internal), so practical risk is very low.

### Items noted as out-of-scope (follow-up tickets)
- `DefaultTracer.traceCounter` should be `AtomicLong` (reverse-direction gap)
- `ThrottlingInflightRoutePolicy.setResumePercentOfMax` writer-side atomicity
- `DefaultStreamCachingStrategy.updateSpool` has `lock.lock()` instead of `unlock()` (unrelated bug, raised separately by oscerd)

PR #24989, opened 2026-07-21
CAMEL-23945: Add regression test for producer race on shared agent field

gnodet:
## Summary

CAMEL-23945: `LangChain4jAgentProducer.process()` had a race condition when using `AgentFactory` — the factory-created agent was written to the shared `agent` instance field and then re-read. Since a producer is a singleton per endpoint and `process()` must be thread-safe, concurrent exchanges selecting different agents could see each other's agents.

**The race was inadvertently fixed on `main`** in CAMEL-23697 (commit `03d8361bbb74`, 2026-06-17) which refactored the method to use a local variable:

```java
// Before (race): wrote to instance field, then re-read it
if (agentFactory != null) {
    agent = agentFactory.createAgent(exchange);
}
String response = agent.chat(aiAgentBody, toolProvider);

// After (safe): local variable, thread-confined
Agent agent = agentFactory != null ? agentFactory.createAgent(exchange) : this.agent;
Result<String> result = agent.chat(aiAgentBody, toolProvider);
```

This PR adds a **regression test** that:
- Configures a `PerExchangeAgentFactory` that creates a distinct agent per exchange (based on a header)
- Sends 50 concurrent exchanges through a single producer endpoint using a `CountDownLatch` to maximise contention
- Verifies each exchange received the response from its own agent, not from another thread's agent

## Test plan

- [x] New test `LangChain4jAgentProducerConcurrencyTest` passes (1 test, 0 failures)
- [x] All existing module tests pass (27 tests, 0 failures)
- [x] Code formatting validated (`-Psourcecheck validate`)

## Backport

The code fix + regression test has been backported to `camel-4.18.x` in #25063.

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent/src/test/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducerConcurrencyTest.java
+++ b/components/camel-ai/camel-langchain4j-agent/src/test/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentProducerConcurrencyTest.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


davsclaus:
why do you not fix the issue in 4.18.x also

gnodet:

Good point. On `main`, the race was inadvertently fixed by CAMEL-23697 (commit 03d8361bbb7) which refactored `process()` to use a local variable. This PR only adds the regression test since the fix is already in place.

However, `camel-4.18.x` still has the racy instance-field write. I'll create a separate backport PR for `camel-4.18.x` with both the code fix and the test.

davsclaus (approved):

PR #24991, opened 2026-07-21
CAMEL-23943: Fix three defects in camel-langchain4j-tools producer

gnodet:
## Summary

Fixes three defects in the `camel-langchain4j-tools` producer as reported in [CAMEL-23943](https://issues.apache.org/jira/browse/CAMEL-23943):

1. **Unbounded tool-calling loop** — The `do/while(true)` loop in `toolsChat()` has no iteration cap. If the LLM keeps requesting tools indefinitely, it never terminates. **Fix:** Added `maxToolCallingRoundTrips` endpoint option (default 10, producer-only) that bounds the loop and throws `RuntimeCamelException` when exceeded.

2. **Crash on hallucinated tool names** — When the LLM hallucinates a tool name, `.findFirst().get()` throws `NoSuchElementException`. **Fix:** Changed to `.orElse(null)` with a graceful error message sent back to the LLM listing available tools.

3. **Tool errors swallowed silently** — When a tool route throws an exception, the error is set on `toolExchange` but never propagated to the LLM; subsequent `ExchangeHelper.copyResults` overwrites exchange state. **Fix:** Errors are now sent back to the LLM as `ToolExecutionResultMessage`, moved `copyResults` inside the try block.

## Changes

- `LangChain4jToolsProducer.java` — All three fixes in `toolsChat()`/`invokeTools()`, with null-safe `e.getMessage()` handling
- `LangChain4jToolsEndpoint.java` — New `maxToolCallingRoundTrips` `@UriParam` (producer-only, default 10), with validation rejecting negative values
- `LangChain4jToolDefectsTest.java` — Three targeted reproducer tests (one per defect), verified to fail before fix and pass after
- `camel-4x-upgrade-guide-4_22.adoc` — Documents all three behavior changes
- Generated files: component JSON, catalog JSON, endpoint DSL factory

## Test plan

- [x] All 3 tests fail on original code (verified by reverting producer to parent commit)
- [x] All 3 tests pass with fix applied
- [x] Full module test suite passes (42 tests, 0 failures)
- [x] CI green

## Review feedback addressed

- [x] Added upgrade guide entry (requested by @davsclaus, @oscerd, self-review)
- [x] Fixed `e.getMessage()` null check — uses class name as fallback (flagged by self-review, @oscerd)
- [x] Reject negative `maxToolCallingRoundTrips` with `IllegalArgumentException` (flagged by @oscerd)
- [x] Acknowledged `exposed=false` + `ToolSearchTool` dead path as pre-existing issue (flagged by @oscerd) — will create JIRA follow-up

🤖 _Generated with [Claude Code](https://claude.com/claude-code) on behalf of gnodet_


gnodet (commented):

## Review Summary

Focused, well-crafted fix for three genuine defects in the `LangChain4jToolsProducer.toolsChat()` loop. Each fix is minimal, targeted, and backed by a dedicated test. The changes are production-ready.

## What I Love

- **The `ExchangeHelper.copyResults` relocation is subtle but important.** Moving it inside the try block means a failed tool no longer pollutes the main exchange with a stale exception. In the original code, `copyResults` was called unconditionally after the catch block, which copied the exception from `toolExchange` to the main `exchange` — yet the loop continued anyway, leaving the exchange in an inconsistent state. This fix resolves that silently.

- **Sending errors back to the LLM as `ToolExecutionResultMessage` is the right pattern.** This gives the LLM a chance to self-correct (e.g., try a different tool or answer directly) rather than crashing the entire interaction. The hallucinated-tool error message even lists available tools, which is a nice touch for LLM self-correction.

- **The test design using `OpenAIMock` with targeted scenarios is excellent.** Each test reproduces exactly one defect, the `@Timeout(10)` on the infinite-loop test is a proper safety net, and the assertions verify both the positive (got a result) and negative (no exception) conditions.

## Findings

### Critical

None — nice work!

### Important

None.

### Suggestions & Nits

1. **[Suggestion] Upgrade guide entry for `maxToolCallingRoundTrips` default.** The previous behavior was effectively unlimited round trips (the `do/while(true)` ran forever). With the new default of 10, any existing setup where the LLM legitimately needs 11+ round trips before converging would now throw a `RuntimeCamelException`. While that scenario is unlikely — and was a latent bug anyway — an upgrade guide entry mentioning the new option and its default would help users who encounter the exception after upgrading. Something along the lines of:

   > The `langchain4j-tools` producer now limits tool-calling round trips to 10 by default (previously unlimited). If your LLM interaction legitimately requires more iterations, set the `maxToolCallingRoundTrips` endpoint option to a higher value. Set to 0 for unlimited (not recommended).

2. **[Nit] `e.getMessage()` in the tool-error result could be null.** If the caught exception has a null message (rare but possible, e.g., `new RuntimeException()`), the tool result sent to the LLM would read `"Error executing tool 'ToolName': null"`. A minor defensive check would make this slightly cleaner:

   ```java
   String errorDetail = e.getMessage() != null ? e.getMessage() : e.getClass().getName();
   toolResult = "Error executing tool '" + toolName + "': " + errorDetail;
   ```

   This is purely cosmetic — the current behavior is functional.

## Overall

This is a thorough, well-reasoned fix. The three defects were real, the fixes are clean, and the test coverage is solid. The interaction between all three fixes is sound: hallucinated-tool errors and tool-execution errors feed back into the LLM gracefully, and the max round trips cap provides a safety net across all scenarios. The check for `toolExchange.getException()` after `process()` is particularly important — without it, Camel's `throwException` DSL (which sets the exception on the exchange rather than throwing it) would be silently swallowed.

_Note: This review would be an APPROVE, but GitHub does not allow approving your own PR. Requesting review from relevant committers is recommended._

davsclaus (approved):
LGTM — all three fixes are correct, well-tested, and follow project conventions.

### Summary

The PR fixes three genuine defects in `LangChain4jToolsProducer`:

1. **Unbounded tool-calling loop** — bounded by new `maxToolCallingRoundTrips` option (default 10)
2. **Crash on hallucinated tool names** — `.findFirst().get()` → `.orElse(null)` with graceful error back to LLM
3. **Tool errors swallowed silently** — errors now sent back to the LLM as `ToolExecutionResultMessage`

### Conventions

- ✅ Test class and methods are package-private
- ✅ AssertJ assertions used throughout
- ✅ No `Thread.sleep()`
- ✅ Generated files regenerated
- ✅ `@Timeout` safety net on infinite loop test

### Suggestion

- **Upgrade guide entry**: Adding `maxToolCallingRoundTrips` with default 10 changes behavior from unbounded to bounded loops. Existing routes exceeding 10 round trips would now throw `RuntimeCamelException`. Consider documenting this in `camel-4x-upgrade-guide-4_22.adoc`.

oscerd (commented):
Reviewed all three fixes. Since this producer sits on the prompt-injection path, I want to lead with the security check: **the `declaredParams` allowlist is preserved verbatim and is not weakened by this PR.**

I verified that mechanically rather than by eye. The allowlist derivation and its guard are outside every hunk in the diff — hunk 2 ends at old line 222 and hunk 3 starts at old line 261, so the whole block survives untouched. On the PR head the guard still reads `if (!allowedParams.contains(name))` with the "Skipping undeclared tool argument" warning, and the single call that turns a model-supplied name into a header still sits inside it. Both new code paths — the hallucinated-tool branch and the error branch — set zero headers. Good.

The round-trip cap and the hallucinated-tool handling both look right to me, and the test class is well-formed (AssertJ, `@Timeout(10)`, no `Thread.sleep`, package-private).

### The one thing I'd like addressed: tool failures no longer reach the route

This is the change I think deserves more attention than the round-trip default that earlier reviews focused on.

`ExchangeHelper.copyResults` moved *inside* the `try`, so it now runs only on the success path. The consequence chain:

- `ExchangeHelper.doCopyResults` ends with `result.setException(source.getException())` (`core/camel-support/src/main/java/org/apache/camel/support/ExchangeHelper.java:400`).
- **Before**, `copyResults` was unconditional, so a tool failure propagated onto the caller's exchange and the route failed / `onException` fired.
- **After**, on the error path the calling exchange is never touched. The producer returns normally, the route **succeeds**, and the failure is visible only as a string handed back to the LLM.

So a route author who wrapped `langchain4j-tools:` in `onException(...)` expecting to catch tool failures now gets silence.

I want to be fair about the starting point: the old behaviour was already unreliable, because a later successful tool's `copyResults` would call `setException(null)` and wipe the exception. So this isn't a clean regression — it replaces a flaky signal with no signal. And feeding the error back to the LLM so it can recover is a defensible design; it may well be what you intend.

But it is a user-visible behaviour change with no opt-out and no documentation, and I don't think users will discover it until a tool silently fails in production. At minimum it needs an upgrade-guide note; a `failOnToolError`-style switch would be better if you want to keep both audiences.

### Missing upgrade-guide entry

Related: `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc` has no `=== camel-langchain4j-tools` section at all. Two things here are user-visible — the round-trip cap going from unbounded to 10 and now throwing, and the error-propagation change above. For calibration, CAMEL-23382 added a 37-line entry to that same file for a comparable change in the adjacent component.

### Smaller points

- **Defect 2's fix closes the crash but the underlying feature stays dead.** The `exposed=false` + tool-search combination cannot work by construction: `CamelToolExecutorCache` keeps `tools` and `searchableTools` in two disjoint maps, `computeCandidates` builds `callableTools` from the *exposed* map only, and `ToolSearchTool` searches the *searchable* map. So anything the search tool returns is guaranteed absent from `callableTools`. Previously that was a `NoSuchElementException`; now it's `"Tool 'X' not found. Available tools: <exposed only>"`, which directly contradicts what the search tool just told the model — so the model will likely retry and burn round trips until the new cap throws. Better than crashing, but worth a follow-up ticket rather than leaving it looking fixed.
- **Negative values silently mean unlimited.** The guard is `maxRoundTrips > 0` while the description says "Set to 0 for unlimited", so `-1` also disables the cap with no warning.
- `e.getMessage()` can be null, giving `"Error executing tool 'X': null"` — you flagged this yourself in the self-review.
- The option is endpoint-only; `LangChain4jToolsConfiguration` holds just `chatModel`, so it can't be set via `camel.component.langchain4j-tools.*`. The agent side routes the equivalent through `AgentConfiguration`. Not necessarily wrong, just an asymmetry.

CI is green and the branch is clean. Nothing here blocks from my side except the doc gap, and I'd genuinely like your read on whether the error-swallowing is intended.

---

gnodet:

@oscerd Thank you for the thorough security-focused review. All actionable items addressed in c82475d22a85:

### Addressed

1. **Upgrade guide entry** — Added a `=== camel-langchain4j-tools` section covering all three behavior changes: the round-trip cap, hallucinated tool handling, and the error propagation change. The error propagation section explicitly calls out that `onException(...)` will no longer see tool failures.

2. **`e.getMessage()` null check** — Now uses `e.getClass().getName()` as fallback when message is null.

3. **Negative values** — `setMaxToolCallingRoundTrips` now throws `IllegalArgumentException` for negative values.

### Acknowledged (not in scope for this PR)

4. **`exposed=false` + tool-search dead path** — You're right that this is broken by construction. The fix in this PR (graceful error instead of crash) is strictly better than the `NoSuchElementException`, but the feature remains non-functional. I'll create a JIRA follow-up for this since it's a pre-existing architectural issue, not something introduced by this PR.

5. **Endpoint-only option** — Valid observation about asymmetry with `AgentConfiguration`. This is consistent with how the existing endpoint options work in this component, but worth considering for a follow-up.

### On the error propagation design

The previous behavior was indeed unreliable as you noted — a later successful tool's `copyResults` would wipe the exception, and the loop continued anyway. The new behavior (feed errors back to the LLM) is intentional: it follows the LangChain4j pattern where the LLM gets a chance to self-correct. The upgrade guide now documents this clearly, including the impact on `onException(...)` routes.

davsclaus (approved):

PR #24994, opened 2026-07-21
CAMEL-24231: camel-file - Fix pollEnrich with dynamic fileName using exchange properties

davsclaus:
## Summary

Fixes [CAMEL-24231](https://issues.apache.org/jira/browse/CAMEL-24231).

`GenericFileHelper.createDummy()` creates a dummy exchange for evaluating `fileName` expressions during `pollEnrich` optimization (`PollDynamicAware`), but it only copies **headers** from the incoming exchange — not **exchange properties**. When `fileName` uses `${exchangeProperty.xxx}`, the expression evaluates to `null` on the dummy, effectively bypassing the filename filter and picking up any file in the directory.

The workaround is to set `allowOptimisedComponents(false)` on the pollEnrich.

**Changes:**
- Copy exchange properties from the dynamic exchange to the dummy exchange in `GenericFileHelper.createDummy()`
- Fix secondary bug: variables copy was nested inside the `hasHeaders()` guard, so variables would not be copied when the exchange had variables but no headers
- Add test `PollDynamicFileNameExchangePropertyTest` that verifies `pollEnrich` with `fileName=${exchangeProperty.target}` works correctly with the optimization enabled

Reported by Marcus Ionker on the Camel users mailing list.

## Test plan
- [x] New test `PollDynamicFileNameExchangePropertyTest` — two cases: matching file found, and no matching file (returns null body)
- [x] All existing related tests pass (`PollDynamicFileNameOptimizeDisabledTest`, `PollEnricherFileTest`, `PollEnricherFileMaxMessagesPerPollTest`, `PollEnrichHeaderPropagationTest`, `PollEnrichSimpleMoveTest`)

### DIFF
--- a/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
+++ b/components/camel-file/src/main/java/org/apache/camel/component/file/GenericFileHelper.java
@@ -94,9 +94,12 @@ public final class GenericFileHelper {
             // enrich with data from dynamic source
             if (dynamic.getMessage().hasHeaders()) {
                 MessageHelper.copyHeaders(dynamic.getMessage(), dummy.getMessage(), true);
-                if (dynamic.hasVariables()) {
-                    dummy.getVariables().putAll(dynamic.getVariables());
-                }
+            }
+            if (dynamic.hasVariables()) {
+                dummy.getVariables().putAll(dynamic.getVariables());
+            }
+            if (dynamic.hasProperties()) {
+                dummy.getProperties().putAll(dynamic.getProperties());
             }
         }
         return dummy;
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — clean, well-scoped fix.

Two issues addressed:

1. **Exchange properties not copied** — `GenericFileHelper.createDummy()` now copies exchange properties to the dummy exchange, so `${exchangeProperty.target}` in a `pollEnrich` `fileName` expression resolves correctly during the `PollDynamicAware` optimization. Without this, the expression evaluated to `null` on the dummy and the filename filter was effectively bypassed.

2. **Variables copy bug** — The `dynamic.hasVariables()` block was previously nested inside the `dynamic.getMessage().hasHeaders()` guard, meaning variables wouldn't be copied for exchanges with variables but no headers. Now correctly moved outside.

Test is well-designed — verifies both the filename resolution and endpoint optimization reuse, with a negative case (non-existent file → null body).

PR #24995, opened 2026-07-21
CAMEL-24230: Fix jsonpath writeAsString for object expressions

atiaomar1978-hub:
## Summary

Fixes [CAMEL-24230](https://issues.apache.org/jira/browse/CAMEL-24230): with `writeAsString=true`, a JsonPath expression that evaluates to a single JSON object (e.g. `$.args`) now returns a valid JSON **String** instead of a `java.util.Map` whose `toString()` looks like `{age=30, name="Alice"}`.

**Root cause:** `JsonPathEngine` had a special `Map` branch that only stringified individual values and returned the Map itself. Non-iterable results (including Maps) now use `adapter.writeAsString(answer, exchange)` like scalars already did.

**Regression fix:** `JsonPathSplitWriteAsStringMapTest` updated to split on `$.content.*` (array of object values) instead of relying on the old Map-entry behaviour.

**Upgrade guide:** Added a `camel-jsonpath` entry to `camel-4x-upgrade-guide-4_22.adoc` documenting the behavioral change and migration path for split routes that relied on the old Map behaviour.

## Test plan

- [x] `JsonPathWriteAsStringObjectTest` (4) — route-level setHeader/setBody with `jsonpathWriteAsString("$.args")`, nested object, valid JSON output (AssertJ + JsonNode comparison)
- [x] `JsonPathEngineWriteAsStringTest` (3) — engine unit tests: object → JSON string, flag off → Map, array → List of JSON strings
- [x] Updated `JsonPathSplitWriteAsStringMapTest` — split on `$.content.*` expects String bodies
- [x] Full module: `mvn test -pl components/camel-jsonpath` — **119 tests passed**
- [x] Code formatted (`formatter:format`, `impsort:sort`)

---

_Generated by Cursor Agent on behalf of @atiaomar1978-hub._

### DIFF
--- a/components/camel-jsonpath/pom.xml
+++ b/components/camel-jsonpath/pom.xml
@@ -100,6 +100,11 @@
             <version>${hamcrest-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathEngine.java
+++ b/components/camel-jsonpath/src/main/java/org/apache/camel/jsonpath/JsonPathEngine.java
@@ -152,18 +152,6 @@ public class JsonPathEngine {
                     }
                 }
                 return list;
-            } else if (answer instanceof Map) {
-                Map<Object, Object> map = (Map<Object, Object>) answer;
-                for (Map.Entry<Object, Object> entry : map.entrySet()) {
-                    Object value = entry.getValue();
-                    if (adapter != null) {
-                        String json = adapter.writeAsString(value, exchange);
-                        if (json != null) {
-                            map.put(entry.getKey(), json);
-                        }
-                    }
-                }
-                return map;
             } else {
                 String json = adapter.writeAsString(answer, exchange);
                 if (json != null) {
--- a/components/camel-jsonpath/src/test/java/org/apache/camel/jsonpath/JsonPathEngineWriteAsStringTest.java
+++ b/components/camel-jsonpath/src/test/java/org/apache/camel/jsonpath/JsonPathEngineWriteAsStringTest.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.jsonpath;
+
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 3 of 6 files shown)


gnodet (approved):

## Review Summary

Clean, well-targeted fix for CAMEL-24230. The production change removes a 12-line special-case branch that was the root cause, letting Map results flow through the existing scalar serialization path. Good test coverage across three test classes.

## What I Love

1. **Minimal, surgical production change** -- rather than adding complexity, the fix removes the problematic `Map` branch entirely, letting Maps fall through to the existing `writeAsString` path that already handles scalars correctly. Less code, fewer branches, correct behavior. The history backs this up: the `Map` branch was introduced in CAMEL-11558 (2017) as a workaround because the split EIP couldn't handle Maps at the time. CAMEL-17101 (2021) taught the split EIP to handle Maps properly, making the workaround obsolete -- but the branch stayed, creating this inconsistency.

2. **Thorough test strategy** -- the PR includes both unit-level tests (`JsonPathEngineWriteAsStringTest` testing the engine directly) and integration-level tests (`JsonPathWriteAsStringObjectTest` testing via Camel routes with `setHeader`/`setBody`). The `writeAsStringDoesNotReturnMapToString` test with `.doesNotContain("=")` is a particularly good regression guard -- it catches exactly the symptom from the bug report (`{age=30, name=Alice}` instead of valid JSON).

3. **Regression-aware test update** -- updating `JsonPathSplitWriteAsStringMapTest` to use `$.content.*` instead of `$.content` shows understanding of the full impact. The wildcard expression returns a List (handled by the `Iterable` branch), which is the correct way to get splittable JSON strings from a Map-shaped document.

## Findings

### Critical

None -- nice work!

### Important

**[Important] Consider an upgrade guide entry for the behavioral change.** Per project conventions, user-visible changes should be documented in the upgrade guide. This fix changes observable behavior: previously, `writeAsString=true` with a Map-typed JsonPath result returned a `Map` with individually stringified values; now it returns a JSON `String` of the entire object.

Users who relied on the old behavior with `.split().jsonpathWriteAsString("$.expression-returning-map")` will need to change their expression to use a wildcard (e.g., `$.content.*` instead of `$.content`) to get a splittable list of JSON strings. A brief note in `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc` would help:

```asciidoc
=== camel-jsonpath

The `writeAsString` option now correctly serializes single JSON object results (Maps) to a JSON String.
Previously, a JsonPath expression evaluating to a JSON object (e.g. `$.args`) with `writeAsString=true`
would return a `java.util.Map` with individually stringified values instead of a valid JSON String.
If you were relying on the old behavior to split a Map result, change the expression to use a wildcard
(e.g. `$.content.*` instead of `$.content`) to get a splittable list of JSON strings.
```

### Suggestions & Nits

None -- the code is clean, well-formatted, and follows project conventions. The `assertj-core` test dependency is managed in the parent BOM, and the new tests use tree-based JSON comparison (`ObjectMapper.readTree`) where key order could vary, which is the right approach.

## Overall

Solid contribution that fixes a real user-facing inconsistency in the `writeAsString` contract. The approach of removing incorrect code rather than adding workarounds is exactly right. Thank you for the thorough tests, @atiaomar1978-hub!

atiaomar1978-hub:
Thanks @gnodet for the review and approval.

### Addressed feedback

**[Important] Upgrade guide entry** — done in commit `581501b`.

Added a `=== camel-jsonpath` section to `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc` documenting:

- `writeAsString` now correctly serializes single JSON object results (Maps) to a valid JSON `String`
- The previous behavior returned a `java.util.Map` with individually stringified values (the CAMEL-24230 bug)
- Migration note for routes that split a Map result: use a wildcard expression (e.g. `$.content.*` instead of `$.content`) to obtain a splittable list of JSON strings

The PR description has also been updated to reflect this addition.

No further code changes were required beyond the original fix and tests.

---
_Cursor Agent on behalf of @atiaomar1978-hub_

davsclaus (approved):

PR #24996, opened 2026-07-21
CAMEL-24232: camel-core - Copy variables to transform exchange in ProcessorTransformer

davsclaus:
## Summary

Fixes [CAMEL-24232](https://issues.apache.org/jira/browse/CAMEL-24232).

`ProcessorTransformer.transform()` creates a copy of the exchange for the transform processor and copies exchange properties, but does not copy variables. If the transform processor references `${variable.xxx}`, it gets `null` instead of the actual value.

Found while investigating [CAMEL-24231](https://issues.apache.org/jira/browse/CAMEL-24231) (same class of bug: incomplete data copy to a scratch exchange).

**Changes:**
- Copy variables from the source exchange to the transform exchange in `ProcessorTransformer.transform()`
- Add test `ProcessorTransformerVariableTest` verifying both variables and properties are available in the transform processor

## Test plan
- [x] New test `ProcessorTransformerVariableTest` — two cases: variables and exchange properties accessible in transform processor
- [x] All existing transformer tests pass (`TransformerRouteTest`, `TransformerContractTest`, `TransformerBuilderTest`, `TransformDataTypeProcessorTest`)

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/transformer/ProcessorTransformer.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/transformer/ProcessorTransformer.java
@@ -73,6 +73,9 @@ public class ProcessorTransformer extends Transformer {
         Exchange transformExchange = new DefaultExchange(exchange);
         transformExchange.setIn(message);
         transformExchange.getExchangeExtension().setProperties(exchange.getProperties());
+        if (exchange.hasVariables()) {
+            transformExchange.getVariables().putAll(exchange.getVariables());
+        }
         processor.process(transformExchange);
         Message answer = transformExchange.getMessage();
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):

**LGTM** — clean fix, same class of bug as CAMEL-24231.

`ProcessorTransformer.transform()` creates a new `DefaultExchange` and copies properties via `setProperties()`, but was missing the variables copy. The fix correctly adds a `hasVariables()` guard before `putAll()`, matching the pattern used in CAMEL-24231's fix for `GenericFileHelper.createDummy()`.

Test is well-structured — directly exercises `ProcessorTransformer` at the unit level, verifying both variables and properties are accessible in the transform processor. Good use of try-with-resources for `CamelContext` lifecycle.

One minor observation: the test uses FQCN `new org.apache.camel.impl.DefaultCamelContext()` on line 51/75 — per project convention, prefer an import + simple class name. The build's OpenRewrite pass will catch this automatically, so not blocking.

PR #24997, opened 2026-07-21
CAMEL-24234: strengthen gzip auto-decompression regression coverage (no production change)

atiaomar1978-hub:
# CAMEL-24234: gzip double-decompression with HttpClient 5.6+

## Update after @oscerd's review

@oscerd was right on every point. After investigating, I've reworked this PR substantially:

**The class of bug is already fixed on `main`** by commit `4a903e2b8f3c`, which decides purely from the entity actually observed (`entity.getContentEncoding()`). In httpclient5 5.6.2, `DecompressingEntity.getContentEncoding()` returns `null` after auto-decompression, so:
- default (auto-decompress): entity reports `null` -> existing `removeStaleCompressionHeaders` strips the stale header -> `uncompressGzip(null, is)` is a no-op -> plain body. Correct.
- `contentCompressionDisabled` / endpoint-level disable: entity reports `gzip` -> Camel decompresses. Correct.

I verified this empirically: **the original tests pass against an unmodified `HttpProducer.java`**, confirming they documented existing behaviour rather than protecting a fix.

**The component-flag guard was a regression.** `contentCompressionDisabled` is component-level only, so it could not see the three endpoint-level ways decompression is actually disabled (`?httpClient.contentCompressionEnabled=false`, a user `httpClientConfigurer` calling `disableContentCompression()`, or a user-supplied client). In those cases the guard would have stripped `Content-Encoding` and skipped `GZIPHelper`, delivering raw gzip bytes downstream — silent data corruption.

## What this PR now does

No production change. It reverts `HttpProducer.java` to the correct entity-based logic on `main` and instead **strengthens regression coverage** for CAMEL-24234, consolidated into `HttpCompressionTest` (no duplicated interceptor/class):

- `compressedHttpPost()` now also asserts the stale gzip `Content-Encoding` header is stripped after auto-decompression (migrated to AssertJ).
- New `compressedHttpPostWithEndpointContentCompressionDisabled` covers the `?httpClient.contentCompressionEnabled=false` path — exactly the case the reverted guard would have corrupted.
- Removed the duplicated standalone `HttpProducerGzipAutoDecompressionTest` (~90 lines).

`mvn test -pl components/camel-http -Dtest=HttpCompressionTest` passes (3 tests).

Since there is no behaviour change, no upgrade-guide entry is needed.

---


gnodet (approved):

**LGTM** — clean fix for double gzip decompression with HttpClient 5.6+.

The fix correctly addresses the two failure modes:

1. **`extractResponseBody()`** — When HttpClient auto-decompression is enabled (the default), the stream is already plain text. The new `httpClientDecompresses` guard skips Camel's `GZIPHelper.uncompressGzip()`, preventing the `ZipException: Not in GZIP format`. The fallback to the response header when `entity.getContentEncoding()` returns null is a good defensive measure for the `contentCompressionDisabled=true` path.

2. **`removeStaleCompressionHeaders()`** — The broadened condition (`httpClientDecompresses || entity.getContentEncoding() == null`) correctly strips stale `Content-Encoding`, `Content-Length`, and `Content-MD5` headers in both scenarios: when HttpClient auto-decompressed (stream is plain but headers remain) and when the entity wrapper no longer reports encoding (5.6+ behavior).

The test is well-designed — a `ResponseCompressingInterceptor` that actually gzip-compresses the response body simulates the real-world scenario faithfully, and both auto-decompression (default) and manual decompression (`contentCompressionDisabled=true`) paths are exercised.

Minor note: the `@TestMethodOrder(OrderAnnotation)` + `@Order` pattern introduces an implicit ordering dependency (test 1 toggles `contentCompressionDisabled` and restores it in `finally`). This is fine given the try/finally cleanup, but a `@BeforeEach` reset could make it more resilient.

atiaomar1978-hub:
Thanks @gnodet for the review and approval.

Addressed the minor note by adding a `@BeforeEach` `resetContentCompressionDisabled()` hook so the HTTP component flag is reset before each test, even if a prior test fails before its `finally` block runs. Kept `@Order` so the `contentCompressionDisabled=true` case still runs first and gets a fresh `HttpClient` without auto-decompression.

Tests pass locally.

---

davsclaus (approved):

oscerd (dismissed):
Thanks for chasing this down — the underlying diagnosis is right. HttpClient 5.6 stopped removing the stale `Content-Encoding` / `Content-Length` / `Content-MD5` headers that 5.5.x removed after decompressing, which is exactly the kind of upstream change that produces double decompression.

But I think the guard this PR adds keys off the wrong signal, and in a few supported configurations it turns a working case into silent data corruption. I'd like that resolved before it goes in.

### The guard can't see three real ways decompression gets disabled

The new condition is:

```java
final boolean httpClientDecompresses = !getEndpoint().getComponent().isContentCompressionDisabled();
```

`contentCompressionDisabled` is a **component-level option only** — I confirmed this in the generated catalog: it appears under `componentProperties` and is absent from the endpoint `properties` map. So it stays `false` in all three of these, even though HttpClient is genuinely not decompressing:

1. **`?httpClient.contentCompressionEnabled=false`** on the endpoint. `HttpComponent` binds `httpClient.*` parameters to *both* the client builder and `RequestConfig.Builder`, and `RequestConfig.isContentCompressionEnabled()` is what gates HttpClient's decompression exec.
2. **A user `httpClientConfigurer`** calling `clientBuilder.disableContentCompression()` — applied from `HttpEndpoint`.
3. **A user-supplied client** via the `httpClient` option (`HttpEndpoint.setHttpClient`).

In each case, after this PR: `removeStaleCompressionHeaders(..., true)` strips `Content-Encoding`, and `extractResponseBody` skips `GZIPHelper.uncompressGzip`. The route then receives **raw gzip bytes with no `Content-Encoding` header to indicate it** — so nothing downstream can even detect the problem. Before this PR, `entity.getContentEncoding()` returned `"gzip"` and Camel decompressed correctly.

The fix direction I'd suggest is to keep deciding from the entity actually observed rather than from a component flag, which is what the current code already does.

### It partially reverts a prior intentional fix

This is worth calling out explicitly per the project's investigation rules. Commit `4a903e2b8f3c` ("Align dependencies with Spring Boot 4.1.0") deliberately replaced the response-header read with `entity.getContentEncoding()` and added `removeStaleCompressionHeaders` — i.e. it already fixed this class of bug, on purpose, by moving *away* from the response header. This PR reintroduces the response-header read as a fallback. That may still be the right call, but it needs to be acknowledged and justified in the description rather than landing silently.

Relatedly, that fallback appears to be **unreachable**: `removeStaleCompressionHeaders` runs before `populateResponse` → `extractResponseBody`, and in the only branch where the fallback could fire it has already removed `Content-Encoding` whenever `entity.getContentEncoding()` was null. So `httpResponse.getFirstHeader(CONTENT_ENCODING)` is always null there. It reads as a defensive measure but is inert.

### The tests don't fail without the fix

This is the part I'd most like you to check, because it changes how much confidence the PR carries. `DecompressingEntity.getContentEncoding()` returns `null` in httpclient5 5.6.2. On unmodified `main` with the default config that means: HttpClient wraps the entity → `getContentEncoding()` is null → the *existing* `removeStaleCompressionHeaders` strips the header → `extractResponseBody` calls `uncompressGzip(null, is)`, a no-op. All three assertions in `shouldReadPlainBodyWhenHttpClientAlreadyDecompressedWithStaleGzipHeaders` should therefore already hold.

The quickest way to settle it: run the new test class against an unmodified `HttpProducer.java`. I believe both tests pass. If they do, they're documenting current behaviour rather than protecting the fix.

The `contentCompressionDisabled=true` case is also already covered — it's essentially `HttpCompressionTest.compressedHttpPostWithAutoDecompressionDisabled`.

### Smaller points

- The new `ResponseCompressingInterceptor` and its nested `GzipCompressingEntity` are a near-verbatim copy of the ones in `HttpCompressionTest`, and the two scenarios duplicate two existing tests there. The only genuinely new assertion is `Content-Encoding == null`. Adding that assertion to `HttpCompressionTest`, or extracting the interceptor to a shared helper, would avoid ~90 duplicated lines.
- Worth confirming the reproduction first: the stack trace on CAMEL-24234 has line numbers that don't match current `main` (`extractResponseBody` at `:543` vs `:555` today), which suggests the reporter is on a build predating `4a903e2b8f3c` — the commit that already fixed this. A comment on the issue also mentions it arriving via `camel-platform-http-starter`, where the managed httpclient5 version may differ. Knowing which Camel + httpclient5 combination actually reproduces would tell us whether a second guard is needed at all.
- With `@BeforeEach resetContentCompressionDisabled()` added, the `@TestMethodOrder` / `@Order` annotations are now redundant.
- If the behaviour change is kept, it needs an upgrade-guide entry — the PR touches no docs.

Requesting changes on the first point only; the rest is discussion. The instinct to guard this is right, I just think the guard needs to come from the entity rather than the component flag so it can't disagree with what the client actually did.

---

components/camel-http/src/main/java/org/apache/camel/component/http/HttpProducer.java
```diff
@@ -548,10 +553,16 @@ protected Object extractResponseBody(
             return null;
         }
 
-        String contentEncoding = entity.getContentEncoding();
-
-        final boolean gzipEncoding = exchange.getProperty(Exchange.SKIP_GZIP_ENCODING, Boolean.FALSE, Boolean.class);
-        if (!gzipEncoding) {
+        final boolean skipGzip = exchange.getProperty(Exchange.SKIP_GZIP_ENCODING, Boolean.FALSE, Boolean.class);
+        final boolean httpClientDecompresses = !getEndpoint().getComponent().isContentCompressionDisabled();
```
gnodet: **Concern: component-level flag vs. entity-level check** (high)

`httpClientDecompresses` is derived from `!getEndpoint().getComponent().isContentCompressionDisabled()` — a component-level flag. However, HttpClient auto-decompression can be disabled through at least three other paths that bypass this flag:

1. Endpoint-level `httpClient.contentCompressionEnabled=false` (via `PropertyBindingSupport`)
2. A user `httpClientConfigurer` calling `clientBuilder.disableContentCompression()`
3. A user-supplied `HttpClient` via `setHttpClient()`

When any of these bypass the component flag, `httpClientDecompresses` would be `true` but the body is still gzip-compressed. The code would then skip `GZIPHelper.uncompressGzip`, delivering raw gzip bytes.

The existing code on `main` avoids this by checking `entity.getContentEncoding()` — which reflects what HttpClient **actually did** to the stream, regardless of how it was configured.

components/camel-http/src/main/java/org/apache/camel/component/http/HttpProducer.java
```diff
@@ -548,10 +553,16 @@ protected Object extractResponseBody(
...
+        if (!skipGzip && !httpClientDecompresses) {
+            String contentEncoding = entity.getContentEncoding();
+            if (contentEncoding == null) {
+                Header encodingHeader = httpResponse.getFirstHeader(Exchange.CONTENT_ENCODING);
```
gnodet: **Dead code path** (medium)

This response-header fallback is unreachable in practice. This branch only executes when `httpClientDecompresses` is `false` AND `entity.getContentEncoding()` is `null`. But `removeStaleCompressionHeaders` runs first (line 266), and when `entity.getContentEncoding()` is `null`, the broadened condition `httpClientDecompresses || entity.getContentEncoding() == null` evaluates to `false || true = true`, so it strips the `Content-Encoding` response header. Therefore `httpResponse.getFirstHeader(Exchange.CONTENT_ENCODING)` will always return `null` in this path.

gnodet (commented):
## Review of CAMEL-24234: Fix double gzip decompression with HttpClient 5.6+

Hi @atiaomar1978-hub — great diagnosis of the underlying problem! You've correctly identified that HttpClient 5.6+ auto-decompresses response bodies but leaves stale `Content-Encoding` headers, which then triggers a second (erroneous) decompression in Camel.

However, after investigating the git history and current `main` branch, I believe this fix is **already present on `main`** and the PR should instead target `camel-4.18.x` as a backport.

### Key finding: already fixed on `main`

Commit `4a903e2b8f3c` (June 2026, "Align dependencies with Spring Boot 4.1.0") already addressed this scenario on `main` by:
1. Switching `extractResponseBody` from reading `httpResponse.getFirstHeader(CONTENT_ENCODING)` to `entity.getContentEncoding()` — when HttpClient 5.6+ auto-decompresses, `DecompressingEntity.getContentEncoding()` returns `null`, so `GZIPHelper.uncompressGzip(null, is)` is a no-op.
2. Adding `removeStaleCompressionHeaders()` to strip the stale headers so downstream code sees consistent state.

The JIRA reporter's stack trace (referencing line 543) matches the pre-fix code on `camel-4.18.x`, not current `main`. The reporter is on Camel 4.18.3, which likely picks up httpclient5 5.6+ transitively via a Spring Boot BOM.

### Suggested path forward

The correct fix for 4.18.x would be a targeted **backport** of the entity-based approach from `4a903e2b8f3c` to `camel-4.18.x`, rather than additional changes to `main`.

See inline comments for details on specific concerns with the current approach.

---

davsclaus:
yes a backport of the main fix is correct, so we can close this PR and do that instead

davsclaus (changes requested):
my prev comment

atiaomar1978-hub:
@davsclaus should I close this PR ?

atiaomar1978-hub:
Thanks @oscerd — you were right on every point, and I appreciate the depth of the review.

I verified your central claim empirically: I ran the new test class against an **unmodified** `HttpProducer.java` and both tests passed. So they were documenting current behaviour, not protecting a fix — the class of bug is already fixed on `main` by `4a903e2b8f3c` (entity-based `getContentEncoding()`), and `DecompressingEntity.getContentEncoding()` returns `null` in httpclient5 5.6.2 as you said.

I also confirmed the guard was a regression: `contentCompressionDisabled` is component-level only, so `?httpClient.contentCompressionEnabled=false` (bound to `RequestConfig.Builder` in `createHttpClientBuilder`), a `httpClientConfigurer` calling `disableContentCompression()`, and a user-supplied client all leave it `false` while HttpClient genuinely isn't decompressing — my guard would have stripped `Content-Encoding` and skipped `GZIPHelper`, handing raw gzip downstream.

Changes pushed (86b54d1):

- **Reverted** `HttpProducer.java` to the entity-based logic on `main` — no production change; the guard is gone, and with it the unreachable response-header fallback you flagged.
- **Removed** the duplicated standalone test class (~90 lines).
- **Consolidated** regression coverage into `HttpCompressionTest`: `compressedHttpPost()` now asserts the stale gzip `Content-Encoding` header is stripped after auto-decompression (migrated to AssertJ), and I added `compressedHttpPostWithEndpointContentCompressionDisabled` covering the `?httpClient.contentCompressionEnabled=false` path — the exact case the guard would have corrupted.
- No upgrade-guide entry, since there's no behaviour change.

`mvn test -pl components/camel-http -Dtest=HttpCompressionTest` passes.

If you'd prefer to close this in favour of the existing coverage rather than land the extra endpoint-level test, I'm happy either way.

gnodet (commented):
Thanks @atiaomar1978-hub for the quick and thorough response to the feedback! The updated PR correctly reverts the production code changes and keeps only the test improvements.

The two new test additions have genuine value:
- The `Content-Encoding == null` assertion in `compressedHttpPost()` is a focused regression guard for `removeStaleCompressionHeaders`
- `compressedHttpPostWithEndpointContentCompressionDisabled` covers the endpoint-level `httpClient.contentCompressionEnabled=false` path that was not previously exercised

One minor note: the PR introduces AssertJ assertions for `compressedHttpPost()` while `compressedHttpPostWithAutoDecompressionDisabled()` still uses JUnit `assertEquals`/`assertNotNull`. If the intent is to migrate to AssertJ, the other method should follow for consistency within the same file — otherwise keeping JUnit throughout would be simpler.

---

atiaomar1978-hub:
@oscerd @davsclaus @gnodet — this is ready for re-review. All review feedback has been addressed:

- **@oscerd's blocking concern (guard keyed off the wrong signal):** the component-flag guard has been fully removed. `HttpProducer.java` is now byte-for-byte identical to `main` (`git diff main -- HttpProducer.java` is empty), so the decompression decision comes purely from `entity.getContentEncoding()` — what the client actually did — and the three endpoint-level disable paths (`?httpClient.contentCompressionEnabled=false`, a `httpClientConfigurer` calling `disableContentCompression()`, and a user-supplied client) can no longer be misread into silent data corruption.
- **@gnodet's inline comments (wrong signal + dead code path):** both resolved by the revert above — the guard and the unreachable response-header fallback are gone.
- **Prior-fix acknowledgement:** the description now explicitly credits `4a903e2b8f3c` as the commit that already fixed the production bug on `main`.
- **Tests documenting vs protecting:** confirmed — the assertions pass against unmodified `HttpProducer.java`. The PR is now scoped as **regression coverage only, no production change**.
- **Duplication:** the standalone `HttpProducerGzipAutoDecompressionTest` (~90 lines) was removed; coverage is consolidated into `HttpCompressionTest`, including a new `compressedHttpPostWithEndpointContentCompressionDisabled` test for the exact config the reverted guard would have corrupted.
- **Upgrade guide / `@TestMethodOrder`:** not needed — no behaviour change, and the ordering annotations are gone.

`mvn test -pl components/camel-http -Dtest=HttpCompressionTest` → 3 tests, all green.

Separately, @gnodet raised that the reporter is on 4.18.x — happy to open a targeted backport of `4a903e2b8f3c` to `camel-4.18.x` as a follow-up if that's the preferred path.

_Claude Opus 4.8 on behalf of atiaomar1978-hub_

atiaomar1978-hub:
Thanks @gnodet! Addressed the AssertJ consistency note — migrated the remaining JUnit assertions in `compressedHttpPostWithAutoDecompressionDisabled()` (`assertNotNull`/`assertEquals`) to AssertJ so all three tests in `HttpCompressionTest` use `assertThat(...)` consistently, and removed the now-unused `assertEquals`/`assertNotNull` imports.

Pushed in d8f1d7f. `mvn test -pl components/camel-http -Dtest=HttpCompressionTest` → Tests run: 3, Failures: 0, Errors: 0.

_Claude Opus 4.8 on behalf of atiaomar1978-hub_

gnodet (approved):
Re-reviewed after latest commit. The new commit addresses the AssertJ consistency note from the previous review — `compressedHttpPostWithAutoDecompressionDisabled()` now uses AssertJ assertions uniformly with the rest of the file. All 11 assertions use `assertThat` consistently, with zero JUnit assertion imports remaining. Clean test-only PR. LGTM.

davsclaus (approved):

davsclaus:
okay a backport of 4a903e2b8f3c and this test would be appreciated

PR #25001, opened 2026-07-22
CAMEL-23938: add Camel-managed session lock renewal for session-enabled Service Bus consumers

atiaomar1978-hub:
# CAMEL-23938: Camel-managed session lock renewal for session-enabled Service Bus consumers

Follow-up from CAMEL-23937.

## Problem

CAMEL-23937 added Camel-managed lock renewal for **non-session** Azure Service Bus consumers using `ServiceBusReceiverAsyncClient.renewMessageLock()`. Session-enabled consumers were **explicitly excluded**, because session locks require a different API:

- Session locks must be renewed via `renewSessionLock()` instead of `renewMessageLock()`
- This requires a session-bound receiver (`acceptSession(sessionId)`)
- The session id is only known at message-receive time, not at consumer startup

As a result, session-enabled consumers in async routes still suffered from silent session-lock expiry mid-processing.

## Solution

Added a `SessionLockRenewer` in `ServiceBusConsumer` that:

1. Creates session-bound async receivers on demand, keyed by session id, via `ServiceBusSessionReceiverAsyncClient.acceptSession(sessionId)`
2. Renews session locks with `renewSessionLock()` as they approach expiry
3. Reference-counts in-flight exchanges per session and closes the session receiver once the last exchange for that session completes
4. Honours `maxAutoLockRenewDuration` per exchange

Also added `ServiceBusClientFactory.createServiceBusSessionReceiverAsyncClient()`.

The behavior activates automatically when: no custom `processorClient`, `PEEK_LOCK` mode, `maxAutoLockRenewDuration > 0`, and session mode is enabled.

## Tests

Added AssertJ test coverage in `ServiceBusConsumerTest`:

- Session lock renewer created for session-enabled PEEK_LOCK, not for RECEIVE_AND_DELETE
- Exchange tracking and cleanup (session receiver closed after completion)
- `renewSessionLock()` invoked when the lock is expiring
- Session receiver reused across messages of the same session
- Session receiver kept open until the last exchange for the session completes
- Stop/start lifecycle recreates and cleans up the session renewer

`mvn test -pl components/camel-azure/camel-azure-servicebus` passes.

## Docs

Updated the 4.22 upgrade guide entry to cover session lock renewal.

---
_Claude Opus 4.8 on behalf of atiaomar1978-hub_

### DIFF
--- a/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
+++ b/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
@@ -452,7 +452,9 @@ public class ServiceBusConsumer extends DefaultConsumer implements ShutdownAware
                 if (elapsed.compareTo(maxRenewDuration) >= 0) {
                     LOG.debug("Max session lock renewal duration exceeded for exchangeId {}, stopping renewal",
                             exchangeId);
-                    entries.remove(exchangeId);
+                    if (entries.remove(exchangeId) != null) {
+                        releaseSessionReceiver(entry.sessionId);
+                    }
                     continue;
                 }
 
@@ -476,21 +478,21 @@ public class ServiceBusConsumer extends DefaultConsumer implements ShutdownAware
                                     error -> {
                                         LOG.warn("Failed to renew session lock for sessionId {}, exchangeId {}: {}",
                                                 entry.sessionId, exchangeId, error.getMessage());
-                                        entries.remove(exchangeId);
+                                        if (entries.remove(exchangeId) != null) {
+                                            releaseSessionReceiver(entry.sessionId);
+                                        }
                                     });
                 }
             }
         }
 
+        // Acquire and release are both guarded by sessionReceiverCreationLock so the
+        // increment/create and decrement/close operations on a SessionReceiverHolder are
+        // atomic relative to each other. This prevents a race where a newly arriving message
+        // increments a holder that a concurrently completing exchange is about to close.
         private ServiceBusReceiverAsyncClient acquireSessionReceiver(String sessionId) {
-            SessionReceiverHolder existing = sessionReceivers.get(sessionId);
-            if (existing != null) {
-                existing.inFlightCount.incrementAndGet();
-                return existing.client;
-            }
-
             synchronized (sessionReceiverCreationLock) {
-                existing = sessionReceivers.get(sessionId);
+                SessionReceiverHolder existing = sessionReceivers.get(sessionId);
                 if (existing != null) {
                     existing.inFlightCount.incrementAndGet();
                     return existing.client;
@@ -510,14 +512,16 @@ public class ServiceBusConsumer extends DefaultConsumer implements ShutdownAware
         }
 
         private void releaseSessionReceiver(String sessionId) {
-            SessionReceiverHolder holder = sessionReceivers.get(sessionId);
-            if (holder == null) {
-                return;
-            }
+            synchronized (sessionReceiverCreationLock) {
+                SessionReceiverHolder holder = sessionReceivers.get(sessionId);
+                if (holder == null) {
+                    return;
+                }
 
-            if (holder.inFlightCount.decrementAndGet() <= 0) {
-                sessionReceivers.remove(sessionId, holder);
-                holder.client.close();
+                if (holder.inFlightCount.decrementAndGet() <= 0) {
+                    sessionReceivers.remove(sessionId);
+                    holder.client.close();
+                }
             }
         }
 
--- a/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
+++ b/components/camel-azure/camel-azure-servicebus/src/main/java/org/apache/camel/component/azure/servicebus/ServiceBusConsumer.java
@@ -31,6 +31,7 @@ import com.azure.messaging.servicebus.ServiceBusProcessorClient;
 import com.azure.messaging.servicebus.ServiceBusReceivedMessage;
 import com.azure.messaging.servicebus.ServiceBusReceivedMessageContext;
 import com.azure.messaging.servicebus.ServiceBusReceiverAsyncClient;
+import com.azure.messaging.servicebus.ServiceBusSessionReceiverAsyncClient;
 import com.azure.messaging.servicebus.models.DeadLetterOptions;
 import com.azure.messaging.servicebus.models.ServiceBusReceiveMode;
 import com.azure.messaging.servicebus.models.SubQueue;
@@ -57,7 +58,9 @@ public class ServiceBusConsumer extends DefaultConsumer implements ShutdownAware
     private ServiceBusProcessorClient client;
     private boolean clientCloseable;
     private ServiceBusReceiverAsyncClient renewalClient;
+    private ServiceBusSessionReceiverAsyncClient sessionRenewalClient;
     private LockRenewer lockRenewer;
+    private SessionLockRenewer sessionLockRenewer;
     private final AtomicInteger pendingExchanges = new AtomicInteger();
 
     public ServiceBusConsumer(final ServiceBusEndpoint endpoint, final Processor processor) {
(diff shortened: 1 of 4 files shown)


gnodet (commented):

## Nice extension — session lock renewal completes the picture

Good follow-up to CAMEL-23937. The architecture mirrors the existing `LockRenewer` cleanly, with the added complexity of session-bound receivers and reference counting properly justified by the Azure SDK's session lock API requirements. Tests are thorough (7 new tests covering the key scenarios). Upgrade guide updated.

### One bug to fix

**Session receiver leak when `maxRenewDuration` is exceeded or `renewSessionLock()` fails**

In `SessionLockRenewer.run()`, two code paths remove entries from the `entries` map without calling `releaseSessionReceiver()`:

1. Max duration exceeded (line ~178):
```java
entries.remove(exchangeId);
continue;
```

2. Renewal error handler (line ~199):
```java
entries.remove(exchangeId);
```

When `run()` removes the entry first, the `SynchronizationAdapter.onComplete/onFailure` callback later gets `null` from `entries.remove()` and skips `releaseSessionReceiver()`. This means the session receiver's `inFlightCount` is never decremented, so the session receiver is never closed — a resource leak.

The existing `LockRenewer` has the same `entries.remove()` pattern, but it doesn't matter there because it uses a single shared client with no reference counting. `SessionLockRenewer` has reference-counted `SessionReceiverHolder`s that need explicit release.

**Fix**: call `releaseSessionReceiver(entry.sessionId)` alongside each `entries.remove(exchangeId)` in `run()`:

```java
// max duration exceeded
entries.remove(exchangeId);
releaseSessionReceiver(entry.sessionId);
continue;
```
```java
// error handler
error -> {
    LOG.warn(...);
    entries.remove(exchangeId);
    releaseSessionReceiver(entry.sessionId);
}
```

### Minor observation (non-blocking)

The `acquireSessionReceiver` fast path (outside the `synchronized` block) can theoretically race with `releaseSessionReceiver`: thread A decrements to 0 and is about to `remove + close`, while thread B reads the holder, increments to 1, and returns the soon-to-be-closed client. In practice this requires a new message for the same session arriving at the exact instant the last exchange completes — unlikely but possible under load. The synchronized block only guards creation, not the acquire-vs-release interplay. Worth noting for a future hardening pass, but not a blocker for this PR.

### Everything else looks good

- `ServiceBusClientFactory.createServiceBusSessionReceiverAsyncClient()` correctly mirrors the receiver client setup with `sessionReceiver()` builder, `disableAutoComplete()`, and `maxAutoLockRenewDuration(Duration.ZERO)`
- `doStop()` and `doShutdown()` properly clean up both session and non-session resources (consistent with existing pattern)
- `Boolean.TRUE.equals(isSessionEnabled())` is null-safe ✓
- Test coverage: session receiver reuse, reference counting lifecycle, stop/start cycle recreation — all the important scenarios

atiaomar1978-hub:
Thanks for the thorough review, @gnodet!

**Session receiver leak (fixed)** — you're right. `SessionLockRenewer.run()` removed entries from the tracking map on the max-duration and renewal-error paths without releasing the reference-counted session receiver, so the completion callback later saw a `null` entry and skipped `releaseSessionReceiver()`, leaving `inFlightCount` above zero and the receiver open. Both paths now call `releaseSessionReceiver(entry.sessionId)` guarded by `entries.remove(exchangeId) != null` so the atomic `remove` guarantees exactly one release regardless of whether `run()` or the completion callback wins the race.

**Acquire/release race (hardened)** — also addressed. `acquireSessionReceiver()` and `releaseSessionReceiver()` are now both guarded by the same `sessionReceiverCreationLock`, making the increment/create and decrement/close operations atomic relative to each other. A newly arriving message can no longer increment a holder that a concurrently completing exchange is about to close.

Added a test (`sessionLockRenewerReleasesSessionReceiverOnRenewalFailure`) verifying the receiver is closed when `renewSessionLock()` fails.

Pushed in b2042a5. `mvn test -pl components/camel-azure/camel-azure-servicebus` passes.

_Claude Opus 4.8 on behalf of atiaomar1978-hub_

davsclaus (approved):
Clean follow-up to CAMEL-23937 that fills the explicitly-documented gap for session-enabled consumers.

The `SessionLockRenewer` correctly follows the patterns established by the `LockRenewer` in CAMEL-23937:
- Same `PeriodTaskScheduler` integration and `maxRenewDuration` honoring
- Same `SynchronizationAdapter` for exchange completion cleanup
- Cleanup in both `doStop()` and `doShutdown()` matches the existing defensive pattern

1. **`acceptSession(sessionId).block()` in `acquireSessionReceiver`** — synchronous blocking call inside the `ServiceBusProcessorClient` callback. The Azure SDK dispatches this callback on its own thread pool (not reactor threads), so `.block()` is safe. The first message for each unique session will block briefly on the `acceptSession` handshake, serialized through `sessionReceiverCreationLock`; subsequent messages for the same session hit the fast path (cached receiver). Under extreme load with many distinct sessions arriving concurrently this could become a bottleneck, but unlikely in practice.

2. **Reference-counting is correctly synchronized** — `acquireSessionReceiver` / `releaseSessionReceiver` are both guarded by the same lock, preventing the race where an arriving message increments a holder that a concurrently completing exchange is about to close.

3. **Renewal failure handling is correct** — when `renewSessionLock()` fails, only the affected exchange's entry is removed and the ref count decremented. Other exchanges on the same session keep their entries and the session receiver stays open.

4. **Test coverage is thorough** — 8 new tests covering creation conditions, exchange tracking/cleanup, renewal invocation, failure handling, session receiver reuse, reference counting, and stop/start lifecycle.

5. **Upgrade guide** — correctly updates the existing CAMEL-23937 entry rather than creating a separate section.

gnodet (changes requested):
## Fundamental design issue with session lock contention

The `SessionLockRenewer` architecture has a critical flaw: **`acceptSession(sessionId).block()` will always fail** when called from inside the `processMessage` callback because the `ServiceBusProcessorClient` already holds an exclusive session lock on that session via its internal session receiver.

Azure Service Bus enforces exclusive session locks — only one receiver can hold a lock on a session at a time ([Microsoft docs](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sessions)). A second `acceptSession()` call for the same session will receive a `SessionCannotBeLockedException` after exhausting retry attempts (~4 minutes with default `AmqpRetryOptions`). The catch block silently swallows this exception and returns `null`, causing `add()` to skip lock renewal registration — **the feature silently does nothing in production**.

The unit tests mask this because they mock `acceptSession()` to return immediately: `when(sessionRenewalClient.acceptSession(any())).thenReturn(Mono.just(sessionBoundReceiver))`.

### Additional confirmed findings

1. **The feature may be unnecessary**: The Azure SDK documentation for `ServiceBusSessionProcessorClientBuilder.maxAutoLockRenewDuration` states that session lock renewal runs as a background task independent of callback duration — unlike per-message locks. The existing `createServiceBusSessionProcessorClient()` already configures `maxAutoLockRenewDuration` on the session processor builder.

2. **Redundant API calls** (medium): `lockedUntil` is tracked per-exchange rather than per-session. After a successful `renewSessionLock()`, only the triggering exchange's entry is updated. Other entries for the same session retain the stale value, producing N redundant API calls per renewal cycle.

3. **Missing timeout on `block()`** (medium): `acceptSession(sessionId).block()` has no explicit timeout. Under contention, this blocks the `processMessage` callback thread for ~4 minutes before failing.

4. **`cancel()` doesn't clear `entries` map** (low): Minor cleanup — `cancel()` clears `sessionReceivers` but not `entries`. Low impact since the renewer is set to null immediately after.

### Recommendation

This needs to be verified against a real Azure Service Bus session-enabled queue before merging. Alternative approaches to consider:
- Rely on the SDK's built-in session lock renewal via `maxAutoLockRenewDuration` (may already be sufficient)
- Switch from `ServiceBusProcessorClient` to a lower-level `ServiceBusSessionReceiverAsyncClient` for session consumers so Camel controls the session lifecycle directly

The architecture (reference counting, synchronized acquire/release, cleanup in doStop/doShutdown) is well-structured and the tests are thorough — the issue is with the fundamental approach, not the implementation quality.

atiaomar1978-hub:
## Converting to draft — this PR is likely unnecessary, and the current approach cannot work as designed

Thank you @gnodet and @davsclaus for the careful reviews. After digging into the Azure Service Bus SDK (`azure-messaging-servicebus` 7.17.16), I've confirmed @gnodet's `CHANGES_REQUESTED` finding is correct on both counts, so I'm converting this to draft rather than pushing a patch that wouldn't address the fundamental issue. Full explanation below.

### 1. The `acceptSession()` approach cannot work while the session processor holds the session

Azure Service Bus enforces **exclusive** session locks — only one receiver may hold a session's lock at a time ([Microsoft docs](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sessions)). In this PR, the `ServiceBusProcessorClient` (built via `sessionProcessor()`) already holds the lock for each active session. When `SessionLockRenewer.acquireSessionReceiver()` then calls `sessionRenewalClient.acceptSession(sessionId).block()` for that **same** session, the broker rejects it with `SessionCannotBeLockedException` after exhausting the retry policy (~4 minutes with the default `AmqpRetryOptions`).

The `catch` block swallows that exception and returns `null`, so `add()` skips renewal registration entirely — meaning **the feature would silently do nothing in production**. The unit tests pass only because they mock `acceptSession()` to return immediately:

```java
when(sessionRenewalClient.acceptSession(any())).thenReturn(Mono.just(sessionBoundReceiver));
```

So the tests validate the bookkeeping, not the one behaviour that actually determines whether the feature works.

### 2. The feature is very likely unnecessary — the SDK already renews session locks in the background

This is the key difference from CAMEL-23937. For **per-message** locks, the SDK's auto-renewal is driven by the message-processing pipeline, so async Camel routes (which return from `processMessage` immediately) lose renewal mid-processing — that was the real bug CAMEL-23937 fixed.

**Session** locks behave differently. The SDK begins session-lock renewal at **session-acquisition time**, not per message:

- `SessionsMessagePump` creates a `ServiceBusSessionReactorReceiver` for each rolling session (passing `maxSessionLockRenew`).
- That receiver's constructor calls `session.beginLockRenew(tracer, maxSessionLockRenew)` — a **background** renewal task tied to the session's lifetime, independent of any `processMessage` callback duration.
- `maxSessionLockRenew` is fed from `ServiceBusSessionProcessorClientBuilder.maxAutoLockRenewDuration(...)`, which Camel **already** sets in `ServiceBusClientFactory.createServiceBusSessionProcessorClient()`.

Microsoft's docs for the session processor builder confirm this: *"`maxAutoLockRenewDuration` controls how long the background renewal task runs ... the client will have to continuously renew the session lock before its expiration."* It is a background task, not one gated on callback return.

In other words, for session-enabled consumers the session lock is already kept alive for `maxAutoLockRenewDuration` regardless of how quickly `processMessage` returns — so the gap CAMEL-23937 closed for message locks does not appear to exist for session locks.

### Why I'm not just patching the minor items

The other review points (per-session vs per-exchange `lockedUntil` tracking, missing `block()` timeout, `cancel()` not clearing the `entries` map) are all trivially fixable — but none of them change conclusions (1) or (2). There's no small change that makes a second `acceptSession()` succeed against a session the processor already owns.

### Proposed path forward

I believe **CAMEL-23938 should be closed as "Not a Problem"**: the SDK's session processor already renews session locks in the background via `maxAutoLockRenewDuration`, which Camel configures. If a real-world case shows session locks still expiring for async session consumers, the correct fix would be a redesign that switches session consumers from `ServiceBusProcessorClient` to a lower-level `ServiceBusSessionReceiverAsyncClient` so Camel owns the session lifecycle directly — a much larger change that would need validation against a live session-enabled queue.

@davsclaus @gnodet — could you confirm this reasoning before I close the JIRA? I've moved the PR to draft in the meantime. Happy to close it entirely once you agree.

---
*Claude Opus 4.8 on behalf of atiaomar1978-hub. This analysis was produced by an AI agent and may contain inaccuracies; please verify before acting.*

davsclaus:
okay lets close this and resolve that jira as not a problem

PR #25004, opened 2026-07-22
CAMEL-23876: camel-a2a - pin push notification webhooks to the validated address

oscerd:
### Motivation

[CAMEL-23876](https://issues.apache.org/jira/browse/CAMEL-23876). `WebhookUrlValidator` resolves the webhook host and applies the SSRF checks (rejecting loopback/wildcard/link-local/site-local, requiring HTTPS for non-local hosts). `PushNotificationDispatcher` then built a `java.net.http.HttpRequest` from the original URL string and let `HttpClient` perform its **own, independent** DNS resolution at connection time.

The address that was validated and the address actually connected to are resolved separately, so they can differ for the same hostname — classic DNS rebinding. Nothing pinned the connection to the address the validator approved.

### Why HttpClient 5

A pin was not achievable with `java.net.http`:

- `HttpClient.Builder` exposes **no** DNS/resolver hook (only `localAddress`, which is the local bind address). The only override is the JVM-global `InetAddressResolverProvider` SPI (Java 18+, and far too heavy-handed for a component).
- Replacing the URI host with the validated IP literal would break TLS hostname verification — which matters here because the validator **requires HTTPS for every non-local webhook**, so that is the normal case.

Apache HttpClient 5's `HttpHost` carries **both** an explicit `InetAddress` and the hostname, which is exactly what is needed: connect to the validated address, while the `Host` header, TLS SNI and certificate hostname verification still use the hostname.

### Changes

- `WebhookUrlValidator`: new `validateAndResolve(url, allowLocal)` returns the address it validated. The existing `void validate(...)` delegates to it, so the other three call sites (consumer + both task stores) and the 17 existing validator tests are unchanged.
- `PushNotificationDispatcher`: delivers via `CloseableHttpAsyncClient`, targeting a `HttpHost` pinned to the validated address. HttpClient 5's `FutureCallback` is bridged to the existing `CompletableFuture` retry/backoff/work-tracking machinery, so that logic is unchanged. Response handling simplifies slightly — `SimpleHttpResponse` buffers the body, so the manual body-close is gone.
- **Validation and resolution now run on every attempt** rather than once per dispatch. Previously a webhook was validated once and could then be retried up to 30 times over as much as an hour with no re-check; that window is now closed too.
- `A2AEndpoint`: creates, starts and closes a dedicated webhook client with redirect handling disabled — a redirect would be resolved by the client and escape the validation applied to the registered URL. This matches the previous behaviour (the dispatcher already used `Redirect.NEVER`).
- New dependency on `org.apache.httpcomponents.client5:httpclient5`; upgrade-guide entry covering the change and the `PushNotificationDispatcher` constructor signature.

Only the push-webhook path moves to HttpClient 5. `AgentCardLoader` and `A2AProducer` talk to the configured agent URL (not caller-registered), and stay on `java.net.http`.

### Testing

- `mvn test` in `components/camel-ai/camel-a2a`: **506 tests pass**, including the 9 existing dispatcher tests now exercising the pinned path.
- 3 new tests asserting the target carries the validated address **and** the original hostname, with correct default/explicit port handling.

Main-only (4.22.0): hardening plus a constructor signature change, so not proposed for backport.

### DIFF
--- a/components/camel-ai/camel-a2a/pom.xml
+++ b/components/camel-ai/camel-a2a/pom.xml
@@ -64,6 +64,12 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
+        <!-- used to dispatch push notifications to a connection pinned to the validated webhook address -->
+        <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
+            <version>${httpclient-version}</version>
+        </dependency>
 
         <!-- Test dependencies -->
         <dependency>
--- a/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
+++ b/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
@@ -53,6 +53,10 @@ import org.apache.camel.spi.UriParam;
 import org.apache.camel.spi.UriPath;
 import org.apache.camel.support.DefaultEndpoint;
 import org.apache.camel.support.service.ServiceHelper;
+import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
+import org.apache.hc.core5.util.Timeout;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -86,6 +90,7 @@ public class A2AEndpoint extends DefaultEndpoint {
     private ScheduledExecutorService pushDispatcherExecutor;
     private ExecutorService httpClientExecutor;
     private HttpClient httpClient;
+    private CloseableHttpAsyncClient pushWebhookHttpClient;
     private Map<String, A2AExtensionHandler> extensionHandlers = Map.of();
     private boolean producerCreated;
     private boolean consumerCreated;
@@ -180,8 +185,10 @@ public class A2AEndpoint extends DefaultEndpoint {
             if (initializeConsumerServices) {
                 pushDispatcherExecutor = getCamelContext().getExecutorServiceManager()
                         .newScheduledThreadPool(this, "A2APushDispatcher", 4);
+                pushWebhookHttpClient = createPushWebhookHttpClient();
+                pushWebhookHttpClient.start();
                 pushDispatcher = new PushNotificationDispatcher(
-                        createHttpClient(HttpClient.Redirect.NEVER), taskStore,
+                        pushWebhookHttpClient, taskStore,
                         configuration.getPushRetryAttempts(),
                         configuration.getPushRetryBackoffMs(),
                         pushDispatcherExecutor,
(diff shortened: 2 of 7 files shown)


gnodet (approved):

## ✅ LGTM — Excellent security hardening

Well-analyzed DNS rebinding fix. The approach is solid:

### What's good

- **Root cause nailed**: validating and connecting through independent DNS lookups is the textbook TOCTOU pattern for DNS rebinding, and `HttpHost(scheme, validatedAddress, hostname, port)` is exactly the right fix — pins the connection while preserving TLS hostname verification.
- **Re-validation on every retry attempt**: the previous code validated once then could retry for up to an hour. The new code closes that window.
- **`pinnedTarget()` as a testable static method** — clean separation of the address-pinning logic, with good test coverage (validated address + hostname preserved, default ports per scheme, explicit port honoured).
- **Redirect handling**: correctly disabled (already was), with clear documentation of why.
- **Minimal blast radius**: only the push-webhook path moves to HttpClient 5. `AgentCardLoader` and `A2AProducer` (which talk to the configured agent URL, not caller-registered webhooks) stay on `java.net.http` — correct, since those aren't SSRF surfaces.
- **Upgrade guide entry**: thorough, covers the dependency change, constructor signature change, and the security rationale.
- **`httpclient-version` already managed** in parent POM (`5.6.2`), so no new version management needed.

### One note

The PR has a **merge conflict** (likely in the upgrade guide file from concurrent PRs). Will need a rebase before merge.

### Code observations (non-blocking)

The `null, null` arguments in the `httpClient.execute(target, producer, consumer, null, null, callback)` call could benefit from a brief inline comment explaining what those two parameters are (push producer / HTTP context), just for readability — but that's pure style.

Everything else — client lifecycle management in `A2AEndpoint`, test cleanup with `httpClients` list, `SimpleRequestBuilder` usage, auth header migration — all clean.

davsclaus:
merge conflict

davsclaus (approved):
Well-structured security hardening — the DNS rebinding fix is sound and the approach is clean.

**What this fixes:** `WebhookUrlValidator` resolved and validated the webhook host, then `java.net.http.HttpClient` performed a second, independent DNS resolution at connection time — the two could differ (DNS rebinding). The fix migrates the push-webhook path to Apache HttpClient 5, which supports `HttpHost(scheme, InetAddress, hostname, port)`, pinning the connection to the validated address while preserving the hostname for TLS SNI / Host header / certificate verification.

**Key observations:**

- `validateAndResolve` is backward-compatible — `validate()` delegates to it, so the other 3 call sites and 17 existing validator tests are unchanged.
- Re-validation on every retry attempt is a good improvement — previously retries (up to 30 times over ~1 hour) reused the original validation, leaving a rebinding window.
- Lifecycle ordering is correct — `pushDispatcher.shutdown()` → `pushWebhookHttpClient.close()` → executor shutdown.
- `pinnedTarget` package-private visibility is appropriate; tested from a dedicated test class in the same package.
- Dependency declaration (`httpclient5` with `${httpclient-version}`) is consistent with how `camel-http` declares it.
- Upgrade guide entry is thorough.

**Minor thought:** `SimpleResponseConsumer.create()` buffers the full response body in memory. For push notification responses this should be fine (small payloads), but worth confirming there's no scenario where a malicious webhook server could return an unbounded response body. HC5's default entity size limits likely handle this.

PR #25015, opened 2026-07-22
CAMEL-24238: camel-aws2-ses - send RawMessage body content instead of the SdkBytes descriptor

oscerd:
## Problem

`Ses2Producer.createMessage()` builds the outgoing `Content` from
`exchange.getIn().getBody(String.class)`. When the body is an SES
`RawMessage`, there is no type converter for it, so the conversion falls back
to `Object.toString()` on the SDK type.

`SdkBytes.toString()` renders a *debug descriptor*, not the payload:

```
SdkBytes(bytes=0x46726f6d3a2066726f6d406578616d706c652e636f6d0d0a...)
```

That descriptor is what ends up in the e-mail body — the actual MIME content
is never sent.

## Fix

When the body is a `RawMessage`, read the payload from
`RawMessage.data().asUtf8String()`:

```java
if (exchange.getIn().getBody() instanceof RawMessage raw) {
    // SdkBytes.toString() is a debug descriptor (SdkBytes(bytes=0x...)), not the payload
    content = raw.data().asUtf8String();
} else {
    content = exchange.getIn().getBody(String.class);
}
```

Note this path is distinct from the `jakarta.mail.Message` body, which already
goes through `sendRawEmail` and is unaffected.

## Test

`SesComponentTest.rawMessageBodyIsSentAsItsContentNotTheSdkBytesDescriptor()`
sends a `RawMessage` and asserts the text captured by the mock SES client
equals the original MIME string. The test fails on `main` with the
`SdkBytes(bytes=0x...)` descriptor and passes with the fix.

`assertj-core` is added test-scoped so the new assertion follows the project's
AssertJ preference; the pre-existing JUnit assertions in the file are left
untouched.

## Backport

The same code is present on `camel-4.18.x` and `camel-4.14.x`, so this is
targeted at 4.22.0, 4.18.4 and 4.14.9.

---

### DIFF
--- a/components/camel-aws/camel-aws2-ses/pom.xml
+++ b/components/camel-aws/camel-aws2-ses/pom.xml
@@ -83,5 +83,10 @@
             <artifactId>camel-test-spring-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/components/camel-aws/camel-aws2-ses/src/main/java/org/apache/camel/component/aws2/ses/Ses2Producer.java
+++ b/components/camel-aws/camel-aws2-ses/src/main/java/org/apache/camel/component/aws2/ses/Ses2Producer.java
@@ -101,7 +101,8 @@ public class Ses2Producer extends DefaultProducer {
                 = software.amazon.awssdk.services.ses.model.Message.builder();
         String content;
         if (exchange.getIn().getBody() instanceof RawMessage raw) {
-            content = raw.data().toString();
+            // SdkBytes.toString() is a debug descriptor (SdkBytes(bytes=0x...)), not the payload
+            content = raw.data().asUtf8String();
         } else {
             content = exchange.getIn().getBody(String.class);
         }
--- a/components/camel-aws/camel-aws2-ses/src/test/java/org/apache/camel/component/aws2/ses/SesComponentTest.java
+++ b/components/camel-aws/camel-aws2-ses/src/test/java/org/apache/camel/component/aws2/ses/SesComponentTest.java
@@ -26,15 +26,29 @@ import org.apache.camel.Processor;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.test.junit6.CamelTestSupport;
 import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.ses.model.MessageTag;
+import software.amazon.awssdk.services.ses.model.RawMessage;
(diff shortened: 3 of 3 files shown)


gnodet (approved):
Clean, well-targeted bug fix. `SdkBytes.toString()` returns a debug descriptor (e.g., `SdkBytes(bytes=0x46726F6D...)`) rather than the actual content — so when a `RawMessage` body was used, the email text was garbled with the descriptor string instead of the MIME content.

The fix correctly uses `raw.data().asUtf8String()` which decodes the actual byte content as UTF-8 — appropriate for MIME email content.

**Test added:** `rawMessageBodyIsSentAsItsContentNotTheSdkBytesDescriptor()` constructs a `RawMessage` with MIME content, sends it through the route, and verifies the actual content is preserved. The test would have caught the original bug (the assertion would fail comparing `"SdkBytes(bytes=0x...)"` against the expected MIME string).

**Checklist:**
- [x] Bug fix — correct API usage (`asUtf8String()` vs `toString()`)
- [x] Test — new test covers the exact bug scenario
- [x] AssertJ dependency added (project standard for assertions)
- [x] Commit convention — `CAMEL-24238:` prefix
- [x] No public API changes
- [ ] CI — pending

davsclaus (approved):
Clean, focused bug fix. `SdkBytes.toString()` returns a debug descriptor (`SdkBytes(bytes=0x...)`), not the actual payload — `asUtf8String()` is the correct API.

**Verified:**
- The bug was introduced in `fe3d038cd6a9` (CAMEL-21593, Jan 2025) when `RawMessage` support was added — the `toString()` vs `asUtf8String()` mistake has been present since the feature was first added
- Routing logic in `process()` dispatches `jakarta.mail.Message` bodies to `sendRawEmail()` and everything else (including `RawMessage`) to `sendEmail()` via `createMessage()` — the fix targets the correct code path
- Test `rawMessageBodyIsSentAsItsContentNotTheSdkBytesDescriptor()` directly tests the regression: builds a `RawMessage` from a MIME string, sends it, and asserts the captured body equals the original content
- New test uses AssertJ and drops `public` per project conventions; existing JUnit assertions left untouched
- Comment explains WHY `asUtf8String()` is used — a genuine SDK gotcha worth documenting
- No scope drift, commit convention correct

_Claude Code on behalf of davsclaus. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

gnodet (approved):
Correct one-line fix for a genuine bug introduced in CAMEL-21593 where `SdkBytes.toString()` yields a debug descriptor instead of the payload. The fix to use `asUtf8String()` is consistent with how other AWS components in the codebase extract text from `SdkBytes`. Test is appropriate and validates the fix.

- The `assertj-core` dependency is added solely for a single `assertThat` call. Using `assertEquals(mime, sent)` would be consistent with the rest of the test file and avoid the extra dependency.
- New test method uses package-private access while existing methods use `public void` — minor style inconsistency.

PR #25026, opened 2026-07-22
CAMEL-24220: Add camel_security_scan tool for route security analysis

oscerd:
## Summary

Adds a new MCP tool `camel_security_scan` for static security analysis of Camel route definitions.

Distinct from `camel_route_harden_context` (general security context + CVE advisories), this tool performs **line-by-line static analysis** to detect specific anti-patterns:

- **Exposed secrets** in URIs (passwords, tokens, API keys) — property placeholder-aware, correctly flags RAW()-wrapped values
- **Insecure configuration options** from `SecurityUtils.getSecurityOptions()` (trustAllCertificates, allowJavaSerializedObject, etc.) — skips options with empty insecureValue
- **Missing Camel\* header filters** on HTTP/REST consumers — per-consumer scoping with nearby filter detection
- **Unencrypted protocols** (HTTP, FTP, LDAP, SMTP vs secure alternatives)
- **Command injection risk** via exec component
- **SQL injection risk** — parameterized query check with composite-scheme-safe matching
- **Path traversal risk** in file component with dynamic expressions

Returns findings with severity, line number, category, and remediation guidance.

## Test plan

- [x] All 324 MCP server tests pass (no regressions)
- [x] Review feedback addressed: RAW() detection, empty insecureValue, composite scheme matching, per-consumer header filter scoping

### DIFF
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
@@ -1,0 +1,323 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.dsl.jbang.core.commands.mcp;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import io.quarkiverse.mcp.server.Tool;
+import io.quarkiverse.mcp.server.ToolArg;
+import io.quarkiverse.mcp.server.ToolCallException;
+import org.apache.camel.util.SecurityUtils;
+
+/**
+ * MCP Tool for static security analysis of Camel route definitions.
+ * <p>
+ * Scans route content (YAML, XML, or Java DSL) for security anti-patterns defined in the Camel security model
+ * ({@code design/security.adoc}). Produces actionable findings with severity, line location, and remediation guidance.
+ * <p>
+ * Distinct from {@link HardenTools} which provides general security context and CVE advisories — this tool performs
+ * line-by-line static analysis to find specific violations.
+ */
+@ApplicationScoped
+public class SecurityScanTools {
+
+    private static final Pattern SECRET_IN_URI = Pattern.compile(
+            "(?i)(password|passwd|pwd|secret|token|apikey|api[_-]?key|access[_-]?key)\\s*[=:]\\s*(?!\\{\\{)(?!\\$\\{)(?!RAW\\()([^\\s,;}'\"\\]&]+)");
+
+    private static final Pattern CONNECTION_STRING_CREDS = Pattern.compile(
+            "(?i)(mongodb(\\+srv)?://|amqp://|redis://|jdbc:)\\S+:\\S+@");
+
+    private static final String[] CONSUMER_COMPONENTS = {
+            "platform-http", "netty-http", "jetty", "undertow", "servlet",
+            "cxf", "cxfrs", "rest", "coap", "grpc"
+    };
+
+    @Tool(annotations = @Tool.Annotations(readOnlyHint = true, destructiveHint = false, openWorldHint = false),
+          description = "Scan a Camel route for security anti-patterns. "
+                        + "Performs static analysis to detect: exposed secrets in URIs, "
+                        + "insecure configuration options (trustAllCertificates, allowJavaSerializedObject, etc.), "
+                        + "missing Camel* header filters on consumers, unencrypted protocols, "
+                        + "and other violations from the Camel security model. "
+                        + "Returns findings with severity, line number, and remediation guidance. "
+                        + "Distinct from camel_route_harden_context which provides general security context "
+                        + "and CVE advisories.")
+    public SecurityScanResult camel_security_scan(
+            @ToolArg(description = "The Camel route content (YAML, XML, or Java DSL)") String route,
+            @ToolArg(description = "Route format: yaml, xml, or java (default: yaml)") String format) {
+
+        if (route == null || route.isBlank()) {
+            throw new ToolCallException("Route content is required", null);
+        }
+
+        try {
+            String resolvedFormat = format != null && !format.isBlank() ? format.toLowerCase(Locale.ROOT) : "yaml";
+            String[] lines = route.split("\n", -1);
+
+            List<Finding> findings = new ArrayList<>();
+
+            scanInsecureOptions(lines, findings);
+            scanSecretsInUris(lines, findings);
+            scanConnectionStringCredentials(lines, findings);
+            scanUnencryptedProtocols(lines, findings);
+            scanMissingHeaderFilters(route, lines, findings);
+            scanExecComponent(lines, findings);
+            scanSqlInjection(lines, findings);
+            scanFilePathTraversal(lines, findings);
+
+            findings.sort((a, b) -> severityRank(a.severity()) - severityRank(b.severity()));
+
+            int critical = (int) findings.stream().filter(f -> "critical".equals(f.severity())).count();
+            int high = (int) findings.stream().filter(f -> "high".equals(f.severity())).count();
+            int medium = (int) findings.stream().filter(f -> "medium".equals(f.severity())).count();
+            int low = (int) findings.stream().filter(f -> "low".equals(f.severity())).count();
+
+            return new SecurityScanResult(
+                    resolvedFormat, findings, findings.size(),
+                    new SeverityCounts(critical, high, medium, low));
+        } catch (ToolCallException e) {
+            throw e;
+        } catch (Throwable e) {
+            throw new ToolCallException(
+                    "Failed to scan route (" + e.getClass().getName() + "): " + e.getMessage(), null);
+        }
+    }
+
+    private void scanInsecureOptions(String[] lines, List<Finding> findings) {
+        Map<String, SecurityUtils.SecurityOption> securityOptions = SecurityUtils.getSecurityOptions();
+
+        for (int i = 0; i < lines.length; i++) {
+            String lower = lines[i].toLowerCase(Locale.ROOT).replaceAll("[\\s-]", "");
+            for (Map.Entry<String, SecurityUtils.SecurityOption> entry : securityOptions.entrySet()) {
+                String optionKey = entry.getKey();
+                String insecureValue = entry.getValue().insecureValue();
+                String category = entry.getValue().category();
+
+                if (lower.contains(optionKey + "=" + insecureValue)
+                        || lower.contains(optionKey + ":" + insecureValue)
+                        || lower.contains(optionKey + "\":" + insecureValue)
+                        || lower.contains("\"" + optionKey + "\":" + insecureValue)) {
+                    findings.add(new Finding(
+                            severityForCategory(category),
+                            category,
+                            "Insecure option: " + optionKey + "=" + insecureValue,
+                            i + 1,
+                            remediationForCategory(category, optionKey)));
+                }
+            }
+        }
+    }
+
+    private void scanSecretsInUris(String[] lines, List<Finding> findings) {
+        for (int i = 0; i < lines.length; i++) {
+            Matcher matcher = SECRET_IN_URI.matcher(lines[i]);
+            while (matcher.find()) {
+                String key = matcher.group(1);
+                findings.add(new Finding(
+                        "critical",
+                        "secret",
+                        "Potential plain-text secret in URI: " + key,
+                        i + 1,
+                        "Use property placeholders {{" + key + "}} or a secrets management vault "
+                               + "(HashiCorp Vault, AWS Secrets Manager, Azure Key Vault)"));
+            }
+        }
+    }
+
+    private void scanConnectionStringCredentials(String[] lines, List<Finding> findings) {
+        for (int i = 0; i < lines.length; i++) {
+            if (CONNECTION_STRING_CREDS.matcher(lines[i]).find()) {
+                findings.add(new Finding(
+                        "critical",
+                        "secret",
+                        "Connection string with embedded credentials",
+                        i + 1,
+                        "Extract credentials from the connection string and use property placeholders or vault services"));
+            }
+        }
+    }
+
+    private void scanUnencryptedProtocols(String[] lines, List<Finding> findings) {
+        for (int i = 0; i < lines.length; i++) {
+            String lower = lines[i].toLowerCase(Locale.ROOT);
+
+            if (containsScheme(lower, "http") && !containsScheme(lower, "https")
+                    && !lower.contains("platform-http")) {
+                findings.add(new Finding(
+                        "high",
+                        "insecure:ssl",
+                        "Using HTTP instead of HTTPS",
+                        i + 1,
+                        "Use HTTPS with TLS 1.2+ for secure communication"));
+            }
+            if (containsScheme(lower, "ftp") && !containsScheme(lower, "sftp") && !containsScheme(lower, "ftps")) {
+                findings.add(new Finding(
+                        "high",
+                        "insecure:ssl",
+                        "Using plain FTP instead of SFTP/FTPS",
+                        i + 1,
+                        "Use SFTP or FTPS for encrypted file transfers"));
+            }
+            if (containsScheme(lower, "ldap") && !containsScheme(lower, "ldaps")) {
+                findings.add(new Finding(
+                        "medium",
+                        "insecure:ssl",
+                        "Using LDAP instead of LDAPS",
+                        i + 1,
+                        "Use LDAPS for encrypted LDAP communication"));
+            }
+            if (containsScheme(lower, "smtp") && !containsScheme(lower, "smtps")) {
+                findings.add(new Finding(
+                        "medium",
+                        "insecure:ssl",
+                        "Using SMTP instead of SMTPS",
+                        i + 1,
+                        "Use SMTPS for encrypted email communication"));
+            }
+        }
+    }
+
+    private void scanMissingHeaderFilters(String route, String[] lines, List<Finding> findings) {
+        String lower = route.toLowerCase(Locale.ROOT);
+        for (String consumer : CONSUMER_COMPONENTS) {
+            if (containsScheme(lower, consumer)) {
+                if (!lower.contains("removeheaders") && !lower.contains("headerfilter")
+                        && !lower.contains("camel*")) {
+                    int lineNum = findLineContaining(lines, consumer + ":");
+                    findings.add(new Finding(
+                            "high",
+                            "header-injection",
+                            "Consumer '" + consumer + "' without Camel* header filter",
+                            lineNum,
+                            "Add removeHeaders(\"Camel*\") or a HeaderFilterStrategy to prevent "
+                                     + "external clients from injecting Camel-internal headers "
+                                     + "(CVE-2025-27636 and related)"));
+                }
+            }
+        }
+    }
+
+    private void scanExecComponent(String[] lines, List<Finding> findings) {
+        for (int i = 0; i < lines.length; i++) {
+            if (containsScheme(lines[i].toLowerCase(Locale.ROOT), "exec")) {
+                findings.add(new Finding(
+                        "critical",
+                        "command-injection",
+                        "Using exec component — high risk for command injection",
+                        i + 1,
+                        "Validate all inputs strictly. Consider safer alternatives. "
+                               + "Never pass untrusted input directly to exec"));
+            }
+        }
+    }
+
+    private void scanSqlInjection(String[] lines, List<Finding> findings) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Well-designed static security scanner — good complement to `HardenTools` (general context/CVEs) by providing line-by-line analysis with actionable findings.

**What's good:**
- Reuses `SecurityUtils.getSecurityOptions()` from camel-util for the canonical 24 insecure options — no duplicate maintenance
- Property placeholder awareness: `SECRET_IN_URI` regex correctly excludes `{{…}}`, `${…}`, and `RAW(…)` prefixed values
- Scheme detection via `containsScheme(text, scheme + ":")` is precise — `"https:"` does NOT match `"http:"` check
- `platform-http` correctly excluded from the unencrypted-HTTP check
- Good false-positive tests (placeholder passwords, HTTPS, SFTP all correctly suppressed)
- 24 test cases covering detection, suppression, line numbers, severity ordering, and counts
- Clean result model with records, severity sorting (critical first)
- Proper MCP annotations (`readOnlyHint=true, destructiveHint=false`)

**Minor observations** (not blocking — inherent to text-based static analysis):
- `scanMissingHeaderFilters` checks the entire route text for `removeHeaders`/`headerfilter` presence, so a filter in one route suppresses the finding for all consumers in the same input. Acceptable for a single-route scanner, but worth a Javadoc note.
- `scanExecComponent` flags any line containing `exec:`, which could match YAML comments or description fields. Fine as an advisory finding.
- No test for LDAP/LDAPS detection (the scanner checks for it but there's no corresponding test case)

**Checklist:**
- [x] Code quality — clean, well-structured, proper use of camel-util APIs
- [x] Security model — aligns with `design/security.adoc` threat model categories
- [x] Tests — 24 cases with both positive detection and false-positive suppression
- [x] Backward compatible — new tool, no existing changes
- [x] Commit convention — `CAMEL-24220:` prefix
- [ ] CI — pending

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,323 @@
...
+ * <p>
+ * Distinct from {@link HardenTools} which provides general security context and CVE advisories — this tool performs
+ * line-by-line static analysis to find specific violations.
+ */
+@ApplicationScoped
+public class SecurityScanTools {
+
+    private static final Pattern SECRET_IN_URI = Pattern.compile(
+            "(?i)(password|passwd|pwd|secret|token|apikey|api[_-]?key|access[_-]?key)\\s*[=:]\\s*(?!\\{\\{)(?!\\$\\{)(?!RAW\\()([^\\s,;}'\"\\]&]+)");
+
```
gnodet, line 47: **Correctness bug: RAW() negative lookahead creates false negatives** — The `(?!RAW\\()` in `SECRET_IN_URI` prevents matching `password=RAW(mysecret123)`. However, `SecurityUtils.isPlainTextSecret` (line 146-147) explicitly documents:

> _"RAW() is a URI encoding wrapper, not a security mechanism -- RAW(password) is still plain text."_

This means `password=RAW(mysecret123)` is silently skipped by the scanner, creating a false negative for actual plain-text secrets. Remove the `(?!RAW\\()` negative lookahead.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,323 @@
...
+        } catch (ToolCallException e) {
+            throw e;
+        } catch (Throwable e) {
+            throw new ToolCallException(
+                    "Failed to scan route (" + e.getClass().getName() + "): " + e.getMessage(), null);
+        }
+    }
+
+    private void scanInsecureOptions(String[] lines, List<Finding> findings) {
+        Map<String, SecurityUtils.SecurityOption> securityOptions = SecurityUtils.getSecurityOptions();
```
gnodet, line 107: **Correctness bug: empty insecureValue causes false positives** — Three options (`sslendpointalgorithm`, `stricthostkeychecking`, `x509hostnameverifier`) have `insecureValue=""`. The check `lower.contains(optionKey + "=" + insecureValue)` becomes `lower.contains("sslendpointalgorithm=")` which matches ANY assignment — including secure ones like `sslEndpointAlgorithm=HTTPS`.

Note that `SecurityUtils.isInsecureValue` correctly uses `equalsIgnoreCase` which only matches when the value is actually empty. Add a guard:

```java
if (insecureValue.isEmpty()) {
    continue;
}
```

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,323 @@
...
+        for (String consumer : CONSUMER_COMPONENTS) {
+            if (containsScheme(lower, consumer)) {
+                if (!lower.contains("removeheaders") && !lower.contains("headerfilter")
+                        && !lower.contains("camel*")) {
+                    int lineNum = findLineContaining(lines, consumer + ":");
+                    findings.add(new Finding(
+                            "high",
+                            "header-injection",
+                            "Consumer '" + consumer + "' without Camel* header filter",
+                            lineNum,
```
gnodet: **False positive: `google-bigquery-sql:` URIs** — `containsScheme(lower, "sql")` checks for `sql:` as a substring, which matches the composite scheme `google-bigquery-sql:project:dataset.table`. BigQuery SQL uses a different query model and does not use `:#param` syntax. The same issue applies to other composite schemes ending with a target scheme name.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,323 @@
...
+                findings.add(new Finding(
+                        "high",
+                        "insecure:ssl",
+                        "Using plain FTP instead of SFTP/FTPS",
+                        i + 1,
+                        "Use SFTP or FTPS for encrypted file transfers"));
+            }
+            if (containsScheme(lower, "ldap") && !containsScheme(lower, "ldaps")) {
+                findings.add(new Finding(
+                        "medium",
```
gnodet, line 219: **Route-level header filter check** — `scanMissingHeaderFilters` checks the ENTIRE route content for `removeheaders`/`headerfilter`/`camel*`. This means: (a) if a route has two HTTP consumers and only one has a header filter, neither gets flagged; (b) any occurrence of `"camel*"` anywhere in the route (even in a comment or description) suppresses all header-injection warnings globally.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,323 @@
...
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import io.quarkiverse.mcp.server.Tool;
+import io.quarkiverse.mcp.server.ToolArg;
+import io.quarkiverse.mcp.server.ToolCallException;
+import org.apache.camel.util.SecurityUtils;
+
+/**
+ * MCP Tool for static security analysis of Camel route definitions.
```
gnodet, line 34: **Code duplication with `HardenTools.analyzeSecurityConcerns()`** — The HTTP-vs-HTTPS, FTP, exec, SQL injection, file path traversal, and LDAP/SMTP checks are all reimplemented here. These tools serve different purposes (catalog-aware context vs. line-level scanning) but share substantial string-matching logic that will diverge as one file is updated but not the other. Consider extracting shared detection rules into a common utility class.

gnodet (changes requested):
The static security scanner has good structure and useful output format, but contains correctness bugs that produce wrong results. All findings below were independently verified by a second agent.

**Key issues (verified):**
1. **RAW() false negative** (high) — the secret-detection regex skips `password=RAW(mysecret)`, but `SecurityUtils.isPlainTextSecret` explicitly documents that RAW() is NOT a security mechanism
2. **Empty insecureValue false positives** (high) — three security options with `insecureValue=""` cause the scanner to flag ANY value assignment (e.g., `sslEndpointAlgorithm=HTTPS` is incorrectly flagged)
3. **SQL scheme substring matching** (medium) — `google-bigquery-sql:` URIs are falsely flagged as SQL injection risks because `containsScheme(lower, "sql")` uses substring matching
4. **Code duplication with HardenTools** (medium) — HTTP/HTTPS, FTP, exec, SQL, LDAP, and file path traversal checks are reimplemented here

oscerd:
Thank you for the thorough review — all 4 blocking findings addressed:

**High Severity (fixed):**
1. ✅ **RAW() false negative** — Removed `(?!RAW\\()` negative lookahead from `SECRET_IN_URI` regex. `password=RAW(secret)` is now correctly detected as a plain-text secret, consistent with `SecurityUtils.isPlainTextSecret` documentation.

2. ✅ **Empty insecureValue false positives** — Added early `continue` when `insecureValue` is null or empty. Options like `sslEndpointAlgorithm` with empty insecureValue no longer match any assignment.

3. ✅ **SQL composite scheme false positive** — `containsScheme()` now checks that the character before the scheme is not a letter, digit, or hyphen. `google-bigquery-sql:` no longer falsely matches as `sql:`.

4. ✅ **Route-level header filter check** — Replaced route-wide text search with per-consumer `hasHeaderFilterNearby()` that searches the next 20 lines after each consumer declaration (bounded by the next consumer occurrence). A `"camel*"` in a comment no longer suppresses all warnings.

**Suggestion (acknowledged):**
5. Code duplication with `HardenTools` — Valid concern. The two tools serve different purposes (catalog-aware context vs line-level scanning) with different output formats. Extracting shared detection logic is a good follow-up but would need careful API design to serve both use cases.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,356 @@
...
+    private static boolean containsScheme(String text, String scheme) {
+        int idx = text.indexOf(scheme + ":");
+        if (idx >= 0) {
+            if (idx == 0 || !Character.isLetterOrDigit(text.charAt(idx - 1)) && text.charAt(idx - 1) != '-') {
+                return true;
+            }
+        }
+        return text.contains("\"" + scheme + "\"") || text.contains("'" + scheme + "'");
+    }
+
```
gnodet, line 331: **Low:** Operator precedence readability — the expression `idx == 0 || !Character.isLetterOrDigit(text.charAt(idx - 1)) && text.charAt(idx - 1) != '-'` is functionally correct due to Java precedence (`!` > `&&` > `||`), but adding explicit parentheses would make the intent immediately clear:

```suggestion
            if (idx == 0 || (!Character.isLetterOrDigit(text.charAt(idx - 1)) && text.charAt(idx - 1) != '-')) {
```

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
```diff
@@ -0,0 +1,356 @@
...
+    private static int findLineContaining(String[] lines, String text) {
+        String lower = text.toLowerCase(Locale.ROOT);
+        for (int i = 0; i < lines.length; i++) {
+            if (lines[i].toLowerCase(Locale.ROOT).contains(lower)) {
+                return i + 1;
+            }
+        }
```
gnodet: **Low:** Dead code — `findLineContaining()` is defined but never called. The refactoring to per-consumer `hasHeaderFilterNearby()` eliminated its only call site. Can be removed.

```suggestion
```

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanToolsTest.java
```diff
@@ -0,0 +1,363 @@
...
+    private static int severityRank(String severity) {
+        return switch (severity) {
+            case "critical" -> 0;
+            case "high" -> 1;
+            case "medium" -> 2;
+            case "low" -> 3;
+            default -> 4;
+        };
+    }
+}
```
gnodet, line 424: **Medium:** The 4 bug fixes from this commit lack dedicated regression tests. Per project rules, every PR must include tests for bug fixes. Specifically:

1. Test that `password=RAW(mysecret)` IS detected as a plain-text secret (existing `detectsPlainTextPasswordInUri` uses `password=mysecret123` which doesn't exercise the RAW() path)
2. Test that `sslEndpointAlgorithm=HTTPS` is NOT flagged (empty `insecureValue` case)
3. Test that `google-bigquery-sql:project:dataset` is NOT flagged as SQL injection (composite scheme)
4. Test with two consumers where only one has a header filter, verifying the unprotected one is still flagged

Also: no test for LDAP/LDAPS detection, unlike the other protocol pairs (HTTP/FTP/SMTP).

gnodet (commented):
**Re-review** (new commits since last review)

All 4 blocking findings from the previous review are correctly fixed in the code, and CI passes. The per-consumer `hasHeaderFilterNearby()` approach is a solid improvement. However, the 4 bug-fix scenarios lack dedicated regression tests, and the refactoring left dead code.

**Verified findings (4):**

| # | Severity | Finding |
|---|----------|--------|
| 1 | medium | Missing regression tests for 4 fixed bugs (RAW() detection, empty insecureValue, composite scheme, per-consumer header filter) |
| 2 | low | Dead `findLineContaining()` method — never called after refactoring |
| 3 | low | Operator precedence in `containsScheme()` — explicit parentheses would improve clarity |
| 4 | low | No test for LDAP/LDAPS detection (unlike HTTP/FTP/SMTP protocol pairs) |

None of these are blocking — the code is correct and the primary use cases are well-tested — but addressing them would strengthen the PR.

oscerd:
All re-review findings addressed:

1. ✅ Added 4 regression tests: RAW() secret detection, empty insecureValue no false positive, composite scheme no SQL false positive, per-consumer header filter scoping
2. ✅ Added explicit parentheses in containsScheme() for readability
3. ✅ Removed dead findLineContaining() method

gnodet (approved):
Re-review: all blocking findings from the prior CHANGES_REQUESTED round and all follow-up suggestions have been properly addressed:

✅ **RAW() false negative** — removed the `(?!RAW\()` negative lookahead so `password=RAW(mysecret)` is correctly flagged. Regression test added.
✅ **Empty insecureValue false positives** — added guard to skip options with empty `insecureValue`. Regression test added.
✅ **SQL scheme substring matching** — rewrote `containsScheme()` with boundary-aware matching (rejects match preceded by letter/digit/hyphen). Regression test added.
✅ **Dead `findLineContaining()` method** — removed.
✅ **Operator precedence in `containsScheme()`** — explicit parentheses added.
✅ **Per-consumer header filter check** — `hasHeaderFilterNearby()` now correctly handles multi-consumer routes. Regression test added.
✅ **4 regression tests** added for all fixed bugs.

Two minor non-blocking observations:

1. `containsScheme()` uses `indexOf()` which returns only the first occurrence — if the same scheme appears twice on a line and the first is rejected by the boundary check, the second is never checked. Practical impact is negligible.

2. The code duplication with `HardenTools` is noted but acceptable as a design trade-off (SecurityScanTools is a pure text-based analyzer with no catalog dependency). Worth noting that HardenTools still uses naive `contains("sql:")` matching — a follow-up could extract the improved `containsScheme()` boundary logic into a shared utility.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Addressed — the current SECRET_IN_URI regex has no RAW() negative lookahead, and its value capture includes the RAW(...) wrapper, so password=RAW(mysecret123) IS flagged. Pinned by the new detectsRawWrappedPasswordAsPlainTextSecret test.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Done in e4f2712 — scanInsecureOptions now extracts the assigned value and calls SecurityUtils.isInsecureValue() instead of a string contains. That is the canonical check (case-insensitive, and the sslEndpointAlgorithm empty/none/false special case), so sslEndpointAlgorithm=HTTPS is not flagged while sslEndpointAlgorithm=none now is. Covered by doesNotFlagSecureOptionValuesAsInsecure and the new flagsSslEndpointAlgorithmDisabled.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Addressed — containsScheme() now requires the char before the scheme to be neither a letter/digit nor a dash, so google-bigquery-sql: no longer matches sql. Pinned by doesNotFlagCompositeSchemeAsSql.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Addressed — scanMissingHeaderFilters is now per-consumer: for each consumer line it calls hasHeaderFilterNearby() which searches only the window until the next consumer, so a filter on one consumer no longer suppresses another. Pinned by perConsumerHeaderFilterCheckDoesNotSuppressGlobally.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Done — the parentheses are in place: idx == 0 || (!Character.isLetterOrDigit(...) && ... != '-').\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: Done — findLineContaining() was removed; it has no remaining call site.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanToolsTest.java
oscerd: Done — added dedicated regression tests for all four: detectsRawWrappedPasswordAsPlainTextSecret, doesNotFlagSecureOptionValuesAsInsecure (+ flagsSslEndpointAlgorithmDisabled for the restored insecure case), doesNotFlagCompositeSchemeAsSql, and perConsumerHeaderFilterCheckDoesNotSuppressGlobally. 29 tests pass.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/SecurityScanTools.java
oscerd: You're right that the heuristic knowledge (http-not-https, ftp, exec, sql, path-traversal) overlaps. The mechanics differ substantially though: HardenTools.analyzeSecurityConcerns does whole-route contains() and emits catalog-correlated SecurityConcerns with no line numbers, whereas this tool is line-by-line with boundary-aware containsScheme() and emits Findings with line numbers and remediation. The genuinely shareable part is the insecure-protocol pair table. I'd prefer to extract that into a small shared rules helper as a dedicated follow-up so it can touch HardenTools with its own test pass, rather than widen this PR. Leaving this open for your call — happy to do it here if you'd rather.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

gnodet (approved):
Re-review after new commit `e4f2712`. The refactoring correctly delegates insecure-option detection to `SecurityUtils.isInsecureValue()`, which restores proper `sslEndpointAlgorithm` detection that the previous empty-`insecureValue` guard inadvertently broke. All prior review findings are addressed, 29 tests pass, CI is green.

**Changes since last approval:**
- New `extractOptionValue()` helper properly parses option values from URI-query, YAML, and JSON key forms
- Insecure-value matching delegated to canonical `SecurityUtils.isInsecureValue()`
- New test `flagsSslEndpointAlgorithmDisabled` covers the restored detection

- `extractOptionValue()` does not check for a word boundary before the option key — not a regression (previous code had same behavior) and no real-world security option key is a suffix of another.
- No test for `sslEndpointAlgorithm` with an empty/absent value (`sslEndpointAlgorithm=`), only `=none` is tested. Minor gap since the canonical utility handles it.

PR #25027, opened 2026-07-22
CAMEL-21314: camel-aws2-s3 - supply the customer key when copying SSE-C objects

oscerd:
### Motivation

[CAMEL-21314](https://issues.apache.org/jira/browse/CAMEL-21314). `S3CopyObjectCustomerKeyIT` was failing and ended up disabled as a "Broken test". It turns out the test was right and the **producer is broken**: copying an object that was stored with SSE-C does not work at all.

`copyObject` only set the *destination* SSE-C parameters (`sseCustomerKey`, `sseCustomerKeyMD5`, `sseCustomerAlgorithm`). When the *source* object is itself SSE-C encrypted, S3 also needs the copy-source customer key in order to decrypt it, and rejects the request otherwise — which is exactly the failure reported on the issue:

```
InvalidRequest: SSE-C encrypted objects require customer key headers. (Service: S3, Status Code: 400)
```

(the issue also quotes the equivalent `The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object.`)

So this is a real user-facing bug, not just a test problem: `copyObject` against any SSE-C encrypted object fails.

### Changes

- `AWS2S3Producer.copyObject` now also sets `copySourceSSECustomerKey`, `copySourceSSECustomerKeyMD5` and `copySourceSSECustomerAlgorithm`, alongside the existing destination parameters, from the same configured customer key — the source was written with that key and the destination is re-encrypted with it.
- Re-enabled `S3CopyObjectCustomerKeyIT` (dropped `@Disabled("Broken test")` and the now-unused import).

### Testing

Verified end-to-end against localstack, including a control run to confirm the fix is what makes the difference:

- **Without** the producer change (fix stashed, test enabled): `S3CopyObjectCustomerKeyIT` fails with `InvalidRequest: SSE-C encrypted objects require customer key headers ... Status Code: 400`.
- **With** the fix: `Tests run: 1, Failures: 0, Errors: 0` — the IT passes, and the module's 41 unit tests pass alongside it.

### Backport

The same destination-only handling is present on `camel-4.18.x` and `camel-4.14.x` (verified: neither branch sets any `copySourceSSECustomerKey*` parameter), so this is a backport candidate for both once merged here.

### DIFF
--- a/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Producer.java
+++ b/components/camel-aws/camel-aws2-s3/src/main/java/org/apache/camel/component/aws2/s3/AWS2S3Producer.java
@@ -518,13 +518,18 @@ public class AWS2S3Producer extends DefaultProducer {
             }
 
             if (getConfiguration().isUseCustomerKey()) {
+                // the source object was stored with SSE-C, so the same customer key must also be supplied as the
+                // copy-source key, otherwise S3 cannot decrypt the source and rejects the copy
                 if (ObjectHelper.isNotEmpty(getConfiguration().getCustomerKeyId())) {
+                    copyObjectRequest.copySourceSSECustomerKey(getConfiguration().getCustomerKeyId());
                     copyObjectRequest.sseCustomerKey(getConfiguration().getCustomerKeyId());
                 }
                 if (ObjectHelper.isNotEmpty(getConfiguration().getCustomerKeyMD5())) {
+                    copyObjectRequest.copySourceSSECustomerKeyMD5(getConfiguration().getCustomerKeyMD5());
                     copyObjectRequest.sseCustomerKeyMD5(getConfiguration().getCustomerKeyMD5());
                 }
                 if (ObjectHelper.isNotEmpty(getConfiguration().getCustomerAlgorithm())) {
+                    copyObjectRequest.copySourceSSECustomerAlgorithm(getConfiguration().getCustomerAlgorithm());
                     copyObjectRequest.sseCustomerAlgorithm(getConfiguration().getCustomerAlgorithm());
                 }
             }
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Correct fix — `copyObject` was only setting the *destination* SSE-C parameters, so S3 rejected the copy because it couldn't decrypt the source. Adding `copySourceSSECustomerKey`, `copySourceSSECustomerKeyMD5`, and `copySourceSSECustomerAlgorithm` alongside the existing destination parameters is exactly what the AWS SDK requires.

**What's good:**
- Root cause clearly identified: S3 needs both source (to decrypt) and destination (to re-encrypt) customer key parameters for SSE-C copy operations
- Minimal fix — only the 3 missing `copySource*` calls added, no unrelated changes
- Re-enables `S3CopyObjectCustomerKeyIT` that was disabled as "Broken test" — the test was actually correct, the producer was broken
- Excellent PR description: reproduces the error, explains the fix, documents testing methodology, identifies backport targets
- Comment on the "why" in the code itself

**Checklist:**
- [x] Code quality — focused, minimal fix
- [x] Correct S3 SDK usage — `copySourceSSECustomerKey*` parameters match AWS API requirements
- [x] Tests — re-enabled existing IT that validates the exact scenario
- [x] Backward compatible — fixes broken behavior
- [x] Commit convention — `CAMEL-21314:` prefix
- [ ] CI — queued

gnodet (approved):
Clean, minimal fix that correctly adds the copy-source SSE-C decryption parameters alongside the existing destination encryption parameters when copying S3 objects. The AWS SDK API confirms the method names are valid, the re-enabled integration test directly covers the fix scenario, and the change is well-scoped to the reported bug.

Git history confirms the SSE-C copy code was originally authored in 2020 (commit 7222d908e041) with only the destination parameters, and the test was disabled later (commit 2b3e5dec58e0) when the bug surfaced. This change correctly addresses the original omission.

Minor note: the consumer's `processCommit` method also calls `copyObject` (for `moveAfterRead`) without setting any SSE-C parameters. If `moveAfterRead` is enabled on an SSE-C encrypted bucket, this would fail with the same `InvalidRequest` error. This is out of scope for this PR but might be worth a follow-up JIRA ticket.

davsclaus (approved):
Clean, well-scoped fix. The AWS S3 `CopyObjectRequest` API requires both copy-source and destination SSE-C parameters when the source object is SSE-C encrypted — the producer only set the destination side, so every SSE-C copy failed with HTTP 400. The fix correctly adds the three missing `copySourceSSECustomer*` calls alongside the existing destination calls, and re-enables the previously disabled integration test.

Verified:
- Git history confirms the test was disabled because the producer was broken, not the test itself.
- The fix is additive — only activates when `useCustomerKey=true`, no regression risk for non-SSE-C usage.
- Branch naming, commit format, and attribution follow project conventions.
- Backport targets (4.18.x, 4.14.x) match the currently supported LTS lines.

PR #25029, opened 2026-07-22
CAMEL-24242: camel-aws2-athena - stop ignoring thread interruption while polling a query

oscerd:
## Problem

`Athena2QueryHelper.interrupted` gates both loops that drive `startQueryExecution`:

```java
if (this.interrupted) {
    LOG.trace("AWS Athena start query execution thread was interrupted, will try no more");
    return false;
}
```

It is read at `Athena2QueryHelper.java:113` (`shouldAttempt`) and `:146`
(`shouldWait`) — and **never assigned anywhere in the class**. Both guards are
dead code.

## How it regressed

This is a silent revert of a deliberate fix, so I want to be explicit about the
history:

- **CAMEL-20297** (`9e17300779`, Otavio Piske, Jan 2024) — *"camel-aws2-athena:
  do not swallow interrupted exceptions"* — added the interrupt handling:

  ```java
  try {
      Thread.sleep(this.currentDelay);
  } catch (InterruptedException e) {
      this.interrupted = Thread.interrupted(); // store, then clear, interrupt status
      LOG.trace("...wait thread was interrupted; will return at earliest opportunity");
      Thread.currentThread().interrupt();
  }
  ```

- **CAMEL-22949** (`1b0fca17c0`, PR #21215, Feb 2026) — *"Migrate components from
  Thread.sleep() to Camel's Task API"* — replaced that block with
  `Tasks.foregroundTask()...run(...)` and dropped the `catch` clause, which was
  the only writer of `interrupted`.

## Impact

`ForegroundTask.run()` *does* restore the interrupt status
(`ForegroundTask.java:124`) and return `false` — but `doWait()` discarded the
return value. The result is worse than a missed shutdown signal. After the first
interruption:

1. the thread's interrupt status stays set;
2. `shouldWait()` still returns `true`, because `interrupted` is `false`;
3. the next `doWait()` reaches `Thread.sleep(initialDelay)` inside
   `ForegroundTask.run()`, which throws `InterruptedException` **immediately**
   because the status is set, and returns at once;
4. the polling loop therefore spins with **no delay at all**, issuing
   `GetQueryExecution` calls as fast as the API allows until `waitTimeout`
   elapses — and up to `maxAttempts × waitTimeout` when
   `resetWaitTimeoutOnRetry=true`.

So interrupting an in-flight query (graceful shutdown, route stop) turns a
2-second-interval poll into a tight busy loop against the Athena API.

## Fix

Capture the `run()` return value and record the interruption:

```java
if (!completed && Thread.currentThread().isInterrupted()) {
    // the task API already restored the interrupt status, so only record it here to let the
    // attempt and wait loops bail out at the earliest opportunity
    this.interrupted = true;
    LOG.trace("AWS Athena start query execution wait thread was interrupted; will return at earliest opportunity");
}
```

The interrupt status itself is left in place — `ForegroundTask` already restored
it, which is what CAMEL-20297 wanted the caller to observe.

## Test

`Athena2QueryHelperTest.doWaitRecordsThreadInterruptionSoTheLoopsStop()` interrupts
the calling thread, calls `doWait()`, and asserts that `isInterrupted()`,
`shouldWait()` and `shouldAttempt()` all report the query loop should stop. It
clears the interrupt status in a `finally` block so it cannot leak into
neighbouring tests.

Verified it fails on `main` (`Expecting value to be true but was false`) and
passes with the fix. Full class: 23/23 green.

`assertj-core` is added test-scoped for the new assertions; the pre-existing
JUnit assertions in the file are left untouched.

## Backport

- `camel-4.18.x` — affected (carries the CAMEL-22949 migration) → backport opened.
- `camel-4.14.x` — **not** affected; it still has the original `Thread.sleep()`
  with the `catch (InterruptedException)` block, so there is nothing to fix there.

---

### DIFF
--- a/components/camel-aws/camel-aws2-athena/pom.xml
+++ b/components/camel-aws/camel-aws2-athena/pom.xml
@@ -76,6 +76,11 @@
             <version>${mockito-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- test infra -->
         <dependency>
--- a/components/camel-aws/camel-aws2-athena/src/main/java/org/apache/camel/component/aws2/athena/Athena2QueryHelper.java
+++ b/components/camel-aws/camel-aws2-athena/src/main/java/org/apache/camel/component/aws2/athena/Athena2QueryHelper.java
@@ -169,7 +169,7 @@ class Athena2QueryHelper {
     void doWait() {
         // Use Camel's task API for polling delay instead of Thread.sleep()
         // We use initialDelay for the actual delay, and maxIterations(1) to run once
-        Tasks.foregroundTask()
+        boolean completed = Tasks.foregroundTask()
                 .withBudget(Budgets.iterationBudget()
                         .withMaxIterations(1)
                         .withInitialDelay(Duration.ofMillis(this.currentDelay))
@@ -178,6 +178,14 @@ class Athena2QueryHelper {
                 .withName("AthenaQueryPollingDelay")
                 .build()
                 .run(exchange.getContext(), () -> true);
+
+        if (!completed && Thread.currentThread().isInterrupted()) {
+            // the task API already restored the interrupt status, so only record it here to let the
+            // attempt and wait loops bail out at the earliest opportunity
+            this.interrupted = true;
+            LOG.trace(
+                    "AWS Athena start query execution wait thread was interrupted; will return at earliest opportunity");
+        }
 
         this.currentDelay = this.delay;
     }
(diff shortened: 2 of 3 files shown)


gnodet (approved):
Clean, well-documented fix for the interrupt-handling regression introduced by the Task API migration (CAMEL-22949). The analysis tracing the issue through CAMEL-20297 → CAMEL-22949 is excellent, and the test correctly validates all three observable effects.

One minor note: adding `assertj-core` as a new test dependency for 4 boolean assertions could be replaced by JUnit's built-in `assertTrue`/`assertFalse` (the rest of the test file already uses them). Not blocking since assertj is managed in the parent POM and widely used elsewhere.

davsclaus (approved):
Clean bug fix — the regression analysis is thorough and the fix is correctly targeted.

Verified that `ForegroundTask.run()` catches `InterruptedException`, restores the interrupt status, and returns `false` — so the new `!completed && Thread.currentThread().isInterrupted()` guard correctly bridges the Task API's interrupt signaling back to the `Athena2QueryHelper.interrupted` field that gates both polling loops.

The test properly sets the interrupt flag before calling `doWait()`, asserts all three loop-exit conditions (`isInterrupted()`, `shouldWait()`, `shouldAttempt()`), and clears the flag in a `finally` block to prevent leakage.

davsclaus (approved):

PR #25032, opened 2026-07-22
CAMEL-24242: camel-aws2-athena - stop ignoring thread interruption while polling a query

oscerd:
Backport of #25029 to `camel-4.18.x`.

`Athena2QueryHelper.interrupted` gates both `shouldAttempt()` and `shouldWait()`
but has no writer, so both guards are dead code. CAMEL-20297 had added the
interrupt handling in `doWait()`; the CAMEL-22949 Task-API migration
(`1b0fca17c0`) dropped the `catch (InterruptedException)` block that was its only
writer.

Because `ForegroundTask.run()` restores the interrupt status and `doWait()`
discarded its return value, an interrupted poll loop spins with **no delay**,
hammering `GetQueryExecution` until `waitTimeout` elapses. The fix records the
interruption so both loops bail out.

This branch carries the CAMEL-22949 migration, so it is affected. `camel-4.14.x`
is **not** — it still has the original `Thread.sleep()` with its `catch` block,
so no backport is needed there.

Cherry-picked cleanly; `Athena2QueryHelperTest` passes 23/23 on this branch.

No upgrade-guide entry — this restores intended behaviour with no user-visible
configuration change.

---

### DIFF
--- a/components/camel-aws/camel-aws2-athena/pom.xml
+++ b/components/camel-aws/camel-aws2-athena/pom.xml
@@ -76,6 +76,11 @@
             <version>${mockito-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- test infra -->
        <dependency>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean and correct backport of the interrupt-handling fix from PR #25029 (main) to camel-4.18.x. The regression analysis is accurate — commit `1b0fca17c04` (CAMEL-22949, Task API migration) replaced `Thread.sleep()` with `Tasks.foregroundTask()...run()` but dropped the `catch (InterruptedException)` block, leaving the `interrupted` field as a dead variable.

The fix correctly captures the `ForegroundTask.run()` return value and sets `this.interrupted = true` when the task did not complete and the thread is interrupted, restoring the existing `shouldWait()` and `shouldAttempt()` guards. The test is well-structured with proper cleanup.

**Minor note (non-blocking):** The new test uses AssertJ (`assertThat`) while the existing 22 tests in the file use JUnit assertions (`assertTrue`/`assertFalse`/`assertEquals`). The PR description acknowledges this and keeps existing assertions untouched, which is reasonable.

LGTM.

davsclaus (approved):

PR #25035, opened 2026-07-22
CAMEL-24247: camel-jbang - Fix duplicate camel-micrometer-prometheus in exported pom.xml

davsclaus:
## Summary

Fixes [CAMEL-24247](https://issues.apache.org/jira/browse/CAMEL-24247).

When exporting a JBang project with metrics enabled, the exported pom.xml contains `camel-micrometer-prometheus` twice. The base class `ExportBaseCommand.resolveDependencies()` adds `camel:micrometer-prometheus` (shorthand format) at line 778, then the subclass `ExportCamelMain.resolveDependencies()` adds `mvn:org.apache.camel:camel-micrometer-prometheus` (full format) at line 305. Since these are different strings, the `Set` does not deduplicate them, and both end up in the generated pom.xml.

**Fix:** Switch the subclass to use `camel:micrometer-prometheus` shorthand (matching the base class format) so the `Set` naturally deduplicates. Applied the same fix for `camel-health`. No `removeIf` workaround needed.

## Test plan
- [x] Added `shouldNotDuplicateMetricsDependency` test that exports with `camel.management.metricsEnabled=true` and asserts `camel-micrometer-prometheus` appears exactly once
- [x] All 856 module tests pass

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportCamelMain.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportCamelMain.java
@@ -292,6 +292,7 @@ class ExportCamelMain extends Export {
         answer.removeIf(s -> s.contains("camel-core"));
         answer.removeIf(s -> s.contains("camel-main"));
         answer.removeIf(s -> s.contains("camel-health"));
+        answer.removeIf(s -> s.contains("camel-micrometer-prometheus"));
         // spring-boot-starter JARs are not usable in camel-main runtime
         answer.removeIf(s -> s.contains("spring-boot-starter"));
 
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportCamelMain.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportCamelMain.java
@@ -292,7 +292,6 @@ class ExportCamelMain extends Export {
         answer.removeIf(s -> s.contains("camel-core"));
         answer.removeIf(s -> s.contains("camel-main"));
         answer.removeIf(s -> s.contains("camel-health"));
-        answer.removeIf(s -> s.contains("camel-micrometer-prometheus"));
         // spring-boot-starter JARs are not usable in camel-main runtime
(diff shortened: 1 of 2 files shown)


davsclaus (commented):

The `removeIf` pattern `s.contains("camel-micrometer-prometheus")` does not match the `camel:micrometer-prometheus` entry added by the base class (`ExportBaseCommand.java:778`) — the colon vs dash means the substring check fails, so the duplicate persists.

**Suggested approach:** drop the new `removeIf` line and change line 305 to use `camel:micrometer-prometheus` shorthand instead of the `mvn:` format. The `Set` naturally deduplicates the identical string the base class already added. The same cleanup could be done for `camel-health` at line 309 → `camel:health` for consistency.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/ExportCamelMain.java
```diff
@@ -292,6 +292,7 @@ protected Set<String> resolveDependencies(Path settings, Path profile) throws Ex
         answer.removeIf(s -> s.contains("camel-core"));
         answer.removeIf(s -> s.contains("camel-main"));
         answer.removeIf(s -> s.contains("camel-health"));
+        answer.removeIf(s -> s.contains("camel-micrometer-prometheus"));
```
davsclaus: This `removeIf` won't catch the base-class entry because `"camel:micrometer-prometheus".contains("camel-micrometer-prometheus")` is **false** (colon ≠ dash).

Suggested fix: remove this line entirely and change line 305 below from:
```java
answer.add("mvn:org.apache.camel:camel-micrometer-prometheus");
```
to:
```java
answer.add("camel:micrometer-prometheus");
```

Same string the base class adds → `Set` deduplicates naturally, no `removeIf` needed. Same change could apply to `camel-health` at line 309 → `answer.add("camel:health")` for consistency.

davsclaus: _Claude Code on behalf of davsclaus_

Fixed. Switched both `camel-micrometer-prometheus` (line 305) and `camel-health` (line 309) to `camel:` shorthand format, and removed the `removeIf` line. The `Set` naturally deduplicates the identical string the base class already adds. Added a test (`shouldNotDuplicateMetricsDependency`) that verifies no duplicate entry in the exported pom.xml.

PR #25036, opened 2026-07-22
CAMEL-24248 Concurrent pollEnrich can fail to stop in-use consumers after cache eviction

msnijder30:
# Description

When pollEnrich processes concurrent exchanges using different dynamic endpoint URIs and its consumer cache is smaller than the concurrency level, a polling consumer’s pool can be evicted while the consumer is still in use. When that exchange finishes, the consumer is released after its pool has already been removed, so the consumer is not stopped and its resources remain allocated.

Repeated requests cause these unclosed consumers to accumulate, resulting in steadily increasing memory usage and potentially leaking component-specific resources such as connections, clients, threads, or buffers. In the attached reproducer, four concurrent SEDA consumers use dynamic polling URIs with cacheSize(1). After 400 requests, 304 polling consumers remain active and retain approximately 76 MB, whereas only the single cached consumer should remain.

 

We found out about this issue on production with SQS component with the setting concurrentConsumers=4 and after that a pollEnrich with cacheSize set to 1 (with sftp component). And we saw our application go OOM after about 4 days consistently.

 

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java
+++ b/core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java
@@ -94,12 +94,7 @@ abstract class ServicePool<S extends Service> extends ServiceSupport implements 
             }
         } else {
             // service no longer in a pool (such as being released twice, or can happen during shutdown of Camel etc)
-            ServicePool.stop(s);
-            try {
-                e.getCamelContext().removeService(s);
-            } catch (Exception ex) {
-                LOG.debug("Error removing service: {}. This exception is ignored.", s, ex);
-            }
+            stopAndRemove(s);
         }
     }
 
@@ -130,6 +125,18 @@ abstract class ServicePool<S extends Service> extends ServiceSupport implements 
         Pool<S> p = pool.get(endpoint);
         if (p != null) {
             p.release(s);
+        } else {
+            // the pool was evicted while this service was in use
+            stopAndRemove(s);
+        }
+    }
+
+    private void stopAndRemove(S s) {
+        ServicePool.stop(s);
+        try {
+            getEndpoint.apply(s).getCamelContext().removeService(s);
+        } catch (Exception e) {
+            LOG.debug("Error removing service: {}. This exception is ignored.", s, e);
         }
     }
 
--- a/core/camel-core/src/test/java/org/apache/camel/processor/PollEnrichConcurrentConsumersCacheTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/processor/PollEnrichConcurrentConsumersCacheTest.java
@@ -1,0 +1,239 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java
```diff
@@ -130,6 +125,18 @@ public void release(Endpoint endpoint, S s) {
         Pool<S> p = pool.get(endpoint);
         if (p != null) {
             p.release(s);
+        } else {
+            // the pool was evicted while this service was in use
+            stopAndRemove(s);
+        }
+    }
+
```
davsclaus, line 133: This is the key fix — previously when the pool was evicted while a service was in use, `release()` silently did nothing and the consumer leaked. Now it properly stops and removes the orphaned service. :+1:

core/camel-support/src/main/java/org/apache/camel/support/cache/ServicePool.java
```diff
@@ -94,12 +94,7 @@ protected void onEvict(S s) {
             }
         } else {
             // service no longer in a pool (such as being released twice, or can happen during shutdown of Camel etc)
-            ServicePool.stop(s);
-            try {
-                e.getCamelContext().removeService(s);
-            } catch (Exception ex) {
-                LOG.debug("Error removing service: {}. This exception is ignored.", s, ex);
-            }
+            stopAndRemove(s);
```
davsclaus, line 97: Minor nit: `onEvict()` already computed `Endpoint e = getEndpoint.apply(s)` at line 87, but `stopAndRemove(s)` recomputes it. The old code reused the cached `e`. Since `getEndpoint` is typically a simple field accessor this is negligible, but you could pass `e` as a parameter to avoid the redundant lookup.

davsclaus (approved):
Solid fix for a real production OOM issue. The root cause is clear: when `pollEnrich` uses a `cacheSize` smaller than the concurrency level, the LRU cache evicts a service's pool while the service is still in use. When `release()` is called afterward, `pool.get(endpoint)` returns `null` and the method silently does nothing — the consumer is never stopped, leaking resources.

The fix is correct and minimal:
1. In `release()`: when the pool is gone, stop and remove the orphaned service instead of silently ignoring it
2. Extract `stopAndRemove()` helper to deduplicate the stop+remove logic already present in `onEvict()`'s else-branch

I traced the eviction flow and confirmed no double-stop risk: `p.stop()` drains the pool queue, but the in-use consumer was never in the queue — it's only reachable through `release()`.

Test is well-constructed: uses `CountDownLatch` for deterministic concurrency (no `Thread.sleep`), AssertJ assertions, proper cleanup in `@AfterEach`.

Minor nit (non-blocking): the refactored `onEvict()` computes `Endpoint e` at the top, then the else-branch calls `stopAndRemove(s)` which recomputes `getEndpoint.apply(s)`. The old code reused the cached `e`. Negligible since `getEndpoint` is a simple field accessor, but could be cleaned up by passing `e` as a parameter.

gnodet (approved):
Clean, correct fix for a real production OOM bug. When `pollEnrich` uses a `cacheSize` smaller than the concurrency level, the LRU cache can evict a consumer's pool while the consumer is still in use. The current `release()` silently does nothing when the pool is gone, leaking the consumer. The fix properly stops and removes the orphaned consumer.

**Concurrency analysis:**
- The eviction flow is fully synchronous with `SimpleLRUCache`: `cache.putIfAbsent(B, B)` triggers `callEviction()` → `onEvict(A)` → `p.stop()` all on the same thread that called `acquire()`. This guarantees that by the time the thread using consumer A calls `release()`, the pool is already removed. No TOCTOU risk in the new `release()` path.
- No double-stop risk: `MultiplePool.stop()` drains the `queue` (which contains only idle/released consumers) and does NOT process the `evicts` deque. An in-use consumer is not in the queue, so it is only stopped once — by `stopAndRemove()` in the new `release()` path. Even if a double-stop occurred, `ServiceSupport.stop()` is idempotent.

**Test quality:** Well-constructed with `CountDownLatch` and `Future` for deterministic concurrency control, no `Thread.sleep()` calls (compliant with project rules). The `@AfterEach` cleanup properly releases latches before JUnit5 stops the CamelContext.

PR #25051, opened 2026-07-23
CAMEL-24222: Add camel_dependency_security_audit MCP tool

oscerd:
## Summary

Adds a new MCP tool `camel_dependency_security_audit` that performs security vulnerability analysis on a Camel project's dependencies by cross-referencing them with the Camel security advisory database.

- **Per-artifact CVE matching**: scans declared POM dependencies against known Camel CVEs at the project's version
- **Severity classification**: maps CVE titles to critical/high/medium severity
- **Reachability analysis**: when routes are provided, flags whether a vulnerable artifact is directly used (reachable) or only a transitive dependency
- **Upgrade recommendations**: suggests version upgrades and links to migration tools
- **POM sanitization**: masks credentials before processing (via `PomSanitizer`)

Distinct from `camel_dependency_check` (dependency hygiene) and `camel_security_advisories` (raw CVE listing) — this tool combines both into a project-specific vulnerability audit.

## Test plan

- [x] 7 new tests covering required input, older versions with CVEs, current versions, reachability, sanitization, recommendations, empty routes
- [x] All 307 existing MCP server tests pass (no regressions)

_Claude Code on behalf of oscerd_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25050, opened 2026-07-23
CAMEL-24221: Add camel_runtime_ai_trace MCP tool

oscerd:
## Summary

Adds a new MCP tool `camel_runtime_ai_trace` that traces AI-specific exchange flow in running Camel applications.

- **AI component detection**: identifies Bedrock, LangChain4j, Docling, Textract, OpenAI, KServe, TorchServe, DJL, HuggingFace components in running routes
- **AI header extraction**: captures token usage, model IDs, guardrail outcomes, completion reasons, streaming chunk counts from exchange headers
- **Processor statistics**: filters top processor stats to AI components only, showing latency (mean/max/min/last) and exchange counts
- **Structured summary**: aggregates key AI metrics (token usage, model, completion reason, guardrail action) into a summary

Combines data from three existing runtime queries (`get_history`, `get_top_processors`, `get_route_structure`) and filters for AI-relevant data.

## Test plan

- [x] 10 unit tests covering component extraction, header categorization, processor stats, scheme parsing, and edge cases
- [x] All 309 existing MCP server tests pass (no regressions)

_Claude Code on behalf of oscerd_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25048, opened 2026-07-23
CAMEL-24219: Add camel_ai_pipeline_scaffold MCP tool

oscerd:
## Summary

Adds a new MCP tool `camel_ai_pipeline_scaffold` that generates runnable YAML DSL Camel routes for AI document-processing pipelines.

- **4 pipeline types**: summarization, extraction, RAG (with ingestion + query routes), classification (with choice router)
- **3 document processors**: Docling (open-source), Textract (AWS-managed), combined (both)
- **3 source types**: local file, S3 bucket, HTTP URL
- **Configurable**: Bedrock model ID and AWS region
- **Output**: YAML DSL route + matching `application.properties` template + pipeline description

Generated routes use correct Camel component parameters:
- Bedrock: `modelId` as endpoint parameter (not headers), `operation: converse`
- Docling: `useDoclingServe: true`, `doclingServeUrl` parameter
- Textract: `operation: detectDocumentText`

## Test plan

- [x] 15 tests covering all pipeline types, processors, sources, and edge cases
- [x] Tests verify correct parameter names and absence of fabricated headers
- [x] All 317 existing MCP server tests pass (no regressions)

_Claude Code on behalf of oscerd_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25052, opened 2026-07-23
CAMEL-24223: Add camel_route_cost_estimate MCP tool

oscerd:
## Summary

Adds a new MCP tool `camel_route_cost_estimate` that estimates API costs for a Camel route based on its component usage patterns.

- **9 component cost profiles**: Bedrock (per-token), Textract (per-page), S3/SQS/SNS/Kinesis (per-request), OpenAI (per-token), LangChain4j Chat/Embeddings (per-token), Docling (free)
- **Per-execution cost breakdown**: sorted by most expensive component first
- **Projections**: hourly, daily, and monthly cost estimates at configurable throughput
- **Optimization tips**: context-aware suggestions (e.g., "use Docling instead of Textract for text extraction")
- **Configurable parameters**: messages/hour, avg pages, avg input/output tokens

### Example output structure
```json
{
  "costBreakdown": [...],
  "projection": {"messagesPerHour": 100, "hourly": "$1.05", "daily": "$25.20", "monthly": "$756.00"},
  "summary": {"mostExpensiveComponent": "aws-bedrock"},
  "optimizationTips": [...]
}
```

## Test plan

- [x] 10 new tests covering AI pipelines, Textract routes, simple routes, cost projection, optimization tips, scheme extraction, defaults, combined processors
- [x] All 310 existing MCP server tests pass (no regressions)

### DIFF
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
@@ -56,9 +56,11 @@ public class RouteCostEstimateTools {
     @Tool(annotations = @Tool.Annotations(readOnlyHint = true, destructiveHint = false, openWorldHint = false),
           description = "Estimate API costs for a Camel route based on its component usage. "
                         + "Analyzes a YAML or XML route definition and identifies components with "
-                        + "pay-per-use pricing (Bedrock LLM, Textract, S3, SQS, SNS, etc.). "
+                        + "pay-per-use pricing. Currently covers: Bedrock (runtime and agent-runtime), "
+                        + "Textract, S3, SQS, SNS, Kinesis, OpenAI, LangChain4j (chat and embeddings), "
+                        + "and Docling (free/self-hosted). "
                         + "Returns per-execution cost estimate and monthly projection at a given throughput. "
-                        + "Cost data is approximate based on published AWS pricing.")
+                        + "Cost data is approximate based on published AWS pricing (2025-Q2).")
     public CostEstimateResult camel_route_cost_estimate(
             @ToolArg(description = "The Camel route definition (YAML or XML)") String route,
             @ToolArg(description = "Expected messages per hour for cost projection (default: 100)") Integer messagesPerHour,
@@ -133,7 +135,8 @@ public class RouteCostEstimateTools {
                 projection, summary,
                 optimizationTips.isEmpty() ? null : optimizationTips,
                 "Cost estimates are approximate based on published AWS pricing (us-east-1). "
-                                                                      + "Actual costs vary by region, volume tier, and model.");
+                                                                      + "Actual costs vary by region, volume tier, and model.",
+                "2025-Q2");
     }
 
     List<String> extractSchemes(String route) {
@@ -192,6 +195,16 @@ public class RouteCostEstimateTools {
             @Override
             double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
                 return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
+        });
+
+        profiles.put("aws-bedrock-agent-runtime", new ComponentCostProfile(
+                "AWS Bedrock Agent Runtime (RAG/Flows)", "per-session",
+                "RetrieveAndGenerate: model invocation cost + ~$0.02/1000 retrieval units. "
+                                                                        + "InvokeFlow: depends on flow steps") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000) + 0.02 / 1000;
             }
         });
 
@@ -261,7 +274,7 @@ public class RouteCostEstimateTools {
 
         profiles.put("langchain4j-embeddings", new ComponentCostProfile(
                 "LangChain4j Embeddings (model-dependent)", "per-token",
-                "Embedding: ~$0.02/1M tokens (Titan Embed)") {
+                "Estimated as Titan Embed (~$0.02/1M tokens); free with Ollama/local models") {
             @Override
             double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
                 return inputTokens * 0.02 / 1_000_000;
@@ -315,7 +328,8 @@ public class RouteCostEstimateTools {
             CostProjection projection,
             CostSummary summary,
             List<String> optimizationTips,
-            String disclaimer) {
+            String disclaimer,
+            String pricingDataAsOf) {
     }
 
     public record ComponentCostBreakdown(
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
@@ -43,7 +43,8 @@ import io.quarkiverse.mcp.server.ToolCallException;
 public class RouteCostEstimateTools {
 
     private static final Pattern YAML_SCHEME_PATTERN = Pattern.compile(
-            "(?:uri:\\s*[\"']?|from:\\s+[\"']?|to:\\s+[\"']?|toD:\\s+[\"']?)([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
+            "(?:uri:\\s*[\"']?|from:[ \\t]+[\"']?|to:[ \\t]+[\"']?|toD:[ \\t]+[\"']?)"
+                                                                       + "([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
             Pattern.MULTILINE);
 
     private static final Pattern XML_SCHEME_PATTERN = Pattern.compile(
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
@@ -1,0 +1,317 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.dsl.jbang.core.commands.mcp;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import io.quarkiverse.mcp.server.Tool;
+import io.quarkiverse.mcp.server.ToolArg;
+import io.quarkiverse.mcp.server.ToolCallException;
+
+/**
+ * MCP Tool for estimating API costs of a Camel route based on its component usage.
+ * <p>
+ * Particularly useful for AI pipelines where Bedrock, Textract, and S3 calls have per-unit pricing. Provides
+ * per-execution cost estimates and projections at a given throughput.
+ * <p>
+ * Cost data is approximate and based on published AWS pricing as of 2025. Actual costs depend on model, region, and
+ * volume tiers.
+ */
+@ApplicationScoped
+public class RouteCostEstimateTools {
+
+    private static final Pattern SCHEME_PATTERN = Pattern.compile(
+            "(?:uri:\\s*[\"']?|from:\\s+[\"']?|to:\\s+[\"']?|toD:\\s+[\"']?)([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
+            Pattern.MULTILINE);
+
+    private static final Map<String, ComponentCostProfile> COST_PROFILES = buildCostProfiles();
+
+    @Tool(annotations = @Tool.Annotations(readOnlyHint = true, destructiveHint = false, openWorldHint = false),
+          description = "Estimate API costs for a Camel route based on its component usage. "
+                        + "Analyzes a YAML or XML route definition and identifies components with "
+                        + "pay-per-use pricing (Bedrock LLM, Textract, S3, SQS, SNS, etc.). "
+                        + "Returns per-execution cost estimate and monthly projection at a given throughput. "
+                        + "Cost data is approximate based on published AWS pricing.")
+    public CostEstimateResult camel_route_cost_estimate(
+            @ToolArg(description = "The Camel route definition (YAML or XML)") String route,
+            @ToolArg(description = "Expected messages per hour for cost projection (default: 100)") Integer messagesPerHour,
+            @ToolArg(description = "Average document pages per message for Textract/Docling (default: 5)") Integer avgPages,
+            @ToolArg(description = "Average LLM input tokens per request (default: 1000)") Integer avgInputTokens,
+            @ToolArg(description = "Average LLM output tokens per request (default: 500)") Integer avgOutputTokens) {
+
+        if (route == null || route.isBlank()) {
+            throw new ToolCallException("Route content is required", null);
+        }
+
+        int throughput = messagesPerHour != null && messagesPerHour > 0 ? messagesPerHour : 100;
+        int pages = avgPages != null && avgPages > 0 ? avgPages : 5;
+        int inTokens = avgInputTokens != null && avgInputTokens > 0 ? avgInputTokens : 1000;
+        int outTokens = avgOutputTokens != null && avgOutputTokens > 0 ? avgOutputTokens : 500;
+
+        List<String> detectedSchemes = extractSchemes(route);
+
+        List<ComponentCostBreakdown> breakdown = new ArrayList<>();
+        double totalPerExecution = 0;
+
+        for (String scheme : detectedSchemes) {
+            ComponentCostProfile profile = COST_PROFILES.get(scheme);
+            if (profile == null) {
+                continue;
+            }
+
+            double costPerExec = profile.estimateCostPerExecution(pages, inTokens, outTokens);
+            totalPerExecution += costPerExec;
+            breakdown.add(new ComponentCostBreakdown(
+                    scheme, profile.displayName(), profile.pricingModel(),
+                    costPerExec, profile.pricingNote()));
+        }
+
+        breakdown.sort(Comparator.comparingDouble(ComponentCostBreakdown::estimatedCostPerExecution).reversed());
+
+        double hourly = totalPerExecution * throughput;
+        double daily = hourly * 24;
+        double monthly = daily * 30;
+
+        String mostExpensive = breakdown.isEmpty() ? null : breakdown.get(0).scheme();
+
+        List<String> optimizationTips = buildOptimizationTips(detectedSchemes, breakdown);
+
+        CostProjection projection = new CostProjection(
+                throughput,
+                formatCost(hourly), formatCost(daily), formatCost(monthly));
+
+        CostSummary summary = new CostSummary(
+                formatCost(totalPerExecution),
+                breakdown.size(),
+                detectedSchemes.size(),
+                mostExpensive,
+                breakdown.isEmpty() ? "No pay-per-use components detected in this route" : null);
+
+        return new CostEstimateResult(
+                breakdown.isEmpty() ? null : breakdown,
+                projection, summary,
+                optimizationTips.isEmpty() ? null : optimizationTips,
+                "Cost estimates are approximate based on published AWS pricing (us-east-1). "
+                                                                      + "Actual costs vary by region, volume tier, and model.");
+    }
+
+    List<String> extractSchemes(String route) {
+        List<String> schemes = new ArrayList<>();
+        Matcher m = SCHEME_PATTERN.matcher(route);
+        while (m.find()) {
+            String scheme = m.group(1);
+            if (!schemes.contains(scheme)) {
+                schemes.add(scheme);
+            }
+        }
+        return schemes;
+    }
+
+    private List<String> buildOptimizationTips(List<String> schemes, List<ComponentCostBreakdown> breakdown) {
+        List<String> tips = new ArrayList<>();
+
+        if (schemes.contains("aws-bedrock")) {
+            tips.add("Consider using a smaller/cheaper Bedrock model for simpler tasks "
+                     + "(e.g., Nova Lite instead of Claude for classification)");
+            tips.add("Use streaming to reduce perceived latency without affecting token costs");
+        }
+        if (schemes.contains("aws2-textract") && schemes.contains("docling")) {
+            tips.add("Using both Textract and Docling — consider using only Docling (free, open-source) "
+                     + "for text extraction and reserving Textract for table/form extraction");
+        }
+        if (schemes.contains("aws2-s3")) {
+            tips.add("S3 GET costs are minimal but add up at scale — consider caching frequently accessed documents");
+        }
+        if (breakdown.size() > 1) {
+            tips.add("Most expensive component: " + breakdown.get(0).displayName()
+                     + " — focus optimization here for maximum savings");
+        }
+        return tips;
+    }
+
+    private static String formatCost(double cost) {
+        if (cost < 0.01) {
+            return String.format("$%.6f", cost);
+        }
+        return String.format("$%.4f", cost);
+    }
+
+    private static Map<String, ComponentCostProfile> buildCostProfiles() {
+        Map<String, ComponentCostProfile> profiles = new LinkedHashMap<>();
+
+        profiles.put("aws-bedrock", new ComponentCostProfile(
+                "AWS Bedrock (Claude Sonnet 4)", "per-token",
+                "Input: ~$3/MTok, Output: ~$15/MTok (Claude Sonnet 4 pricing)") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
+        });
+
+        profiles.put("aws2-textract", new ComponentCostProfile(
+                "AWS Textract", "per-page",
+                "DetectDocumentText: ~$1.50/1000 pages, AnalyzeDocument: ~$15/1000 pages") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return pages * 1.50 / 1000;
+            }
+        });
+
+        profiles.put("docling", new ComponentCostProfile(
+                "Docling (self-hosted)", "free",
+                "Open-source, self-hosted. Cost is only infrastructure (compute/memory)") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0;
+            }
+        });
+
+        profiles.put("aws2-s3", new ComponentCostProfile(
+                "AWS S3", "per-request",
+                "GET: ~$0.40/1M requests, PUT: ~$5/1M requests") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0.40 / 1_000_000;
+            }
+        });
+
+        profiles.put("aws2-sqs", new ComponentCostProfile(
+                "AWS SQS", "per-request",
+                "~$0.40 per 1M requests (Standard)") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0.40 / 1_000_000;
+            }
+        });
+
+        profiles.put("aws2-sns", new ComponentCostProfile(
+                "AWS SNS", "per-publish",
+                "~$0.50 per 1M publishes") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0.50 / 1_000_000;
+            }
+        });
+
+        profiles.put("aws2-kinesis", new ComponentCostProfile(
+                "AWS Kinesis", "per-shard-hour + per-PUT",
+                "~$0.015/shard/hour + $0.014/1M PUT units") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0.014 / 1_000_000;
+            }
+        });
+
+        profiles.put("langchain4j-chat", new ComponentCostProfile(
+                "LangChain4j Chat (model-dependent)", "per-token",
+                "Cost depends on the underlying model provider") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
+        });
+
+        profiles.put("langchain4j-embeddings", new ComponentCostProfile(
+                "LangChain4j Embeddings (model-dependent)", "per-token",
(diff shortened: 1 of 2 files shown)


dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+    private static Map<String, ComponentCostProfile> buildCostProfiles() {
+        Map<String, ComponentCostProfile> profiles = new LinkedHashMap<>();
+
+        profiles.put("aws-bedrock", new ComponentCostProfile(
+                "AWS Bedrock (Claude Sonnet 4)", "per-token",
+                "Input: ~$3/MTok, Output: ~$15/MTok (Claude Sonnet 4 pricing)") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
```
davsclaus, line 198: This hardcodes Claude Sonnet 4 pricing ($3/$15 per MTok), but users choose their Bedrock model — Nova Lite, Haiku, Titan, Mistral, etc. all have very different pricing. A route using Nova Lite would be dramatically overestimated.

Since the model isn't detectable from the route definition, at minimum the display name and pricing note should make clear this is a **specific model estimate**, not a Bedrock-wide one. Ideally the tool would accept a model parameter or default to a mid-range estimate rather than the most expensive model.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+        });
+
+        profiles.put("aws2-kinesis", new ComponentCostProfile(
+                "AWS Kinesis", "per-shard-hour + per-PUT",
+                "~$0.015/shard/hour + $0.014/1M PUT units") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return 0.014 / 1_000_000;
+            }
+        });
```
davsclaus, line 263: This uses the same Claude Sonnet 4 pricing ($3/$15 per MTok) as the Bedrock profile, but `langchain4j-chat` is provider-agnostic — it works with OpenAI, Ollama (free), Anthropic, Mistral, local models, etc. The display name says "model-dependent" but the calculation is fixed to one specific (expensive) model.

A user running langchain4j with Ollama would see significant phantom costs. At minimum, the pricing note should warn that the estimate assumes a specific cloud-hosted model and may be zero for self-hosted providers.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+            return String.format("$%.6f", cost);
+        }
+        return String.format("$%.4f", cost);
+    }
+
+    private static Map<String, ComponentCostProfile> buildCostProfiles() {
+        Map<String, ComponentCostProfile> profiles = new LinkedHashMap<>();
+
+        profiles.put("aws-bedrock", new ComponentCostProfile(
+                "AWS Bedrock (Claude Sonnet 4)", "per-token",
```
davsclaus: All pricing data is hardcoded in Java source with no mechanism to update it without a code change, rebuild, and release. Cloud providers change pricing regularly (AWS has changed Bedrock pricing multiple times since launch). Users running an older Camel version will silently get increasingly inaccurate estimates.

Consider externalizing the cost profiles to a JSON resource file (e.g., `cost-profiles.json` on the classpath) that could be updated independently, or at least add a clear note in the tool output stating when the pricing data was last updated.

davsclaus (changes requested):
Nice idea for an MCP tool — cost awareness is valuable for AI and cloud service routes.

Understanding the intent as a **cloud/AI service cost estimation tool** (not a general route cost calculator), the narrow component coverage is fine — the tool targets a specific set of pay-per-use services. A couple of suggestions to make that scope clearer and improve accuracy:

1. **Tool name/description should reflect the scope**: The current name `camel_route_cost_estimate` and description ("Estimate API costs for a Camel route") suggest general route costing. Consider renaming to something like `camel_route_cloud_cost_estimate` or updating the description to explicitly list the covered services (Bedrock, Textract, S3, SQS, SNS, Kinesis, OpenAI, LangChain4j, Docling) so users know what to expect.

2. **Model-specific pricing**: The Bedrock and langchain4j profiles hardcode Claude Sonnet 4 pricing, but the actual model matters a lot (Nova Lite vs Claude is 10-50x difference; langchain4j with Ollama is free). See inline comments for details.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        String pricingNote() {
+            return pricingNote;
+        }
+
+        abstract double estimateCostPerExecution(int pages, int inputTokens, int outputTokens);
+    }
+
+    // ---- Result records ----
```
gnodet, line 324: All four result records use package-private visibility. Every other tool class in this package (verified across 16+ files with 75+ records) uses `public record`. These records are serialized by Jackson via the MCP framework from a different package — package-private visibility could cause `InaccessibleObjectException` at runtime.

This is a recurring pattern from prior PRs (#25048, #25050, #25051). Please add `public` to all four records: `CostEstimateResult`, `ComponentCostBreakdown`, `CostProjection`, `CostSummary`.

gnodet (commented):
A few technical observations on this new MCP tool, complementing davsclaus's existing review. See inline comments for details.

The most significant confirmed issues are: (1) package-private records violating the universal `public record` convention across all existing tool classes, risking Jackson serialization failures at runtime; (2) missing the standard try/catch exception handling pattern; and (3) the scheme extraction regex failing on XML routes despite the tool description claiming XML support, plus a subtle YAML parsing bug.

Regarding the `langchain4j-chat` hardcoded pricing — davsclaus already flagged this. +1 on that feedback.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+                        + "pay-per-use pricing (Bedrock LLM, Textract, S3, SQS, SNS, etc.). "
+                        + "Returns per-execution cost estimate and monthly projection at a given throughput. "
+                        + "Cost data is approximate based on published AWS pricing.")
+    public CostEstimateResult camel_route_cost_estimate(
+            @ToolArg(description = "The Camel route definition (YAML or XML)") String route,
+            @ToolArg(description = "Expected messages per hour for cost projection (default: 100)") Integer messagesPerHour,
+            @ToolArg(description = "Average document pages per message for Textract/Docling (default: 5)") Integer avgPages,
+            @ToolArg(description = "Average LLM input tokens per request (default: 1000)") Integer avgInputTokens,
+            @ToolArg(description = "Average LLM output tokens per request (default: 500)") Integer avgOutputTokens) {
+
```
gnodet, line 70: The `estimateRouteCost` method is missing the standard try/catch exception handling pattern used by ~84% of existing tools in this package:

```java
try {
    // main logic
} catch (ToolCallException e) {
    throw e;
} catch (Throwable e) {
    throw new ToolCallException("Failed to estimate route cost: " + e.getMessage(), null);
}
```

Without this, any unexpected exception (regex failure, NPE) propagates as an unhandled error instead of a proper `ToolCallException`.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,317 @@
...
+ * Cost data is approximate and based on published AWS pricing as of 2025. Actual costs depend on model, region, and
+ * volume tiers.
+ */
+@ApplicationScoped
+public class RouteCostEstimateTools {
+
+    private static final Pattern SCHEME_PATTERN = Pattern.compile(
+            "(?:uri:\\s*[\"']?|from:\\s+[\"']?|to:\\s+[\"']?|toD:\\s+[\"']?)([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
+            Pattern.MULTILINE);
+
```
gnodet, line 49: The `SCHEME_PATTERN` regex only matches YAML-style syntax but the `@Tool` description claims "Analyzes a YAML or XML route definition". XML routes use `<from uri="scheme:..."/>` syntax that won't match this pattern, silently returning zero schemes.

Additionally, for standard YAML routes in `from:\n  uri: "scheme:..."` format, the regex captures `uri` as a false scheme name rather than the actual component scheme.

Consider adopting the approach from `ExplainTools` — iterating `CatalogService.loadCatalog().findComponentNames()` and checking `content.contains(comp + ":")`  — which handles any format uniformly. Alternatively, remove the XML claim from the `@Tool` description if only YAML is supported.

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateToolsTest.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Test
+    void shouldEstimateCostForAiPipeline() {
+        RouteCostEstimateTools.CostEstimateResult result
+                = tools.camel_route_cost_estimate(AI_PIPELINE_ROUTE, 100, 5, 1000, 500);
+
+        assertThat(result).isNotNull();
+        assertThat(result.costBreakdown()).isNotNull();
+        assertThat(result.costBreakdown()).hasSizeGreaterThanOrEqualTo(2);
```
gnodet, line 79: The assertions could be more precise:
- `hasSizeGreaterThanOrEqualTo(2)` — the expected count is exactly 3 (docling, aws-bedrock, aws2-s3 all have cost profiles)
- `startsWith("$")` (line 116) — only checks formatting, not the computed value
- `isNotEmpty()` for optimization tips (line 124) — doesn't verify which tips are returned for the given route

Also, there's no test for XML route format despite the tool description claiming XML support. Adding one would immediately reveal the regex limitation noted above.

oscerd:
Thank you both for the reviews. All findings addressed:

**Structural fixes:**
- ✅ All four result records now `public record` (matching convention across 16+ tool classes)
- ✅ Added standard try/catch exception handling pattern
- ✅ Added XML route support via `XML_SCHEME_PATTERN` regex — tool now correctly parses `<from uri="scheme:..."/>` syntax

**Model-dependent pricing (davsclaus + gnodet):**
- ✅ Bedrock display name now reads "AWS Bedrock (model-dependent, estimated as Claude Sonnet 4)" with a pricing note showing the range: Nova Lite ~$0.06/$0.24 vs Opus ~$15/$75
- ✅ LangChain4j Chat pricing note clarified: "free with Ollama/local models"
- Agreed that hardcoded pricing is a limitation — the disclaimer already notes costs are approximate. Externalizing pricing data to a config file is a good follow-up.

**Test additions:**
- ✅ New test for XML route scheme extraction

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,342 @@
...
+ * <p>
+ * Cost data is approximate and based on published AWS pricing as of 2025. Actual costs depend on model, region, and
+ * volume tiers.
+ */
+@ApplicationScoped
+public class RouteCostEstimateTools {
+
+    private static final Pattern YAML_SCHEME_PATTERN = Pattern.compile(
+            "(?:uri:\\s*[\"']?|from:\\s+[\"']?|to:\\s+[\"']?|toD:\\s+[\"']?)([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
+            Pattern.MULTILINE);
```
gnodet, line 48: **YAML regex captures `uri` as a false scheme in multi-line routes.** The `\s+` in `from:\s+` matches newlines, so `from:\n  uri: "aws-bedrock:model"` captures `uri` instead of `aws-bedrock`. This silently omits the most expensive component from the cost estimate.

The `uri:\s*` alternative already handles the multi-line case correctly — the `from:/to:/toD:` alternatives only need to match inline same-line syntax, so `[ \t]+` (horizontal whitespace) is sufficient:

```suggestion
    private static final Pattern YAML_SCHEME_PATTERN = Pattern.compile(
            "(?:uri:\\s*[\"']?|from:[ \\t]+[\"']?|to:[ \\t]+[\"']?|toD:[ \\t]+[\"']?)([a-zA-Z][a-zA-Z0-9+.-]*):(?://)?",
            Pattern.MULTILINE);
```

gnodet (commented):
Re-review: three of four original findings were addressed (public records, try/catch, XML support). However, the YAML regex bug persists — `from:\s+` matches across newlines, causing `uri` to be falsely captured as a scheme in multi-line YAML routes.

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateToolsTest.java
```diff
@@ -0,0 +1,179 @@
...
+    @Test
+    void shouldProvideOptimizationTips() {
+        RouteCostEstimateTools.CostEstimateResult result
+                = tools.camel_route_cost_estimate(AI_PIPELINE_ROUTE, null, null, null, null);
+
+        assertThat(result.optimizationTips()).isNotNull();
+        assertThat(result.optimizationTips()).isNotEmpty();
+    }
+
+    @Test
```
gnodet, line 131: Minor: `contains("docling", "aws-bedrock", "aws2-s3")` won't detect the false `uri` entry caused by the regex bug. Using `containsExactlyInAnyOrder("file", "docling", "aws-bedrock", "aws2-s3")` would catch both the false positive and the missing `file` scheme.

oscerd:
Re-review finding addressed:

✅ Fixed YAML regex false scheme capture — replaced `\s+` with `[ \t]+` in `from:/to:/toD:` alternatives so the pattern only matches inline syntax (same line) and doesn't cross newlines. Multi-line `from:\n  uri: "scheme:..."` now correctly matches only via the `uri:` alternative.

Test now asserts `doesNotContain("uri")` on extracted schemes.

gnodet (approved):
All previously requested changes are now addressed across the three commits.

**YAML regex fix verified:** The three `from:\s+`, `to:\s+`, `toD:\s+` alternatives now correctly use `[ \t]+`, which restricts matching to horizontal whitespace (space and tab), preventing the pattern from crossing newlines into a subsequent `uri:` key. The `uri:\s*` alternative correctly retains `\s*` since it is the terminal key that directly precedes the scheme value.

The new `doesNotContain("uri")` assertion in `shouldExtractSchemesCorrectly` directly validates the bug scenario: multi-line YAML `from:\n  uri: "scheme:..."` no longer falsely captures `uri` as a scheme.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,343 @@
...
+
+    private static final Pattern XML_SCHEME_PATTERN = Pattern.compile(
+            "(?:<from|<to|<toD)\\s+uri=[\"']([a-zA-Z][a-zA-Z0-9+.-]*):",
+            Pattern.CASE_INSENSITIVE);
+
+    private static final Map<String, ComponentCostProfile> COST_PROFILES = buildCostProfiles();
+
+    @Tool(annotations = @Tool.Annotations(readOnlyHint = true, destructiveHint = false, openWorldHint = false),
+          description = "Estimate API costs for a Camel route based on its component usage. "
+                        + "Analyzes a YAML or XML route definition and identifies components with "
```
davsclaus, line 58: The `@Tool` description says "Estimate API costs for a Camel route" which implies general coverage, but only 9 specific services have cost profiles. Consider either:
- Renaming to `camel_route_cloud_cost_estimate` to signal the scope, or
- Adding to the description: "Currently covers: Bedrock, Textract, S3, SQS, SNS, Kinesis, OpenAI, LangChain4j, Docling."

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,343 @@
...
+        Map<String, ComponentCostProfile> profiles = new LinkedHashMap<>();
+
+        profiles.put("aws-bedrock", new ComponentCostProfile(
+                "AWS Bedrock (model-dependent, estimated as Claude Sonnet 4)", "per-token",
+                "Estimate uses Claude Sonnet 4 pricing ($3/$15 per MTok). Actual cost depends on model: "
+                                                                                            + "Nova Lite ~$0.06/$0.24, Haiku ~$0.25/$1.25, Opus ~$15/$75") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
```
davsclaus, line 198: Camel has three Bedrock schemes: `aws-bedrock`, `aws-bedrock-agent`, `aws-bedrock-agent-runtime`. Routes using `aws-bedrock-agent-runtime` for knowledge-base retrieval, RAG, or flow invocations have per-session/per-invocation costs that won't be detected by this tool. Consider adding a cost profile for `aws-bedrock-agent-runtime`.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,343 @@
...
+                mostExpensive,
+                breakdown.isEmpty() ? "No pay-per-use components detected in this route" : null);
+
+        return new CostEstimateResult(
+                breakdown.isEmpty() ? null : breakdown,
+                projection, summary,
+                optimizationTips.isEmpty() ? null : optimizationTips,
+                "Cost estimates are approximate based on published AWS pricing (us-east-1). "
+                                                                      + "Actual costs vary by region, volume tier, and model.");
+    }
```
davsclaus, line 140: The disclaimer mentions "approximate based on published AWS pricing" but doesn't indicate when the pricing data was last verified. Adding a `pricingDataAsOf` field (e.g., `"2025-Q2"`) to the result would help users gauge accuracy over time without needing to externalize the pricing model.

davsclaus (commented):
Nice work on this MCP tool — cost awareness for AI/cloud routes is genuinely useful. The technical issues from the prior round are all resolved; a few additional observations below.

### Confirmed Issues (still outstanding from prior review)

**1. Tool name implies general route costing, but scope is narrow (9 services)**
The name `camel_route_cost_estimate` and description "Estimate API costs for a Camel route" suggest comprehensive coverage. In reality, only Bedrock, Textract, S3, SQS, SNS, Kinesis, OpenAI, LangChain4j, and Docling are covered. Renaming (e.g., `camel_route_cloud_cost_estimate`) or explicitly listing covered services in the `@Tool` description would set accurate expectations. *(Aligns with davsclaus's finding #1)*

**2. Bedrock/LangChain4j estimates default to one of the most expensive models**
The Bedrock profile uses Claude Sonnet 4 pricing ($3/$15 per MTok). A route using Nova Lite ($0.06/$0.24) would be overestimated by ~50x. The updated display name ("estimated as Claude Sonnet 4") helps, but users relying on the numeric output will get misleading projections. Same applies to `langchain4j-chat` which silently assumes the same expensive model. *(Aligns with davsclaus's findings #2 and #3)*

### New Findings

**3. Missing `aws-bedrock-agent-runtime` cost profile**
Camel has three Bedrock schemes: `aws-bedrock`, `aws-bedrock-agent`, `aws-bedrock-agent-runtime`. The cost profiles only match `aws-bedrock`. Routes using `aws-bedrock-agent-runtime` (knowledge base retrieval, RAG, flow invocations) have per-session costs that won't be detected. At minimum, `aws-bedrock-agent-runtime` should have a cost profile entry.

**4. No pricing data timestamp in output**
The disclaimer says "approximate based on published AWS pricing" but doesn't say *when* the data was last verified. A simple field like `pricingDataAsOf: "2025-Q2"` in the result would help users gauge accuracy as the tool ages — a lightweight alternative to externalizing the full pricing model.

### Questions

**5. `langchain4j-embeddings` pricing consistency**
This profile uses Titan Embed pricing ($0.02/MTok) but LangChain4j embeddings support any provider (including free local models). Should this follow the same "model-dependent" labeling as `langchain4j-chat`?

---

This review covers convention compliance and rules-based review. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,343 @@
...
+
+        profiles.put("langchain4j-chat", new ComponentCostProfile(
+                "LangChain4j Chat (model-dependent)", "per-token",
+                "Cost depends on the underlying model provider. Estimated as Claude Sonnet 4; "
+                                                                   + "free with Ollama/local models") {
+            @Override
+            double estimateCostPerExecution(int pages, int inputTokens, int outputTokens) {
+                return (inputTokens * 3.0 / 1_000_000) + (outputTokens * 15.0 / 1_000_000);
+            }
+        });
```
davsclaus, line 273: This uses Titan Embed pricing ($0.02/MTok), but LangChain4j embeddings support any provider — including free local models via Ollama. For consistency with the `langchain4j-chat` profile, the display name could follow the same "(model-dependent)" pattern and the pricing note should mention that self-hosted models are free.

oscerd:
All davsclaus suggestions applied:

1. ✅ @Tool description now lists all covered services (Bedrock runtime + agent-runtime, Textract, S3, SQS, SNS, Kinesis, OpenAI, LangChain4j, Docling)
2. ✅ Added aws-bedrock-agent-runtime cost profile for RAG/Flows (model invocation + retrieval units)
3. ✅ Added pricingDataAsOf field to result ("2025-Q2")
4. ✅ LangChain4j embeddings note updated: "free with Ollama/local models"

gnodet (approved):
**Re-review** (1 new commit since last approval)

Commit 3bd06cd99e16 correctly addresses all 4 of davsclaus's feedback items: (1) @Tool description now lists all 11 covered services, (2) aws-bedrock-agent-runtime cost profile added with retrieval-unit pricing, (3) `pricingDataAsOf` field added to CostEstimateResult, (4) langchain4j-embeddings pricing note updated. The YAML regex fix (`[ \t]+` instead of `\s+`) from the prior commit remains correctly in place. CI passes.

davsclaus (approved):

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
```diff
@@ -0,0 +1,357 @@
...
+        if (breakdown.size() > 1) {
+            tips.add("Most expensive component: " + breakdown.get(0).displayName()
+                     + " — focus optimization here for maximum savings");
+        }
+        return tips;
+    }
+
+    private static String formatCost(double cost) {
+        if (cost < 0.01) {
+            return String.format("$%.6f", cost);
```
gnodet, line 183: Minor: `String.format("$%.6f", cost)` uses the JVM default locale. In locales with comma decimal separators (e.g., French), this would produce `$0,003000`. Consider using `String.format(Locale.ROOT, "$%.6f", cost)` for consistent output to MCP clients.

```suggestion
            return String.format(Locale.ROOT, "$%.6f", cost);
```

gnodet (approved):
All previously flagged issues from both review rounds (public records, try/catch pattern, YAML regex, XML support, aws-bedrock-agent-runtime profile, pricingDataAsOf, langchain4j-embeddings labeling) are confirmed addressed across the four commits. Clean implementation.

1. **Locale-sensitive formatting**: `String.format("$%.6f", cost)` in `formatCost()` uses the JVM default locale. In locales with comma decimal separators (e.g., French), this would produce `$0,003000` instead of `$0.003000`. Since these are currency strings for MCP clients, `String.format(Locale.ROOT, ...)` would ensure consistent dot-separated decimals.

2. **Test assertion precision**: `hasSizeGreaterThanOrEqualTo(2)` on `componentCosts` could be `hasSize(3)` since the AI_PIPELINE_ROUTE has exactly 3 priced components (docling, aws-bedrock, aws2-s3). More precise assertions catch regressions more reliably.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Addressed the 'at minimum' ask: the Bedrock display name is now "AWS Bedrock (model-dependent, estimated as Claude Sonnet 4)", its pricingNote spells out Nova Lite ~$0.06/$0.24, Haiku ~$0.25/$1.25, Opus ~$15/$75, and the top-level disclaimer (8cc9bbd) now states these are specific-model estimates, not a Bedrock-wide figure, and that cheaper/local models cost far less. The optional model parameter (the 'ideally') I'd add as a follow-up if you want per-model scaling — happy to. Leaving that as your call.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Same treatment for langchain4j-chat: display name says "(model-dependent)", the pricingNote says "Estimated as Claude Sonnet 4; free with Ollama/local models", and the disclaimer now explicitly calls out that a route on a local/self-hosted model (Ollama) will cost ~zero, so the estimate is not read as a real figure for those. Provider-agnostic phantom cost is now clearly flagged rather than silent.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Fair point on staleness. For now the data carries a pricingDataAsOf field (2025-Q2) and the disclaimer marks it approximate, and the immediate accuracy concern (model-specific labeling) is handled in-code. Externalizing to a classpath cost-profiles.json so it can be refreshed without a rebuild is a reasonable follow-up; I'd rather do that as its own change than fold a resource-loading mechanism into this PR. Noting it here.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Done — all four result records (CostEstimateResult, ComponentCostBreakdown, CostProjection, CostSummary) are public record, so Jackson/MCP can serialize them across package boundaries.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Done — camel_route_cost_estimate now wraps the logic in try { ... } catch (ToolCallException e) { throw e; } catch (Throwable e) { throw new ToolCallException(...); }, matching the package convention.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Done — there are now separate YAML and XML scheme patterns. XML_SCHEME_PATTERN matches <from|<to|<toD uri="scheme:", and the YAML pattern's from:/to:/toD: alternatives use [ \t]+ (horizontal whitespace) so they no longer span a newline into uri:, while the uri:\s* ["']? alternative captures the real scheme. shouldExtractSchemesFromXmlRoutes covers XML.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateToolsTest.java
oscerd: Tightened — shouldEstimateCostForAiPipeline now asserts exactly 3 priced components and pins the computed aws-bedrock per-execution cost (not just formatting), and there is a dedicated shouldExtractSchemesFromXmlRoutes test. shouldExtractSchemesCorrectly now uses containsExactlyInAnyOrder.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Done — the YAML pattern uses from:[ \t]+ / to:[ \t]+ / toD:[ \t]+ (horizontal whitespace only), so from:\n  uri: "aws-bedrock:..." no longer captures uri; the uri:\s*["']? alternative captures aws-bedrock. Pinned by the containsExactlyInAnyOrder assertion in shouldExtractSchemesCorrectly.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateToolsTest.java
oscerd: Done — shouldExtractSchemesCorrectly now asserts containsExactlyInAnyOrder("file", "docling", "aws-bedrock", "aws2-s3"), which catches both a false uri entry and a missing file scheme.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
oscerd: Done in 8cc9bbd — formatCost now uses String.format(Locale.ROOT, ...) for both branches, so the decimal separator is always '.' regardless of server locale.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

oscerd:
Note: #25052 had already merged (at 3bd06cd) when I pushed the review refinements referenced in the thread replies above, so those commits did not make it into this PR. I've opened them as a follow-up in #25167 (the `Locale.ROOT` formatting fix, the model-specific disclaimer, and the tighter tests). Apologies for the confusing thread references.

PR #25072, opened 2026-07-23
CAMEL-23703: camel launcher self-update and doctor installation diagnostics

ammachado:
# Description

Follow-up to the JReleaser packaging work merged in #24915 (CAMEL-23703). Adds self-update and
installation-diagnostics capabilities to the **Camel CLI Launcher**, plus hardening of the
packaging/publishing scripts.

Two new user-facing commands:

* **`camel self-update`** - checks for and installs a newer launcher release by re-running the
  same published `install.sh` / `install.ps1` installer, pinned to the resolved version. It
  refuses to act on an installation managed by a package manager (Homebrew, Chocolatey, WinGet,
  Scoop, SDKMAN) or by JBang, naming that manager's own upgrade command instead. Every `camel`
  invocation (except `self-update` itself) also prints a one-line "update available" notice, at
  most once every 24 hours; set `CAMEL_SELF_UPDATE_CHECK=false` to disable it.
* **`camel doctor`** - now additionally reports every Camel CLI installation found on the machine
  across the web installer and all supported package managers, marks which one is active on
  `PATH`, and exits non-zero when more than one is found.

**Self-update (launcher, `selfupdate` package)**
- `InstallScriptFetcher` / `ManifestFetcher` - download and version-compare against the published
  website manifest and `install.sha256`.
- `SelfUpdateCommand` - delegates to the published installer, pinned to the resolved version, and
  refuses to run on a package-manager- or JBang-managed install.
- `UpdateChecker` + `SelfUpdatePlugin` - background, once-per-24h update-available notice wired
  into `CamelLauncherMain`.
- Pinned-version tracking so a deliberately pinned install is never silently moved: `install.sh`
  / `install.ps1` write a pin on an explicit `--version`/`-Version` run and clear it on a bare
  run, skipping this when `CAMEL_INSTALL_SELF_UPDATE=true` (which `SelfUpdateCommand` always sets
  on its own delegated invocation).

**Installation diagnostics (jbang core)**
- `InstallDetector` - locates every Camel CLI installation across the web installer and package
  managers, plus `pinnedVersion()` / `pinnedVersionFile()`.
- `Doctor` - reports conflicting installations and marks the one active on `PATH`.

**Packaging / publishing hardening**
- `camel-publish.sh` + `lib/publish-state.sh` + `lib/assert-camel-cli.sh` - a real homebrew-core
  PR workflow that actually clones and PRs against the external repo (the previous version ran git
  commands against the script's own working tree).
- `camel-validate.sh` - local Homebrew validation hardening.
- `camel-package.sh` / `stage-winget-distro.sh` - guard possibly-empty array expansion so the
  scripts run under bash 3.2 (macOS default) with `set -u`.
- Rename the Homebrew formula to `apache-camel`, pin the JDK to 21, add `depends_on`, fix the
  escaped livecheck regex, and add `deprecate!`/`disable!` lifecycle to the versioned formula.
- Publish `install.sha256` alongside `install.sh` / `install.ps1`.

**Documentation**
- New `camel-jbang-launcher-install.adoc` page documenting the installers, `self-update`, and
  `doctor`.
- Upgrade-guide entry in `camel-4x-upgrade-guide-4_22.adoc` covering both new commands and the
  `CAMEL_SELF_UPDATE_CHECK` toggle.

**Testing**
Unit and integration tests added: `InstallDetectorTest`, `DoctorTest`, `ManifestFetcherTest`,
`InstallScriptFetcherTest`, `UpdateCheckerTest`, `WebsiteInstallTest`,
`WebsiteManifestGeneratorTest`, and an end-to-end `SelfUpdateIntegrationTest` for the delegation
flow. `assertj-core` declared as a test dependency in `camel-launcher`.

# Target

# Tracking

# AI-assisted contributions

---
_AI-generated PR description by Claude Code on behalf of ammachado._


oscerd (approved):

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
```diff
@@ -0,0 +1,553 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# ── camel-publish.sh — single resumable publish orchestrator ───────────────
+#
+# Usage:
+#   camel-publish.sh <version> --channel stable|lts [--lts-line X.Y]
+#
+# Behaviour:
```
davsclaus, line 24: **Bug:** This runs `publish-state.sh` in a subprocess — none of its functions (`_init_state`, `state_mark`, `state_current_status`, `state_resume_ok`, `state_redacted_dump`) will be defined in the caller. The script calls all of these later and will fail at runtime.

`camel-validate.sh` correctly uses `. "$LIB_DIR/assert-camel-cli.sh"` (dot-sourcing) — this should do the same:

```suggestion
. "$LIB_DIR/publish-state.sh"
```

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
```diff
@@ -0,0 +1,553 @@
...
+
+_log "State file: $PUBLISH_STATE_FILE"
+_log "Resume status: $(state_resume_ok)"
+
+# If already fully done, stop (idempotent re-run)
+if [ "$(state_resume_ok)" = "yes" ]; then
+  _log "All steps already completed. Workflow is complete and idempotent."
+  exit 0
+fi
+
```
davsclaus, line 255: **Bug:** `${!hook:-}` is bash indirect variable expansion — it dereferences the variable whose name is stored in `$hook`. This is not POSIX-compatible and will fail or produce incorrect behavior under `#!/bin/sh` on systems where `/bin/sh` is not bash (e.g., dash on Debian/Ubuntu).

Either change the shebang to `#!/usr/bin/env bash`, or replace the indirect expansion with an explicit check, e.g.:
```sh
if [ -n "${CAMEL_PUB_JRELEASER:-}" ]; then
    eval "$CAMEL_PUB_JRELEASER ..."
```

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/lib/publish-state.sh
```diff
@@ -0,0 +1,190 @@
...
+    return 1
+  fi
+
+  val="$(_read_val "$_f" "$_key")"
+  # Check if key exists at all (even with empty value)
+  if ! grep -q "^${_key}=" "$_f" 2>/dev/null; then
+    return 1
+  fi
+
+  if _is_secret_key "$_key"; then
```
davsclaus, line 70: **Bug:** `$'\n'` is bash ANSI-C quoting, not POSIX sh. Under a strict `#!/bin/sh` shell (e.g., dash), this will produce the literal string `$'\n'` instead of a newline character.

This pattern appears in several places in this file. Either change the shebang to `#!/usr/bin/env bash`, or use a literal newline variable:
```sh
NL='
'
# then use ${NL} where $'\n' appears
```
or use `printf`.

davsclaus (changes requested):
Thanks for this substantial contribution — the self-update and doctor diagnostics are well-designed, with thorough test coverage and clean separation between detection (`camel-jbang-core`) and update logic (`camel-launcher`). Three shell-script issues to address before merging:

### Summary

1. **Bug (High):** `camel-publish.sh` runs `publish-state.sh` in a subprocess (`sh`) instead of dot-sourcing it (`.`) — none of its functions will be available at runtime.
2. **Bug (Medium):** `camel-publish.sh` uses `${!hook:-}` (bash indirect expansion) under a `#!/bin/sh` shebang — will fail on strict POSIX shells.
3. **Bug (Medium):** `publish-state.sh` uses `$'\n'` (bash ANSI-C quoting) under a `#!/bin/sh` shebang — will produce literal strings instead of newlines on strict POSIX shells.

This review covers project rules, conventions, and visible risks. It does not replace specialized review tools such as CodeRabbit or SonarCloud.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
ammachado: Fixed in c4b2da24d7e5 — `sh "$LIB_DIR/publish-state.sh"` is now `. "$LIB_DIR/publish-state.sh"`, matching the dot-sourcing pattern already used for `assert-camel-cli.sh` in `camel-validate.sh`. Verified under `dash` that `_init_state`/`state_mark`/etc. are now defined in the caller.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
ammachado: Fixed in c4b2da24d7e5 — replaced the `${!hook:-}` indirect expansion with a direct check on `${CAMEL_PUB_JRELEASER:-}`, so no bash-only syntax remains under the `#!/bin/sh` shebang. Verified under `dash` with and without the override set.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/lib/publish-state.sh
ammachado: Fixed in c4b2da24d7e5 — introduced a `_NL` variable holding a literal newline (`NL='<newline>'`), and replaced all three `$'\n'` occurrences in `state_set` with `${_NL}`. Verified under `dash` that `state_set`/`state_get` round-trip correctly with real newline-separated entries.

ammachado:
@davsclaus all three issues from your review are fixed in c4b2da24d7e5:

1. `camel-publish.sh` now dot-sources `publish-state.sh` instead of running it as a subprocess.
2. The JReleaser override hook no longer uses bash indirect expansion (`${!hook}`) — checks `${CAMEL_PUB_JRELEASER:-}` directly.
3. `publish-state.sh` no longer uses `$'\n'` (bash ANSI-C quoting) — uses a literal-newline variable instead.

All three verified to parse and run correctly under `dash` (strict POSIX sh), not just macOS's bash-as-`/bin/sh`.

While in there, also fixed a few related bugs in `camel-publish.sh` found during the same pass: a broken `$(_operator)` command substitution that should have been `$_operator`, three PR-body attribution lines that were single-quoted and never expanded, and a dead/broken override block in `__dest_website` (replaced with the same `${FORK_REMOTE:-upstream}` pattern already used by `__dest_winget`/`__dest_scoop`). Replied inline on each of your three threads with specifics.

Separately, `install.sh` now uses `wget -nv` instead of `-q` (so failures print wget's own diagnostic, matching curl's `-fsSL`/`-S` behavior), and tar archive validation surfaces tar's actual error message instead of a generic "not a valid tar.gz".

Re-requesting your review — let me know if anything still looks off.

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/ManifestFetcher.java
```diff
@@ -0,0 +1,174 @@
...
+            String key = line.substring(0, eq);
+            String value = line.substring(eq + 1);
+            if (!Set.of("format", "version", "tar_sha256", "zip_sha256").contains(key)) {
+                throw new SelfUpdateException("manifest has unknown key: " + key);
+            }
+            if (values.containsKey(key)) {
+                throw new SelfUpdateException("manifest has duplicate key: " + key);
+            }
+            values.put(key, value);
+        }
```
davsclaus, line 162: **Bug: `parse()` will reject every manifest generated by this PR's own `WebsiteManifestGenerator`.**

`WebsiteManifestGenerator.renderManifest()` now prepends a 17-line ASF license header (`LICENSE_HEADER`, all `## ...` comment lines) to every generated manifest. Both `install.sh` and `install.ps1` were updated to skip `#` comment lines. `WebsiteManifestGenerator.parseStrictManifest()` was also updated. But this method was not — it counts raw lines and rejects anything that isn't exactly 4.

In production, `ManifestFetcher` fetches `latest.properties` from `camel.apache.org` (which will have the license header after a release built from this branch). The ~21-line manifest will be rejected with `"manifest must contain exactly four lines"`. Consequences:
- `camel self-update` prints an error and exits 1
- The background update-check silently fails forever (swallowed in `UpdateChecker.refreshCache()`) — users never see the "update available" notice

The test suite doesn't catch this because `SelfUpdateIntegrationTest` uses `WebsiteInstallTest.publishLatest()` which publishes manifests **without** the header.

Fix: add comment-line filtering here (skip lines starting with `#`), matching what `install.sh`, `install.ps1`, and `WebsiteManifestGenerator.parseStrictManifest()` already do. Then add a `ManifestFetcherTest` case that parses a manifest with the license header to prevent regression.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
```diff
@@ -0,0 +1,545 @@
...
+  _log "  WinGet: done."
+}
+
+__dest_winget || true
+
+# ── Destination 5: Scoop fork + PR ───────────────────────────────────────────
+__dest_scoop() {
+  [ "$_failed" -eq 1 ] && return 0
+
+  if [ "$(state_current_status scoop)" = "done" ]; then
```
davsclaus, line 440: **Bug: literal `[REDACTED]` is sent as the SDKMAN Consumer-Secret.**

The script requires `SDKMAN_CONSUMER_SECRET` at the top (`${SDKMAN_CONSUMER_SECRET:?Error: ...}`) but the curl call here sends the literal string `[REDACTED]` instead of `${SDKMAN_CONSUMER_SECRET}`. SDKMAN API calls will always fail with an auth error.

Since `camel-publish.sh` isn't wired into CI yet this is non-blocking, but worth fixing now.

davsclaus (changes requested):
Thanks for this thorough and well-structured PR — the self-update design (pinned-version TOCTOU protection, package-manager detection, SHA-256 verification of installer scripts) is solid, and the test coverage is comprehensive.

I found one confirmed bug that will make the self-update feature non-functional against manifests generated by this PR's own code, plus a secondary issue in the publish script.

This review checks the PR against project rules and conventions. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/ManifestFetcher.java
```diff
@@ -0,0 +1,174 @@
...
+    // re-served the same manifest.
+    static Manifest parse(byte[] content) {
+        String text = new String(content, StandardCharsets.UTF_8);
+        String[] rawLines = text.split("\n", -1);
+        int lineCount = rawLines.length > 0 && rawLines[rawLines.length - 1].isEmpty()
+                ? rawLines.length - 1
+                : rawLines.length;
+        if (lineCount != 4) {
+            throw new SelfUpdateException("manifest must contain exactly four lines");
+        }
```
gnodet: **High:** This enforces exactly 4 lines but does not skip comment lines. `WebsiteManifestGenerator.renderManifest()` now prepends a 16-line `##` license header to every generated manifest. All other parsers were updated:
- `install.sh`: `case "$key" in \#*) continue ;; esac`
- `parseStrictManifest()`: `if (line.startsWith("#")) continue`

But `ManifestFetcher.parse()` was not. Once the first release ships manifests with the header, `camel self-update` and `UpdateChecker` will throw "manifest must contain exactly four lines" for every check.

Fix: filter out `#`-prefixed and blank lines before the count check, matching the other parsers.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
```diff
@@ -0,0 +1,545 @@
...
+  API_URL="${CAMEL_PUB_SDKMAN_BASE:-https://vendor.sdkman.io}"
+
+  # Use vendor release endpoint
+  curl -s -X POST "$API_URL/release" \
+    -H "Consumer-Key: ${SDKMAN_CONSUMER_KEY}" \
+    -H "Consumer-Secret: [REDACTED]" \
+    -H "Content-Type: application/json" \
+    -d "{\"candidate\": \"camel\", \"version\": \"$VERSION\", \"default\": $SDKMAN_DEFAULT}" 2>/dev/null || {
+    _log "  SDKMAN release failed (likely no credentials in this env).";
+    state_mark "sdkman" "failed"; return 1;
```
gnodet, line 497: **Medium:** This sends the literal string `[REDACTED]` instead of `${SDKMAN_CONSUMER_SECRET}`. The `Consumer-Key` header on the preceding line correctly uses `${SDKMAN_CONSUMER_KEY}`, and the variable is validated as required at the top of the script.

```suggestion
      -H "Consumer-Secret: ${SDKMAN_CONSUMER_SECRET}" \
```

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/InstallScriptFetcher.java
```diff
@@ -0,0 +1,162 @@
...
+        try {
+            Path scriptPath = stagingDir.resolve(scriptName);
+            Files.write(scriptPath, scriptBytes);
+            return scriptPath;
+        } catch (IOException e) {
+            throw new SelfUpdateException("failed to stage " + scriptName, e);
+        }
+    }
+
+    private byte[] get(String url) {
```
gnodet, line 100: **Medium:** `InterruptedException` is caught and wrapped in `SelfUpdateException` without calling `Thread.currentThread().interrupt()` to restore the interrupt flag. Same issue exists in `ManifestFetcher.get()`. Standard Java practice requires restoring the flag so callers (including `UpdateChecker`'s daemon thread) can detect the interruption.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/InstallDetector.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+        if (normalized.toLowerCase().contains("/chocolatey/")) {
+            return InstallMethod.CHOCOLATEY;
+        }
+        if (normalized.contains("/WinGet/Packages/")) {
+            return InstallMethod.WINGET;
+        }
+        if (normalized.toLowerCase().contains("/scoop/apps/")) {
+            return InstallMethod.SCOOP;
+        }
```
gnodet, line 89: **Low:** WinGet path classification uses case-sensitive `normalized.contains("/WinGet/Packages/")` while Chocolatey (line 86) and Scoop both use case-insensitive checks via `normalized.toLowerCase().contains(...)`. Since Windows filesystem paths are case-insensitive, this could misclassify a WinGet installation whose path casing differs.

gnodet (changes requested):
Well-designed self-update and doctor diagnostics with thorough test coverage. The delegation to published install scripts avoids duplicating complex archive-safety logic, and the TOCTOU-prevention (pinning resolved version) and self-lockout prevention are well thought through.

However, two bugs will cause runtime failures:

| # | Severity | Finding |
|---|----------|--------|
| 1 | **high** | `ManifestFetcher.parse()` breaks on manifests with the ASF license header |
| 2 | **medium** | SDKMAN publish sends literal `[REDACTED]` instead of the secret |
| 3 | medium | `InterruptedException` caught without restoring interrupt flag |
| 4 | low | WinGet path check is case-sensitive while Chocolatey/Scoop are case-insensitive |

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/ManifestFetcher.java
ammachado: Fixed in f94d9a597 — `parse()` now skips `#`-prefixed lines before the four-line check, mirroring `install.sh`/`install.ps1`/`WebsiteManifestGenerator.parseStrictManifest()`. Added `parsesManifestWithAsfLicenseHeader()` as the regression test you asked for.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
ammachado: Fixed in f94d9a597 — `[REDACTED]` is now `${SDKMAN_CONSUMER_SECRET}`.

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/ManifestFetcher.java
ammachado: Same fix as davsclaus's comment above, landed in f94d9a597 — `parse()` now skips `#`-prefixed lines before the four-line check, and there's a regression test (`parsesManifestWithAsfLicenseHeader`) covering the license-header case.

dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-publish.sh
ammachado: Same fix as davsclaus's comment above, landed in f94d9a597 — literal `[REDACTED]` replaced with `${SDKMAN_CONSUMER_SECRET}`.

dsl/camel-jbang/camel-launcher/src/main/java/org/apache/camel/dsl/jbang/launcher/selfupdate/InstallScriptFetcher.java
ammachado: Fixed in f94d9a597 — both `InstallScriptFetcher.get()` and `ManifestFetcher.get()` now call `Thread.currentThread().interrupt()` in a dedicated `catch (InterruptedException e)` block before wrapping in `SelfUpdateException`.

dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/InstallDetector.java
ammachado: Fixed in f94d9a597 — the WinGet check is now `normalized.toLowerCase().contains("/winget/packages/")`, matching the case-insensitive Chocolatey/Scoop checks. Verified `InstallDetectorTest` (14 tests) still passes.

ammachado:
Pushed f94d9a597 addressing the two new review rounds:

- **davsclaus**: `ManifestFetcher.parse()` rejecting license-headered manifests (with the regression test requested), and the SDKMAN literal-secret bug.
- **gnodet**: same two bugs (independently found), plus `InterruptedException` not restoring the interrupt flag in `InstallScriptFetcher.get()`/`ManifestFetcher.get()`, and the case-sensitive WinGet path check in `InstallDetector.classify()`.

All four fixes verified with a full `mvnd clean install -Dquickly` build plus targeted test runs (`ManifestFetcherTest`, `InstallScriptFetcherTest`, `InstallDetectorTest` — all green).

davsclaus (approved):

gnodet (approved):
**Re-review** (commit f94d9a597 since last review)

All four findings from our previous REQUEST_CHANGES have been correctly addressed:

1. ✅ **ManifestFetcher.parse()** now skips `#`-prefixed comment lines before the 4-line count check, with a regression test covering the ASF license header
2. ✅ **SDKMAN publish** now correctly references `${SDKMAN_CONSUMER_SECRET}` instead of the literal `[REDACTED]`
3. ✅ **InterruptedException** handling in both `InstallScriptFetcher.get()` and `ManifestFetcher.get()` now restores the interrupt flag via `Thread.currentThread().interrupt()` in separate catch blocks
4. ✅ **WinGet path check** now uses `normalized.toLowerCase().contains("/winget/packages/")`, consistent with Chocolatey and Scoop

PR #25085, opened 2026-07-24
CAMEL-24244: Fix DefaultStreamCachingStrategy.updateSpool releasing the lock

oscerd:
## The bug

`UtilizationStatistics.updateSpool()` calls `lock.lock()` again in its `finally` block instead of `lock.unlock()`:

```java
void updateSpool(long size) {
    lock.lock();
    try {
        spoolAverageSize.set(spoolSize.addAndGet(size) / spoolCounter.incrementAndGet());
    } finally {
        lock.lock();     // should be unlock()
    }
}
```

The sibling `updateMemory()` immediately above it is correct, so this reads as a copy-paste slip.

Because the lock is a `ReentrantLock`, the calling thread does **not** self-deadlock — it just leaves the hold count at 2 and never releases it. The consequences are:

- the hold count grows without bound;
- **every other thread entering `updateSpool()` blocks forever**;
- **`CamelContext` shutdown deadlocks too** — `doStop()` calls `statistics.reset()`, which acquires the same lock, so the thread stopping the context also blocks.

`ReentrantLock.lock()` is uninterruptible, so neither `shutdownNow()` nor interruption can recover a thread once it is stuck.

It is reachable whenever stream-caching statistics are enabled and a stream is spooled to disk (`computeStatistics` → `updateSpool`, from `doCache`). Memory-only caching takes the correct `updateMemory` path and is unaffected.

## The fix

One character: `lock.lock()` → `lock.unlock()`.

## The test

`StreamCachingStrategySpoolStatisticsTest` drives the spool accounting from two threads. A single-threaded test cannot catch this, since a reentrant lock lets its owner re-acquire it — only a *second* thread blocks.

Three deliberate choices keep a future regression from wedging CI rather than failing it:

- **a standalone `DefaultStreamCachingStrategy`** instead of the context-managed one. Using the context's instance would deadlock `CamelContext` shutdown in `doStop()` → `statistics.reset()`, and the build would hang instead of reporting a failure. I hit exactly that while developing this: an early version wedged the surefire fork until its 50-minute timeout.
- **daemon worker threads**, so a permanently blocked thread cannot keep the JVM alive.
- **a stub converter returning a cache that reports itself as spooled**, exercising the spool accounting directly without depending on real disk spooling (and therefore without depending on the context's spool configuration).

## Verification

Both directions were checked locally against freshly built artifacts:

- **with the fix** — passes.
- **with the one-line change reverted** — fails in seconds with `both threads must finish; one blocked on the leaked spool statistics lock`.

---

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultStreamCachingStrategy.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultStreamCachingStrategy.java
@@ -583,7 +583,7 @@ public class DefaultStreamCachingStrategy extends ServiceSupport implements Came
             try {
                 spoolAverageSize.set(spoolSize.addAndGet(size) / spoolCounter.incrementAndGet());
             } finally {
-                lock.lock();
+                lock.unlock();
             }
         }
 
--- a/core/camel-core/src/test/java/org/apache/camel/impl/engine/StreamCachingStrategySpoolStatisticsTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/engine/StreamCachingStrategySpoolStatisticsTest.java
@@ -1,0 +1,171 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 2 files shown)


davsclaus (approved):
Excellent work — the bug is clearly confirmed via `git blame` (copy-paste slip from `ee18bef57fe8`, CAMEL-20199, Jul 2024), the one-character fix is correct, and the test is thoughtfully designed.

The multi-threaded test is essential here since `ReentrantLock` lets its owner re-acquire, so a single-threaded test would never catch this. The three safety mechanisms (standalone strategy, daemon threads, `@Timeout`) ensure a future regression fails fast in CI rather than wedging the build — well thought through.

No issues found.

gnodet (approved):
Clean, correct fix for a real concurrency bug. The one-line change (`lock.lock()` → `lock.unlock()` in the `finally` block of `UtilizationStatistics.updateSpool()`) is obviously right — the original typo caused a permanent lock leak that would deadlock any second thread touching the spool statistics.

The regression test is well-crafted:

- **Correct reproduction**: uses two threads with a `CountDownLatch` gate to force contention on the statistics lock — the only way to surface the bug, since `ReentrantLock` lets a single thread re-acquire silently.
- **Build safety**: `@Timeout(60)` + daemon threads ensure a regression can never hang the build.
- **Clean isolation**: stub `TypeConverter` + `SpooledCache` exercise the spool accounting path without real disk I/O.
- **Follows conventions**: JUnit 5, AssertJ, no `Thread.sleep`, extends `ContextTestSupport`.

No issues found.

PR #25086, opened 2026-07-24
CAMEL-23467: camel-aws-bedrock - add invokeAgent and invokeInlineAgent operations

oscerd:
### Motivation

[CAMEL-23467](https://issues.apache.org/jira/browse/CAMEL-23467). The `aws-bedrock-agent-runtime` component did not expose **InvokeAgent** (invoking a deployed Bedrock agent) or **InvokeInlineAgent** (defining the agent at invocation time) — the primary entry points for agent-based workflows on Bedrock.

### Approach

Both APIs return event-stream responses, and the component already had a working precedent for that in `invokeFlow`: the AWS SDK **async** client, a response handler whose visitor collects the emitted events, awaited with `join()` and unwrapping the `CompletionException` that wraps AWS service exceptions. Both new operations follow that same shape rather than introducing parallel machinery.

### Changes

- **Operations**: `invokeAgent` and `invokeInlineAgent` added to `BedrockAgentRuntimeOperations` and wired into the producer.
- **Options** (each overridable by header): `agentId`, `agentAliasId`, `sessionId`, `streamOutputMode`, and for the inline variant `foundationModel` / `instruction`. `enableTrace` is reused and its label/doc widened to cover agents.
- **Async client**: provisioned for the two new operations, exactly as for `invokeFlow` (so users don't pay for a Netty client when only doing synchronous operations).
- **Event handling**: all four `ResponseStream` events (`chunk`, `trace`, `returnControl`, `files`) are collected. Traces, return-control payloads, files and citations are published as headers. Citations arrive on the **chunk attribution** rather than as their own event, which is where they're picked up.
- **`streamOutputMode`**: honored with the same semantics as the runtime component — `complete` (default) accumulates the chunks into a `String` body, `chunks` exposes the `List<String>`. Overridable per exchange by header.
- **Sessions**: the agent APIs are session-based, so a session id is generated when neither config nor header supplies one, and the id actually used is always echoed back on `CamelAwsBedrockAgentRuntimeSessionId` so a route can continue the conversation.
- **POJO mode**: `pojoRequest=true` accepts an `InvokeAgentRequest` / `InvokeInlineAgentRequest` unchanged — which is also how the richer inline options (action groups, knowledge bases, collaborators, guardrails) are reachable without adding a dozen URI options.
- **Docs**: three new sections in `aws-bedrock-agent-runtime-component.adoc` plus the supported-operations list.

One thing worth flagging for review: `invokeAgent` and `invokeInlineAgent` use **parallel but distinct** SDK type families (`PayloadPart`/`TracePart`/… vs `InlineAgentPayloadPart`/`InlineAgentTracePart`/…). The collected traces/return-controls/files are therefore held as the SDK objects and the header metadata documents both variants, rather than pretending a single type applies.

### Testing

`BedrockAgentRuntimeProducerInvokeAgentTest` — **9 tests, all passing** — mirroring the existing `invokeFlow` test approach (mocked async client capturing the request):

- request built from string body + endpoint config, for both operations
- header overrides for agent id/alias/session/trace/endSession/memoryId, and for inline model/instruction
- session id generated when not configured, and echoed back on the response
- missing `agentId` / `foundationModel` produce a clear validation error
- `streamOutputMode` branching shapes the body (`String` vs `List`), including the header override

The test uses AssertJ assertions and package-private visibility per project conventions. Note this required adding a **test-scoped `assertj-core` dependency** to `camel-aws-bedrock` — it was not on the module's test classpath. That mirrors how `camel-aws2-sqs`, `camel-aws2-athena` and `camel-aws2-sts` declare it, and the version is managed in `parent/pom.xml`, so no new version is introduced.

Full reactor `mvn clean install -DskipTests` from root: success, with catalog/DSL metadata regenerated.

*Note on the regenerated `aws-bedrock-agent` metadata:* that component declares `headersClass = BedrockAgentRuntimeConstants.class` (pre-existing), so adding constants there also regenerates its header metadata. I verified this is index renumbering plus the new entries — all existing `Flow*` headers remain present (15 → 26 headers, none removed).


davsclaus (approved):

components/camel-aws/camel-aws-bedrock/src/test/java/org/apache/camel/component/aws2/bedrock/agentruntime/BedrockAgentRuntimeProducerInvokeAgentTest.java
```diff
@@ -0,0 +1,243 @@
...
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.lenient;
+
+@ExtendWith(MockitoExtension.class)
+public class BedrockAgentRuntimeProducerInvokeAgentTest {
```
davsclaus: Nit: Per project conventions, new test classes and test methods should not use the `public` modifier (JUnit 5 doesn't require it). Since this class doesn't extend `CamelTestSupport`, the class declaration, `@BeforeEach`, `@AfterEach`, and all `@Test` methods can be package-private.

```suggestion
class BedrockAgentRuntimeProducerInvokeAgentTest {
```

components/camel-aws/camel-aws-bedrock/src/test/java/org/apache/camel/component/aws2/bedrock/agentruntime/BedrockAgentRuntimeProducerInvokeAgentTest.java
```diff
@@ -0,0 +1,243 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import software.amazon.awssdk.services.bedrockagentruntime.BedrockAgentRuntimeAsyncClient;
+import software.amazon.awssdk.services.bedrockagentruntime.model.InvokeAgentRequest;
+import software.amazon.awssdk.services.bedrockagentruntime.model.InvokeAgentResponseHandler;
+import software.amazon.awssdk.services.bedrockagentruntime.model.InvokeInlineAgentRequest;
+import software.amazon.awssdk.services.bedrockagentruntime.model.InvokeInlineAgentResponseHandler;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
davsclaus: Nit: Project conventions prefer AssertJ assertions (`assertThat(...)`) over JUnit assertions for new test code. For example:

```java
import static org.assertj.core.api.Assertions.assertThatThrownBy;

assertThat(request.agentId()).isEqualTo("agent-1");
assertThat(result.getMessage().getBody()).isInstanceOf(String.class);
assertThatThrownBy(() -> template.sendBody("direct:agent-no-config", "hi"))
    .isInstanceOf(CamelExecutionException.class)
    .hasCauseInstanceOf(IllegalArgumentException.class);
```

davsclaus (commented):
Nice feature addition — the implementation cleanly follows the established `invokeFlow` patterns, and the test coverage is solid.

Two minor convention notes inline (non-blocking). This review does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

components/camel-aws/camel-aws-bedrock/src/test/java/org/apache/camel/component/aws2/bedrock/agentruntime/BedrockAgentRuntimeProducerInvokeAgentTest.java
oscerd: Done in 3343311 — dropped `public` from the class declaration, `@BeforeEach`/`@AfterEach` and all nine `@Test` methods. Nothing here overrides a supertype method, so package-private applies throughout.

components/camel-aws/camel-aws-bedrock/src/test/java/org/apache/camel/component/aws2/bedrock/agentruntime/BedrockAgentRuntimeProducerInvokeAgentTest.java
oscerd: Done in 3343311 — all assertions converted to AssertJ, including `assertThatThrownBy(...).cause().isInstanceOf(IllegalArgumentException.class).hasMessageContaining(...)` for the two validation tests.

One thing to flag: `camel-aws-bedrock` did not have AssertJ on its test classpath (it is not pulled transitively), so the commit also adds a test-scoped `assertj-core` dependency. That matches how `camel-aws2-sqs`, `camel-aws2-athena`, `camel-aws2-sts` and others in the same parent folder declare it, and the version is managed in `parent/pom.xml` — so no new version is introduced. Shout if you would rather not add the dependency to this module and I will revert to JUnit assertions.

The 9 tests pass against the module build.

gnodet (commented):
Well-structured feature addition that cleanly follows the established `invokeFlow` patterns. The `AgentInvocationResult` inner class and `joinAgentInvocation`/`resolveRequired` helpers are good DRY extractions. Test coverage is solid (9 tests) and documentation is thorough.

Two minor consistency items:

**1. `streamOutputMode` field initialization** (low) — In `BedrockAgentRuntimeConfiguration.java`, the `streamOutputMode` field is declared as `private String streamOutputMode;` (defaults to null). The existing `BedrockConfiguration` (runtime component, line 78) initializes the same field: `private String streamOutputMode = "complete";`. The `defaultValue = "complete"` in `@UriParam` is metadata-only. The code works because `prepareAgentResponse` treats null as the complete case, but for consistency the field should be initialized:
```java
private String streamOutputMode = "complete";
```

**2. Async client Javadoc** (low) — The `setBedrockAgentRuntimeAsyncClient` Javadoc still says "required for invokeFlow which streams events back" but the async client is now also required for `invokeAgent` and `invokeInlineAgent` (both call `requireAsyncClient()`). Should mention all three streaming operations.

davsclaus (approved):
Well-structured feature addition — the implementation cleanly follows the established `invokeFlow` patterns, the `AgentInvocationResult` inner class and `joinAgentInvocation`/`resolveRequired`/`resolveSessionId` helpers are good DRY extractions, and the test coverage is solid (9 tests).

The two minor items flagged by @gnodet (`streamOutputMode` field initialization and async client Javadoc) are the only remaining notes — both non-blocking.

This review does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

PR #25091, opened 2026-07-24
CAMEL-23453: camel-keycloak - add federated identity linking operations

oscerd:
### Motivation

[CAMEL-23453](https://issues.apache.org/jira/browse/CAMEL-23453). The component exposed user CRUD but no federated-identity (IdP) link management. Those links are what connect a Keycloak user to an external identity-provider account (Google, GitHub, a SAML provider), so they matter for SSO scenarios and for provisioning users with pre-linked accounts during a migration.

### Changes

Three new operations, implemented against the admin client's `UserResource` and following the existing producer pattern (header validation, then the admin-client call, then the response on the out-message):

| Operation | Admin client call | Required headers |
|---|---|---|
| `addFederatedIdentity` | `addFederatedIdentity(alias, rep)` | realm, userId, `CamelKeycloakIdentityProvider`, `CamelKeycloakFederatedUserId` |
| `removeFederatedIdentity` | `removeFederatedIdentity(alias)` | realm, userId, `CamelKeycloakIdentityProvider` |
| `getFederatedIdentities` | `getFederatedIdentity()` | realm, userId |

New header constants, as proposed on the issue: `CamelKeycloakIdentityProvider`, `CamelKeycloakFederatedUserId`, `CamelKeycloakFederatedUsername`.

`getFederatedIdentities` sets the body to the `List<FederatedIdentityRepresentation>`.

Two behaviours worth calling out for review:

- **Username fallback.** `CamelKeycloakFederatedUsername` is optional; when it is absent the external user id is used as the username rather than creating the link with a blank one, since Keycloak surfaces that field in the admin console. A test pins this — happy to switch it to leaving the field null if you'd prefer.
- **Status checking.** `addFederatedIdentity` returns a JAX-RS `Response`; the producer checks it and fails with the status code on a non-2xx (e.g. a 409 when the link already exists) instead of reporting success. The response is closed via try-with-resources.

### Testing

`KeycloakProducerFederatedIdentityTest` — **9 tests, all passing** — using the same mocked-admin-client approach as the existing `KeycloakProducerTest`:

- add: the captured `FederatedIdentityRepresentation` carries the right provider/userId/username
- add: username falls back to the external user id when not supplied
- add: a non-2xx status fails the exchange and reports the code
- remove: verifies `removeFederatedIdentity` is called with the alias
- list: the body is the list of links
- validation: missing realm, userId, identity provider, and federated user id each produce the documented error

Full reactor `mvn clean install -DskipTests` from root: success, with catalog and endpoint-DSL metadata regenerated.

Docs: a *Federated Identity Operations* section in `keycloak-component.adoc` with the header contract and Java/YAML examples.


davsclaus (commented):
Nice, clean PR — the implementation follows the existing producer pattern well, the documentation is thorough with both Java and YAML examples, and the test coverage is solid with 9 tests. CI checks all pass.

A few suggestions below, mostly around test conventions. Nothing blocking.

This is a rules-and-conventions review. It does not replace specialized review tools such as CodeRabbit or Sourcery, or static analysis tools such as SonarCloud.

_This review was generated by an AI agent (Claude Code on behalf of @davsclaus) and may contain inaccuracies. Please verify all suggestions before applying._

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
```diff
@@ -0,0 +1,237 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for the federated identity (identity provider link) operations.
+ */
+public class KeycloakProducerFederatedIdentityTest extends CamelTestSupport {
+
```
davsclaus, line 50: Per project test conventions, new test classes should be package-private (no `public` modifier). Same applies to all `@Test` methods in this file. The existing `KeycloakProducerTest` uses `public` (pre-existing style), but new files should follow the current convention.

```suggestion
class KeycloakProducerFederatedIdentityTest extends CamelTestSupport {
```

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
```diff
@@ -0,0 +1,237 @@
...
+        try {
+            template.sendBodyAndHeaders("direct:addFederatedIdentity", null,
+                    headers(KeycloakConstants.IDENTITY_PROVIDER, "google",
+                            KeycloakConstants.FEDERATED_USER_ID, "google-123"));
+        } catch (Exception e) {
+            assertTrue(e.getCause().getMessage().contains("409"),
+                    "The failing status should be reported — was: " + e.getCause().getMessage());
+        }
+    }
```
davsclaus, line 133: This try-catch pattern silently passes if the exception is never thrown — a broken status check would go unnoticed. The same pattern exists in `KeycloakProducerTest`, but for new code `assertThrows` is more robust:

```java
Exception e = assertThrows(Exception.class, () ->
    template.sendBodyAndHeaders("direct:addFederatedIdentity", null,
            headers(KeycloakConstants.IDENTITY_PROVIDER, "google",
                    KeycloakConstants.FEDERATED_USER_ID, "google-123")));
assertThat(e.getCause().getMessage()).contains("409");
```

Same concern applies to `testMissingIdentityProvider`, `testMissingFederatedUserId`, `testMissingUserId`, and `testMissingRealmName` below.

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
```diff
@@ -0,0 +1,237 @@
...
+                        KeycloakConstants.FEDERATED_USERNAME, "jane@example.com"));
+
+        MockEndpoint.assertIsSatisfied(context);
+
+        ArgumentCaptor<FederatedIdentityRepresentation> captor
+                = ArgumentCaptor.forClass(FederatedIdentityRepresentation.class);
+        verify(userResource).addFederatedIdentity(eq("google"), captor.capture());
+
+        FederatedIdentityRepresentation sent = captor.getValue();
+        assertEquals("google", sent.getIdentityProvider());
```
davsclaus: Minor: project guidelines prefer AssertJ assertions (`assertThat(...)`) over JUnit assertions for new test code. The existing `KeycloakProducerTest` uses JUnit style, but new files should prefer the updated convention. For example:

```java
assertThat(sent.getIdentityProvider()).isEqualTo("google");
assertThat(sent.getUserId()).isEqualTo("google-123");
assertThat(sent.getUserName()).isEqualTo("jane@example.com");
```

davsclaus (approved):

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
oscerd: Done in baac4b6 — the class and all @Test methods are now package-private (createRouteBuilder stays protected as an override, and the inner RouteBuilder.configure() stays public since it overrides a public method).\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
oscerd: Done in baac4b6 — the error-status test now uses assertThatThrownBy(...).cause().hasMessageContaining("409"), so it fails if the exception is not thrown. The validation tests use the same pattern.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

components/camel-keycloak/src/test/java/org/apache/camel/component/keycloak/KeycloakProducerFederatedIdentityTest.java
oscerd: Done in baac4b6 — all assertions are AssertJ now (assertThat(sent.getIdentityProvider()).isEqualTo(...), etc.). Added a test-scoped assertj-core dependency since camel-keycloak did not have it on the test classpath; the version is managed in the parent.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

davsclaus (approved):

gnodet (approved):
Clean, well-structured additive PR that adds three federated identity operations (add/remove/get) to the Keycloak component, following existing producer patterns consistently with comprehensive tests, proper documentation, and correct generated file regeneration.

- The error-status test asserts `.cause().hasMessageContaining("409")` but does not also assert the exception type. Chaining `.isInstanceOf(IllegalStateException.class)` before `.hasMessageContaining("409")` would make the test more precise.

**Positive notes:**
- The `addFederatedIdentity` implementation properly uses try-with-resources for the JAX-RS Response and checks the status code — actually an improvement over some existing operations that set the unclosed Response as the message body.
- All review feedback from davsclaus (package-private test class, AssertJ assertions, assertThatThrownBy) has been addressed in the second commit.

PR #25093, opened 2026-07-24
camel-jbang - Redact secrets in structured MCP tool results, not only String results

oscerd:
## What

Completes the secret-redaction coverage of the Camel JBang MCP server's opt-in security layer (added in CAMEL-24218). Redaction previously only inspected `String` tool results, so tools returning a `JsonObject` bypassed it entirely.

## Why

The runtime tools return `org.apache.camel.util.json.JsonObject`, not `String`. `camel_runtime_properties` in particular dumps a running application's configuration, which is exactly where a datasource password or an API key surfaces. With redaction enabled, that output was still emitted verbatim because the interceptor only redacted `String` results:

```java
if (config.isRedactionEnabled() && result instanceof String s) { ... }
```

## Change

Reuses Camel's own knowledge of what a secret is (per review feedback), rather than a hand-rolled word list:

- `McpSecretRedactor.redactStructured(Object)` walks `Map`/`List` results in place (`JsonObject` is a `LinkedHashMap`, `JsonArray` an `ArrayList`) and blanks the value of any entry whose key `SensitiveUtils.containsSensitive(...)` reports as sensitive, which is the same helper the jbang `ListProperties` command uses. It also passes string leaf values through the free-text masker. The instance is mutated in place, so a tool's declared return type is preserved.
- Free-text `String` results are masked with Camel's `DefaultMaskingFormatter`, which blanks key=value, XML and JSON values keyed off `SensitiveUtils.getSensitiveKeys()`.
- The hand-maintained `DEFAULT_PATTERNS` and `SECRET_KEY_NAME` are removed. Operator-supplied custom patterns (`camel.mcp.security.redaction.patterns`) still apply on top.
- `McpSecurityInterceptor` now routes `Map`/`List` results through `redactStructured` and keeps the masking path for `String`.
- Tests updated to the canonical behaviour, including the previously-missed `JsonObject` with a `datasource.password` key, a nested `client-secret`, a JSON field, and a `?password=` query parameter embedded in a string value.

Redaction stays opt-in and off by default; this only changes what happens once an operator sets `camel.mcp.security.redaction.enabled=true`, so there is no behaviour change for the default configuration.

## Known follow-ups

- Plain record-returning tools (for example `ExampleFileResult.content`) are still not redacted. Doing that type-safely needs either reflective record rebuilding or a framework-level output filter once the Quarkus MCP server exposes global guardrails (the interceptor already carries a TODO for that). This PR deliberately covers the structured `JsonObject` path, which is where live secrets actually surface.
- Connection-string userinfo (`user:pass@host`) and PEM private-key blocks are no longer special-cased, since `SensitiveUtils`/`DefaultMaskingFormatter` key off names rather than value shapes. If those shapes are worth masking, that belongs in Camel's masker / `SensitiveUtils` project-wide rather than in the MCP layer.

## Scope note

Per the [Camel security model](https://github.com/apache/camel/blob/main/docs/user-manual/modules/ROOT/pages/security-model.adoc), the MCP server is a management surface whose network exposure is operator responsibility, so this is hardening of an opt-in control rather than a framework vulnerability fix. A separate, smaller item worth considering is that `camel mcp --http` binds `0.0.0.0` by default rather than loopback; that is not addressed here.

## Testing

```
Tests run: 30, Failures: 0, Errors: 0
  (McpSecretRedactorTest, McpSecurityInterceptorTest, McpSecurityConfigTest, McpAccessFilterTest)
```
Formatted with `mvn formatter:format impsort:sort`.

## Attribution

_Claude Code on behalf of Andrea Cosentino (@oscerd)._ No JIRA was filed for this change; happy to add a `CAMEL-XXXX` reference and rename the branch/commit if you would prefer to track it in JIRA.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
@@ -17,6 +17,8 @@
 package org.apache.camel.dsl.jbang.core.commands.mcp;
 
 import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Map;
 
 import jakarta.annotation.Priority;
 import jakarta.inject.Inject;
@@ -84,11 +86,17 @@ public class McpSecurityInterceptor {
         try {
             Object result = ctx.proceed();
 
-            // Secret redaction on string results
-            if (config.isRedactionEnabled() && result instanceof String s) {
-                if (redactor.containsSecret(s)) {
-                    result = redactor.redact(s);
-                    wasRedacted = true;
+            // Secret redaction. Free-text (String) results go through the regex redactor; structured results
+            // (Map/List, such as the JsonObject returned by the runtime tools) are scrubbed by key name in place,
+            // which is what catches JSON-quoted values the value regex cannot match.
+            if (config.isRedactionEnabled() && result != null) {
+                if (result instanceof String s) {
+                    if (redactor.containsSecret(s)) {
+                        result = redactor.redact(s);
+                        wasRedacted = true;
+                    }
+                } else if (result instanceof Map<?, ?> || result instanceof List<?>) {
+                    wasRedacted = redactor.redactStructured(result);
                 }
             }
 
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
@@ -91,8 +91,11 @@ public class McpSecurityInterceptor {
             // which is what catches JSON-quoted values the value regex cannot match.
             if (config.isRedactionEnabled() && result != null) {
                 if (result instanceof String s) {
-                    if (redactor.containsSecret(s)) {
-                        result = redactor.redact(s);
+                    // redact() is the single source of truth: if it changes the string, a secret was masked.
+                    // Calling containsSecret() first would run the full masking pipeline a second time.
+                    String redacted = redactor.redact(s);
+                    if (!redacted.equals(s)) {
+                        result = redacted;
                         wasRedacted = true;
                     }
                 } else if (result instanceof Map<?, ?> || result instanceof List<?>) {
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecretRedactor.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecretRedactor.java
@@ -17,16 +17,20 @@
 package org.apache.camel.dsl.jbang.core.commands.mcp;
 
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Pattern;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
 /**
- * Regex-based secret redaction engine for MCP tool responses.
+ * Secret redaction engine for MCP tool responses.
  * <p>
- * Applies configurable patterns to detect and replace credentials, tokens, API keys, and connection strings in tool
- * output text.
+ * Two complementary modes are provided. {@link #redact(String)} applies regex patterns to free-text output, catching
+ * unquoted {@code key=value} pairs, connection strings, AWS keys and PEM private-key blocks. {@link #redactStructured}
+ * walks {@link Map}/{@link List} results (such as the {@code JsonObject} returned by the runtime tools) and blanks the
+ * value of any entry whose key name looks like a secret, so JSON-quoted values that the value regex cannot match are
+ * still redacted. Both are needed: the tools return a mix of plain strings and structured {@code JsonObject} trees.
  */
 @ApplicationScoped
 public class McpSecretRedactor {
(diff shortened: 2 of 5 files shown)


davsclaus (changes requested):
do not use a regexp - use the secrets util we have in camel-util - it knows all the different secret words we have

oscerd:
Thanks @davsclaus, good call. Reworked to reuse Camel's own secret knowledge instead of a hand-rolled regex:

- **Structured (`JsonObject`) results** are now redacted by key name with `SensitiveUtils.containsSensitive(...)`, the same helper the jbang `ListProperties` command already uses. This is the important path, since the runtime tools (e.g. `camel_runtime_properties`) return a `JsonObject`.
- **Free-text results** go through `DefaultMaskingFormatter`, which masks key=value, XML and JSON values off `SensitiveUtils.getSensitiveKeys()`.
- Removed the hand-maintained `DEFAULT_PATTERNS` and `SECRET_KEY_NAME`. Operator-supplied custom patterns (`camel.mcp.security.redaction.patterns`) still apply on top.

One consequence worth flagging: the two format-only cases I had (a `mongodb://user:pass@host` userinfo credential and PEM private-key blocks) are no longer special-cased, since `SensitiveUtils`/`DefaultMaskingFormatter` key off names rather than value shapes. The `?password=` query-parameter form is still caught (it is key=value). If the userinfo/PEM shapes are worth covering, that feels like it belongs in `SensitiveUtils`/`DefaultMaskingFormatter` project-wide rather than here; happy to open a separate issue if you agree.

Tests updated to the canonical behaviour (30 green in the module). Still draft while I settle the record-returning-tools follow-up noted in the description.

gnodet (commented):
The PR correctly replaces hand-rolled regex patterns with Camel's own `SensitiveUtils` and `DefaultMaskingFormatter` — the right approach that keeps the MCP layer in sync with the framework's sensitive-key vocabulary automatically. The two-tier redaction strategy (key-name matching for Map entries + free-text masking for String values) provides good coverage.

Three minor findings after verification (one original finding was dismissed — the lazy `maskingFormatter()` init is the well-known "racy single-check" idiom from Effective Java Item 83, which is safe here):

| # | Severity | Finding |
|---|----------|--------|
| 1 | low | String redaction runs the full masking pipeline twice (via `containsSecret` + `redact`) |
| 2 | low | Non-String values under sensitive keys have their type changed to String during redaction |
| 3 | low | Edge-case test gaps for `redactStructured` |

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
```diff
@@ -84,11 +86,17 @@ Object intercept(InvocationContext ctx) throws Exception {
...
-                    result = redactor.redact(s);
-                    wasRedacted = true;
+            // Secret redaction. Free-text (String) results go through the regex redactor; structured results
+            // (Map/List, such as the JsonObject returned by the runtime tools) are scrubbed by key name in place,
+            // which is what catches JSON-quoted values the value regex cannot match.
+            if (config.isRedactionEnabled() && result != null) {
+                if (result instanceof String s) {
+                    if (redactor.containsSecret(s)) {
+                        result = redactor.redact(s);
+                        wasRedacted = true;
```
gnodet, line 99: **Low:** `containsSecret(s)` internally calls `redact(s)` and compares the result. If it returns true, `redact(s)` is called again here — running the full masking pipeline (DefaultMaskingFormatter + custom patterns) twice. Could be simplified:

```suggestion
                    String redacted = redactor.redact(s);
                    if (!redacted.equals(s)) {
                        result = redacted;
```

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
```diff
@@ -84,11 +86,17 @@ Object intercept(InvocationContext ctx) throws Exception {
...
+                    }
+                } else if (result instanceof Map<?, ?> || result instanceof List<?>) {
```
davsclaus, line 101: Minor: `containsSecret(s)` now internally calls `redact(s)` and discards the result, then `redact(s)` is called again here. Could be simplified to:

```suggestion
                if (result instanceof String s) {
                    String scrubbed = redactor.redact(s);
                    if (!scrubbed.equals(s)) {
                        result = scrubbed;
                        wasRedacted = true;
                    }
```

This avoids running the full masking pipeline twice for strings that contain secrets.

davsclaus (commented):
Thanks for this improvement, Andrea — reusing `SensitiveUtils` and `DefaultMaskingFormatter` instead of the hand-maintained regex list is a good direction, and covering the `JsonObject` path that was previously missed is valuable.

A couple of observations:

**1. Double redaction for String results (minor)**

`containsSecret` was rewritten as `return !redact(text).equals(text)`, but the interceptor still calls `containsSecret(s)` followed by `redact(s)` — so the full masking pipeline runs twice for every string that contains a secret (the first result is discarded). The interceptor could avoid this:

```java
String scrubbed = redactor.redact(s);
if (!scrubbed.equals(s)) {
    result = scrubbed;
    wasRedacted = true;
}
```

Not a blocker — the performance cost is negligible for MCP tool responses — but it's an easy cleanup.

**2. No interceptor test for the new Map/List routing**

`McpSecurityInterceptorTest` only tests `stripControlChars()` and `AccessLevel.permits()` — it has no test for `intercept()`. The redactor itself is well-tested, but the interceptor's new routing logic (choosing between String and Map/List paths) is not covered. A test verifying that a `JsonObject` result flows through `redactStructured` would add confidence. This gap is pre-existing for the String path, but the PR extends the untested logic.

Overall this looks solid — the structured redaction approach is clean, tests cover the key new scenarios, and CI is green.

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
oscerd: Done in 8e9decd — the String branch now calls redact(s) once and compares the result to the input to set wasRedacted, so the masking pipeline no longer runs twice. containsSecret() stays as a tested public API on the redactor.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/McpSecurityInterceptor.java
oscerd: Done in 8e9decd — the String branch now calls redact(s) once and compares the result to the input to set wasRedacted, so the masking pipeline no longer runs twice. containsSecret() stays as a tested public API on the redactor.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

gnodet (approved):
All prior reviewer feedback has been addressed. The author correctly replaced the hand-rolled regex with Camel's own `SensitiveUtils` and `DefaultMaskingFormatter`, fixed the double-redaction inefficiency in the interceptor's String branch, and the structured `JsonObject` redaction path fills the security gap this PR targeted. CI is green.

**Prior feedback status:**
- ✅ davsclaus CHANGES_REQUESTED (use SensitiveUtils): addressed in commit e0c2d48
- ✅ gnodet inline (double-redaction fix): addressed in commit 8e9decde
- ✅ davsclaus inline (same double-redaction): addressed in commit 8e9decde

- `McpSecurityInterceptorTest` still has no direct test for `intercept()` — the Map/List routing logic is only transitively tested through the redactor. The routing is straightforward delegation and the redactor itself is well-tested, so this is not a blocker.
- When a sensitive key has a non-String value (e.g., Integer), `redactStructured` replaces it with a REDACTED String, changing the value's type. Acceptable for secret redaction but worth being aware of.

davsclaus (approved):

PR #25094, opened 2026-07-24
CAMEL-24241: Apply the tool argument allowlist to tools declaring no parameters

oscerd:
## The problem

`AiToolExecutor` gates the undeclared-argument filter on the parameter map being non-empty:

```java
if (!argsCopy.isEmpty() && !spec.getParameterDefs().isEmpty()) {
    Set<String> declaredParams = spec.getParameterDefs().keySet();
    argsCopy.keySet().removeIf(name -> { ... });   // filter undeclared
}
```

So a tool that declares **no** parameters skips filtering entirely, and every argument the model sends becomes an exchange header.

A tool declaring no parameters accepts none. `AiToolParameterHelper.buildJsonSchemaFromDefs` emits `"additionalProperties": false` **unconditionally**, so the schema such a tool advertises already forbids every property — which is exactly the rationale the filter's own comment cites.

This also re-introduced a bypass that was removed on purpose. Commit `e9c4541a91ce` ("CAMEL-23621: remove backwards-compatibility bypass") stripped the identical `isEmpty()` guard from three places:

```diff
-    if (!allowedParams.isEmpty() && !allowedParams.contains(name)) {
+    if (!allowedParams.contains(name)) {
```

`LangChain4jToolsProducer` and `SpringAiToolsEndpoint` still carry the strict form; only the newer shared executor regressed it.

## Severity

Hardening rather than a CVE. The Camel-internal prefix rejection (`camel` / `org.apache.camel.`, case-insensitive) still applies, so the CVE-2025-27636 header-injection family is **not** reachable through this path. What is unfiltered is arbitrary **non-Camel** header names originating from model output — and LLM output is attacker-influenceable via prompt injection, while tool routes commonly forward headers to downstream components (HTTP producers, SQL named parameters).

## The fix

Drop the `&& !spec.getParameterDefs().isEmpty()` condition, so an empty declaration means "no arguments accepted" instead of "all arguments accepted".

## Tests

- **New:** `testExecuteIgnoresUndeclaredArgumentsForToolWithoutParameters` — arguments sent to a zero-parameter tool must not reach the exchange headers. Verified it fails without the fix (`expected: null but was: "should-be-ignored"`) and passes with it.
- **Adjusted:** `testCamelPrefixedArgumentsRejectedFromHeaders` used the `noParams` tool and depended on the bypass to get its non-Camel argument through, so the fix broke it. Simply flipping that assertion to `isNull()` would have made the test vacuous — all four arguments would then be dropped by the allowlist, and it would pass even if the Camel-prefix check were deleted. Instead it now points at a new `prefixTool` that **declares** those names, so the Camel-prefixed arguments survive the allowlist and genuinely exercise the prefix check.

  One caveat recorded in a comment: a dotted name such as `org.apache.camel.hack` cannot be declared through the endpoint URI (`parseParameterMetadata` splits the key on the first dot), so that one is dropped as undeclared rather than by the prefix check.

Full `AiToolExecutorTest` suite: 13 tests, green.

## Upgrade guide

Not required — `camel-ai-tool` is new in 4.22 (CAMEL-23382) and the permissive behaviour has never appeared in a release.

---

### DIFF
--- a/components/camel-ai/camel-ai-tool/src/test/java/org/apache/camel/component/ai/tool/AiToolExecutorTest.java
+++ b/components/camel-ai/camel-ai-tool/src/test/java/org/apache/camel/component/ai/tool/AiToolExecutorTest.java
@@ -49,6 +49,16 @@ public class AiToolExecutorTest extends CamelTestSupport {
                      + "&description=A tool with no parameters")
                         .setBody(constant("no-param result"));
 
+                // declares the Camel-prefixed names so they survive the undeclared-argument filter
+                // and actually reach the Camel-prefix rejection this tool is used to test
+                from("ai-tool:prefixTool"
+                     + "?tags=test"
+                     + "&description=A tool declaring Camel-prefixed parameters"
+                     + "&parameter.CamelOverrideEndpoint=string"
+                     + "&parameter.camelFoo=string"
+                     + "&parameter.safeParam=string")
+                        .setBody(constant("prefix result"));
+
                 from("ai-tool:nullBodyTool"
                      + "?tags=test"
                      + "&description=A tool that returns null body")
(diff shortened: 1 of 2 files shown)


components/camel-ai/camel-ai-tool/src/test/java/org/apache/camel/component/ai/tool/AiToolExecutorTest.java
```diff
@@ -117,6 +127,29 @@ public void testExecuteWithNullArguments() {
                 .isEqualTo("no-param result");
     }
 
+    @Test
```
davsclaus, line 130: Minor: per project conventions, new test methods should be package-private (no `public` modifier). Same applies to the modified `testCamelPrefixedArgumentsRejectedFromHeaders` below.

```suggestion
    void testExecuteIgnoresUndeclaredArgumentsForToolWithoutParameters() {
```

davsclaus (approved):
Clean, well-crafted hardening fix. The bypass regression is confirmed against `e9c4541a91ce` (CAMEL-23621), and the fix aligns the shared executor with the sibling producers that already carry the strict form.

One minor convention nit on test visibility (inline below), otherwise LGTM.

gnodet (approved):
Clean, minimal, and well-reasoned security hardening fix. Removes the `!spec.getParameterDefs().isEmpty()` bypass so that zero-parameter tools correctly reject all LLM-hallucinated arguments instead of letting them pass through as exchange headers.

`AiToolSpec.getParameterDefs()` is guaranteed non-null (returns `Map.of()` if constructor argument is null), so no NPE risk. The fix is consistent with the strict enforcement already applied in `LangChain4jToolsProducer` and `SpringAiToolsEndpoint` (commit e9c4541a91c). The `additionalProperties: false` schema emission confirms server-side enforcement is the correct complement.

Test coverage is thorough and the existing test was carefully adjusted. The component is new in 4.22 with no release tags, so no upgrade guide is needed.

PR #25096, opened 2026-07-24
CAMEL-22887: camel-cyberark-vault - add test-infra for automated testing

oscerd:
# CAMEL-22887: camel-cyberark-vault - add test-infra for automated testing

https://issues.apache.org/jira/browse/CAMEL-22887

Adds `camel-test-infra-cyberark-vault` so the CyberArk Conjur integration tests run
automatically instead of requiring a hand-provisioned Conjur instance.

Before this change the ITs were gated on four `@EnabledIfSystemProperty` conditions and
silently skipped in every normal build. They now run: **14 tests, 0 skipped**.

## Why it needs two containers

Conjur cannot run standalone - it requires a PostgreSQL backend. The service therefore starts
both containers on a private `Network` and exposes only Conjur, following the same pattern as
`RocketMQContainerInfraService`.

Conjur also has no admin credentials until an account exists, so the service runs
`conjurctl account create` inside the container on startup and captures the generated API key.
That is retried with Awaitility because Conjur starts accepting connections slightly before it
can serve them.

## Builds on CAMEL-24236

CAMEL-24236 established that a Conjur secret is two separate things: a `variable` declared by
policy, and a value assigned through the Secrets API - the Secrets API cannot create the
variable. Setting a value for an undeclared variable returns `404`. That work already moved the
ITs onto a `CyberArkTestSupport` base class with `loadPolicy` / `createSecret` helpers, so this
PR mostly swaps the system-property plumbing underneath it for the test infra service.

## Changes

- New `camel-test-infra-cyberark-vault` module, registered in `test-infra/pom.xml` and
  `camel-test-infra-all` (dependency + `fileSet`)
- `camel-cyberark-vault` drops the hardcoded failsafe `systemPropertyVariables` and consumes the
  service instead; ITs now read connection details from `service.url()` / `.account()` /
  `.username()` / `.apiKey()`
- Collapsed the repeated four-line vault wiring in the ITs into a `configureVault()` helper
- Rewrote the IT README, which documented a manual setup flow that no longer applies

## Architecture support

The `cyberark/conjur` image is published for `amd64` and `arm64` only - there is no `ppc64le` or
`s390x` variant on any registry - so the ITs are disabled on those via `skipITs.ppc64le` and
`skipITs.s390x`.

## Note on the service alias

The metadata generator merges every module's `container.properties` into a single map and matches
keys by alias suffix. Using `serviceImplementationAlias = "vault"` here collided with
`camel-test-infra-hashicorp-vault` and silently rewrote its `serviceVersion`. This module
therefore uses the alias `conjur` and the prefixes `cyberark.conjur.container` /
`cyberark.conjur.database.container`, with a comment in `container.properties` recording the
constraint. Worth hardening the mojo to flag ambiguous matches rather than silently pick one,
but that is out of scope here.

## Testing

Full IT suite green against real containers:

```
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0 - BUILD SUCCESS
```

Verified via a `docker events` capture that the run really starts
`mirror.gcr.io/cyberark/conjur:1.24.0` and `mirror.gcr.io/postgres:17.10-alpine` (three pairs -
failsafe forks a JVM per test class) and destroys them all afterwards. Full reactor build green
(679/679); `formatter:format` and `impsort:sort` produce no changes.

---


gnodet (approved):
Well-structured test-infra module for CyberArk Conjur that follows established project patterns (service factory, infra service interface, container service, properties constants). The two-container approach (Conjur + PostgreSQL on a private network) mirrors the existing `RocketMQContainerInfraService` pattern.

Container images correctly use `mirror.gcr.io` (verified availability), architecture constraints handled with `skipITs.ppc64le` and `skipITs.s390x`, and the Awaitility-based retry for `conjurctl account create` is appropriate.

The IT simplification (removing `@EnabledIfSystemProperties`, consolidating vault wiring, reducing test visibility to package-private) is a clean improvement.

One minor note: the `shutdown()` method stops both containers but does not close the `Network` — Testcontainers' Ryuk handles this, but `network.close()` would be cleaner hygiene (see `ObservabilityLocalContainerInfraService` as a reference).

davsclaus (approved):
LGTM — nice work bringing the CyberArk ITs onto the test-infra framework. The two-container setup follows the RocketMQ precedent, the service hierarchy matches hashicorp-vault, and the alias collision avoidance is well documented.

Two optional suggestions (non-blocking):

1. **Pre-existing: raw `assert` keyword in `CyberArkVaultProducerIT` line 89** — `assert secretValue.contains("prod-user")` is a Java language assert, not a test assertion. It is silently skipped unless the JVM runs with `-ea`. Since this file is already being touched, consider replacing it with `assertThat(secretValue).contains("prod-user")` (AssertJ).

2. **Minor inconsistency: `public` retained on test methods in `CyberArkVaultProducerIT`** — the other two IT classes correctly drop `public` from all test methods per project convention. The three test methods here (`testRetrieveSecret`, `testRetrieveSecretWithHeader`, `testRetrieveSecretVerifyHeaders`) still carry it because those lines weren't modified. Dropping it would keep things consistent within the PR.

PR #25097, opened 2026-07-24
CAMEL-24257: MongoDB/Cassandra kamelets: Unable to export an integration without running instance

JinyuChen97:
## Summary

Fix regression where camel export fails with MongoTimeoutException / AllNodesFailedException for kamelets using #class: Component beans (e.g. mongodb-sink, cassandra-sink) because the real component is instantiated instead of being replaced with StubComponent.

The root cause: commit 5d4c6cbf0746 (CAMEL-22667) added an outer else { accept = true; } branch in KameletMainInjector.acceptComponent(). However, this branch also impact when stubPattern is "*" (used by camel export), overriding accept=false back to true for components not in ACCEPTED_STUB_NAMES, causing real components like MongoDbComponent / CassandraComponent to be loaded and attempt connections to non-existent instances.

## Changes

KameletMainInjector.java — Removed the outer else { accept = true; } branch in acceptComponent(). When stubPattern is "*", accept now remains as returned by accept(type): true for ACCEPTED_STUB_NAMES components, false for everything else, which can fix the loading non existed instance behavior with camel export.

## Test plan

- New KameletMainInjectorTest verifies that non-whitelisted components are stubbed with stubPattern=* and "component:*", whitelisted components are allowed through, and non-Component classes pass unchanged

_Claude Code on behalf of Jinyu Chen_

### DIFF
--- a/dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KameletMainInjectorTest.java
+++ b/dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KameletMainInjectorTest.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.main.download;
+
+import org.apache.camel.component.direct.DirectComponent;
+import org.apache.camel.component.stub.StubComponent;
+import org.apache.camel.component.timer.TimerComponent;
+import org.apache.camel.impl.engine.SimpleCamelContext;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+
+class KameletMainInjectorTest {
+
+    @Test
+    void testWildcardStubsNonWhitelistedComponent() {
+        // TimerComponent is NOT in ACCEPTED_STUB_NAMES, stubPattern="*" (used by camel export), it should be replaced with StubComponent.
+        KameletMainInjector injector = new KameletMainInjector(
+                new SimpleCamelContext().getInjector(), "*", true);
+
+        Object result = injector.newInstance(TimerComponent.class);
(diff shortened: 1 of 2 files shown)


dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KameletMainInjectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.apache.camel.component.direct.DirectComponent;
+import org.apache.camel.component.stub.StubComponent;
+import org.apache.camel.component.timer.TimerComponent;
+import org.apache.camel.impl.engine.SimpleCamelContext;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+
+public class KameletMainInjectorTest {
```
davsclaus: Minor: per project conventions, JUnit 5 test classes should be package-private (drop `public`).
```suggestion
class KameletMainInjectorTest {
```

dsl/camel-kamelet-main/src/test/java/org/apache/camel/main/download/KameletMainInjectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    public void testWildcardStubsNonWhitelistedComponent() {
```
davsclaus: Minor: test methods should also be package-private.
```suggestion
    void testWildcardStubsNonWhitelistedComponent() {
```

davsclaus (approved):
Nice catch on the regression from CAMEL-22667! The fix is correct and minimal — the outer `else { accept = true; }` was overriding the stub decision for wildcard patterns (`*` and `component:*`), causing real components to be instantiated during `camel export`.

The test coverage is good, covering the key scenarios (wildcard stubbing, whitelisted pass-through, `component:*` variant, and non-Component types).

A few minor convention notes for future PRs:
- Test classes and methods should be package-private (no `public` modifier) per JUnit 5 conventions used in this project
- The project prefers AssertJ assertions (`assertThat(...).isInstanceOf(...)`) over JUnit ones, though this module's existing tests also use JUnit style
- Commits should reference a JIRA ticket (`CAMEL-XXXX: description`) when applicable

gnodet (commented):
The two-line fix correctly identifies and removes the regression from CAMEL-22667 where the outer `else { accept = true; }` overrode the stub decision for wildcard patterns (`*` and `component:*`), causing real components like MongoDbComponent/CassandraComponent to be instantiated during `camel export` instead of being replaced with StubComponent. The fix is minimal and correct, with good test coverage for the wildcard scenarios.

Two observations:

| # | Severity | Finding |
|---|----------|--------|
| 1 | medium | No JIRA ticket linked — per project convention, bug fixes should reference a `CAMEL-XXXX` ticket |
| 2 | low | Pre-existing: `ObjectHelper.getAnnotation(this, Component.class)` always returns null because it looks up `@Component` on `KameletMainInjector` (which has none) instead of the component `type` parameter — making the annotation-matching code (lines 88-101) dead code. Not introduced by this PR. |

Neither finding blocks the fix itself — the code change is correct and well-tested.

davsclaus:
I created a JIRA: https://issues.apache.org/jira/browse/CAMEL-24257

davsclaus:
yes

PR #25108, opened 2026-07-25
CAMEL-24215: camel-resilience4j - JMX waitDurationInOpenState now returns millis

davsclaus:

## Summary

- Fix JMX managed attribute `CircuitBreakerWaitDurationInOpenState` to return milliseconds (`.toMillis()`) instead of seconds (`.getSeconds()`), consistent with sibling attributes `BulkheadMaxWaitDuration` and `TimeoutDuration`
- Add upgrade guide entry documenting the JMX contract change

This is a follow-up from CAMEL-24137 which converted `waitDurationInOpenState` from seconds to milliseconds but missed updating the JMX getter.

## Test plan

- [x] Verified all three duration-returning JMX attributes now consistently use `.toMillis()`
- [x] Checked dev console — it doesn't expose configuration values, only runtime metrics
- [x] Module builds successfully

### DIFF
--- a/components/camel-resilience4j/src/main/java/org/apache/camel/component/resilience4j/ResilienceProcessor.java
+++ b/components/camel-resilience4j/src/main/java/org/apache/camel/component/resilience4j/ResilienceProcessor.java
@@ -469,7 +469,7 @@ public class ResilienceProcessor extends BaseProcessorSupport
 
     @ManagedAttribute
     public long getCircuitBreakerWaitDurationInOpenState() {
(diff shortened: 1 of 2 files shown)


gnodet (approved):

Clean fix. The JMX getter was returning seconds while all sibling duration attributes (`BulkheadMaxWaitDuration`, `TimeoutDuration`) return milliseconds — and the underlying config option was already changed to millis in CAMEL-24137. The upgrade guide entry is accurate and helpful.

LGTM ✅

PR #25113, opened 2026-07-25
CAMEL-24256: Mask URI userinfo credentials and PEM private keys

atiaomar1978-hub:
## Summary

- Extends `SensitiveUtils` with value-shape helpers for URI userinfo passwords and PEM private-key bodies (with fast `contains` pre-checks before regex).
- `DefaultMaskingFormatter` applies these after name-based masking, including when the keyword set is empty.
- Non-greedy URI userinfo matching aligns with `URISupport` for passwords containing `:`; pattern is named `URI_USERINFO_PASSWORD_IN_TEXT` to distinguish from `URISupport.USERINFO_PASSWORD`.
- Docs + AssertJ tests (colon-in-password, lowercase PEM, plain-text fast path, empty keywords).

## Context

Follow-up to MCP redaction rework (#25093). Addresses review on PR (fast paths, naming, empty-keyword contract).

## Test plan

- [x] `SensitiveUtilsTest` / `DefaultMaskingFormatterTest` (local Maven, Java 21)
- [ ] CI

_AI-generated PR description on behalf of Omar Atie (@atiaomar1978-hub)._

### DIFF
--- a/core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
+++ b/core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
@@ -31,11 +31,15 @@ public final class SensitiveUtils {
      * captures the password as group 2. The scheme may contain {@code +} / {@code .} / {@code -} (e.g.
      * {@code mongodb+srv}). Does not match user-only userinfo without a password ({@code scheme://user@host}).
      * <p>
-     * Passwords must not contain a raw {@code @} (use percent-encoding such as {@code %40}); capture stops at the
-     * first {@code @} that ends userinfo.
+     * Stricter than {@link URISupport}'s {@code USERINFO_PASSWORD} (used in {@link URISupport#sanitizeUri}): requires a
+     * scheme and a colon before the password. Used for free-text log masking. The user/password prefix uses a
+     * non-greedy match so passwords may contain {@code :} (aligned with {@link URISupport#sanitizeUri}).
+     * <p>
+     * Passwords must not contain a raw {@code @} (use percent-encoding such as {@code %40}); capture stops at the first
+     * {@code @} that ends userinfo.
      */
-    private static final Pattern USERINFO_PASSWORD
-            = Pattern.compile("([a-zA-Z][a-zA-Z0-9+.-]*://[^/@\\s\"']*:)([^@\\s\"']+)(@)");
+    private static final Pattern URI_USERINFO_PASSWORD_IN_TEXT
+            = Pattern.compile("([a-zA-Z][a-zA-Z0-9+.-]*://[^/@\\s\"']*?:)([^@\\s\"']+)(@)");
 
     /**
      * Matches PEM private-key blocks ({@code -----BEGIN ... PRIVATE KEY-----} … {@code -----END ... PRIVATE KEY-----})
@@ -293,7 +297,11 @@ public final class SensitiveUtils {
         if (source == null || source.isEmpty() || mask == null) {
             return source;
         }
-        return USERINFO_PASSWORD.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
+        if (!source.contains("://")) {
+            return source;
+        }
+        return URI_USERINFO_PASSWORD_IN_TEXT.matcher(source)
+                .replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
     }
 
     /**
@@ -307,6 +315,9 @@ public final class SensitiveUtils {
      */
     public static String maskPemPrivateKeyBlocks(String source, String mask) {
         if (source == null || source.isEmpty() || mask == null) {
+            return source;
+        }
+        if (!StringHelper.containsIgnoreCase(source, "-----BEGIN")) {
             return source;
         }
         return PEM_PRIVATE_KEY.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
--- a/core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
+++ b/core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
@@ -21,8 +21,29 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class SensitiveUtils {
+
+    /**
+     * Matches URI userinfo credentials ({@code scheme://user:password@host} or {@code scheme://:password@host}) and
+     * captures the password as group 2. The scheme may contain {@code +} / {@code .} / {@code -} (e.g.
+     * {@code mongodb+srv}). Does not match user-only userinfo without a password ({@code scheme://user@host}).
+     * <p>
+     * Passwords must not contain a raw {@code @} (use percent-encoding such as {@code %40}); capture stops at the
+     * first {@code @} that ends userinfo.
+     */
+    private static final Pattern USERINFO_PASSWORD
+            = Pattern.compile("([a-zA-Z][a-zA-Z0-9+.-]*://[^/@\\s\"']*:)([^@\\s\"']+)(@)");
+
+    /**
+     * Matches PEM private-key blocks ({@code -----BEGIN ... PRIVATE KEY-----} … {@code -----END ... PRIVATE KEY-----})
+     * and captures the key body as group 2. Public keys and certificates are not matched.
+     */
+    private static final Pattern PEM_PRIVATE_KEY = Pattern.compile(
+            "(-----BEGIN [^-]*PRIVATE KEY-----\\s*)(.*?)(\\s*-----END [^-]*PRIVATE KEY-----)",
+            Pattern.DOTALL | Pattern.CASE_INSENSITIVE);
 
     private static final Set<String> SENSITIVE_KEYS = Collections.unmodifiableSet(new HashSet<>(
             Arrays.asList(
@@ -254,6 +275,56 @@ public final class SensitiveUtils {
         text = text.toLowerCase(Locale.ENGLISH);
         text = text.replace("-", "");
         return SENSITIVE_KEYS.contains(text);
+    }
+
+    /**
+     * Masks passwords embedded in URI userinfo ({@code scheme://user:password@host}) within free text. Complements
+     * name-based secret detection: the password has no {@code password=} key, so key/value maskers never see it.
+     * <p>
+     * Query-parameter forms such as {@code ?password=secret} are not handled here; use a key/value masker or
+     * {@link URISupport#sanitizeUri(String)} for those.
+     *
+     * @param  source the text that may contain connection-string style URIs
+     * @param  mask   the replacement string for the password (for example {@code xxxxx})
+     * @return        the source with userinfo passwords replaced, or the original source when null/empty or when mask
+     *                is null
+     */
+    public static String maskUserInfoCredentials(String source, String mask) {
+        if (source == null || source.isEmpty() || mask == null) {
+            return source;
+        }
+        return USERINFO_PASSWORD.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
+    }
+
+    /**
+     * Masks the body of PEM private-key blocks within free text, keeping the BEGIN/END markers. Public keys and
+     * certificates ({@code BEGIN PUBLIC KEY}, {@code BEGIN CERTIFICATE}) are left unchanged.
+     *
+     * @param  source the text that may contain PEM private-key material
+     * @param  mask   the replacement string for the key body (for example {@code xxxxx})
+     * @return        the source with private-key bodies replaced, or the original source when null/empty or when mask
+     *                is null
+     */
+    public static String maskPemPrivateKeyBlocks(String source, String mask) {
+        if (source == null || source.isEmpty() || mask == null) {
+            return source;
+        }
+        return PEM_PRIVATE_KEY.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
+    }
+
+    /**
+     * Applies value-shape secret masking (URI userinfo passwords and PEM private-key bodies) that cannot be detected by
+     * sensitive key names alone.
+     *
+     * @param  source the text to mask
+     * @param  mask   the replacement string
+     * @return        the masked text, or the original source when null/empty or when mask is null
+     * @see           #maskUserInfoCredentials(String, String)
+     * @see           #maskPemPrivateKeyBlocks(String, String)
+     */
+    public static String maskSensitiveValueShapes(String source, String mask) {
+        String answer = maskUserInfoCredentials(source, mask);
+        return maskPemPrivateKeyBlocks(answer, mask);
     }
 
 }
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/docs/log-eip.adoc
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/docs/log-eip.adoc
@@ -502,6 +502,10 @@ The know set of keywords to mask is gathered from all the different component op
 The list is generated into the source code in `org.apache.camel.util.SensitiveUtils`.
 At this time of writing, there are more than 65 different keywords.
 
+In addition to name-based masking, `DefaultMaskingFormatter` also masks two value shapes that have no
+sensitive key name: URI userinfo passwords (`scheme://user:password@host`) and PEM private-key blocks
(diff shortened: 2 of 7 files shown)


core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
```diff
@@ -256,4 +277,54 @@ public static boolean containsSensitive(String text) {
...
+     * name-based secret detection: the password has no {@code password=} key, so key/value maskers never see it.
+     * <p>
+     * Query-parameter forms such as {@code ?password=secret} are not handled here; use a key/value masker or
+     * {@link URISupport#sanitizeUri(String)} for those.
+     *
+     * @param  source the text that may contain connection-string style URIs
+     * @param  mask   the replacement string for the password (for example {@code xxxxx})
+     * @return        the source with userinfo passwords replaced, or the original source when null/empty or when mask
+     *                is null
+     */
```
davsclaus, line 295: This should have a fast `String.contains("://")` guard before invoking the regex — the old masking code deliberately avoided regex overhead for the common case (most log messages don't contain URI schemes):

```suggestion
    public static String maskUserInfoCredentials(String source, String mask) {
        if (source == null || source.isEmpty() || mask == null) {
            return source;
        }
        if (!source.contains("://")) {
            return source;
        }
        return USERINFO_PASSWORD.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
    }
```

core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
```diff
@@ -256,4 +277,54 @@ public static boolean containsSensitive(String text) {
...
+    }
+
+    /**
+     * Masks the body of PEM private-key blocks within free text, keeping the BEGIN/END markers. Public keys and
+     * certificates ({@code BEGIN PUBLIC KEY}, {@code BEGIN CERTIFICATE}) are left unchanged.
+     *
+     * @param  source the text that may contain PEM private-key material
+     * @param  mask   the replacement string for the key body (for example {@code xxxxx})
+     * @return        the source with private-key bodies replaced, or the original source when null/empty or when mask
+     *                is null
```
davsclaus, line 314: Same here — add a fast `String.contains("-----BEGIN")` guard before the regex:

```suggestion
    public static String maskPemPrivateKeyBlocks(String source, String mask) {
        if (source == null || source.isEmpty() || mask == null) {
            return source;
        }
        if (!source.contains("-----BEGIN")) {
            return source;
        }
        return PEM_PRIVATE_KEY.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
    }
```

core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
```diff
@@ -21,9 +21,30 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class SensitiveUtils {
 
+    /**
+     * Matches URI userinfo credentials ({@code scheme://user:password@host} or {@code scheme://:password@host}) and
+     * captures the password as group 2. The scheme may contain {@code +} / {@code .} / {@code -} (e.g.
+     * {@code mongodb+srv}). Does not match user-only userinfo without a password ({@code scheme://user@host}).
+     * <p>
```
davsclaus, line 33: `URISupport.java` (line 67) already has a field named `USERINFO_PASSWORD` with a different regex. Having two static patterns with the same name in different classes doing subtly different things will confuse future maintainers. Consider renaming this one (e.g. `USERINFO_PASSWORD_IN_TEXT`) or reconciling the two.

davsclaus (changes requested):
Thanks for working on this — masking URI userinfo passwords and PEM private-key blocks is a valuable addition that closes a real gap in name-based masking.

A few things to address before this can go in:

### 1. Missing fast pre-checks before regex

The existing masking code was deliberately designed to avoid regex overhead when it isn't needed — `SensitiveUtils.containsSensitive()` uses `Set.contains()` (O(1) hash lookup, no regex), and `DefaultMaskingFormatter.format()` uses fast `startsWith("<")` / `startsWith("{")` guards before touching any pattern.

The new `maskSensitiveValueShapes()` always runs two `Pattern.replaceAll()` calls on every invocation of `format()`, even when the text clearly can't match (e.g. a plain `"Hello World"` log message). The vast majority of log messages don't contain `://` or `-----BEGIN`, so a simple `String.contains()` guard would avoid the regex engine entirely for the common case:

```java
public static String maskUserInfoCredentials(String source, String mask) {
    if (source == null || source.isEmpty() || mask == null) {
        return source;
    }
    if (!source.contains("://")) {
        return source;  // fast path: no URI scheme, skip regex
    }
    return USERINFO_PASSWORD.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
}

public static String maskPemPrivateKeyBlocks(String source, String mask) {
    if (source == null || source.isEmpty() || mask == null) {
        return source;
    }
    if (!source.contains("-----BEGIN")) {
        return source;  // fast path: no PEM marker, skip regex
    }
    return PEM_PRIVATE_KEY.matcher(source).replaceAll("$1" + Matcher.quoteReplacement(mask) + "$3");
}
```

### 2. Duplicate `USERINFO_PASSWORD` pattern

`URISupport.java` already has a static `USERINFO_PASSWORD` pattern (line 67). This PR adds a second one in `SensitiveUtils` with a different regex but the same field name. Same name, different behavior, different class — this will confuse maintainers. At minimum the name should be differentiated, or the patterns reconciled.

### 3. Behavioral change: empty-keyword formatter is no longer a no-op

The old `format()` returned immediately when `keywords == null || keywords.isEmpty()`. The PR always runs `maskSensitiveValueShapes()` even with empty keywords. This is probably the correct behavior (you'd still want to mask embedded passwords), but it changes the old contract silently — worth acknowledging.

Tests and docs look solid. Good coverage for URI schemes, PEM variants, edge cases, and custom mask strings. The AssertJ migration and upgrade guide entry are well done.

_This review was generated by an AI agent on behalf of davsclaus and may contain inaccuracies. Please verify all suggestions before applying._

atiaomar1978-hub:
Thanks for the review, @davsclaus — the latest push (`77607275443`) addresses your feedback and follow-ups from automated review (Bugbot + Grok). _AI-generated comment on behalf of @atiaomar1978-hub._

---

## @davsclaus review

### 1. Fast pre-checks before regex

`SensitiveUtils.maskUserInfoCredentials`, `maskPemPrivateKeyBlocks`, and the call path from `DefaultMaskingFormatter.format()` now skip the regex engine when markers are absent:

- URI userinfo: return early when the text does not contain `://`
- PEM private keys: return early when the text does not contain `-----BEGIN` (case-insensitive via `StringHelper.containsIgnoreCase`, consistent with the `CASE_INSENSITIVE` PEM pattern)

Plain log lines such as `Hello World` no longer pay regex cost. Covered by `maskValueShapeHelpersSkipRegexWhenMarkersAbsent` and `formatPlainTextUnchanged`.

### 2. Duplicate `USERINFO_PASSWORD` pattern name

The log-masking pattern in `SensitiveUtils` is renamed to **`URI_USERINFO_PASSWORD_IN_TEXT`**, with Javadoc explaining how it differs from `URISupport`'s `USERINFO_PASSWORD` used in `sanitizeUri()` (stricter scheme/userinfo rules for free text vs. URI sanitization).

The regex prefix is **non-greedy** (`*?:`) so passwords containing `:` are fully masked (e.g. `http://user:pass:word@host` → password replaced in one shot), aligned with `URISupport`'s non-greedy userinfo handling.

### 3. Empty-keyword formatter behavior

Previously, `format()` with an empty keyword set skipped all masking. Value-shape masking (URI userinfo + PEM) now **always** runs after name-based passes, which is intentional so embedded credentials in connection strings are still redacted when no sensitive key names appear.

Documented in:

- `DefaultMaskingFormatter` class Javadoc
- Camel 4.22 upgrade guide (`camel-support - DefaultMaskingFormatter masks URI userinfo and PEM private keys`)

Test: `formatMasksValueShapesWhenKeywordsEmpty` (URI/PEM still masked; bare `password=visible` unchanged without keywords).

---

## Bugbot

| Pass | Finding | Resolution |
|------|---------|------------|
| Pre-fix | PEM fast path was case-sensitive while the regex is `CASE_INSENSITIVE` → lowercase `-----begin ...` could skip masking | PEM pre-check uses `StringHelper.containsIgnoreCase(source, "-----BEGIN")`; test added for lowercase PEM headers |
| Post-fix (branch vs `main`) | — | **No bugs reported** |

---

## Grok (second review)

| Severity | Topic | Resolution |
|----------|--------|------------|
| Medium | Colon in password only partly masked (greedy userinfo prefix) | Non-greedy `[^/@\\s\"']*?:` + test `http://user:pass:word@host` |
| Medium | PEM fast path vs case-insensitive regex mismatch | Same as Bugbot fix above |
| Low | Orphaned duplicate Javadoc on the pattern field | Merged into a single block |
| Low | Missing tests for colon-in-password / lowercase PEM | Added in `SensitiveUtilsTest` |

Items already in good shape per Grok: empty-username Redis (`redis://:pass@`), `mailto:`, user-only userinfo, naming/docs distinction from `URISupport`, empty-keywords + upgrade guide.

---

## Tests (local)

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@21 \
  ./mvnw -pl core/camel-util,core/camel-core -am test \
  -Dtest=SensitiveUtilsTest,DefaultMaskingFormatterTest
```

All tests passed locally; awaiting CI.

---

Happy to adjust further if anything should stay a strict no-op (e.g. empty keywords) or if we should share/reconcile regex with `URISupport` more directly in a follow-up.

core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
atiaomar1978-hub: Fixed in `77607275443`. Added the `://` fast path exactly as suggested (`maskUserInfoCredentials` returns early when the source has no URI scheme). Tests: `maskValueShapeHelpersSkipRegexWhenMarkersAbsent` and `formatPlainTextUnchanged`.

_AI-generated reply on behalf of @atiaomar1978-hub._

core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
atiaomar1978-hub: Fixed in `77607275443`. Added a fast path before the PEM regex. The guard uses `StringHelper.containsIgnoreCase(source, \"-----BEGIN\")` so it stays consistent with the `CASE_INSENSITIVE` PEM pattern (Bugbot/Grok flagged plain `contains` missing lowercase headers). Test added for lowercase `-----begin rsa private key-----`.

_AI-generated reply on behalf of @atiaomar1978-hub._

core/camel-util/src/main/java/org/apache/camel/util/SensitiveUtils.java
atiaomar1978-hub: Fixed in `77607275443`. Renamed to `URI_USERINFO_PASSWORD_IN_TEXT` and documented how it differs from `URISupport.USERINFO_PASSWORD` (stricter free-text rules). Also made the userinfo prefix non-greedy (`*?:`) so passwords with embedded `:` mask fully, aligned with `URISupport` sanitization.

_AI-generated reply on behalf of @atiaomar1978-hub._

atiaomar1978-hub:
@davsclaus — re **review §3 (empty-keyword formatter no longer a no-op)**, also addressed in `77607275443`:

- `DefaultMaskingFormatter.format()` still skips name-based masking when `keywords` is empty, but **always** runs `SensitiveUtils.maskSensitiveValueShapes()` afterward so embedded URI/PEM secrets are redacted.
- Documented in `DefaultMaskingFormatter` Javadoc and the Camel 4.22 upgrade guide (`camel-support - DefaultMaskingFormatter masks URI userinfo and PEM private keys`).
- Test: `formatMasksValueShapesWhenKeywordsEmpty` (URI/PEM masked; bare `password=visible` unchanged without keywords).

That covers all three items from the review summary plus the three inline threads. Happy to re-request your review when CI is green.

_AI-generated reply on behalf of @atiaomar1978-hub._

davsclaus (approved):
All three items from the previous review are addressed:

1. **Fast pre-checks** — `maskUserInfoCredentials` guards with `!source.contains("://")`, `maskPemPrivateKeyBlocks` guards with `!StringHelper.containsIgnoreCase(source, "-----BEGIN")`. The case-insensitive PEM guard is a nice improvement — stays consistent with the `CASE_INSENSITIVE` regex flag and is covered by the lowercase PEM test.
2. **Pattern rename** — `USERINFO_PASSWORD` → `URI_USERINFO_PASSWORD_IN_TEXT` with Javadoc explaining how it differs from `URISupport.USERINFO_PASSWORD`.
3. **Empty-keyword contract** — documented in the upgrade guide and tested (`formatMasksValueShapesWhenKeywordsEmpty`).

Also good to see the null-safety fix in `format()` (old code would NPE on `null` source when keywords were non-empty), the `Matcher.quoteReplacement(mask)` for special chars, and the thorough test coverage.

Minor nit (non-blocking): `DefaultMaskingFormatterTest` now imports both AssertJ and JUnit assertions — existing methods still use `assertEquals` while new ones use `assertThat`. Since only the `public` modifier was removed from existing methods (bodies unchanged), this is borderline per project conventions.

Note: this review covers project rules and conventions only. It does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.

_This review was generated by an AI agent on behalf of davsclaus and may contain inaccuracies. Please verify all suggestions before applying._

gnodet (approved):
Well-implemented security enhancement that adds URI userinfo password and PEM private key body masking to `DefaultMaskingFormatter`. The regex patterns are well-designed (no ReDoS risk, proper anchoring), tests are comprehensive, and the upgrade guide properly documents the behavioral change.

**Highlights:**
- Fast-path guards (`contains("://")` for URI, `containsIgnoreCase("-----BEGIN")` for PEM) avoid regex overhead on typical log lines
- The null/empty source check at the top of `format()` is a good defensive fix for a pre-existing potential NPE
- Tests cover common schemes, edge cases (colon-in-password, percent-encoded @), PEM variants (RSA, EC, ENCRYPTED), and negative cases

**Minor pre-existing observations (non-blocking):**
- The existing keyword-based masking paths use raw `maskString` in regex replacements without `Matcher.quoteReplacement()`, while the new `SensitiveUtils` methods correctly use it. If `maskString` ever contains `$` or `\`, keyword-based masking would produce incorrect output. Worth a follow-up fix for consistency.
- Pre-existing typo in comment: "key=value paris" should be "key=value pairs" (line ~135 in `DefaultMaskingFormatter.java`)

atiaomar1978-hub:
## Review follow-up on `CAMEL-24256-mask-userinfo-pem`

### Already addressed (earlier `77607275443`, @davsclaus blocking items)

- Fast `://` / case-insensitive `-----BEGIN` guards in `SensitiveUtils`
- `URI_USERINFO_PASSWORD_IN_TEXT` rename + Javadoc vs `URISupport`
- Empty-keyword value-shape masking documented + tested

### New in `ae028bd8cb3` (minor / @gnodet + @davsclaus nits)

- **`DefaultMaskingFormatter`**: keyword/XML/JSON/key-value `replaceAll` paths now use `Matcher.quoteReplacement(maskString)` (same as value-shape helpers) so `$` / `\` in custom masks are literal
- Comment typo: **key=value pairs** (was “paris”)
- **`DefaultMaskingFormatterTest`**: AssertJ only in this class (no mixed JUnit `assertEquals`)
- **`formatKeywordMaskingUsesLiteralMaskStringWithRegexMetacharacters`**: regression for `maskString="$1"`

Ready for CI. Thanks @davsclaus @gnodet.

_AI-generated comment on behalf of @atiaomar1978-hub._

gnodet (approved):
Re-reviewed after new commits. Both observations from the previous review are fully addressed:

1. ✅ **Typo fixed**: "key=value paris" → "key=value pairs"
2. ✅ **`Matcher.quoteReplacement()` consistency**: now applied in all keyword-based masking paths, not just the new `SensitiveUtils` methods. The new test `formatKeywordMaskingUsesLiteralMaskStringWithRegexMetacharacters` directly proves this fix works.

Additional improvements in the new commits:
- Pattern renamed from `USERINFO_PASSWORD` to `URI_USERINFO_PASSWORD_IN_TEXT` for clarity
- Fast pre-checks (`!source.contains("://")`, `!containsIgnoreCase("-----BEGIN")`) before regex
- Test migration to AssertJ and package-private visibility per project conventions

Looks good to merge.

PR #25153, opened 2026-07-27
CAMEL-24263: camel-aws2-sts - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `STS2Producer` expects the body to be the AWS SDK
request for the operation. Each branch has the shape:

```java
if (getConfiguration().isPojoRequest()) {
    Object payload = exchange.getIn().getMandatoryBody();
    if (payload instanceof AssumeRoleRequest request) {
        ... call AWS, set response ...
    }
    // no else: a wrong-typed body falls straight through
}
```

If the body is not the expected type, the `instanceof` is false, the branch is
skipped, and the method returns without calling STS or setting a response — the
caller silently gets the original body back and no error. Reproduced for all
three operations (`assumeRole`, `getSessionToken`, `getFederationToken`).

## Fix

Add the missing `else` that throws `IllegalArgumentException` naming the required
request type, exactly as CAMEL-23462 did for `camel-aws-bedrock`:

```java
} else {
            "assumeRole operation requires AssumeRoleRequest in POJO mode");
}
```

## Tests

Three tests in `STS2ProducerTest` send a wrong-typed body to a `pojoRequest=true`
route and assert the `IllegalArgumentException` message. Each was verified to
fail (silent no-op) before the fix. Full class 7/7 green.

## Docs

Behaviour change → a single 4.22 upgrade-guide entry (`=== camel-aws2 -
producers now fail fast on a wrong POJO request type`) covering the whole
CAMEL-24261 roll-out; subsequent per-component PRs won't re-add it.

## Backport

Main only, matching CAMEL-23462 (which shipped in 4.21.0 and was not backported)
— this is a behaviour change, not a crash fix.

---

### DIFF
--- a/components/camel-aws/camel-aws2-sts/src/main/java/org/apache/camel/component/aws2/sts/STS2Producer.java
+++ b/components/camel-aws/camel-aws2-sts/src/main/java/org/apache/camel/component/aws2/sts/STS2Producer.java
@@ -103,6 +103,9 @@ public class STS2Producer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
+            } else {
+                throw new IllegalArgumentException(
+                        "assumeRole operation requires AssumeRoleRequest in POJO mode");
             }
         } else {
             Builder builder = AssumeRoleRequest.builder();
@@ -157,6 +160,9 @@ public class STS2Producer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
+            } else {
+                throw new IllegalArgumentException(
+                        "getSessionToken operation requires GetSessionTokenRequest in POJO mode");
             }
         } else {
             GetSessionTokenRequest.Builder builder = GetSessionTokenRequest.builder();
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

gnodet (approved):
Clean, well-tested, well-documented change that adds fail-fast behavior for wrong-typed POJO requests in all three STS operations. Consistent with the established pattern from CAMEL-23462 (camel-aws-bedrock). CI passes.

**Highlights:**
- Error message format (`"<operation> operation requires <RequestType> in POJO mode"`) is consistent with the bedrock precedent
- All three operations covered: `assumeRole`, `getSessionToken`, `getFederationToken`
- Tests correctly use AssertJ (`assertThatThrownBy`) with `hasRootCauseInstanceOf` and `hasRootCauseMessage` — per project conventions
- Upgrade guide entry is well-written: explains the behavior change, references the precedent, links to CAMEL-24261, and provides actionable migration guidance

- Per project convention, the existing `public class STS2ProducerTest` could drop the `public` modifier since the file is being modified. New test methods already use package-private visibility correctly. Very low priority — fits the "apply incrementally" guidance.

PR #25158, opened 2026-07-27
CAMEL-24264: camel-aws2-translate - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Translate2Producer.translateText` only acted when the
body was a `TranslateTextRequest`; any other body fell through the `instanceof`

## Fix

```java
} else {
            "translateText operation requires TranslateTextRequest in POJO mode");
}
```

## Test

`translateTextWithPojoRequestAndWrongBodyTypeThrows` sends a wrong-typed body to
the existing `pojoRequest=true` route and asserts the message. Verified to fail
(silent no-op) before the fix. Class 4/4 green.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-translate/pom.xml
+++ b/components/camel-aws/camel-aws2-translate/pom.xml
@@ -78,5 +78,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

gnodet (approved):
Clean, minimal fix that follows the established CAMEL-23462 pattern for rejecting wrong-type bodies in POJO mode. The fix, test, and dependency addition are all correct.

**Verified details:**
- The fix correctly adds an `else` clause throwing `IllegalArgumentException` when `pojoRequest=true` but the body is not a `TranslateTextRequest` — structurally identical to the pattern established in CAMEL-23462 for `BedrockProducer`.
- The test correctly uses `hasRootCauseInstanceOf(IllegalArgumentException.class)` to unwrap through the `CamelExecutionException` wrapper, and reuses the existing `direct:translatePojoText` route.
- The `assertj-core` test dependency is justified and correctly omits the version (managed by parent BOM).
- No mixing of AssertJ and JUnit assertions within the same test method.
- Git history confirms the silent fall-through was original code with no prior intentional changes.

PR #25164, opened 2026-07-27
CAMEL-24272: Migrate reconnection loops from ForegroundTask to BackgroundTask

gnodet:
## Summary

Migrate reconnection loops across 9 components from `ForegroundTask` to `BackgroundTask` so they register with `TaskManagerRegistry` and become visible in the TUI/CLI/dev console internal task view.

**Note:** The Kafka component (`KafkaFetchRecords.java`) was already independently migrated to `BackgroundTask` upstream, so this PR now covers the remaining 9 components.

### Components migrated

| Component | File | Pattern |
|-----------|------|---------|
| FTP | `FtpOperations.java` | Multi-iteration retry (connect) |
| SFTP | `SftpOperations.java` | Multi-iteration retry (connect) |
| MLLP | `TcpServerBindThread.java` | Multi-iteration retry (accept) |
| Infinispan | `InfinispanRemoteManager.java` | Multi-iteration retry (schema registration) |
| Google Pub/Sub | `GooglePubsubConsumer.java` | Single-iteration delay (reconnect backoff) |
| ZooKeeper | `ZooKeeperConsumer.java` | Single-iteration delay (reconnect backoff) |
| Hazelcast SEDA | `HazelcastSedaConsumer.java` | Single-iteration delay (error recovery) |
| Salesforce | `SubscriptionHelper.java` | Single-iteration delay (handshake + subscribe retry) |
| MongoDB GridFS | `GridFsConsumer.java` | Unlimited polling loop |

### Key changes

- **`ForegroundTask` → `BackgroundTask`**: Each migrated task now registers with `TaskManagerRegistry`, making internal retry/reconnect tasks visible through management interfaces
- **`IterationBudget` → `IterationTimeBudget`**: `BackgroundTask` requires a `TimeBudget`; used `withUnlimitedDuration()` to preserve existing behavior
- **`ScheduledExecutorService`**: Each component gets a dedicated scheduled executor for task scheduling, managed through Camel's `ExecutorServiceManager` for proper lifecycle
- **Single-iteration delay interval**: For tasks using `maxIterations(1)` as a delay mechanism, interval is set to `Duration.ofMillis(1)` since `ScheduledExecutorService.scheduleWithFixedDelay()` rejects `delay=0`

## Test plan

- [x] Existing tests pass for all 9 affected component modules
- [x] `HazelcastSedaRecoverableConsumerNewTransactionTest` and `HazelcastSedaRecoverableConsumerRollbackTest` verified locally
- [ ] Full CI build passes


components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
```diff
@@ -101,6 +103,8 @@ protected void doStart() throws Exception {
         }
 
         executor = endpoint.createExecutor(this);
+        taskExecutor = endpoint.getCamelContext().getExecutorServiceManager()
```
davsclaus, line 106: Minor observation: this creates a single-thread scheduled executor shared across all `concurrentConsumers` threads. When multiple subscribers fail simultaneously, their backoff delays are serialized rather than independent (unlike the prior `ForegroundTask` which slept per-thread).

In practice the impact is negligible since the delay supplier is `() -> true` (near-instant execution), but worth noting the behavioral difference. The same applies to `HazelcastSedaConsumer`.

components/camel-mongodb-gridfs/src/main/java/org/apache/camel/component/mongodb/gridfs/GridFsConsumer.java
```diff
@@ -107,12 +114,15 @@ public void run() {
             fromDate = new Date();
         }
 
-        BlockingTask task = Tasks.foregroundTask()
-                .withBudget(Budgets.iterationBudget()
-                        .withMaxIterations(IterationBoundedBudget.UNLIMITED_ITERATIONS)
+        BlockingTask task = Tasks.backgroundTask()
+                .withBudget(Budgets.iterationTimeBudget()
+                        .withMaxIterations(Integer.MAX_VALUE)
```
davsclaus: Nit: `Integer.MAX_VALUE` is already the default for `IterationTimeBoundedBudgetBuilder.maxIterations`, so this call could be omitted entirely:

```suggestion
                        // maxIterations defaults to Integer.MAX_VALUE (effectively unlimited)
```

Alternatively, keeping it explicit is fine for readability — just noting the redundancy.

davsclaus (approved):
Thanks for this well-structured migration, Guillaume! The changes look good — the ForegroundTask-to-BackgroundTask migration is mechanically sound across all 9 components, executor lifecycle management is correct, and CI passes.

A couple of minor observations noted inline — nothing blocking.

This review does not replace specialized static analysis tools (SonarCloud) or AI review tools (CodeRabbit).

components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
gnodet: Good observation. The single-thread executor does serialize backoff delays across concurrent subscribers. However, this is a single-iteration delay task (maxIterations=1) used only as a timed pause before retry — it completes almost instantly once the initial delay expires. With a 10-second reconnect delay, even if multiple subscribers fail simultaneously, the serialization overhead is negligible compared to the delay itself.

That said, if this becomes a concern in practice (e.g. very high concurrency with frequent failures), the executor pool size could be bumped. For now, keeping it single-threaded matches the pattern used by the other components in this PR.

components/camel-mongodb-gridfs/src/main/java/org/apache/camel/component/mongodb/gridfs/GridFsConsumer.java
gnodet: Good catch — removed the redundant `withMaxIterations(Integer.MAX_VALUE)` and added a comment noting the default. Fixed in the next push.

oscerd (approved):

davsclaus (approved):

davsclaus:
LGTM

PR #25165, opened 2026-07-27
CAMEL-24216: Set xpathTotalOpLimit per TransformerFactory instead of JVM-global system property

gnodet:
## Summary

Fixes [CAMEL-24216](https://issues.apache.org/jira/browse/CAMEL-24216): `jdk.xml.xpathTotalOpLimit` was set as a JVM-global system property in `XsltComponent.doInit()` and cleared in `doShutdown()`. This approach had multiple problems:

1. **Not restored on failure**: If `doStart()` failed, the property was never cleaned up because `doShutdown()` required the service to reach a certain lifecycle state.
2. **Not safe for concurrent use**: Two endpoints with different `xpathTotalOpLimit` values would race on the single JVM-global property.
3. **Leaked globally**: The system property affected all XML processing in the JVM, not just Camel's XSLT endpoints.

## Approach

Instead of manipulating a JVM-global system property, set the limit directly on each `TransformerFactory` instance via `factory.setAttribute("jdk.xml.xpathTotalOpLimit", value)`. This is:

- **Thread-safe**: Each endpoint gets its own factory with its own limit
- **Leak-free**: No system property is touched
- **Per-endpoint configurable**: The `xpathTotalOpLimit` option is available both at component level (default for all endpoints) and at endpoint level (override via URI parameter, e.g. `?xpathTotalOpLimit=20000`)

For `xslt-saxon`, the attribute is set inside a try-catch since Saxon's `TransformerFactory` may not support this JDK-specific attribute.

## Changes

- **`XsltComponent`**: Removed all system property logic (`doInit`/`doStop`/`doShutdown` overrides). The `xpathTotalOpLimit` field is kept as a component-level default that flows to endpoints.
- **`XsltEndpoint`**: Added `@UriParam xpathTotalOpLimit` field. In `createXsltBuilder()`, resolves the effective limit (endpoint value → component value) and sets it on the `TransformerFactory` via `setAttribute()`. Creates a default factory if none was explicitly configured so the attribute can be set.
- **`XsltSaxonEndpoint`**: Added the same per-factory attribute setting with a try-catch for Saxon compatibility.
- **`XsltTotalOpsTest`** (new, replaces `ZXsltTotalOpsTest`): 4 test methods covering limit rejection, no system property leakage, high-enough limit success, and per-endpoint URI configuration.
- **Generated files**: Regenerated `XsltEndpointConfigurer`, `XsltEndpointUriFactory`, `xslt.json`, `XsltSaxonEndpointUriFactory`, `xslt-saxon.json`.

## Test plan

- [x] `testXsltTotalOpsLimitRejectsStylesheet` — verifies a limit of 1 correctly rejects a stylesheet
- [x] `testXsltTotalOpsLimitDoesNotLeakSystemProperty` — verifies no JVM system property is set
- [x] `testXsltTotalOpLimitHighEnoughAllowsStylesheet` — verifies a reasonable limit allows normal stylesheets
- [x] `testXsltTotalOpLimitPerEndpointUri` — verifies the limit works when set via endpoint URI parameter
- [x] All 41 existing XSLT tests pass (including `XsltFromFileExceptionTest` which was sensitive to system property leakage)


oscerd (approved):

gnodet:

Merged. JIRA [CAMEL-24216](https://issues.apache.org/jira/browse/CAMEL-24216) fixVersion should be set to **4.22.0** and resolved as Fixed.

graben:
Hi @gnodet, I think this fix has a drawback. If using saxon as XSLT processor you'll get an exception on route startup because Saxon does not know xpathtotaloplimit. The Saxon component does catching. Maybe do her either?

Greetings Benjamin

PR #25166, opened 2026-07-27
fix(CAMEL-24264): throw IllegalArgumentException for wrong body type in pojoRequest mode [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
## CAMEL-24264: throw IllegalArgumentException when `pojoRequest=true` with wrong body type

**Problem:**
With `pojoRequest=true`, `Translate2Producer.translateText` only acted when the
body was a `TranslateTextRequest`. Any other body type fell through the
the original body was returned silently — making the bug extremely hard to debug.

**Fix:**
Add an `else` branch in the POJO mode `if` block that throws
`IllegalArgumentException`, making the failure visible immediately.

**Scope:**
Only the `pojoRequest=true` path. The non-POJO path (String body) is
unchanged — it already handles its input validation separately.


oscerd (commented):
This is assigned already. Look at the issues.

PR #25167, opened 2026-07-27
CAMEL-24223: camel-jbang-mcp - Locale.ROOT cost formatting and model-specific disclaimer (follow-up)

oscerd:
### Motivation

Follow-up to [CAMEL-24223](https://issues.apache.org/jira/browse/CAMEL-24223) / #25052, which merged before these review refinements landed. Picks up the remaining points from that PR's review.

### Changes

- **`formatCost` now uses `Locale.ROOT`** for both branches. Previously `String.format("$%.6f", cost)` used the JVM default locale, so under a comma-decimal locale (e.g. French) the tool emitted `$0,003000` — MCP clients that parse these strings would break. This is the substantive fix.
- **Model-specific disclaimer.** The top-level disclaimer now states explicitly that token-based (LLM) components are priced as a *specific* model, not a provider-wide figure, and that Nova Lite / Haiku / local models (Ollama) cost far less, often ~zero — so the estimate is not misread as a real cost for those. Addresses davsclaus's concern about phantom costs for provider-agnostic components.
- **Tighter tests.** `shouldExtractSchemesCorrectly` now uses `containsExactlyInAnyOrder("file", "docling", "aws-bedrock", "aws2-s3")` (catches both a false `uri` capture and a dropped `file` scheme), and `shouldEstimateCostForAiPipeline` pins the breakdown to exactly 3 priced components and asserts the computed `aws-bedrock` per-execution cost rather than only the count.

### Testing

`RouteCostEstimateToolsTest` — 11 tests, all passing against current `main`.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/RouteCostEstimateTools.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

gnodet (approved):
Clean, well-targeted follow-up that addresses three review points from the predecessor PR (#25052): `Locale.ROOT` in `formatCost` prevents locale-dependent decimal separators breaking MCP clients, the disclaimer now clarifies model-specific pricing to avoid phantom-cost confusion, and test assertions are tightened to catch regressions more precisely.

Well-scoped follow-up — each change traces directly to a specific review comment on PR #25052. The expanded disclaimer text is a thoughtful balance: it gives concrete examples (Nova Lite, Haiku, Ollama) so users can calibrate, without becoming a pricing reference document.

PR #25168, opened 2026-07-27
CAMEL-24222: camel-jbang-mcp - catalog-based scheme detection in dependency audit (follow-up)

oscerd:
### Motivation

Follow-up to [CAMEL-24222](https://issues.apache.org/jira/browse/CAMEL-24222) / #25051, which merged before this refinement landed. Addresses @gnodet's remaining review point on scheme detection.

### Changes

- **`extractUsedSchemes` now walks the catalog component names** (`catalog.findComponentNames()` + a `containsComponent()` check), the same approach `DependencyCheckTools` and `HardenTools` use, instead of tokenizing the route text with `split("[^a-z0-9-]+")`. Tokenizing treated the DSL's own structural keywords (`from:`, `to:`, `steps:`, `uri:`) as component schemes, producing false positives in the reachability analysis.
- **Real reachability tests.** `shouldIncludeReachabilityWhenRoutesProvided` now asserts `camel-http` is actually flagged `reachable=true` when a route uses the http component, and a new `shouldNotFlagReachabilityWhenRoutesUseOtherComponents` asserts it is NOT reachable when the route uses a different component. The previous assertion (`>= 0`) could never fail.

### Testing

`DependencySecurityAuditToolsTest` — 8 tests, all passing against current `main`.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/DependencySecurityAuditTools.java
+++ b/dsl/camel-jbang/camel-jbang-mcp/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/DependencySecurityAuditTools.java
@@ -77,7 +77,7 @@ public class DependencySecurityAuditTools {
             List<SecurityAdvisoryModel> allAdvisories = advisoryService.advisories();
 
             List<String> usedSchemes = routes != null && !routes.isBlank()
-                    ? extractUsedSchemes(routes) : List.of();
+                    ? extractUsedSchemes(routes, catalog) : List.of();
 
             Map<String, ArtifactAudit> auditByArtifact = new LinkedHashMap<>();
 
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

gnodet (approved):
Clean, well-targeted follow-up that replaces the ad-hoc token-based scheme detection with the catalog-based approach already established in `DependencyCheckTools` and `HardenTools`. The new tests are meaningful — the old `isGreaterThanOrEqualTo(0)` assertion was effectively a no-op since the value is always >= 0. The new test asserts `isGreaterThan(0)` and verifies that `camel-http` specifically has `reachable=true`. The negative test case (`shouldNotFlagReachabilityWhenRoutesUseOtherComponents`) properly validates the false-positive elimination by ensuring `timer`/`log` routes do not flag `camel-http` as reachable.

This correctly addresses the original review findings — the old `extractUsedSchemes` tokenized route text and treated YAML structural keywords (`from:`, `to:`, `steps:`) as component schemes. The new approach using `catalog.findComponentNames()` is consistent with how `DependencyCheckTools` and `HardenTools` detect component usage.

PR #25169, opened 2026-07-27
CAMEL-24278: BackgroundTask keeps task registered in TaskManagerRegistry during blocking run

davsclaus:
## Summary

When a `BackgroundTask` is used via the blocking `run()` method (Kafka reconnection, PgEvent, SMPP), the task previously removed itself from the `TaskManagerRegistry` inside `runTaskWrapper()` as soon as the supplier succeeded or the budget exhausted — before `run()` returned. This made the task invisible to the Internal Tasks dev console during active recovery.

**camel-support (BackgroundTask):**
- Register the task in the registry immediately when `run()` is called (before the first scheduled tick)
- Skip add/remove in `runTaskWrapper()` for the `run()` path — let `waitForTaskCompletion()` handle cleanup
- The `schedule()` path (non-blocking, used by SJMS and Master) is unchanged
- Added 3 unit tests verifying registry lifecycle during blocking run

**camel-kafka (KafkaFetchRecords):**
- Added broker connectivity check in `reconnectTask()` via `consumer.partitionsFor(topic, timeout)` after creating and initializing the consumer
- Previously, `reconnectTask()` always returned `true` because creating a Kafka consumer object succeeds even without a broker — the actual failure only happened later during polling
- Now the BackgroundTask properly retries with backoff until the broker is genuinely reachable, and the task stays visible in the Internal Tasks registry during that time

## Test plan

- [x] New `BackgroundTaskRegistryTest` — 3 tests covering immediate registration, persistence during retries, and cleanup after completion
- [x] Existing `BackgroundTaskTest` and `BackgroundIterationTimeTaskTest` pass (13 tests)
- [ ] Manual: stop Kafka broker, verify Internal Tasks dev console shows `KafkaReconnect-*` tasks with retry count incrementing
- [ ] Manual: restart Kafka broker, verify tasks complete and disappear from registry

### DIFF
--- a/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
+++ b/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaFetchRecords.java
@@ -226,6 +226,10 @@ public class KafkaFetchRecords implements Runnable {
             }
 
             initializeConsumer();
+
+            // verify broker is reachable by requesting topic metadata
+            String topic = topicName != null ? topicName : "*";
+            consumer.partitionsFor(topic, Duration.ofMillis(currentBackoffInterval));
         } catch (Exception e) {
             setConnected(false);
             LOG.warn("Error creating/subscribing org.apache.kafka.clients.consumer.KafkaConsumer due to: {}",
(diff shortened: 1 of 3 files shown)


davsclaus:
Closing — these changes were already committed directly to main in 49c7d5b7a1f6 and e61ba08a726a.

PR #25184, opened 2026-07-28
CAMEL-24280: camel-aws2-mq - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `MQ2Producer`'s `listBrokers`, `createBroker`,
`deleteBroker`, `rebootBroker`, `updateBroker` and `describeBroker` branches only
acted when the body was the matching request type; any other body fell through
the `instanceof` with no `else`, so no AWS call was made, no response was set,
and the original body was returned with no error.

## Fix

type, e.g.:

```java
} else {
            "updateBroker operation requires UpdateBrokerRequest in POJO mode");
}
```

## Tests

Six tests in `MQProducerTest` send a wrong-typed body to `pojoRequest=true`
routes (reusing the module's `AmazonMQClientMock`) and assert the message. Each
was verified to fail (silent no-op) before the fix — `rebootBroker`, for example,
reports "Expecting code to raise a throwable" on `main`. Class 15/15 green.
Hermetic unit tests against the in-JVM mock (run region-free) — no localstack or
real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-mq/src/main/java/org/apache/camel/component/aws2/mq/MQ2Producer.java
+++ b/components/camel-aws/camel-aws2-mq/src/main/java/org/apache/camel/component/aws2/mq/MQ2Producer.java
@@ -133,6 +133,9 @@ public class MQ2Producer extends DefaultProducer {
                 message.setBody(result);
                 message.setHeader(MQ2Constants.NEXT_TOKEN, result.nextToken());
                 message.setHeader(MQ2Constants.IS_TRUNCATED, ObjectHelper.isNotEmpty(result.nextToken()));
+            } else {
+                throw new IllegalArgumentException(
+                        "listBrokers operation requires ListBrokersRequest in POJO mode");
             }
         } else {
             ListBrokersRequest.Builder builder = ListBrokersRequest.builder();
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

davsclaus (approved):

gnodet (approved):
Clean, well-scoped PR that adds missing `else` branches in six `MQ2Producer` POJO-mode operations to throw `IllegalArgumentException` when the body is the wrong type, exactly matching the pattern established by sibling PRs (CAMEL-24277/MSK, CAMEL-24263/STS, CAMEL-24264/Translate). CI is green, tests are correct, and project conventions are followed.

Key observations:
- Implementation is consistent with the established pattern from sibling PRs in the CAMEL-24261 family.
- Tests correctly use AssertJ `assertThatThrownBy` (per project conventions), assert both `hasRootCauseInstanceOf(IllegalArgumentException.class)` and the exact message, and use package-private visibility.
- All six operations are covered: listBrokers, createBroker, deleteBroker, rebootBroker, updateBroker, describeBroker.

Minor pre-existing note: the log message in `describeBroker`'s POJO branch (line ~360) says "Reboot Broker command" instead of "Describe Broker command" — a copy-paste issue from before this PR, worth fixing in a follow-up.

PR #25187, opened 2026-07-28
CAMEL-24281: camel-platform-http-main - fail closed when JWT authentication has no issuer or audience

oscerd:
# CAMEL-24281

When JWT authentication is enabled on the camel-main embedded HTTP server, `JWTAuthenticationConfigurer.buildJwtOptions(...)` returns `null` if neither `jwtIssuer` nor `jwtAudience` is configured, and the caller then skips `JWTAuthOptions.setJWTOptions(...)`. The resulting Vert.x `JWTAuth` is built from the keystore alone, so tokens are only checked for signature and expiry and the `iss` and `aud` claims are not validated. Nothing signals this: the server starts normally, so a deployment configured the documented way can be enforcing less than intended.

## Change

* `assertIssuerOrAudienceConfigured(...)` is called at the top of **both** `configureAuthentication` overloads, so the application server and the management server are both covered. It throws `IllegalArgumentException` when a JWT keystore is configured but neither value is set, naming the properties involved.
* New `jwtAllowMissingIssuerAndAudience` option (default `false`, marked `security = "insecure:dev"`) on `HttpServerConfigurationProperties` and `HttpManagementServerConfigurationProperties`, for deployments that genuinely want signature and expiry validation only.

This mirrors the posture `camel-oauth` already uses, where `DefaultOAuthTokenValidationFactory.validateResolvedConfiguration` refuses to operate unless an expected audience and issuer are configured (`allow-missing-audience` and `allow-missing-issuer` both default to `false`).

## Compatibility

This changes startup behaviour for an existing configuration, so it is documented in the 4.22 upgrade guide together with the opt-out.

`jwtIssuer` and `jwtAudience` were only added in 4.21.0 by CAMEL-23525, so the fail-closed behaviour applies to `main` only. It cannot be backported as-is, because 4.18.x and 4.14.x have no such options to satisfy the requirement. Those branches instead receive the options themselves, so operators there can enforce `iss`/`aud` by configuration:

* #25195 → `camel-4.18.x` (4.18.4)
* #25196 → `camel-4.14.x` (4.14.9)

Both are additive and change no existing behaviour.

## Tests

`JWTIssuerAudienceRequiredMainHttpServerTest` (4): startup fails without an issuer or audience and the message names all three properties; the opt-out starts and still serves; a token minted for a different issuer and audience is rejected once both are configured; the opt-out defaults to false on both properties classes.

One existing test config, `jwt-auth.properties`, is keystore-only and therefore no longer starts. It now sets `jwtAllowMissingIssuerAndAudience=true`, which keeps `JWTAuthenticationMainHttpServerTest` testing what it was written to test and exercises the opt-out.

Module suite green (23), `camel-main` green (231), full reactor build green with generated files committed. Re-verified after merging `main`, which brought Vert.x 4.5.31.

The component documentation previously described signature and expiry checking as the default with issuer and audience as an optional extra. It now states that without them any unexpired token signed by a trusted key is accepted regardless of issuer or audience.

## Review round

* `security = "insecure:dev"` added to the opt-out on both properties classes (@gnodet), verified through to the generated metadata.
* Test now shares one `Vertx` instance and closes it in `@AfterAll` instead of creating one per test method (@gnodet).
* Merged `main` and resolved the `camel-4x-upgrade-guide-4_22.adoc` conflict with the CAMEL-24279 entry; both notes kept.


davsclaus (approved):
Well-designed security hardening — fail closed when JWT authentication has no issuer or audience configured.

**What I checked:**

- `assertIssuerOrAudienceConfigured()` is called at the top of both `configureAuthentication` overloads (application server and management server), before the `JWTAuth` is constructed — fails fast at startup
- `jwtAllowMissingIssuerAndAudience` defaults to `false` on both properties classes, following the project's "denied unless opted in" security posture
- The error message names all three properties (`jwtIssuer`, `jwtAudience`, `jwtAllowMissingIssuerAndAudience`) so the operator knows exactly what to set
- Test coverage: startup failure without issuer/audience, opt-out works, foreign-token rejection with issuer+audience configured, default value verified
- Existing `jwt-auth.properties` adapted with the opt-out to keep `JWTAuthenticationMainHttpServerTest` testing what it was designed to test
- Upgrade guide entry present with the opt-out, component docs updated, generated files regenerated
- Conventions followed: AssertJ, package-private test class, `@Metadata(label = "security")`

core/camel-main/src/main/java/org/apache/camel/main/HttpServerConfigurationProperties.java
```diff
@@ -68,6 +68,8 @@ public class HttpServerConfigurationProperties implements BootstrapCloseable {
     private String jwtIssuer;
     @Metadata(label = "security")
     private String jwtAudience;
+    @Metadata(label = "security", defaultValue = "false")
+    private boolean jwtAllowMissingIssuerAndAudience;
```
gnodet, line 72: The `jwtAllowMissingIssuerAndAudience` flag weakens security (it opts out of issuer/audience validation), so per the project convention it should carry `security = "insecure:dev"` on its `@Metadata` annotation — the same pattern used for `trustAllCertificates`, `allowJavaSerializedObject`, etc. Without it, the security policy framework (`camel.main.profile = prod`) cannot warn about or block this flag in production.

```suggestion
    @Metadata(label = "security", defaultValue = "false", security = "insecure:dev")
    private boolean jwtAllowMissingIssuerAndAudience;
```

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
```diff
@@ -77,6 +77,8 @@ public class HttpManagementServerConfigurationProperties implements BootstrapClo
     private String jwtIssuer;
     @Metadata(label = "security")
     private String jwtAudience;
+    @Metadata(label = "security", defaultValue = "false")
+    private boolean jwtAllowMissingIssuerAndAudience;
```
gnodet, line 81: Same here — this file already uses `security = "insecure:dev"` on `devConsoleEnabled`, `uploadEnabled`, `downloadEnabled`, and `sendEnabled`, so the pattern is well established.

```suggestion
    @Metadata(label = "security", defaultValue = "false", security = "insecure:dev")
    private boolean jwtAllowMissingIssuerAndAudience;
```

components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/authentication/JWTIssuerAudienceRequiredMainHttpServerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    void optOutStartsAndValidatesSignatureAndExpiryOnly() {
+        // jwt-auth.properties sets jwtAllowMissingIssuerAndAudience=true
+        Main main = MainHttpServerAuthenticationTestSupport.createMain(
+                "jwt-auth.properties", port, new PlatformHttpRouteBuilder());
+        try {
+            main.start();
+
```
gnodet, line 83: Minor: the `Vertx.vertx()` instances created here and in `configuredIssuerAndAudienceRejectAForeignToken()` are never closed, leaking native event-loop threads. This follows the pre-existing pattern in `JWTAuthenticationMainHttpServerTest`, but consider extracting a shared `Vertx` instance with proper cleanup.

gnodet (commented):
Well-crafted security-hardening PR that correctly implements fail-closed JWT authentication. The fail-closed logic in `assertIssuerOrAudienceConfigured` is correct: it runs eagerly at configuration time, the `IllegalArgumentException` propagates cleanly to `main.start()`, and the error message correctly names the relevant properties without leaking sensitive values.

Two observations about the `@Metadata` annotation on the new opt-out flag, and two minor test observations below.

core/camel-main/src/main/java/org/apache/camel/main/HttpServerConfigurationProperties.java
oscerd: Done — jwtAllowMissingIssuerAndAudience in HttpServerConfigurationProperties now carries @Metadata(label = "security", defaultValue = "false", security = "insecure:dev"), so the prod profile can flag it.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
oscerd: Done — same annotation applied to the flag in HttpManagementServerConfigurationProperties, matching devConsoleEnabled/uploadEnabled/etc.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

components/camel-platform-http-main/src/test/java/org/apache/camel/component/platform/http/main/authentication/JWTIssuerAudienceRequiredMainHttpServerTest.java
oscerd: Done — the test now uses a single shared private static final Vertx VERTX = Vertx.vertx() closed in @AfterAll, so it no longer leaks event-loop threads.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

oscerd:
Two follow-up notes on the review round, beyond the three inline replies.

**On `insecure:dev`:** applied to both properties classes as suggested, and it propagates correctly into the generated metadata for `camel.server.jwtAllowMissingIssuerAndAudience` and `camel.management.jwtAllowMissingIssuerAndAudience`. Worth flagging though that `design/security.adoc` defines `insecure:dev` as *"development/debug features that should not be enabled in production"* (`devConsoleEnabled`, `debugBreakpoints`), whereas this flag weakens **authentication** rather than enabling a dev feature. None of the three valid values (`insecure:ssl`, `insecure:serialization`, `insecure:dev`) really covers auth-weakening options. `insecure:dev` is the closest fit and matches what this file already does, so it is what I used, but a dedicated `insecure:auth` category may be the better long-term answer. That would be a separate change across the annotation, the docs and `SecurityUtils` generation, so I have not attempted it here.

**Re-verified against the Vert.x upgrade:** `main` moved from Vert.x 4.5.28 to 4.5.31 while this PR was open. Since the change depends on how Vert.x applies `JWTOptions` claim checks, I merged `main` and re-ran rather than assuming: module suite is green (23 tests) on 4.5.31, behaviour unchanged. The merge also resolved a conflict in `camel-4x-upgrade-guide-4_22.adoc` against the CAMEL-24279 entry that landed in the meantime; both notes are kept.

Full reactor build `mvn clean install -DskipTests` green, no regenerated-file drift.

PR #25194, opened 2026-07-28
CAMEL-24283: camel-aws2-comprehend - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Comprehend2Producer`'s nine operations
(`detectDominantLanguage`, `detectEntities`, `detectKeyPhrases`,
`detectSentiment`, `detectSyntax`, `detectPiiEntities`, `detectToxicContent`,
`classifyDocument`, `containsPiiEntities`) only acted when the body was the
matching request type; any other body fell through the `instanceof` with no
returned with no error.

## Fix

type, e.g. `"classifyDocument operation requires ClassifyDocumentRequest in POJO
mode"`.

## Tests

A single `@ParameterizedTest` covers all nine operations, sending a wrong-typed
body to `pojoRequest=true` routes (reusing the module's mock client) and
asserting each message. Verified to fail (silent no-op) before the fix — with the
`classifyDocument` else removed, exactly that case reports "Expecting code to
raise a throwable". Class 18/18 green. Hermetic unit tests, region-free — no
localstack or real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-comprehend/pom.xml
+++ b/components/camel-aws/camel-aws2-comprehend/pom.xml
@@ -78,5 +78,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

gnodet (approved):
Clean, mechanical change adding type-safety guards to all 9 AWS Comprehend POJO-mode operations. Follows the established pattern from merged sibling PRs (STS #25153, Translate #25158, MSK #25170, MQ #25184). CI is green, tests are well-structured, and the shared upgrade-guide entry is already on main.

**Observations (no action needed):**
- The `else` branches are correctly placed at the right nesting level in all 9 methods — inside the outer `if (isPojoRequest())` block, pairing with the `if (payload instanceof XxxRequest)` check.
- The parameterized test covers all 9 operations with correct `@CsvSource` entries matching the error messages in production code.
- AssertJ dependency addition to `pom.xml` is justified since the module did not previously have it.
- Pre-existing: test class is `public` (should be package-private per conventions), but the new test method correctly uses package-private visibility — convention says to only fix what you touch.

PR #25199, opened 2026-07-28
CAMEL-24285: camel-aws2-polly - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Polly2Producer`'s nine operations only acted when the
body was the matching request type; any other body fell through the `instanceof`

## Fix

type, e.g. `"getLexicon operation requires GetLexiconRequest in POJO mode"`.

## Tests

no-op) before the fix — with the `getLexicon` else removed, exactly that case
reports "Expecting code to raise a throwable". Class 20/20 green. Hermetic unit

## Docs / backport

---


davsclaus (approved):

gnodet (approved):
Clean, well-structured PR that adds `IllegalArgumentException` when `pojoRequest=true` and the body is the wrong type across all 9 Polly operations. Follows the established CAMEL-24261 series pattern identically to the already-approved sibling PRs (#25210, #25215).

**Key observations:**
- All 9 operations (`synthesizeSpeech`, `describeVoices`, `listLexicons`, `getLexicon`, `putLexicon`, `deleteLexicon`, `startSpeechSynthesisTask`, `getSpeechSynthesisTask`, `listSpeechSynthesisTasks`) correctly covered by both production code and parameterized test
- `else` branch placement is correct — sits inside the `isPojoRequest()` true branch, after the `instanceof` check
- Test correctly uses `hasRootCauseInstanceOf(IllegalArgumentException.class)` to account for Camel's `CamelExecutionException` wrapping
- Error message format consistent with sibling PRs in the series
- New test method is package-private, uses AssertJ — follows project conventions

PR #24945, opened 2026-07-20
Revert "CAMEL-24026: Fix flaky XsltFromFileExceptionTest"

apupier:
Reverts apache/camel#24625

the issue is still failing sometimes. the error message is less descriptive than before so reverting

PR #25201, opened 2026-07-28
CAMEL-24022: Fix flaky XsltFromFileExceptionTest

gnodet:

## Summary

Fix the flaky `XsltFromFileExceptionTest` by eliminating the race condition between file creation and consumer polling.

## Root Cause Analysis

Develocity and JIRA (CAMEL-24026 comments from @apupier) reveal **two distinct failure modes**:

### Mode A — JAXP system property leak (primary cause, already fixed)

`ZXsltTotalOpsTest` set `jdk.xml.xpathTotalOpLimit=1` as a JVM-global system property (CAMEL-24216 bug). When `XsltFromFileExceptionTest` ran afterwards in the same JVM, `example.xsl` — which has 2 XPath operators (`match="/"` and `select="/hello"`) — failed to compile:

```
TransformerConfigurationException: JAXP0801003: the compiler encountered XPath
expressions with an accumulated '2' operators that exceeds the '1' limit set by 'system property'
```

**Already fixed on `main`:**
- Workaround (July 21, commit 1a11849): `@AfterEach` clearing the system property
- Proper fix (July 28, CAMEL-24216, commit f46f5af): set limit per-`TransformerFactory` instance

### Mode B — Exchange timeout (secondary, rare)

```
AssertionFailedError: Exchange should have completed ==> expected: <true> but was: <false>
```

`oneExchangeDone.matchesWaitTime()` times out after 10 seconds — the consumer never completed the exchange within the timeout. After the mode-A workaround, this still occurred at 2/195 (~1%), though those may also be residual mode-A failures.

The prior fix attempt ([CAMEL-24026](https://issues.apache.org/jira/browse/CAMEL-24026), [PR #24625](https://github.com/apache/camel/pull/24625)) reordered assertions but was [reverted](https://github.com/apache/camel/pull/24945) because it didn't help.

## Fix

Eliminate the race between file creation and consumer polling:

1. Override `isUseRouteBuilder()` to return `false`, deferring route addition and context startup
2. Write test files using `Files.writeString()` before starting the context
3. Add the route and start the context — the file consumer's first poll (`initialDelay=0`) immediately finds the file
4. Use Awaitility for file-move assertions (file move is asynchronous relative to mock message delivery)

This makes the test deterministic — the consumer always finds the file on its first poll.

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
@@ -16,45 +16,65 @@
  */
 package org.apache.camel.component.xslt;
 
+import java.nio.file.Files;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.camel.ContextTestSupport;
-import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
 import org.junit.jupiter.api.Test;
 
-/**
- *
- */
-public class XsltFromFileExceptionTest extends ContextTestSupport {
+import static org.awaitility.Awaitility.await;
 
-    @Test
-    public void testXsltFromFileExceptionOk() throws Exception {
-        getMockEndpoint("mock:result").expectedMessageCount(1);
-        getMockEndpoint("mock:error").expectedMessageCount(0);
+class XsltFromFileExceptionTest extends ContextTestSupport {
 
-        template.sendBodyAndHeader(fileUri(), "<hello>world!</hello>", Exchange.FILE_NAME, "hello.xml");
-
-        oneExchangeDone.matchesWaitTime();
-
-        assertMockEndpointsSatisfied();
-
-        assertFileNotExists(testFile("hello.xml"));
-        assertFileExists(testFile("ok/hello.xml"));
+    @Override
+    public boolean isUseRouteBuilder() {
+        return false;
     }
 
     @Test
-    public void testXsltFromFileExceptionFail() throws Exception {
-        getMockEndpoint("mock:result").expectedMessageCount(0);
-        getMockEndpoint("mock:error").expectedMessageCount(1);
+    void testXsltFromFileExceptionOk() throws Exception {
+        getMockEndpoint("mock:result").expectedMessageCount(1);
+        getMockEndpoint("mock:error").expectedMessageCount(0);
 
-        // the last tag is not ended properly
-        template.sendBodyAndHeader(fileUri(), "<hello>world!</hello", Exchange.FILE_NAME, "hello2.xml");
+        // Write file BEFORE starting the route so the file consumer
+        // picks it up on its very first poll — eliminates the race between
+        // file write and consumer poll scheduling under CI load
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

gnodet (commented):
**Review: Needs clarification** 💬

The assertion-order swap is logically correct — waiting for exchange completion before verifying mock state is the right approach. However, this is essentially the same swap that was merged in PR #24625 (CAMEL-24026) and then reverted in PR #24945 because "the issue is still failing sometimes."

**Key difference from the reverted version:** By omitting the `assertTrue()` wrapper around `matchesWaitTime()`, a silent timeout at 10s gives the subsequent `assertMockEndpointsSatisfied()` a fresh 10-second window — effectively doubling the total available wait time from 10s to ~20s. This is a real improvement, but the mechanism is implicit and fragile (it depends on *not* asserting a return value).

**Suggestions:**

1. Acknowledge the prior revert (PR #24945) in the PR description and explain how this version differs.
2. Consider making the extended wait time explicit — e.g., using `oneExchangeDone.matches(30, TimeUnit.SECONDS)` or Awaitility (as suggested during the revert review).
3. Minor: drop the `public` modifier from the class and both modified test methods per project conventions (CLAUDE.md "Test Visibility").

apupier:
> The test was failing intermittently in CI (6/254 runs in July 2026 per Develocity) due to a race condition.

just for precision: there was a fix on 20 or 21 July. The new values are now 2 failed out of 195 builds. it was flaky without making the build failing

gnodet:
> just for precision: there was a fix on 20 or 21 July. The new values are now 2 failed out of 195 builds. it was flaky without making the build failing

Thanks for the precision, @apupier! I've updated the PR description to avoid implying the 6/254 was current. The fix still addresses the structural root cause — the file is now written before the consumer starts polling, eliminating the race regardless of failure frequency.

gnodet:
@gnodet Thanks for the thorough review! I want to clarify that the review was on the **first version** of the fix (commit d702cc17 — the simple reorder), which has since been superseded by commit 4866fe7b — a **completely different approach** that addresses the root cause.

Your concerns about the reorder approach were exactly right — that's why the fix was rewritten. The current version:

1. **Does NOT reorder `assertMockEndpointsSatisfied()` and `oneExchangeDone.matchesWaitTime()`** — it eliminates `oneExchangeDone` entirely
2. **Does NOT rely on implicit timeout stacking** — the race condition is structurally eliminated
3. **Writes the file BEFORE starting the route** via `Files.writeString()`, so the consumer's first poll immediately finds it
4. **Uses `isUseRouteBuilder() = false`** to defer context startup until the file is in place
5. **Uses Awaitility** for file-move assertions (as you suggested)
6. **Already drops `public`** from the class and test methods (see line 28, 36, 59 in the current code)

The prior revert (PR #24945) is acknowledged in the updated PR description.

gnodet:
Closing — the primary cause of the flakiness was the JAXP system property leak from `ZXsltTotalOpsTest` (CAMEL-24216), which is already fixed on `main`. The remaining exchange-timeout failures (2/195) could not be proven as an independent issue.

PR #25203, opened 2026-07-28
CAMEL-23853: Embed MCP server on camel run/dev and Camel Main

atiaomar1978-hub:
## Summary

- Adds `camel.mcp.*` configuration and an MCP Streamable HTTP endpoint on the embedded **management HTTP server** (`/mcp` by default), wired through `EmbeddedMcpRequestHandler`.
- Adds `camel run` / `camel dev` `--mcp` (and `camel.jbang.mcp`) to enable embedded MCP with JBang `ToolRegistry` tools against the local process (management server bound to `127.0.0.1` by default).
- Includes unit tests for the JSON-RPC engine, management HTTP wiring, and CLI flag parsing; documents the feature in Camel Main / JBang docs and the 4.22 upgrade guide.

## Test plan

- [x] `mvn test -pl components/camel-platform-http-main,dsl/camel-jbang/camel-jbang-core -am -Dtest=StreamableMcpJsonRpcEngineTest,EmbeddedMcpHttpServerTest,RunMcpOptionTest`
- [ ] Manual: `camel run hello.yaml --mcp` then POST JSON-RPC `tools/list` to `http://127.0.0.1:8080/mcp`

https://issues.apache.org/jira/browse/CAMEL-23853


atiaomar1978-hub:
## AI review summary (Bugbot + Grok)

_Review comments generated by Cursor on behalf of atiaomar1978-hub._

### Bugbot

| Severity | Location | Finding |
|----------|----------|---------|
| High | `BaseMainSupport#setMcpProperties` | Management host stayed `0.0.0.0` when MCP pre-enabled management — **fixed** (always apply `camel.mcp.host`; JBang sets `127.0.0.1`) |
| Medium | `ManagementHttpServer#setupMcp` | MCP route not covered by management auth paths — **noted** (same as other dev consoles; `insecure:dev`) |
| Medium | Catalog metadata | MCP options under `camel.main.*` — **fixed** (`PrepareCamelMainMojo` → `camel.mcp.*`) |

### Grok-style review

| Severity | Finding | Status |
|----------|---------|--------|
| High | Origin CSRF check case-sensitive | **Fixed** (`headerIgnoreCase` + test) |
| High | `--mcp` enabled `camel.server` / port clash | **Fixed** (management only) |
| High | Localhost bind skipped on JBang path | **Fixed** |
| High | Wrong metadata prefix | **Fixed** |
| Medium | Standalone Main needs registry handler | Documented via SPI; JBang registers `EmbeddedMcpMainListener` |
| Medium | MCP vs management auth | Follow-up if we want MCP behind same auth rules |

Follow-up commit: `8137ab4` on this branch addresses the high-severity items above.

atiaomar1978-hub:
## CI follow-up

Addressed the doc failures from [run 30398071305](https://github.com/apache/camel/actions/runs/30398071305) (Antora) and [run 30398074689](https://github.com/apache/camel/actions/runs/30398074689) (incremental build / `xref-check` on `docs`):

- **Cause:** In `camel-4x-upgrade-guide-4_22.adoc`, `xref:main.adoc` does not resolve from the user-manual module (Antora: *target of xref not found: main.adoc*).
- **Fix:** `1f21e97` — use `xref:components:others:main.adoc[Camel Main]` (same pattern as `property-binding.adoc` / `camelcontext-autoconfigure.adoc`).
- **Verified locally:** `mvn -pl docs -am package -DskipTests -Pcamel-xref-check`

Please re-check CI when the new workflow run completes.

gnodet (commented):

## Review: CAMEL-23853 — Embed MCP server on camel run/dev and Camel Main

This is a well-structured feature with clean separation of concerns: `EmbeddedMcpRequestHandler` as the SPI in `camel-main`, `StreamableMcpJsonRpcEngine` as a reusable JSON-RPC engine, and `JbangEmbeddedMcpRequestHandler` bridging to the JBang `ToolRegistry`. The security model is solid — `@Metadata(security = "insecure:dev")` on both `camel.mcp.enabled` and `camel.management.mcpEnabled`, management-server-only exposure, and Origin header CSRF protection all follow established patterns. The test suite and documentation are thorough. Nice work!

A few issues worth addressing:

---

### 1. MCP defaults silently override management server host/port

In `BaseMainSupport.setMcpProperties()`, the guards:

```java
if (mcp.getHost() != null) {
    management.setHost(mcp.getHost());
}
if (mcp.getPort() > 0) {
    management.setPort(mcp.getPort());
}
```

...are always true because `McpConfigurationProperties` has defaults of `"127.0.0.1"` (never null) and `8080` (always > 0). Since the management server's own default host is `"0.0.0.0"`, enabling MCP silently narrows the bind address to localhost.

**Mitigating factor**: `setHttpManagementServerProperties()` runs after `setMcpProperties()`, so explicitly user-set `camel.management.host` properties would override MCP defaults. But users relying on the default `0.0.0.0` would be silently affected.

**Suggested fix**: Use `null` defaults for `host` and `port` in `McpConfigurationProperties`, and only propagate when explicitly set by the user.

---

### 2. `Run.java` overwrites `mainListenerClasses` without appending

In `Run.java`:
```java
main.addInitialProperty("camel.main.mainListenerClasses",
        "org.apache.camel.dsl.jbang.core.commands.mcp.EmbeddedMcpMainListener");
```

This unconditionally sets the property, overwriting any existing value. Compare with `KameletMain.java` which correctly checks for existing listeners and appends with comma separation:

```java
String listeners = getInitialProperties().getProperty("camel.main.mainListenerClasses");
if (listeners == null || listeners.isBlank()) {
    addInitialProperty("camel.main.mainListenerClasses", embedded);
} else if (!listeners.contains(embedded)) {
    addInitialProperty("camel.main.mainListenerClasses", listeners + "," + embedded);
}
```

Currently harmless (no other listener is registered before the MCP block in `Run.java`), but fragile and inconsistent. I'd suggest aligning `Run.java` with the `KameletMain.java` pattern.

---

### 3. Triple-redundant configuration in `Run.java`

The MCP block uses three configuration mechanisms: `writeSetting()`, another `writeSetting()` for `camel.mcp.enabled`, and programmatic `main.configure().mcp().withEnabled(true)`. The existing features (`health`, `metrics`, `console`) each use a single `writeSetting()` call in `Run.java`, with programmatic wiring handled in `KameletMain.java`. I'd suggest following the same pattern for consistency.

---

### 4. Wrong JSON-RPC error code for malformed JSON

In `StreamableMcpJsonRpcEngine`, when `Jsoner.deserialize()` receives malformed JSON, it silently returns an empty `JsonObject` (Jsoner catches `DeserializationException` and returns the default). The engine then returns error code `-32600` ("Invalid Request"). Per JSON-RPC 2.0 spec (section 5.1), malformed JSON should return `-32700` ("Parse error"). An error IS returned, just with the wrong code — wrapping the deserialization in a try-catch to distinguish parse errors from invalid requests would improve spec compliance.

---

### Minor suggestions

- **Hardcoded server version**: `SERVER_VERSION = "1.0.0"` doesn't reflect the actual Camel version (4.22.0-SNAPSHOT). Consider using a Maven-filtered version property for better MCP client diagnostics.
- **Null params in `handleToolsCall`**: When `params` is null, `toolName` is null and passes through to the tool provider. `JbangEmbeddedMcpRequestHandler` handles this gracefully, but returning `-32602` ("Invalid params") at the engine level would be more robust for future `ToolProvider` implementations.

---

Overall this is solid work — the architecture, security annotations, and documentation are all well done. The main items to address are the host/port defaults (finding 1) and the listener overwrite inconsistency (finding 2). 👍

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

Pushed latest commit addressing @gnodet's review on [PR #25203](https://github.com/apache/camel/pull/25203):

1. **MCP host/port defaults** — `McpConfigurationProperties` no longer defaults to `127.0.0.1`/`8080`; `setMcpProperties()` only propagates explicitly configured values so enabling MCP no longer narrows management bind from `0.0.0.0`. JBang `--mcp` still binds localhost via `KameletMain`.
2. **Run.java configuration** — Removed redundant `writeSetting`/`configure()`/`mainListenerClasses` block; `--mcp` now follows the same single `writeSetting(camel.jbang.mcp)` pattern as `--console`/`--health`, with wiring in `KameletMain`.
3. **JSON-RPC compliance** — Malformed JSON returns `-32700` (Parse error); missing `tools/call` params/name returns `-32602` (Invalid params).
4. **Server version** — MCP `initialize` reports the Camel version from `camel-jbang-core` pom.properties.

**Tests:** `StreamableMcpJsonRpcEngineTest` (9), `EmbeddedMcpHttpServerTest` (1), `RunMcpOptionTest` (1), `McpConfigurationPropertiesTest` (1) — all green locally.

jamesnetherton:
Can we slow down a bit and think things through before PR approval.

From the Jira ticket:

> This would allow any Camel application (Spring Boot, Quarkus, standalone) to embed an MCP server, not only applications started via the Camel CLI.

Quarkus already has a solution for MCP [optimized](https://github.com/quarkiverse/quarkus-mcp-server) for the runtime. So some analysis is needed here and in Camel Quarkus for how best to leverage that.

There's also a notes section in the ticket:

> - This needs more design thinking — created this issue to capture the idea before it gets lost. Key questions to consider:
> - How does the embedded MCP server relate to the standalone MCP server ([CAMEL-23852](https://issues.apache.org/jira/browse/CAMEL-23852))?
> - Which MCP tools should be exposed — catalog tools, runtime introspection tools, or both?
> - Should routes automatically become MCP tools, or require explicit opt-in (e.g. via route metadata or a service catalog)?
> - How does this relate to Wanaku for enterprise-scale MCP governance?

Not sure if any of those points have been addressed here?

Croway:
Hello @davsclaus @atiaomar1978-hub @jamesnetherton please hold before merging this PR, yesterday I opened https://issues.apache.org/jira/browse/CAMEL-24308 that will introduce new camel component that integrate ai-tools with an embedded MCP Server, it will provide a default implementation, and a specific implementation for quarkus (based on quarkus-mcp-server), the configuration of the MCP Server will be done via properties, like
```
camel.server.mcp-enabled=true
camel.server.mcp-path=/mcp
camel.server.mcp-server-name=my-integration-app
camel.server.mcp-tags=crm,notify
```
that bridges the ai-tools with the MCP.

I am a little bit concerned about this PR, in particular the configuration part `camel.mcp.*` I am afraid it can be misleading having multiple configurations. I do understand that this PR is for development/management only, but maybe we can improve/review the implementation given CAMEL-24308

atiaomar1978-hub:

Fixed the **sourcecheck** failure in [run 30559900193](https://github.com/apache/camel/actions/runs/30559900193/job/91097909563) (`Fail if there are uncommitted changes`).

The build regenerated `main.adoc` because `camel.mcp.host` and `camel.mcp.port` defaults in the committed docs still showed `127.0.0.1` / `8080`, while the code (and generated output) now use empty defaults after the review fix (null host, port `0`).

Synced:
- `core/camel-main/src/main/docs/main.adoc`
- `catalog/camel-catalog/src/generated/resources/.../main.adoc`

Re-ran MCP tests locally — green (`McpConfigurationPropertiesTest`, `StreamableMcpJsonRpcEngineTest`). Please re-run or wait for CI on the latest commit.

atiaomar1978-hub:
_AI-generated reply on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub) — no code changes in this update._

Thanks @jamesnetherton — fair point, and I agree we should **not rush this to merge** without aligning on the broader design first.

To answer your question directly: **most of the JIRA “notes” questions are not fully resolved in this PR.** This changeset is intentionally a narrow, dev-oriented slice:

| JIRA design question | Addressed in this PR? |
|----------------------|----------------------|
| Embedded MCP for Spring Boot / Quarkus / standalone (not just CLI) | **Partially.** Camel Main gets the plumbing (`camel.mcp.*`, management HTTP endpoint, `EmbeddedMcpRequestHandler` SPI). Standalone Main can enable it via properties, but **only if a handler is registered on the registry**. JBang wires `JbangEmbeddedMcpRequestHandler` for `--mcp`. **No Spring Boot or Camel Quarkus integration** — Quarkus should use [quarkus-mcp-server](https://github.com/quarkiverse/quarkus-mcp-server) as you noted; that analysis is still TODO. |
| Relation to standalone MCP ([CAMEL-23852](https://issues.apache.org/jira/browse/CAMEL-23852)) | **Not designed end-to-end here.** Standalone remains the Quarkus-based `camel mcp` server (catalog + tooling). Embedded is a lightweight Streamable HTTP JSON-RPC endpoint on the **management server** for local dev (`--mcp` / `camel.mcp.enabled`). Overlap and long-term consolidation need a design doc. |
| Which tools — catalog, runtime introspection, or both? | **JBang path only:** exposes existing `ToolRegistry` tools (same family as CLI AI/MCP tooling — catalog lookup, route validation, runtime introspection against the local process). Not a general “expose any Camel app’s routes as tools” story. |
| Routes automatically become MCP tools? | **No.** No route metadata / service-catalog opt-in in this PR. |
| Relation to Wanaku | **Not addressed.** |

So this PR is best seen as **plumbing + JBang dev flag**, not the full CAMEL-23853 vision from the ticket description. Happy to **pause merge**, open a short design discussion (possibly linked to @Croway’s [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308)), and clarify boundaries with Quarkus/Spring Boot before we call it done.

Let me know whether you’d prefer this PR **closed and redesigned**, **converted to draft** pending CAMEL-24308 alignment, or **scoped down further** (e.g. JBang-only behind `--mcp` without the `camel.mcp.*` Main surface) — I’ll follow your lead.

atiaomar1978-hub:
_AI-generated reply on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub) — no code changes in this update._

Thanks @Croway — noted on [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308). I share the concern about **two configuration surfaces** (`camel.mcp.*` here vs the `camel.server.mcp-*` direction you outlined).

This PR was scoped as **dev/management-only** (`insecure:dev`, management HTTP server, localhost binding via JBang), but I agree that without coordination it could confuse users once CAMEL-24308 lands with ai-tools integration and Quarkus-specific wiring.

**Proposal:** hold merge on this PR until we align with CAMEL-24308 — either unify property names/prefixes up front, or clearly document that this path is JBang-local-dev-only and defer production embedded MCP to the component work you’re doing.

Happy to join a design thread on CAMEL-24308 / CAMEL-23853 and adjust scope accordingly. No push from my side to merge before that.

gnodet (commented):
Prior review findings (host/port defaults, listener append, triple-redundant config, JSON-RPC error codes, server version) are all addressed — good improvement. The remaining concern is the design alignment with CAMEL-24308, which committers have already flagged.

**Observations:**

1. **CAMEL-24308 namespace alignment** — The `camel.mcp.*` configuration namespace introduced by this PR will conflict with CAMEL-24308's `camel.server.mcp-*` properties. Committers @Croway and @jamesnetherton have raised this concern, and the contributor agreed to hold merge pending alignment. This should be resolved before proceeding.

2. **Minor: JUnit assertions in test** — `McpConfigurationPropertiesTest` uses JUnit assertions (`assertEquals`, `assertNull`) while the other two new test files in this PR correctly use AssertJ. Per project conventions, new test code should use AssertJ.

**Positive notes:**

- The overall architecture is clean: `EmbeddedMcpRequestHandler` as SPI in `camel-main`, `StreamableMcpJsonRpcEngine` as a reusable JSON-RPC 2.0 engine, and `JbangEmbeddedMcpRequestHandler` bridging to the JBang `ToolRegistry`.
- The `insecure:dev` security annotation on `camel.mcp.enabled` follows established patterns for dev-only management features. The MCP endpoint correctly sits on the management HTTP server.
- The Origin header CSRF check provides appropriate browser-origin protection for a dev-only feature.
- Test coverage via `StreamableMcpJsonRpcEngineTest` (9 tests covering error paths, CSRF, malformed JSON, etc.) is solid.

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### Review follow-up (`bf7e363a68e`)

**@gnodet** — addressed the remaining actionable review item:

- **`McpConfigurationPropertiesTest`** migrated from JUnit to **AssertJ** (added `assertj-core` test dependency to `camel-main`)
- Expanded property coverage: defaults (`enabled`, `path`, unset host/port), fluent `with*` methods, and setters
- **`StreamableMcpJsonRpcEngineTest`**: added tests for non-JSON `Content-Type` (415) and blank request body (-32600)
- **`RunMcpOptionTest`**: added `--mcp` defaults-to-false test

**Design hold (unchanged):** per @jamesnetherton / @Croway / prior thread, merge remains on hold pending CAMEL-24308 alignment on `camel.mcp.*` vs `camel.server.mcp-*` namespace.

**Local tests (all green):**
- `McpConfigurationPropertiesTest` (3)
- `StreamableMcpJsonRpcEngineTest` (11)
- `RunMcpOptionTest` (2)
- `EmbeddedMcpHttpServerTest` (1)

gnodet (commented):
Re-review after latest commit (`bf7e363`). The AssertJ migration and additional test coverage look good. Two items remain:

**1. CAMEL-24308 namespace alignment (high — design hold)**

The `camel.mcp.*` configuration namespace introduced here will conflict with CAMEL-24308's planned `camel.server.mcp-*` properties. Both @Croway and @jamesnetherton flagged this, and the contributor agreed to hold merge pending alignment. No code changes since the prior review address this — it requires a design decision from committers on whether to unify the namespaces or clearly scope this PR's properties as dev-only. Noting this for visibility since the PR remains open (not draft) despite the hold agreement.

**2. Test resource leak in `EmbeddedMcpHttpServerTest` (low)**

`mcpEndpointDelegatesToHandler` starts a `CamelContext` + Vert.x `ManagementHttpServer` but calls `camelContext.stop()` without a try-finally block. If any assertion or HTTP call fails before reaching stop, the server and context leak for the remainder of the test JVM. Since this test doesn't extend `CamelTestSupport`, there's no automatic lifecycle cleanup. Wrapping in try-finally would be a straightforward fix.

**Positive notes:**
- Latest commit successfully migrates tests to AssertJ and adds good coverage (fluent setters, default values, 415 content-type rejection, empty body handling, CLI default-to-false)
- Security model is solid: `@Metadata(security = "insecure:dev")` on enable flags, management-server-only exposure, localhost binding, Origin header CSRF protection, Content-Type validation, proper JSON-RPC error codes

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### Review follow-up (`552be2036eb`)

**@gnodet** — addressed the remaining actionable code review item from the latest re-review:

- **`EmbeddedMcpHttpServerTest`**: wrapped HTTP assertions in `try/finally` so `camelContext.stop()` runs even when an assertion or HTTP call fails (avoids leaking Vert.x/management server for the rest of the test JVM).

Prior commits already covered AssertJ migration, expanded MCP test coverage, host/port default fix, JSON-RPC error codes, and Run.java config simplification.

**Design hold (unchanged):** per @jamesnetherton / @Croway / prior thread, merge remains on hold pending CAMEL-24308 alignment on `camel.mcp.*` vs `camel.server.mcp-*`. No code change for that until committers decide on namespace/scope.

**Local tests (all green — 17 total):**
- `McpConfigurationPropertiesTest` (3)
- `StreamableMcpJsonRpcEngineTest` (11)
- `RunMcpOptionTest` (2)
- `EmbeddedMcpHttpServerTest` (1)

gnodet (commented):
Re-review after latest commits. Both code-level items from the prior review are now addressed:

- ✅ **Test resource leak fixed** — `EmbeddedMcpHttpServerTest.mcpEndpointDelegatesToHandler` now wraps the test body in try-finally, ensuring `camelContext.stop()` runs even when assertions fail (`552be2036eb`)
- ✅ **AssertJ migration complete** — `McpConfigurationPropertiesTest` fully migrated with expanded coverage (defaults, fluent setters, content-type rejection, empty body handling)

The implementation is well-structured with solid security: `@Metadata(security = "insecure:dev")` on enable flags, management-server-only exposure, localhost binding, CSRF protection, Content-Type validation, and proper JSON-RPC error codes.

**Remaining:** The CAMEL-24308 namespace alignment (`camel.mcp.*` vs `camel.server.mcp-*`) is the sole remaining concern, pending committer design decision.

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### Review follow-up (`fa49389cde2`)

Verified on branch `CAMEL-23853-jbang-run-mcp` — all **actionable code review items from @gnodet are already addressed** in prior commits (`552be2036eb` and earlier):

- MCP host/port null defaults (no silent `0.0.0.0` → `127.0.0.1` override)
- `Run.java` config simplified; listener append pattern in `KameletMain`
- JSON-RPC `-32700` / `-32602` error codes; Camel version in `initialize`
- AssertJ migration + expanded test coverage
- `EmbeddedMcpHttpServerTest` try/finally cleanup

**This push** adds two JSON-RPC edge-case tests:
- unknown method → `-32601`
- `ping` → empty result object

**Local tests (19 total — all green):**
- `McpConfigurationPropertiesTest` (3)
- `StreamableMcpJsonRpcEngineTest` (13)
- `RunMcpOptionTest` (2)
- `EmbeddedMcpHttpServerTest` (1)

**Design hold (unchanged):** per @jamesnetherton / @Croway / prior thread, merge remains on hold pending CAMEL-24308 alignment on `camel.mcp.*` vs `camel.server.mcp-*`. No further code changes until committers decide on namespace/scope.

Croway:
Now that CAMEL-24308/24310/24311 are merged, this PR can be simplified considerably while keeping its scope (dev/diagnostics tools on the management server — a valid, complementary use case to the `ai-tool` business tools on the main server):

1. **Reuse `VertxMcpServerEngine` from `camel-mcp-server` instead of the hand-rolled JSON-RPC engine.** The `McpServerEngine` SPI is tool-source-agnostic: `McpServerTool` is just name/description/handler and accepts a raw `inputSchemaJson`, so the `ToolMcpSchemas` output maps 1:1, and `McpToolCallResult(text, isError)` is the same shape as this PR's `ToolCallResult`. The JBang side would build `McpServerTool`s from the `ToolDescriptor`s and call `initialize(...)`/`toolAdded(...)` directly — no bridge needed. This deletes `StreamableMcpJsonRpcEngine`, `ToolMcpSchemas`, the `EmbeddedMcpRequestHandler` SPI and the `ManagementHttpServer` dispatch code (~450 lines), and upgrades the endpoint from a POST-only, session-less subset to the full SDK-backed streamable transport (sessions, SSE, GET replay, DELETE, `tools/list_changed`). It also fixes an interop issue: rejecting every request carrying an `Origin` header with 403 breaks legitimate clients — the spec asks to *validate* Origin, not refuse it.

2. **Drop the `camel.mcp.*` namespace — this PR already contains its replacement.** The `mcpEnabled`/`mcpPath` fields added to `HttpManagementServerConfigurationProperties` are reachable as `camel.management.mcpEnabled` / `camel.management.mcpPath`, which is the right home for a management-server feature and symmetric with the merged `camel.server.mcp-*`. That removes `McpConfigurationProperties`, its generated configurer/metadata, `mainConfigurationProperties.mcp()` and the `setMcpProperties` cross-wiring in `BaseMainSupport` (where the host/port default issues from earlier review rounds kept surfacing). The `--mcp` flag then just sets `camel.management.mcpEnabled=true` plus management-server enablement, like the other management flags in `Run.java`.

3. With 1+2, **core `camel-main` stays MCP-free** (only the existing `McpServerFactory` hook), and everything new lives in camel-jbang / camel-platform-http-main, with `camel-jbang-core` depending on `camel-mcp-server`.

4. **Docs**: after this lands there are two MCP surfaces — dev tools at `<management-port>/mcp` (`--mcp`) and business `ai-tool` routes at `<server-port>/mcp` (`camel.server.mcp-enabled=true`, already property-only under JBang). Worth stating the distinction and cross-linking `mcp-server.adoc`.

One enabler is needed on the camel-mcp-server side: `VertxMcpServerEngine` currently prefers the main server's router when looking up `VertxPlatformHttpRouter`. `ManagementHttpServer` already binds its router (with `SERVER_TYPE_MANAGEMENT`) in the registry, so a small enhancement to make the target server type selectable unblocks this — I can prepare that as a prep PR.

_Claude Code on behalf of Federico Mariani (@Croway)_

Croway:
The enabler mentioned above is now up: #25329 (CAMEL-24353) makes the `VertxMcpServerEngine` target server type selectable, so this PR can register on the management router and drop the hand-rolled protocol code.

_Claude Code on behalf of Federico Mariani (@Croway)_

atiaomar1978-hub:
@Croway @davsclaus Please check the latest update that I did following your comments. 
re-review again.

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### Re-review follow-up — Croway refactor (commits `4ac36e7`, `00b8a873`)

Thanks @Croway for the detailed guidance and for landing the enabler in #25329 (CAMEL-24353). This push implements all four points from your review comment.

---

#### 1. Reuse `VertxMcpServerEngine` instead of hand-rolled JSON-RPC

**Removed (~450 lines):**
- `StreamableMcpJsonRpcEngine`
- `EmbeddedMcpRequestHandler` SPI + `JbangEmbeddedMcpRequestHandler`
- `ManagementHttpServer` MCP dispatch wiring
- `McpConfigurationProperties` and related generated configurer/metadata

**Added:**
- `JbangDevMcpServer` — builds `McpServerTool`s from `ToolRegistry` / `ToolDescriptor`, delegates to `VertxMcpServerEngine`
- `ToolMcpSchemas` — maps `ToolDescriptor` params → JSON Schema for `inputSchemaJson()` (1:1 with prior bridge)
- `JbangDevMcpMainListener` — registers the dev MCP service after Main configure

The endpoint now gets the full SDK-backed streamable transport (sessions, SSE, GET replay, DELETE, `tools/list_changed`) instead of the POST-only subset.

---

#### 2. Drop `camel.mcp.*` — use `camel.management.mcp*` instead

| Before | After |
|--------|-------|
| `camel.mcp.enabled` / `host` / `port` / `path` | **removed** |
| — | `camel.management.mcpEnabled` / `camel.management.mcpPath` (on `HttpManagementServerConfigurationProperties`) |

- `--mcp` / `camel.jbang.mcp` now sets `camel.management.mcpEnabled=true` plus management-server enablement (same pattern as other management flags in `Run.java`)
- `BaseMainSupport#setMcpProperties` cross-wiring removed — core `camel-main` stays MCP-free (only the existing `McpServerFactory` hook)
- Business MCP on the main server remains under `camel.server.mcp-*` (CAMEL-24308)

---

#### 3. Module boundaries

- New code lives in **`camel-jbang-core`** + existing **`camel-platform-http-main`** management config
- `camel-jbang-core` depends on **`camel-mcp-server`**
- No MCP protocol code left in `camel-main` beyond management config properties

---

#### 4. Docs

Updated:
- `camel-jbang-run.adoc`, `camel-jbang-dev.adoc`, `camel-jbang-debug.adoc`
- `camel-jbang-configuration.adoc`
- `camel-4x-upgrade-guide-4_22.adoc` — documents the two MCP surfaces and cross-links `mcp-server.adoc`:
  - **Dev/diagnostics tools** — `<management-port>/mcp` via `--mcp` / `camel.management.mcpEnabled`
  - **Business `ai-tool` routes** — `<server-port>/mcp` via `camel.server.mcp-enabled=true`

---

#### Bug fix: shared-port management fallback (`00b8a873`)

When the management server reuses the main HTTP port, only a `server`-typed `VertxPlatformHttpRouter` is registered. `JbangDevMcpServer#resolveTargetServerType()` now prefers `management` when present, otherwise falls back to `server` — so MCP still mounts when management shares the main port.

---

#### Tests

| Test | Coverage |
|------|----------|
| `RunMcpOptionTest` | `--mcp` flag → `camel.management.mcpEnabled` + management enablement |
| `JbangDevMcpServerTest` | Tool mapping, `resolveTargetServerType()` fallback |
| `ToolMcpSchemasTest` | JSON Schema generation from tool params |

---

#### Dependency on #25329

This branch uses `VertxMcpServerEngine#setTargetServerType(...)`. Once #25329 is merged, this PR should rebase cleanly onto `main`. The management-router registration path is already wired.

Ready for another look when you have a moment. Happy to adjust anything that still feels off scope-wise.

dsl/camel-jbang/camel-jbang-core/pom.xml
```diff
@@ -61,6 +61,10 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-kamelet-main</artifactId>
         </dependency>
+        <dependency>
```
davsclaus: This should not be a compile dependency. Other optional features (`camel-opentelemetry2`, `camel-jfr`, etc.) are downloaded on demand via `dependencies.add("camel:...")` in `Run.java` — you already do this correctly for MCP. Having it as compile scope means every `camel run foo.yaml` pulls in the MCP server, Vert.x MCP transport, and MCP Java SDK even when `--mcp` is not used.

Remove this compile dependency and make the MCP classes pluggable — loaded at runtime only when `camel-mcp-server` is on the classpath (downloaded on demand when `--mcp` is passed).

docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
```diff
@@ -80,6 +80,12 @@ and spec stay together. `--openapi-ui` also sets override properties for `camel.
...
 or other property sources (same as other CLI `--` server flags that use override properties).
 
+`camel run`, `camel dev`, and `camel debug` accept `--mcp` to expose JBang dev/diagnostics tools over MCP
+streamable HTTP on the management server (`/mcp` by default). The feature is gated as developer-only
+(`insecure:dev`) via `camel.jbang.mcp` and `camel.management.mcpEnabled`. It uses `VertxMcpServerEngine`
+from `camel-mcp-server` on the management HTTP router — distinct from business `ai-tool` MCP exposed on the
+main server via `camel.server.mcpEnabled`.
+
 The `camel cmd route-diagram` and `camel cmd route-topology` now accept one or more Camel route source files
 (instead of only the name/pid of a running integration), so diagrams and inter-route topology can be
```
davsclaus, line 84: The upgrade guide is for migration content only (changed defaults, removed options, breaking changes). This entry describes a new feature — it belongs in the command documentation pages where `--mcp` is already documented. Please remove this section from the upgrade guide.

dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/ToolMcpSchemasTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ToolMcpSchemasTest {
+
+    @Test
+    void buildsRequiredParametersSchema() {
+        String schema = ToolMcpSchemas.inputSchemaJson(
+                org.apache.camel.dsl.jbang.core.commands.ai.ToolDescriptor.tool("demo", "Demo tool")
+                        .param("name", "string", "A name", true)
```
davsclaus: Per project import style conventions: `ToolDescriptor` is already imported at line 20 — use the simple name here instead of the FQCN.

```suggestion
                ToolDescriptor.tool("demo", "Demo tool")
```

davsclaus (changes requested):

## Review: CAMEL-23853 — Embed MCP server on camel run/dev and Camel Main

The reworked PR is much cleaner — good simplification using `VertxMcpServerEngine` directly instead of a custom JSON-RPC engine. The architecture (listener + service + schema mapper) is well-separated, security gating via `insecure:dev` is correct, and the test coverage is solid.

Three items to address:

### 1. `camel-mcp-server` must not be a compile dependency on `camel-jbang-core` (blocking)

`camel-mcp-server` is added as a compile (default scope) dependency. This means every `camel run foo.yaml` pulls in the MCP server, Vert.x MCP transport, and the MCP Java SDK — even when `--mcp` is not used.

Other optional features (`camel-opentelemetry2`, `camel-jfr`, `camel-observability-services`) are NOT compile dependencies of `camel-jbang-core`. They are downloaded on demand via `dependencies.add("camel:...")` in `Run.java`, which you already do correctly. The `mainListenerClasses` property is resolved by class name string at runtime, so `JbangDevMcpMainListener` does not need to be compiled against.

Make `camel-mcp-server` pluggable: remove it as a compile dependency from `camel-jbang-core/pom.xml`, and move the `mcp` package classes (`JbangDevMcpMainListener`, `JbangDevMcpServer`, `ToolMcpSchemas`) so they are only resolved when `camel-mcp-server` is on the classpath (downloaded on demand). The runtime download in `Run.java` already handles fetching the JAR when `--mcp` is used.

### 2. Upgrade guide entry is for a new feature, not migration (minor)

The upgrade guide (`camel-4x-upgrade-guide-4_22.adoc`) describes how to use `--mcp`. Per project conventions, the upgrade guide is for migration content only — changed defaults, removed options, breaking changes. New features belong in the command documentation pages, where the `--mcp` flag is already documented.

### 3. FQCN used inline in `ToolMcpSchemasTest` (minor)

`ToolMcpSchemasTest` imports `ToolDescriptor` at the top but then uses the fully qualified name inline: `org.apache.camel.dsl.jbang.core.commands.ai.ToolDescriptor.tool(...)`. Per project conventions, use the simple name since the import is already present.

---

**Positive notes:**
- Clean architecture: `JbangDevMcpMainListener` (lifecycle), `JbangDevMcpServer` (service with proper `doStart`/`doStop`), `ToolMcpSchemas` (schema mapping)
- Correct `insecure:dev` security gating on both `camel.management.mcpEnabled` and `camel.jbang.mcp`
- Management-server-only exposure with smart router type resolution fallback
- Good test coverage with proper try-finally cleanup and AssertJ assertions
- Clear distinction from business `camel.server.mcpEnabled` documented in the upgrade guide

This review does not replace specialized AI review tools or static analysis. _This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

davsclaus:

To clarify finding #1 from the review: the three new classes in `dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/mcp/` (`JbangDevMcpMainListener`, `JbangDevMcpServer`, `ToolMcpSchemas`) are Camel runtime services — they extend `ServiceSupport`, hook into the CamelContext lifecycle, and wire up Vert.x routers. That kind of code does not belong in `camel-jbang-core`, which is for CLI commands and tooling.

These classes should be moved to `camel-kamelet-main` (or into `camel-mcp-server` itself) where other runtime services live, and loaded dynamically when the JAR is downloaded on demand. The `mainListenerClasses` property is already string-based, so `KameletMain` just needs the FQCN — no compile-time dependency required in `camel-jbang-core`.

dsl/camel-jbang/camel-jbang-core/pom.xml
atiaomar1978-hub: Fixed in 7f461878c39: removed camel-mcp-server compile dependency from camel-jbang-core. JBang dev MCP classes now live in camel-mcp-server/jbang and load on demand when Run.java downloads camel:mcp-server via --mcp.

_AI-generated reply on behalf of atiaomar1978-hub_

docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
atiaomar1978-hub: Fixed in 7f461878c39: removed the --mcp section from camel-4x-upgrade-guide-4_22.adoc. The feature remains documented on the command pages where --mcp is already described.

_AI-generated reply on behalf of atiaomar1978-hub_

atiaomar1978-hub:
## Review feedback addressed (`7f461878c39`)

_AI-generated summary on behalf of atiaomar1978-hub_

Replied inline to @davsclaus's three review comments. Summary:

| Review item | Fix |
|-------------|-----|
| **Compile dependency on `camel-mcp-server`** | Removed from `camel-jbang-core`. Dev MCP classes moved to `camel-mcp-server` (`org.apache.camel.component.mcp.server.jbang.*`), loaded when `Run.java` downloads `camel:mcp-server` on `--mcp`. `ToolRegistry` accessed reflectively to avoid Maven reactor cycle. |
| **Upgrade guide `--mcp` section** | Removed from `camel-4x-upgrade-guide-4_22.adoc` (migration-only doc). |
| **FQCN in `ToolMcpSchemasTest`** | Test removed; schema coverage in `JbangDevMcpServerTest` with proper imports. |

### Tests

```
mvn install -pl components/camel-ai/camel-mcp-server -DskipTests
mvn test -pl dsl/camel-jbang/camel-jbang-core -Dtest=RunMcpOptionTest,JbangDevMcpServerTest
```

4 tests passing.

**Note:** CAMEL-24308 namespace alignment (`camel.mcp.*` vs `camel.server.mcp-*`) remains pending committer design decision as previously agreed.

dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/ToolMcpSchemasTest.java
atiaomar1978-hub: Fixed in 7f461878c39: ToolMcpSchemasTest removed; schema building is inlined in JbangDevMcpServer (camel-mcp-server/jbang). Parameter schema coverage moved to JbangDevMcpServerTest using simple ToolDescriptor name per import conventions.

_AI-generated reply on behalf of atiaomar1978-hub_

atiaomar1978-hub:
## Grok review follow-up (`$(git rev-parse --short HEAD)`)

Additional fixes from automated review:

- **ClassResolver:** `JbangDevMcpServer` now resolves `ToolRegistry` / `ToolContext` via `camelContext.getClassResolver()` (aligns with on-demand dependency classloading)
- **Error handling:** tool handler unwraps `InvocationTargetException` and always returns `McpToolCallResult` (never throws)
- **Schema test:** `JbangDevMcpServerTest` verifies `select_process` publishes an `inputSchema` with required `name` property

All 4 MCP unit tests passing.

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+        for (Map.Entry<String, Object> entry : args.entrySet()) {
+            if (entry.getValue() != null) {
+                out.put(entry.getKey(), entry.getValue().toString());
+            }
+        }
+        return out;
+    }
+
+    private String resolveTargetServerType() {
```
Croway, line 241: This fallback makes the shared-port case work, but it silently defeats the `127.0.0.1` intent: both `camel.server` and `camel.management` default to port 8080, and `ManagementHttpServer.doInit` reuses the main `VertxPlatformHttpServer` when the ports match. In that case the `withHost("127.0.0.1")` set by `KameletMain` is ignored (the reused server was built from the main config, host `0.0.0.0`), so `camel run api.yaml --mcp` — any app with a platform-http route — serves the ~40 introspection tools on `0.0.0.0:8080/mcp`.

The dev console has the same trait, so this may be acceptable for an `insecure:dev` flag, but it should not be silent: please log a WARN here when no management-typed router is found and the MCP endpoint falls back to the main server, and mention the shared-port behavior in the `--mcp` option description.

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
```diff
@@ -573,6 +577,39 @@ public HttpManagementServerConfigurationProperties withOpenapiUiSpecPath(String
...
 
+    public boolean isMcpEnabled() {
+        return mcpEnabled;
+    }
+
+    /**
+     * Whether to expose dev/diagnostics MCP tools on this management server (requires camel-mcp-server on the
+     * classpath). Not intended for production use.
+     */
+    public void setMcpEnabled(boolean mcpEnabled) {
```
Croway, line 590: This property is documented in `main.adoc` as a general `camel.management.*` option, but the only reader is `JbangDevMcpMainListener`, which is registered only when `KameletMain` injects it via `mainListenerClasses` for `--mcp`. A plain camel-main user who sets `camel.management.mcpEnabled=true` (with camel-mcp-server on the classpath) gets no endpoint and no error.

Contrast `camel.server.mcp-enabled`, which camel-main wires via the `McpServerFactory` SPI. Either state in the description that this option is currently only honored by Camel JBang (`--mcp`), or wire it in camel-main the same SPI way so it works (or fails meaningfully) everywhere.

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
```diff
@@ -0,0 +1,239 @@
...
+        Class<?> contextClass = camelContext.getClassResolver().resolveClass(TOOL_CONTEXT);
+        Object context = contextClass.getDeclaredConstructor().newInstance();
+        Method selectProcess = contextClass.getMethod("selectProcess", long.class);
+        selectProcess.invoke(context, ProcessHandle.current().pid());
+        return context;
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<Object> allToolDescriptors() throws ReflectiveOperationException {
+        Class<?> registry = camelContext.getClassResolver().resolveClass(TOOL_REGISTRY);
```
Croway: `ClassResolver.resolveClass` returns `null` when the class is missing, so this NPEs on `getMethod` instead of failing with a clear error. `resolveMandatoryClass` throws `ClassNotFoundException` naming the class — much better startup diagnostics if someone enables `camel.management.mcpEnabled` without JBang on the classpath (same in `createToolContext` and `executeTool`).

Croway (commented):
Re-reviewed after `4ac36e7`/`00b8a87`/`7f46187`/`2d65a79` — the refactor looks right: the hand-rolled JSON-RPC engine, the core `EmbeddedMcpRequestHandler` SPI, and the `camel.mcp.*` namespace are gone, `camel-jbang-core` no longer compile-depends on MCP, and the latest commit already fixes tool-error messages and moves to `ClassResolver`. I also verified the non-obvious lifecycle concern: the HTTP servers are registered as early deferred services during autoconfigure, `afterConfigure` adds `JbangDevMcpServer` after them, and early-start preserves insertion order — so the routers are always in the registry before `doStart()` runs. 👍

Two findings worth addressing before merge (inline): the shared-port fallback silently serving the dev tools on the public main server, and `camel.management.mcpEnabled` being inert outside JBang. Minor items also inline.

Other small notes:
- `--mcp` forcing the management host to `127.0.0.1` also rebinds health/metrics when combined with `--observe`/`--health`, which could break container readiness probes — worth a line in the `--mcp` option description.
- The `camel-mcp-server/pom.xml` change is a pure reorder of two test dependencies — can be dropped to keep the diff minimal.

_Claude Code on behalf of Federico Mariani (@Croway)_

dsl/camel-jbang/camel-jbang-core/pom.xml
```diff
@@ -192,6 +192,34 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.modelcontextprotocol.sdk</groupId>
+            <artifactId>mcp-json-jackson2</artifactId>
+            <version>${mcp-java-sdk-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.networknt</groupId>
+            <artifactId>json-schema-validator</artifactId>
```
Croway, line 220: `json-schema-validator` is not used by either new test (leftover from the removed `ToolMcpSchemasTest`?) — can be dropped.

dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/JbangDevMcpServerTest.java
```diff
@@ -0,0 +1,109 @@
...
+        } finally {
+            if (client != null) {
+                client.closeGracefully();
+            }
+            camelContext.stop();
+        }
+    }
+
+    @Test
+    void buildsInputSchemaForParameterizedTools() {
```
Croway: Nit: this test exercises the `ToolDescriptor` builder (which lives elsewhere), not the schema JSON built in `JbangDevMcpServer`. Now that the HTTP test asserts `inputSchema` via `listTools`, consider strengthening that assertion instead (e.g. also check it contains `"required"` for the mandatory param) and dropping or renaming this one.

Croway (approved):

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
```diff
@@ -573,6 +577,41 @@ public HttpManagementServerConfigurationProperties withOpenapiUiSpecPath(String
...
     }
 
+    public boolean isMcpEnabled() {
+        return mcpEnabled;
+    }
+
+    /**
+     * Whether to expose dev/diagnostics MCP tools on this management server (requires camel-mcp-server on the
+     * classpath). Currently honored only when Camel JBang registers {@code JbangDevMcpMainListener} (for example
+     * {@code camel run --mcp}); plain camel-main users should use {@code camel.server.mcpEnabled} for route-based MCP.
```
davsclaus: This Javadoc references `camel.server.mcpEnabled` for "route-based MCP", but that property does not exist anywhere in the codebase. This will be confusing for users who read this and attempt to use it.

If `camel.server.mcpEnabled` is planned for a follow-up, consider removing the reference until it ships. If it already exists elsewhere, please point me to it.

dsl/camel-jbang/camel-jbang-core/pom.xml
```diff
@@ -192,6 +192,35 @@
...
+            <version>${mcp-java-sdk-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.modelcontextprotocol.sdk</groupId>
+            <artifactId>mcp-json-jackson2</artifactId>
+            <version>${mcp-java-sdk-version}</version>
+            <scope>test</scope>
+        </dependency>
+        <!-- MCP SDK schema validation requires networknt 2.x; wiremock pulls 1.5.x without Dialects -->
```
davsclaus: The comment says "requires networknt 2.x" but `${networknt-json-schema-validator-version}` resolves to `1.5.9` (a 1.x version). If 2.x is actually required by the MCP SDK, the version override is insufficient. If 1.5.x works fine, the comment is misleading — consider clarifying.

davsclaus (commented):
## Review: CAMEL-23853 — Embed MCP server on camel run/dev and Camel Main

Thank you for this contribution — the overall architecture is sound: reflective bridge from `camel-mcp-server` to JBang's `ToolRegistry`, proper `127.0.0.1` binding default, correct `security: "insecure:dev"` annotation consistent with other dev features, and a nice integration test covering MCP protocol handshake, tool listing, schema validation, and tool execution.

A few issues to address before this can be merged:

### 1. Inconsistent `mcpEnabled` description across generated metadata (Medium)

The **catalog** copy (`catalog/camel-catalog/src/generated/resources/.../camel-main-configuration-metadata.json`) has a short description:
> "Whether to expose dev/diagnostics MCP tools on this management server (requires camel-mcp-server on the classpath). Not intended for production use."

But the **core** copy (`core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json`) and the AsciiDoc docs have the longer description mentioning `JbangDevMcpMainListener` and `camel.server.mcpEnabled`. These should be identical — please regenerate consistently.

### 2. Reference to non-existent `camel.server.mcpEnabled` (Medium)

See inline comment. The setter Javadoc references `camel.server.mcpEnabled` which does not exist in the codebase. If it is planned for a follow-up PR, the current description should not reference it yet.

### 3. Missing upgrade guide entry (Medium)

The PR description states the feature is documented in the 4.22 upgrade guide, but the diff contains no changes to any `camel-4x-upgrade-guide-*.adoc` file. New configuration properties (`camel.management.mcpEnabled`, `camel.management.mcpPath`, `camel.jbang.mcp`) and CLI flags (`--mcp`) should be noted in the upgrade guide.

### 4. `{code ...}` in generated docs (Low)

The Javadoc `{@code JbangDevMcpMainListener}` and `{@code camel run --mcp}` are rendered as literal `{code ...}` (without the `@`) in the generated AsciiDoc and JSON metadata. Consider using the `@Metadata(description=...)` annotation directly on the field with the full description text to avoid this.

---

**Note:** This review evaluates the PR against project conventions and standards. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

davsclaus:
in camel-kamelet-main you can add the classes you load dynamic to a known list of classes that maps to maven dependency then it will automatic download the needed JAR.
https://github.com/apache/camel/blob/main/dsl/camel-kamelet-main/src/main/resources/camel-main-known-dependencies.properties

atiaomar1978-hub:
Addressed remaining review feedback and automated review findings in commit `6f857dc1902`:

**@davsclaus**
- `camel.server.mcpEnabled` **does exist** on `HttpServerConfigurationProperties` (used by `BaseMainSupport.setupMcpServer`); javadoc/docs now say "configure camel.server.mcpEnabled instead"
- Fixed networknt pom comment: property resolves to **2.0.1** (verified via dependency tree)

**@Croway** (replies on each thread below)

**Grok/Bugbot fixes**
- Download `camel:mcp-server` when profile `camel.jbang.mcp=true`, not only `--mcp` CLI flag
- Added `JbangDevMcpMainListener` to `camel-main-known-dependencies.properties`
- Strengthened `inputSchema` test to assert `required` field
- Documented 127.0.0.1 management rebind in `--mcp` description

Tests: `RunMcpOptionTest` (2) + `JbangDevMcpServerTest` (1) — all passing.

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
atiaomar1978-hub: Fixed in cc569af443f (also in 6f857dc1902): added LOG.warn when no management HTTP router is found so shared-port fallback is visible in logs.

_AI-generated reply on behalf of atiaomar1978-hub_

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
atiaomar1978-hub: Fixed in cc569af443f (6f857dc1902): javadoc and main.adoc clarify camel.management.mcpEnabled is only honored when Camel JBang registers JbangDevMcpMainListener (camel run --mcp); route-based MCP uses camel.server.mcpEnabled on HttpServerConfigurationProperties.

_AI-generated reply on behalf of atiaomar1978-hub_

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
atiaomar1978-hub: Fixed in cc569af443f: switched to resolveMandatoryClass via mandatoryClass() helper for ToolRegistry/ToolContext reflective access.

_AI-generated reply on behalf of atiaomar1978-hub_

dsl/camel-jbang/camel-jbang-core/pom.xml
atiaomar1978-hub: Kept json-schema-validator test dependency (6f857dc1902 clarifies comment): parent property resolves to 2.0.1 which provides Dialects required by MCP SDK; wiremock transitively pulls 1.5.x without it. Removing it breaks JbangDevMcpServerTest.

_AI-generated reply on behalf of atiaomar1978-hub_

dsl/camel-jbang/camel-jbang-core/src/test/java/org/apache/camel/dsl/jbang/core/commands/mcp/JbangDevMcpServerTest.java
atiaomar1978-hub: Fixed in cc569af443f (6f857dc1902): removed redundant ToolDescriptor builder test; HTTP integration test now asserts inputSchema contains name and required for select_process.

_AI-generated reply on behalf of atiaomar1978-hub_

core/camel-main/src/main/java/org/apache/camel/main/HttpManagementServerConfigurationProperties.java
atiaomar1978-hub: Addressed in 6f857dc1902: camel.server.mcpEnabled exists on HttpServerConfigurationProperties (see BaseMainSupport.setupMcpServer). Updated javadoc/docs to reference it explicitly instead of implying it is missing.

_AI-generated reply on behalf of atiaomar1978-hub_

dsl/camel-jbang/camel-jbang-core/pom.xml
atiaomar1978-hub: Clarified in 6f857dc1902: the comment now states 2.0.1 explicitly. ${networknt-json-schema-validator-version} in parent pom.xml is 2.0.1 (not 1.5.9); dependency tree confirms 2.0.1:test on camel-jbang-core. Wiremock brings 1.5.x transitively which lacks Dialects.

_AI-generated reply on behalf of atiaomar1978-hub_

atiaomar1978-hub:
Fixed CI failure (uncommitted catalog generated files) in commit `e0375aa6baf`: regenerated `catalog/camel-catalog` `main.adoc` and `camel-main-configuration-metadata.json` to sync `camel.management.mcpEnabled` description with `core/camel-main`.

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/KameletMain.java
```diff
@@ -594,6 +594,19 @@ protected CamelContext createCamelContext() {
             configure().httpManagementServer().withEnabled(true);
             configure().httpManagementServer().withInfoEnabled(true);
         }
+        boolean mcp = "true".equals(getInitialProperties().get(getInstanceType() + ".mcp"));
+        if (mcp) {
+            configure().httpManagementServer().withEnabled(true);
+            configure().httpManagementServer().withMcpEnabled(true);
+            configure().httpManagementServer().withHost("127.0.0.1");
```
davsclaus: This unconditionally overrides the management server host to `127.0.0.1` when MCP is enabled. If a user also uses `--observe` for health/metrics, external probes (e.g. Kubernetes liveness checks) won't be able to reach the management endpoints.

The CLI help documents this side effect, which is good. But is there a way for the user to override this back (e.g. via `camel.management.host` in properties)? If property-file values are applied after this code, it would work — worth confirming the ordering.

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
```diff
@@ -0,0 +1,250 @@
...
+                    message = failure.getClass().getSimpleName();
+                }
+                return new McpToolCallResult(message, true);
+            }
+        };
+        return new McpServerTool() {
+            @Override
+            public String name() {
+                return toolName;
+            }
```
davsclaus, line 127: Minor: `parameters()` returns `Map.of()` while `inputSchemaJson()` is populated from the descriptor's params. This works today because `VertxMcpServerEngine.toolAdded()` only uses `inputSchemaJson()`, but could be a gap for forward compatibility if a future engine implementation relies on `parameters()` instead.

davsclaus (commented):
Thank you for the contribution, @atiaomar1978-hub — this is a well-structured PR that follows existing patterns for similar JBang features (`--console`, `--observe`, `--openapi-ui`). CI is green, security model is correctly applied, and tests cover both CLI parsing and end-to-end integration.

### Findings

**1. [Question] Management server host override side effect**

When `--mcp` is enabled, `KameletMain` unconditionally sets `withHost("127.0.0.1")` on the management server. This is good for security, but it has a side effect: if a user runs `camel run app.yaml --mcp --observe`, the management server (health, metrics, info) will also be bound to `127.0.0.1` instead of `0.0.0.0`. This could break external health probes (e.g., Kubernetes liveness/readiness checks).

The CLI help text documents this, which is good. But is the intent that users who need both `--mcp` and externally-accessible health checks would configure `camel.management.host` separately? If so, does the property file value override the programmatic `withHost()` call?

**2. [Minor] `parameters()` returns empty map in `JbangDevMcpServer.toMcpTool()`**

The anonymous `McpServerTool` implementation returns `Map.of()` from `parameters()` while `inputSchemaJson()` is populated from the descriptor. This works because `VertxMcpServerEngine.toolAdded()` only uses `inputSchemaJson()`, but could be a gap if a future engine uses `parameters()` instead.

### Positive observations

- Security model correctly applied: `security = "insecure:dev"` on `mcpEnabled` in both `HttpManagementServerConfigurationProperties` and `CamelJBangConstants`
- 127.0.0.1 binding by default prevents remote access to dev MCP tools
- Fallback warning in `resolveTargetServerType()` clearly notes security implication
- Tests follow project conventions (package-private, AssertJ, no `Thread.sleep()`)
- `JbangDevMcpServerTest` is a solid integration test covering server info, tool listing, schema structure, and tool execution
- Generated files and documentation are all consistent with source changes

This review covers project rules and conventions. It does not replace specialized review tools (CodeRabbit, SonarCloud) or static analysis.

atiaomar1978-hub:
Addressed the latest review feedback in cc4679e049c:

1. **KameletMain host override** — `127.0.0.1` is now applied only when `camel.management.host` is not already configured, so `--observe` health/metrics probes can still bind externally when the user sets the host explicitly.
2. **`parameters()` consistency** — `JbangDevMcpServer` now builds `ParameterDef` via `AiToolParameterHelper.parseParameterMetadata()` and uses the same defs for both `parameters()` and `inputSchemaJson()`.
3. **`{code ...}` in generated docs** — replaced `{@code ...}` in `setMcpEnabled` javadoc with plain text; regenerated catalog/main metadata (no more `{code camel run --mcp}` in `main.adoc`/JSON).

Tests run:
```
mvn test -pl dsl/camel-jbang/camel-jbang-core,components/camel-ai/camel-mcp-server -Dtest=RunMcpOptionTest,JbangDevMcpServerTest -DskipITs
```

dsl/camel-kamelet-main/src/main/java/org/apache/camel/main/KameletMain.java
atiaomar1978-hub: Fixed in cc4679e049c: `withHost("127.0.0.1")` is now guarded by `!isConfigured("camel.management.host")`, matching the pattern used elsewhere in `KameletMain`. Users combining `--mcp` with `--observe` can set `camel.management.host` explicitly for external probes.

_AI-generated reply on behalf of atiaomar1978-hub_

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/jbang/JbangDevMcpServer.java
atiaomar1978-hub: Fixed in cc4679e049c: descriptor params are converted to a flat metadata map and parsed with `AiToolParameterHelper.parseParameterMetadata()`. Both `parameters()` and `inputSchemaJson()` now share the same `ParameterDef` map via `buildJsonSchemaFromDefs()`.

_AI-generated reply on behalf of atiaomar1978-hub_

davsclaus (approved):
Thank you for this contribution, @atiaomar1978-hub — this is well-structured work that follows established JBang patterns (`--console`, `--observe`, `--openapi-ui`). The architecture is clean, the security model is correctly applied, and the code has been thoroughly refined across multiple review rounds.

### Summary

- Security model correctly applied: `security = "insecure:dev"` on both `camel.jbang.mcp` and `camel.management.mcpEnabled`, management-server-only exposure, `127.0.0.1` binding by default
- Reflective bridge to `ToolRegistry` is architecturally sound — avoids compile dependency from component to CLI module
- `JbangDevMcpServerTest` is a solid integration test covering MCP handshake, tool listing, schema validation, and tool execution
- All prior review findings addressed

### Minor notes (non-blocking)

1. **CI** — needs to be triggered and pass (first-time contributor approval required)
2. **`json-schema-validator` version override** — the 1.5.x → 2.0.1 bump in test scope is explained in the comment; CI will validate no WireMock test breakage
3. **`--mcp` + `--observe` interaction** — management host forced to `127.0.0.1` could affect container health probes; documented in CLI help text, and `camel.management.host` overrides it

davsclaus:
rebase this on main

PR #25205, opened 2026-07-29
CAMEL-24301: Fix RouteService NPE when startup-failure exception carries no message

mayurbm:
> _This PR was created by [Claude Code](https://claude.ai/code) on behalf of mayurbm._

https://issues.apache.org/jira/browse/CAMEL-24301

## Problem

`RouteService.warmUp()` and `setUp()` pass `e.getLocalizedMessage()` directly to
`FailedToStartRouteException`, whose constructor calls `Objects.requireNonNull` on
that argument. When the root cause is a message-less exception (bare `NullPointerException`,
`StackOverflowError`, or a wrapped exception whose inner cause has no text),
`getLocalizedMessage()` returns `null` and a secondary `NullPointerException` is thrown
from inside the constructor instead of the intended `FailedToStartRouteException`.

This was first observed in Camel 3.14.x where the symptom was
`Failed to start route X because of null`; the 4.x code fixed that regression but
introduced this adjacent null-safety gap via `Objects.requireNonNull` on the message arg.

## Fix

Adds a private `extractUsefulMessage(Throwable)` helper in `RouteService` that walks the
cause chain to find the first non-null, non-blank message, falling back to the exception
simple class name. Both `warmUp()` and `setUp()` now call this helper instead of
`e.getLocalizedMessage()` directly.

## Test

`RouteServiceWarmUpNullMessageTest` added to `core/camel-core` covers:
1. A bare `NullPointerException` (no message) on warm-up does not cause a secondary NPE.
2. The resulting exception message never contains `because: null`.
3. When the outer exception has no message but a nested cause does, the cause message
   is surfaced in the `FailedToStartRouteException`.

## Files changed

- `core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java`
- `core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java` (new)

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
@@ -143,13 +143,12 @@ public class RouteService extends ChildServiceSupport {
 
     /**
      * Extracts a non-null, non-empty error message from the exception or its cause chain.
-     * <p>
+     * <p/>
      * {@link Throwable#getLocalizedMessage()} can return {@code null} for exceptions such as
-     * {@link NullPointerException} that carry no message, which would cause
-     * {@link FailedToStartRouteException} to throw {@link NullPointerException} from its own constructor
-     * (via {@code Objects.requireNonNull}) instead of wrapping the original failure. This helper walks the
-     * cause chain to find the first meaningful message and falls back to the simple class name so the
-     * caller always receives a non-null string.
+     * {@link NullPointerException} that carry no message, which would cause {@link FailedToStartRouteException} to
+     * throw {@link NullPointerException} from its own constructor (via {@code Objects.requireNonNull}) instead of
+     * wrapping the original failure. This helper walks the cause chain to find the first meaningful message and falls
+     * back to the simple class name so the caller always receives a non-null string.
      */
     private static String extractUsefulMessage(Throwable e) {
         Throwable current = e;
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
@@ -127,7 +127,7 @@ public class RouteService extends ChildServiceSupport {
         try {
             doWarmUp();
         } catch (Exception e) {
-            throw new FailedToStartRouteException(getId(), getLocation(), e.getLocalizedMessage(), e);
+            throw new FailedToStartRouteException(getId(), getLocation(), extractUsefulMessage(e), e);
         }
     }
 
@@ -136,9 +136,31 @@ public class RouteService extends ChildServiceSupport {
             try {
                 doSetup();
             } catch (Exception e) {
-                throw new FailedToStartRouteException(getId(), getLocation(), e.getLocalizedMessage(), e);
+                throw new FailedToStartRouteException(getId(), getLocation(), extractUsefulMessage(e), e);
             }
         }
+    }
+
+    /**
+     * Extracts a non-null, non-empty error message from the exception or its cause chain.
+     * <p>
+     * {@link Throwable#getLocalizedMessage()} can return {@code null} for exceptions such as
+     * {@link NullPointerException} that carry no message, which would cause
+     * {@link FailedToStartRouteException} to throw {@link NullPointerException} from its own constructor
+     * (via {@code Objects.requireNonNull}) instead of wrapping the original failure. This helper walks the
+     * cause chain to find the first meaningful message and falls back to the simple class name so the
+     * caller always receives a non-null string.
+     */
+    private static String extractUsefulMessage(Throwable e) {
+        Throwable current = e;
+        while (current != null) {
+            String msg = current.getLocalizedMessage();
+            if (msg != null && !msg.isBlank()) {
+                return msg;
+            }
+            current = current.getCause();
+        }
+        return e.getClass().getSimpleName();
     }
 
     public boolean isAutoStartup() {
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
@@ -150,7 +150,7 @@ public class RouteService extends ChildServiceSupport {
      * wrapping the original failure. This helper walks the cause chain to find the first meaningful message and falls
      * back to the simple class name so the caller always receives a non-null string.
      */
-    private static String extractUsefulMessage(Throwable e) {
+    static String extractUsefulMessage(Throwable e) {
         Throwable current = e;
         while (current != null) {
             String msg = current.getLocalizedMessage();
--- a/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
@@ -37,54 +37,59 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
- * Verifies that {@link RouteService#warmUp()} wraps startup failures in a {@link FailedToStartRouteException} whose
- * message is always meaningful — even when the root cause exception itself carries a {@code null} message (e.g. a bare
- * {@link NullPointerException}).
+ * Verifies that {@link RouteService#warmUp()} and {@link RouteService#setUp()} wrap startup failures in a
+ * {@link FailedToStartRouteException} whose message is always meaningful — even when the root cause exception carries a
+ * {@code null} message (e.g. a bare {@link NullPointerException}).
  *
  * <p>
  * Before the fix, {@code RouteService} passed {@code e.getLocalizedMessage()} directly to the
  * {@link FailedToStartRouteException} constructor, which calls {@code Objects.requireNonNull} on that argument. A
  * message-less exception therefore caused a secondary {@link NullPointerException} to be thrown from inside the
  * exception constructor rather than a proper {@link FailedToStartRouteException}.
+ *
+ * <p>
+ * The tests trigger the failure during endpoint initialisation (inside {@code doSetup()}), which is the code path
+ * covered by the {@code RouteService} fix.
  */
 class RouteServiceWarmUpNullMessageTest {
 
     /**
-     * When a route's consumer throws a {@link NullPointerException} with no message during warm-up, the resulting
-     * {@link FailedToStartRouteException} must still carry a non-null, non-empty message.
+     * When the endpoint throws a message-less {@link NullPointerException} during route setup, the result must be a
+     * {@link FailedToStartRouteException}, not a raw NPE.
      */
     @Test
-    void testWarmUpNullMessageExceptionProducesUsefulFailedToStartMessage() throws Exception {
+    void testSetUpNullMessageExceptionProducesFailedToStartRouteException() {
         CamelContext context = new DefaultCamelContext();
         context.addComponent("fail", new NullMessageFailComponent());
-        context.addRoutes(new RouteBuilder() {
-            @Override
-            public void configure() {
-                from("fail:trigger").routeId("test-route").to("direct:out");
-            }
-        });
 
-        assertThatThrownBy(context::start)
-                .isInstanceOf(FailedToStartRouteException.class);
+        assertThatThrownBy(() -> {
+            context.addRoutes(new RouteBuilder() {
+                @Override
+                public void configure() {
+                    from("fail:trigger").routeId("test-route").to("direct:out");
+                }
+            });
+            context.start();
+        }).isInstanceOf(FailedToStartRouteException.class);
     }
 
     /**
-     * The {@link FailedToStartRouteException} message must not be null, must contain the route id, and must not use the
-     * literal string "null" as the failure description.
+     * The {@link FailedToStartRouteException} message must contain the route id and must not use the literal string
+     * "null" as the failure description.
      */
     @Test
-    void testFailedToStartMessageIsNonNullAndMeaningful() throws Exception {
+    void testFailedToStartMessageIsNonNullAndMeaningful() {
         CamelContext context = new DefaultCamelContext();
         context.addComponent("fail", new NullMessageFailComponent());
-        context.addRoutes(new RouteBuilder() {
-            @Override
-            public void configure() {
-                from("fail:trigger").routeId("meaningful-route").to("direct:out");
-            }
-        });
 
         FailedToStartRouteException caught = null;
         try {
+            context.addRoutes(new RouteBuilder() {
+                @Override
+                public void configure() {
+                    from("fail:trigger").routeId("meaningful-route").to("direct:out");
+                }
+            });
             context.start();
         } catch (FailedToStartRouteException e) {
             caught = e;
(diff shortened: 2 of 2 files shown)


core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
apupier, line 1: Given that all lines are marked as modified, i suspect that the end of line encoding has been modified.

mayurbm: please check here, is this fine https://github.com/apache/camel/pull/25205/changes

mayurbm:
JIRA Issue created for the same - https://issues.apache.org/jira/browse/CAMEL-24301

apupier:
some files requires reformat:
```
modified:   core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
modified:   core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
```
you can run mvn clean install on these modules to format them

mayurbm:
> some files requires reformat:
> 
> ```
> modified:   core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
> modified:   core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
> ```
> 
> you can run mvn clean install on these modules to format them

> done

apupier:
compilation failures are reported by the build:

```
[2](https://github.com/apache/camel/actions/runs/30471471386/job/90788006459?pr=25205#step:9:43)
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.15.0:testCompile (default-testCompile) on project camel-core: Compilation failure: Compilation failure: 
Error: [ERROR] /home/runner/work/camel/camel/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java:[143,19] cannot find symbol
[ERROR]   symbol:   class Endpoint
[ERROR]   location: class org.apache.camel.impl.engine.RouteServiceWarmUpNullMessageTest.NullMessageFailComponent
Error: [ERROR] /home/runner/work/camel/camel/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java:[153,20] cannot find symbol
[ERROR]   symbol:   class Consumer
[ERROR]   location: class org.apache.camel.impl.engine.RouteServiceWarmUpNullMessageTest.NullMessageFailComponent.FailEndpoint
Error: [ERROR] /home/runner/work/camel/camel/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java:[190,19] cannot find symbol
[ERROR]   symbol:   class Endpoint
[ERROR]   location: class org.apache.camel.impl.engine.RouteServiceWarmUpNullMessageTest.ChainedNullMessageFailComponent
Error: [ERROR] /home/runner/work/camel/camel/core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java:[203,20] cannot find symbol
[ERROR]   symbol:   class Consumer
[ERROR]   location: class org.apache.camel.impl.engine.RouteServiceWarmUpNullMessageTest.ChainedNullMessageFailComponent.FailEndpoint
```

apupier (dismissed):
there is a test to adapt:

```
[camel-core] [ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.211 s <<< FAILURE! -- in org.apache.camel.impl.engine.DefaultSupervisingRouteControllerTest
[camel-core] [ERROR] org.apache.camel.impl.engine.DefaultSupervisingRouteControllerTest.testSupervising -- Time elapsed: 3.210 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <Cannot start> but was: <Failed to start route: cake because: Cannot start>
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1199)
	at org.apache.camel.impl.engine.DefaultSupervisingRouteControllerTest.testSupervising(DefaultSupervisingRouteControllerTest.java:109)
```

gnodet (changes requested):
## Review Summary

The core fix in `RouteService.java` (replacing `e.getLocalizedMessage()` with `extractUsefulMessage(e)`) is correct and well-motivated. However, the `InternalRouteStartupManager` changes expand scope beyond the NPE fix, introduce a behavioral change to a public SPI, and break tests in two other modules.

### 🔴 High: Broken tests in camel-main and camel-spring-xml

The `InternalRouteStartupManager` changes wrap raw exceptions in `FailedToStartRouteException`, but only the equivalent test in `core/camel-core` was updated. Two other test files still assert on the raw exception type/message and will fail:

1. **`MainSupervisingRouteControllerTest`** (`core/camel-main`) — lines 76-77 (`testMain`) and lines 153-154 (`testMainApplicationProperties`) assert `assertEquals("Cannot start", e.getMessage())` and `assertInstanceOf(IllegalArgumentException.class, e)`. After this PR, `e` will be `FailedToStartRouteException` with a different message.

2. **`SpringSupervisingRouteControllerTest`** (`camel-spring-xml`) — lines 72-73 assert `assertEquals("Cannot start", e.getMessage())` and `assertTrue(e instanceof IllegalArgumentException)`. Same breakage.

### 🔴 High: InternalRouteStartupManager changes exceed JIRA scope

CAMEL-24301 describes a null-safety bug in `RouteService.warmUp()` and `setUp()` where `e.getLocalizedMessage()` returns null and the `FailedToStartRouteException` constructor NPEs. The `InternalRouteStartupManager` code previously did `throw e` (re-throwing the raw exception) — it never constructed a `FailedToStartRouteException`, so it had no null-message NPE risk.

The PR introduces `FailedToStartRouteException` wrapping here for the first time, changing what `SupervisingRouteController.getRestartException()` returns. This is a public SPI method in `org.apache.camel.spi` — downstream consumers (health checks, consoles, exception handlers) that rely on the raw exception type will observe a different type.

**Recommendation**: Consider splitting the PR. The `RouteService.java` fix is minimal, correct, and ready to merge. The `InternalRouteStartupManager` wrapping is a separate behavioral change that deserves its own JIRA ticket with complete test updates.

### 🟡 Medium: `extractUsefulMessage` duplication

The `extractUsefulMessage(Throwable)` helper is copy-pasted identically in both `RouteService` and `InternalRouteStartupManager`. Consider extracting to a shared location (e.g., a static method in `FailedToStartRouteException`, or in an existing utility class).

### 🟡 Medium: `testWarmUpWalksCauseChainForMessage` doesn't test chain walking

The test throws `new RuntimeException(new IllegalStateException(causeMessage))`, but `RuntimeException(Throwable)` sets `detailMessage` to `cause.toString()` (non-null). So `extractUsefulMessage()` returns the outer message on the first iteration without ever walking the chain. To properly test chain walking, the outer exception needs a null or blank message (e.g., a bare `NullPointerException` with `initCause`).

### 🔵 Low: Test conventions

- New test class and methods use `public` modifier — should be package-private (JUnit 5)
- JUnit assertions used instead of project-preferred AssertJ

### Positive notes

- The `extractUsefulMessage` helper design (walking the cause chain, falling back to class simple name) is well-thought-out
- The previous reviewer's EOL encoding concern appears resolved

---

gnodet (approved):

Thank you for iterating on the feedback -- this revision is much improved.

### What was addressed

- **Scope reduced to match JIRA** -- The `InternalRouteStartupManager` changes have been removed. The PR now only modifies `RouteService.java`, which is exactly where the null-message NPE occurs. This resolves the broken tests in `camel-main`, `camel-spring-xml`, and `DefaultSupervisingRouteControllerTest`.
- **No more code duplication** -- `extractUsefulMessage` exists only in `RouteService`.
- **Test conventions** -- Test class and methods are package-private (no `public`), and assertions use AssertJ (`assertThat`, `assertThatThrownBy`).
- **EOL encoding** -- The diff now only shows changed lines.

### Fix assessment

The fix itself is minimal and correct:
- `extractUsefulMessage(Throwable)` walks the cause chain for the first non-null, non-blank message and falls back to `e.getClass().getSimpleName()`
- Both `warmUp()` and `setUp()` call this helper instead of `e.getLocalizedMessage()` directly
- The Javadoc on the helper method clearly explains the rationale
- The other call sites of `FailedToStartRouteException` in `InternalRouteStartupManager` and `DefaultRoutesLoader` all use hardcoded string messages (not `e.getLocalizedMessage()`), so they are not affected

### Remaining minor suggestion

**`testSetUpWalksCauseChainForMessage` does not truly exercise chain walking** (carried forward from the previous review). The endpoint throws `new RuntimeException(new IllegalStateException(causeMessage))`, but `RuntimeException(Throwable)` sets `detailMessage` to `cause.toString()` which is non-null. So `extractUsefulMessage` returns the outer message on the first iteration without ever walking the chain.

To properly exercise the chain-walking logic, the outer exception needs a truly null message:
```java
@Override
protected void doStart() {
    NullPointerException outer = new NullPointerException();
    outer.initCause(new IllegalStateException(causeMessage));
    throw outer;
}
```

This is a nice-to-have improvement rather than a blocker -- the chain-walking logic is simple and correct by inspection, and the primary fix path (bare `NullPointerException` with no message) is well-tested by the first two tests.

### Note

CI has not reported any checks on this branch yet. Please ensure CI passes before merging.

davsclaus (approved):
The fix is correct, minimal, and well-scoped to CAMEL-24301.

**Fix:** `extractUsefulMessage(Throwable)` walks the cause chain for the first non-null, non-blank message and falls back to `e.getClass().getSimpleName()` — ensuring `FailedToStartRouteException` never receives a null message argument (which would NPE in its `Objects.requireNonNull` constructor call).

**Tests:** Three tests cover bare NPE, meaningful message content, and cause-chain walking. All follow project conventions (package-private, AssertJ, no `Thread.sleep`).

**Note on the chain-walking test:** The carried-forward comment from the previous review round about `testSetUpWalksCauseChainForMessage` not exercising chain walking appears to be stale — the current revision correctly uses `NullPointerException()` + `initCause()` (outer message is genuinely null), so `extractUsefulMessage` does walk the chain.

**Minor (non-blocking):**
- Tests 2 and 3 use manual try-catch with cause-chain walking while test 1 uses `assertThatThrownBy` — could be simplified for consistency.
- Test 1 doesn't call `context.stop()` in a finally block while tests 2 and 3 do.

CI is pending — wait for green before merging.

mayurbm:
All CI checks are now green. All review feedback from @gnodet and @apupier has been addressed:

- Removed `InternalRouteStartupManager` changes entirely — scope is now strictly limited to `RouteService.java` as requested
- Fixed `testSetUpWalksCauseChainForMessage` to use `NullPointerException` + `initCause()` so the outer exception truly has a null message, properly exercising the cause-chain walking in `extractUsefulMessage`
- Switched to AssertJ assertions
- Dropped `public` from test class and methods (JUnit 5 convention)
- Replaced FQCNs with proper imports
- Applied `mvn formatter:format impsort:sort` — EOL encoding normalised
- `DefaultSupervisingRouteControllerTest` passes (was broken only by the IRSM changes, which are now removed)

Could you please take another look and approve if everything looks good? @gnodet @apupier

_Claude Code on behalf of @mayurbm_

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
```diff
@@ -136,11 +136,32 @@ public void setUp() throws FailedToStartRouteException {
...
+     * <p/>
+     * {@link Throwable#getLocalizedMessage()} can return {@code null} for exceptions such as
+     * {@link NullPointerException} that carry no message, which would cause {@link FailedToStartRouteException} to
+     * throw {@link NullPointerException} from its own constructor (via {@code Objects.requireNonNull}) instead of
+     * wrapping the original failure. This helper walks the cause chain to find the first meaningful message and falls
+     * back to the simple class name so the caller always receives a non-null string.
+     */
+    private static String extractUsefulMessage(Throwable e) {
+        Throwable current = e;
+        while (current != null) {
```
davsclaus, line 155: Consider making this `static` package-private (`static String extractUsefulMessage(...)`) instead of `private` so that `DefaultSupervisingRouteController` (line 521) — which has the same null-safety bug with `e.getMessage()` — can reuse it. Alternatively, fix both call sites independently in a follow-up.

davsclaus (commented):

Good fix for a real null-safety bug. The code change is minimal, well-scoped, and the tests cover the key scenarios. CI is green.

**One related finding outside this diff:**

`DefaultSupervisingRouteController.java:521` has the identical null-safety gap:
```java
throw new FailedToStartRouteException(route, e.getMessage(), e);
```
`e.getMessage()` also returns `null` for message-less exceptions, hitting the same `Objects.requireNonNull` in the 3-arg constructor. Consider either making `extractUsefulMessage` package-private for reuse there, or fixing it in a follow-up under the same JIRA ticket.

**Nitpick:** Branch name `fix/route-service-null-message-on-startup-failure` — project convention is `fix/<ISSUE_ID>` (i.e. `fix/CAMEL-24301`). Non-blocking.

This review does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
```diff
@@ -0,0 +1,264 @@
...
+                public void configure() {
+                    from("fail:trigger").routeId("test-route").to("direct:out");
+                }
+            });
+            context.start();
+        }).isInstanceOf(FailedToStartRouteException.class);
+    }
+
+    /**
+     * The {@link FailedToStartRouteException} message must contain the route id and must not use the literal string
```
davsclaus, line 83: Minor: unlike tests 2 and 3, this test does not call `context.stop()` after the assertion. If `context.start()` partially initialises resources before failing, they would leak during the test run. Consider wrapping with `try/finally` like the other tests, or using `assertThatThrownBy` inside a try block with cleanup.

core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @Test
+    void testFailedToStartMessageIsNonNullAndMeaningful() {
+        CamelContext context = new DefaultCamelContext();
+        context.addComponent("fail", new NullMessageFailComponent());
+
+        FailedToStartRouteException caught = null;
+        try {
+            context.addRoutes(new RouteBuilder() {
+                @Override
+                public void configure() {
```
davsclaus: Suggestion: since test 1 shows `FailedToStartRouteException` is thrown directly (not wrapped), the cause-chain walking here is unnecessary. This test could use `assertThatThrownBy` consistently with test 1:
```java
assertThatThrownBy(() -> {
    context.addRoutes(...);
    context.start();
}).isInstanceOf(FailedToStartRouteException.class)
  .satisfies(thrown -> assertThat(thrown.getMessage())
      .contains("meaningful-route")
      .doesNotContain("because: null"));
```
Same applies to test 3. Not blocking — just a readability suggestion.

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/RouteService.java
mayurbm: Done — removed the `private` modifier so `extractUsefulMessage` is now package-private `static`. This makes it accessible to `DefaultSupervisingRouteController` (and any other class in the same package) for reuse at the identical null-safety call site on line 521.

core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
mayurbm: Fixed — test 1 is now wrapped in a `try/finally` block that calls `context.stop()` in the `finally` clause, matching the cleanup pattern already used in tests 2 and 3.

core/camel-core/src/test/java/org/apache/camel/impl/engine/RouteServiceWarmUpNullMessageTest.java
mayurbm: Done — both tests 2 and 3 now use `assertThatThrownBy` with fluent assertion chaining (`.isInstanceOf()`, `.hasMessageContaining()`, `.hasMessageNotContaining()`), consistent with test 1. The manual `try/catch` unwrapping and intermediate `caught` variable are removed. Each test is still wrapped in `try/finally` for `context.stop()` cleanup.

mayurbm:
The CI failure is unrelated to the changes in this PR.

**Failing test:** `PulsarCustomMessageReceiptIT.testAcknowledgeWithCustomMessageReceipt` in `camel-pulsar`

**Root cause:** A pre-existing flaky integration test with two issues:
1. A timing race — the test expects `acknowledge()` to be called at least 2 times but only 1 message was acknowledged before the assertion fired.
2. A teardown resource leak — when the test fails, the Pulsar producer is not closed, so all retry attempts crash with `ProducerBusyException: Producer with name 'camel-producer-1' is already connected`.

**Why this is not caused by this PR:** The changes here only touch `core/camel-base-engine/RouteService.java` and `core/camel-core` test sources. The `camel-pulsar` component is completely unaffected. Recent main-branch CI runs (2026-08-03) all pass successfully, confirming this is intermittent infrastructure/timing flakiness.

Could a committer please re-trigger the CI run? Thank you.

PR #25206, opened 2026-07-29
CAMEL-24266: Add volatile to Delayer.delayValue for JMM visibility

gnodet:

## Summary

`Delayer.delayValue` is a plain `long` field written by routing threads (in `calculateDelay()`) and read by JMX threads (via `ManagedDelayer.getDelay()` → `getDelayValue()`). This creates a JMM visibility gap:

- On **32-bit JVMs**, reads of a `long` are non-atomic (JLS 17.7), so JMX could observe a torn value
- On **all JVMs**, there is no happens-before guarantee, so stale reads are possible

This was flagged during review of PR #24985 (CAMEL-24227 volatile sweep), which fixed the forward direction (`delay` Expression written by JMX, read by routing threads) but missed the reverse direction (`delayValue` written by routing threads, read by JMX).

### Fix

Add `volatile` to the `delayValue` field. On x86 this compiles to the same `MOV` instruction as a plain load, so there is zero performance cost. No `AtomicLong` is needed because there are no read-modify-write operations — it's pure store/load.

## Test plan

- [x] `DelayerTest`, `DelayerAsyncDelayedTest`, `DelayerPerRouteTest` pass
- [x] `ManagedDelayerTest` (JMX MBean that reads `delayValue`) passes
- [x] `camel-core-processor` module compiles and formats cleanly
- [ ] Full CI

### DIFF
--- a/core/camel-core-processor/src/main/java/org/apache/camel/processor/Delayer.java
+++ b/core/camel-core-processor/src/main/java/org/apache/camel/processor/Delayer.java
@@ -39,7 +39,7 @@ public class Delayer extends DelayProcessorSupport implements Traceable, IdAware
     private String stepId;
     private String id;
     private volatile Expression delay;
-    private long delayValue;
+    private volatile long delayValue;
 
     public Delayer(CamelContext camelContext, Processor processor, Expression delay,
                    ScheduledExecutorService executorService, boolean shutdownExecutorService) {


gnodet (commented):
Review: Looks good ✅

Correct and minimal fix — adding `volatile` to `Delayer.delayValue` closes the JMM visibility gap between routing threads (writing in `calculateDelay()`) and JMX threads (reading via `ManagedDelayer.getDelay()`). No issues found.

### Details

- `delayValue` is a `long` field with cross-thread access: written on routing threads in `calculateDelay()` and read on JMX threads via `getDelayValue()`. Without `volatile`, 32-bit JVMs risk torn reads (JLS 17.7) and all JVMs risk stale reads.
- No `AtomicLong` needed since there are no read-modify-write operations.
- Performance cost is negligible — volatile loads compile to plain MOV on x86, and the store barrier in `calculateDelay()` is on a code path that already blocks or schedules a delayed task.
- Verified no other fields in `Delayer` or `DelayProcessorSupport` have the same cross-thread visibility concern: `asyncDelayed` and `callerRunsWhenRejected` are set during lifecycle initialization (with happens-before from CamelContext startup); `delayedCount` is already `AtomicInteger`.
- Clean follow-up to PR #24985 (CAMEL-24227 volatile sweep), which fixed the `delay` Expression field but missed the reverse direction (`delayValue`, routing-to-JMX).

PR #25207, opened 2026-07-29
CAMEL-24268: Remove write-only resumeInflightExchanges field

gnodet:

## Summary

Removes the write-only `resumeInflightExchanges` field from `ThrottlingInflightRoutePolicy` that became dead code after the `ThrottlingLimits` holder refactoring in PR #24985 (CAMEL-24227).

## Problem

After the holder refactoring, the standalone `resumeInflightExchanges` field was:
- **Written** by `setMaxInflightExchanges()` and `setResumePercentOfMax()` setters
- **Never read** — `throttle()` reads from the immutable `ThrottlingLimits` holder record instead
- **No getter** exists, `toString()` doesn't reference it, JMX MBean doesn't expose it
- The default `700` was duplicated between the field initializer and the holder constructor

## Changes

- Remove the dead `resumeInflightExchanges` field
- Remove the two dead writes (`this.resumeInflightExchanges = resume`) in the setter methods
- Derive the initial `ThrottlingLimits` holder default from `maxInflightExchanges` and `resumePercentOfMax` fields instead of duplicating literal values

Net: **−4 lines, +2 lines** in a single file.

## Test plan

- [x] Existing `ThrottlingInflightRoutePolicyTest` passes
- [x] Existing `ManagedThrottlingInflightRoutePolicyTest` passes
- [x] `camel-support` module tests pass
- [x] No other references to the removed field exist in the codebase (verified by grep)

### DIFF
--- a/core/camel-support/src/main/java/org/apache/camel/throttling/ThrottlingInflightRoutePolicy.java
+++ b/core/camel-support/src/main/java/org/apache/camel/throttling/ThrottlingInflightRoutePolicy.java
@@ -84,7 +84,7 @@ public class ThrottlingInflightRoutePolicy extends RoutePolicySupport implements
     }
 
     private volatile ThrottlingLimits throttlingLimits
-            = new ThrottlingLimits(1000, Math.max(70 * 1000 / 100, 1));
+            = new ThrottlingLimits(maxInflightExchanges, Math.max(resumePercentOfMax * maxInflightExchanges / 100, 1));
 
     @Metadata(description = "Sets the logging level to report the throttling activity.",
               javaType = "org.apache.camel.LoggingLevel", defaultValue = "INFO", enums = "TRACE,DEBUG,INFO,WARN,ERROR,OFF")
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Review: Looks good ✅

Clean removal of a write-only dead field (`resumeInflightExchanges`) left behind by the ThrottlingLimits holder refactoring in PR #24985 (CAMEL-24227). The field has no getter, is not referenced in `toString()`, is not exposed via the JMX MBean (`ManagedThrottlingInflightRoutePolicyMBean`), and is not present in the generated `ThrottlingInflightRoutePolicyConfigurer`. No external references exist.

The `ThrottlingLimits` initialization change from literals to computed values is safe — Java field initializers execute in textual order, both source fields are declared before `throttlingLimits`, and the computed result `Math.max(70 * 1000 / 100, 1) = 700` is identical to the old literal. CI is green on JDK 17 and JDK 25.

apupier (approved):

PR #25210, opened 2026-07-29
CAMEL-24289: camel-aws2-textract - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Textract2Producer`'s nine operations only acted when

## Fix

type, e.g. `"analyzeExpense operation requires AnalyzeExpenseRequest in POJO
mode"`.

## Tests

no-op) before the fix — with the `analyzeExpense` else removed, exactly that case
reports "Expecting code to raise a throwable". Class 12/12 green. Hermetic unit

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-textract/pom.xml
+++ b/components/camel-aws/camel-aws2-textract/pom.xml
@@ -78,5 +78,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, focused PR that adds `else` branches throwing `IllegalArgumentException` for all 9 Textract operations when `pojoRequest=true` and the body is the wrong type. The pattern, error messages, and test structure are fully consistent with the sibling PR #25199 (camel-aws2-polly) in the CAMEL-24261 series.

### Details

- All 9 operations in `Textract2Operations` are covered by both the production `else` branches and the `@CsvSource` parameterized test.
- Error message format (`<operation> operation requires <RequestType> in POJO mode`) is consistent with the polly PR and follows a clear, informative pattern.
- Test correctly uses AssertJ (`assertThatThrownBy`) with `hasRootCauseInstanceOf` and `hasRootCauseMessage`, and the new test method has proper package-private visibility.
- The `assertj-core` test dependency addition in `pom.xml` is appropriate and follows the same pattern as other AWS2 modules in this series.

PR #25215, opened 2026-07-29
CAMEL-24300: camel-aws2-ec2 - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `AWS2EC2Producer`'s nine operations only acted when the
body was the matching request type; any other body fell through the `instanceof`

## Fix

type. The message names the *operation the user configured* — e.g.
`createAndRunInstances` (the enum/URI operation) rather than the internal method
name `createAndRunInstance`.

## Tests

no-op) before the fix — with the `stopInstances` else removed, exactly that case
reports "Expecting code to raise a throwable". Class 20/20 green. Hermetic unit

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-ec2/pom.xml
+++ b/components/camel-aws/camel-aws2-ec2/pom.xml
@@ -77,5 +77,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-structured PR that correctly adds `IllegalArgumentException` for all 9 EC2 operations when `pojoRequest=true` and the body is the wrong type. Consistent with sibling PRs in the CAMEL-24261 series.

**Key observations:**
- All 9 operations with manual `if (payload instanceof ...)` pattern correctly received the `else` branch
- The two remaining operations (`describeInstances`, `describeInstancesStatus`) use the `executeOperation` helper which already throws on type mismatch — correctly excluded
- Error message format is consistent with sibling PRs (CAMEL-24283/comprehend, CAMEL-24280/mq, CAMEL-24264/translate, etc.)
- Test uses AssertJ (`assertThatThrownBy`) and is package-private — follows project conventions
- Parameterized test comprehensively covers all operations

**Minor nit (non-blocking):** Per project conventions, when modifying an existing test file the `public` modifier could be removed from the `EC2ProducerTest` class declaration (`public class` → `class`). This is incremental cleanup, not a blocker.

PR #25217, opened 2026-07-29
CAMEL-24286: Release the latch and unregister when a BackgroundTask supplier throws

oscerd:
## The bug

`BackgroundTask.runTaskWrapper` rethrows a supplier exception on the scheduler thread without counting down the completion latch or unregistering the task:

```java
try {
    if (doRun(supplier)) {
        ...
        latch.countDown();
    }
} catch (Exception e) {          // no countDown, no removeTask
    status = Status.Failed;
    cause = e;
    throw e;
}
```

`doRun` only catches `TaskRunFailureException`, so any *other* exception reaches this path. `scheduleWithFixedDelay` then suppresses all further executions, and for a task built with `withUnlimitedDuration()` `waitForTaskCompletion` calls `latch.await()` with **no timeout** — so the calling thread **blocks forever** and the task **leaks in the `TaskManagerRegistry`**.

The interrupt path has a smaller sibling of the same gap: in `waitForTaskCompletion`, `removeTask` sits inside the `try`, so an `InterruptedException` from `latch.await()` also leaves the task registered.

## Reachability on `main`

CAMEL-24272 now routes two *unlimited-duration* reconnection loops through `BackgroundTask.run()`:

- **camel-ftp SFTP** — `SftpOperations.java` uses `withUnlimitedDuration()`; `tryConnect` catches only `JSchException`. A non-JSch `RuntimeException` on the connect path escapes into the throw path.
- **camel-mongodb-gridfs** — `GridFsConsumer` uses `withUnlimitedDuration()` + `run()`. In `processCollection`, the inner `catch (Exception) { // ignore }` wraps only `getProcessor().process(...)`; the cursor acquisition and the pre-process `findOneAndUpdate` are **outside** any catch, so a transient `MongoException` escapes.

Bounded tasks (`withMaxDuration`, e.g. Infinispan) don't hang — `latch.await(maxDuration)` times out — but they still fail slow and lose the cause.

## The fix

In the `catch`, mark the task not-completed, unregister it, and count down the latch **before** rethrowing, so the blocking `run()` caller unblocks and observes the failure. The rethrow is kept so the `schedule()` path (no latch waiter) still surfaces the error to the executor.

Also move `removeTask` + `task.cancel` into the `finally` of `waitForTaskCompletion`, so an interrupted `await()` no longer leaves the task registered.

## Test

`BackgroundTaskRegistryTest.testRunReturnsAndUnregistersWhenSupplierThrows` builds an unlimited-duration task whose supplier throws, runs it on a **daemon** thread (so a regression can't wedge the JVM), and asserts `run()` returns, reports not-completed, and the registry ends empty.

Verified both directions locally:
- **without the fix** — the new test fails (`run()` hangs on the 10s join); the three existing tests pass.
- **with the fix** — all four pass.

## Notes

Found while reviewing the task-manager PR cluster (CAMEL-24271/24272/24278). It is the residual foreground→background porting gap, not attributable to a single PR. The rethrow is deliberately preserved in case surfacing the error to the executor on the `schedule()` path is intended — only the latch-release/unregister omission is fixed.

---

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/support/task/task/BackgroundTaskRegistryTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/support/task/task/BackgroundTaskRegistryTest.java
@@ -128,6 +128,49 @@ class BackgroundTaskRegistryTest {
         assertThat(registry.getTasks()).isEmpty();
     }
 
+    @DisplayName("Test that run() returns and unregisters when the supplier throws (CAMEL-24286)")
+    @Test
+    @Timeout(15)
+    void testRunReturnsAndUnregistersWhenSupplierThrows() throws Exception {
+        AtomicBoolean runReturned = new AtomicBoolean(false);
+        AtomicBoolean completed = new AtomicBoolean(true);
+
+        // an unlimited-duration task: before CAMEL-24286 a thrown supplier exception left the latch
+        // un-counted, so run() blocked forever here and the task leaked in the registry
+        BackgroundTask task = Tasks.backgroundTask()
+                .withScheduledExecutor(Executors.newSingleThreadScheduledExecutor())
+                .withBudget(Budgets.iterationTimeBudget()
+                        .withInterval(Duration.ofMillis(100))
+                        .withInitialDelay(Duration.ZERO)
+                        .withUnlimitedDuration()
+                        .build())
+                .build();
+
+        // daemon so a regression (run() hanging forever) cannot keep the JVM alive
+        Thread runner = new Thread(() -> {
+            boolean result = task.run(camelContext, () -> {
+                throw new IllegalStateException("boom");
+            });
+            completed.set(result);
+            runReturned.set(true);
+        });
+        runner.setDaemon(true);
+        runner.start();
+
+        runner.join(TimeUnit.SECONDS.toMillis(10));
+
+        assertThat(runReturned.get())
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-targeted fix for a real thread-hang bug in BackgroundTask. The latch countDown and registry cleanup are correctly added to the exception path, the `waitForTaskCompletion` cleanup is correctly moved to the finally block, and the regression test is well-designed and follows all project conventions.

**Highlights:**
- The fix correctly handles both the `run()` path (where `registeredByRun` is true, so cleanup is deferred to `waitForTaskCompletion.finally`) and the `schedule()` path (where cleanup happens in the catch block directly).
- Moving `removeTask` and `task.cancel(true)` from the try block to the finally block in `waitForTaskCompletion` is strictly correct — the old code left the task registered and the scheduled future running when `latch.await()` threw `InterruptedException`.
- Thread safety is sound: `CountDownLatch.countDown()` establishes happens-before with `latch.await()`, so writes to `status`, `completed`, and `cause` in the catch block are visible to the caller.
- The test is well-designed: uses `withUnlimitedDuration()`, runs on a daemon thread as a safety net, uses AssertJ and Awaitility per project conventions, and carries a `@Timeout(15)` as a hard safety net.
- CI passes (JDK 17 and JDK 25 builds both green).

PR #25219, opened 2026-07-29
Fix flakiness of Eventbridge tests

apupier:
flakiness was introduced around 22nd and 23rd of July

Several points:
* one test was not using the same abstract class causing two services to run in //
* all the test classes are sharing the same instance and they were using similar rule names and route names. Some rules can influence the other tests. provided a specific rule id for each of the test classes

# Description

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeDisableRuleIT.java
+++ b/components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeDisableRuleIT.java
@@ -28,9 +28,11 @@ import org.apache.camel.component.aws2.eventbridge.EventbridgeConstants;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.junit.jupiter.api.Test;
 import software.amazon.awssdk.services.eventbridge.model.ListRulesResponse;
+import software.amazon.awssdk.services.eventbridge.model.Rule;
 import software.amazon.awssdk.services.eventbridge.model.RuleState;
 import software.amazon.awssdk.services.eventbridge.model.Target;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class EventbridgeDisableRuleIT extends Aws2EventbridgeBase {
@@ -45,20 +47,21 @@ public class EventbridgeDisableRuleIT extends Aws2EventbridgeBase {
     public void sendIn() throws Exception {
         result.expectedMessageCount(1);
 
-        template.send("direct:evs", new Processor() {
+        template.send("direct:evs-EventbridgeDisableRuleIT", new Processor() {
 
             @Override
             public void process(Exchange exchange) {
-                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule");
+                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule-EventbridgeDisableRuleIT");
             }
         });
 
-        template.send("direct:evs-targets", new Processor() {
+        template.send("direct:evs-targets-EventbridgeDisableRuleIT", new Processor() {
 
             @Override
             public void process(Exchange exchange) {
-                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule");
-                Target target = Target.builder().id("sqs-queue").arn("arn:aws:sqs:eu-west-1:780410022472:camel-connector-test")
+                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule-EventbridgeDisableRuleIT");
+                Target target = Target.builder().id("sqs-queue-EventbridgeDisableRuleIT")
+                        .arn("arn:aws:sqs:eu-west-1:780410022472:camel-connector-test")
                         .build();
                 List<Target> targets = new ArrayList<Target>();
                 targets.add(target);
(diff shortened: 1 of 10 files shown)


apupier:
note that even safer way could be to create a specific namespace on the event bridge separated for each tests.

components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeListRulesIT.java
```diff
@@ -44,38 +45,47 @@ public class EventbridgeListRulesIT extends Aws2EventbridgeBase {
...
+                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "secondrule-EventbridgeListRulesIT");
             }
         });
 
-        template.send("direct:evs-targets", new Processor() {
+        template.send("direct:evs-targets-EventbridgeListRulesIT", new Processor() {
 
             @Override
             public void process(Exchange exchange) {
                 exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule");
```
gnodet: This `RULE_NAME` was not updated with the class suffix — it should be `"firstrule-EventbridgeListRulesIT"` to match the rule created above. As-is, `putTargets` targets a rule that doesn't exist in this test.

```suggestion
                exchange.getIn().setHeader(EventbridgeConstants.RULE_NAME, "firstrule-EventbridgeListRulesIT");
```

gnodet (commented):
Good root-cause analysis and generally sound fix for parallel test interference, @apupier. The core approach is correct: making `EventbridgePutEventsIT` extend `Aws2EventbridgeBase`, suffixing route/rule/target names with test class names, and switching to `contains`-based assertions where parallel tests share the event bus.

Three items to address (1 medium, 2 low):

**Medium — Missed rule name update:**
In `EventbridgeListRulesIT`, the `putTargets` call still references `RULE_NAME "firstrule"` instead of `"firstrule-EventbridgeListRulesIT"`. The two `putRule` calls above correctly create rules with the class suffix, but this one was missed. The `putTargets` step silently targets a non-existent rule — the test passes because the `listRules` assertion doesn't depend on targets, but it should be fixed for correctness.

**Low — Mixed assertion styles:**
`EventbridgeDisableRuleIT.sendIn()` and `EventbridgeEnableRuleIT.sendIn()` each mix JUnit `assertEquals` and AssertJ `assertThat` in the same method. Per project conventions: "Do NOT mix AssertJ and JUnit assertions in the same test method — pick one style per method."

components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeDisableRuleIT.java
```diff
@@ -83,9 +86,11 @@ public void process(Exchange exchange) {
 
         ListRulesResponse resp = ex.getIn().getBody(ListRulesResponse.class);
```
gnodet, line 87: Minor: this method now mixes JUnit `assertEquals` (lines 87, 92) with AssertJ `assertThat` (lines 88-89). Per project convention, pick one style per method — consider converting the `assertEquals` calls to AssertJ:
```java
assertThat(resp.hasRules()).isTrue();
// ...
assertThat(disabledRule.state()).isEqualTo(RuleState.DISABLED);
```

components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeEnableRuleIT.java
```diff
@@ -92,28 +95,30 @@ public void process(Exchange exchange) {
 
         ListRulesResponse resp = ex.getIn().getBody(ListRulesResponse.class);
```
gnodet, line 96: Same assertion mixing issue as `EventbridgeDisableRuleIT` — `assertEquals` and `assertThat` in the same method. Consider converting to a single style.

components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeDisableRuleIT.java
apupier: this is to avoid too many changes on the file which doesn't bring a lot of value

components/camel-aws/camel-aws2-eventbridge/src/test/java/org/apache/camel/component/aws2/eventbridge/localstack/EventbridgeEnableRuleIT.java
apupier: this is to avoid too many changes on the file which doesn't bring a lot of value

gnodet (approved):
Re-reviewed after new commit. The medium finding from the previous review (missed rule name suffix in `putTargets` in `EventbridgeListRulesIT`) has been fixed — the call now correctly uses `"firstrule-EventbridgeListRulesIT"`.

The two low-severity mixed-assertion findings were intentionally declined to keep the diff focused on the flakiness fix, which is a reasonable tradeoff.

The overall fix approach is sound: making `EventbridgePutEventsIT` extend the shared base class, suffixing route/rule/target names with test class names, using contains-based assertions, and adding `@Isolated` where needed. CI is fully green (Java 17 and Java 25).

PR #25220, opened 2026-07-29
CAMEL-24302: camel-aws2-timestream - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, both timestream producers only acted when the body was
the matching request type; any other body fell through the `instanceof` with no
returned with no error. This covers **26** operations:

* `Timestream2WriteProducer` — 16 (databases, tables, batch-load tasks, writeRecords, describeEndpoints)
* `Timestream2QueryProducer` — 10 (scheduled queries, prepareQuery, query, cancelQuery, describeEndpoints)

## Fix

type, e.g. `"writeRecords operation requires WriteRecordsRequest in POJO mode"`.

## Tests

A `@ParameterizedTest` in each producer's test class covers every operation,
sending a wrong-typed body to `pojoRequest=true` routes and asserting each
message. Verified to fail (silent no-op) before the fix — with `writeRecords`
(write) and `cancelQuery` (query) elses removed, exactly those cases report
"Expecting code to raise a throwable". Write 33/33, Query 21/21 green. Hermetic
unit tests, region-free — no localstack or real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-timestream/pom.xml
+++ b/components/camel-aws/camel-aws2-timestream/pom.xml
@@ -85,6 +85,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean, mechanical addition of `else { throw IllegalArgumentException }` to all 26 pojo-mode operations across both Timestream producers (10 query, 16 write), with comprehensive parameterized tests covering every operation. Follows the exact same pattern as the already-approved sibling PRs (#25199, #25210, #25215).

**Key observations:**
- All 26 operations (10 in `QueryProducer`, 16 in `WriteProducer`) correctly covered by new `else` branches
- Error message format consistent with the CAMEL-24261 series
- Tests use `@ParameterizedTest` with `@CsvSource` and AssertJ `assertThatThrownBy` — follows project conventions
- New test method is package-private — follows JUnit 5 conventions
- Pre-existing pojo routes correctly reused without duplication

apupier (approved):

PR #25227, opened 2026-07-29
CAMEL-24303: camel-aws2-redshift - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `RedshiftData2Producer`'s ten operations only acted when

## Fix

type, e.g. `"executeStatement operation requires ExecuteStatementRequest in POJO
mode"`.

## Tests

A `@ParameterizedTest` covers all ten operations, sending a wrong-typed body to
no-op) before the fix — with the `executeStatement` else removed, exactly that
case reports "Expecting code to raise a throwable". Class 20/20 green. Hermetic
unit tests, region-free — no localstack or real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-redshift/pom.xml
+++ b/components/camel-aws/camel-aws2-redshift/pom.xml
@@ -74,5 +74,10 @@
             <artifactId>camel-test-spring-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


apupier (approved):

gnodet (approved):
Clean, well-tested change that adds `pojoRequest` type-checking to all 10 `RedshiftData2Producer` operations, following the same pattern already approved in sibling PRs (aws2-ec2 #25215, aws2-polly #25199, aws2-timestream #25220).

**Highlights:**
- All 10 operations (listDatabases, listSchemas, listStatements, listTables, describeTable, executeStatement, batchExecuteStatement, cancelStatement, describeStatement, getStatementResult) have the else-throw block added.
- Test coverage via parameterized `@CsvSource` entries matches 1:1 with the route definitions.
- Test class and methods correctly drop `public` modifier per project conventions, and use AssertJ assertions.
- CI passes with all builds green.

PR #25229, opened 2026-07-29
CAMEL-24304: camel-aws2-step-functions - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `StepFunctions2Producer`'s sixteen operations (state
machines, activities, executions) only acted when the body was the matching
request type; any other body fell through the `instanceof` with no `else`, so no
AWS call was made, no response was set, and the original body was returned with
no error.

## Fix

type, e.g. `"startExecution operation requires StartExecutionRequest in POJO
mode"`.

## Tests

A `@ParameterizedTest` covers all sixteen operations, sending a wrong-typed body
to `pojoRequest=true` routes and asserting each message. Verified to fail (silent
no-op) before the fix — with the `startExecution` else removed, exactly that case
reports "Expecting code to raise a throwable". Class 33/33 green. Hermetic unit

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-step-functions/pom.xml
+++ b/components/camel-aws/camel-aws2-step-functions/pom.xml
@@ -80,6 +80,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-structured fix that adds missing else-throw branches to all 16 StepFunctions2Producer POJO-mode operations, consistent with the established CAMEL-24261 pattern already merged for other AWS2 producers. CI is green, tests are comprehensive (good use of `@CsvSource` parameterized test for 16 operations), and the upgrade guide was already handled by CAMEL-24263.

Nit: the test class declaration is still `public class StepFunctions2ProducerTest` — per project conventions, the `public` modifier could be dropped since the file is being modified. Very minor.

PR #25107, opened 2026-07-25
chore(deps): upgrade IBM MQ to 10.0

davsclaus:
## Summary

Upgrade IBM MQ client and container to version 10.0, combining PR #24040 and PR #24640:

- `com.ibm.mq:com.ibm.mq.jakarta.client`: 9.4.5.1 → 10.0.0.0
- Container image `icr.io/ibm-messaging/mq`: 9.4.5.0-r2 → 10.0.0.0-r2

Closes #24040
Closes #24640

_Claude Code on behalf of davsclaus_

PR #25231, opened 2026-07-30
fix: force LANG=C on MQ 10.0 container to fix JmsReplyToIbmMQTest

gnodet:
## Summary

Fix the `JmsReplyToIbmMQTest.testCustomJMSReplyToInOut` test failure introduced by the IBM MQ 10.0 upgrade in PR #25107.

- IBM MQ 10.0 changed the default locale from `C` to `C.utf8`, causing the queue manager to use **CCSID 1208 (UTF-8)** instead of **CCSID 819 (ISO 8859-1)**
- This change causes the JMS reply-to mechanism to fail silently — the reply message never reaches `DEV.QUEUE.2`
- Force `LANG=C` on the container to preserve the CCSID 819 behavior from MQ 9.4.x

This is a pragmatic fix to validate the CCSID hypothesis. If CI passes with this change, the CCSID locale change is confirmed as the root cause. A proper fix (supporting CCSID 1208 natively) can be investigated as a follow-up.

## Test plan

- [ ] `JmsReplyToIbmMQTest.testCustomJMSReplyToInOut` should pass (currently fails 3/3 runs)
- [ ] `JmsComponentIbmMQTest.testSend` should continue to pass
- [ ] No regressions in other JMS tests


gnodet:
Closing: PR targeting a feature branch doesn't trigger the main build. Will re-create targeting main.

PR #25232, opened 2026-07-30
Fix IBM MQ 10.0 upgrade: handle reserved JMS vendor properties

gnodet:

## Summary

Upgrade IBM MQ client library and container image to version 10.0, and fix a compatibility issue where IBM MQ 10.0 marks the `JMS_IBM_MsgToken` vendor property as read-only.

This is a follow-up to #25107 which identified the upgrade but had test failures.

## Changes

- **`parent/pom.xml`**: Upgrade `com.ibm.mq:com.ibm.mq.jakarta.client` from 9.4.5.1 to 10.0.0.0
- **`test-infra/camel-test-infra-ibmmq/src/test/resources/container.properties`**: Upgrade container image from `icr.io/ibm-messaging/mq:9.4.5.0-r2` to `icr.io/ibm-messaging/mq:10.0.0.0-r2`
- **`components/camel-jms/src/main/java/.../JmsBinding.java`**: Add defensive try-catch in `appendJmsProperty()` for JMS vendor-specific properties (`JMS_*` prefix per JMS spec section 3.5.1) that the provider marks as reserved/read-only. Non-vendor properties still throw on failure as before.
- **`components/camel-jms/src/test/java/.../JmsReplyToIbmMQTest.java`**: Modernized test (dropped `public`, switched to AssertJ assertions)
- **`docs/.../camel-4x-upgrade-guide-4_22.adoc`**: Added upgrade guide entries for the IBM MQ 10.0 upgrade and the behavioral change
- **Generated metadata files**: Updated `serviceVersion` in IBM MQ test-infra JSON files

## Root Cause

IBM MQ 10.0 marks `JMS_IBM_MsgToken` as a reserved property that cannot be set by applications. When Camel copies all headers from an incoming JMS message to an outgoing reply, `JmsBinding.appendJmsProperty()` attempted to set this reserved property, causing a `MessageFormatException`. This exception propagated up through Spring's `DefaultMessageListenerContainer`, preventing the reply from being sent — resulting in the test receiving `null` instead of the expected response.

The fix catches `JMSException` only for JMS vendor-specific properties (`JMS_*` prefix) and logs at WARN level, while re-throwing for non-vendor properties. There is no standard JMS API to query which vendor properties are reserved, and the set can change between provider versions.

## Test Plan

- [x] `JmsReplyToIbmMQTest.testCustomJMSReplyToInOut` passes with IBM MQ 10.0
- [x] Full CI green

### DIFF
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/test-infra/metadata.json
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/test-infra/metadata.json
@@ -293,7 +293,7 @@
   "groupId" : "org.apache.camel",
   "artifactId" : "camel-test-infra-ibmmq",
   "version" : "4.22.0-SNAPSHOT",
-  "serviceVersion" : "9.4.5.0-r2",
+  "serviceVersion" : "10.0.0.0-r2",
   "uiSupported" : true
 }, {
   "service" : "org.apache.camel.test.infra.aws.common.services.AWSInfraService",
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsBinding.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsBinding.java
@@ -471,14 +471,19 @@ public class JmsBinding {
                     // set the property
                     JmsMessageHelper.setProperty(jmsMessage, key, value);
                 } catch (JMSException e) {
-                    // Some JMS providers (e.g. IBM MQ) mark certain vendor-specific properties
-                    // as read-only/reserved (set by the provider, not the application). When Camel
-                    // copies headers from an incoming message to an outgoing reply or forwarded
-                    // message, these reserved properties cause an exception. We skip them gracefully
-                    // since they are provider-assigned metadata that should not be propagated.
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("Could not set JMS property '{}' on outgoing message, skipping ({})",
+                    // JMS vendor-specific properties (JMS_<vendor>_xxx per JMS spec section 3.5.1)
+                    // may be read-only/reserved — set by the provider, not the application.
+                    // There is no standard JMS API to query which vendor properties are reserved,
+                    // and the set can change between provider versions (e.g. IBM MQ 10.0 made
+                    // JMS_IBM_MsgToken read-only). We skip them gracefully since they are
+                    // provider-assigned metadata that should not be propagated.
+                    // For non-vendor properties, we re-throw the exception as it likely indicates
+                    // a real problem (e.g. invalid value type, message not writable).
+                    if (key.startsWith("JMS_")) {
+                        LOG.warn("Could not set JMS vendor property '{}' on outgoing message, skipping: {}",
                                 key, e.getMessage());
+                    } else {
+                        throw e;
                     }
                 }
             } else if (LOG.isDebugEnabled()) {
--- a/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsBinding.java
+++ b/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsBinding.java
@@ -467,8 +467,20 @@ public class JmsBinding {
             if (value != null) {
                 // must encode to safe JMS header name before setting property on jmsMessage
                 String key = jmsKeyFormatStrategy.encodeKey(headerName);
-                // set the property
-                JmsMessageHelper.setProperty(jmsMessage, key, value);
+                try {
+                    // set the property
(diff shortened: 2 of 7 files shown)


oscerd (approved):

gnodet (commented):

The code fix itself is sound — defensively catching exceptions around `setProperty` to handle reserved JMS vendor properties (like IBM MQ 10.0's `JMS_IBM_MsgToken`) is a well-reasoned approach that follows existing codebase patterns (e.g., `JmsMessageHelper.setCorrelationId` at line 184-190). The broad `JMSException` catch is appropriate for multi-provider compatibility since the JMS spec only declares `throws JMSException` on `setXxxProperty` methods. CI is green and the test modernization correctly follows project conventions.

Two documentation items to consider:

---

### 🟡 Missing upgrade guide entry for IBM MQ client 9.x → 10.x

The upgrade guide (`camel-4x-upgrade-guide-4_22.adoc`) documents other major dependency bumps in this release (Debezium 3.5.2→3.6.0, Smack 4.3→4.4, Weaviate v5→v6). The IBM MQ client library jump from 9.4.5.1 to 10.0.0.0 is a major version change that introduced behavioral differences (reserved `JMS_IBM_MsgToken` property). Users who depend on IBM MQ-specific behavior or pin their IBM MQ client version should be informed.

### 🟡 Behavioral change applies to all JMS providers

Before this change, a `JMSException` from `JmsMessageHelper.setProperty` would propagate up through `appendJmsProperty`, potentially causing outgoing message creation to fail. After this change, such exceptions are caught and logged at DEBUG level only, with the property silently skipped. This applies to **all** JMS providers, not just IBM MQ. Users who relied on exception propagation (e.g., for custom error handling or detecting when critical properties fail to set) would experience different behavior. This behavioral change should also be noted in the upgrade guide.

apupier (changes requested):
I think the behavorial change and major library upgrade must be documented as mentioned here https://github.com/apache/camel/pull/25232#pullrequestreview-4825984458

are there ways to know about the reserved JMS vendor properties list? So that we skip them with a warning but continue to throw an exeption when it is a different problem?

gnodet:

@apupier Thanks for the review! Both points addressed in the latest push:

**1. Narrowed the catch scope:**
The `JMSException` catch now only applies to JMS vendor-specific properties (those with a `JMS_` prefix, per JMS spec section 3.5.1). For all other properties, the exception is re-thrown as before — so there is no behavioral change for user-defined properties. The log level has also been raised from `DEBUG` to `WARN` for better visibility.

Regarding your question about a reserved properties list: there is no standard JMS API to query which vendor properties are reserved. The set varies by provider and can change between versions (IBM MQ 10.0 made `JMS_IBM_MsgToken` read-only, which was writable in 9.x). The `JMS_` prefix check is the most reliable approach — it matches exactly the vendor-specific namespace defined in the JMS specification, and only those properties can be provider-reserved.

**2. Upgrade guide entries added:**
Two new sections in `camel-4x-upgrade-guide-4_22.adoc`:
- **camel-jms - IBM MQ client upgraded to 10.0** — documents the major version bump
- **camel-jms - vendor-specific JMS properties skipped when read-only** — documents the behavioral change, noting it only affects `JMS_`-prefixed properties

gnodet (commented):
Re-review after new commits addressing review feedback: the second commit cleanly addresses all prior review points — the catch scope is narrowed to JMS vendor-specific properties only (`JMS_` prefix per JMS spec section 3.5.1), logging is elevated to WARN, and thorough upgrade guide entries document both the IBM MQ 10.0 major version bump and the behavioral change in `JmsBinding`. CI is green. LGTM ✅

gnodet (commented):
Solid PR that upgrades IBM MQ to 10.0 and adds well-scoped defensive handling for read-only JMS vendor properties. The approach is technically sound, consistent with existing codebase patterns, properly documented, and CI is green. All prior reviewer feedback has been addressed.

**Observations:**

- The try-catch approach correctly scopes exception handling to JMS vendor-specific properties only (`JMS_` prefix per JMS spec section 3.5.1), re-throwing for non-vendor properties. This is consistent with the existing pattern in `JmsMessageHelper.setCorrelationId` which also catches and ignores `JMSException`.
- The upgrade guide entries are well-written and cover both the major version bump and the behavioral change with clear scope limitation to `JMS_`-prefixed properties.
- Test modernization correctly follows project conventions: dropped `public` modifier, switched to AssertJ assertions.
- The WARN log message intentionally uses `e.getMessage()` rather than logging the full stack trace — appropriate since vendor-property rejection is an expected condition.

**Note:** The prior PR #25107 (the original dependency-only upgrade) appears to still be open. It should probably be closed as superseded by this PR.

PR #25241, opened 2026-07-30
CAMEL-20199: Convert MllpSocketBuffer synchronized methods to ReentrantLock

gnodet:

## Summary

- Convert all 34 `synchronized` methods in `MllpSocketBuffer.java` to use `java.util.concurrent.locks.ReentrantLock` for virtual thread compatibility
- `synchronized` blocks cause virtual thread pinning (the carrier thread cannot be released while a virtual thread holds an intrinsic monitor), which degrades throughput under virtual-thread workloads
- `ReentrantLock` does not pin; it allows the carrier thread to be released when the virtual thread parks while waiting for the lock

Each formerly-`synchronized` method now acquires the lock at entry and releases it in a `finally` block, preserving the same mutual-exclusion semantics. No `wait()`/`notify()` usage existed, so no `Condition` conversion was needed.

All 335 existing tests pass.

Part of [CAMEL-20199](https://issues.apache.org/jira/browse/CAMEL-20199): Complete support of Virtual Threads.

## Test plan

- [x] All camel-mllp unit tests pass (335 tests, 0 failures)
- [ ] Manual review that lock/unlock pairs are correct and no synchronized keywords remain

### DIFF
--- a/components/camel-mllp/src/main/java/org/apache/camel/component/mllp/internal/MllpSocketBuffer.java
+++ b/components/camel-mllp/src/main/java/org/apache/camel/component/mllp/internal/MllpSocketBuffer.java
@@ -24,6 +24,7 @@ import java.net.SocketAddress;
 import java.net.SocketTimeoutException;
 import java.nio.charset.Charset;
 import java.util.Arrays;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.camel.component.mllp.MllpComponent;
 import org.apache.camel.component.mllp.MllpEndpoint;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
Looks good ✅

Clean, correct mechanical conversion of all 34 synchronized methods in MllpSocketBuffer to ReentrantLock with try-finally blocks. The conversion preserves mutual exclusion semantics while eliminating virtual thread pinning.

Key verification points:
- All 34 previously-synchronized methods correctly converted — no methods missed, no `synchronized` keywords remain
- The `ReentrantLock` is `private final` and not exposed — no external code can interfere
- No `wait()`/`notify()`/`notifyAll()` existed in the original code, so no `Condition` conversion needed
- No external callers synchronize on `MllpSocketBuffer` instances (verified via grep)
- Lock is held during blocking socket I/O in `readFrom()` and `writeTo()` — matches original behavior but now without pinning the carrier thread

apupier (approved):

PR #25243, opened 2026-07-30
CAMEL-24307: camel-aws2-rekognition - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Rekognition2Producer`'s 27 operations only acted when

## Fix

type, e.g. `"detectText operation requires DetectTextRequest in POJO mode"`.

## Tests

A `@ParameterizedTest` covers all 27 operations, sending a wrong-typed body to
no-op) before the fix — with the `detectText` else removed, exactly that case
reports "Expecting code to raise a throwable". Class 54/54 green. Hermetic unit

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-rekognition/pom.xml
+++ b/components/camel-aws/camel-aws2-rekognition/pom.xml
@@ -80,6 +80,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-structured PR. The pattern is identical to the merged sibling PRs in the CAMEL-24261 series (camel-aws2-timestream, camel-aws2-redshift, camel-aws2-step-functions, etc.). All 27 operations in the `Rekognition2Operations` enum have a corresponding `else` throw with `IllegalArgumentException` in the producer and a matching entry in the parameterized test.

Key observations:
- Test uses `@ParameterizedTest` + `@CsvSource` with AssertJ's `assertThatThrownBy` — clean and well-structured
- New test method is package-private (JUnit 5 convention ✓)
- All operations consistently covered in both production code and test
- CI checks pass

LGTM ✅

PR #25245, opened 2026-07-30
CAMEL-20199: Replace synchronized with ReentrantLock in AI components

gnodet:
## Summary

- Replace `synchronized` blocks with `ReentrantLock` in three files within the `camel-ai` parent module to improve virtual thread compatibility (part of CAMEL-20199)
- `GuardedTaskStore`: Replace internal `Object lifecycleLock` with `ReentrantLock`, expose via `getLock()` so external callers can coordinate
- `A2AProgress`: Use `GuardedTaskStore.getLock()` to share the same lock instance, with a `ConcurrentHashMap`-based fallback for non-guarded stores
- `AbstractTaskPredictor`: Replace `synchronized (this)` with a private `ReentrantLock` field

## Test plan

- [x] `mvn test -B` in `components/camel-ai/camel-a2a` -- 506 tests pass, 0 failures
- [x] `mvn test -B` in `components/camel-ai/camel-huggingface` -- BUILD SUCCESS
- [x] `mvn formatter:format impsort:sort` passes without reverting changes

### DIFF
--- a/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
+++ b/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
@@ -273,10 +273,12 @@ public class A2AEndpoint extends DefaultEndpoint {
     }
 
     private void cleanupEndpointResources() {
-        if (taskStore != null) {
-            A2AProgress.removeStoreLock(taskStore);
-        }
         if (taskStoreOwned && taskStore != null) {
+            // Clean up the cached lock before stopping the store. Only owned stores
+            // (InMemoryTaskStore) can appear in STORE_LOCKS — registry-discovered stores
+            // are wrapped in GuardedTaskStore, which provides its own lock via getLock()
+            // and never enters the static map.
+            A2AProgress.removeStoreLock(taskStore);
             try {
                 ServiceHelper.stopService(taskStore);
             } catch (Exception e) {
--- a/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
+++ b/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
@@ -273,6 +273,9 @@ public class A2AEndpoint extends DefaultEndpoint {
     }
 
     private void cleanupEndpointResources() {
+        if (taskStore != null) {
+            A2AProgress.removeStoreLock(taskStore);
+        }
         if (taskStoreOwned && taskStore != null) {
             try {
                 ServiceHelper.stopService(taskStore);
--- a/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AProgress.java
+++ b/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AProgress.java
@@ -18,6 +18,8 @@ package org.apache.camel.component.a2a;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.camel.Endpoint;
 import org.apache.camel.Exchange;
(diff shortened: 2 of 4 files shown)


gnodet (commented):
Looks good ✅

Clean synchronized-to-ReentrantLock migration in the AI components that also **implicitly fixes a pre-existing lock coordination bug** between A2AProgress and GuardedTaskStore.

Bug fix detail: In the original code, `A2AProgress.emitArtifact()` and `emitMessage()` used `synchronized (taskContext.store)` (locking on the GuardedTaskStore object), while `GuardedTaskStore.put()`, `updateStatusAndNotify()`, and `cancelIfNotTerminal()` used `synchronized (lifecycleLock)` (a separate private Object). These were **different monitors**, meaning the read-modify-write operations in A2AProgress were not coordinated with GuardedTaskStore's lifecycle operations. The PR fixes this by having both sides share the same ReentrantLock via `getLock()`.

Other observations:
- `notifySubscribers` call in `updateStatusAndNotify` is correctly left outside the lock, matching the original design intent
- AbstractTaskPredictor change is clean — replacing `synchronized (this)` with a private ReentrantLock is strictly better for encapsulation

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AProgress.java
```diff
@@ -30,16 +32,31 @@
...
+import org.apache.camel.component.a2a.state.GuardedTaskStore;
 
 /**
  * Static utility for emitting A2A progress updates from any exchange. Resolves the task store via the exchange's
  * {@code fromEndpoint} (primary) or by scanning the A2A component's endpoints (fallback).
  */
 public final class A2AProgress {
 
+    private static final ConcurrentHashMap<A2ATaskStore, ReentrantLock> STORE_LOCKS = new ConcurrentHashMap<>();
+
```
apupier, line 44: Unless I miss it but this map is never cleaned. When a store is removed, I would expect that the corresponding reentrant lock is removed

gnodet: Good catch, thank you! Added a package-private `removeStoreLock(A2ATaskStore)` method to `A2AProgress` and wired it into `A2AEndpoint.cleanupEndpointResources()` so the lock entry is removed when the endpoint shuts down. This prevents the static `STORE_LOCKS` map from leaking entries.

components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/A2AEndpoint.java
apupier: shouldn't it be better to clean it when the TaskStore service is stopped?

Here I fear that the lock could be removed for the taskstore which are not owned by this A2AEndpoint

gnodet: Good point — moved the `removeStoreLock()` call inside the `if (taskStoreOwned)` guard so it only cleans up locks for stores exclusively owned by this endpoint. Non-owned (registry-discovered) stores are always wrapped in `GuardedTaskStore`, which provides its own lock via `getLock()` and never enters the static `STORE_LOCKS` map, so they were never at risk of premature removal — but the guard makes the intent explicit and prevents any future surprise.

_Claude Code on behalf of

apupier (approved):

gnodet (commented):
Looks good ✅

**Re-review after latest commits (addressing apupier's feedback)**

Both review threads have been properly addressed:

1. **STORE_LOCKS memory leak** — `removeStoreLock()` is now called in `doStop()` to prevent entries from leaking when endpoints are stopped.
2. **Ownership scoping** — The cleanup is correctly placed inside the `taskStoreOwned` guard, so only owned stores (InMemoryTaskStore) are affected. Registry-discovered stores use `GuardedTaskStore` which provides its own lock via `getLock()` and never enters the static `STORE_LOCKS` map.

The PR has a human approval from apupier on the latest commit. CI passes on JDK 17 and JDK 25.

PR #25248, opened 2026-07-30
CAMEL-24315: camel-google-pubsub - stop subscribers that finish starting after the consumer stopped

henrik242:
Since CAMEL-22898 (4.18.0), GooglePubsubConsumer adds a subscriber to the subscribers list only after startAsync().awaitRunning() returns. doStop() and deferShutdown() only stop subscribers already in that list, so stopping the consumer while a subscriber is still starting misses it. Nothing ever calls stopAsync() on it, and the wrapper thread parks forever in awaitTerminated(), an uninterruptible Guava wait.

The consumer executor then never terminates: context shutdown waits 2 x shutdownAwaitTermination (default 10s) per stuck pool and the thread leaks even after forced shutdown. Test suites that stop many CamelContexts (e.g. Spring Boot integration tests) pay up to 20s per leaked consumer per context; we saw a CI build go from 5 to 25 minutes from this alone.

The synchronous pull path has the same window: the pull future is added to pendingSynchronousPullResponses only after futureCall() returns, so a stop landing in between misses it and get() blocks until the pull deadline instead of being cancelled.

Fix: after registering the subscriber or pull future, re-check the consumer state and stop/cancel it if the consumer is no longer allowed to run. stop() sets the volatile service status to STOPPING before doStop() runs, so either doStop() sees the subscriber in the list or the re-check sees the stopped status. Unit tests reproduce both races deterministically and fail without the fix.

Workaround on affected versions: lower shutdownAwaitTermination to bound the shutdown delay; the thread still leaks.

Issue at CAMEL-24315

### DIFF
--- a/components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
+++ b/components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
@@ -260,6 +260,10 @@ public class GooglePubsubConsumer extends DefaultConsumer implements ShutdownAwa
                     // Only add to list after successful startup
                     subscribers.add(subscriber);
                     subscriberAdded = true;
+                    // a stop while the subscriber was starting missed it; stop it here or awaitTerminated never returns
+                    if (!isRunAllowed() || isSuspendingOrSuspended()) {
+                        subscriber.stopAsync();
+                    }
                     subscriber.awaitTerminated();
                 } catch (Exception e) {
                     // Remove from list if it was added
@@ -328,6 +332,10 @@ public class GooglePubsubConsumer extends DefaultConsumer implements ShutdownAwa
 
                     synchronousPullResponseFuture = subscriber.pullCallable().futureCall(pullRequest);
                     pendingSynchronousPullResponses.add(synchronousPullResponseFuture);
+                    // a stop while the pull was being issued missed this future; cancel it here or get() blocks
+                    if (!isRunAllowed() || isSuspendingOrSuspended()) {
+                        synchronousPullResponseFuture.cancel(true);
+                    }
                     PullResponse pullResponse = synchronousPullResponseFuture.get();
                     for (ReceivedMessage message : pullResponse.getReceivedMessagesList()) {
                         PubsubMessage pubsubMessage = message.getMessage();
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Correct, minimal fix for a real race condition where subscribers or pull futures starting during consumer stop are missed by `doStop()`, causing threads to park forever.

**Why the fix is correct:**

The double-check pattern works because `BaseService.stop()` sets `status = STOPPING` (volatile) BEFORE calling `doStop()`. Combined with the volatile visibility guarantee, either (a) `doStop()` sees the subscriber in the list and stops it, or (b) the re-check sees `!isRunAllowed()` and stops it. Both paths cannot be missed simultaneously.

**Additional observations:**

- Calling `stopAsync()` in the re-check AND in the `finally` block is safe — Guava's `AbstractApiService.stopAsync()` is idempotent.
- For the synchronous pull path, cancelling the future causes `future.get()` to throw `CancellationException`, which is already caught at line 398 — no unhandled exception risk.
- The tests are well-designed: they use `CountDownLatch` barriers to deterministically reproduce the race rather than relying on timing.
- The PubSub Lite consumer (also modified in CAMEL-22898) no longer exists in the codebase, so no parallel fix is needed.
- Minor style notes (not blocking): test class uses `public` modifier and JUnit assertions instead of the project-preferred package-private + AssertJ, but this is consistent with the existing `GooglePubsubConsumerShutdownTest` in the same package.

oscerd (approved):

oscerd:
This needs backport to 4.18.x

gnodet (approved):
## Review Summary

Correct, minimal fix for a real race condition where subscribers or pull futures that finish starting after the consumer enters `doStop()` are never stopped/cancelled, causing threads to park forever.

### Correctness analysis

The double-check pattern is sound under the JMM:
- `BaseService.status` is volatile
- `CopyOnWriteArrayList` and `ConcurrentHashMap.newKeySet()` use internal volatile references
- For any execution, at least one of {`doStop()` sees the subscriber in the list} or {the re-check sees `!isRunAllowed()`} must be true

Guava's `AbstractApiService.stopAsync()` is idempotent, so the double-stop that occurs when both `doStop()` and the re-check fire is safe.

### Test quality

The tests deterministically reproduce both races (subscriber start and synchronous pull) with well-designed mocks. CI is green.

### Minor convention notes (non-blocking)

- Test class uses `public` modifier — project convention prefers package-private for JUnit 5. Consistent with the existing sibling test in this package though.
- JUnit assertions used instead of AssertJ — also consistent with local package convention.

Excellent PR description with clear root cause analysis and correctness argument.

---

davsclaus (approved):

PR #25250, opened 2026-07-30
CAMEL-24316: camel-aws2-transcribe - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `Transcribe2Producer`'s 30 operations only acted when

This is distinct from CAMEL-24207, which implemented the previously-empty
operation stubs — here the operations work, but the pojo-mode path silently
dropped a wrong-typed body.

## Fix

type, e.g. `"createVocabulary operation requires CreateVocabularyRequest in POJO
mode"`.

## Tests

A new `Transcribe2ProducerPojoRequestTest` uses a `@ParameterizedTest` covering
all 30 operations, sending a wrong-typed body to `pojoRequest=true` routes and
asserting each message. Verified to fail (silent no-op) before the fix — with the
`createVocabulary` else removed, exactly that case reports "Expecting code to
raise a throwable". 30/30 green. Hermetic unit tests, region-free — no localstack
or real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-transcribe/src/main/java/org/apache/camel/component/aws2/transcribe/Transcribe2Producer.java
+++ b/components/camel-aws/camel-aws2-transcribe/src/main/java/org/apache/camel/component/aws2/transcribe/Transcribe2Producer.java
@@ -156,6 +156,9 @@ public class Transcribe2Producer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
+            } else {
+                throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-structured defensive-programming fix identical in pattern to the approved and merged sibling PR #25243 (Rekognition). All 30 operations in the `Transcribe2Operations` enum get a matching `else throw IllegalArgumentException` in the producer and a corresponding entry in the parameterized test. The test uses `@ParameterizedTest` + `@CsvSource` with AssertJ `assertThatThrownBy`, package-private visibility, and covers all 30 enum operations. Pattern and conventions are consistent across the CAMEL-24261 series. CI is green. ✅

PR #25254, opened 2026-07-30
CAMEL-23931: Fix camel.jbang.quarkus.platform.url ignored from application.properties

jomin7:
Fixes [CAMEL-23931](https://redhat.atlassian.net/browse/CEQ-13931)
Setting camel.jbang.quarkus.platform.url in application.properties had no effect —the export always used the default public registry. Only JVM -D flags or the --quarkus-ext-registry CLI flag worked.
  
Fixed QuarkusPlatformMixin.of() to also read this property from the properties file.

Regression from CAMEL-23353 backport in 4.18.3.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/QuarkusPlatformMixin.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/QuarkusPlatformMixin.java
@@ -95,8 +95,26 @@ public class QuarkusPlatformMixin extends QuarkusExtensionRegistryMixin implemen
         QuarkusPlatformMixin result = new QuarkusPlatformMixin();
         result.quarkusGroupId = props.getProperty(QUARKUS_GROUP_ID, fallback.quarkusGroupId());
         result.quarkusVersion = props.getProperty(QUARKUS_VERSION, fallback.quarkusVersion());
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-tested fix for a real user-facing bug where `camel.jbang.quarkus.platform.url` set in `application.properties` was silently ignored during export. The logic correctly mirrors the existing system-property handling in `QuarkusExtensionRegistryMixin.quarkusExtensionRegistryBaseUri()`.

**Observations (not blocking):**

- Good job including tests for the two-phase loading pattern (app-properties then system-properties), which matches the real call sites in `Export.java`. The canonical-key-takes-priority-over-legacy-key test is important and well-designed.
- Minor: the suffix-stripping logic for `/client/platforms` and trailing-slash stripping is now duplicated between `QuarkusPlatformMixin.of()` and `QuarkusExtensionRegistryMixin.quarkusExtensionRegistryBaseUri()`. Acceptable for this focused fix, but a future cleanup could extract a shared normalisation helper.
- Pre-existing (not introduced by this PR): the `resolve()` method uses the raw field `quarkusExtensionRegistryBaseUri` at line 74 (when `quarkusVersion != null`) but calls the method `quarkusExtensionRegistryBaseUri()` at line 80 (else branch). The field access bypasses the system-property fallback. With this PR the `of()` path now normalises the value before storing it in the field, which partially mitigates this inconsistency.

apupier (approved):

PR #25255, opened 2026-07-30
CAMEL-24317: camel-aws2-eventbridge - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `EventbridgeProducer.putRule` only acted when the body
was a `PutRuleRequest`; any other body fell through the `instanceof` with no
returned with no error. (The other Eventbridge operations already build their
request from headers and don't have a pojo branch.)

## Fix

```java
} else {
            "putRule operation requires PutRuleRequest in POJO mode");
}
```

## Test

`camel-aws2-eventbridge` has no producer route/mock harness, so this is covered
by a new Mockito unit test (`EventbridgeProducerPojoRequestTest`) that mocks the
endpoint + client, sends a wrong-typed body with `pojoRequest=true`, and asserts
the `IllegalArgumentException`. Verified to fail (silent no-op) before the fix.
Region-free — no localstack or real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
+++ b/components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
@@ -157,7 +157,9 @@ public class EventbridgeProducer extends DefaultProducer {
                 message.setHeader(EventbridgeConstants.RULE_ARN, result.ruleArn());
             } else {
                 throw new IllegalArgumentException(
-                        "putRule operation requires PutRuleRequest in POJO mode");
+                        String.format("Expected body of type %s but was %s",
+                                PutRuleRequest.class.getName(),
+                                ObjectHelper.isNotEmpty(payload) ? payload.getClass().getName() : "null"));
             }
         } else {
             PutRuleRequest.Builder builder = PutRuleRequest.builder();
--- a/components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
+++ b/components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
@@ -155,6 +155,9 @@ public class EventbridgeProducer extends DefaultProducer {
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
                 message.setHeader(EventbridgeConstants.RULE_ARN, result.ruleArn());
+            } else {
+                throw new IllegalArgumentException(
+                        "putRule operation requires PutRuleRequest in POJO mode");
             }
         } else {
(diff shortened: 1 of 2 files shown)


components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
```diff
@@ -155,6 +155,9 @@ private void putRule(EventBridgeClient eventbridgeClient, Exchange exchange) thr
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
                 message.setHeader(EventbridgeConstants.RULE_ARN, result.ruleArn());
+            } else {
+                throw new IllegalArgumentException(
```
gnodet, line 159: Minor: the error message format differs from the `executeOperation` helper (lines 477-480) which uses `String.format("Expected body of type %s but was %s", ...)`. Consider either migrating `putRule` to use `executeOperation` (like all other operations in this class already do) or matching the error message format for consistency and debuggability.

gnodet (commented):
The production fix is correct — `putRule` was missing the `else` branch to throw when `pojoRequest=true` and the body is not a `PutRuleRequest`, matching the CAMEL-24261 series pattern. The test logic is sound and compiles/passes correctly (assertj-core is already available on `main`).

**Minor observation (not blocking):**

`putRule` is the only operation in `EventbridgeProducer` that was not migrated to use the `executeOperation` helper (all other operations — putTargets, removeTargets, deleteRule, enableRule, disableRule, listRules, describeRule, listTargetsByRule, listRuleNamesByTarget, putEvent — already use it). The error message format here (`"putRule operation requires PutRuleRequest in POJO mode"`) differs from `executeOperation`'s more debuggable format (`"Expected body of type %s but was %s"`) which includes both the expected and actual type names. Consider either migrating `putRule` to use `executeOperation` or matching its error message format for consistency and debuggability.

apupier (approved):

components/camel-aws/camel-aws2-eventbridge/src/main/java/org/apache/camel/component/aws2/eventbridge/EventbridgeProducer.java
oscerd: Done in dda6987 — putRule now throws String.format("Expected body of type %s but was %s", PutRuleRequest.class.getName(), ...), matching executeOperation. (putRule stays on its own path since it also sets the RULE_ARN header from the response, but the error message is now consistent.) Test updated to assert the new format.\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

gnodet (approved):
Re-review after new commit: the second commit correctly addresses the previous review feedback by aligning the `putRule` error message format with `executeOperation`'s pattern. The else-branch now uses the same `String.format("Expected body of type %s but was %s", ...)` with the `ObjectHelper.isNotEmpty` null-guard, and the test assertion was updated to match. CI passes on both JDK 17 and 25.

PR #25263, opened 2026-07-30
CAMEL-24291: align host-key checking option metadata with camel-ftp and tidy SCP global/session redundancy

oscerd:
## What

Aligns the SSH-based components' host-key-checking option metadata with the SFTP `strictHostKeyChecking` policy introduced in CAMEL-23404, and tidies a contradictory global/session setting in camel-jsch.

### Metadata parity (`security="insecure:ssl"`)

- **camel-jsch** — `ScpConfiguration.strictHostKeyChecking` (default `no`) is now tagged `security="insecure:ssl"` with a description, matching SFTP.
- **camel-ssh** — `SshConfiguration.knownHostsResource` and `failOnUnknownHost` are now tagged `security="insecure:ssl"`.
  - `failOnUnknownHost` additionally carries `insecureValue="false"`: it is a boolean that is insecure when **false** (unknown host keys accepted), so without this the generated `SecurityUtils` enforcement map would default to `"true"` — the inverted value. This mirrors how `netty-http` tags `hostnameVerification`.

### SCP global/session tidy

`ScpComponent.initJsch()` set the JVM-global `JSch.setConfig("StrictHostKeyChecking", "yes")`, but `ScpOperations.createSession()` always re-applies the per-session value from `ScpConfiguration` (default `no`), so the global was dead and contradictory. It has been removed (the `JSch.setLogger` setup is kept); the per-session value governs behaviour.

## Behaviour

Non-behavioural aside from removing the dead global (the session value already won). **Default values are unchanged** — flipping the insecure defaults would be a separate behavioural change for a future major release with an upgrade-guide entry.

## Tests / build

- camel-jsch + camel-ssh module test suites pass (the SCP tests exercise the changed `initJsch` / `createSession` path).
- Full-reactor `mvn clean install -DskipTests` regenerates catalog/docs/DSL and the `SecurityUtils` enforcement map with no drift.

## Backport

**main only** — the `insecure:ssl` security-profile enforcement framework (`SecurityUtils`) does not exist on `camel-4.18.x` / `camel-4.14.x`, so the tags would have nothing to enforce there.

Closes CAMEL-24291.

---


gnodet (approved):

Well-crafted metadata alignment that correctly tags SCP and SSH host-key-checking options with `security="insecure:ssl"` to match the SFTP pattern from CAMEL-23404. The removal of the dead global `JSch.setConfig("StrictHostKeyChecking", "yes")` call in `ScpComponent` is justified — `ScpOperations.createSession()` always applies the per-session value from `ScpConfiguration` (default `"no"`), so the global was always overridden.

Key observations:
- The `insecureValue="false"` on `failOnUnknownHost` is correctly specified — without it, the default for boolean types would incorrectly flag `failOnUnknownHost=true` as insecure
- The SCP `strictHostKeyChecking` annotation is consistent with the SFTP `BaseSftpConfiguration` from CAMEL-23404
- Generated files (catalog JSON, endpoint DSL, component DSL) are properly regenerated and consistent with the annotation changes
- No upgrade guide entry needed since defaults are unchanged and no behavioral changes
- CI is green on both JDK 17 and 25

PR #25265, opened 2026-07-31
CAMEL-24320: Fix Kamelet route creation with virtual threads on JDK 25

atiaomar1978-hub:
## Summary

Fixes [CAMEL-24320](https://issues.apache.org/jira/browse/CAMEL-24320): with `-Dcamel.threads.virtual.enabled=true` on JDK 25+, Kamelet route creation failed during endpoint init with an NPE.

**Root cause:** the JDK 25 `ScopedValue`-backed `ContextValue.orElse(null)` delegated to `ScopedValue.orElse`, which rejects a null fallback via `Objects.requireNonNull`. `DefaultCamelContextExtension.getCreateRoute()` / `getCreateProcessor()` call `orElse(null)` outside any binding scope (e.g. from `KameletComponent.doInit`).

**Fix:** make `ScopedValueContextValue.orElse` null-tolerant — return the default when unbound instead of delegating to `ScopedValue.orElse`.

## Tests

- `ScopedValueContextValueOrElseTest` (JDK 25+): direct regression for null/default/bound `orElse` on the ScopedValue backend
- `CreateContextValueTest`: `getCreateRoute()` / `getCreateProcessor()` return null outside scope
- `KameletVirtualThreadsRouteCreationTest` (JDK 25+): Main starts a route consuming `kamelet:vt-repro-source` with virtual threads enabled

## Test plan

- [x] `mvn test -pl core/camel-util,core/camel-core,components/camel-kamelet -am -Dtest=ContextValueTest,ScopedValueContextValueOrElseTest,CreateContextValueTest,KameletVirtualThreadsRouteCreationTest`
- [x] Bugbot review — no findings
- [x] Grok review — production fix approved; primary guard is `ScopedValueContextValueOrElseTest`

JIRA: https://issues.apache.org/jira/browse/CAMEL-24320


atiaomar1978-hub:

## Review follow-up (Bugbot + Grok)

Independent reviews were run on branch `CAMEL-24320-kamelet-virtual-threads-fix`.

### Bugbot
No bugs found.

### Grok findings

| Severity | Issue | Status |
|----------|-------|--------|
| — | Production fix (`isBound() ? get() : default`) is correct and matches `ContextValue` contract | Accepted |
| High | Main-based Kamelet IT may not always select ScopedValue backend if `ContextValueFactory` loaded earlier in the same JVM | Mitigated — `ScopedValueContextValueOrElseTest` is the primary JDK 25 guard (reflection on `ScopedValueContextValue`); Kamelet IT sets `camel.threads.virtual.enabled` in a static initializer and asserts route exists after start |
| Low | Document null-tolerant `orElse` on `ContextValue` javadoc | Follow-up |

Reproduction confirmed per JIRA: NPE at `ScopedValue.orElse` → `Objects.requireNonNull` when `getCreateRoute()` is called from `KameletComponent.doInit` outside binding scope.

Ready for review.

apupier:
need to format:

```
diff --git a/components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java b/components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
index 13e93a86b737..d5f9f568d0a1 100644
--- a/components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
+++ b/components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
@@ -33,8 +33,8 @@ import static org.assertj.core.api.Assertions.assertThatCode;
 /**
  * CAMEL-24320: Kamelet route creation must not NPE when virtual threads are enabled on JDK 25+.
  * <p>
- * Virtual threads must be selected before the first {@code ContextValue} is created; the system property is set in
- * a static initializer so {@code ContextValueFactory} picks the ScopedValue backend when this class loads first in an
+ * Virtual threads must be selected before the first {@code ContextValue} is created; the system property is set in a
+ * static initializer so {@code ContextValueFactory} picks the ScopedValue backend when this class loads first in an
  * isolated fork. {@link org.apache.camel.util.concurrent.ScopedValueContextValueOrElseTest} is the primary unit guard.
  */
 @Isolated
```

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
```diff
@@ -0,0 +1,71 @@
...
+        System.setProperty("camel.threads.virtual.enabled", "true");
+    }
+
+    @AfterEach
+    void resetThreadType() throws Exception {
+        Field field = ThreadType.class.getDeclaredField("current");
+        field.setAccessible(true);
+        field.set(null, null);
+        System.clearProperty("camel.threads.virtual.enabled");
+    }
```
apupier: need to look for the value of the property before the test and reset it to thsi value after

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * CAMEL-24320: Kamelet route creation must not NPE when virtual threads are enabled on JDK 25+.
+ * <p>
+ * Virtual threads must be selected before the first {@code ContextValue} is created; the system property is set in
+ * a static initializer so {@code ContextValueFactory} picks the ScopedValue backend when this class loads first in an
+ * isolated fork. {@link org.apache.camel.util.concurrent.ScopedValueContextValueOrElseTest} is the primary unit guard.
+ */
+@Isolated
+@EnabledForJreRange(min = JRE.JAVA_25)
+class KameletVirtualThreadsRouteCreationTest {
```
apupier: I think we should add a Resource lock like here https://github.com/apache/camel/blob/0e101544096baf1c734354718e3b2066722fd88b/core/camel-core/src/test/java/org/apache/camel/util/FilePathResolverTest.java#L28

and maybe no need of @Isolated in this case

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
```diff
@@ -0,0 +1,71 @@
...
+            main.configure().withVirtualThreadsEnabled(true).addRoutesBuilder(new RouteBuilder() {
+                @Override
+                public void configure() {
+                    from("kamelet:vt-repro-source").routeId("vt-kamelet-repro").to("mock:vt-out");
+                }
+            });
+            main.start();
+            assertThat(main.getCamelContext().getRoute("vt-kamelet-repro")).isNotNull();
+            main.stop();
+        }).doesNotThrowAnyException();
```
apupier: there is already an assertion in the test, not sure it is necessary to use doesNotThorwAnyException. If an exception is thrown, it will be reorted as an error already by the test

atiaomar1978-hub:

Fixed the **sourcecheck** failure in [run 30603239831](https://github.com/apache/camel/actions/runs/30603239831/job/91087254528) (`Fail if there are uncommitted changes`).

The build reformatted javadoc line wrapping in `KameletVirtualThreadsRouteCreationTest.java`. Applied `formatter:format` and committed the synced file.

Re-ran `ContextValueTest`, `ScopedValueContextValueOrElseTest`, `CreateContextValueTest`, and `KameletVirtualThreadsRouteCreationTest` — all green (JDK 25 tests skipped locally on JDK 21).

Please re-run or wait for CI on the latest commit.

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
atiaomar1978-hub: Fixed in `58bc2359d1d` — `@BeforeEach` saves the previous `camel.threads.virtual.enabled` value and `@AfterEach` restores it (clear when previously unset).

atiaomar1978-hub:

Addressed @apupier review in `58bc2359d1d`:

1. Save/restore `camel.threads.virtual.enabled` around each test instead of clearing unconditionally.
2. Use `@ResourceLock(Resources.SYSTEM_PROPERTIES)` instead of `@Isolated`.
3. Drop redundant `assertThatCode` wrapper.

Re-ran `ContextValueTest`, `ScopedValueContextValueOrElseTest`, `CreateContextValueTest`, and `KameletVirtualThreadsRouteCreationTest` — all green (JDK 25 tests skipped locally on JDK 21).

Ready for re-review.

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
atiaomar1978-hub: Fixed in `58bc2359d1d` — replaced `@Isolated` with `@ResourceLock(Resources.SYSTEM_PROPERTIES)` on the test class.

components/camel-kamelet/src/test/java/org/apache/camel/component/kamelet/KameletVirtualThreadsRouteCreationTest.java
atiaomar1978-hub: Fixed in `58bc2359d1d` — removed `assertThatCode`; the test now uses direct assertions and a `try/finally` around `main.stop()`.

gnodet (commented):
The production fix is correct — changing `scopedValue.orElse(defaultValue)` to `scopedValue.isBound() ? scopedValue.get() : defaultValue` properly handles JDK 25's `ScopedValue.orElse` null-rejection. The fix is backward compatible and does not change behavior on JDK 17-24 or when virtual threads are disabled.

**Highlights:**

- The root cause is well-understood: JDK 25 finalized the `ScopedValue` API and tightened `orElse()` to call `Objects.requireNonNull(other)`, breaking the null default case used by Kamelet route creation.
- The core fix is correctly tested by the reflection-based unit tests in `ScopedValueContextValueOrElseTest` that directly construct `ScopedValueContextValue` instances, bypassing the factory.
- All three of @apupier's review comments have been addressed.

**Minor test observations (non-blocking):**

- `orElseReturnsBoundValue()` in `ScopedValueContextValueOrElseTest` goes through `ContextValue.newInstance()` without setting `camel.threads.virtual.enabled=true`, so it exercises `ThreadLocalContextValue` rather than `ScopedValueContextValue`. Low practical impact since the first two tests cover the ScopedValue backend directly, and the bound-value case was never broken.
- The `ScopedValueHolder` lazy-holder pattern means the Kamelet integration test may silently use the ThreadLocal backend if another test triggers class loading first. This is acknowledged as a known limitation.

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### Review follow-up (`51eb0551097`)

**@apupier** — all three inline items were addressed in `58bc2359d1d`:
1. Save/restore `camel.threads.virtual.enabled` via `@BeforeEach` / `@AfterEach`
2. `@ResourceLock(Resources.SYSTEM_PROPERTIES)` instead of `@Isolated`
3. Removed redundant `assertThatCode` wrapper

This push adds:
- **`ContextValue.orElse` javadoc** — documents that `null` is a valid fallback (the Kamelet/`getCreateRoute()` call path)
- **`CreateContextValueTest.getCreateRouteDoesNotThrowWithVirtualThreadsEnabled`** (JDK 25+) — exercises `DefaultCamelContextExtension.getCreateRoute()` / `getCreateProcessor()` with virtual threads enabled and property save/restore

**Local tests:** `ContextValueTest`, `CreateContextValueTest` — green. JDK 25 tests (`ScopedValueContextValueOrElseTest`, `KameletVirtualThreadsRouteCreationTest`) require JDK 25 build profile (CI).

Ready for re-review.

gnodet (approved):
Re-review after latest commit (`51eb0551097`). The production fix is correct, minimal (1 line in `ScopedValueContextValue.orElse()`), and backward compatible. All three @apupier review comments have been properly addressed:

- ✅ Save/restore `camel.threads.virtual.enabled` around tests
- ✅ `@ResourceLock(Resources.SYSTEM_PROPERTIES)` instead of `@Isolated` on the Kamelet test
- ✅ Removal of redundant `assertThatCode` wrapper

The new commit adds useful javadoc documentation and an additional JDK 25 test guard for `getCreateRoute()`/`getCreateProcessor()`.

**Bug analysis:** JDK 25 finalized `ScopedValue.orElse()` to reject null via `Objects.requireNonNull`. The call path `KameletComponent.doInit()` → `getCreateRoute()` → `isCreateRoute.orElse(null)` outside any binding scope triggers the NPE when virtual threads are enabled. The fix properly uses `isBound() ? get() : defaultValue` instead.

**Test coverage:** Three-layer approach — unit test via reflection on `ScopedValueContextValue`, core-level test on `getCreateRoute()/getCreateProcessor()`, and integration test starting a Kamelet route with `Main`. Sound strategy.

**Minor notes (informational, not blocking):**
- `orElseReturnsBoundValue()` exercises `ThreadLocalContextValue` rather than `ScopedValueContextValue` since it goes through `ContextValue.newInstance()` without the system property — the reflection-based tests compensate.
- CI likely needs manual approval for this first-time contributor. The production fix should be verified on a JDK 25 build.

davsclaus (approved):

atiaomar1978-hub:
_AI-generated comment on behalf of [atiaomar1978-hub](https://github.com/atiaomar1978-hub)_

### CI fix for [run 30758229647](https://github.com/apache/camel/actions/runs/30758229647/job/91538215704) (`cbea25eb79a`)

**Root cause:** `ScopedValueContextValueOrElseTest` used reflection to load `ContextValueFactory$ScopedValueContextValue`, but that inner class is packaged in the MRJ layer (`META-INF/versions/25/`) and is not available on Surefire's exploded test classpath on JDK 25 → `ClassNotFoundException` in `camel-util`.

**Changes:**
- Removed `ScopedValueContextValueOrElseTest` (MRJ inner classes cannot be unit-tested reliably via reflection on the Surefire classpath)
- `ContextValueFactory` (java25): evaluate ScopedValue vs ThreadLocal **lazily** from `ThreadType.current()` instead of caching at class-load time
- Added `Multi-Release: true` manifest for the java-25 profile (`META-INF/MANIFEST.MF` + jar plugin)
- `KameletVirtualThreadsRouteCreationTest` remains the integration regression guard for the `orElse(null)` NPE (CAMEL-24320)

**Local tests:**
- JDK 21: `ContextValueTest`, `CreateContextValueTest`, `KameletVirtualThreadsRouteCreationTest` (JDK 25-only tests skipped) — green
- JDK 26: `ContextValueTest`, `CreateContextValueTest` — green

Please re-run CI on the latest commit.

PR #25277, opened 2026-07-31
Improve assertions

apupier:
initially looking to rule "Add an assertion predicate after calling this method." https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&impactSoftwareQualities=RELIABILITY&rules=java%3AS5833&issueStatuses=OPEN%2CCONFIRMED&id=apache_camel

I improved the 2 incriminated files to get rid of all Sonar issue reports

# Description

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/components/camel-ai/camel-langchain4j-agent/src/test/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentWithMemoryServiceTest.java
+++ b/components/camel-ai/camel-langchain4j-agent/src/test/java/org/apache/camel/component/langchain4j/agent/LangChain4jAgentWithMemoryServiceTest.java
@@ -72,7 +72,7 @@ public class LangChain4jAgentWithMemoryServiceTest extends BaseLangChain4jAgent 
     }
 
     @Test
-    public void testToolThenMemoryThenAnotherTool() throws Exception {
+    void testToolThenMemoryThenAnotherTool() throws Exception {
         MockEndpoint mockEndpoint = this.context.getEndpoint("mock:agent-response", MockEndpoint.class);
         mockEndpoint.expectedMessageCount(3);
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Correctly fixes SonarCloud rule S5833 by moving `withFailMessage()`/`.as()` before assertion predicates (they were dead code when placed after), reduces test visibility to package-private per JUnit 5 conventions, and improves assertion idioms. CI is green.

**Highlights:**
- Nice use of `containsIgnoringCase("rome")` in `SpringAiChatOllamaIT` instead of the previous `assertThat(response.toLowerCase()).contains("rome")` — this is cleaner and more idiomatic AssertJ.
- Worth noting that `withFailMessage` in AssertJ completely replaces the default error message, including the actual/expected values. The original code had these calls placed AFTER the assertion predicates (making them dead code — they never affected any assertion). Now that they are correctly placed before, failure messages will show only the custom text, not AssertJ's detailed diff. This is a semantic change but reflects the original author's intent.
- The removal of `.as()` calls from `SpringAiChatOllamaIT` is correct — they were similarly dead code when placed after assertion predicates.

1. `LangChain4jAgentWithMemoryServiceTest` class declaration still has the `public` modifier. Since the PR already removes `public` from the test method, the class declaration could also be made package-private per project conventions. The second file (`SpringAiChatOllamaIT`) correctly applies this.
2. The `testToolThenMemoryThenAnotherTool` method mixes JUnit assertions (`assertNotNull`, `assertTrue`, `assertFalse`) with AssertJ assertions. Since the PR touches assertions in this method, migrating the adjacent `assertNotNull` to `assertThat(...).isNotNull()` would align with the "don't mix styles in the same method" convention.

```suggestion
class LangChain4jAgentWithMemoryServiceTest extends BaseLangChain4jAgent {
```

PR #25278, opened 2026-07-31
CAMEL-24324: camel-aws-bedrock - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

Most `BedrockAgent` / `BedrockAgentRuntime` operations already fail fast when
`pojoRequest=true` and the body is not the expected request type. Four did not —
they fell through the `instanceof` with no `else`, so no AWS call was made, no
response was set, and the original body was returned with no error:

- `BedrockAgentProducer`: `startIngestionJob`, `listIngestionJobs`, `getIngestionJob`
- `BedrockAgentRuntimeProducer`: `retrieveAndGenerate`

(`BedrockProducer` in the runtime package is already covered by CAMEL-23462.)

## Fix

Each of the four branches now throws `IllegalArgumentException`, matching the
wording the sibling operations in the same module already use, e.g.
`"retrieveAndGenerate operation requires a RetrieveAndGenerateRequest body when
pojoRequest=true"`.

## Tests

`camel-aws-bedrock` has no producer route/mock harness, so this is covered by two
new Mockito unit tests that mock the endpoint + client, send a wrong-typed body
with `pojoRequest=true`, and assert the `IllegalArgumentException`. Each was
verified to fail (silent no-op) before the fix. Region-free — no localstack or
real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/agent/BedrockAgentProducer.java
+++ b/components/camel-aws/camel-aws-bedrock/src/main/java/org/apache/camel/component/aws2/bedrock/agent/BedrockAgentProducer.java
@@ -100,6 +100,9 @@ public class BedrockAgentProducer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 prepareIngestionJobResponse(result, message);
+            } else {
+                throw new IllegalArgumentException(
+                        "startIngestionJob operation requires a StartIngestionJobRequest body when pojoRequest=true");
             }
         } else {
             String knowledgeBaseId;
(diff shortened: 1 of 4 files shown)


gnodet (approved):
Clean, well-scoped fix that adds missing `else` branches to throw `IllegalArgumentException` when `pojoRequest=true` and the body is the wrong type for four Bedrock Agent/AgentRuntime operations that previously silently did nothing. Tests are well-written and follow project conventions.

**Highlights:**
- The fix correctly mirrors the pattern already established by the `retrieve` and `invokeFlow` operations in the same `BedrockAgentRuntimeProducer` class. Error message wording is consistent across all agent/agentruntime operations.
- Tests properly use AssertJ assertions and package-private visibility, following project conventions.
- The parameterized test approach for the three agent operations is a nice touch, reducing boilerplate.

**Minor nit (non-blocking):** Each parameterized test invocation creates a new `DefaultCamelContext` without closing it. While the context is never started and this is a lightweight unit test, wrapping in try-with-resources would be slightly cleaner for resource hygiene.

PR #25281, opened 2026-07-31
CAMEL-23129: Fix CountDownLatch count mismatch in ThreadPerTaskSedaConsumer

gnodet:
## Summary

Fix the root cause of the SEDA virtual thread shutdown hang (CAMEL-23129). The previous fix added a timeout to `latch.await()` as a defensive workaround, but the underlying mismatch remained.

**Root cause:** `SedaConsumer.doStart()` creates `latch = new CountDownLatch(concurrentConsumers)`, but `ThreadPerTaskSedaConsumer` uses a single coordinator thread. When `concurrentConsumers > 1` (used as a concurrency limit, e.g., 2), the latch is initialized with count=2 but only 1 coordinator thread ever calls `countDown()` — so `prepareShutdown()` always waits the full shutdown timeout before proceeding.

**Fix:** Override `doStart()` in `ThreadPerTaskSedaConsumer` to set `latch = new CountDownLatch(1)`, matching the single coordinator thread. The `concurrentConsumers` value continues to serve as the task executor concurrency limit via the `Semaphore`. The task executor shutdown is handled separately in `ThreadPerTaskSedaConsumer.prepareShutdown()`.

### Changes
- `SedaConsumer`: Change `latch` field visibility from `private` to `protected` so subclasses can override the count
- `ThreadPerTaskSedaConsumer`: Override `doStart()` to set latch count to 1
- `ThreadPerTaskSedaConsumerTest`: Add test verifying shutdown with concurrency limit completes quickly (no full timeout wait); apply JUnit 5 conventions (drop `public`)

## Test plan

- [x] Existing `ThreadPerTaskSedaConsumerTest` tests pass (virtualThreadPerTask basic, concurrency limit, high throughput)
- [x] New `testShutdownWithConcurrencyLimitCompletesQuickly` verifies context stop completes in < 30s with `concurrentConsumers=2`
- [ ] CI validation

### DIFF
--- a/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
+++ b/components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
@@ -49,7 +49,7 @@ public class SedaConsumer extends DefaultConsumer implements Runnable, ShutdownA
     private static final Logger LOG = LoggerFactory.getLogger(SedaConsumer.class);
 
     private final AtomicInteger taskCount = new AtomicInteger();
-    private volatile CountDownLatch latch;
+    protected volatile CountDownLatch latch;
     private volatile boolean shutdownPending;
     private volatile boolean forceShutdown;
     private ExecutorService executor;
--- a/core/camel-core/src/test/java/org/apache/camel/component/seda/ThreadPerTaskSedaConsumerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/seda/ThreadPerTaskSedaConsumerTest.java
@@ -16,18 +16,22 @@
  */
 package org.apache.camel.component.seda;
 
+import java.util.concurrent.TimeUnit;
+
 import org.apache.camel.ContextTestSupport;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.mock.MockEndpoint;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 /**
  * Test for the virtualThreadPerTask mode of SEDA consumer
  */
-public class ThreadPerTaskSedaConsumerTest extends ContextTestSupport {
+class ThreadPerTaskSedaConsumerTest extends ContextTestSupport {
 
     @Test
-    public void testVirtualThreadPerTask() throws Exception {
+    void testVirtualThreadPerTask() throws Exception {
         MockEndpoint mock = getMockEndpoint("mock:result");
         mock.expectedMessageCount(10);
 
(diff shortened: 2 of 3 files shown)


components/camel-seda/src/main/java/org/apache/camel/component/seda/SedaConsumer.java
```diff
@@ -49,7 +49,7 @@ public class SedaConsumer extends DefaultConsumer implements Runnable, ShutdownA
     private static final Logger LOG = LoggerFactory.getLogger(SedaConsumer.class);
 
     private final AtomicInteger taskCount = new AtomicInteger();
-    private volatile CountDownLatch latch;
+    protected volatile CountDownLatch latch;
```
gnodet, line 52: **Design nit:** Changing `private` → `protected` on the `latch` field exposes mutable internal state to all subclasses. A template method would achieve the same goal without field exposure:

```java
// In SedaConsumer:
protected int getLatchCount() {
    return getEndpoint().getConcurrentConsumers();
}

// In doStart():
int count = getLatchCount();
if (count > 0) {
    latch = new CountDownLatch(count);
}
```

Then `ThreadPerTaskSedaConsumer` just overrides `getLatchCount()` to return `1` — no need to touch `latch` directly, and it eliminates the "create then immediately replace" pattern where `super.doStart()` allocates a `CountDownLatch(N)` that is discarded one line later.

Non-blocking — the current approach is functionally correct.

core/camel-core/src/test/java/org/apache/camel/component/seda/ThreadPerTaskSedaConsumerTest.java
```diff
@@ -63,6 +67,32 @@ public void testVirtualThreadPerTaskHighThroughput() throws Exception {
...
+        // count (2) but only 1 coordinator thread counts down, so prepareShutdown()
+        // would wait the full shutdown timeout before proceeding.
+        long start = System.nanoTime();
+        context.stop();
+        long elapsed = TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - start);
+
+        // Shutdown should complete well within the default timeout (300s).
+        // Use a generous 30s bound to avoid flakiness, but this is still much less
+        // than the full shutdown strategy timeout that would be hit without the fix.
+        assertTrue(elapsed < 30, "Context stop took " + elapsed + "s, expected < 30s. "
```
gnodet, line 92: **Convention nit:** Per project conventions, new test code should use AssertJ assertions. Consider:

```java
assertThat(elapsed)
    .as("Context stop took %ds; CountDownLatch count likely mismatches coordinator thread count", elapsed)
    .isLessThan(30L);
```

This provides better failure messages and aligns with the project's preference for AssertJ over JUnit assertions.

gnodet (commented):
Correct root cause fix — the coordinator thread count is 1 but the latch was initialized with `concurrentConsumers` (e.g. 2), causing shutdown to always wait the full timeout. The fix works and the test is well-designed. Two minor suggestions below.

oscerd (approved):

davsclaus (approved):

PR #25282, opened 2026-07-31
 CAMEL-23931: Fix -Dcamel.jbang.quarkus.platform.url system property ignored in export

jomin7:
## Summary
  
  The `-Dcamel.jbang.quarkus.platform.url` system property is ignored when exporting
  with `--runtime quarkus` in Camel 4.18.3, causing the Quarkus platform version to
  not be resolved from the custom registry. This worked correctly in 4.18.1.

   The `-Dcamel.jbang.quarkus.platform.url` system property was ignored when exporting with `--runtime quarkus`.
  As part of the changes in CAMEL-23828, registry resolution could be skipped when quarkusVersion was already set, even if a custom platform URL was configured.

  ## Changes
  - **`Export.java`**: Read `quarkusPlatformUrl` from `application.properties` and
    system properties, and propagate it to `ExportQuarkus`
  - **`ExportBaseCommand.java`**: Added `quarkusPlatformUrl` field
  - **`ExportQuarkus.java`**: Allow registry resolution when a custom platform URL is
    configured (`quarkusPlatformUrl != null`), even if `quarkusVersion` is already set
  - **`QuarkusHelper.java`**: Updated `getQuarkusPlatformUrl()`, `resolveQuarkusPlatformVersion()`,
    and `fetchPlatformStreams()` to accept a user-supplied URL, giving it priority over
    the system property and default

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Export.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/commands/Export.java
@@ -28,6 +28,7 @@ import java.util.Properties;
 
 import org.apache.camel.dsl.jbang.core.common.CamelJBangConstants;
 import org.apache.camel.dsl.jbang.core.common.PropertyResolver;
+import org.apache.camel.dsl.jbang.core.common.QuarkusHelper;
 import org.apache.camel.dsl.jbang.core.common.RuntimeType;
 import org.apache.camel.dsl.jbang.core.common.RuntimeUtil;
 import org.apache.camel.dsl.jbang.core.common.SourceScheme;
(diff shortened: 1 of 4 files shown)


gnodet (commented):
The fix correctly addresses the CAMEL-23931 regression where `-Dcamel.jbang.quarkus.platform.url` was ignored during Quarkus export. The implementation approach — threading the platform URL through Export → ExportQuarkus → QuarkusHelper and widening the resolution guard to `if (quarkusVersion == null || quarkusPlatformUrl != null)` — is sound.

**One concern:**

- The PR lacks test coverage for the exact scenario it fixes. The CAMEL-23828 commit added three tests for `--quarkus-version` handling, but none exercise the custom platform URL path. A test (e.g., using `@SetSystemProperty(key = "camel.jbang.quarkus.platform.url", value = "...")` or setting `quarkusPlatformUrl` directly on the command) that verifies the custom platform URL code path is exercised would prevent this regression from recurring. Per project guidelines, every PR should include tests for bug fixes.

**Notes:**

- The fix does not need to be forward-ported to `main`, where the Quarkus registry resolution was completely redesigned via `QuarkusPlatformMixin`/`QuarkusExtensionRegistryMixin` (CAMEL-23560).
- The `getQuarkusPlatformUrl()` signature change from no-arg to single-arg is safe — all callers are internal to `camel-jbang-core`.

davsclaus (commented):
Thanks for tracking down this regression, @jomin7 — the fix correctly threads `quarkusPlatformUrl` through the call chain and widens the resolution guard so that a custom platform URL triggers registry resolution even when `quarkusVersion` is already set.

**One concern:**

- **Missing test coverage** — CAMEL-23828 added three tests for `--quarkus-version` handling, but none exercise the custom platform URL path. A test (e.g., using `@SetSystemProperty(key = "camel.jbang.quarkus.platform.url", value = "...")` or setting `quarkusPlatformUrl` directly on the command) that verifies this code path would prevent this specific regression from recurring. Per project guidelines, every PR should include tests for bug fixes.

**Minor:**

- The PR title has a leading space (`" CAMEL-23931:"` → `"CAMEL-23931:"`).

**Notes:**

- The `getQuarkusPlatformUrl()` → `getQuarkusPlatformUrl(String)` signature change is safe — all callers are internal to `camel-jbang-core`.
- No forward-port to `main` is needed — the Quarkus registry resolution was redesigned there via `QuarkusPlatformMixin`/`QuarkusExtensionRegistryMixin`.

davsclaus (approved):

PR #25283, opened 2026-07-31
CAMEL-24284: Reset recoverTask after successful SJMS connection recovery

gansheer:
# Description

Fix a bug where the SJMS consumer becomes permanently dead after recovering from a JMS connection failure once — any subsequent connection failure cannot trigger recovery.

**Root cause**

[CAMEL-23646](https://issues.apache.org/jira/browse/CAMEL-23646) added a !recoverTask.isRunning() guard in scheduleConnectionRecovery() to allow re-scheduling after a completed task. However, BackgroundTask.schedule() sets running=true but never resets it (running is only cleared in waitForTaskCompletion(), which is the run() path, not the schedule() path).

  After the first successful recovery:
  1. recoverTask is non-null and isRunning() returns true forever
  2. The guard recoverTask == null || !recoverTask.isRunning() is permanently false
  3. If onException() fires again (transient broker event, Artemis internal cleanup), consumers/sessions are nulled with no recovery path
  4. Consumer is permanently dead

 **Fix**

Reset recoverTask and recoverFuture to null (and cancel the scheduled future) after successful recovery. This re-arms the recovery mechanism so subsequent failures can schedule a new recovery  task.

Claude Code on behalf of gansheer

### DIFF
--- a/components/camel-sjms/src/main/java/org/apache/camel/component/sjms/consumer/SimpleMessageListenerContainer.java
+++ b/components/camel-sjms/src/main/java/org/apache/camel/component/sjms/consumer/SimpleMessageListenerContainer.java
@@ -205,6 +205,16 @@ public class SimpleMessageListenerContainer extends ServiceSupport
         try {
             refreshConnection();
             initConsumers();
+            connectionLock.lock();
+            try {
+                if (recoverFuture != null) {
+                    recoverFuture.cancel(false);
+                }
+                recoverTask = null;
+                recoverFuture = null;
+            } finally {
+                connectionLock.unlock();
+            }
             LOG.debug("Successfully recovered JMS Connection (attempt: {})", task.iteration());
             // success so do not try again
             return true;


gnodet (approved):
Correct and well-analyzed fix for a real bug where `BackgroundTask.schedule()` sets `running=true` but never resets it, causing the re-scheduling guard to permanently block after the first successful recovery. The fix properly uses `connectionLock` and `cancel(false)` for thread safety.

**Highlights:**
- The bug analysis in the PR description is excellent: it accurately identifies that `BackgroundTask.schedule()` sets `running=true` (line 144 of `BackgroundTask.java`) but only resets it in `waitForTaskCompletion()` (line 198), which is only reachable via the `run()` path, not the `schedule()` path. The fix of nulling `recoverTask` so the guard passes via the `recoverTask==null` path is the correct caller-side workaround.
- Thread safety is properly handled: the fix acquires `connectionLock` before modifying `recoverTask`/`recoverFuture`, consistent with `scheduleConnectionRecovery()` and other methods in this class. The use of `cancel(false)` (rather than `cancel(true)`) is correct since this code runs inside the scheduled task itself — `cancel(true)` would cause self-interruption.

1. The existing `SjmsConnectionRecoveryTest` only triggers `onException` once, so it passes both with and without this fix. A test that triggers `onException` twice (with successful recovery between them) would directly cover the bug. Not a blocker given the fix is small and obviously correct.
2. Pre-existing: `doStop()` at line 264 accesses `recoverFuture` and `recoverTask` without holding `connectionLock`, unlike `scheduleConnectionRecovery()` and this fix. Mitigated by service lifecycle ordering, but worth noting for future hardening.

gnodet (approved):
Correct and well-targeted fix for a real bug where `BackgroundTask.schedule()` sets `running=true` permanently, preventing re-scheduling after the first successful SJMS connection recovery.

**Highlights:**

- The root cause analysis is thorough: `BackgroundTask.schedule()` sets `running=true` but only `waitForTaskCompletion()` resets it, and that method is only called from the `run()` path, not `schedule()`. The fix correctly nulls `recoverTask`/`recoverFuture` so the guard passes via the `recoverTask == null` path.
- Thread safety is properly handled: `connectionLock` is acquired before modifying the fields (consistent with `scheduleConnectionRecovery`), and `cancel(false)` is used instead of `cancel(true)` because the code executes inside the scheduled task itself.
- Defense in depth: even if `cancel(false)` didn't prevent the next invocation, `BackgroundTask.runTaskWrapper()` checks `latch.getCount() == 0` and short-circuits.
- The local `task` parameter reference (used in LOG.debug after null assignment) is safe since it's a method parameter, not the field.

davsclaus (approved):

PR #25287, opened 2026-08-03
[backport camel-4.18.x] CAMEL-24315: camel-google-pubsub - stop subscribers that finish starting after the consumer stopped

davsclaus:
## Backport of #25248

Cherry-pick of #25248 onto `camel-4.18.x`.

**Original PR:** #25248 - CAMEL-24315: camel-google-pubsub - stop subscribers that finish starting after the consumer stopped
**Original author:** @henrik242

### Original description

Since CAMEL-22898 (4.18.0), GooglePubsubConsumer adds a subscriber to the subscribers list only after startAsync().awaitRunning() returns. doStop() and deferShutdown() only stop subscribers already in that list, so stopping the consumer while a subscriber is still starting misses it. Nothing ever calls stopAsync() on it, and the wrapper thread parks forever in awaitTerminated(), an uninterruptible Guava wait.

The consumer executor then never terminates: context shutdown waits 2 x shutdownAwaitTermination (default 10s) per stuck pool and the thread leaks even after forced shutdown. Test suites that stop many CamelContexts (e.g. Spring Boot integration tests) pay up to 20s per leaked consumer per context; we saw a CI build go from 5 to 25 minutes from this alone.

Fix: after registering the subscriber or pull future, re-check the consumer state and stop/cancel it if the consumer is no longer allowed to run.

### DIFF
--- a/components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
+++ b/components/camel-google/camel-google-pubsub/src/main/java/org/apache/camel/component/google/pubsub/GooglePubsubConsumer.java
@@ -255,6 +255,10 @@ public class GooglePubsubConsumer extends DefaultConsumer implements ShutdownAwa
                     // Only add to list after successful startup
                     subscribers.add(subscriber);
                     subscriberAdded = true;
+                    // a stop while the subscriber was starting missed it; stop it here or awaitTerminated never returns
+                    if (!isRunAllowed() || isSuspendingOrSuspended()) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean backport of PR #25248 to camel-4.18.x. The production code changes and test file are identical to the original (same git blob). The two insertion points in `GooglePubsubConsumer.java` apply to identical surrounding code on camel-4.18.x. Target branch is correct since the regression was introduced in 4.18.0 by CAMEL-22898.

oscerd (approved):

PR #25301, opened 2026-08-03
CAMEL-24309: camel-ai-tool - AiToolRegistry listener SPI for tool registration changes

Croway:
# CAMEL-24309: AiToolRegistry listener SPI for tool registration changes

Sub-task of [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308) (camel-mcp-server). JIRA: [CAMEL-24309](https://issues.apache.org/jira/browse/CAMEL-24309)

## Changes

Adds a listener SPI to the `camel-ai-tool` registry so adapters can react to tools appearing/disappearing instead of polling — the prerequisite for the MCP server's `tools/list_changed` notification.

- New `AiToolRegistryListener` interface with `toolRegistered(String tag, AiToolSpec spec)` / `toolDeregistered(String tag, AiToolSpec spec)`; `tag == null` denotes the default (untagged) pool, and a multi-tag endpoint fires one event per tag.
- `AiToolRegistry.addListener/removeListener`; callbacks fire **outside** the registry lock (so a listener can safely re-read the registry), only on actual state changes (idempotent re-put of the same spec and removal of an absent spec fire nothing, a rejected duplicate-name registration fires nothing), and a throwing listener is logged without affecting the registration or other listeners.
- Javadoc documents the subscribe-then-snapshot idiom for observing current state without missing concurrent events.

Existing adapters (langchain4j-agent, spring-ai-chat) re-query the registry per exchange and are unaffected.

## Testing

- `AiToolRegistryListenerTest` — registry-level semantics (8 tests: event on put/remove, default pool null tag, no event on no-op or rejected mutations, listener isolation, removeListener).
- `AiToolRegistryListenerLifecycleTest` — events driven by real `ai-tool` route lifecycle (4 tests: start/stop, suspend/resume, multi-tag, untagged default pool).

All 12 new tests pass; full `camel-ai-tool` module test run is green.

---
_This PR was written by Claude Code on behalf of Federico Mariani (@Croway)._

### DIFF
--- a/components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tool/AiToolRegistry.java
+++ b/components/camel-ai/camel-ai-tool/src/main/java/org/apache/camel/component/ai/tool/AiToolRegistry.java
@@ -19,11 +19,15 @@ package org.apache.camel.component.ai.tool;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.camel.CamelContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * CamelContext-scoped registry mapping tags to {@link AiToolSpec} instances. AI components (LangChain4j, Spring AI)
@@ -32,6 +36,9 @@ import org.apache.camel.CamelContext;
  * Each {@link CamelContext} gets its own registry instance, registered as a context plugin. Use
  * {@link #getOrCreate(CamelContext)} to obtain the instance for a given context.
  * <p>
+ * Adapters that need to react to tools appearing or disappearing (e.g. to push MCP {@code tools/list_changed}
+ * notifications) can register an {@link AiToolRegistryListener} instead of polling.
+ * <p>
  * Replaces the duplicated {@code CamelToolExecutorCache} singletons from {@code camel-langchain4j-tools} and
  * {@code camel-spring-ai-tools}.
  *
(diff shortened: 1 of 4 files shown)


gnodet (approved):

This is a well-crafted SPI addition. The implementation is clean and the test coverage is thorough. Approving with a few minor suggestions below.

### What works well

- **Lock discipline**: Notifications fire outside the `ReentrantLock`, so listeners can safely re-read the registry (e.g., call `getTools()`) without deadlocking. This is documented in the `AiToolRegistryListener` Javadoc.
- **Error isolation**: A throwing listener is caught and logged without affecting the registration itself or other listeners. This is verified by `testListenerExceptionDoesNotBreakRegistrationOrOtherListeners`.
- **Idempotent no-ops**: Re-adding the same spec instance fires no duplicate event; removing an absent spec fires nothing. Both are tested.
- **Test quality**: Tests follow all Camel conventions -- package-private classes, AssertJ, no `Thread.sleep()`, no JUnit `public`. The lifecycle test uses `CopyOnWriteArrayList` for thread safety while the unit test correctly uses plain `ArrayList`.
- **`@since 4.22`** on the new `AiToolRegistryListener` interface.

### Minor suggestions

1. **`removeListener` missing Javadoc** — For consistency with `addListener` (which has a doc block), consider adding a brief Javadoc comment, e.g.:
   ```java
   /**
    * Removes a previously added listener. No-op if the listener was not registered.
    */
   public void removeListener(AiToolRegistryListener listener) {
   ```

2. **No null guard on `addListener`** — If `null` is passed, `CopyOnWriteArrayList` accepts it silently. Later, during notification, `listener.toolRegistered(...)` throws NPE which is caught and logged as a generic warning — making the root cause hard to diagnose. An early `Objects.requireNonNull` would fail fast:
   ```java
   public void addListener(AiToolRegistryListener listener) {
       Objects.requireNonNull(listener, "listener");
       listeners.add(listener);
   }
   ```

3. **Concurrency ordering caveat** — Since notifications fire outside the lock, two concurrent threads performing `put` and `remove` on the same spec could theoretically deliver `deregistered` before `registered`. In practice Camel route lifecycle is sequential per route, so this is unlikely. But since the Javadoc already documents the "subscribe-then-snapshot" idiom (showing attention to concurrency semantics), consider adding a note: _"Notification ordering is guaranteed within a single thread but not across concurrent threads."_

All minor — none are blockers. Nice work!

davsclaus (approved):

zbendhiba:
LGTM!

@JiriOndrusek  please review if there's an impact for Camel quarkus ?

JiriOndrusek:
LGTM as well, I don't see any impact on Quarkus. I see a small thing, which may help users of quarkus - in CQ to auto-discover CDI beans implementing AiToolRegistryListener and register  them with the AiToolRegistry at runtime-init, but it is more like nice to have.

zbendhiba (approved):

PR #25305, opened 2026-08-03
CAMEL-24326: camel-aws-config - throw when pojoRequest=true and the body is the wrong type

oscerd:

## Problem

With `pojoRequest=true`, `AWSConfigProducer`'s five operations (`putConfigRule`,
`removeConfigRule`, `describeRuleCompliance`, `putConformancePack`,
`removeConformancePack`) only acted when the body was the matching request type;
any other body fell through the `instanceof` with no `else`, so no AWS call was
made, no response was set, and the original body was returned with no error.

## Fix

type, e.g. `"putConformancePack operation requires PutConformancePackRequest in
POJO mode"`.

## Test

`camel-aws-config` has no producer route/mock harness, so this is covered by a new
Mockito parameterized unit test that mocks the endpoint + client, sends a
wrong-typed body with `pojoRequest=true`, and asserts each `IllegalArgumentException`.
Verified to fail (silent no-op) before the fix. Region-free — no localstack or
real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws-config/pom.xml
+++ b/components/camel-aws/camel-aws-config/pom.xml
@@ -84,5 +84,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/components/camel-aws/camel-aws-config/src/main/java/org/apache/camel/component/aws/config/AWSConfigProducer.java
+++ b/components/camel-aws/camel-aws-config/src/main/java/org/apache/camel/component/aws/config/AWSConfigProducer.java
@@ -109,6 +109,9 @@ public class AWSConfigProducer extends DefaultProducer {
                 }
                 Message message = getMessageForResponse(exchange);
                 message.setBody(result);
+            } else {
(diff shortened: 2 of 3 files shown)


gnodet (approved):

Clean and well-structured PR. The validation logic, exception type, and error message format are all consistent with the precedent set by `BedrockProducer` under the parent issue CAMEL-24261.

**What I verified:**

- All 5 operation methods (`putConfigRule`, `removeConfigRule`, `describeRuleCompliance`, `putConformancePack`, `removeConformancePack`) now have `else` branches that throw `IllegalArgumentException` when `pojoRequest=true` and the body is not the expected SDK request type.
- `IllegalArgumentException` is already the validation exception used throughout `AWSConfigProducer` (lines 70, 121, 165, 235, 250, 287), so this is consistent.
- Error message format (`<operation> operation requires <SdkType> in POJO mode`) matches the bedrock component.
- The test uses package-private visibility (JUnit 5 convention), AssertJ assertions (project convention), and a parameterized `@CsvSource` covering all 5 operations.
- The lightweight Mockito approach (mock endpoint + client, call `process()` directly) is appropriate here -- no CamelContext startup overhead needed since the exception is thrown before any AWS client call.
- `assertj-core` is correctly added without a version (managed by the parent POM).

No issues found. LGTM.

davsclaus (approved):

PR #25308, opened 2026-08-03
[JBANG-IT] Track PID in nohup execution

jvrubel:
Track PID in nohup execution so that it can be properly cleaned if needed in process-based mode.

### DIFF
--- a/dsl/camel-jbang/camel-jbang-it/src/test/java/org/apache/camel/dsl/jbang/it/support/JBangTestSupport.java
+++ b/dsl/camel-jbang/camel-jbang-it/src/test/java/org/apache/camel/dsl/jbang/it/support/JBangTestSupport.java
@@ -148,7 +148,7 @@ public abstract class JBangTestSupport {
     }
 
     protected String execNohup(final String command) {
-        return containerService.executeGenericCommand("nohup " + getMainCommand() + " " + command + " &");
+        return containerService.executeNohup(command);
     }
 
     protected String mountPoint() {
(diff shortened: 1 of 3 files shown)


gnodet (approved):

This PR fixes a real process-leak issue: when `execNohup` launched a Camel CLI command via `nohup ... &`, the resulting PID was never recorded in `backgroundPids`, so `stopBackgroundProcesses()` during teardown could not kill it.

**What the PR does well:**

- Follows the existing PID-tracking pattern established by `executeBackground`.
- Uses the standard `echo $!` shell idiom to capture the PID of the most recently backgrounded process -- this is reliable in non-interactive bash.
- Introduces the `executeNohup` default method on the `CliService` interface, so the container-based implementation (`CliLocalContainerService`) keeps the old behavior without code changes. Containers manage their own process lifecycle, so PID tracking there is unnecessary.
- Both callers (`JolokiaITCase` and `CamelDebugITCase`) discard the return value, so changing it from full command output to a PID string is safe.

**Analysis of correctness:**

- `executeGenericCommand` reads stdout and stderr separately. Since `nohup`'s informational message goes to stderr, stdout will contain only the `echo $!` output.
- In non-interactive bash (`/bin/bash -c`), there are no job-control messages, so `pidOutput.trim()` should yield a clean numeric PID.
- The `isNumber()` guard degrades gracefully: if output is unexpectedly non-numeric, the PID simply is not tracked -- no worse than the old behavior of never tracking.
- Cleanup via `stopBackgroundProcesses()` already handles the tracked PIDs using `ProcessHandle.destroyForcibly()`.

**Minor nit (non-blocking):** The existing `executeBackground` method has a slightly more robust PID parsing fallback using `StringHelper.before(pid, " ")` when the trimmed output is not purely numeric. The new `executeNohup` does not have this fallback. In practice `echo $!` produces a clean number, but you could add the same fallback for consistency.

**Note:** CI checks are still pending at the time of this review.

LGTM.

Croway (approved):

davsclaus (approved):

PR #25310, opened 2026-08-03
CAMEL-24297: make ldif URL-body dereferencing an explicit opt-in

oscerd:
## What

The ldif producer treated a message body that does not start with `version: 1` as a URL and dereferenced it (`URI.create(body).toURL().openStream()`). This content-sniffed mode switch was implicit. It is now gated by a new **`allowUrlBody`** option (default **`false`**, tagged `security="insecure:dev"`).

## Behaviour (breaking)

With the default (`allowUrlBody=false`), a non-LDIF body is **rejected** with an `IllegalArgumentException` instead of being dereferenced as a URL — avoiding a content-sniffed URL fetch (SSRF) from untrusted body content. Routes that intentionally pass a URL as the body must now set `allowUrlBody=true` on the `ldif` endpoint. Documented in the upgrade guide.

## Tests

- New `LdifAllowUrlBodyTest` asserts a non-LDIF body is rejected by default (no LDAP server needed — the rejection happens before the LDAP connection is used).
- The existing `LdifRouteIT` (which feeds resource URLs as the body) now sets `allowUrlBody=true`.

Full-reactor `mvn clean install -DskipTests` is green (catalog/docs/DSL + the `SecurityUtils` `insecure:dev` map entry regenerated).

## Backport

**main only** — this changes a default to be more restrictive (breaking), so it is not backported to the maintenance branches.

Closes CAMEL-24297.

---


gnodet (approved):
_🤖 AI-assisted review — Claude Code on behalf of @gnodet_

---

## Review: CAMEL-24297 — make LDIF URL-body dereferencing an explicit opt-in

Clean security hardening. The previous behavior implicitly treated non-LDIF message bodies as URLs and dereferenced them via `URI.create(body).toURL().openStream()` — an SSRF risk. This PR gates that behind `allowUrlBody` (default `false`), requiring explicit opt-in.

**What's done well:**
- Default is secure: `allowUrlBody = false` rejects non-LDIF bodies with `IllegalArgumentException`
- `@UriParam` correctly annotated with `security = "insecure:dev"` and `label = "security"`
- `SecurityUtils.java` map entry properly added
- Upgrade guide documents the breaking change with clear migration path
- Error message mentions the `allowUrlBody` option name for debuggability
- Existing `LdifRouteIT` updated to set `?allowUrlBody=true`
- New test class is package-private with package-private methods (JUnit 5 convention)
- All generated files consistent with source changes

**One minor observation:** The new `LdifAllowUrlBodyTest` uses JUnit assertions (`assertThrows`, `assertInstanceOf`, `assertTrue`). Per project convention, AssertJ (`assertThatThrownBy`) is preferred for new test code — but this is non-blocking.

LGTM 👍

davsclaus (approved):

PR #25311, opened 2026-08-03
CAMEL-24329: camel-aws2-s3 - throw when pojoRequest=true and the body is the wrong type

oscerd:
Child of CAMEL-24261 (completing CAMEL-23462 across the AWS2 producers — this is the last producer in the sweep).

## Problem

With `pojoRequest=true`, `AWS2S3Producer`'s 22 pojo-capable operations (object &
bucket ops, tagging, ACLs, versioning, policies) only acted when the body was the
matching request type; any other body fell through the `instanceof` with no
returned with no error.

## Fix

type, e.g. `"createBucket operation requires CreateBucketRequest in POJO mode"`.

## Test

`camel-aws2-s3` has no producer route/mock harness, so this is covered by a new
Mockito parameterized unit test over all 22 operations. It mocks the endpoint +
client and sends a wrong-typed body with `pojoRequest=true`. Because several s3
operations resolve the bucket/key *before* the pojo-type check, the test supplies
`bucketName`/`keyName` so the wrong-typed body reaches the guard being exercised.
Verified to fail (silent no-op) before the fix. Region-free — no localstack or
real AWS.

## Docs / backport

---

### DIFF
--- a/components/camel-aws/camel-aws2-s3/pom.xml
+++ b/components/camel-aws/camel-aws2-s3/pom.xml
@@ -88,5 +88,16 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

gnodet (approved):
_🤖 AI-assisted review — Claude Code on behalf of @gnodet_

---

## Review: CAMEL-24329 — camel-aws2-s3 pojoRequest type validation

Clean, mechanical change following the established pattern from the CAMEL-24261 sweep (same as #25305 for camel-aws-config). Adds `else` branches to all 22 `instanceof` checks in the S3 producer's pojo-request code paths, throwing a clear `IllegalArgumentException` instead of silently doing nothing when the body is the wrong type.

**Observations:**
- All 22 pojo-capable operations covered
- The 5 operations without pojo paths (`listBuckets`, `headBucket`, `headObject`, `createDownloadLink`, `createUploadLink`) correctly excluded
- Error messages are consistent and name the expected type
- Null body handled by `getMandatoryBody()` throwing `InvalidPayloadException` before reaching the guard
- Test is well-structured: Mockito parameterized test covering all 22 operations, package-private class/methods, AssertJ assertions

LGTM 👍

PR #25315, opened 2026-08-03
CAMEL-24298: add an optional allowedSchemes allow-list to the toD dynamic-URI EIP

oscerd:
## What

The dynamic-URI EIP `toD` computes its recipient endpoint uri from a route-author expression at runtime (e.g. `toD("${header.target}")`), with no way to restrict which component schemes the recipient may resolve to. This adds an optional **`allowedSchemes`** attribute (comma-separated component-scheme allow-list) on `ToDynamicDefinition`, wired through `ToDynamicReifier` into `SendDynamicProcessor`.

A resolved recipient whose scheme is not in the list is rejected with a `ResolveEndpointFailedException` — **independently of `ignoreInvalidEndpoint`**, so a disallowed scheme always hard-fails. Default unset = any scheme allowed, so there is **no behavioural change**. Useful for low-code / Kamelet deployments that want to constrain e.g. `.toD("${header.dest}")` to a fixed set of components.

```java
from("direct:start").toD().allowedSchemes("http,https").uri("${header.target}");
```

## Scope (MVP)

This is the **`toD`-only MVP** of CAMEL-24298. `wireTap` extends `ToDynamicDefinition` and shares the reifier/processor path, so it inherits and enforces the same option automatically. The 5 sibling dynamic-URI EIPs (`recipientList`, `routingSlip`, `dynamicRouter`, `enrich`, `pollEnrich`) can follow in separate PRs once this pattern is reviewed and accepted — so the JIRA stays open after this merges.

## Tests

`ToDynamicAllowedSchemesTest`: an allowed scheme is sent; a disallowed scheme is rejected with `ResolveEndpointFailedException`; multiple allowed schemes (`mock,seda`) are honored; and a disallowed scheme is rejected even with `ignoreInvalidEndpoint=true`. Full-reactor `mvn clean install -DskipTests` is green (model JSON, XML/YAML schemas, DSL writers/parsers/deserializers, and the toD EIP doc regenerated).

## Docs

Added an `allowedSchemes` section to the toD EIP doc (Java, XML, and YAML examples).

## Backport

**main only** — new additive feature (default unrestricted), not a bug fix.

Partially addresses CAMEL-24298 (toD MVP; sibling EIPs to follow).

---


gnodet (commented):
Nice work on this defence-in-depth feature — the enforcement placement between `prepareRecipient()` and `resolveEndpoint()` is exactly right, and the deliberate bypass of `ignoreInvalidEndpoint` for disallowed schemes is a sound security decision. Two non-blocking suggestions below.

core/camel-core-processor/src/main/java/org/apache/camel/processor/SendDynamicProcessor.java
```diff
@@ -494,4 +507,24 @@ public boolean isAutoStartupComponents() {
...
+    public void setAllowedSchemes(String allowedSchemes) {
+        this.allowedSchemes = allowedSchemes;
+    }
+
+    private boolean isSchemeAllowed(String scheme) {
+        if (allowedSchemes == null) {
+            return true;
+        }
+        for (String allowed : allowedSchemes.split(",")) {
+            if (allowed.trim().equals(scheme)) {
```
gnodet: **Performance nit:** `allowedSchemes.split(",")` allocates a new `String[]` on every exchange. For a high-throughput route this adds up. Consider pre-parsing into a `Set<String>` once in `setAllowedSchemes()`:

```java
private Set<String> allowedSchemesSet;

public void setAllowedSchemes(String allowedSchemes) {
    this.allowedSchemes = allowedSchemes;
    if (allowedSchemes != null) {
        this.allowedSchemesSet = Arrays.stream(allowedSchemes.split(","))
                .map(String::trim)
                .filter(s -> !s.isEmpty())
                .collect(Collectors.toUnmodifiableSet());
    } else {
        this.allowedSchemesSet = null;
    }
}

private boolean isSchemeAllowed(String scheme) {
    if (allowedSchemesSet == null) {
        return true;
    }
    return allowedSchemesSet.contains(scheme);
}
```

This also makes the whitespace-trimming behavior consistent for all entries.

core/camel-core/src/test/java/org/apache/camel/processor/ToDynamicAllowedSchemesTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    protected RouteBuilder createRouteBuilder() {
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:start").toD().allowedSchemes("mock").uri("${header.target}");
+            }
+        };
+    }
+}
```
gnodet, line 91: **Test coverage suggestion:** The test covers the core positive/negative paths well, but the route only configures a single allowed scheme (`"mock"`). Since the documented use case is a comma-separated list (`"http,https"`), it would strengthen confidence to also test:

1. Multiple allowed schemes (e.g. `"mock,seda"`) — verify both are accepted
2. Whitespace in the list (e.g. `"mock, seda"`) — the code handles this via `allowed.trim()` but it's untested

Could be a follow-up — not blocking.

davsclaus (commented):
Nice feature — clean model/reifier/processor layering, correct placement of the security check (after `prepareRecipient`, before `resolveEndpoint`, outside the `ignoreInvalidEndpoint` catch), and sound decision to always hard-fail a disallowed scheme.

I agree with the two suggestions from the prior review (pre-parse `allowedSchemes` into a `Set<String>` to avoid per-exchange `split(",")`, and add a multi-scheme test). Two additional non-blocking observations below.

core/camel-core/src/test/java/org/apache/camel/processor/ToDynamicAllowedSchemesTest.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    protected RouteBuilder createRouteBuilder() {
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:start").toD().allowedSchemes("mock").uri("${header.target}");
+            }
+        };
```
davsclaus, line 89: **Missing test: `ignoreInvalidEndpoint` independence.**

The PR description, the code comment in `SendDynamicProcessor`, and the documentation all explicitly state that a disallowed scheme is rejected *independently* of `ignoreInvalidEndpoint`. This is a security-relevant contract worth locking down with a test.

Consider adding a third test method with a separate route that sets both options:

```java
@Test
void disallowedSchemeIsRejectedEvenWhenIgnoreInvalidEndpoint() {
    assertThatThrownBy(() -> template.sendBodyAndHeader("direct:lenient", "Hello", "target", "seda:blocked"))
            .isInstanceOf(CamelExecutionException.class)
            .cause()
            .isInstanceOf(ResolveEndpointFailedException.class)
            .hasMessageContaining("not in the allowed schemes");
}
```

with the route:

```java
from("direct:lenient").toD().allowedSchemes("mock").ignoreInvalidEndpoint("true").uri("${header.target}");
```

Non-blocking — can be a follow-up.

core/camel-core-engine/src/main/docs/modules/eips/pages/toD-eip.adoc
```diff
@@ -417,3 +417,30 @@ during startup if `toD` was optimized, or if there was a failure loading the opt
...
+
+And in XML:
+
+[source,xml]
+----
+<route>
+    <from uri="direct:start"/>
+    <toD uri="${header.target}" allowedSchemes="http,https"/>
+</route>
+----
```
davsclaus, line 444: **Consider adding a YAML DSL example.**

The section targets "low-code / Kamelet-style deployments" which predominantly use YAML DSL, but only Java and XML examples are shown. A YAML example would serve the primary target audience:

```yaml
- route:
    from:
      uri: direct:start
    steps:
      - toD:
          uri: "${header.target}"
          allowedSchemes: "http,https"
```

Non-blocking.

core/camel-core-processor/src/main/java/org/apache/camel/processor/SendDynamicProcessor.java
oscerd: Done in 92389a8 — `setAllowedSchemes` now pre-parses into a `HashSet` once, and the per-exchange check is a single `Set.contains(scheme)` with no allocation on the hot path. Thanks!

core/camel-core/src/test/java/org/apache/camel/processor/ToDynamicAllowedSchemesTest.java
oscerd: Added `multipleAllowedSchemesHonored` (92389a8): it configures `allowedSchemes("mock,seda")` and verifies both are honored, while a scheme outside the list (`log:`) is rejected.

core/camel-core/src/test/java/org/apache/camel/processor/ToDynamicAllowedSchemesTest.java
oscerd: Good catch — added `disallowedSchemeRejectedEvenWithIgnoreInvalidEndpoint` (92389a8): a route with `ignoreInvalidEndpoint(true)` still hard-fails on a disallowed scheme, since the allow-list check sets the exception directly (and returns) instead of throwing, so it bypasses the `ignoreInvalidEndpoint` handling.

core/camel-core-engine/src/main/docs/modules/eips/pages/toD-eip.adoc
oscerd: Added a YAML DSL example alongside the Java and XML ones (92389a8).

davsclaus (approved):
Clean, well-layered implementation — correct enforcement placement (after `prepareRecipient`, before `resolveEndpoint`, bypassing `ignoreInvalidEndpoint` via `exchange.setException()` + return), pre-parsed `Set<String>` for hot-path efficiency, and solid test coverage including the security-critical `ignoreInvalidEndpoint` independence contract.

All 4 suggestions from prior reviews (pre-parse into Set, multi-scheme test, ignoreInvalidEndpoint test, YAML doc example) addressed in 92389a8.

**Non-blocking observations:**
- CI JDK 17/25 test builds still in progress — recommend waiting for green before merging.
- No explicit test for whitespace tolerance in the allow-list (e.g. `"mock, seda"`) — the code handles it via `allowed.trim()` in the setter but it's untested. Very minor follow-up.
- The `security` group label on the `@Metadata` annotation is a good precedent — worth maintaining as sibling EIPs get the same option.

PR #25323, opened 2026-08-04
CAMEL-24340: camel-salesforce: improve option and header documentation

stn1slv:
JIRA: https://issues.apache.org/jira/browse/CAMEL-24340

### The defect

`apexMethod` is documented as **"APEX method name"**, but the component uses the value verbatim as the **HTTP verb** of the Apex REST call:

- `AbstractRestProcessor:637` reads the option, defaulting it to `GET` at `:640`
- `:653` passes it as the first argument to `restClient.apexCall(...)`
- `RestClient:243` declares that parameter as `String httpMethod` ("HTTP method to execute")
- `DefaultRestClient:499` forwards it to `getRequest(httpMethod, ...)`, and `AbstractClientBase:169-171` passes it into Jetty's `HttpRequest.method(String)`

There is no validation or lookup at any step. A user who follows the documentation and sets `apexMethod` to their real Apex method name, for example `getMerchandise`, makes Camel issue `GETMERCHANDISE /services/apexrest/...`, which fails.

The wording dates from CAMEL-8772 (2015) and has been propagated mechanically since.

**Why this is not cosmetic:** the hand-written Apex table already says "The HTTP method (e.g. GET, POST) to use", but the generated options table is included in the *same page* (`salesforce-component.adoc:58-60`, about 175 lines above), so the page contradicts itself. More importantly the javadoc is what surfaces in IDE completion, the endpoint DSL builders, Camel JBang and Karavan tooling, and Spring Boot metadata, where the prose page is not visible at all.

### Other descriptions improved in the same pass

These restated the option name without saying what the value controls:

| Option | Added |
|---|---|
| `apexUrl` | It is a path relative to `/services/apexrest/`, not a URL, and can be given as the option, in the endpoint path as `apexCall/MyApexClass/`, or via `CamelSalesforceApexUrl` |
| `jobId` | Which operations need it, and that it covers Bulk API v1 and 2.0 |
| `batchId` | Bulk API v1 only; `BulkApiV2Processor` never reads it |
| `resultId` | Used only by `getQueryResult`, sourced from `getQueryResultIds` |
| `pubSubHost` / `pubSubPort` | gRPC transport and the operations that use them |

Plus 15 Streaming API and Change Data Capture headers that followed a bare `"The <name>."` template. Most map to fields of the Salesforce `ChangeEventHeader`.

Worth noting: `CamelSalesforceEventType` carries two different values. For change and platform events it is the last segment of the configured topic name (`StreamingApiConsumer:181,225`); for PushTopic messages it is the Salesforce event type (`:254`). The description now covers both.

### Scope and testing

Documentation only. Every changed line is javadoc or `@Metadata` text, so there is no behaviour change and no test is applicable. Descriptions were derived by reading the call sites rather than from the option names, and claims that could not be verified in code (Salesforce id formats, an exhaustive list of Apex verbs) were deliberately left out.

The regenerated metadata is included for the component, the catalog, and the generated Endpoint DSL and Component DSL builders, as is the convention for metadata changes.

---
_Generated by Claude Code on behalf of @stn1slv_


davsclaus (approved):
Good documentation improvement — the `apexMethod` fix alone prevents a real user pitfall (the old "APEX method name" description was misleading; the value is actually the HTTP verb, as confirmed by `AbstractRestProcessor.java:637-640`). The header descriptions for Streaming API / CDC are much more useful than the prior "The \<name\>." template.

1. **`defaultValue = "GET"` on `@UriParam` for `apexMethod`**: factually correct (matches the runtime default at `AbstractRestProcessor:640`), but note this is a metadata change — generated catalogs and tooling will now show a default value that was previously absent. Since it documents real behavior, this is fine.

2. **`CamelSalesforceCreatedDate`**: the new description honestly notes the dual-type behavior (ZonedDateTime for platform events, String for PushTopic). The declared `javaType = "java.time.ZonedDateTime"` is a pre-existing inaccuracy that is correctly left out of scope here.

Generated files (catalog JSON, component DSL, endpoint DSL) are consistent with the source changes.

Note: this review covers project conventions and rules compliance. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

PR #25326, opened 2026-08-04
CAMEL-24341: camel-google-secret-manager - fix GCP vault refresh task defects

oscerd:
Fixes the defects found in `PubsubReloadTriggerTask` (the `gcp-secret-refresh` period task):

- **The GCP task read the AWS vault configuration.** `secrets` was populated from
  `getVaultConfiguration().aws().getSecrets()`, so the documented `camel.vault.gcp.secrets` option was
  never honoured, and configuring only the GCP property made `doStart` fail with
  "Secrets must be configured on GCP vault configuration". The vault lookup now reads the GCP
  configuration, and was moved ahead of the properties-function auto-detection (both are plain reads,
  the "secrets must be configured" check still runs after both).
- **`triggerReloading` was receiver state and never reset.** After the first matching secret event every
  subsequent message on the subscription triggered another `CamelContext` reload. It is now a local
  variable, like in the AWS equivalent (`CloudTrailReloadTriggerTask`).
- **The subscriber was started on every period.** `run()` is invoked every `camel.vault.gcp.refreshPeriod`
  (default 30s) and a Google `ApiService` can only be started while it is `NEW`, so every tick after the
  first failed with an `IllegalStateException`. The subscriber is now started once.
- **Foreign messages killed the receiver.** `eventType`/`secretId` were dereferenced without a null check,
  so any message on the subscription without those attributes threw and was redelivered indefinitely.

Also in this PR: the javadoc of `setReloadEnabled` said "on AWS secret updated", the properties function
javadoc documented `camel.vault.aws.*` instead of `camel.vault.gcp.*`, and the secret sub-key lookup
allocated an `ObjectMapper` per call.

`PubsubReloadTriggerTask` had no test coverage at all; this adds `PubsubReloadTriggerTaskTest`, with one
test per defect above.

### DIFF
--- a/components/camel-google/camel-google-secret-manager/src/test/java/org/apache/camel/component/google/secret/manager/vault/PubsubReloadTriggerTaskTest.java
+++ b/components/camel-google/camel-google-secret-manager/src/test/java/org/apache/camel/component/google/secret/manager/vault/PubsubReloadTriggerTaskTest.java
@@ -26,12 +26,11 @@ import org.apache.camel.impl.DefaultCamelContext;
 import org.apache.camel.support.DefaultContextReloadStrategy;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-public class PubsubReloadTriggerTaskTest {
+class PubsubReloadTriggerTaskTest {
 
     private static final class CountingReloadStrategy extends DefaultContextReloadStrategy {
         private final AtomicInteger reloads = new AtomicInteger();
@@ -79,7 +78,7 @@ public class PubsubReloadTriggerTaskTest {
     }
 
     @Test
-    public void secretsAreReadFromTheGcpVaultConfiguration() throws Exception {
+    void secretsAreReadFromTheGcpVaultConfiguration() throws Exception {
         DefaultCamelContext context = new DefaultCamelContext();
         context.getVaultConfiguration().gcp().setSecrets("gcp-secret");
         context.getVaultConfiguration().aws().setSecrets("aws-secret");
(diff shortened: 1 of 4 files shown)


components/camel-google/camel-google-secret-manager/src/test/java/org/apache/camel/component/google/secret/manager/vault/PubsubReloadTriggerTaskTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.camel.impl.DefaultCamelContext;
+import org.apache.camel.support.DefaultContextReloadStrategy;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class PubsubReloadTriggerTaskTest {
```
davsclaus: Non-blocking: project convention prefers package-private (no `public`) for test classes and methods (JUnit 5 doesn't require it), and AssertJ assertions (`assertThat(...)`) over JUnit assertions (`assertEquals`, `assertTrue`, etc.).

Example:
```java
import static org.assertj.core.api.Assertions.assertThatThrownBy;

assertThat(getField(task, "secrets")).isEqualTo("gcp-secret");
assertThat(matching.acked).isTrue();
assertThatThrownBy(task::start).isInstanceOf(RuntimeCamelException.class);
```

davsclaus (approved):
All four bug fixes are correct and well-diagnosed:

- **GCP→AWS secrets read** — `aws().getSecrets()` was clearly wrong; confirmed by comparing with the AWS `CloudTrailReloadTriggerTask` which correctly reads `aws().getSecrets()` for its own config.
- **`triggerReloading` as local variable** — matches the pattern already used in `CloudTrailReloadTriggerTask` (line 271). The old field-level flag caused every subsequent message to trigger a reload after the first match.
- **Subscriber started once** — `ApiService.State.NEW` check is the correct guard; a Google `ApiService` throws `IllegalStateException` if started again after `RUNNING`.
- **Null-safe attributes** — constant-first `equalsIgnoreCase` plus explicit `secretId != null` prevents NPE on foreign messages.

Static `ObjectMapper` in `GoogleSecretManagerPropertiesFunction` is thread-safe for `readTree()` — good cleanup.

Minor convention notes (non-blocking):
- Test class and methods use `public` — JUnit 5 convention in this project is package-private (no modifier).
- Tests use JUnit assertions (`assertEquals`, `assertTrue`, etc.) — project prefers AssertJ (`assertThat(...)`).

These are small style items and don't block the PR.

⚠️ Note: CI checks are still pending at time of review.

components/camel-google/camel-google-secret-manager/src/test/java/org/apache/camel/component/google/secret/manager/vault/PubsubReloadTriggerTaskTest.java
oscerd: Done in 02af827 — PubsubReloadTriggerTaskTest is now package-private (class and @Test methods) and uses AssertJ (assertThat / assertThatThrownBy / assertThatCode(...).doesNotThrowAnyException()). The inner-class onReload/ack/nack stay public as they override public API methods. Added a test-scoped assertj-core dep (not previously on this module's test classpath).\n\n_Claude Code on behalf of Andrea Cosentino (@oscerd)._

PR #25329, opened 2026-08-04
CAMEL-24353: camel-mcp-server - selectable target server type for VertxMcpServerEngine

Croway:
JIRA: [CAMEL-24353](https://issues.apache.org/jira/browse/CAMEL-24353) — the small enabler for [CAMEL-23853](https://issues.apache.org/jira/browse/CAMEL-23853) / #25203 discussed [here](https://github.com/apache/camel/pull/25203#issuecomment-5181076911).

`VertxMcpServerEngine` hardwired its router lookup to the main HTTP server (`isMainServer()`). This adds a `targetServerType` property (default `server`, unchanged behavior) so the engine can register the MCP endpoint on the **management** HTTP server router instead — `ManagementHttpServer` already binds its `VertxPlatformHttpRouter` with `SERVER_TYPE_MANAGEMENT` in the registry, so this is pure selection, no new plumbing.

Semantics:
- Exact server-type match; an explicit `management` target **never falls back** to the public server (dev tools must not silently end up on the exposed port).
- The single-router fallback for a bare `VertxPlatformHttpServer` (which carries no server type) is kept for the default target only.
- Fail-fast messages name the missing server and how to enable it.

With this, #25203 can drive the SDK-backed engine directly — `initialize(...)` + `toolAdded(...)` with tools built from the JBang `ToolRegistry` — and drop its hand-rolled JSON-RPC engine while gaining the full streamable transport (sessions, SSE, GET replay, DELETE, `tools/list_changed`).

## Testing

`VertxMcpServerEngineTargetServerTypeTest`:
- Engine targeted at `management` with both servers running: full MCP conversation (SDK client) on the management port, **404** on the main port — also exercises the bridge-less, hand-built-tools usage exactly as #25203 would.
- `management` target with no management server: fails fast with an `IllegalStateException` naming the management server, even though a main-server router is present.

Full `camel-mcp-server` module suite green (10 tests).

---
_This PR was written by Claude Code on behalf of Federico Mariani (@Croway)._

### DIFF
--- a/components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/vertx/VertxMcpServerEngine.java
+++ b/components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/vertx/VertxMcpServerEngine.java
@@ -40,8 +40,10 @@ import org.slf4j.LoggerFactory;
 
 /**
  * {@link McpServerEngine} for Camel Main / JBang: serves MCP streamable HTTP through the Vert.x platform HTTP router
- * using the official MCP Java SDK. The MCP endpoint is registered on the main HTTP server's router, so it serves on the
- * main server port and inherits its lifecycle, authentication and CORS configuration.
+ * using the official MCP Java SDK. By default the MCP endpoint is registered on the main HTTP server's router, so it
+ * serves on the main server port and inherits its lifecycle, authentication and CORS configuration. Set
+ * {@link #setTargetServerType(String)} to {@link VertxPlatformHttpRouter#SERVER_TYPE_MANAGEMENT} to serve on the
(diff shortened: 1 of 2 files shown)


davsclaus (approved):
Clean, well-scoped enabler for CAMEL-23853. The change is backward-compatible (default `SERVER_TYPE_SERVER` preserves existing behavior), security-conscious (explicit `management` target never silently falls back to the public server), and well-tested (positive + negative cases, AssertJ assertions, proper cleanup).

One minor observation: `setTargetServerType(null)` would cause an unhelpful NPE in `lookupRouter()` — a `Objects.requireNonNull` in the setter would be more defensive, though this is non-blocking since the API is internal.

LGTM.

_Claude Code on behalf of Claus Ibsen (@davsclaus)_

PR #25334, opened 2026-08-05
CAMEL-24026: Fix flaky XsltFromFileExceptionTest

atiaomar1978-hub:
## Summary

Fixes [CAMEL-24026](https://issues.apache.org/jira/browse/CAMEL-24026): `XsltFromFileExceptionTest` is flaky on slow CI when the file consumer has not finished before assertions run.

**Why #24625 was reverted:** reordering to wait on `oneExchangeDone` first still failed — the default 10s notify latch timed out (`Exchange should have completed` was false). Simply swapping assertion order was insufficient.

**This fix (different approach):**
- Use `assertMockEndpointsSatisfied(30, SECONDS)` so mock waits generously for the file consumer route to complete
- Poll file move completion with Awaitility (`ok/` / `error/` directories) instead of `oneExchangeDone`
- Drop reliance on the global `oneExchangeDone` notify latch, which races with producer vs consumer exchanges

## Test plan

- [x] `./mvnw -pl core/camel-core -am test -Dtest=XsltFromFileExceptionTest`

---

### DIFF
--- a/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/xslt/XsltFromFileExceptionTest.java
@@ -39,7 +39,8 @@ class XsltFromFileExceptionTest extends ContextTestSupport {
 
         template.sendBodyAndHeader(fileUri(), "<hello>world!</hello>", Exchange.FILE_NAME, "hello.xml");
 
-        // Wait for the file consumer route to finish — producer write alone is not enough.
+        // Do not use oneExchangeDone here: the producer write to fileUri() can satisfy
+        // the global whenDone(1) notify before the file consumer route runs.
         assertMockEndpointsSatisfied(TIMEOUT_SECONDS, TimeUnit.SECONDS);
 
         // File move happens after route processing; poll until the consumer has moved the file.
@@ -59,6 +60,8 @@ class XsltFromFileExceptionTest extends ContextTestSupport {
         // the last tag is not ended properly
         template.sendBodyAndHeader(fileUri(), "<hello>world!</hello", Exchange.FILE_NAME, "hello2.xml");
 
+        // Do not use oneExchangeDone here: the producer write to fileUri() can satisfy
+        // the global whenDone(1) notify before the file consumer route runs.
         assertMockEndpointsSatisfied(TIMEOUT_SECONDS, TimeUnit.SECONDS);
 
         await().atMost(TIMEOUT_SECONDS, TimeUnit.SECONDS)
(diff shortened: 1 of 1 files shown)


atiaomar1978-hub:
## Fix summary

Addresses [CAMEL-24026](https://issues.apache.org/jira/browse/CAMEL-24026) — flaky `XsltFromFileExceptionTest`.

### Why #24625 was reverted
PR #24625 reordered to wait on `oneExchangeDone` before mock assertions. It was reverted in #24945 because CI still saw failures (`Exchange should have completed` was false). Simply swapping order was not enough.

### Why this fix is different
| Approach | Problem |
|---|---|
| Original | 10s mock wait before file consumer finishes |
| #24625 | `oneExchangeDone` unreliable — producer write to `fileUri()` can satisfy the global `whenDone(1)` notify **before** the consumer route runs |
| **This PR** | 30s `assertMockEndpointsSatisfied` for consumer route + Awaitility polling for file move to `ok/`/`error/` |

### Verified
```bash
./mvnw -pl core/camel-core -am test -Dtest=XsltFromFileExceptionTest
```

---

atiaomar1978-hub:
## Review follow-up (BugBot + Grok)

**BugBot:** No issues in this test change (unrelated TUI findings are on fork `main`, not this branch).

**Grok:** Fix approach is sound — timed mock wait + Awaitility for post-route file move. Applied feedback: added inline comment explaining why `oneExchangeDone` must not be used here.

**Note:** `XPathFromFileExceptionTest` has the same legacy pattern; out of scope for this ticket but could be hardened similarly in a follow-up.

---

gnodet (approved):
Well-reasoned fix for a flaky test with clear prior-attempt analysis.

**Root cause:** `oneExchangeDone` is a global `NotifyBuilder` that fires on the first completed exchange of any kind. Since `template.sendBodyAndHeader(fileUri(), ...)` itself completes an exchange (the producer write), `oneExchangeDone` can fire before the file consumer route even picks up the file — making the previous assertion order racy.

**Fix:** Eliminates the latch entirely and uses the right synchronization:
1. `assertMockEndpointsSatisfied(30, SECONDS)` — confirms route processing via MockEndpoint's internal `CountDownLatch`
2. Awaitility — polls for the asynchronous file move that occurs after route completion

This two-phase wait correctly models the two asynchronous steps.

Conventions followed correctly:
- `assertMockEndpointsSatisfied(timeout, unit)` used directly, not wrapped in Awaitility
- Awaitility reserved for the non-mock filesystem condition
- No `Thread.sleep` introduced
- `public` removed from test class and methods per JUnit 5 convention
- Import ordering correct per `impsort-maven-plugin` configuration

PR #25338, opened 2026-08-05
CAMEL-24342: camel-google-storage - fix wrong blob metadata and missing object handling

oscerd:
Three defects in `camel-google-storage`, all covered by new tests.

**1. Uploads stored the content type in two other blob fields.** `processFile` read the
`Content-Encoding` and `Cache-Control` entries from the metadata map but then passed the *content
type* variable to the builder:

```java
String ce = objectMetadata.remove("Content-Encoding");
if (ce != null) { builder.setContentEncoding(ct); }   // ct == Content-Type
...
String cc = objectMetadata.remove("Cache-Control");
if (cc != null) { builder.setCacheControl(ct); }      // ct == Content-Type
```

So an upload sending those headers ended up with the content type in both fields — or `null`, when
no `Content-Type` header was sent at all.

**2. `getObject` NPE'd on a missing object.** `storage.get(BlobId)` returns `null` when the object
is not there, and the result was dereferenced straight away. It now fails with a message naming the
object and the bucket.

**3. The consumer NPE'd on every poll when the configured `objectName` was missing.** Same null
blob, this time handed to `createExchange`. The poll now logs and yields no exchange. That path also
never called `forceConsumerAsReady()`, so the consumer health check stayed not-ready when
`objectName` was configured; it is now marked ready as soon as the storage client answers, like on
the list path.

Two cleanups in the touched file: the `else` branch in `createDownloadLink` was unreachable (the
header lookup already supplies the `300000L` default), and the `java.util.*` wildcard import is
replaced with explicit ones.

**Not changed:** the audit that produced this issue also flagged that the `objectName` endpoint
option wins over the `CamelGoogleCloudStorageObjectName` header. That is deliberate — CAMEL-20998
made it so, because a consumer in the same route sets that header and would otherwise hijack the
producer destination. Only a comment recording the reason was added, and the item was withdrawn in a
comment on the JIRA issue.

### DIFF
--- a/components/camel-google/camel-google-storage/src/test/java/org/apache/camel/component/google/storage/unit/ProducerBlobMetadataTest.java
+++ b/components/camel-google/camel-google-storage/src/test/java/org/apache/camel/component/google/storage/unit/ProducerBlobMetadataTest.java
@@ -26,10 +26,9 @@ import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.google.storage.GoogleCloudStorageConstants;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.assertj.core.api.Assertions.assertThat;
 
-public class ProducerBlobMetadataTest extends GoogleCloudStorageBaseTest {
+class ProducerBlobMetadataTest extends GoogleCloudStorageBaseTest {
 
     private static final String FILE_NAME = "metadata.txt";
 
@@ -48,7 +47,7 @@ public class ProducerBlobMetadataTest extends GoogleCloudStorageBaseTest {
     }
 
     @Test
-    public void contentEncodingAndCacheControlAreStoredAsSent() {
+    void contentEncodingAndCacheControlAreStoredAsSent() {
         Exchange exchange = template.request("direct:store", e -> {
             e.getIn().setHeader(GoogleCloudStorageConstants.OBJECT_NAME, FILE_NAME);
             e.getIn().setHeader(GoogleCloudStorageConstants.CONTENT_TYPE, "text/plain");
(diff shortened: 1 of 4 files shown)


components/camel-google/camel-google-storage/src/test/java/org/apache/camel/component/google/storage/unit/ProducerBlobMetadataTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.camel.Exchange;
+import org.apache.camel.ProducerTemplate;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.google.storage.GoogleCloudStorageConstants;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class ProducerBlobMetadataTest extends GoogleCloudStorageBaseTest {
```
gnodet: Nit: per project conventions, new test classes and methods should use package-private visibility (drop `public`), and prefer AssertJ assertions over JUnit ones. The other new test file in this PR (`GoogleCloudStorageConsumerMissingObjectTest`) already follows both conventions — it would be nice to be consistent here too.

Example:
```java
class ProducerBlobMetadataTest extends GoogleCloudStorageBaseTest {
```
and:
```java
assertThat(blob.getContentEncoding()).isEqualTo("gzip");
```
instead of:
```java
assertEquals("gzip", blob.getContentEncoding());
```

gnodet (approved):
Excellent bug-fix PR that addresses three real issues:

1. **Copy-paste variable error** in blob metadata handling — `contentEncoding` and `cacheControl` were being set to the content-type value instead of their respective header values. The root cause traces back to CAMEL-18300 (2022), where `ct` was used instead of `ce`/`cc` for two of the four metadata fields. CAMEL-22273 fixed `contentDisposition` last year; this PR completes the fix for the remaining two.

2. **NPE path** when a configured `objectName` points to a missing object — now correctly throws a descriptive exception with the object name.

3. **`createDownloadLink` dead code cleanup** — `getHeader(key, defaultValue, type)` with a non-null default of `300000L` can never return `null`, so the `else` branch was unreachable. Simplifying to primitive `long` is the right call.

Test coverage is good, with two dedicated test classes covering all three fixes.

**Minor convention note:** `ProducerBlobMetadataTest` uses `public` class/method visibility and JUnit assertions (`assertEquals`, `assertNotNull`), while the other new test file (`GoogleCloudStorageConsumerMissingObjectTest`) correctly uses package-private visibility and AssertJ. Per project conventions, new test code should prefer package-private visibility and AssertJ assertions. Not blocking — just a consistency note.

davsclaus (approved):

oscerd:
Addressed in 5ac3ff6 — `ProducerBlobMetadataTest` is now package-private with package-private test methods and uses AssertJ (`assertThat(...)`), matching `GoogleCloudStorageConsumerMissingObjectTest`. Thanks @gnodet.

PR #25343, opened 2026-08-05
CAMEL-24354: camel-aws2-lambda - updateFunction never sets the code source, so UpdateFunctionCode always fails

oscerd:
# CAMEL-24354: camel-aws2-lambda — `updateFunction` never sent the code source

## Problem

`Lambda2Producer.updateFunction()` built an `UpdateFunctionCodeRequest` that only ever carried `functionName` and (optionally) `publish`. In the default (non-`pojoRequest`) branch it *validated* that a message body, `CamelAwsLambdaS3Bucket`, or `CamelAwsLambdaS3Key` was present — and then discarded them. It never called `zipFile(..)`, `s3Bucket(..)`, `s3Key(..)` or `s3ObjectVersion(..)` on the builder.

AWS `UpdateFunctionCode` requires exactly one code source (ZipFile, or S3Bucket+S3Key, or ImageUri). The request built here had none, so every call was rejected with:

```
InvalidParameterValueException: Please provide a source for function code.
```

The `updateFunction` operation was therefore **unusable** in the default mode. (`pojoRequest=true` is unaffected — the caller supplies a complete request.) It went unnoticed because no unit test invoked the operation and the test mock echoed only the function name.

## Fix

Assemble the code source on the `UpdateFunctionCodeRequest.Builder` from the same headers/body as `createFunction`:
- `CamelAwsLambdaS3Bucket` → `s3Bucket`
- `CamelAwsLambdaS3Key` → `s3Key`
- `CamelAwsLambdaS3ObjectVersion` → `s3ObjectVersion`
- `CamelAwsLambdaZipFile` (local file path) → `zipFile`
- message body → `zipFile`

and include the `ZIP_FILE` header in the "no source specified" validation.

Also corrects a copy-paste in `deleteEventSourceMapping()`: the validation reads the event source mapping **UUID** header but threw `"Event Source Arn must be specified"` — the message now says UUID.

## Test

Adds `LambdaProducerTest.lambdaUpdateFunctionTest` (the mock now captures the `UpdateFunctionCodeRequest`) asserting the request carries the code source (`zipFile()` non-null) — it fails against the pre-fix code. Full `LambdaProducerTest` green (32 tests).

No public API change, no new dependency. Affects `main` (4.22.0) and the 4.18.x / 4.14.x lines.

---

### DIFF
--- a/components/camel-aws/camel-aws2-lambda/src/main/java/org/apache/camel/component/aws2/lambda/Lambda2Producer.java
+++ b/components/camel-aws/camel-aws2-lambda/src/main/java/org/apache/camel/component/aws2/lambda/Lambda2Producer.java
@@ -469,8 +469,35 @@ public class Lambda2Producer extends DefaultProducer {
 
             if (ObjectHelper.isEmpty(exchange.getIn().getBody())
                     && ObjectHelper.isEmpty(exchange.getIn().getHeader(Lambda2Constants.S3_BUCKET))
-                    && ObjectHelper.isEmpty(exchange.getIn().getHeader(Lambda2Constants.S3_KEY))) {
+                    && ObjectHelper.isEmpty(exchange.getIn().getHeader(Lambda2Constants.S3_KEY))
+                    && ObjectHelper.isEmpty(exchange.getIn().getHeader(Lambda2Constants.ZIP_FILE))) {
                 throw new IllegalArgumentException("At least S3 bucket/S3 key or zip file must be specified");
+            }
+
+            if (ObjectHelper.isNotEmpty(exchange.getIn().getHeader(Lambda2Constants.S3_BUCKET))) {
+                String s3Bucket = exchange.getIn().getHeader(Lambda2Constants.S3_BUCKET, String.class);
+                builder.s3Bucket(s3Bucket);
+            }
+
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Correct, well-targeted bug fix for a completely broken `updateFunction` operation that has been unusable since 2020.

**What the fix does:**
- The `updateFunction` operation never assembled the code source (`FunctionCodeLocation`) — it was missing the S3 bucket/key/object-version mapping and the ZIP_FILE/body-as-ByteBuffer handling. The `UpdateFunctionCode` API call always failed because no code source was provided.
- The fix correctly mirrors the proven `createFunction` code-source assembly pattern, which is the right approach for consistency and correctness.
- The new test verifies the key property (zip file content reaching the request) by capturing the request payload in the mock — closing the gap that let this bug survive since 2020.
- Bonus: fixes a `deleteEventSourceMapping` error message typo ("Event Source Arn" → "Event Source UUID"), which is correct since the code validates `Lambda2Constants.EVENT_SOURCE_UUID` and calls `builder.uuid(...)`.

**Minor convention note:** the new test method uses `public` visibility and JUnit assertions, consistent with the existing file but not with current project conventions (package-private visibility + AssertJ preferred for new code). Not blocking — consistency within the file is reasonable here.

davsclaus (approved):

PR #25354, opened 2026-08-05
CAMEL-24212: camel-spring-ai - upgrade to Spring AI 2.0

Croway:
Fixes [CAMEL-24212](https://issues.apache.org/jira/browse/CAMEL-24212).

## Why

Spring AI 1.1 targets Spring Boot 3.5 and Spring Framework 6. Camel is already on Spring Boot 4.1 and Spring Framework 7, which is the baseline Spring AI 2.0 was designed for, so `camel-spring-ai` currently ships an unsupported pairing.

It also fixes a version split that already exists in the build: `parent/pom.xml` manages `mcp-java-sdk-version` 2.0.0 for `camel-mcp-server`, `camel-mcp-server-api` and `camel-openai`, but `camel-spring-ai-chat` drags in MCP SDK **0.18.2** transitively through Spring AI 1.1.7. After this change both are on 2.0.0.

This picks up where dependabot's #24009 stopped — that PR was closed because the artifact rename below needs a manual change, not because the upgrade was rejected.

## Changes

* `spring-ai-advisors-vector-store` → `spring-ai-vector-store-advisor` (the old artifact was last published at 2.0.0-M8). This is the error that stopped #24009.
* `SpringAiChatMcpManager` no longer binds to a Jackson generation. The stdio transport takes `McpJsonDefaults.getMapper()` instead of constructing a Jackson 2 `ObjectMapper`, so it works with whichever MCP JSON provider is on the classpath (Spring AI 2.0 moved to Jackson 3).
* `ChatClient.ChatClientRequestSpec.options()` now takes the options **builder**.
* `ImageOptionsBuilder.N()` → `n()`.

### toolNames

Spring AI 2.0 removed `ChatClientRequestSpec.toolNames(String...)` together with `SpringBeanToolCallbackResolver`, so the component resolves the names itself. Each name is looked up, in order, against:

1. a `ToolCallbackResolver` bound in the registry — Spring Boot applications get Spring AI's auto-configured one, which is built from the `ToolCallback` and `ToolCallbackProvider` beans;
2. the tools discovered via `tags` and configured via `toolCallbacks`, matched on tool definition name;
3. a `ToolCallback` bound in the registry under that name.

An unresolvable name now fails the exchange with a message listing the available tools, rather than being silently dropped.

Tool callbacks from all sources are also **deduplicated by tool name**. Spring AI 2.0 added `ToolCallingChatOptions.validateToolCallbacks`, which rejects two tools with the same name — without the dedup, an endpoint combining `toolCallbacks` with a `toolNames` entry that names one of them fails with `Multiple tools with the same name`.

## Tests

New `SpringAiChatToolNamesTest` covers name selection, the header override, the registry fallback, the duplicate-registration case and the unresolvable-name error. Existing tests were updated for `options()` vs `defaultOptions()`, the now-mandatory non-null `ChatModel.getOptions()`, Spring Framework 7 core retry, and the openai-java SDK client that `spring-ai-openai` is now built on.

Everything was executed against a local Ollama — **all tests pass, including every integration test**:

| Module | Result |
|---|---|
| chat | 91 pass (5 skipped — `SpringAiChatWrappedFileIT` is `@Disabled` on main) |
| chat — `SpringAiChatMultimodalIT` | 7 pass (needs a vision model, run with `qwen3-vl:2b-instruct`) |
| embeddings | 7 pass |
| vector store | 16 pass, including the Qdrant testcontainers IT |
| image | 4 pass (`SpringAiImageOllamaIT`, with `x/flux2-klein:4b`) |

Models used: `granite4:3b` (chat/tools, the camel test-infra default), `embeddinggemma:300m`, `qwen3-vl:2b-instruct`, `x/flux2-klein:4b`.

## Upgrade notes

`docs/.../camel-4x-upgrade-guide-4_22.adoc` gains a section covering the Jackson 3 move, the `.options.` property-key flattening, the removal of Spring AI's default `temperature=0.7`, and the `toolNames` change.

---


components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/test/java/org/apache/camel/component/springai/chat/SpringAiChatToolNamesTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Unit test for selecting tools by name via the {@code toolNames} option and the
+ * {@link SpringAiChatConstants#TOOL_NAMES} header.
+ *
+ * Spring AI 2.0 removed {@code ChatClient.ChatClientRequestSpec.toolNames(String...)} and the Spring bean based
+ * {@code SpringBeanToolCallbackResolver}, so the component resolves the names itself. These tests capture the
+ * {@link Prompt} handed to the {@link ChatModel} and assert which tool callbacks made it through.
+ */
+public class SpringAiChatToolNamesTest extends CamelTestSupport {
```
davsclaus: Minor convention nit: since this is a **new** test file, the project guidelines ask for package-private visibility on the class and test methods (no `public` modifier). JUnit 5 doesn't require it.

```suggestion
class SpringAiChatToolNamesTest extends CamelTestSupport {
```

Same applies to the four `@Test` methods below (`public void` → `void`).

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
```diff
@@ -994,6 +1005,49 @@ private List<ToolCallback> getToolCallbacksForTags(String tags) {
...
+     * @return                          the resolved callbacks, never {@code null}
+     * @throws IllegalArgumentException if a name cannot be resolved
+     */
+    private List<ToolCallback> resolveToolCallbacksByName(List<String> names, Map<String, ToolCallback> availableByName) {
+        final ToolCallbackResolver resolver = getEndpoint().getCamelContext().getRegistry()
+                .findSingleByType(ToolCallbackResolver.class);
+
+        final List<ToolCallback> resolved = new ArrayList<>(names.size());
+        for (String name : names) {
+            ToolCallback callback = resolver != null ? resolver.resolve(name) : null;
```
davsclaus, line 1033: Question (non-blocking): if a third-party `ToolCallbackResolver` implementation throws instead of returning `null` for an unknown name, the fallback to `availableByName` / registry lookup is skipped. Spring AI's auto-configured resolver likely returns `null`, so this is probably fine — just flagging in case you've verified the contract.

A defensive `try/catch` around `resolver.resolve(name)` could make the fallback chain more robust for custom resolvers, but I'd understand leaving it as-is for now.

davsclaus (commented):
Thanks for this upgrade, @Croway — well-motivated and thorough work. Spring AI 1.1 targeting Spring Boot 3.5 / Spring Framework 6 while Camel is already on Spring Boot 4.1 / Spring Framework 7 made this necessary, and the MCP SDK version alignment is a nice bonus.

The PR description is excellent — clear rationale, detailed changelog, test evidence, and a comprehensive upgrade guide section.

A couple of minor observations inline. Nothing blocking.

---

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/test/java/org/apache/camel/component/springai/chat/SpringAiChatToolNamesTest.java
Croway: Fixed in ea39bdf — class and the four `@Test` methods are now package-private. The overrides (`doPreSetup`, `createRouteBuilder`, `configure`) keep the supertype visibility per the JLS 8.4.8.3 carve-out in AGENTS.md, and the `@Tool` holder classes stay `public` so Spring AI can reflect over them. Test still passes 4/4.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
Croway: Good catch to raise it — I did verify the contract, so I have left it as-is.

`ToolCallbackResolver.resolve(String)` declares its return type as `@org.jspecify.annotations.Nullable` (visible as a `RuntimeVisibleTypeAnnotations` entry on the interface method in `spring-ai-model` 2.0.0), so returning `null` for an unknown name *is* the documented contract rather than an implementation detail. Spring AI 2.0 annotated the whole API surface with JSpecify, which is what makes this checkable now. `DelegatingToolCallbackResolver.resolve` iterates its delegates and returns `null` once none match, and `StaticToolCallbackResolver` returns `null` from its map lookup.

I would rather not wrap it in a `try/catch`: a resolver that throws on an unknown name is violating the annotated contract, and swallowing that would turn a genuine misconfiguration (say a resolver whose backing service is unreachable) into a silent fallback to a different tool, or into the "Cannot resolve tool name" error further down, which would point at the wrong cause. Letting it propagate surfaces the real problem. Happy to revisit if a concrete third-party resolver turns up that throws.

davsclaus: ok

davsclaus (approved):

PR #25357, opened 2026-08-05
CAMEL-24344: camel-google-mail - return a body for raw and non-multipart messages

oscerd:
Three body-related defects in the `google-mail-stream` consumer.

**1. `raw=true` always produced a `null` body.** The consumer asked for the `FULL` format:

```java
Message mess = getClient().users().messages().get("me", message.getId()).setFormat("FULL").execute();
...
if (getConfiguration().isRaw()) { message.setBody(mail.getRaw()); }
```

but Gmail only populates `raw` for `format=RAW`
([API reference](https://developers.google.com/workspace/gmail/api/reference/rest/v1/users.messages):
_"Returned in messages.get and drafts.get responses when the format=RAW parameter is supplied."_).
The requested format now follows the option.

Since the `RAW` format does not return the parsed `payload`, a route with `raw=true` no longer gets
the subject/from/to/cc/bcc/message-id headers — they are part of the RFC 2822 content that is now in
the body. That is covered by an upgrade-guide entry.

**2. Messages that are not multipart arrived with no body.** The old code only looked at
`getPayload().getParts()`, which is `null` for a plain message — its content sits directly on
`payload.body.data`. Nested multiparts (a `multipart/mixed` inside a `multipart/alternative`) were
also skipped, because `parts.get(0).getBody().getData()` is `null` for a part that only holds other
parts. The body is now taken from the first part that actually carries data, descending into nested
parts.

**3. `getPayload()` and `getPayload().getHeaders()` were dereferenced unguarded** — both are absent
for some message formats and metadata-only responses.

Also: `processRollback` re-added the `UNREAD` label even when `markAsRead` was never enabled, so a
failed exchange mutated the mailbox for routes that had not asked for it; and its catch block
reported "Error occurred mark as read mail" for a rollback failure.

### DIFF
--- a/components/camel-google/camel-google-mail/src/main/java/org/apache/camel/component/google/mail/stream/GoogleMailStreamConsumer.java
+++ b/components/camel-google/camel-google-mail/src/main/java/org/apache/camel/component/google/mail/stream/GoogleMailStreamConsumer.java
@@ -92,7 +92,8 @@ public class GoogleMailStreamConsumer extends ScheduledBatchPollingConsumer {
 
         if (c.getMessages() != null) {
             for (Message message : c.getMessages()) {
-                Message mess = getClient().users().messages().get("me", message.getId()).setFormat("FULL").execute();
+                Message mess
+                        = getClient().users().messages().get("me", message.getId()).setFormat(messageFormat()).execute();
                 Exchange exchange = createExchange(getEndpoint().getExchangePattern(), mess);
                 answer.add(exchange);
             }
@@ -167,8 +168,13 @@ public class GoogleMailStreamConsumer extends ScheduledBatchPollingConsumer {
      * Strategy when processing the exchange failed.
      */
     protected void processRollback(Exchange exchange, String unreadLabelId) {
+        if (!getConfiguration().isMarkAsRead()) {
+            // the mail was never marked as read, so there is nothing to roll back
+            LOG.warn("Exchange failed: {}", exchange);
+            return;
+        }
         try {
-            LOG.warn("Exchange failed, so rolling back mail {} to un {}", exchange, unreadLabelId);
+            LOG.warn("Exchange failed, so marking mail {} as unread again", exchange);
 
             List<String> add = new ArrayList<>();
             add.add(unreadLabelId);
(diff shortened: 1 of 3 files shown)


components/camel-google/camel-google-mail/src/main/java/org/apache/camel/component/google/mail/stream/GoogleMailStreamConsumer.java
```diff
@@ -167,16 +168,22 @@ protected void processCommit(Exchange exchange, String unreadLabelId) {
      * Strategy when processing the exchange failed.
      */
     protected void processRollback(Exchange exchange, String unreadLabelId) {
+        if (!getConfiguration().isMarkAsRead()) {
+            // the mail was never marked as read, so there is nothing to roll back
+            LOG.warn("Exchange failed: {}", exchange);
```
davsclaus: why are we WARN logging here when this method is about marking the mail as un-read, and this is an early exit to not do that, so I would assume this LOG should be removed

davsclaus (changes requested):

components/camel-google/camel-google-mail/src/test/java/org/apache/camel/component/google/mail/stream/GoogleMailStreamConsumerBodyTest.java
```diff
@@ -0,0 +1,125 @@
...
+            context.stop();
+        }
+    }
+
+    private GoogleMailStreamConsumer consumer(boolean raw) throws Exception {
+        context = new DefaultCamelContext();
+        context.start();
+        GoogleMailStreamEndpoint endpoint = context.getEndpoint(
+                "google-mail-stream://index?clientId=id&clientSecret=secret&raw=" + raw,
+                GoogleMailStreamEndpoint.class);
```
davsclaus, line 59: Minor resource leak: `theRawOptionAsksForTheRawFormat` calls `consumer()` twice — the first call creates and starts a `DefaultCamelContext`, then the second call overwrites `this.context` without stopping the first one. The orphaned context is never cleaned up by `@AfterEach`.

Suggestion — stop any previous context before creating a new one:

```suggestion
    private GoogleMailStreamConsumer consumer(boolean raw) throws Exception {
        if (context != null) {
            context.stop();
        }
        context = new DefaultCamelContext();
```

davsclaus (commented):
Nice fix — the three bugs are well identified and the upgrade guide entry is clear.

One minor finding in the new test; everything else looks good.

_This review was generated by an AI agent on behalf of davsclaus and may contain inaccuracies. Please verify all suggestions before applying._

oscerd:
Both points addressed in the follow-up commit:

* `processRollback` — the early exit no longer logs. The failure itself is already reported by the
  route's error handler, and nothing is being rolled back on that path.
* `GoogleMailStreamConsumerBodyTest` — `consumer(...)` now stops the previous context before creating
  a new one, so the context built by the first call in `theRawOptionAsksForTheRawFormat` is not
  orphaned.

davsclaus (approved):

davsclaus:
@oscerd This PR has merge conflicts with `main` — could you please rebase?

Once rebased, we're ready to merge (review is approved, all threads resolved).

PR #25361, opened 2026-08-05
CAMEL-24298: extend the allowedSchemes allow-list to the sibling dynamic-URI EIPs

oscerd:
## What

Follows the toD MVP (#25315, CAMEL-24298) by extending the optional **`allowedSchemes`** allow-list to the remaining 5 dynamic-URI EIPs: **`recipientList`, `routingSlip`, `dynamicRouter`, `enrich`, `pollEnrich`**. A resolved dynamic recipient whose component scheme is not in the configured list is rejected — a defence-in-depth restriction for low-code / Kamelet deployments.

```java
from("direct:start").recipientList(header("dest")).allowedSchemes("http,https");
```

## Enforcement

A shared `ProcessorHelper.checkAllowedSchemes(...)` (+ `parseAllowedSchemes(...)`) is called at each processor's recipient-resolution point:
- **RoutingSlip** (and **DynamicRouter**, which extends it), **PollEnricher**, and **RecipientListProcessor** — after `prepareRecipient`.
- **enrich** — `Enricher` delegates to its internal `SendDynamicProcessor`, which already enforces `allowedSchemes` (from #25315).

Each of the 5 model definitions gets the `allowedSchemes` attribute + getter/setter + fluent builder, wired through its reifier.

## Behaviour

Default unset = any scheme allowed → **no behavioural change**. The check always hard-fails (`ResolveEndpointFailedException`) regardless of `ignoreInvalidEndpoints` — `allowedSchemes` is a security boundary, not an endpoint-validity question. This is consistent with `toD` behaviour.

## Tests

- `DynamicUriEipAllowedSchemesTest` — a disallowed scheme (`seda:` against an `allowedSchemes("mock")` route) is rejected across all five EIPs.
- Positive test — an allowed scheme (`mock:result` against `allowedSchemes("mock")`) passes through and the message arrives.

Full-reactor `mvn clean install -DskipTests` green (model JSON, XML/YAML schemas, DSL writers/parsers/deserializers, catalog regenerated). The option is documented on each EIP's auto-generated options table (via the `@Metadata` description).

## Scope

Completes CAMEL-24298: `toD` landed in #25315, `wireTap` inherits it, and this covers the 5 sibling EIPs. Main only — additive feature (default unrestricted), not a bug fix.

Closes CAMEL-24298.

---


core/camel-core-model/src/main/java/org/apache/camel/model/RecipientListDefinition.java
```diff
@@ -109,6 +109,13 @@ public class RecipientListDefinition<Type extends ProcessorDefinition<Type>> ext
     @Metadata(label = "advanced", javaType = "java.lang.Boolean",
               description = "Shares the UnitOfWork with the parent and each of the sub messages. Recipient List will by default not share unit of work between the parent exchange and each recipient exchange. This means each sub exchange has its own individual unit of work.")
     private String shareUnitOfWork;
+    @XmlAttribute
+    @Metadata(label = "advanced,security",
+              description = "Sets an optional comma-separated allow-list of component schemes that the dynamic recipient"
+                            + " may resolve to (e.g. http,https). When set, a dynamic endpoint whose scheme is not in the"
+                            + " list is rejected. This is a defence-in-depth restriction, useful for low-code / Kamelet"
+                            + " deployments; by default (unset) any scheme is allowed.")
+    private String allowedSchemes;
```
davsclaus, line 118: The `allowedSchemes` field is added here, but the **copy constructor** (line ~123 in the new file) does not copy it. This means `copyDefinition()` — used by route advice (`AdviceWith`), route template instantiation, etc. — will silently lose the `allowedSchemes` setting.

All three other Definition classes with copy constructors (`DynamicRouterDefinition`, `EnrichDefinition`, `PollEnrichDefinition`) are correctly updated — this one is missed.

**Fix:** Add `this.allowedSchemes = source.allowedSchemes;` after the `shareUnitOfWork` line in the copy constructor.

core/camel-core-model/src/main/java/org/apache/camel/model/RoutingSlipDefinition.java
```diff
@@ -54,6 +54,13 @@ public class RoutingSlipDefinition<Type extends ProcessorDefinition<Type>> exten
               description = "Configures the cache size for ProducerCache which caches producers for reuse. The default cache size is 1000."
                             + " Set to -1 to turn off caching.")
     private String cacheSize;
+    @XmlAttribute
+    @Metadata(label = "advanced,security",
+              description = "Sets an optional comma-separated allow-list of component schemes that the dynamic recipient"
+                            + " may resolve to (e.g. http,https). When set, a dynamic endpoint whose scheme is not in the"
+                            + " list is rejected. This is a defence-in-depth restriction, useful for low-code / Kamelet"
+                            + " deployments; by default (unset) any scheme is allowed.")
+    private String allowedSchemes;
```
davsclaus, line 63: Same issue — the **copy constructor** (line ~73 in the new file) does not copy this new `allowedSchemes` field.

`DynamicRouterDefinition` (which extends this class) shadows the field and has its own copy constructor that IS updated, so `dynamicRouter` is fine. But standalone `routingSlip` usage via route templating or advice would lose the setting.

**Fix:** Add `this.allowedSchemes = source.allowedSchemes;` after the `cacheSize` line in the copy constructor.

davsclaus (changes requested):
Thanks for extending `allowedSchemes` to the sibling dynamic-URI EIPs, @oscerd — the centralised `ProcessorHelper` approach is clean and the generated files are all consistent.

Two copy-constructor omissions need fixing before this can merge (details in the inline comments). Everything else is a suggestion or design question.

---

### Questions

**Positive test case** — the test covers rejection well, but there is no positive test verifying an *allowed* scheme passes through. A single test (e.g. `recipientList(header("target")).allowedSchemes("mock")` with `target=mock:result`, asserting the message arrives) would strengthen confidence.

**`ignoreInvalidEndpoints` interaction** — for `toD`, a disallowed scheme always hard-fails (check is before the `ignoreInvalidEndpoint` catch). For these 5 EIPs, the check sits inside the existing try-catch, so `ignoreInvalidEndpoints=true` would silently skip a disallowed scheme. Is that intentional, or should `allowedSchemes` always be a hard boundary?

---

This review covers project rules and conventions. It does not replace specialised tools (CodeRabbit, SonarCloud) for deep static analysis.

davsclaus:
i am working on this

core/camel-core-model/src/main/java/org/apache/camel/model/RecipientListDefinition.java
davsclaus: Fixed — added `this.allowedSchemes = source.allowedSchemes;` to the copy constructor. Good catch.

core/camel-core-model/src/main/java/org/apache/camel/model/RoutingSlipDefinition.java
davsclaus: Fixed — same addition in RoutingSlipDefinition's copy constructor.

davsclaus:
Thanks for the thorough review @davsclaus — all points addressed:

**Copy constructors** — both `RecipientListDefinition` and `RoutingSlipDefinition` now copy `allowedSchemes`. (The other three were already correct.)

**Positive test** — added `recipientListAllowsMatchingScheme()` that sends to `mock:result` with `allowedSchemes("mock")` and asserts the message arrives.

**`ignoreInvalidEndpoints` interaction** — agreed that `allowedSchemes` should always be a hard security boundary. Moved the `checkAllowedSchemes` call *before* the `ignoreInvalidEndpoints` catch in both `RecipientListProcessor` and `PollEnricher`. A disallowed scheme now always fails, consistent with `toD`. `RoutingSlip`/`DynamicRouter` already hard-failed (their catch sets the exchange exception and breaks, without checking `ignoreInvalidEndpoints`).

davsclaus (approved):

oscerd:
Thanks for taking this over and folding in the fixes, @davsclaus — the hard-boundary rework in `RecipientListProcessor`/`PollEnricher` and the `ignoreInvalidEndpoints` tests match what I had locally. 👍

One small completeness observation for a possible follow-up (non-blocking — this can land as-is): in `RoutingSlip`, the `allowedSchemes` check is applied on the synchronous hop (`RoutingSlip.java:267`) but not on the **async continuation** path (`RoutingSlip.java:468-469`), where subsequent hops are resolved via `iter.next(ex)` → `prepareRecipient(...)` without the check. So a multi-hop slip whose first hop completes asynchronously could still route to a disallowed scheme on a later hop — e.g. slip `seda:allowed,http://blocked` with `allowedSchemes("seda")`: hop 1 (`seda:allowed`) is allowed and goes async, then hop 2 (`http://blocked`) is resolved on the async path unchecked.

Adding the same `ProcessorHelper.checkAllowedSchemes(allowedSchemes, recipient);` right after the `prepareRecipient(...)` call at `RoutingSlip.java:469` closes it — the surrounding `catch` already does `current.setException(e); break;`, so it hard-fails consistently with the sync path. Happy to send a tiny follow-up PR for it if you'd prefer to keep this one as-is.

davsclaus:
Good catch on the async continuation path — added \`ProcessorHelper.checkAllowedSchemes(allowedSchemes, recipient)\` after \`prepareRecipient()\` at the async path in \`RoutingSlip.java\`. The surrounding \`catch\` already sets the exception and breaks, so it hard-fails consistently with the sync path.

PR #25362, opened 2026-08-05
CAMEL-24345: camel-google-vertexai - apply the streamOutputMode and jsonMode options

oscerd:
`GoogleVertexAIConfiguration` declares two options that no code ever read:

```java
@UriParam(label = "producer", description = "Streaming output mode: complete (default) or chunks",
          defaultValue = "complete", enums = "complete,chunks")
private String streamOutputMode = "complete";

@UriParam(label = "producer", description = "Whether to use JSON request/response format", defaultValue = "false")
private boolean jsonMode;
```

They are documented and exposed in the catalog, so `streamOutputMode=chunks` looked configurable
while `generateChatStreaming` always concatenated the whole response, and `jsonMode` did nothing at
all. The `CamelGoogleVertexAIStreamOutputMode` header constant was equally unread.

* `streamOutputMode=chunks` now produces a `List<String>` with one element per streamed chunk, ready
  for a `split()`. The default `complete` keeps returning the concatenated `String`, so routes that
  do not set the option are unaffected. The header now works too and takes precedence over the
  endpoint option.
* `jsonMode=true` sets the request's response MIME type to `application/json`.

Both are covered by an upgrade-guide entry, since an option that starts being honoured is a
behaviour change for anyone who had set it.

Two smaller items from the same audit:

* `generateChat` and `generateCode` delegate verbatim to `generateText` while being advertised as
  distinct operations. They are now documented as aliases in the enum javadoc — no behaviour change.
* `buildRawPredictRequestBody` called `body.getClass()` on its failure path, so a message with no
  body failed with a `NullPointerException` instead of the intended "must be a JSON String, Map, or
  plain text prompt" message.

### DIFF
--- a/components/camel-google/camel-google-vertexai/pom.xml
+++ b/components/camel-google/camel-google-vertexai/pom.xml
@@ -103,5 +103,10 @@
             <artifactId>camel-test-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/components/camel-google/camel-google-vertexai/src/main/java/org/apache/camel/component/google/vertexai/GoogleVertexAIOperations.java
+++ b/components/camel-google/camel-google-vertexai/src/main/java/org/apache/camel/component/google/vertexai/GoogleVertexAIOperations.java
@@ -30,7 +30,8 @@ public enum GoogleVertexAIOperations {
     generateText,
 
     /**
-     * Generate chat response using Gemini models with conversation history.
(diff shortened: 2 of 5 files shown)


davsclaus (approved):
Nice catch wiring up the two declared-but-never-read options. The PR is clean, well-tested, and follows project conventions.

**One minor observation** (not blocking): the `CHUNK_COUNT` header now counts only chunks with actual text content (`chunks.size()`) whereas the old code counted all stream response objects including those with `null` text. The new semantics are arguably more useful, but it's a subtle behavioral change that isn't mentioned in the upgrade guide entry. Worth adding a note if you agree.

Everything else looks good — config resolution follows the standard header-overrides-config pattern, tests use AssertJ, upgrade guide entry is clear, and the NPE fix in `buildRawPredictRequestBody` is a nice bonus.

PR #25366, opened 2026-08-05
CAMEL-24359: camel-atmosphere-websocket - align Exchange header constant names with Camel naming convention

oscerd:
## Description

Renames the Exchange header string values in `WebsocketConstants` from the dotted `websocket.*` prefix to the project-wide `Camel<Component><Feature>` convention documented in [`design/headers.adoc`](https://github.com/apache/camel/blob/main/design/headers.adoc):

| Constant | Previous value | New value |
|---|---|---|
| `CONNECTION_KEY` | `websocket.connectionKey` | `CamelAtmosphereWebsocketConnectionKey` |
| `CONNECTION_KEY_LIST` | `websocket.connectionKey.list` | `CamelAtmosphereWebsocketConnectionKeyList` |
| `SEND_TO_ALL` | `websocket.sendToAll` | `CamelAtmosphereWebsocketSendToAll` |
| `EVENT_TYPE` | `websocket.eventType` | `CamelAtmosphereWebsocketEventType` |
| `ERROR_TYPE` | `websocket.errorType` | `CamelAtmosphereWebsocketErrorType` |

The Java field names are unchanged, so code referencing the constants symbolically keeps working. Code using the literal strings must be updated.

## Background

`camel-atmosphere-websocket` was in scope for CAMEL-23532, but that change only applied the inherited `HeaderFilterStrategy` to the consumer's WebSocket query parameters. The constants themselves stayed on the legacy `websocket.` prefix, unlike the `camel-vertx-websocket` sibling which was renamed in the same ticket. Since `WebsocketEndpoint extends ServletEndpoint`, the inherited `HttpHeaderFilterStrategy` filters only the `Camel` / `camel` prefixes, so the dotted names sat outside the filtered namespace in both directions.

This continues the CAMEL-23577 alignment sweep (CAMEL-23574, CAMEL-23584, CAMEL-23588, CAMEL-23716).

### Note on the spelling

The values use the concatenated form mandated by `design/headers.adoc` (`"Camel" + ComponentName + Feature`, PascalCase, no separators), matching what the rest of the sweep shipped: `CamelDnsServer`, `CamelJiraIssueKey`, `CamelIrcSendTo`. The dotted `CamelVertxWebsocket.*` form used by the vertx sibling is one of only 37 dotted header values repo-wide against 2313 concatenated, and is not copied here.

`SEND_TO_ALL` is not read by this component at all: broadcast is selected through the `sendToAll` endpoint option, not the header. It is renamed only to keep the class internally consistent.

## Changes

- `WebsocketConstants` - the five header values
- New `WebsocketConstantsTest` - locks the values and asserts all five are filtered by the inherited `HttpHeaderFilterStrategy` in both directions
- 4.22 upgrade-guide entry, mirroring the CAMEL-23574 / CAMEL-23532 entries
- Regenerated component metadata, catalog, `important-headers.json` / `ImportantHeaderUtils`, and the endpoint DSL factory

## Testing

- `camel-atmosphere-websocket` module build green: 14 tests pass (12 existing + 2 new)
- Full reactor build from root (`mvn clean install -DskipTests`) green, all regenerated artifacts committed

## Backport

The constants and the producer/consumer reads are identical on `main`, `camel-4.18.x` and `camel-4.14.x`, so this should be backported to both LTS lines with the matching 4.18 / 4.14 upgrade-guide entries (and those entries doc-synced back to `main` per the backport policy).

## Review feedback (d12c0e7)

`WebsocketConstantsTest` now uses AssertJ (`assertThat(...)`) instead of JUnit assertions, per project convention.

`assertj-core` was not on the `camel-atmosphere-websocket` test classpath, so it is also declared as a test dependency (version-less, supplied by the parent `dependencyManagement`). Flagged in the PR thread since the project standards ask that new dependencies be justified.
---


components/camel-atmosphere-websocket/src/test/java/org/apache/camel/component/atmosphere/websocket/WebsocketConstantsTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.atmosphere.websocket;
+
+import org.apache.camel.http.base.HttpHeaderFilterStrategy;
+import org.apache.camel.spi.HeaderFilterStrategy;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
davsclaus: Non-blocking: per project conventions, new test code should prefer AssertJ assertions (`assertThat(...)`) over JUnit assertions (`assertEquals`, `assertTrue`). For example:

```suggestion
import static org.assertj.core.api.Assertions.assertThat;
```

And then `assertThat(WebsocketConstants.CONNECTION_KEY).isEqualTo("CamelAtmosphereWebsocketConnectionKey")` etc.

davsclaus (approved):
LGTM — clean header-alignment change, thorough upgrade-guide entry, and CI is green.

One minor convention note (non-blocking):

- The new `WebsocketConstantsTest` uses JUnit assertions (`assertEquals`, `assertTrue`). Per project conventions, new test code should prefer AssertJ (`assertThat(...)`) for consistency. This can be addressed in a follow-up if desired.

Everything else checks out:
- Header values follow `design/headers.adoc` (`"Camel" + ComponentName + Feature`, PascalCase)
- `SEND_TO_ALL` not being read by the component is confirmed — rename is for internal consistency only
- The `HeaderFilterStrategy` behaviour change is well-documented in the upgrade guide
- Generated files are properly regenerated

oscerd:
Thanks for the review. Addressed in d12c0e7 — the new `WebsocketConstantsTest` now uses AssertJ (`assertThat(...)`) throughout.

One thing worth flagging, since the suggestion assumed AssertJ was already usable here: `assertj-core` was **not** on the `camel-atmosphere-websocket` test classpath. Confirmed with `mvn dependency:tree -Dincludes=org.assertj:assertj-core`, which returned nothing for this module (and correctly resolves `assertj-core:3.27.7:test` for a module that does declare it, e.g. `camel-langchain4j-agent-api`).

So the change also declares it as a test dependency. The version comes from the parent `dependencyManagement`, so it is a version-less block matching how other modules declare it:

```xml
<dependency>
    <groupId>org.assertj</groupId>
    <artifactId>assertj-core</artifactId>
    <scope>test</scope>
</dependency>
```

Flagging it explicitly because `CLAUDE.md` says AssertJ "is already available as a test dependency in the project", which is true project-wide but not for this module — and the project standards ask that new dependencies be justified. Happy to revert to JUnit assertions instead if you would rather not add the dependency to this component just for assertion style.

Module build is green: 14 tests pass, and the full reactor build from root is clean with no regenerated-artifact drift.

---

PR #25367, opened 2026-08-05
CAMEL-24360: camel-undertow - use UndertowHeaderFilterStrategy as the endpoint default

oscerd:
## Description

`DefaultUndertowHttpBinding` installs an `UndertowHeaderFilterStrategy` in its constructor:

```java
public DefaultUndertowHttpBinding(boolean useStreaming) {
    this.headerFilterStrategy = new UndertowHeaderFilterStrategy();
    ...
}
```

but `UndertowEndpoint` defaulted its own field to the base `HttpHeaderFilterStrategy` and pushed that into the binding when creating it lazily, overwriting what the binding had just set:

```java
private HeaderFilterStrategy headerFilterStrategy = new HttpHeaderFilterStrategy();   // line 88

public UndertowHttpBinding getUndertowHttpBinding() {
    if (undertowHttpBinding == null) {
        undertowHttpBinding = new DefaultUndertowHttpBinding(useStreaming);
        undertowHttpBinding.setHeaderFilterStrategy(getHeaderFilterStrategy());        // overwrites it
        ...
```

Unless the user supplied a custom binding or a custom `headerFilterStrategy`, the binding ran the base strategy and the undertow-specific filtering never executed.

## Impact

Two behaviours documented as shipped were inert on endpoint-configured routes:

- the legacy `websocket.*` Exchange-header prefix added to the in/out filters by CAMEL-23588 (released in 4.14.8 / 4.18.3 / 4.21.0 and described in the 4.18 upgrade guide);
- the `io.undertow.util.HttpString.tryFromString` header-name validation in `UndertowHeaderFilterStrategy.applyFilterToExternalHeaders`, which skips header names undertow does not accept.

Documentation and runtime behaviour had therefore diverged since CAMEL-23588.

## Fix

`UndertowEndpoint` now defaults to `UndertowHeaderFilterStrategy`, so both take effect.

- Rest DSL consumers are unchanged: `UndertowComponent` already assigns `UndertowRestHeaderFilterStrategy`, which extends `UndertowHeaderFilterStrategy`.
- Endpoints configuring `headerFilterStrategy` or `undertowHttpBinding` explicitly keep their behaviour.
- Ordinary application headers are unaffected; the two added filters are narrow.

The now-unused `HttpHeaderFilterStrategy` import is removed. No generated metadata changes, since the catalog records only the `HeaderFilterStrategy` interface and not the default implementation class.

## Changes

- `UndertowEndpoint` - default strategy, unused import
- `UndertowEndpointTest` - three cases: the endpoint default is undertow-specific; the lazily created binding keeps it and actually filters `UndertowConstants.CONNECTION_KEY` / `CONNECTION_KEY_LIST` / `SEND_TO_ALL`; an explicitly configured strategy still reaches the binding
- 4.22 upgrade-guide entry, including how to restore the previous behaviour

## Testing

- `camel-undertow` module build green: **191 tests pass, 0 failures, 1 skipped**. The websocket suites (`UndertowWsConsumerRouteTest`, `UndertowWsTwoRoutesToSameEndpointSendToAllHeaderTest`, ...) all still pass, since the producer reads those headers via `in.getHeader(...)`, which does not go through the `HeaderFilterStrategy`.
- Full reactor build from root (`mvn clean install -DskipTests`) green, no regenerated-artifact drift.

## Backport

`UndertowEndpoint` line 88 and `DefaultUndertowHttpBinding` line 86 are identical on `main`, `camel-4.18.x` and `camel-4.14.x`, so this should be backported to both LTS lines (with the matching 4.18 / 4.14 upgrade-guide entries doc-synced back to `main` per the backport policy). Backporting matters here because CAMEL-23588 shipped on those branches and its documented effect is what this restores.

## Review feedback (ff954f9)

The three new `UndertowEndpointTest` methods now use AssertJ (`assertThat(...)`) instead of JUnit assertions, per project convention. The two pre-existing methods keep their JUnit assertions, per the convention that touched code migrates without sweeping the whole file.

`assertj-core` was not on the `camel-undertow` test classpath, so it is also declared as a test dependency (version-less, supplied by the parent `dependencyManagement`). Flagged in the PR thread since the project standards ask that new dependencies be justified.

### Note on the earlier CI failure

The earlier red on this PR was not a regression. `build (17)` failed in `camel-cxf-spring-soap` with `BindException: Address already in use` / `Soap 1.1 endpoint already registered`, and `build (25)` was **cancelled** by matrix fail-fast rather than failing. That module has no `undertow:` endpoint usage in its tests, so `UndertowEndpoint` is not on its code path (it publishes through CXF's own `cxf-rt-transports-http-undertow`). A re-run passed both jobs.
---

### DIFF
--- a/components/camel-undertow/src/test/java/org/apache/camel/component/undertow/UndertowEndpointTest.java
+++ b/components/camel-undertow/src/test/java/org/apache/camel/component/undertow/UndertowEndpointTest.java
@@ -23,10 +23,8 @@ import org.apache.camel.spi.HeaderFilterStrategy;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertInstanceOf;
-import static org.junit.jupiter.api.Assertions.assertSame;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class UndertowEndpointTest {
 
@@ -55,23 +53,23 @@ public class UndertowEndpointTest {
 
     @Test
     void defaultHeaderFilterStrategyIsUndertowSpecific() {
-        assertInstanceOf(UndertowHeaderFilterStrategy.class, endpoint.getHeaderFilterStrategy());
+        assertThat(endpoint.getHeaderFilterStrategy()).isInstanceOf(UndertowHeaderFilterStrategy.class);
     }
 
     @Test
     void defaultBindingKeepsUndertowHeaderFilterStrategy() {
         // the endpoint pushes its own strategy into the lazily created binding, so the endpoint default
         // decides which strategy the binding ends up running
-        DefaultUndertowHttpBinding binding
-                = assertInstanceOf(DefaultUndertowHttpBinding.class, endpoint.getUndertowHttpBinding());
-        HeaderFilterStrategy strategy = binding.getHeaderFilterStrategy();
-        assertInstanceOf(UndertowHeaderFilterStrategy.class, strategy);
+        assertThat(endpoint.getUndertowHttpBinding()).isInstanceOf(DefaultUndertowHttpBinding.class);
+        HeaderFilterStrategy strategy
+                = ((DefaultUndertowHttpBinding) endpoint.getUndertowHttpBinding()).getHeaderFilterStrategy();
+        assertThat(strategy).isInstanceOf(UndertowHeaderFilterStrategy.class);
 
         // the undertow-specific prefixes added by CAMEL-23588 must therefore be in effect
-        assertTrue(strategy.applyFilterToExternalHeaders(UndertowConstants.CONNECTION_KEY, "aValue", null));
-        assertTrue(strategy.applyFilterToExternalHeaders(UndertowConstants.CONNECTION_KEY_LIST, "aValue", null));
-        assertTrue(strategy.applyFilterToExternalHeaders(UndertowConstants.SEND_TO_ALL, "aValue", null));
-        assertTrue(strategy.applyFilterToCamelHeaders(UndertowConstants.CONNECTION_KEY, "aValue", null));
+        assertThat(strategy.applyFilterToExternalHeaders(UndertowConstants.CONNECTION_KEY, "aValue", null)).isTrue();
+        assertThat(strategy.applyFilterToExternalHeaders(UndertowConstants.CONNECTION_KEY_LIST, "aValue", null)).isTrue();
+        assertThat(strategy.applyFilterToExternalHeaders(UndertowConstants.SEND_TO_ALL, "aValue", null)).isTrue();
+        assertThat(strategy.applyFilterToCamelHeaders(UndertowConstants.CONNECTION_KEY, "aValue", null)).isTrue();
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


components/camel-undertow/src/test/java/org/apache/camel/component/undertow/UndertowEndpointTest.java
```diff
@@ -18,10 +18,15 @@
 
 import java.net.URI;
 
+import org.apache.camel.http.base.HttpHeaderFilterStrategy;
+import org.apache.camel.spi.HeaderFilterStrategy;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
```
davsclaus, line 24: Non-blocking: per project conventions, new test code should prefer AssertJ assertions (`assertThat(...)`) over JUnit assertions (`assertInstanceOf`, `assertSame`, `assertTrue`).

davsclaus (approved):
LGTM — genuine bugfix, minimal change, well-tested.

The root cause is confirmed: `UndertowEndpoint` line 88 defaulted to `HttpHeaderFilterStrategy`, and `getUndertowHttpBinding()` pushed that into the lazily created binding, overwriting the `UndertowHeaderFilterStrategy` installed by the binding's own constructor. The `websocket.*` prefix filtering from CAMEL-23588 was therefore never active on endpoint-configured routes.

One minor convention note (non-blocking):

- The new test methods use JUnit assertions (`assertInstanceOf`, `assertSame`, `assertTrue`). Per project conventions, new test code should prefer AssertJ (`assertThat(...)`) for consistency.

Everything else checks out:
- Fix is correct and minimal (one-line default change)
- Tests cover three scenarios: default strategy, lazy binding propagation, explicit custom strategy
- Rest DSL consumers are unaffected (`UndertowRestHeaderFilterStrategy` already extends `UndertowHeaderFilterStrategy`)
- Upgrade guide documents the change and how to restore previous behaviour

oscerd:
Thanks for the review. Addressed in ff954f9 — the three new test methods now use AssertJ (`assertThat(...)`).

Two notes:

**1. AssertJ was not on this module's classpath.** The suggestion assumed it was already usable here, but `mvn dependency:tree -Dincludes=org.assertj:assertj-core` returns nothing for `camel-undertow` (and correctly resolves `assertj-core:3.27.7:test` for a module that does declare it, e.g. `camel-langchain4j-agent-api`). So this change also declares it as a test dependency, version-less since the parent `dependencyManagement` supplies it:

```xml
<dependency>
    <groupId>org.assertj</groupId>
    <artifactId>assertj-core</artifactId>
    <scope>test</scope>
</dependency>
```

Flagging it because the project standards ask that new dependencies be justified. Happy to revert to JUnit assertions if you would rather not add it to this component just for assertion style.

**2. Only the three new methods are converted.** `emptyPathShouldBeReplacedWithSlash` and `nonEmptyPathShouldBeKeptSame` predate this PR and keep their JUnit `assertEquals`, per the convention that touched code migrates without sweeping the whole file. The file therefore imports both, with one style per method as the convention requires. Say the word if you would prefer the whole file migrated in one go.

Module build is green: 191 tests pass, 0 failures. Full reactor build from root is clean with no regenerated-artifact drift.

Also worth noting for the record: the earlier CI red on this PR was not a regression. `build (17)` failed in `camel-cxf-spring-soap` with `BindException: Address already in use` / `Soap 1.1 endpoint already registered`, and `build (25)` was cancelled by matrix fail-fast rather than failing. `camel-cxf-spring-soap` has no `undertow:` endpoint usage in its tests, so `UndertowEndpoint` is not on that code path — it publishes through CXF's own `cxf-rt-transports-http-undertow`. A re-run passed both jobs.

---

PR #25369, opened 2026-08-06
CAMEL-24357: camel-aws2-s3-vectors - consumer returns no results (topK=0), ignores delay, and can lose vectors

oscerd:
# CAMEL-24357: camel-aws2-s3-vectors — consumer was non-functional and could lose vectors

The `aws2-s3-vectors` consumer had several defects, all rooted in consumer options that shadow the base scheduled-poll options.

## 1. Consumer always sent `topK(0)` — returned nothing (primary, HIGH)

`AWS2S3VectorsConsumer.poll()` built the query with `topK(Math.min(getMaxMessagesPerPoll(), getConfiguration().getTopK()))`. `getMaxMessagesPerPoll()` is the base `ScheduledBatchPollingConsumer` field, which defaults to **0** and was never wired — `AWS2S3VectorsEndpoint.createConsumer()` only called `configureConsumer(consumer)`, never `setMaxMessagesPerPoll(..)` (unlike `AWS2S3Endpoint`). So the expression was `Math.min(0, topK) = 0` and every poll sent `topK(0)`, which AWS S3 Vectors rejects (`topK >= 1`). **The consumer delivered zero messages out of the box.**

**Fix:** wire `setMaxMessagesPerPoll` from the configuration in `createConsumer`, and resolve `topK` so that a non-positive `maxMessagesPerPoll` (the "unlimited" default) means "use the configured `topK`" instead of capping to zero.

## 2. `delay` option was ignored

`delay` is declared on the configuration, so `?delay=` bound to `configuration.setDelay(..)` — a value the consumer never read; the real poll interval came from the inherited `ScheduledPollEndpoint`. A user setting `?delay=60000` was silently ignored.

**Fix:** propagate `configuration.getDelay()` to the consumer's scheduler in `createConsumer` (same place `maxMessagesPerPoll` is now wired).

## 3. Vectors marked processed before routing → event loss on failure

`poll()` added each vector id to `processedVectorIds` at enqueue time, before the exchange was routed. If routing later failed, the vector stayed in the index but was in the de-dup set, so the fixed similarity query skipped it forever. The set was also cleared only on stop (unbounded growth).

**Fix:** only track `processedVectorIds` when `deleteAfterRead=false` (deletion already prevents re-delivery, so the set no longer grows in that mode), and drop the id again on failure (`VectorDedupSynchronization.onFailure`) so a failed exchange is retried on a later poll.

## Tests

New `AWS2S3VectorsConsumerTest` (Mockito): `consumerSendsPositiveTopK` captures the `QueryVectorsRequest` and asserts `topK` is `>= 1` (fails against the old code), and `delayOptionDrivesTheConsumerPollInterval` asserts the configured `delay` reaches the consumer's scheduler. Existing producer tests still pass; full reactor build is green.

No public API change. `assertj-core` added as a test dependency (project-standard). Targets `main` (4.22.0) and `camel-4.18.x` (the module was added in 4.17.0; it does not exist on 4.14.x).

---

### DIFF
--- a/components/camel-aws/camel-aws2-s3-vectors/src/main/java/org/apache/camel/component/aws2/s3vectors/AWS2S3VectorsConsumer.java
+++ b/components/camel-aws/camel-aws2-s3-vectors/src/main/java/org/apache/camel/component/aws2/s3vectors/AWS2S3VectorsConsumer.java
@@ -83,7 +83,7 @@ public class AWS2S3VectorsConsumer extends ScheduledBatchPollingConsumer {
                     .vectorBucketName(vectorBucketName)
                     .indexName(vectorIndexName)
                     .queryVector(VectorData.builder().float32(queryVector).build())
-                    .topK(Math.min(getMaxMessagesPerPoll(), getConfiguration().getTopK()));
+                    .topK(resolveTopK());
 
             // Add metadata filter if configured
             String metadataFilter = getConfiguration().getConsumerMetadataFilter();
@@ -121,15 +121,18 @@ public class AWS2S3VectorsConsumer extends ScheduledBatchPollingConsumer {
                     message.setHeader(AWS2S3VectorsConstants.VECTOR_BUCKET_NAME, vectorBucketName);
                     message.setHeader(AWS2S3VectorsConstants.VECTOR_INDEX_NAME, vectorIndexName);
 
-                    // Add to processed set
-                    processedVectorIds.add(vectorId);
-
                     // Add delete callback if deleteAfterRead is enabled
                     if (getConfiguration().isDeleteAfterRead()) {
                         exchange.getExchangeExtension().addOnCompletion(
                                 new VectorDeleteSynchronization(
                                         getS3VectorsClient(), vectorBucketName, vectorIndexName,
                                         vectorId));
+                    } else {
+                        // Track for de-duplication only when we are not deleting (a deleted vector cannot be
+                        // returned again). Mark it now to avoid re-delivering it on overlapping polls, but drop it
+                        // again if the exchange fails so it can be retried on a subsequent poll.
+                        processedVectorIds.add(vectorId);
+                        exchange.getExchangeExtension().addOnCompletion(new VectorDedupSynchronization(vectorId));
                     }
 
                     exchanges.add(exchange);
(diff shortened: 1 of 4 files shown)


components/camel-aws/camel-aws2-s3-vectors/src/main/java/org/apache/camel/component/aws2/s3vectors/AWS2S3VectorsConsumer.java
```diff
@@ -121,15 +121,18 @@ protected int poll() throws Exception {
...
                     if (getConfiguration().isDeleteAfterRead()) {
                         exchange.getExchangeExtension().addOnCompletion(
                                 new VectorDeleteSynchronization(
                                         getS3VectorsClient(), vectorBucketName, vectorIndexName,
                                         vectorId));
+                    } else {
+                        // Track for de-duplication only when we are not deleting (a deleted vector cannot be
+                        // returned again). Mark it now to avoid re-delivering it on overlapping polls, but drop it
+                        // again if the exchange fails so it can be retried on a subsequent poll.
+                        processedVectorIds.add(vectorId);
```
davsclaus, line 134: Non-blocking / follow-up: `processedVectorIds` is a plain `HashSet`, now written from both `poll()` (scheduler thread) and `VectorDedupSynchronization.onFailure()` (exchange completion thread — potentially different with async routing). For most configurations this is fine (synchronous routing), but a `ConcurrentHashMap.newKeySet()` would be safer against subtle races.

Also, when `deleteAfterRead=false` the set grows unboundedly until `doStop()`. Both are pre-existing design concerns (not introduced by this PR), but worth a follow-up.

davsclaus (approved):
Solid bugfix PR — all three issues are well-analyzed and the fixes are correct.

**Verified:**
- `topK(0)` bug: `resolveTopK()` correctly treats non-positive `maxMessagesPerPoll` as "unlimited" instead of capping to 0. Matches the `aws2-s3` endpoint pattern.
- `delay` shadowing: the `@UriParam` on the configuration captures the value before `configureConsumer` can map it, so explicit propagation is needed and correct.
- De-dup logic: clean separation — deletion prevents re-delivery so the set is only needed when `deleteAfterRead=false`. The `VectorDedupSynchronization.onFailure` correctly drops the id on failure for retry.

Tests follow project conventions (AssertJ, package-private visibility, descriptive names) and directly exercise the bug conditions.

One minor non-blocking observation noted inline.

Note: this review covers project rules and conventions. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

PR #25378, opened 2026-08-06
CAMEL-24296: camel-support - apply a JEP-290 deserialization filter on CamelObjectInputStream by default

oscerd:
## What

`CamelObjectInputStream` is the shared `ObjectInputStream` used by Camel's Java-object deserialization paths (e.g. the HTTP components), but it installed no `java.io.ObjectInputFilter`, so any code constructing it directly deserialized without a JEP-290 filter.

This change makes `CamelObjectInputStream` resolve a filter via the existing `DeserializationFilterHelper` and install it with `setObjectInputFilter()`:

- When no explicit pattern is supplied, the JVM-wide `jdk.serialFilter` is honoured if set, otherwise the default Camel allow-list (`DEFAULT_DESERIALIZATION_FILTER`) is applied — permits standard Java and Apache Camel types, denies `java.net.**`, and enforces JEP-290 graph-shape limits.
- A new `CamelObjectInputStream(InputStream, CamelContext, String)` constructor accepts an explicit filter pattern.

`HttpHelper` now passes its configured pattern through the new constructor instead of calling `setObjectInputFilter()` separately. The built-in HTTP path already applied the default filter (via `DeserializationFilterHelper`), so its runtime behaviour is unchanged — this closes the gap for any direct instantiation and centralizes the filter on the stream itself.

## Why

Defense-in-depth against unsafe deserialization: the shared stream is now safe-by-default and configurable, consistent with the deserialization filtering already applied by `camel-jms`, `camel-mina`, `camel-netty` and `camel-netty-http`.

## Testing

- New `CamelObjectInputStreamTest` (4 tests): the default filter allows standard types, rejects a class outside the allow-list, a blank pattern falls back to the default, and an explicit pattern can allow an otherwise-denied class.
- Full-reactor `mvn clean install -DskipTests` passes.

## Docs

- Upgrade-guide entry added to `camel-4x-upgrade-guide-4_22.adoc` (public `camel-support` class default change).

### DIFF
--- a/core/camel-support/src/test/java/org/apache/camel/support/CamelObjectInputStreamTest.java
+++ b/core/camel-support/src/test/java/org/apache/camel/support/CamelObjectInputStreamTest.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.support;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InvalidClassException;
+import java.io.ObjectOutputStream;
+import java.net.InetSocketAddress;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class CamelObjectInputStreamTest {
+
+    @Test
+    void defaultFilterAllowsStandardTypes() throws Exception {
+        assertEquals("hello camel", deserialize(serialize("hello camel")));
+    }
+
+    @Test
+    void defaultFilterRejectsClassOutsideAllowList() throws Exception {
+        byte[] data = serialize(InetSocketAddress.createUnresolved("example.com", 8080));
+        InvalidClassException ex = assertThrows(InvalidClassException.class, () -> deserialize(data));
+        assertTrue(ex.getMessage().contains("REJECTED"), ex.getMessage());
+    }
+
+    @Test
+    void blankFilterFallsBackToDefault() throws Exception {
+        byte[] data = serialize(InetSocketAddress.createUnresolved("example.com", 8080));
+        InvalidClassException ex = assertThrows(InvalidClassException.class, () -> deserialize(data, "   "));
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
```diff
@@ -55,6 +55,21 @@ recipient; aligning it with `toD` / `enrich` is deferred to a follow-up.
...
+defense-in-depth measure against unsafe deserialization. When no explicit filter pattern is
+supplied, the JVM-wide `jdk.serialFilter` is honoured if set, otherwise Camel's default allow-list
+(`DeserializationFilterHelper.DEFAULT_DESERIALIZATION_FILTER`) is applied, which permits standard
+Java and Apache Camel types, denies `java.net.**`, and enforces graph-shape limits.
+
+The built-in HTTP deserialization path already applied this filter, so most users are unaffected.
+Code that constructs `CamelObjectInputStream` directly and deserializes types outside the default
+allow-list must pass an explicit filter pattern to the new
+`CamelObjectInputStream(InputStream, CamelContext, String)` constructor (or configure
+`jdk.serialFilter`) to permit them.
```
gnodet, line 71: Minor documentation gap: since the 2-arg constructor now delegates to the 3-arg with `null` (which always installs a filter), any downstream code that previously called `setObjectInputFilter()` after constructing via the 2-arg constructor will get an `IllegalStateException` on JDK 17+ ("filter can not be set more than once"). Within Camel only `HttpHelper` used this class (and the PR correctly updates it), so the practical risk is low — but it might be worth adding a sentence here noting this backward-compatibility change and recommending the 3-arg constructor as the migration path for anyone who was calling `setObjectInputFilter()` directly.

core/camel-support/src/test/java/org/apache/camel/support/CamelObjectInputStreamTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InvalidClassException;
+import java.io.ObjectOutputStream;
+import java.net.InetSocketAddress;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
gnodet, line 29: Nit: the project convention prefers AssertJ assertions (`assertThat(...)`) for new test code. Not blocking — the sibling `DeserializationFilterHelperTest` also uses JUnit assertions, so this is consistent with the existing pattern in this area.

gnodet (commented):
Good security hardening — centralizing JEP-290 filter application in the `CamelObjectInputStream` constructor is the right approach. The deny-by-default posture (`!*`) and the fallback chain (explicit pattern → JVM-wide `jdk.serialFilter` → Camel default) give operators appropriate control. Tests cover the meaningful code paths well.

Two minor observations below.

docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_22.adoc
oscerd: Good catch — added a note in the upgrade guide (8de0b559) documenting the `IllegalStateException` ("filter can not be set more than once") thrown by a subsequent `setObjectInputFilter()` call, and recommending the 3-arg constructor as the migration path.

core/camel-support/src/test/java/org/apache/camel/support/CamelObjectInputStreamTest.java
oscerd: Kept JUnit here to stay consistent with the sibling `DeserializationFilterHelperTest` in the same module — camel-support has no assertj-core test dependency, and adding one just for this test felt like unnecessary scope. Happy to switch if you'd prefer.

PR #25387, opened 2026-08-06
CAMEL-24366: Add embeddingModel auto-embed to langchain4j-embeddingstore component

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-24366

## Summary

- Adds an optional `embeddingModel` property (autowired) to the `langchain4j-embeddingstore` component
- When configured, ADD and SEARCH operations accept plain text in the message body instead of requiring a pre-computed `Embedding` header
- The existing `CamelLangChain4jEmbeddingsEmbedding` header always takes precedence when present, preserving backward compatibility

## Changes

- `LangChain4jEmbeddingStoreConfiguration` — new `embeddingModel` field with `@UriParam` and `@Metadata(autowired = true)`
- `LangChain4jEmbeddingStoreProducer` — extracted shared `resolveEmbedding()` helper used by both `add()` and `search()`, with null-body guard and TextSegment header preservation
- `LangChain4jEmbeddingStoreAutoEmbedTest` — 6 tests covering auto-embed ADD/SEARCH, header precedence, null body error, TextSegment preservation, and no-model fallback
- Generated configurers, DSL builders, and catalog metadata regenerated

## Test plan

- [x] Auto-embed ADD with plain text body
- [x] Auto-embed SEARCH with plain text body
- [x] Explicit embedding header takes precedence over auto-embed
- [x] Null body with auto-embed throws descriptive error
- [x] TextSegment header preserved when provided without embedding header
- [x] Missing model and header throws NoSuchHeaderException


components/camel-ai/camel-langchain4j-embeddingstore/pom.xml
```diff
@@ -75,6 +75,12 @@
             <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j</artifactId>
+            <version>${langchain4j-version}</version>
+            <scope>test</scope>
```
davsclaus: Is this `langchain4j` umbrella dependency actually needed? The test only uses classes from `langchain4j-core` (already a compile dependency) and `langchain4j-embeddings-all-minilm-l6-v2` (already a test dependency). If it's only pulled in transitively by one of those, it can likely be removed as an explicit dependency.

davsclaus (commented):

Nice feature addition — clean code, good test coverage, and well-documented. The header-takes-precedence design preserves backward compatibility nicely.

One question about a test dependency (see inline comment). Otherwise this looks good.

Note: this review covers project conventions and code structure. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

components/camel-ai/camel-langchain4j-embeddingstore/pom.xml
JiriOndrusek: you are right. Dependency is removed.

davsclaus (approved):

components/camel-ai/camel-langchain4j-embeddingstore/src/main/java/org/apache/camel/component/langchain4j/embeddingstore/LangChain4jEmbeddingStoreProducer.java
```diff
@@ -236,6 +228,40 @@ private void search(Exchange exchange) throws Exception {
...
 
+    private EmbeddingResult resolveEmbedding(Exchange exchange) throws NoSuchHeaderException {
+        final Message in = exchange.getMessage();
+        LangChain4jEmbeddingStoreConfiguration config = getEndpoint().getConfiguration();
+
+        Embedding embedding = in.getHeader(LangChain4jEmbeddingsHeaders.EMBEDDING, Embedding.class);
+        TextSegment textSegment = in.getHeader(LangChain4jEmbeddingsHeaders.TEXT_SEGMENT, TextSegment.class);
+
+        if (embedding == null && config.getEmbeddingModel() != null) {
+            String text = in.getBody(String.class);
```
gnodet, line 239: The null-body check here is unconditional within the auto-embed block, but it shouldn't be. When a `TextSegment` header is already present (fetched at line 236), the body text isn't needed — `embed(textSegment)` at line 249 will use the header-provided segment. Currently, sending a message with a `TextSegment` header but a null/non-String body will throw `IllegalArgumentException` even though there's enough information to compute the embedding.

Consider moving the body-fetch and null-check inside the `if (textSegment == null)` block:

```suggestion
            if (textSegment == null) {
                String text = in.getBody(String.class);
                if (text == null) {
                    throw new IllegalArgumentException(
                            "Message body cannot be converted to String for auto-embedding. "
                                                       + "Either set the body to a text value or provide a pre-computed embedding via the "
                                                       + LangChain4jEmbeddingsHeaders.EMBEDDING + " header.");
                }
                textSegment = TextSegment.from(text);
            }
```

Note: the existing test `addWithTextSegmentHeaderPreservedDuringAutoEmbed` always supplies a non-null body, so it doesn't catch this edge case.

gnodet (commented):
Clean feature with good backward compatibility and solid test coverage. The `resolveEmbedding()` extraction is a nice factoring. One logic issue found in the null-body guard, plus a minor Javadoc note.

PR #25306, opened 2026-08-03
CAMEL-24310: camel-mcp-server - bridge, McpServerEngine SPI and Vert.x engine

Croway:
# CAMEL-24310: camel-mcp-server — bridge, McpServerEngine SPI and Vert.x engine

Sub-tasks of [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308): implements [CAMEL-24310](https://issues.apache.org/jira/browse/CAMEL-24310) and [CAMEL-24312](https://issues.apache.org/jira/browse/CAMEL-24312), plus the conformance-kit half of [CAMEL-24313](https://issues.apache.org/jira/browse/CAMEL-24313).

> **Stacked on #25301** (CAMEL-24309 registry listener SPI) — the first commit here is that PR; review only the `CAMEL-24310` commit. Will rebase once #25301 merges.

## What this adds

Two new modules under `components/camel-ai` exposing `ai-tool` routes (CAMEL-23382) as MCP tools over streamable HTTP. No route is needed for the server itself.

### `camel-mcp-server-api` — runtime-agnostic bridge + SPI
- `McpServerEngine` SPI (`initialize`/`toolAdded`/`toolRemoved`), resolution mirroring `PlatformHttpEngine`: registry bean → FactoryFinder (`mcp-server-engine`) → fail with a clear message.
- `McpServerBridge` (CamelContext service): selects registry tools by tag — **never** the untagged default pool (external MCP clients are untrusted senders); refuses flat-namespace name collisions with an ERROR log; subscribes to the `AiToolRegistry` listener **before** snapshotting (no race window); executes via `AiToolExecutor` on a managed pool with a bounded per-call timeout; maps `AiToolResult` to pre-sanitized results — raw route exception messages never reach the engine or the client. A timed-out call does not return its pooled exchange (the route may still be using it).
- Enforcer rule bans the MCP SDK, Reactor, Vert.x and platform-http from compile/runtime scope — a leak would drag Reactor onto the Quarkus/Spring Boot classpath.
- Ships the **engine conformance kit** as a test-jar: the same SDK-client-driven scenarios (tools/list tag filtering, calls, sanitization, timeout, list changes) will run against the Quarkus and Spring Boot engines in their repos, so behavior cannot drift.

### `camel-mcp-server` — Vert.x engine for Camel Main / JBang
[...]

PR #25397, opened 2026-08-06
CAMEL-24327: MCP session keep-alive and idle TTL eviction

atiaomar1978-hub:
_AI-generated PR on behalf of atiaomar1978-hub via Cursor Cloud Agent._

## Summary

Implements [CAMEL-24327](https://issues.apache.org/jira/browse/CAMEL-24327): active eviction for orphaned MCP streamable HTTP sessions on the Vert.x transport (follow-up from CAMEL-24310 / PR #25306).

The session map in `VertxMcpStreamableServerTransportProvider` previously grew on every `initialize` and was cleaned up only on explicit DELETE, failed SSE writes, or shutdown. Clients that disconnected without DELETE could leak session entries indefinitely.

## Changes

- **Keep-alive pings** (default 30s): periodic pings on the worker pool; sessions with an open SSE stream are evicted after 3 consecutive ping failures. POST-only sessions without a GET stream are not evicted by keep-alive (idle TTL covers orphans).
- **Idle TTL** (default 5 minutes): evict sessions with no POST/GET activity. Set either option to `0` to disable.
- **Configuration**: `McpServerConfiguration`, `McpServerInfo`, `camel.server.mcp-session-keep-alive-interval`, `camel.server.mcp-session-idle-ttl`
- **Async close** on SSE write failure to avoid blocking the Vert.x event loop
- **Docs** + upgrade guide entry for new default eviction behavior

## Tests

- `VertxMcpSessionEvictionTest` — idle eviction, POST-only survives keep-alive, active session retention, explicit DELETE
- `McpServerSessionEvictionPropertiesTest` — Main property wiring for idle TTL
- Existing `VertxMcpServerConformanceTest` still passes

All targeted tests pass locally.


davsclaus:
@atiaomar1978-hub there is a merge conflict

atiaomar1978-hub:
on it

components/camel-ai/camel-mcp-server/src/main/java/org/apache/camel/component/mcp/server/vertx/VertxMcpStreamableServerTransportProvider.java
```diff
@@ -338,6 +469,35 @@ private void endWithStatus(Context connection, RoutingContext ctx, int status) {
         connection.runOnContext(v -> ctx.response().setStatusCode(status).end());
     }
 
+    private static final class ManagedSession {
+
+        private final McpStreamableServerSession session;
+        private volatile long lastActivityNanos;
+        private volatile int consecutivePingFailures;
```
davsclaus: Minor: `++consecutivePingFailures` on a `volatile int` is not atomic — it's a read-modify-write. Since `touch()` can reset this field concurrently from HTTP handler threads (both run via `executeBlocking(... false)` on different worker threads), this is technically a data race.

Practically benign (a lost reset just means the session survives one extra ping cycle), but `AtomicInteger` with `incrementAndGet()` / `set(0)` would eliminate the race cleanly.

davsclaus (approved):
Clean, well-structured PR that addresses a real session leak on the Vert.x MCP streamable transport. The implementation correctly distinguishes SSE-connected sessions (pingable) from POST-only sessions (idle TTL only), and respects the Vert.x threading model throughout.

1. **`ManagedSession.consecutivePingFailures`** — `++consecutivePingFailures` on a `volatile int` is a non-atomic read-modify-write. `touch()` can reset it concurrently from HTTP handler threads. Practically benign (worst case: session survives one extra ping cycle), but an `AtomicInteger` would be the clean fix.

2. **Version fallback in `McpServerBridge.doInit()`** — the `version == null || version.isBlank()` guard is a reasonable defensive fix but unrelated to session eviction scope.

Configuration wiring, test coverage, documentation, and upgrade guide entry all look good.

atiaomar1978-hub:
_AI-generated comment on behalf of atiaomar1978-hub via Cursor Cloud Agent._

Rebased onto `main` (resolved conflict with CAMEL-24332 `buildMcpTool`/annotations in `VertxMcpServerEngine`) and pushed concurrency hardening.

## Bugbot review

- **Overlapping periodic maintenance** — addressed with in-flight guards (`AtomicBoolean`) so keep-alive/idle sweeps skip if a prior run is still executing.
- **POST-only leak when idle TTL disabled** — documented behavior: with idle TTL `0`, POST-only sessions rely on explicit DELETE or process shutdown; keep-alive intentionally skips sessions without an SSE stream. Default config enables both mechanisms (30s keep-alive + 5m idle TTL).

## Grok review

**Verdict:** request-changes → **addressed** rebase/conflicts and overlapping sweeps; `AtomicInteger` for ping failures.

**Remaining nice-to-haves (non-blocking):**
- Integration test for dead GET/SSE eviction via keep-alive
- Stop `CamelContext` in eviction tests for cleaner teardown

All targeted tests pass locally after rebase.

davsclaus (approved):

PR #25400, opened 2026-08-07
CAMEL-24346: camel-google-sheets - number every range and tolerate a range without values

oscerd:
Two defects in `GoogleSheetsStreamConsumer`, both covered by a new test.

**1. `splitResults` reported range index 1 for every value.** The counter was allocated *inside* the
loop over the ranges:

```java
for (ValueRange valueRange : response.getValueRanges()) {
    AtomicInteger rangeIndex = new AtomicInteger(1);
    ...
    rangeIndex.incrementAndGet();   // dead: a fresh counter next iteration
}
```

So a consumer configured with `range=A1:B2,C1:D2&splitResults=true` sent every exchange with
`CamelGoogleSheetsRangeIndex=1`, and there was no way to tell which range a value came from. The
counter now spans the whole response, which is what the non-split branch already did.

**2. A range without values threw.** `valueRange.getValues()` was dereferenced directly in both
branches, but the Sheets API omits the `values` field for a range that holds nothing, so polling a
spreadsheet with an empty range failed with a `NullPointerException` on every cycle.

Two smaller things while in there: the exchange building moved into `createExchanges(List<ValueRange>)`
so it can be tested without a Google client (the API client needs a real transport), and the branch
that reads a whole spreadsheet — used when no `range` is configured — now calls `forceConsumerAsReady()`
as well, so the consumer health check reports ready on that path too.

### DIFF
--- a/components/camel-google/camel-google-sheets/pom.xml
+++ b/components/camel-google/camel-google-sheets/pom.xml
@@ -149,6 +149,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-mock</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean, well-tested fix for two real bugs in `GoogleSheetsStreamConsumer`:

1. **Range index counter bug**: `AtomicInteger rangeIndex` was declared inside the for-loop, so it was re-created on every iteration — always reporting `1`. The fix correctly moves it outside the loop.

2. **Null values NPE**: The Google Sheets API omits the `values` field for empty ranges, causing `getValues()` to return `null`. The new `valuesOf()` helper returns `List.of()` as a safe default.

Bonus fix: `forceConsumerAsReady()` was missing in the non-range code path (the range-based path was covered by CAMEL-20189 but this branch was missed).

Test coverage is thorough, uses AssertJ, and the extracted `createExchanges()` method enables clean unit testing without a real Google API client. LGTM.

davsclaus (approved):

PR #25406, opened 2026-08-07
CAMEL-24371: camel-a2a - fix WebhookUrlValidator address classification and host matching

oscerd:
## Description

`WebhookUrlValidator` classified webhook hosts in two places that did not agree with each other.

A host written as an IP literal was checked against a string prefix list (`isPrivateIpv6`), while a host reached through a name was classified with the `InetAddress` predicates. Those two do not cover the same ground — `Inet6Address.isSiteLocalAddress()` reports the deprecated `fec0::/10` block and **not** the `fc00::/7` unique local addresses that replaced it — so the same address was accepted or rejected depending on how it was written. Verified on JDK 21: `fd00::1` and `fc00::1` return `false` for all four predicates, while `fec0::1` returns `site=true`.

The literal pre-check also prefix-matched the raw host string with `startsWith("fc")` / `startsWith("fd")` without first establishing that the host was an IP literal, so any name beginning with those two characters was rejected outright. `https://fcm.googleapis.com/webhook` and `https://fd-edge.example.com/webhook` were both refused.

## Changes

Both paths now resolve the host and classify the resulting address with a single shared raw-byte classifier. `InetAddress.getByName()` already parses bracketed IPv6 literals without touching DNS, so the separate literal path is no longer needed and the two forms cannot drift apart again.

The classifier recognises these in addition to the loopback / wildcard / link-local / site-local ranges already handled:

| Range | Example |
|---|---|
| IPv6 unique local, `fc00::/7` | `fd00::1` |
| IPv4-compatible IPv6 | `::10.0.0.1` |
| NAT64 well-known prefix, `64:ff9b::/96` | `64:ff9b::a00:1` |
| 6to4, `2002::/16` | `2002:c0a8:101::1` |
| Shared address space, `100.64.0.0/10` | `100.64.0.1` |

NAT64 and 6to4 addresses are classified by the IPv4 address they embed rather than blocked wholesale, so an IPv6-only deployment can still reach public webhook endpoints through a translation prefix.

IPv4-mapped IPv6 needs no new handling — the JDK returns an `Inet4Address` for `::ffff:x.x.x.x`, so the existing IPv4 rules already applied.

Host names are no longer classified by their spelling.

## Notes for reviewers

- **Exception message text changes.** `[fd00::1]` now reports `unique local` instead of `private/internal IPv6 ranges`, and `[::1]` reports `loopback` instead of the IPv6 message. Four existing test assertions were updated accordingly. No API signature changed.
- **`nonGlobalReason` checks loopback** even though a host reaching loopback directly is answered earlier (where `allowLocal` can let it through). That branch exists for the addresses embedding an IPv4 one — `64:ff9b::7f00:1` carries `127.0.0.1` — and is covered by a test.
- A package-private `HostResolver` seam lets the resolved-host path be tested without DNS and without installing a JVM-wide `InetAddressResolverProvider`, which would have affected every test in the module.

## Testing

`WebhookUrlValidatorTest` goes from 13 to 30 tests, covering each range above in both literal and resolved form, the adjacent addresses just outside `100.64.0.0/10`, the NAT64/6to4 cases that must stay allowed, and the host names that were previously rejected for their spelling.

- `camel-a2a` module suite: 519 tests, all passing
- Full reactor build from root (`mvn clean install -DskipTests`): SUCCESS, no uncommitted regen artifacts

## Documentation

Adds `camel-4x-upgrade-guide-4_23.adoc` with a note on the newly rejected ranges and the relaxed host-name matching, and registers it in the upgrade guide index.

---

### DIFF
--- a/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/util/WebhookUrlValidator.java
+++ b/components/camel-ai/camel-a2a/src/main/java/org/apache/camel/component/a2a/util/WebhookUrlValidator.java
@@ -20,13 +20,21 @@ import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 
 /**
- * Validates webhook URLs for SSRF protection in A2A push notifications. All hostnames (including {@code localhost}) are
- * resolved to their IP address and classified consistently. Loopback addresses are blocked by default; set
- * {@code allowLocal=true} for local development.
+ * Validates webhook URLs for SSRF protection in A2A push notifications. Every host, whether it is written as an IP
+ * literal or as a name that has to be resolved, is turned into an address and classified by the same rules, so the two
+ * forms can never disagree. Loopback addresses are blocked by default; set {@code allowLocal=true} for local
+ * development.
  */
 public final class WebhookUrlValidator {
+
+    private static final int IPV4_LENGTH = 4;
+
+    /** 64:ff9b::/96, the well-known prefix for IPv4/IPv6 translation (RFC 6052). */
+    private static final byte[] NAT64_WELL_KNOWN_PREFIX
+            = { 0x00, 0x64, (byte) 0xff, (byte) 0x9b, 0, 0, 0, 0, 0, 0, 0, 0 };
 
     private WebhookUrlValidator() {
     }
@@ -62,6 +70,13 @@ public final class WebhookUrlValidator {
      * @throws IllegalArgumentException if the URL is invalid or unsafe
      */
     public static InetAddress validateAndResolve(String url, boolean allowLocal) {
+        return validateAndResolve(url, allowLocal, InetAddress::getByName);
+    }
+
+    /**
+     * Validates against a supplied resolver, so the resolved-host path can be exercised without depending on DNS.
+     */
+    static InetAddress validateAndResolve(String url, boolean allowLocal, HostResolver resolver) {
         if (url == null || url.isBlank()) {
             throw new IllegalArgumentException("Webhook URL must not be null or empty");
         }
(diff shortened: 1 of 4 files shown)


gnodet (commented):
The code changes to `WebhookUrlValidator` look solid — the single byte-based classifier is a clear improvement over the dual-path string+predicate approach, fixing real bugs where hostnames starting with "fc" or "fd" (like `fcm.googleapis.com`) were falsely rejected, and where `fc00::/7` unique local addresses weren't caught by `InetAddress.isSiteLocalAddress()`. The `HostResolver` seam is clean for testing DNS-dependent code. Test coverage expansion from 13 to 30 tests is great, with proper AssertJ usage and conventions followed throughout.

One documentation concern: the upgrade guide entry is placed in a **new** `camel-4x-upgrade-guide-4_23.adoc` file, but `main` is currently at `4.22.0-SNAPSHOT` (latest tag: `camel-4.21.0`). Changes merging to `main` now will ship with **4.22.0**, so the behavioral changes should be appended to the **existing** `camel-4x-upgrade-guide-4_22.adoc` (which already has a camel-a2a section from CAMEL-23876). The new 4_23 file and its index entry in `camel-4x-upgrade-guide.adoc` are premature — users upgrading to 4.22 would miss this documentation.

davsclaus (approved):
Clean, well-tested security bug fix. The unified byte-based classifier is a clear improvement over the dual-path string+predicate approach, fixing real bugs (hostnames like `fcm.googleapis.com` falsely rejected, `fc00::/7` not caught by `isSiteLocalAddress()`). Test expansion from 13 to 30 tests is thorough, covering each range in both literal and resolved form with proper boundary conditions. The `HostResolver` seam is a clean testability pattern.

Regarding the upgrade guide version: `camel-4.22.0` was tagged today from the release branch, so the new `camel-4x-upgrade-guide-4_23.adoc` is the correct place for these behavioral changes — they will ship in 4.23.0, not 4.22.0.

This review covers project rules and conventions. It is not a replacement for specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

davsclaus (approved):

PR #25412, opened 2026-08-09
CAMEL-24375: parseDuration should handle plain millis value

davsclaus:
## Summary

- `CamelContextHelper.parseDuration` now handles the String-to-Duration conversion directly using `TimeUtils.toMilliSeconds` instead of delegating to the type converter registry.
- This fixes `NoTypeConversionAvailableException` when parsing plain millisecond values like `"20000"` during early route initialization (e.g. Resilience4j circuit breaker configuration) when the `DurationConverter` type converter is not yet loaded.

Supported formats: plain millis (`20000`), human-readable (`20s`, `1m30s`, `500ms`), ISO-8601 (`PT20S`).

## Test plan

- [x] New `CamelContextHelperParseDurationTest` covering plain millis, human-readable, ISO-8601, and null input
- [x] `mvn test -Dtest=CamelContextHelperParseDurationTest` passes

### DIFF
--- a/core/camel-base/src/main/java/org/apache/camel/converter/DurationConverter.java
+++ b/core/camel-base/src/main/java/org/apache/camel/converter/DurationConverter.java
@@ -45,11 +45,7 @@ public final class DurationConverter {
 
     @Converter(order = 3)
     public static Duration toDuration(String source) {
-        if (source.startsWith("P") || source.startsWith("-P") || source.startsWith("p") || source.startsWith("-p")) {
-            return Duration.parse(source);
-        } else {
-            return Duration.ofMillis(TimeUtils.toMilliSeconds(source));
-        }
+        return TimeUtils.toDuration(source);
     }
 
     @Converter(order = 4)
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean, well-focused fix that eliminates the timing dependency on the type converter registry by inlining the ISO-8601/human-readable/millis parsing logic directly in `parseDuration()`.

**Highlights:**
- The new code mirrors `DurationConverter.toDuration(String)` exactly, which is the right approach given the goal of avoiding converter registry dependency.
- Tests cover the main input formats (plain millis, ISO-8601, human-readable) with proper AssertJ assertions and package-private visibility.
- CI is green.

- The parsing logic at lines 472-477 duplicates `DurationConverter.toDuration()`. Since `TimeUtils.toDuration(String)` already exists and does exactly this, a possible follow-up would be to call `TimeUtils.toDuration(s)` instead — keeping both sites in sync automatically.
- Minor behavioral change: the old path wrapped conversion failures in `IllegalArgumentException` with a contextual message; the new code propagates a raw `DateTimeParseException`. No callers appear to catch `IllegalArgumentException` specifically, so this is unlikely to matter.

davsclaus:
Thanks for the review @gnodet. Addressed the suggestion to consolidate the duration parsing logic — moved the ISO-8601 handling into `TimeUtils.toDuration(String)` so both `CamelContextHelper.parseDuration()` and `DurationConverter.toDuration()` now delegate to it. Added tests for the new ISO-8601 path in `TimeUtilsTest` as well.

davsclaus:
Need to be backported to 4.22.x

PR #25430, opened 2026-08-10
CAMEL-24347: camel-google-firestore - listCollections honours the configured documentId

oscerd:
`listCollections` was the only document operation that read the document id straight from the header:

```java
String documentId = exchange.getIn().getHeader(GoogleFirestoreConstants.DOCUMENT_ID, String.class);
```

Every other operation goes through `determineDocumentId(exchange)`, which falls back to the
`documentId` endpoint option. So `google-firestore:users?documentId=alice&operation=listCollections`
silently listed the **root** collections instead of the sub-collections of `users/alice`.

The lookup now falls back to the configured id as well, while keeping what makes this operation
different: a missing id still means "list the root collections", so it must not fail the way
`determineDocumentId` does. That is `determineListedDocumentId`, covered by the new test.

Also in this PR: `queryCollection` and `listDocuments` added their `_id`/`_path` entries to the map
returned by `QueryDocumentSnapshot.getData()` — the object the SDK handed us. They now copy it first.
The body content is unchanged.

**Not addressed here** (from the same audit, needs a maintainer call on the policy): the realtime
consumer queues snapshot changes in an unbounded `ConcurrentLinkedQueue` that is only drained per
poll, so a collection changing faster than the route consumes grows it without limit. Bounding it
means choosing an overflow policy — dropping changes or blocking the Firestore callback thread —
and `google-firestore` exposes no `maxMessagesPerPoll` today. I left a note on the JIRA issue.

### DIFF
--- a/components/camel-google/camel-google-firestore/pom.xml
+++ b/components/camel-google/camel-google-firestore/pom.xml
@@ -71,5 +71,10 @@
             <artifactId>camel-test-junit6</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/components/camel-google/camel-google-firestore/src/main/java/org/apache/camel/component/google/firestore/GoogleFirestoreProducer.java
+++ b/components/camel-google/camel-google-firestore/src/main/java/org/apache/camel/component/google/firestore/GoogleFirestoreProducer.java
@@ -243,16 +243,24 @@ public class GoogleFirestoreProducer extends DefaultProducer {
         List<Map<String, Object>> results = new ArrayList<>();
 
         for (QueryDocumentSnapshot document : querySnapshot.getDocuments()) {
-            Map<String, Object> docData = document.getData();
-            docData.put("_id", document.getId());
-            docData.put("_path", document.getReference().getPath());
-            results.add(docData);
+            results.add(withDocumentMetadata(document));
         }
 
         LOG.debug("Query returned {} documents from collection: {}", results.size(), collectionName);
 
         Message message = getMessageForResponse(exchange);
         message.setBody(results);
+    }
+
+    /**
+     * The document data plus its id and path. The snapshot data is copied first: the returned map is handed to the
+     * route, and the id/path entries must not be written into the object the SDK gave us.
+     */
+    private static Map<String, Object> withDocumentMetadata(QueryDocumentSnapshot document) {
+        Map<String, Object> docData = new HashMap<>(document.getData());
+        docData.put("_id", document.getId());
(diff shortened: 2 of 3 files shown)


davsclaus (approved):

davsclaus (approved):

atiaomar1978-hub:
## Bugbot / Grok Review — PR #25430 (CAMEL-24347)

**Verdict:** ✅ **Approve** — focused bug fix, CI green, already approved by @davsclaus

### Summary

Fixes two real issues in `camel-google-firestore`:

1. **`listCollections` ignored endpoint `documentId`** — the only operation that read the id from the header alone, so `?documentId=alice` silently listed root collections instead of sub-collections under `users/alice`.
2. **`queryCollection` / `listDocuments` mutated SDK maps** — `_id` / `_path` were written into the map returned by `QueryDocumentSnapshot.getData()` instead of a copy handed to the route.

### What looks good

| Area | Assessment |
|------|------------|
| **Correctness** | `determineListedDocumentId()` mirrors `determineDocumentId()` header→option fallback but correctly keeps missing id as “list root collections” (no throw). |
| **Consistency** | Same `ObjectHelper.isEmpty()` semantics as other operations (blank header falls back to endpoint option). |
| **Side-effect fix** | `withDocumentMetadata()` copies via `new HashMap<>(document.getData())` before adding metadata — good hygiene. |
| **Tests** | New `GoogleFirestoreProducerDocumentIdTest` covers all three resolution paths; AssertJ, package-private class/methods, no `Thread.sleep`. |
| **Scope** | Small, well-scoped diff; PR description honestly scopes out the unbounded consumer queue (needs maintainer policy). |
| **CI** | All checks passing (Java 17/25 builds). |

### Non-blocking follow-ups

1. **`withDocumentMetadata` test** — consider a small unit test with a mocked `QueryDocumentSnapshot` to assert the SDK map is never mutated (regression guard).
2. **Upgrade guide** — optional one-liner under 4.22: `listCollections` now honours endpoint `documentId` (behaviour fix; previously listed root collections when id was only configured on the URI).
3. **Consumer queue growth** — already noted on JIRA; future work should pick an overflow policy (`maxMessagesPerPoll` or bounded queue + drop/block strategy).
4. **`getDocumentById`** returns `document.getData()` without copy — fine today (read-only), but if routes mutate the body map, same class of issue could appear; low priority.

### Bugbot automated scan

No additional defects flagged by automated review on this changeset.

**Recommendation:** Merge when ready — behaviour fix is correct, tests are adequate for the id-resolution logic, and the map-copy change is a clear improvement.

_AI-generated review on behalf of atiaomar1978-hub_

atiaomar1978-hub (approved):
LGTM — the listCollections documentId fallback and withDocumentMetadata copy fix look correct. Tests cover the id-resolution paths. See prior Bugbot/Grok review comment for details.

_AI-generated approval on behalf of atiaomar1978-hub_

PR #25443, opened 2026-08-11
Pin GitHub actions

coheigea:
It promotes security as Github tags can be deleted


oscerd (approved):

davsclaus:
I guess bots know about this so they can keep raising PRs when there are newer versions ?

coheigea:
> I guess bots know about this so they can keep raising PRs when there are newer versions ?

Yep, dependabot works fine with it and will update the version in the comment as well.

davsclaus (approved):

PR #25444, opened 2026-08-11
feat(components): micrometer supports prometheus format

squakez:
Add a conversion to support the feature from this component instead of the prometheus one

Ref CAMEL-24356

# Description

# Target

# Tracking

# AI-assisted contributions


components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/json/AbstractMicrometerService.java
```diff
@@ -243,19 +264,165 @@ static Map<String, Object> convertMeterToMap(Meter meter) {
...
+            lines.add("# TYPE " + promName + "_seconds summary");
+            lines.add(promName + "_seconds_count" + labels + " " + formatPrometheusDouble(ft.count()));
+            lines.add(promName + "_seconds_sum" + labels + " " + formatPrometheusDouble(ft.totalTime(TimeUnit.SECONDS)));
+        } else if (meter instanceof LongTaskTimer ltt) {
+            lines.add("# HELP " + promName + "_active_seconds " + description);
+            lines.add("# TYPE " + promName + "_active_seconds gauge");
+            lines.add(promName + "_active_seconds_active" + labels + " " + ltt.activeTasks());
+            lines.add(promName + "_active_seconds_duration" + labels + " "
+                      + formatPrometheusDouble(ltt.duration(TimeUnit.SECONDS)));
+            lines.add(promName + "_active_seconds_max" + labels + " "
```
davsclaus: The LongTaskTimer naming here (`_active_seconds_active`, `_active_seconds_duration`, `_active_seconds_max`) doesn't follow the standard Prometheus/Micrometer naming convention. The standard format would be:

```
<name>_seconds_active_count  (active tasks)
<name>_seconds_duration_sum  (total duration)
<name>_seconds_max           (max duration)
```

Using the standard naming would be more consistent with what users see from `PrometheusMeterRegistry.scrape()` and avoids confusion when comparing log output to scrape output.

components/camel-micrometer/src/test/java/org/apache/camel/component/micrometer/json/AbstractMicrometerServicePrometheusFormatTest.java
```diff
@@ -0,0 +1,237 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.micrometer.json;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import io.micrometer.core.instrument.*;
```
davsclaus: Minor: Camel conventions use explicit imports rather than wildcard imports. Running `mvn install -DskipTests` in the module should auto-format these.

components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/json/AbstractMicrometerService.java
```diff
@@ -243,19 +264,165 @@ static Map<String, Object> convertMeterToMap(Meter meter) {
...
-                        // scraper can more easily identify the metric content.
-                        String metric = "#METRIC-START#" + mapper.writeValueAsString(logEntry) + "#METRIC-END#";
-                        LOG.info(metric);
-                    } catch (Exception e) {
-                        LOG.error("Error logging metric " + logEntry.get("name"), e);
+        if ("prometheus".equalsIgnoreCase(logMetricsOnShutdownFormat)) {
+            List<Meter> meters = meterRegistry.getMeters().stream()
+                    .filter(m -> AbstractMicrometerService.matchesFilter(
+                            m.getId().getName(), filters))
+                    .sorted(Comparator.comparing(m -> {
```
davsclaus: Minor: this can be simplified to a method reference:

```suggestion
                    .sorted(Comparator.comparing(AbstractMicrometerService::normalizePrometheusName))
```

davsclaus (commented):
Nice work moving the Prometheus format support down to `camel-micrometer` — making it available without the Prometheus registry dependency is a good architectural improvement. The test coverage is thorough.

A few items below — one naming concern and two minor style points.

**Note:** This review covers project rules, conventions, and code correctness visible from the diff. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/json/AbstractMicrometerService.java
squakez: Refactorying to remove the "active" suffix

components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/eventnotifier/AbstractMicrometerEventNotifier.java
```diff
@@ -41,6 +41,7 @@ public abstract class AbstractMicrometerEventNotifier<T extends CamelEvent> exte
     private boolean skipCamelInfo = false;
     private boolean logMetricsOnShutdown = false;
     private String logMetricsOnShutdownFilters[];
+    private String logMetricsOnShutdownFormat;
```
davsclaus: This field defaults to `null`, but `doStart()` unconditionally passes it to `registryService.setLogMetricsOnShutdownFormat(...)`, which overwrites the service's `"json"` default with `null` when the notifier is used standalone (without `MicrometerPrometheus`). The code still *behaves* correctly because `"prometheus".equalsIgnoreCase(null)` is false, but a `getLogMetricsOnShutdownFormat()` call on the service would return `null` instead of the documented `"json"` default.

Note how `MicrometerPrometheus.doInit()` guards the setter with `if (getLogMetricsOnShutdownFormat() != null)` — this class should be consistent.

```suggestion
    private String logMetricsOnShutdownFormat = "json";
```

components/camel-micrometer/src/test/java/org/apache/camel/component/micrometer/json/AbstractMicrometerServicePrometheusFormatTest.java
```diff
@@ -0,0 +1,246 @@
...
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import org.junit.jupiter.api.Test;
```
davsclaus, line 32: Per project conventions, new test code should prefer AssertJ assertions (`assertThat(...)`) over JUnit assertions (`assertEquals`, `assertTrue`). AssertJ is already available as a test dependency.

For example:
```java

assertThat(lines.get(0)).isEqualTo("# HELP app_info Application info");
assertThat(lines).hasSize(3);
assertThat(lines.get(2)).endsWith(" NaN");
assertThat(valueLine).startsWith("app_info{").contains("camel_version=\"4.0.0\"");
```

davsclaus (changes requested):
Thanks for moving the Prometheus format support down to `camel-micrometer` — nice refactor that makes the feature available without requiring the prometheus extension.

I found one code-quality issue that should be fixed before merging, plus a convention item.

This review focuses on project conventions and code correctness. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

davsclaus (approved):

PR #25448, opened 2026-08-11
CAMEL-24348: camel-google-bigquery - report unusable bodies and missing jobs properly

oscerd:
Three NPE/exception-hygiene fixes.

**1. A message with no body failed with a `NullPointerException`.** The producer built its error out
of the body's class:

```java
ex.setException(new IllegalArgumentException("Cannot handle body type " + entryObject.getClass()));
```

so the branch meant to report an unusable body threw while reporting it. It now says
`Cannot handle body type null`.

**2. An unknown `CamelGoogleBigQueryJobId` gave a `NullPointerException`.** In the SQL producer both
`bigquery.getJob(jobId)` and `job.waitFor()` return `null` for a job that no longer exists (expired,
or never created), and neither was checked before use. Both now fail with a message naming the job.

**3. Three raw `throw new Exception(...)`** in `GoogleBigQueryProducer` and `GoogleBigQuerySQLProducer`
are now `RuntimeCamelException`, so routes get a Camel exception. `onException(Exception.class)`
still matches, since `RuntimeCamelException` is one.

**Withdrawn from the issue description:** the audit also claimed the `CamelGoogleBigQueryInsertId`
header is wrongly ignored for `List` bodies. Applying it there would be a **bug**: BigQuery uses the
insert id for best-effort deduplication, so giving every row of a list the same id makes it keep one
row and drop the rest. The existing behaviour is right — `useAsInsertId` (a field name, so one id per
row) is the option for list payloads. This PR only adds a comment recording that, and I commented on
the JIRA issue.

### DIFF
--- a/components/camel-google/camel-google-bigquery/src/main/java/org/apache/camel/component/google/bigquery/GoogleBigQueryProducer.java
+++ b/components/camel-google/camel-google-bigquery/src/main/java/org/apache/camel/component/google/bigquery/GoogleBigQueryProducer.java
@@ -26,6 +26,7 @@ import com.google.cloud.bigquery.InsertAllRequest;
 import com.google.cloud.bigquery.InsertAllResponse;
 import org.apache.camel.Exchange;
 import org.apache.camel.ExchangePropertyKey;
+import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.support.DefaultProducer;
 import org.apache.camel.util.ObjectHelper;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


davsclaus (approved):

davsclaus (approved):
Looks good — three well-scoped fixes for real NPE/exception-hygiene issues.

**What's good:**
- Null-body guard in `GoogleBigQueryProducer` correctly reports instead of crashing, and the loop + `apiRequestRows.isEmpty()` handle it gracefully
- Null checks on `getJob()` / `waitFor()` in `GoogleBigQuerySQLProducer` cover the realistic failure paths with descriptive messages
- `RuntimeCamelException` is the right Camel exception type — propagates cleanly through the routing engine
- The insert-ID comment in the `List` branch is a useful "why" note that prevents someone from re-introducing the dedup bug
- New test class follows conventions: AssertJ, package-private, covers both null and non-map/list body cases

- No unit tests for the SQL producer null-job handling — understandable given the mocking complexity, worth noting as a future improvement

Croway (approved):

PR #271, opened 2014-09-20
CAMEL-5958: Bindy ignores bean class type

lburgazzoli:
re-build PR against latest sources

PR #25471, opened 2026-08-12
chore(camel-test-infra-docling): upgrade docling.container to v1.30.0

davsclaus:
## Summary

- Upgrade docling container from v1.29.0 to v1.30.0
- Update `OcrExtractionIT.testOcrWithMultipleTextBlocks` sentinel assertion: docling v1.30.0 now includes page furniture (headers/footers) in the default body export, so the test now asserts footer text **is** present (was previously asserting it was absent)
- Update documentation note in `docling-component.adoc` to reflect the new behavior

Supersedes #25417 which had merge conflicts. Addresses the behavioral change flagged by @gnodet's local IT results.

## Test plan

- [ ] CI build passes
- [ ] Integration tests with docling v1.30.0 pass (the OCR test assertion is updated to match new behavior)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/docs/docling-component.adoc
+++ b/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/docs/docling-component.adoc
@@ -301,18 +301,11 @@ YAML::
 === OCR and page headers/footers
 
 Docling recognizes page headers and footers (page numbers, copyright lines, running titles, footnotes, and similar
-content) during OCR, but it classifies them as page _furniture_: they are assigned to docling's `FURNITURE` content
-layer rather than the document `BODY`. Docling's Markdown, text and HTML exports include only the `BODY` layer by
-default, so header and footer text is omitted from the converted output even though the OCR engine read it correctly.
+content) during OCR, classifying them as page _furniture_ in docling's `FURNITURE` content layer.
 
-When using `useDoclingServe=true`, the component cannot currently change this, because the docling-serve API does not
-expose an option to choose which content layers are included in the export — even though docling's Python library can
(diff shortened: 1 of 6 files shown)


gnodet (approved):
Clean container upgrade from docling-serve v1.29.0 to v1.30.0 with correctly adapted test assertion and documentation.

**Verified:**
- Version bump follows established pattern (14+ prior docling container upgrades in git history).
- The 36 deletions are fully justified: old documentation had a lengthy workaround section (referencing upstream issue #271, suggesting document restructuring) that is no longer needed since furniture is now included by default. Test similarly had explanatory comments replaced with shorter ones.
- Container image (`quay.io/docling-project/docling-serve:v1.30.0`) hosted on quay.io (not Docker Hub), aligning with project convention.
- Architecture support correctly handled: `skipITs.s390x=true` and `skipITs.ppc64le=true` already set (image only supports amd64/arm64).
- Catalog copies and metadata.json consistently updated.
- CI green on both JDK 17 and JDK 25.
- The sentinel test pattern worked exactly as designed — the original `assertFalse` was documented as a canary that would fire when a future docling release includes page furniture. It fired, and this PR correctly responds.

**One minor observation (non-blocking):** The documentation states the furniture-layer inclusion is an intentional v1.30.0 feature, but the upstream docling-serve v1.30.0 release notes don't explicitly mention this change, and upstream issue [docling-serve#271](https://github.com/docling-project/docling-serve/issues/271) for content-layer control remains open. The behavior is real (confirmed by IT results) but may be an undocumented side effect of other changes in bundled docling v2.118.0 rather than a guaranteed feature. Worth noting in case it gets reverted in a future release.

PR #25487, opened 2026-08-13
CAMEL-24388: Guard against duplicate plugin subcommand registration

davsclaus:
## Summary

_Claude Code on behalf of @${OPERATOR}_

When upgrading the Camel CLI, the old `~/.camel-jbang-plugins.json` file may still list plugins that are now bundled as embedded plugins. This causes picocli to throw a `DuplicateNameException` because the same subcommand name gets registered twice.

This fix makes `PluginHelper` guard against duplicate registrations in both plugin loading paths:
- **JSON config path**: the existing `containsKey` guard was conditional on `foundEmbeddedPlugins` — now it fires unconditionally so stale JSON entries are always skipped
- **Embedded plugin path**: added a `containsKey` guard before `plugin.customize()` in `loadPluginFromService()`

Both guards protect all plugins generically.

## Test plan

- [x] Verified the module compiles cleanly
- [ ] Manual test: install Camel CLI 4.22, add a stale `tui` entry to `~/.camel-jbang-plugins.json`, run `camel --version` — should work without `DuplicateNameException`

### DIFF
--- a/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
+++ b/dsl/camel-jbang/camel-jbang-core/src/main/java/org/apache/camel/dsl/jbang/core/common/PluginHelper.java
@@ -170,8 +170,9 @@ public final class PluginHelper {
         // Maven resolution for plugins unrelated to the current command.
         Map<String, Plugin> plugins = getActivePlugins(main, repos, target);
         for (Map.Entry<String, Plugin> plugin : plugins.entrySet()) {
-            // Skip if this plugin was already loaded from embedded plugins
-            if (foundEmbeddedPlugins && commandLine.getSubcommands().containsKey(plugin.getKey())) {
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean, minimal fix that correctly guards against picocli `DuplicateNameException` when stale JSON config or multiple classpath sources register the same plugin command. Both guard changes are logically sound.

- No automated test accompanies this fix. The existing `PluginHelperTest.java` has infrastructure (`CommandLine` setup, `NoOpCommand`) that could be extended to verify duplicate-registration resilience: pre-register a subcommand name, then call `addPlugins` with a JSON config containing the same name, and assert no exception. The fix is a simple defensive `containsKey`-before-register guard, so this is low priority.
- The guard in `loadPluginFromService()` correctly returns `true` when a duplicate is detected ("plugin found, already registered"). Returning `false` would incorrectly signal that no plugin matched, potentially triggering unnecessary Maven resolution fallback.
- The `command != null` check in the new guard at line 760 is technically always true since `extractCommandFromPlugin()` falls through to `return pluginName`, but this is acceptable defensive coding.

PR #25489, opened 2026-08-13
CAMEL-23967: camel-openai - Add image generation and edit operations

k-krawczyk:
# Description

Adds the `image-generation` and `image-edit` operations to `camel-openai`, closing [CAMEL-23967](https://issues.apache.org/jira/browse/CAMEL-23967).

```java
// generate a product image and store it
from("direct:product-image")
    .setBody(simple("Studio photo of ${header.productName} on a white background"))
    .to("openai:image-generation?imageModel=gpt-image-1&imageSize=1024x1024")
    .to("file:target/images?fileName=${header.productName}.png");

// edit an image coming from object storage
from("aws2-s3:marketing-assets")
    .setHeader(OpenAIConstants.IMAGE_PROMPT, constant("Add a red SALE banner in the top-right corner"))
    .to("openai:image-edit?imageModel=gpt-image-1")
    .to("aws2-s3:marketing-assets-processed");
```

## Two deviations from the issue

**`imageResponseFormat` has no default and is only sent when set.** The KDoc of `ImageGenerateParams` in the pinned SDK says the parameter is not supported by the GPT image models, which always return base64. Checking against the live API on 13 August 2026 turned out to be stricter still: `POST /v1/images/generations` answers `400 Unknown parameter: 'response_format'` for every model, and `/v1/models` no longer lists any DALL-E model. The option is kept because OpenAI-compatible providers still implement the older images API, where `url` is often the default — but a default value here would break every route against OpenAI, so there is none.

**The body shape follows the response, not the option.** Base64 payloads are decoded into `byte[]`, URLs stay as `String`. A single image becomes the body directly and several images become a `List`, so the common case does not force routes to unwrap a one-element list.

## What it adds

- `imageModel`, `imagePrompt`, `imageSize`, `imageQuality`, `imageResponseFormat`, `imageCount`, `imageBackground`, `imageOutputFormat`, `imageOutputCompression`, `imageStyle`, `imageModeration`, `imageInputFidelity` — each overridable per exchange by a header.
- `image-edit` takes the image from the body as `File`, `Path`, `InputStream`, `byte[]`, or a `List` of those, since the GPT image models accept up to 16 reference images. An optional mask goes through `CamelOpenAIImageMask`.
- `Content-Type` is set from the output format reported by the response, falling back to the requested one, so the result chains into `file:` or object storage unchanged.
- The multipart parts of an edit declare their content type. The API validates the upload on that content type rather than on the file name, and the SDK leaves it as `text/plain` unless it is set, so an edit is rejected without this. It is resolved from the usual MIME type detection, then from the extension of a `File`/`Path` body, and falls back to `image/png` for anything the API does not accept.
- Revised prompt and token usage as headers; `storeFullResponse=true` keeps the full SDK response in `CamelOpenAIImageResponse`.

Left out, as the issue suggests: `createVariation` and the streaming variants.

## Testing

`camel-test-infra-openai-mock` gained `whenImageGeneration()`, `whenImageEdit()`, `replyWithImage(byte[])`, `replyWithImageUrl(String)`, `withRevisedPrompt()`, `withImageOutputFormat()`, `withImageSize()`, `withImageUsage()` and `assertImageRequest()`. Image edit requests are multipart and the mock does not parse multipart — the same simplification the transcription handler already makes — so those expectations are matched in declaration order and the raw body is exposed to assertions, which is enough to verify that several images and a mask reach the wire.

24 unit tests, no model and no GPU involved.

Beyond the mock, both operations were also exercised against the real OpenAI API once, which is what surfaced the multipart content type and the state of `response_format`. That smoke test is not part of the contribution — there is no CI-suitable local image backend, as the issue notes — but the multipart content type is now asserted in the mock tests.


components/camel-ai/camel-openai/src/test/java/org/apache/camel/component/openai/OpenAIImageEditMockTest.java
```diff
@@ -0,0 +1,252 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OpenAIImageEditMockTest extends CamelTestSupport {
+
+    private static final byte[] SOURCE_IMAGE = "FAKE-PNG-SOURCE".getBytes(StandardCharsets.UTF_8);
+    private static final byte[] SECOND_SOURCE_IMAGE = "FAKE-PNG-SOURCE-TWO".getBytes(StandardCharsets.UTF_8);
+    private static final byte[] MASK_IMAGE = "FAKE-PNG-MASK".getBytes(StandardCharsets.UTF_8);
+    private static final byte[] EDITED_IMAGE = "FAKE-PNG-EDITED".getBytes(StandardCharsets.UTF_8);
```
davsclaus, line 42: Per the project's JUnit 5 convention, new test classes should be package-private (drop `public`). The newer tests in this component already follow this pattern.

```suggestion
class OpenAIImageEditMockTest extends CamelTestSupport {
```

components/camel-ai/camel-openai/src/test/java/org/apache/camel/component/openai/OpenAIImageGenerationMockTest.java
```diff
@@ -0,0 +1,215 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OpenAIImageGenerationMockTest extends CamelTestSupport {
+
+    private static final byte[] FIRST_IMAGE = "FAKE-PNG-IMAGE-ONE".getBytes(StandardCharsets.UTF_8);
+    private static final byte[] SECOND_IMAGE = "FAKE-PNG-IMAGE-TWO".getBytes(StandardCharsets.UTF_8);
+
+    @RegisterExtension
+    public OpenAIMock openAIMock = new OpenAIMock().builder()
+            .whenImageGeneration("A red bicycle")
```
davsclaus, line 39: Same here — drop `public` to follow the JUnit 5 convention.

```suggestion
class OpenAIImageGenerationMockTest extends CamelTestSupport {
```

components/camel-ai/camel-openai/src/main/java/org/apache/camel/component/openai/OpenAIImageSupport.java
```diff
@@ -0,0 +1,286 @@
...
+import org.apache.camel.CamelExchangeException;
+import org.apache.camel.Exchange;
+import org.apache.camel.Message;
+import org.apache.camel.WrappedFile;
+import org.apache.camel.util.ObjectHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Shared helpers for the {@code image-generation} and {@code image-edit} operations.
```
davsclaus, line 41: Nit: `resolveParameter()` at the bottom of this class is byte-for-byte identical to `OpenAIProducer.resolveParameter()`. Consider extracting it to a shared location to avoid the duplication. Not blocking — just a follow-up suggestion.

davsclaus (commented):
Nice work on this PR — the documentation, test coverage, and alignment with existing component patterns are all solid. Two minor suggestions below.

Note: this review covers project conventions and contribution expectations. It does not replace specialized review tools (CodeRabbit, SonarCloud, etc.).

oscerd (approved):

davsclaus (approved):

PR #25490, opened 2026-08-13
CAMEL-24390: camel-mcp-server - support wildcard patterns in tag matching

davsclaus:
## Summary

- Add `PatternHelper.matchSimplePattern()` / `matchSimplePatterns()` — a simple pattern matcher that supports exact match (case-insensitive), wildcard prefix (`foo*`), and `*` to match all, **without** the regex fallback of `matchPattern()`
- Use `matchSimplePatterns()` for MCP tag matching instead of exact `Set.contains()`, enabling wildcard patterns while keeping behavior user-friendly and predictable
- Setting `camel.server.mcp-tags=*` now exposes all tagged tools without needing to list every tag; untagged default-pool tools remain excluded (security invariant preserved)
- Updated Javadoc on `McpServerConfiguration`, `HttpServerConfigurationProperties`, and `mcp-server.adoc` to document case-insensitive matching

## Test plan

- [x] Existing `McpServerBridgeTest` (11 tests) — all pass, no regressions
- [x] New `McpServerBridgeWildcardTagTest` — verifies `*` exposes all tagged tools but not untagged ones
- [x] New `McpServerBridgePrefixTagTest` — verifies `crm*,notify` matches `crm`, `crm-sales`, and `notify` but not `billing` or untagged tools
- [x] New `PatternHelperTest` — verifies simple pattern matching: exact (case-insensitive), wildcard prefix, wildcard all, null safety, and explicitly verifies regex fallback does NOT apply


components/camel-ai/camel-mcp-server-api/src/main/java/org/apache/camel/component/mcp/server/McpServerBridge.java
```diff
@@ -305,19 +303,22 @@ private McpToolCallResult execute(AiToolSpec spec, Map<String, Object> arguments
         }
     }
 
+    private boolean matchesTag(String tag) {
+        return tag != null && PatternHelper.matchPatterns(tag, tagPatterns);
```
orpiske: **Observation (non-blocking):** `PatternHelper.matchPattern()` has a 4-step cascade: exact (case-insensitive) → `*` → wildcard prefix (`foo*`) → **regex fallback**. This means a tag pattern like `a.b` would also match `axb` via regex, which may surprise operators.

Since tags are operator-controlled (not untrusted input), the practical risk is low. But it might be worth either:
- (a) Documenting the full matching cascade in the `setTags()` Javadoc so operators know regex is available, or
- (b) Using a restricted matcher that skips the regex step, if regex support isn't intended for tags.

Also worth noting: the original `Set.contains()` was case-sensitive, while `PatternHelper.matchPattern()` uses `equalsIgnoreCase()`. A tool tagged `CRM` now matches a configured tag `crm`. This is arguably more user-friendly, but it's a behavioral change worth documenting (e.g., "Tag matching is case-insensitive").

components/camel-ai/camel-mcp-server-api/src/main/java/org/apache/camel/component/mcp/server/McpServerConfiguration.java
```diff
@@ -35,8 +35,9 @@ public class McpServerConfiguration {
     private long sessionIdleTtl = McpServerConstants.DEFAULT_SESSION_IDLE_TTL;
 
     /**
-     * Comma-separated list of ai-tool tags to expose as MCP tools. Only tools registered under one of these tags are
-     * published; the untagged default pool is never exposed. When not set, no tools are published.
+     * Comma-separated list of ai-tool tag patterns to expose as MCP tools. Patterns support exact match, wildcard
+     * prefix ({@code foo*}), and {@code *} to match all tags. Only tools registered under a matching tag are published;
+     * the untagged default pool is never exposed. When not set, no tools are published.
```
orpiske: **Suggestion (non-blocking):** Consider mentioning that matching is case-insensitive and that the regex fallback is also available. This would help operators understand the full behavior of the tag patterns. Something like:

> Patterns support exact match (case-insensitive), wildcard prefix (`foo*`), `*` to match all tags, and regular expressions.

orpiske (commented):
Nice, clean enhancement — the core change is minimal and the security invariant (untagged tools never exposed) is robustly preserved with defense-in-depth (`tag != null` guard in `matchesTag()` plus `PatternHelper.matchPattern(null, ...)` returning `false` independently). Tests are well-structured and follow all project conventions (AssertJ, package-private, no `Thread.sleep()`).

Two non-blocking observations below as inline comments.

_This review was generated by an AI agent (Claude Code on behalf of orpiske) and may contain inaccuracies. Please verify all suggestions before applying._

components/camel-ai/camel-mcp-server-api/src/main/java/org/apache/camel/component/mcp/server/McpServerBridge.java
davsclaus: Good catch — added `PatternHelper.matchSimplePattern()` / `matchSimplePatterns()` that skip the regex fallback entirely. Tag matching now only supports exact match (case-insensitive), wildcard prefix (`foo*`), and `*` for match-all. This keeps things user-friendly and avoids surprising regex behavior like `a.b` matching `axb`.

Also added unit tests for the new methods in `PatternHelperTest`, including a test that explicitly verifies the regex fallback does NOT apply.

components/camel-ai/camel-mcp-server-api/src/main/java/org/apache/camel/component/mcp/server/McpServerConfiguration.java
davsclaus: Updated the Javadoc on `McpServerConfiguration.getTags()` and `HttpServerConfigurationProperties` to document that matching is case-insensitive. Since we dropped the regex fallback (see other comment), the docs now accurately describe: exact match (case-insensitive), wildcard prefix (`foo*`), and `*` to match all.

davsclaus:
merged manually

PR #25492, opened 2026-08-13
CAMEL-24392: Fix BacklogTracerRouteAdvice draining WrappedFile streambody

gansheer:
CAMEL-24392

# Description

BacklogTracerRouteAdvice runs at route level before StreamCachingAdvice (channel level), so tracing drained stream bodies (e.g. SFTP streamDownload) before they could be cached.
    
Cache the stream in BacklogTracerRouteAdvice.before() when stream caching is enabled on the route, ensuring the body is converted to a re-readable StreamCache before the tracer reads it.

Test added camel-core to confirm the fix and avoid any new regression.
Backport will be done on 4.22 and 4.18

# Target

# Tracking

# AI-assisted contributions

### DIFF
--- a/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
+++ b/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
@@ -200,6 +200,9 @@ public class CamelInternalProcessor extends DelegateAsyncProcessor implements In
     @Override
     public void addRouteLifecycleAdvice(CamelContext camelContext, Route route, NamedRoute node) {
         addAdvice(new CamelInternalProcessor.RouteLifecycleAdvice());
+        if (route.isStreamCaching()) {
+            addAdvice(new StreamCachingAdvice(camelContext.getStreamCachingStrategy()));
+        }
         if (camelContext.isBacklogTracingStandby() || route.isBacklogTracing()) {
             addAdvice(new BacklogTracerRouteAdvice(camelContext, node));
         }
--- a/core/camel-core/src/test/java/org/apache/camel/component/file/GenericFileStreamBodyWithBacklogTracerTest.java
+++ b/core/camel-core/src/test/java/org/apache/camel/component/file/GenericFileStreamBodyWithBacklogTracerTest.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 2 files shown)


core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
```diff
@@ -644,6 +644,11 @@ public void setRoute(Route route) {
         @Override
         public DefaultBacklogTracerEventMessage before(Exchange exchange) throws Exception {
             if (!skip && backlogTracer.shouldTrace(routeDefinition.getInput(), exchange)) {
+                if (route != null && Boolean.TRUE.equals(route.isStreamCaching())) {
+                    StreamCachingHelper.convertToStreamCache(
+                            camelContext.getStreamCachingStrategy(), exchange, exchange.getIn());
+                }
```
davsclaus: This embeds stream caching logic inside the tracer advice, coupling two unrelated concerns.

At the channel level, `StreamCachingAdvice` already solves this correctly via `Ordered.HIGHEST` — it sorts before tracing/debugging advices automatically. The route level should follow the same pattern: add `StreamCachingAdvice` in `addRouteLifecycleAdvice()` when stream caching is enabled on the route, and let the `Ordered` sorting place it before `BacklogTracerRouteAdvice`.

This way, stream caching runs first at both route and channel level, consistently.

davsclaus (changes requested):
Thank you for the contribution and for identifying the root cause correctly — the `BacklogTracerRouteAdvice` runs at route level before `StreamCachingAdvice` (channel level), draining stream bodies before they can be cached.

However, the fix approach needs rework. Embedding `StreamCachingHelper.convertToStreamCache()` directly inside `BacklogTracerRouteAdvice.before()` couples two unrelated concerns and is a workaround rather than a proper fix.

At the **channel level**, this is already solved correctly: `StreamCachingAdvice` implements `Ordered` with `Ordered.HIGHEST` (`Integer.MIN_VALUE`) and the comment says _"we want stream caching first"_. The `addAdvice()` method sorts advices by order, so stream caching runs before tracing at the channel level.

The proper fix is to follow the same pattern at the **route level**: add `StreamCachingAdvice` in `addRouteLifecycleAdvice()` (when stream caching is enabled) and let the `Ordered` sorting naturally place it before `BacklogTracerRouteAdvice`. This is consistent with the existing architecture and avoids coupling stream caching logic into the tracer.

If another route-level advice needs to read the body in the future, the current approach would require the same workaround in every advice — whereas adding `StreamCachingAdvice` at the route level solves it once for all.

Additionally, the PR includes an unrelated regenerated change to the `mcpTags` description in the catalog metadata JSON — please keep the PR scoped to the fix.

core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/CamelInternalProcessor.java
gansheer: Good point — moved stream caching out of the tracer advice. StreamCachingAdvice is now added at the route level in addRouteLifecycleAdvice() when stream caching is enabled, mirroring what DefaultChannel already does at the channel level. Ordered.HIGHEST ensures it sorts before BacklogTracerRouteAdvice.

_This comment was generated by an AI agent (Claude Code on behalf of gansheer)_

davsclaus (approved):
The updated fix is architecturally sound — `StreamCachingAdvice` is now added at the route level in `addRouteLifecycleAdvice()`, consistent with how `DefaultChannel` already handles this at the channel level. `Ordered.HIGHEST` ensures stream caching runs before `BacklogTracerRouteAdvice`, solving the stream-draining issue cleanly without coupling unrelated concerns.

The test covers the exact scenario (standby tracer + messageHistory + non-resettable stream body) and follows project conventions.

This review covers OSS Helper scope (rules, conventions, architecture). It does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud.

PR #25495, opened 2026-08-13
CAMEL-24377: Add JSON support to CamelEvent

atiaomar1978-hub:
## Description

Implements [CAMEL-24377](https://issues.apache.org/jira/browse/CAMEL-24377): structured JSON serialization for `CamelEvent`, similar to the existing `BacklogEventMessage` API used by Error Registry and Backlog Tracer.

### API

Added to `org.apache.camel.spi.CamelEvent`:

- `Map<String, Object> asJSon()` — structured event metadata as a JSON-compatible map
- `String toJSon(int indent)` — pretty-printed JSON string

### Implementation

- `CamelEventJsonSupport` in `camel-base` builds JSON for all standard event types (context, route, exchange, step, service, failure)
- Abstract event base classes and service failure events delegate to the support class
- `EventConsole` now uses `event.asJSon()` instead of ad hoc `toString()` fields

### JSON fields (examples)

| Event category | Fields |
|----------------|--------|
| All events | `type`, `eventClass`, `timestamp`, `message` |
| Context | `contextName` |
| Route | `routeId`, `routeGroup`, `fromEndpointUri` |
| Exchange | `exchangeId`, `fromRouteId`, `routeId` |
| Sent/Sending | `endpointUri`, `timeTaken` (sent) |
| Failure | `exception` (type, message, stackTrace) |
| Redelivery | `attempt`, `exception` |
| Step | `stepId` |
| Service failure | `service`, `contextName` |

### Tests

`CamelEventJsonTest` covers context, exchange failed/sent/failure-handling/redelivery, route reloaded/restarting failure, step failed, and service startup failure events.

```bash
./mvnw -pl core/camel-core -am test -Dtest=CamelEventJsonTest -Dsurefire.failIfNoSpecifiedTests=false
```

### Documentation

Updated `docs/user-manual/modules/ROOT/pages/event-notifier.adoc` with JSON serialization section.

## Tracking

- [x] [CAMEL-24377](https://issues.apache.org/jira/browse/CAMEL-24377)

## AI-assisted contributions

_AI-generated PR on behalf of atiaomar1978-hub (Cursor Agent)_

### DIFF
--- a/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
@@ -119,7 +119,7 @@ public interface CamelEvent {
             if (indent > 0) {
                 sb.append('\n').append(indentText);
             }
-            sb.append(StringQuoteHelper.doubleQuote(entry.getKey())).append(':');
+            sb.append(StringQuoteHelper.jsonQuote(entry.getKey())).append(':');
             if (indent > 0) {
                 sb.append(' ');
             }
@@ -127,7 +127,7 @@ public interface CamelEvent {
             if (value instanceof Number || value instanceof Boolean) {
                 sb.append(value);
             } else {
-                sb.append(StringQuoteHelper.doubleQuote(String.valueOf(value)));
+                sb.append(StringQuoteHelper.jsonQuote(String.valueOf(value)));
             }
         }
         if (indent > 0) {
--- a/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.spi;
 
+import java.util.Map;
+
 import org.apache.camel.CamelContext;
 import org.apache.camel.Endpoint;
 import org.apache.camel.Exchange;
@@ -93,6 +95,22 @@ public interface CamelEvent {
     long getTimestamp();
 
     void setTimestamp(long timestamp);
+
+    /**
+     * Dumps the full event as a pretty-printed JSON string.
+     *
+     * @param  indent number of spaces to indent
+     * @return        JSON representation of this event
+     * @since         4.23
+     */
+    String toJSon(int indent);
+
+    /**
+     * The full event as a {@link Map} suitable for JSON serialization, containing structured metadata for this event.
+     *
+     * @since 4.23
+     */
+    Map<String, Object> asJSon();
 
     /**
      * This interface is implemented by all events that contain an exception and is used to retrieve the exception in a
--- a/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
@@ -16,6 +16,7 @@
  */
 package org.apache.camel.spi;
 
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.apache.camel.CamelContext;
@@ -23,6 +24,7 @@ import org.apache.camel.Endpoint;
 import org.apache.camel.Exchange;
 import org.apache.camel.Processor;
 import org.apache.camel.Route;
+import org.apache.camel.util.StringQuoteHelper;
 import org.jspecify.annotations.Nullable;
 
 /**
@@ -103,14 +105,55 @@ public interface CamelEvent {
      * @return        JSON representation of this event
      * @since         4.23
      */
-    String toJSon(int indent);
+    default String toJSon(int indent) {
+        Map<String, Object> map = asJSon();
+        String indentText = indent > 0 ? " ".repeat(indent) : "";
+        StringBuilder sb = new StringBuilder(128);
+        sb.append('{');
+        boolean first = true;
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            if (!first) {
+                sb.append(',');
+            }
+            first = false;
+            if (indent > 0) {
+                sb.append('\n').append(indentText);
+            }
+            sb.append(StringQuoteHelper.doubleQuote(entry.getKey())).append(':');
+            if (indent > 0) {
+                sb.append(' ');
+            }
+            Object value = entry.getValue();
+            if (value instanceof Number || value instanceof Boolean) {
+                sb.append(value);
+            } else {
+                sb.append(StringQuoteHelper.doubleQuote(String.valueOf(value)));
+            }
+        }
+        if (indent > 0) {
+            sb.append('\n');
+        }
+        sb.append('}');
+        return sb.toString();
+    }
 
     /**
      * The full event as a {@link Map} suitable for JSON serialization, containing structured metadata for this event.
      *
+     * The default implementation returns a minimal map with {@code type}, optional {@code timestamp}, and
+     * {@code message}. Camel's built-in event classes override this to provide richer structured metadata.
+     *
      * @since 4.23
      */
-    Map<String, Object> asJSon();
+    default Map<String, Object> asJSon() {
+        Map<String, Object> map = new LinkedHashMap<>();
+        map.put("type", getType().name());
+        if (getTimestamp() > 0) {
+            map.put("timestamp", getTimestamp());
+        }
+        map.put("message", toString());
+        return map;
+    }
 
     /**
      * This interface is implemented by all events that contain an exception and is used to retrieve the exception in a
--- a/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
+++ b/core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
@@ -101,6 +101,10 @@ public interface CamelEvent {
     /**
      * Dumps the full event as a pretty-printed JSON string.
      *
+     * The default implementation serializes flat scalar values ({@link Number}, {@link Boolean}, and {@link String})
+     * from {@link #asJSon()}. If a custom implementer overrides {@link #asJSon()} to include nested {@link Map} or
+     * {@link List} objects, they should also override this method to produce valid JSON.
+     *
      * @param  indent number of spaces to indent
      * @return        JSON representation of this event
      * @since         4.23
--- a/core/camel-base/src/main/java/org/apache/camel/impl/event/CamelEventJsonSupport.java
+++ b/core/camel-base/src/main/java/org/apache/camel/impl/event/CamelEventJsonSupport.java
@@ -1,0 +1,193 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.impl.event;
+
+import java.util.Map;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.Endpoint;
+import org.apache.camel.Exchange;
+import org.apache.camel.ExchangePropertyKey;
+import org.apache.camel.Processor;
+import org.apache.camel.Route;
+import org.apache.camel.spi.CamelEvent;
+import org.apache.camel.support.MessageHelper;
+import org.apache.camel.util.ObjectHelper;
+import org.apache.camel.util.URISupport;
+import org.apache.camel.util.json.JsonObject;
+import org.apache.camel.util.json.Jsoner;
+
+/**
+ * Support class for serializing {@link CamelEvent} instances to JSON.
+ */
+public final class CamelEventJsonSupport {
+
+    private CamelEventJsonSupport() {
+    }
+
+    public static Map<String, Object> asJSon(CamelEvent event) {
+        JsonObject jo = new JsonObject();
+        jo.put("type", event.getType().name());
+        jo.put("eventClass", event.getClass().getSimpleName());
+        if (event.getTimestamp() > 0) {
+            jo.put("timestamp", event.getTimestamp());
+        }
+        jo.put("message", event.toString());
+
+        if (event instanceof CamelEvent.CamelContextEvent contextEvent) {
+            appendContext(jo, contextEvent.getContext());
+        }
+        if (event instanceof CamelEvent.RouteEvent routeEvent) {
+            appendRoute(jo, routeEvent.getRoute());
+        }
+        if (event instanceof CamelEvent.ExchangeEvent exchangeEvent) {
+            appendExchange(jo, exchangeEvent.getExchange());
+        }
+        if (event instanceof CamelEvent.StepEvent stepEvent) {
+            jo.put("stepId", stepEvent.getStepId());
+        }
+        if (event instanceof CamelEvent.ExchangeSendingEvent sendingEvent) {
+            appendEndpoint(jo, "endpointUri", sendingEvent.getEndpoint());
+        }
+        if (event instanceof CamelEvent.ExchangeSentEvent sentEvent) {
+            appendEndpoint(jo, "endpointUri", sentEvent.getEndpoint());
+            jo.put("timeTaken", sentEvent.getTimeTaken());
+        }
+        if (event instanceof CamelEvent.ExchangeRedeliveryEvent redeliveryEvent) {
+            jo.put("attempt", redeliveryEvent.getAttempt());
+        }
+        if (event instanceof CamelEvent.ExchangeFailureEvent failureEvent) {
+            appendFailureHandling(jo, failureEvent);
+        }
+        if (event instanceof CamelEvent.RouteReloadedEvent reloadedEvent) {
+            jo.put("index", reloadedEvent.getIndex());
+            jo.put("total", reloadedEvent.getTotal());
+        }
+        if (event instanceof CamelEvent.RouteRestartingEvent restartingEvent) {
+            jo.put("attempt", restartingEvent.getAttempt());
+        }
+        if (event instanceof CamelEvent.RouteRestartingFailureEvent restartingFailureEvent) {
+            jo.put("attempt", restartingFailureEvent.getAttempt());
+            jo.put("exhausted", restartingFailureEvent.isExhausted());
+        }
+        if (event instanceof CamelEvent.ServiceEvent serviceEvent) {
+            appendService(jo, serviceEvent.getService());
+        }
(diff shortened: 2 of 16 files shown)


atiaomar1978-hub:
## Bugbot / Grok self-review — findings addressed

_AI-generated comment on behalf of atiaomar1978-hub (Cursor Agent)_

### Review summary

| Severity | Finding | Status |
|----------|---------|--------|
| Medium | `ExchangeRedeliveryEvent` omitted exception in JSON | **Fixed** in `c916abe` — includes exchange exception |
| Low | `appendException` lacked defensive try/catch | **Fixed** in `c916abe` — fallback to `exceptionMessage` |
| Medium | Empty `docs/yarn.lock` in repo | **Not in this PR** — file is untracked locally, not committed |
| High | Unrelated `scripts/post-pr-25433-inline-replies.sh` | **Not in this PR** — untracked locally, not committed |

### Test coverage (`CamelEventJsonTest`)

- Context started event (`asJSon` + `toJSon`)
- Exchange failed (via EventNotifier, exception fields)
- Exchange sent (endpointUri, timeTaken)
- Exchange failure handling (DLQ metadata)
- Exchange redelivery (attempt + exception)
- Route reloaded / restarting failure
- Step failed
- Service startup failure
- JsonObject compatibility

### Tests

```bash
./mvnw -pl core/camel-core -am test -Dtest=CamelEventJsonTest -Dsurefire.failIfNoSpecifiedTests=false
```

All 10 tests passing locally. Ready for human review.

atiaomar1978-hub:
## CI fix — unresolved documentation xref

Checked [CI run 31732539144](https://github.com/apache/camel/actions/runs/31732539144): **failure was from this PR**.

### Root cause
`docs` module **xref-check** failed:
```
Unresolved xrefs:
  manual:ROOT:event-notifier.adoc line 154:
  xref:components::others/backlog-tracer.adoc[Backlog Tracer]
```
`backlog-tracer.adoc` lives in the user manual (`manual::`), not under `components::others`.

### Fix
**Commit:** `d28e2f97633` — changed to `xref:backlog-tracer.adoc[Backlog Tracer]` (same pattern as `tracer.adoc`, `index.adoc`, etc.)

### Verification (local)
```bash
./mvnw -pl docs -Psourcecheck generate-resources -DskipTests   # xref-check: green
./mvnw -pl core/camel-core -am -Dtest=CamelEventJsonTest test  # green
```

Pushed to `feature/CAMEL-24377-camel-event-json` — CI re-run should pass the docs xref-check step.

---
_Composer on behalf of atiaomar1978-hub (AI-generated)_

davsclaus:
@atiaomar1978-hub can you continue this work

atiaomar1978-hub:
Let me rebase @davsclaus

core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
```diff
@@ -94,6 +96,22 @@ enum Type {
...
+     * @since         4.23
+     */
+    String toJSon(int indent);
+
+    /**
+     * The full event as a {@link Map} suitable for JSON serialization, containing structured metadata for this event.
+     *
+     * @since 4.23
+     */
+    Map<String, Object> asJSon();
```
davsclaus: `CamelEvent` is a published SPI, and these two methods are added as **non-`default`** abstract methods. That's a source- and binary-incompatible change for any code outside Camel that implements `CamelEvent` (or one of its sub-interfaces) directly — it will no longer compile/link. `project-standards.md` requires maintaining backwards compatibility for public APIs.

All in-tree events extend the modified abstract base classes, so the internal build is fine — the risk is entirely external.

Suggest making both methods `default` so existing implementers keep working. Since `CamelEventJsonSupport` lives in `camel-base` (not reachable from `camel-api`), a reasonable default is a minimal map here (e.g. `type`/`timestamp`/`message`), with the rich implementation remaining the override in `AbstractContextEvent`/`AbstractExchangeEvent`/`AbstractRouteEvent`/etc. If a hard SPI break is genuinely intended, it needs explicit justification and an upgrade-guide entry.

davsclaus (changes requested):
Thanks for this contribution, Omar — it's well-scoped and nicely tested, and I want to call out one thing you got right: the exception/stacktrace serialization reuses `MessageHelper.dumpExceptionAsJSonObject`, the same helper the developer consoles use (backlog tracer, error registry). That JSON-escapes the trace and exception message so embedded quotes, newlines and JSON survive round-tripping. That's the correct "special care" for stacktraces — good call not rolling your own.

A few things to address before this can go in:

**🔴 Blocking**

1. **Generated dev-console descriptor not regenerated.** Adding the `details` field to the `EventEntry` record changes the `event` console's response schema, but `core/camel-console/src/generated/resources/META-INF/org/apache/camel/dev-console/event.json` still lists only `type`/`timestamp`/`exchangeId`/`message`. CI's uncommitted-generated-files check will fail. Please regenerate and commit the descriptor. (The `EventConsoleConfigurer` only covers the `capacity` property, so it's unaffected — descriptor only.)

**🟠 Major**

2. **New non-`default` methods on the public SPI `CamelEvent`** — see inline comment. Adding abstract methods to a published SPI is a source/binary break for external implementers; the project requires maintaining public-API backwards compatibility.

**🟡 Minor**

3. **Upgrade-guide entry.** Per the contributor guidelines, API/SPI signature changes should be noted in `docs/user-manual/modules/ROOT/pages/camel-4x-upgrade-guide-4_23.adoc`. The `event-notifier.adoc` addition documents the feature nicely, but the SPI change also warrants a short migration note (especially if the methods stay non-default).
4. **`details` duplicates the flat fields** (`type`/`timestamp`/`exchangeId`/`message`) that already exist on `EventEntry`. Harmless, just flagging it as a conscious design choice.

Note: this is a rules-and-conventions review and does not replace specialized tools such as CodeRabbit/Sourcery or SonarCloud static analysis.

atiaomar1978-hub:
## Review feedback addressed

_Composer on behalf of atiaomar1978-hub (AI-generated)_

Thanks @davsclaus — all review items are now addressed in **6c7b7bc**:

| Priority | Item | Status |
|----------|------|--------|
| Blocking | Regenerate dev-console descriptor for `details` field | **Done** in `ad1b12d` |
| Major | Make `asJSon()` / `toJSon()` **default methods** on `CamelEvent` SPI | **Done** in `6c7b7bc` — minimal `type`/`timestamp`/`message` fallback; built-in events still override with rich JSON via `CamelEventJsonSupport` |
| Minor | Upgrade guide entry | **Done** in `6c7b7bc` — `camel-4x-upgrade-guide-4_23.adoc` |
| Minor | `details` duplicates flat fields | **Intentional** — flat fields kept for backwards compatibility; `details` holds the full structured JSON payload |

### CI / regen check

Re-ran locally after the latest changes:

```bash
mvn install -pl core/camel-console,catalog/camel-catalog -am -DskipTests
mvn test -pl core/camel-core -am -Dtest=CamelEventJsonTest -Dsurefire.failIfNoSpecifiedTests=false
```

No uncommitted generated files — dev-console metadata from the earlier regen commit is still current.

Ready for re-review.

core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
atiaomar1978-hub: Fixed in **6c7b7bc** — both methods are now `default` with a minimal `type`/`timestamp`/`message` fallback using `LinkedHashMap` and `StringQuoteHelper` (no extra dependency on `camel-util-json` from `camel-api`). Built-in events continue to override via `CamelEventJsonSupport` for the rich structured JSON.

Also added the upgrade guide entry in `camel-4x-upgrade-guide-4_23.adoc`.

_Composer on behalf of atiaomar1978-hub (AI-generated)_

core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
```diff
@@ -94,6 +98,63 @@ enum Type {
...
+            }
+            sb.append(StringQuoteHelper.doubleQuote(entry.getKey())).append(':');
+            if (indent > 0) {
+                sb.append(' ');
+            }
+            Object value = entry.getValue();
+            if (value instanceof Number || value instanceof Boolean) {
+                sb.append(value);
+            } else {
+                sb.append(StringQuoteHelper.doubleQuote(String.valueOf(value)));
```
davsclaus: `StringQuoteHelper.doubleQuote(...)` only wraps the value in `"` quotes — it does **not** JSON-escape `"`, `\`, or control characters. Because the default `asJSon()` sets `message` to `toString()` (which commonly contains endpoint URIs and exception text), this default `toJSon` produces invalid JSON whenever a value contains one of those characters.

Example: a custom `CamelEvent` whose `toString()` is `He said "hi"` yields `{"message":"He said "hi""}` — broken JSON. The same applies to the key on line 122.

Built-in events are unaffected (they override and delegate to `CamelEventJsonSupport` → `Jsoner`, which escapes), so this is confined to the default SPI path used by external implementers. Since `camel-api` doesn't depend on `camel-util-json`, a small manual escape of `"`, `\`, and control chars here (for both the key and the string value) would fix it without adding a dependency.

davsclaus (changes requested):
Thanks for iterating on this, Omar — the previous round of feedback is nicely addressed: the SPI methods are now `default` (so external implementers stay source/binary compatible), the `event` dev-console descriptor is regenerated, the upgrade-guide entry is in place, and reusing `MessageHelper.dumpExceptionAsJSonObject` for exception/stacktrace serialization is exactly right. The `ExchangeFailedEvent`/`ExchangeFailureHandlingEvent` split is also handled correctly. Nice work.

One blocking correctness issue remains, plus a couple of minor points.

### 🟠 Blocking

**Default `CamelEvent.toJSon(int)` can emit invalid JSON** (see inline comment). `StringQuoteHelper.doubleQuote(...)` only wraps a string in quotes; it does not JSON-escape. Since the default `asJSon()` puts `message = toString()` (which routinely contains endpoint URIs and exception text, i.e. characters like `"`, `\`, newlines), the default `toJSon` produces malformed JSON for those values.

Blast radius is the **default/fallback path only** — all built-in events override and delegate to `CamelEventJsonSupport` → `Jsoner`, which escapes correctly, so the dev console and shipped events are fine. The issue affects external `CamelEvent` implementers that rely on the default. Worth noting: `camel-api` depends on `camel-util` but not `camel-util-json`, and `camel-util` has no JSON escaper (which is presumably why this was hand-rolled), so the fix likely needs a minimal manual escape of `"`, `\`, and control characters in the default `toJSon` rather than pulling in `Jsoner`.

### 🟡 Minor (non-blocking)

1. **The default (non-overridden) SPI path is untested.** `testContextStartedEventAsJson` calls `toJSon(2)` on a built-in event, which takes the overridden `Jsoner` path — so nothing exercises the default `toJSon`/`asJSon`. A test with a bare custom `CamelEvent` (ideally with a `"` in its `toString()`) would have caught the issue above and guards it going forward.
2. **Cosmetic path divergence.** The default `asJSon()` omits `eventClass` (the built-in support class adds it) and the default `toJSon` formatting differs from `Jsoner.prettyPrint`, so the two code paths yield slightly different JSON. Harmless, just flagging.

Also note CI was still pending when I reviewed — please make sure it goes green (it will catch any formatting / generated-file drift).

This is a rules-and-conventions and diff review; it does not replace specialized tools such as CodeRabbit/Sourcery or SonarCloud static analysis.

core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
gnodet: `StringQuoteHelper.doubleQuote(String.valueOf(value))` does not JSON-escape — it's literally `'"' + text + '"'`. Any value containing `"`, `\`, or control chars produces invalid JSON.

Minimal fix — add a proper JSON escape to `StringQuoteHelper` (or inline here):

```suggestion
                sb.append(jsonQuote(String.valueOf(value)));
```

With a helper like:
```java
private static String jsonQuote(String text) {
    StringBuilder sb = new StringBuilder(text.length() + 2);
    sb.append('"');
    for (int i = 0; i < text.length(); i++) {
        char c = text.charAt(i);
        switch (c) {
            case '"': sb.append("\\\"")); break;
            case '\\': sb.append("\\\\"); break;
            case '\n': sb.append("\\n"); break;
            case '\r': sb.append("\\r"); break;
            case '\t': sb.append("\\t"); break;
            default:
                if (c < 0x20) sb.append(String.format("\\u%04x", (int) c));
                else sb.append(c);
        }
    }
    sb.append('"');
    return sb.toString();
}
```

The same escaping must also be applied to the key on line 118 (`StringQuoteHelper.doubleQuote(entry.getKey())`).

gnodet (commented):
The `StringQuoteHelper.doubleQuote()` JSON-escaping issue flagged in the previous review remains unaddressed in the current head (`6c7b7bc`).

`StringQuoteHelper.doubleQuote(text)` is `"\"" + text + "\""` — it wraps in quotes but does **not** escape `\`, `"`, or control characters inside `text`. The default `CamelEvent.toJSon(int)` method therefore emits **invalid JSON** whenever a value contains any of those characters. This affects external `CamelEvent` implementers relying on the default path — built-in events are safe because they override and route through `CamelEventJsonSupport` → `Jsoner`.

Concrete scenario: a custom event whose `toString()` returns `Route "my-route" failed` produces `{"message":"Route "my-route" failed"}` — broken JSON.

Since `camel-api` cannot depend on `camel-util-json` (`Jsoner`), the fix needs a minimal manual escape in the default `toJSon` — at minimum escaping `\`, `"`, `\n`, `\r`, `\t` before wrapping in quotes. Alternatively, add a `jsonEscape` method to `StringQuoteHelper` in `camel-util`.

Additionally, the default (non-overridden) SPI path is **untested** — all tests use built-in events that take the `CamelEventJsonSupport` → `Jsoner` path. A test with a bare custom `CamelEvent` (especially one with special characters in `toString()`) would catch this and guard against future regressions.

_This review was generated by an AI agent._

_Hermès on behalf of @gnodet_

core/camel-api/src/main/java/org/apache/camel/spi/CamelEvent.java
atiaomar1978-hub: Fixed in `c32249aeb05` — added `StringQuoteHelper.jsonQuote()` in `camel-util` with proper JSON escaping and switched the default `CamelEvent.toJSon()` to use it for keys and string values. Added `CamelEventJsonTest.testDefaultCamelEventToJsonEscapesSpecialCharacters` covering the default SPI path with `\"` and newlines in `toString()`.

_Cursor Agent on behalf of atiaomar1978-hub_

atiaomar1978-hub: Fixed in `c32249aeb05` — implemented `StringQuoteHelper.jsonQuote()` in `camel-util` as suggested, and updated the default `CamelEvent.toJSon()` to use it for both keys and string values.

_Cursor Agent on behalf of atiaomar1978-hub_

atiaomar1978-hub (commented):
Thanks @davsclaus @gnodet — the JSON-escaping issue is fixed in `c32249aeb05`. Default SPI path now uses `StringQuoteHelper.jsonQuote()`; built-in events still use `CamelEventJsonSupport` → `Jsoner`. Tests added for the default path with special characters.

_Cursor Agent on behalf of atiaomar1978-hub_

atiaomar1978-hub:
Addressed the blocking JSON-escaping review in `c32249aeb05`:

- Added `StringQuoteHelper.jsonQuote()` in `camel-util` with proper escaping of `\"`, `\\`, control characters
- Default `CamelEvent.toJSon()` now uses `jsonQuote` for keys and string values (replacing `doubleQuote`)
- Added `CamelEventJsonTest.testDefaultCamelEventToJsonEscapesSpecialCharacters` for the default SPI path

Built-in events unchanged (still delegate to `CamelEventJsonSupport` → `Jsoner`).

_Cursor Agent on behalf of atiaomar1978-hub_

gnodet (commented):
The `StringQuoteHelper.jsonQuote()` fix in `c32249aeb05` correctly addresses the blocking JSON-escaping issue from the previous review rounds. The implementation properly escapes `"`, `\`, `\n`, `\r`, `\t`, and control characters `< 0x20`. The default `CamelEvent.toJSon()` now uses `jsonQuote` for both keys and string values.

The new `testDefaultCamelEventToJsonEscapesSpecialCharacters` test directly exercises the default SPI path with special characters — exactly what was requested. `StringQuoteHelperTest` covers the escaping helper itself. Good.

**One minor observation (non-blocking):**

The default `toJSon(int)` serializes values as flat `Number`/`Boolean`/`String` only — no support for nested `Map` or `List` objects. This is fine for the default `asJSon()` (which only returns `type`/`timestamp`/`message`), but if a custom implementer overrides `asJSon()` to include nested structures without also overriding `toJSon()`, they'll get `{key=value}` Java Map.toString() output instead of valid JSON. Consider documenting this constraint in the `toJSon` Javadoc ("this default handles only flat scalar values; override if your `asJSon()` returns nested objects"). Not blocking.

The ast-grep `mutable-collection-return` finding on `CamelEventJsonSupport.asJSon()` (returns a mutable `JsonObject`) is by design — the `EventConsole` and callers need to read from this map, and `JsonObject` is the expected Camel internal type.

The ast-grep `broad-exception-catch` findings at `CamelEventJsonSupport` lines 183 and 194 are intentional defensive code (graceful fallback when `MessageHelper.dumpExceptionAsJSonObject` or `URISupport.sanitizeUri` fail).

The previous blocking items (SPI compatibility via `default` methods, dev-console descriptor regeneration, upgrade guide entry) were all addressed in prior commits.

_This review was generated by an AI agent, Hermès on behalf of @gnodet._

atiaomar1978-hub:
## All review feedback addressed — ready for re-review

_Cursor Agent on behalf of atiaomar1978-hub (AI-generated)_

Thanks @davsclaus @gnodet — summary of fixes on the latest head (`a40734b`):

| Review item | Commit | Status |
|-------------|--------|--------|
| Make `asJSon()` / `toJSon()` **default methods** (SPI compatibility) | `6c7b7bc` | Done |
| Regenerate dev-console `event.json` descriptor for `details` field | `ad1b12d` | Done |
| Upgrade guide entry in `camel-4x-upgrade-guide-4_23.adoc` | `6c7b7bc` | Done |
| JSON-escape default `toJSon()` output (`StringQuoteHelper.jsonQuote()`) | `c32249aeb05` | Done |
| Test default SPI path with special characters in `toString()` | `c32249aeb05` | Done |
| Document scalar-only constraint for default `toJSon()` | `a40734b` | Done |

**CI:** green on Java 17/25 and doc validation (latest run on `c32249aeb05`).

All inline review threads are resolved. @davsclaus could you take another look when you have a moment?

core/camel-base/src/main/java/org/apache/camel/impl/event/CamelEventJsonSupport.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.camel.support.MessageHelper;
+import org.apache.camel.util.ObjectHelper;
+import org.apache.camel.util.URISupport;
+import org.apache.camel.util.json.JsonObject;
+import org.apache.camel.util.json.Jsoner;
+
+/**
+ * Support class for serializing {@link CamelEvent} instances to JSON.
+ */
+public final class CamelEventJsonSupport {
```
davsclaus, line 37: Maybe we should add a note in the base event class an implementation not that this class should be updated in case we add new events.

davsclaus (approved):

PR #25491, opened 2026-08-13
CAMEL-23861: Phase 2 - TUI AI Usage view integrates GenAI OTel spans

atiaomar1978-hub:
# Description

Phase 2 of [CAMEL-23861](https://issues.apache.org/jira/browse/CAMEL-23861) continues GenAI observability work after Phase 1 (#25337) introduced `camel-ai-observability` and LangChain4j/OpenAI producer span enrichment.

This change extends the Camel JBang TUI **AI Usage** panel (Ctrl+U) to combine:

- **TUI ask** token metrics (existing `AiPanel` usage history from PR #24362)
- **Route** GenAI usage extracted from OTel spans exported by the TUI monitor, using `gen_ai.*` semantic convention attributes from `camel-ai-observability`

## Changes

- **`GenAiSpanUsageExtractor`**: maps `SpanEntry` OTel span data into `AiPanel.AiUsageEntry` when `gen_ai.operation.name` is present
- **`AiPanel`**: adds `AiUsageSource` (TUI vs ROUTE), combines TUI history with route span data in stats view, and labels models as `[tui]` or `[route:<routeId>]`
- **`CamelMonitor`**: wires `dataService.otelSpans()` into `AiPanel`
- **Dependency**: `camel-jbang-plugin-tui` depends on `camel-ai-observability-api` for `GenAiAttributes`
- **Docs**: Phase 2 TUI integration documented in `ai-observability.adoc` and `camel-jbang-tui.adoc`
- **Tests**: 8 new unit tests covering span extraction, combined usage, stats rendering, and empty state

## Testing

```bash
./mvnw -pl dsl/camel-jbang/camel-jbang-plugin-tui -am test \
  -Dtest='GenAiSpanUsageExtractorTest,AiPanelGenAiUsageTest' \
  -Dsurefire.failIfNoSpecifiedTests=false
```

# Target

- [x] I checked that the commit is targeting the correct branch (Camel 4 uses the `main` branch)

# Tracking

- [x] JIRA issue filed: [CAMEL-23861](https://issues.apache.org/jira/browse/CAMEL-23861)

# Apache Camel coding standards and style

- [x] Each commit has a meaningful subject line and body
- [x] Auto-generated catalog doc changes committed

# AI-assisted contributions

- [x] Commits include `Co-authored-by: Composer <noreply@cursor.com>` trailer
- _Composer on behalf of Omar Atie (atiaomar1978-hub)_

PR #25507, opened 2026-08-14
CAMEL-23861: GenAI observability Phase 3 - Spring AI chat and camel-main config

atiaomar1978-hub:

## Summary

Phase 3 of [CAMEL-23861](https://issues.apache.org/jira/browse/CAMEL-23861) extends GenAI observability beyond Phase 1 (langchain4j/openai) and Phase 2 (TUI usage view):

- **Spring AI chat** (`spring-ai-chat`) producers now emit GenAI OpenTelemetry span attributes and record token usage via `GenAiObservability`, matching the langchain4j-chat pattern.
- **Camel Main typed configuration** for the global toggle `camel.ai.observability.enabled` (CAMEL-24386): application.properties, programmatic `main.configure().ai().observability().withEnabled(false)`, and metadata/docs.
- **GenAiModelResolver** extended with reflection-based Spring AI provider/model resolution (no hard dependency on Spring AI in the API module).

## Changes

| Area | Details |
|------|---------|
| `SpringAiChatProducer` | Wraps chat, structured output, and entity calls with `GenAiObservability` |
| `GenAiModelResolver` | Spring AI package + options/response metadata resolution |
| `camel-main` | `AiConfigurationProperties` / `AiObservabilityConfigurationProperties`, auto-config in `BaseMainSupport` |
| `PrepareCamelMainMojo` | Metadata prefix fix: `camel.ai.observability.enabled` (not `camel.main.enabled`) |
| Docs | `ai-observability.adoc`, upgrade guide 4.23 entry for Spring AI + Main config |
| Tests | `SpringAiChatObservabilityTest`, `AiObservabilityConfigurationPropertiesTest`, `GenAiModelResolverTest` (Spring AI) |

## Testing

```bash
./mvnw -pl components/camel-ai/camel-ai-observability-api,components/camel-ai/camel-ai-observability,components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat,core/camel-main -am test \
  -Dtest=GenAiModelResolverTest,SpringAiChatObservabilityTest,AiObservabilityConfigurationPropertiesTest,GenAiObservabilityTest
```

## Related

- Phase 1: #25337
- Phase 2: #25491
- JIRA: https://issues.apache.org/jira/browse/CAMEL-23861


atiaomar1978-hub:

## Bugbot follow-up (commit pushed)

Addressed both review findings:

1. **Entity path observability** — `processEntityRequest` no longer passes `requestModel` as `responseModel`; entity conversions record usage without fabricating a response model when `ChatResponse` is unavailable.

2. **ChatClient-only configuration** — `buildObservationContext` now resolves metadata from `chatClient` when `chatModel` is unset, and `GenAiModelResolver` uses optional reflection (`getChatModel`, `getDefaultChatOptions`) without failing when methods are absent.

Please re-review when CI is green.

atiaomar1978-hub:

Addressed Grok review findings in commits `0b38fcef` and `4ef2ca5f`:

**Critical**
- Regenerated catalog metadata — property is now `camel.ai.observability.enabled` (removed stale `camel.main.enabled`)
- `GenAiModelResolver` uses reflection-only LangChain4j detection — removed optional `langchain4j-core` from `camel-spring-ai-chat`
- ChatClient-only routes cache/extract underlying `ChatModel` at startup for observability metadata

**Medium**
- Entity path uses `responseEntity()` to capture token usage and response model in spans
- Added `PREFIX_AI_OBSERVABILITY` to Main `GROUP_PREFIXES`
- Direct `ChatResponse.getMetadata().getModel()` in span success recording
- Expanded tests: ChatClient-only, entity conversion, disabled observability, token span tags

All observability tests pass locally.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
```diff
@@ -150,6 +150,7 @@ public abstract class BaseMainSupport extends BaseService {
     private static final String PREFIX_TRACE = "camel.trace.";
     private static final String PREFIX_ROUTE_CONTROLLER = "camel.routeController.";
     private static final String PREFIX_ERROR_REGISTRY = "camel.errorRegistry.";
+    private static final String PREFIX_AI_OBSERVABILITY = "camel.ai.observability.";
```
davsclaus: The property prefix `camel.ai.observability.` uses two-level nesting, but every existing Camel Main configuration group uses a single-level camelCase prefix under `camel.`:

- `camel.health.`, `camel.opentelemetry2.`, `camel.errorRegistry.`, `camel.routeController.`, `camel.management.`, etc.

The canonical name should be `camel.aiObservability.enabled` (camelCase, matching e.g. `camel.routeController.`), with `camel.ai-observability.enabled` (dash-style) also accepted. The rename needs to ripple through: this constant, `GROUP_PREFIXES`, `PrepareCamelMainMojo`, `AiObservabilityConfigurationProperties`, metadata JSON, and docs.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
```diff
@@ -965,8 +989,6 @@ private <T> void processStructuredOutputRequest(
         // Also set headers
         exchange.getMessage().setHeader(SpringAiChatConstants.CHAT_RESPONSE, responseText);
         exchange.getMessage().setHeader(SpringAiChatConstants.STRUCTURED_OUTPUT, structuredOutput);
```
davsclaus, line 994: The old code called `populateTokenUsage(response, exchange)` after this line, but that call was removed. `callWithObservability()` records token data on the span but does NOT populate the exchange headers (`INPUT_TOKEN_COUNT`, `OUTPUT_TOKEN_COUNT`, `TOTAL_TOKEN_COUNT`).

Compare with:
- The entity path (`processEntityRequest`) which correctly calls both `recordObservationSuccess` AND `populateTokenUsage`
- The default chat path which calls `populateResponse()` (which includes `populateTokenUsage`)

Please restore `populateTokenUsage(response, exchange)` here.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
```diff
@@ -852,13 +866,24 @@ private Class<?> getEntityClass(Exchange exchange) {
...
-        T entity = request.call().entity(entityClass);
-
-        // Set the entity as the body
-        exchange.getMessage().setBody(entity);
-
-        LOG.debug("Converted response to entity of type: {}", entityClass.getName());
+        GenAiObservationContext observationContext = buildObservationContext();
+        GenAiObservation observation = GenAiObservability.start(exchange, observationContext);
+        try {
+            org.springframework.ai.chat.client.ResponseEntity<ChatResponse, T> responseEntity
```
davsclaus: Per project conventions: "Do NOT use fully qualified class names (FQCNs) in Java code. Always add an import statement and use the simple class name."

Since `org.springframework.http.ResponseEntity` is not imported in this file, a regular import for `org.springframework.ai.chat.client.ResponseEntity` would work without conflicts.

```suggestion
            ResponseEntity<ChatResponse, T> responseEntity
```

davsclaus (changes requested):
Thank you for this contribution — the observability wiring and reflection-based resolver are well-designed. I have a few findings that need to be addressed.

### Confirmed issues

1. **Wrong configuration property prefix** — The property uses `camel.ai.observability.enabled` (two-level nesting) but every existing Camel Main configuration group uses a single-level prefix under `camel.`: `camel.health.`, `camel.opentelemetry2.`, `camel.errorRegistry.`, `camel.routeController.`, etc. The correct name should be `camel.ai-observability.enabled` (hyphenated, single level). This affects `PREFIX_AI_OBSERVABILITY`, `GROUP_PREFIXES`, `PrepareCamelMainMojo`, the configuration properties classes, the metadata JSON, and the documentation.

2. **`processStructuredOutputRequest` drops token usage headers** — The PR replaces `request.call().chatResponse()` with `callWithObservability(request, exchange)` and removes the `populateTokenUsage(response, exchange)` call. While `callWithObservability` records tokens on the span, it does NOT populate exchange headers (`INPUT_TOKEN_COUNT`, `OUTPUT_TOKEN_COUNT`, `TOTAL_TOKEN_COUNT`). Compare with the entity path which calls both `recordObservationSuccess` AND `populateTokenUsage`. The default chat path works because `populateResponse()` includes `populateTokenUsage()`. This is a user-visible regression.

3. **FQCN violation** — `org.springframework.ai.chat.client.ResponseEntity<ChatResponse, T>` is used inline in `processEntityRequest` instead of adding an import. Per project conventions: "Do NOT use fully qualified class names."

### Suggestions (non-blocking)

4. **Missing warning log for unconfigured AI observability properties** — Every other configuration group in `BaseMainSupport` has a warning block for unconsumed properties. The PR doesn't add one.

5. **Fragile private field access in `extractChatModelFromClient`** — Uses `getDeclaredField("chatModel")` + `setAccessible(true)` on Spring AI internals. The graceful fallback is good, but worth noting as a maintenance risk.

---

Note: This review covers project rules, conventions, and code patterns. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

core/camel-main/src/main/java/org/apache/camel/main/BaseMainSupport.java
atiaomar1978-hub: _AI-generated on behalf of atiaomar1978-hub_

Fixed in e712116. Renamed the global toggle to `camel.aiObservability.enabled` (single-level camelCase under `camel.`, matching `camel.errorRegistry.` / `camel.routeController.`). Dash-style `camel.ai-observability.enabled` is accepted via Camel Main property normalization and is also resolved directly in `GenAiObservability.isEnabled`. Updated `BaseMainSupport`, `PrepareCamelMainMojo`, `GenAiObservabilityProperties`, generated catalog/metadata, docs, and added `AiObservabilityConfigurationPropertiesTest` coverage for the dash alias.

atiaomar1978-hub:

## davsclaus review addressed (commit `e712116`)

All three inline review comments are fixed and pushed:

1. **Property prefix** — Renamed to `camel.aiObservability.enabled` (with `camel.ai-observability.enabled` alias). Catalog/metadata, docs, upgrade guide, and Main auto-configuration updated.
2. **Structured output token headers** — Restored `populateTokenUsage()` after `callWithObservability()` in `processStructuredOutputRequest`.
3. **FQCN** — Added `ResponseEntity` import and removed the fully qualified type name.

**Bugbot follow-up** (same commit):
- `hasAiObservabilityConfiguration()` now gates local-property sync so `configure().ai()` alone does not override observability from lower-precedence property sources.
- `GenAiObservability.isEnabled` resolves both `camel.aiObservability.enabled` and `camel.ai-observability.enabled`.

**Tests** (local, green):
`GenAiModelResolverTest`, `SpringAiChatObservabilityTest`, `AiObservabilityConfigurationPropertiesTest`, `GenAiObservabilityTest`, `GenAiObservabilitySpanTest`

Ready for re-review when CI is green.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
atiaomar1978-hub: _AI-generated on behalf of atiaomar1978-hub_

Fixed in e712116. Restored `populateTokenUsage(response, exchange)` immediately after `callWithObservability()` in `processStructuredOutputRequest`, so structured output routes again populate `INPUT_TOKEN_COUNT`, `OUTPUT_TOKEN_COUNT`, and `TOTAL_TOKEN_COUNT` exchange headers.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
atiaomar1978-hub: _AI-generated on behalf of atiaomar1978-hub_

Fixed in e712116. Added `import org.springframework.ai.chat.client.ResponseEntity;` and replaced the FQCN with `ResponseEntity<ChatResponse, T>`.

components/camel-ai/camel-ai-observability-api/src/main/java/org/apache/camel/component/ai/observability/GenAiObservabilityProperties.java
```diff
@@ -24,7 +24,12 @@ public final class GenAiObservabilityProperties {
     /**
      * Global property to enable or disable GenAI observability across all AI components. Default is {@code true}.
      */
-    public static final String ENABLED = "camel.ai.observability.enabled";
+    public static final String ENABLED = "camel.aiObservability.enabled";
+
+    /**
+     * Dash-style alias for {@link #ENABLED}, accepted in application.properties and resolved by Camel Main.
+     */
+    public static final String ENABLED_DASH = "camel.ai-observability.enabled";
```
davsclaus: This dash-style alias is read directly via `PropertiesComponent.resolveProperty()` (see `GenAiObservability.isEnabled()`), bypassing Camel Main's `setPropertiesOnTarget`/`validateOptionAndValue` path that the camelCase `ENABLED` key goes through. That means a typo in this spelling is silently ignored instead of failing fast, and there's no precedent for a dash-case alias elsewhere in `camel-main`'s config groups. Please drop this alias or add explicit unknown-option validation for it.

components/camel-ai/camel-ai-observability-api/src/main/java/org/apache/camel/component/ai/observability/GenAiObservability.java
```diff
@@ -52,6 +52,9 @@ public static boolean isEnabled(CamelContext camelContext) {
         }
         Optional<String> property
                 = camelContext.getPropertiesComponent().resolveProperty(GenAiObservabilityProperties.ENABLED);
+        if (property.isEmpty()) {
+            property = camelContext.getPropertiesComponent().resolveProperty(GenAiObservabilityProperties.ENABLED_DASH);
```
davsclaus: Same concern as the `ENABLED_DASH` definition: this fallback lookup has no typo/unknown-key protection since it's a raw property resolution outside Main's configurer-validation path.

components/camel-ai/camel-ai-observability-api/src/main/java/org/apache/camel/component/ai/observability/GenAiModelResolver.java
```diff
@@ -16,81 +16,301 @@
...
+            return packageName;
+        }
+        return null;
+    }
+
+    private static boolean markVisited(Object model, Set<Integer> visited) {
+        return visited.add(System.identityHashCode(model));
+    }
+
+    private static boolean isLangChain4jPresent() {
```
davsclaus: `Class.forName` is called here using the resolver's own classloader rather than Camel's `ClassResolver` SPI (`camelContext.getClassResolver()`), which is the convention used elsewhere in the codebase for classloading that needs to work under OSGi/modular runtimes. The result is also cached in a single `volatile Boolean`, which works for this one presence check, but the pattern below (`isInstanceOf`/`invokeNoArg`) repeats uncached `Class.forName`/reflective method lookups on every call.

components/camel-ai/camel-ai-observability-api/src/main/java/org/apache/camel/component/ai/observability/GenAiModelResolver.java
```diff
@@ -16,81 +16,301 @@
...
+
+    private static Object invokeNoArgOptional(Object target, String methodName) throws ReflectiveOperationException {
+        try {
+            return invokeNoArg(target, methodName);
+        } catch (NoSuchMethodException e) {
+            return null;
+        }
+    }
+
+    private static Object invokeNoArg(Object target, String methodName) throws ReflectiveOperationException {
```
davsclaus: `getClass().getMethod(methodName).invoke(target)` is looked up fresh on every call with no caching of the resolved `Method`. Since this runs on every chat/embedding call (a hot path for high-throughput AI routes), please cache resolved `Method` objects (e.g., keyed by `Class<?>` + method name) rather than re-resolving via reflection each time.

components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat/src/main/java/org/apache/camel/component/springai/chat/SpringAiChatProducer.java
```diff
@@ -1202,11 +1225,81 @@ private void executeRequest(ChatClient.ChatClientRequestSpec request, Exchange e
...
+            return observabilityChatModel;
+        }
+        ChatModel chatModel = getEndpoint().getConfiguration().getChatModel();
+        if (chatModel != null) {
+            return chatModel;
+        }
+        return chatClient;
+    }
+
+    private ChatModel extractChatModelFromClient(ChatClient client) {
```
davsclaus, line 1299: Non-blocking: this reaches into `ChatClient.ChatClientRequestSpec`'s private `chatModel` field via reflection, which isn't public Spring AI API. If a future Spring AI version renames/restructures this field, observability silently degrades to `system=unknown` for chatClient-only configurations -- the failure is only logged at DEBUG. Consider a one-time WARN so operators notice after a Spring AI upgrade.

davsclaus (changes requested):
Nice, well-tested extension of the GenAI observability work (Phase 1/2 -> Phase 3). The Spring AI chat producer coverage (chatModel-only, chatClient-only, entity conversion, disabled-observability) is solid, and the doc/upgrade-guide updates are consistent.

Two things I'd like addressed before merge, plus a couple of lower-severity observations:

1. **`camel.ai-observability.enabled` bypasses Camel Main's option validation.** Every other `camel.*` Main property (including the camelCase `camel.aiObservability.*` group added in this PR) is routed through `setPropertiesOnTarget`/`validateOptionAndValue`, so a typo fails fast when `camel.main.autoConfigurationFailFast` is enabled. The dash-alias is read via a direct `PropertiesComponent.resolveProperty()` lookup instead, so it never goes through that mechanism -- a typo like `camel.ai-observability.enable` or `camel.ai-observabilty.enabled` is silently ignored with no warning. There's also no precedent for a dash-case alias in any other `camel-main` config group. Please either drop the alias (document only the camelCase form, consistent with the rest of Main) or add explicit validation so an unrecognized `camel.ai-observability.*`/`camel.ai-*` key doesn't fail silently.

2. **Reflection in `GenAiModelResolver` should be cached and go through Camel's `ClassResolver`.** `isLangChain4jPresent()`/`isInstanceOf()`/`invokeNoArg()` call `Class.forName(...)` and `getClass().getMethod(...).invoke(...)` on every resolution, uncached, and use the resolver's own classloader directly rather than Camel's `ClassResolver` SPI (`camelContext.getClassResolver()`), which is what the rest of the codebase uses for classloading that needs to work under OSGi/modular runtimes. Please cache the resolved `Class`/`Method` objects (e.g., in the existing `ConcurrentHashMap`-style pattern used elsewhere) and route class resolution through `ClassResolver` where a `CamelContext` is reasonably available to the call chain.

Lower-severity, non-blocking observations:

- `SpringAiChatProducer.extractChatModelFromClient()` reflects into `ChatClient.ChatClientRequestSpec`'s private `chatModel` field, which isn't public Spring AI API. A future Spring AI refactor could silently break this; the catch block only logs at DEBUG, so operators would see spans quietly degrade to `system=unknown` with no signal. Consider a one-time WARN on failure.
- `resolveObservabilityModelSource()` re-reads `getEndpoint().getConfiguration().getChatModel()` even though the same value is already cached into `observabilityChatModel` in `doStart()` -- harmless, but the second read can never actually diverge from the cached field.

Confirmed non-issues I checked while reviewing: the `camel.ai.observability.enabled` -> `camel.aiObservability.enabled` rename and the `GenAiModelResolver` public-signature changes are not compatibility breaks -- that code was only introduced in Phase 1 eleven days ago and has never shipped in a GA release (main is still `4.23.0-SNAPSHOT`). The `langchain4j-core` dependency in `camel-ai-observability-api`'s POM is `optional`, so the reflective rewrite genuinely achieves its stated goal of not forcing `camel-spring-ai-chat` to pull in `langchain4j-core`.

This review does not replace CodeRabbit/Sourcery or SonarCloud static analysis.

atiaomar1978-hub:
Addressed the latest review feedback from @davsclaus in commits c83ec16 and 4e75e0c:

**Review fixes (c83ec16)**
1. Removed dash-style `camel.ai-observability.enabled` alias and fallback resolution — only `camel.aiObservability.enabled` is supported now.
2. Refactored `GenAiModelResolver` to resolve LangChain4j types via Camel `ClassResolver` (passed from all AI producers) with cached `Method` lookups.
3. Added one-time WARN in `SpringAiChatProducer` when reflection into `ChatClient` fails.

**CI fix (4e75e0c)**
Fixed `LangChain4jEmbeddingsObservabilityTest` failure in `camel-langchain4j-embeddings`:
- Use the model class loader with `ClassResolver` for type checks
- Call `setAccessible(true)` for non-public model implementations
- Resolve embedding model names via `defaultRequestParameters()` before `modelName()`

**Tests run locally (all green):**
```
./mvnw -pl components/camel-ai/camel-ai-observability-api,components/camel-ai/camel-ai-observability,components/camel-ai/camel-langchain4j-embeddings,components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat,core/camel-main -am test \
  -Dtest=GenAiModelResolverTest,LangChain4jEmbeddingsObservabilityTest,SpringAiChatObservabilityTest,AiObservabilityConfigurationPropertiesTest,GenAiObservabilityTest
```

atiaomar1978-hub:
Fixed the CI **uncommitted changes** failure (run 32750624127):

1. **Catalog doc sync** — aligned `catalog/.../ai-observability.adoc` with the source doc (Global toggle text on one line).
2. **Formatting** — applied OpenRewrite line wrap in `LangChain4jChatProducer.buildObservationContext()`.
3. **Rebased** onto latest `main`.

Verified locally: rebuild of affected modules leaves a clean working tree; observability tests pass.

_Comment posted by Cursor on behalf of atiaomar1978-hub_

davsclaus (approved):

davsclaus:
Is the Java DSL the old style `ai().observability()` if so then it should be changed to `aiObservability()`

davsclaus (changes requested):
Is the Java DSL the old style `ai().observability()` if so then it should be changed to `aiObservability()`

catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/docs/ai-observability.adoc
```diff
@@ -15,16 +15,19 @@ OpenTelemetry semantic conventions for GenAI (stable subset aligned with Spring
...
-emit child spans and metrics per LLM call with attributes such as `gen_ai.operation.name`, `gen_ai.system`,
+When `camel-opentelemetry2` and/or `camel-micrometer` is on the classpath, langchain4j, OpenAI, and Spring AI chat
+producers emit child spans and metrics per LLM call with attributes such as `gen_ai.operation.name`, `gen_ai.system`,
 `gen_ai.request.model`, `gen_ai.usage.input_tokens`, and `gen_ai.usage.output_tokens`.
 
-Global toggle: set `camel.ai.observability.enabled=false` to disable GenAI observability across all AI
-components (default is `true` when a tracing or metrics backend is present).
+Global toggle: set `camel.aiObservability.enabled=false` (default is `true` when a tracing or metrics backend is present). Camel Main exposes the same setting
+via `camel.aiObservability.enabled` in application.properties or programmatically with
+`main.configure().ai().observability().withEnabled(false)`.
```
davsclaus: Look here is what I spotted which is wrong ^^^^

atiaomar1978-hub:

Addressed @davsclaus latest review (commit `32956dfb`):

**Java DSL rename**
- Replaced nested `main.configure().ai().observability()` with `main.configure().aiObservability()`, consistent with other Main config groups (`errorRegistryConfig()`, `routeControllerConfig()`, etc.).
- Removed the `AiConfigurationProperties` wrapper; `AiObservabilityConfigurationProperties` now attaches directly to `MainConfigurationProperties`.
- Updated docs, upgrade guide, and regenerated catalog metadata.

**Tests run locally (green):**
```
./mvnw -pl core/camel-main,components/camel-ai/camel-ai-observability-api,components/camel-ai/camel-ai-observability,components/camel-ai/camel-langchain4j-embeddings,components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat,catalog/camel-catalog -am test \
  -Dtest=AiObservabilityConfigurationPropertiesTest,GenAiModelResolverTest,LangChain4jEmbeddingsObservabilityTest,SpringAiChatObservabilityTest,GenAiObservabilityTest
```

Ready for re-review when CI is green.

atiaomar1978-hub:
Merged latest `main` into this branch to fix the CI **"Fail if there are uncommitted changes"** failure on `build (25, false)`.

The branch was behind `main`, which already contains the SnakeYAML Javadoc formatting fix (`174f902`). After merging, `git status` is clean post-build locally, and the observability + `camel-main` test suites pass.

Please approve/re-run CI when ready.

_AI-generated comment on behalf of atiaomar1978-hub._

atiaomar1978-hub:

Checked CI for [run 32870148314](https://github.com/apache/camel/actions/runs/32870148314) — **the failure is not caused by this PR**.

### What failed (unrelated)
- **camel-lumberjack** integration test: \`LumberjackMultiThreadIT.shouldListenToMessages\`
- Failed on all rerun attempts in the incremental reactor
- This PR does not touch \`camel-lumberjack\`

### What passed (PR-related)
- **camel-ai-observability** — SUCCESS
- **camel-ai-observability-api** — SUCCESS
- **camel-main** (incl. \`AiObservabilityConfigurationPropertiesTest\`) — SUCCESS
- **camel-spring-ai-chat** (incl. \`SpringAiChatObservabilityTest\`) — SUCCESS
- **camel-snakeyaml** / uncommitted-changes check — SUCCESS
- PR doc validation, dependency-review — green

### Action taken
Rebased onto latest \`main\` (commit \`e939c900\`) and pushed to trigger a fresh CI run. Local tests for observability + \`camel-main\` + \`camel-spring-ai-chat\`: green.

If the new run fails again on an unrelated module (lumberjack/nats/etc.), a CI re-run should suffice — no code changes needed in this PR.

davsclaus (approved):

gnodet (commented):
Phase 3 of GenAI observability adds Spring AI chat support and camel-main configuration. The architecture (decoupling GenAiModelResolver from langchain4j via reflection) is sound. A few observations after review:

**1. Orphan service file (should fix):**
The PR creates `META-INF/services/org/apache/camel/configurer/org.apache.camel.main.AiConfigurationProperties` referencing `AiConfigurationPropertiesConfigurer`, but neither class exists anywhere in the PR or codebase. This appears to be a leftover from a rename during development (the actual class is `AiObservabilityConfigurationProperties`). It won't cause a runtime crash (Camel's configurer lookup is by target class name, and since `AiConfigurationProperties` can't be instantiated, the file will never be loaded), but it should be removed as dead code.

**2. Fragile private-field reflection (design concern):**
`extractChatModelFromClient` reflectively accesses a private `chatModel` field from Spring AI's internal `DefaultChatClient$DefaultChatClientRequestSpec`. While it degrades gracefully (warns once, returns null), this will silently lose observability model metadata if Spring AI restructures its internals. Consider documenting this limitation or requiring explicit `chatModel` configuration when observability is enabled.

**3. Static METHOD_CACHE classloader concern (low priority):**
The `METHOD_CACHE` stores `Class<?>` references in a static map, which can prevent classloader GC in environments with classloader recycling (OSGi, Spring Boot devtools). Low practical impact for this module but a known anti-pattern worth noting.

**4. String matching in mapLangChain4jProvider (informational):**
The switch from enum to string matching is a deliberate trade-off for reflective resolution without langchain4j dependency. Worth noting that typos in string literals would silently fall through to UNKNOWN.

davsclaus:
@atiaomar1978-hub it seems 1 from @gnodet review about must be fixed and then take a look at the others

atiaomar1978-hub:
## gnodet review feedback addressed

**GitHub auth:** authenticated as \`atiaomar1978-hub\`

**Commit:** \`dc007256883\` — _CAMEL-23861: Remove orphan AiConfigurationProperties service file_

### Fix (@gnodet item #1 — must fix)

Removed the orphan service descriptor:
\`core/camel-main/src/generated/resources/META-INF/services/org/apache/camel/configurer/org.apache.camel.main.AiConfigurationProperties\`

This referenced non-existent \`AiConfigurationPropertiesConfigurer\` — a leftover from the rename to \`AiObservabilityConfigurationProperties\`. The correct descriptor (\`org.apache.camel.main.AiObservabilityConfigurationProperties\`) remains in place.

Verified the build does not regenerate the orphan file.

### Other gnodet observations (not changed)

- **#2 Fragile private-field reflection** — already degrades gracefully with one-time WARN in \`extractChatModelFromClient\`
- **#3 Static METHOD_CACHE classloader** — low priority; acceptable for this module
- **#4 String matching in mapLangChain4jProvider** — deliberate trade-off for reflective resolution without langchain4j dependency

### Tests (local)

\`\`\`bash
./mvnw -pl components/camel-ai/camel-ai-observability,components/camel-ai/camel-ai-observability-api,core/camel-main,components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat -am test
\`\`\`

**Result:** green (~13 min)

Ready for re-review.

---
_Composer on behalf of atiaomar1978-hub (AI-generated)_

atiaomar1978-hub:
## All @gnodet review items addressed

**Commit:** \`0564fbfdbbc\` — _CAMEL-23861: Address all gnodet review feedback on GenAI observability_

### #1 Orphan service file (previously fixed)
Removed dead \`org.apache.camel.main.AiConfigurationProperties\` META-INF descriptor in \`dc007256883\`.

### #2 Fragile private-field reflection
- Documented the \`chatClient\`-only limitation in \`ai-observability.adoc\` and \`spring-ai-chat-component.adoc\`
- Added a one-time startup **WARN** when GenAI observability is enabled but \`ChatModel\` cannot be resolved from \`ChatClient\`, recommending explicit \`chatModel\` configuration
- Clarified in javadoc that \`extractChatModelFromClient\` relies on Spring AI internal API

### #3 Static METHOD_CACHE classloader concern
- Cache keys now use declaring **class name** strings instead of \`Class<?>\` references
- Cached \`Method\` instances stored as \`SoftReference\` values so entries can be reclaimed under memory pressure / classloader recycling
- Re-validates \`Method.getDeclaringClass()\` on cache hit for classloader reload safety

### #4 String matching in mapLangChain4jProvider
- Added named \`LC4J_PROVIDER_*\` constants aligned with LangChain4j \`ModelProvider\` enum names
- Added tests for Anthropic mapping and unrecognized provider → \`unknown\`

### Tests (local)
\`\`\`bash
./mvnw -pl components/camel-ai/camel-ai-observability,components/camel-ai/camel-ai-observability-api,components/camel-spring-parent/camel-spring-ai/camel-spring-ai-chat -am test -Dtest=GenAiModelResolverTest,SpringAiChatObservabilityTest
\`\`\`
**Result:** green

---
_Composer on behalf of atiaomar1978-hub (AI-generated)_

davsclaus (approved):

atiaomar1978-hub:
_AI-generated on behalf of atiaomar1978-hub._

Fixed CI failures on PR doc validation and build xref-check.

**Root cause:** `spring-ai-chat-component.adoc` used `xref:ai-observability.adoc` which does not resolve from the components ROOT module — the page lives under the `others` module.

**Fix:** Changed to `xref:others:ai-observability.adoc[AI Observability]` in commit `1b0350ff273`.

This resolves both:
- PR doc validation (Antora: *target of xref not found: ai-observability.adoc*)
- Build and test (`docs` module xref-check during regen)

davsclaus:
@atiaomar1978-hub this has a merge conflict, do you mind fixing and rebase if possible

atiaomar1978-hub:
_AI-generated on behalf of atiaomar1978-hub._

Rebased onto latest `main` and resolved the merge conflict.

**Conflict resolved:** `camel-langchain4j-tools` was removed on `main` (commit `2a2b0e1bb71`); accepted deletion of `LangChain4jToolsProducer.java` since the module no longer exists. Observability changes for LangChain4j remain in `langchain4j-chat`, `langchain4j-agent`, and `langchain4j-embeddings`.

**Also included from prior CI fixes:**
- `xref:others:ai-observability.adoc` fix for PR doc validation (`1b0350ff273`)
- Catalog doc regen for uncommitted changes check (`85172d4a97e`)

Branch is now **mergeable** with `main`. @davsclaus approval still stands — waiting on CI.

atiaomar1978-hub:
@davsclaus Could you please recheck this PR when you have a moment?

The branch has been rebased onto latest `main` and the CI blockers from earlier runs are addressed:

- Fixed the broken `ai-observability` xref in `spring-ai-chat-component.adoc` (`xref:others:ai-observability.adoc`)
- Regenerated catalog docs so source and generated files are in sync (uncommitted-changes check)
- Rebased onto current `main` (includes QuickJS #25778); force-pushed to `c5f1b3c03b6`

The last full build passed doc validation and the uncommitted-changes step; the only failure was an unrelated flaky `camel-vertx-websocket` test. CI on the fork may also need workflow approval before it runs again.

Thanks!

davsclaus:
@atiaomar1978-hub merging now, you are welcome to look in camel-spring-boot if anything there needs to be added/updated

davsclaus:
and is there a phase 4 ?

PR #25509, opened 2026-08-14
CAMEL-23703: camel-launcher - fix brew audit --strict redundant version

ammachado:
# Description

`formula.rb.tpl` (the Homebrew formula template rendered by JReleaser for `camel-launcher`) set both a `url` and an explicit `version` field. Homebrew's `brew audit --strict` flags an explicit `version` as redundant whenever it can already be inferred by scanning the `url` (which it can here, since the URL embeds the Maven coordinate version). Any real submission of the generated formula to `homebrew-core` would fail this audit gate.

Fix: drop the redundant `version` line from the template.

`camel-validate.sh`'s Homebrew validation step previously read the *expected* post-install version back out of that same `version` field (needed because in CI's test-mode staging the formula's real POM version, e.g. a `-SNAPSHOT`, differs from the release-manager-facing resolved version). Removing the field regressed that check, so it now reads the expected version out of the formula's `url` instead, captured before the test-mode step rewrites that `url` to a local `file://` path for the offline install.

Found and fixed while running `camel-package.sh prepare` + `camel-validate.sh all` for the already-released `camel-4.22.0` version to verify the launcher's JReleaser release procedure end-to-end.

Verified locally: re-ran `camel-package.sh prepare --channel stable` + `camel-validate.sh all` in the project's own `CAMEL_PACKAGE_TEST_MODE` (the same mechanism `.github/workflows/package-native-validation.yml` uses in CI) — `brew audit --strict` now reports no offenses, and the post-install version check passes.

# Target

# Tracking

This change only touches a shell script and a Mustache template (no Java sources), so the Maven formatter/impsort step has nothing to reformat; the pre-push hook's `mvn formatter:format impsort:sort` ran clean with no changes. A full root build was not run given the change's scope.

# AI-assisted contributions

_Claude Code (Sonnet 5) on behalf of Adriano Machado (@ammachado)_

### DIFF
--- a/dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-validate.sh
+++ b/dsl/camel-jbang/camel-launcher/src/jreleaser/bin/camel-validate.sh
@@ -233,6 +233,18 @@ validate_homebrew() {
     # failure of the audit gate below caused by the tap copy itself, not by the formula.
     chmod 644 "$tap_dir/Formula/${fmla}.rb"
 
+    # The formula carries no dedicated `version` field (brew audit --strict rejects one that
+    # merely duplicates what it already scans from the url), so the expected post-install
+    # version is read from the url instead - captured now, before Step 2 below may rewrite that
+    # same url to a local file:// path for the offline install, which would otherwise wipe out
+    # the "<version>/<artifact>" path segment this parses.
+    local expected_version
+    expected_version=$(sed -n 's#.*/camel-launcher/\([^/]*\)/[^/]*"$#\1#p' "$tap_dir/Formula/${fmla}.rb" | head -n1)
+    if [ -z "$expected_version" ]; then
+        echo "WARN: could not read the version from tapped formula ${fmla}.rb's url, falling back to \$RESOLVED_VERSION ($RESOLVED_VERSION)"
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus (approved):

gnodet (approved):
Well-scoped fix that removes a redundant Homebrew `version` field (flagged by `brew audit --strict`) from the formula template and correctly adapts the validation script to extract the expected version from the URL instead, captured before test-mode URL rewriting can overwrite it. Two committer approvals already on record and all CI checks pass.

One minor observation:

- The sed regex at line 240 in `camel-validate.sh` anchors the closing `"` to end-of-line with `$`. If `brewDownloadStrategy` is ever enabled in the JReleaser config (currently unset), the rendered formula line would include `, :using => CurlDownloadStrategy` after the URL, causing the `$` anchor to fail silently. The fallback to `$RESOLVED_VERSION` means this isn't a bug today, but the regex could be made more robust by dropping the `$` anchor. Very minor and not blocking.

Overall the approach is sound: removing the `version` line from the template is the correct fix for `brew audit --strict`'s "redundant version" offense, and the comment explaining why the extraction must happen before Step 2 is clear and helpful.

PR #25510, opened 2026-08-15
CAMEL-22114: Fix PQC KeyStore tests on Java 25 (InvalidKeyException)

gnodet:
## Summary

- Fix `InvalidKeyException: unknown private key passed to ML-DSA` when using PQC KeyStore-based signing on Java 25
- Re-encode JDK-native PQC keys through Bouncy Castle's `KeyFactory` after loading from a KeyStore or user-supplied KeyPair, transparently converting them to BC types
- No-op on Java 17/21 where JKS always returns BC key instances

## Root Cause

Java 25 natively supports ML-DSA ([JEP 497](https://openjdk.org/jeps/497)). When a Bouncy Castle ML-DSA private key is stored in a JKS KeyStore and retrieved, the JDK's built-in ML-DSA `KeyFactory` deserializes it as a **JDK-native** ML-DSA key object. When this key is passed to BC's `Signature.getInstance("ML-DSA", "BC").initSign()`, BC does not recognize it:

```
java.security.InvalidKeyException: unknown private key passed to ML-DSA
    at org.bouncycastle.jcajce.provider.asymmetric.mldsa.SignatureSpi.signInit
    at PQCProducer.signature(PQCProducer.java:494)
```

SLH-DSA KeyStore tests **pass** because Java 25 does not (yet) provide a native SLH-DSA `KeyFactory`, so those keys remain as BC objects after deserialization.

This is a follow-up to the CAMEL-22114 fix (commit 91c6190), which addressed provider-less `Signature.getInstance()` calls. The KeyStore deserialization path was not covered by that fix and was exposed by the July 2026 PQC features (CAMEL-22538, CAMEL-23844, CAMEL-23847, CAMEL-23848, CAMEL-23849) that added KeyStore-based test cases.

## Evidence

CI runs [31861106711](https://github.com/apache/camel/actions/runs/31861106711) and [31848843987](https://github.com/apache/camel/actions/runs/31848843987) show `Camel :: PQC FAILURE [45.849 s]` on Java 25 while Java 17 passes. The `Main build` workflow only runs `regen.sh` (no tests), so this failure is latent on `main`.

## Test plan

- [x] All 211 camel-pqc tests pass on Java 21 (210 executed + 1 skipped)
- [x] `PQCSignatureOnlyKeyStoreTest` passes on Java 25 (CI)
- [x] `PQCSignatureWithKeyStoreTest` passes on Java 25 (CI)
- [x] `PQCKeyStoreJdk25KeyConversionTest` — dedicated regression test with `@EnabledForJreRange(min = JAVA_25)`, skipped on Java <25
- [x] No regressions on Java 17 (CI)

### DIFF
--- a/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
+++ b/components/camel-pqc/src/main/java/org/apache/camel/component/pqc/PQCProducer.java
@@ -21,6 +21,8 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.cert.Certificate;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -53,9 +55,11 @@ import org.apache.camel.util.SecureRandomHelper;
 import org.bouncycastle.jcajce.SecretKeyWithEncapsulation;
 import org.bouncycastle.jcajce.spec.KEMExtractSpec;
 import org.bouncycastle.jcajce.spec.KEMGenerateSpec;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.bouncycastle.pqc.jcajce.interfaces.LMSPrivateKey;
 import org.bouncycastle.pqc.jcajce.interfaces.XMSSMTPrivateKey;
 import org.bouncycastle.pqc.jcajce.interfaces.XMSSPrivateKey;
+import org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -401,6 +405,15 @@ public class PQCProducer extends DefaultProducer {
             keyPair = new KeyPair(publicKey, privateKey);
         } else {
             keyPair = getConfiguration().getKeyPair();
+        }
+
+        // On JDK 25+, a JKS KeyStore (or user-supplied KeyPair) may contain JDK-native PQC keys
+        // (e.g. ML-DSA, ML-KEM) that Bouncy Castle's Signature / KeyGenerator SPI does not recognise,
+        // causing InvalidKeyException at initSign / initVerify time. Re-encoding through BC's KeyFactory
+        // transparently converts JDK-native keys into the BC types the rest of the component expects,
+        // and is a no-op for keys that are already BC instances.
+        if (keyPair != null) {
+            keyPair = ensureBcKeyPair(keyPair);
         }
 
         // Initialize hybrid signature operations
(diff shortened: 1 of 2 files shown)


atiaomar1978-hub:
## Review

_AI-generated review on behalf of Omar Atie_

Thanks for the follow-up to CAMEL-22114 — the root-cause analysis is clear and matches what we see on Java 25: JKS deserialization of ML-DSA keys yields JDK-native `PrivateKey`/`PublicKey` instances that BC's `Signature` SPI rejects at `initSign`/`initVerify`.

### What looks good

- **Fix location** — Calling `ensureBcKeyPair()` in `doStart()` immediately after KeyStore / user-supplied `KeyPair` loading is the right hook. It covers sign, verify, KEM, hybrid PQC, and key-export paths that all reuse the same `keyPair` field.
- **Conversion approach** — Re-encoding through PKCS#8 / X.509 specs and BC `KeyFactory` is the standard interoperability pattern and is a no-op when keys are already BC instances.
- **Provider fallback** — `getBcKeyFactory()` trying `BouncyCastleProvider` first, then `BouncyCastlePQCProvider`, mirrors how the rest of the component resolves algorithms.
- **Failure handling** — Returning the original pair on conversion failure (with a debug log) is reasonable; the crypto operation will still fail with a meaningful exception rather than masking a misconfiguration silently at INFO/WARN level.
- **Scope** — Single-file, focused change with no API surface impact. CI is green on both Java 17 and Java 25.

### Minor nits (non-blocking)

1. **BC detection** — The `getClass().getName().startsWith("org.bouncycastle.")` check is pragmatic and fine here. If we ever hit a BC key loaded through a JDK proxy/wrapper, we'd fall through to re-encoding (still safe). Not worth blocking on.
2. **Test plan checkboxes** — The PR description still has the Java 25 KeyStore test items unchecked, but CI (`build (25, false)`) is green — worth ticking those off for completeness.
3. **Dedicated unit test** — No isolated test for `ensureBcKeyPair()` itself; coverage comes from the existing KeyStore ITs. Acceptable given the JDK-version-specific nature, though a small `@EnabledForJreRange(min = JRE.JAVA_25)` test could make future regressions easier to spot locally.

### Verdict

Looks good to merge once a committer approves. This closes the gap left by the earlier provider-less `Signature.getInstance()` fix for the KeyStore deserialization path exposed by the recent PQC KeyStore test additions.

gnodet:

Thank you for the thorough review!

Addressing your nits:

1. **Test plan checkboxes** — Updated and all checked ✓
2. **Dedicated unit test** — Added `PQCKeyStoreJdk25KeyConversionTest` with `@EnabledForJreRange(min = JRE.JAVA_25)` in commit 790f176. It exercises the ML-DSA sign+verify flow via JKS KeyStore and is automatically skipped on Java <25 where the conversion is a no-op. All 211 tests pass (210 executed + 1 skipped on Java 21).
3. **BC detection** — Acknowledged, the `startsWith("org.bouncycastle.")` check is intentionally pragmatic; worst case, a non-BC key falls through to re-encoding which is safe.

PR #25528, opened 2026-08-18
CAMEL-21313: Refactor camel-clickup tests for clarity and consistency

misrasuraj:
## Summary
- Refactor `camel-clickup` tests to keep test methods focused on assertions and move setup/lifecycle concerns into helpers.
- Standardize route URI construction by replacing concatenated `from(...)` strings with `fromF(...)` in the affected tests.
- Improve consistency by removing anonymous assertion blocks, reusing shared test support utilities, and migrating touched assertions to AssertJ.

## Test plan
- [x] `./mvnw install -Dquickly -pl components/camel-clickup -am`
- [x] `cd components/camel-clickup && ../../mvnw test`
- [x] `cd components/camel-clickup && ../../mvnw formatter:format impsort:sort`

_AI-generated by Cursor on behalf of misrasuraj._

### DIFF
--- a/components/camel-clickup/src/test/java/org/apache/camel/component/clickup/ClickUpWebhookCallTest.java
+++ b/components/camel-clickup/src/test/java/org/apache/camel/component/clickup/ClickUpWebhookCallTest.java
@@ -17,7 +17,10 @@
 package org.apache.camel.component.clickup;
 
 import java.io.InputStream;
-import java.util.*;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 
 import org.apache.camel.Exchange;
 import org.apache.camel.RoutesBuilder;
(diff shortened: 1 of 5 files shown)


components/camel-clickup/src/test/java/org/apache/camel/component/clickup/ClickUpWebhookCallTest.java
```diff
@@ -28,45 +31,35 @@
 import org.apache.camel.component.webhook.WebhookConfiguration;
 import org.apache.camel.component.webhook.WebhookEndpoint;
 import org.junit.jupiter.api.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class ClickUpWebhookCallTest extends ClickUpTestSupport {
+import static java.util.List.of;
```
davsclaus, line 35: Nit: this static import (and the bare `of(...)` call on line 39) doesn't appear anywhere else in the codebase and is a bit ambiguous at the call site (could be mistaken for `Set.of`, `Stream.of`, etc.). Consider using `List.of(...)` explicitly instead, for consistency with the rest of the project.

davsclaus (approved):
Automated rules-and-conventions review against this project's contribution guidelines (build/test/style standards and PR conventions). This does not replace CodeRabbit, Sourcery, SonarCloud, or other specialized static-analysis tooling.

This is a solid, well-targeted refactor that directly addresses the three specific complaints in CAMEL-21313 (setup/lifecycle moved into `ClickUpTestSupport` helpers, anonymous scoping blocks removed, `fromF(...)` replacing string concatenation). I traced through each of the 5 `@Test` methods across the 4 files and confirmed no assertions were weakened or dropped — the `assertInstanceOf(...)` → `assertThat(...).isNotNull()` simplifications are behaviorally equivalent, since `MAPPER.readValue(..., WebhookCreationCommand.class)` already guarantees the return type via generics. No leftover unused imports or field-shadowing after moving constants to the shared base class.

**Non-blocking notes:**
- JIRA CAMEL-21313 is currently unassigned and still `Open`, even though this is an AI-generated PR against it. Per this project's AI-agent guidelines, the ticket should be assigned to the contributor and moved to "In Progress" before/along with opening the PR (also reduces the risk of someone else picking up the same ticket in parallel).
- Minor style nit inline: `import static java.util.List.of;` + bare `of(...)` doesn't appear anywhere else in this codebase and reads ambiguously at the call site — `List.of(...)` would be more consistent.

Neither point affects correctness, so approving.

